Issue #3491, opened 2020-01-02
Custom 404 page - "cannot mix legacy routes with new keys"


iinfin:
Trying to set a custom 404 page as shown [here ](https://khrome.dev/blog/custom-404-pages-on-zeit-now-v2-platform) and [here](https://spectrum.chat/zeit/now/custom-404-page~1f921045-60d9-477b-b23c-1626c9fa1565), getting this:

> Deployment failed with the following error:
> ```
>     Cannot mix legacy routes with new keys:
> ```

I'm using Hugo static generator, and here's the 404 routing part in my `now.json`
```
"routes": [{
	"src": "/.*",
	"dest": "/$1"
}, {
	"handle": "filesystem"
}, {
	"src": "/.*",
	"status": 404,
	"dest": "/404"
}
```
I also tried: 
```
}, {
	"src": "/(.*)",
	...
```

I've looked around for other examples, maybe missing a difference between now v1/2?
Thanks in advance

jasondibenedetto:
Having a similar issue, unsure how to redirect all 404 pages to the home page. Previously used the following solution...

```
"routes": [
  { "handle": "filesystem" },
  { "src": "/(.*)", "status": 404, "dest": "/index.html" }
]
```

Now that routes is deprecated unsure what the implementation should be.

nickgrealy:
I had the same issue, dug into the code (it's very readable) - turns out this is symptomatic of another underlying issue (some config is not inter-compatible). Check here -> https://github.com/zeit/now/blob/9d0b0c6e410343eead5e43f650c942d4b301f84a/packages/now-routing-utils/src/index.ts#L187-L211

To fix, I removed `cleanUrls`.

ivan-kleshnin:
> Now that routes is deprecated

Since when does "Deprecated" == "Suddenly stops working"?

I assume Zeit team will release breaking changes under `version: 3` in `now.json`.

I also see some bugs with 404 handling though... 

`"status": 404` doesn't seem to apply with `now dev`.

cryptiklemur:
routes hasn't "suddenly stop[ped] working"

ivan-kleshnin:
@aequasi just some examples of serious bugs related to routes: 1) https://github.com/ivan-kleshnin/zeit-routes 2) https://github.com/zeit/now/issues/3659 And there are many more (look at the Issue threads at Next / Now GH and Spectrum).

styfle:
Hello,

[...]

PR #4498, opened 2020-05-27
[routing-utils] Improve error messages

styfle:
This PR improves the way we handle routing errors in a few ways:

- The error response is a single error (the first) instead of an array of errors when mixing routing properties
- The error message says which route index has the error
- The error includes `link` and `action` properties to match our API
- The error message for mixed routes with new routing properties has been updated per [ch341](https://app.clubhouse.io/vercel/story/341)

Related to #3491


packages/now-routing-utils/src/index.ts
```diff
@@ -100,79 +92,85 @@ export function normalizeRoutes(inputRoutes: Route[] | null): NormalizedRoutes {
...
-  const error = createNowError(
-    'invalid_routes',
-    'One or more invalid routes were found',
-    errors
-  );
+  const error =
+    errors.length > 0
+      ? createError(
+          'invalid_route',
+          errors[0],
```
ijjk: Why do we only return the first error here? Won't this make debugging harder?

styfle: We don't have a good support for multi-line errors right now.

Today, this error could be displayed in 4+ places:

- cli (works with multiline)
- git integration (does not - truncated to first line)
- build logs (kinda of works, not highlighted properly)
- summary at the top left (does not work - newlines are not preserved)

styfle: The other reason is that each error could have a separate link for the "Learn More"

ijjk: Could we provide the full errors to the environments that do have multi-line support and then provide the first error to the environments that don't?

styfle: No because the builds logs and the summary at the top share the same `error.message` value



It is unlikely that a user will have multiple routes with errors since the one in the screenshot has been combined into a single error anyway.

ijjk: I see, if a user does have multiple routes with errors though it seems like it would be a pretty painful experience to fix one re-deploy and have the next one show and continue for however many incorrect routes they do have. If we could avoid this I think it would be ideal

packages/now-cli/test/dev/integration.js
```diff
@@ -573,7 +573,11 @@ test('[vercel dev] validate mixed routes and rewrites', async t => {
   const output = await exec(directory);
 
   t.is(output.exitCode, 1, formatOutput(output));
-  t.regex(output.stderr, /Cannot define both `routes` and `rewrites`/m);
+  t.regex(
+    output.stderr,
+    /If `rewrites`, `redirects`, `headers`, `cleanUrls` or `trailingSlash` is used, `routes` cannot be present./m
```
TooTallNate: I realize this isn't the place to fix it, but this error message has some grammar errors:

```suggestion
    /If `rewrites`, `redirects`, `headers`, `cleanUrls` or `trailingSlash` are used, then `routes` cannot be present./m
```

packages/now-routing-utils/src/index.ts
styfle: The user doesn't need to deploy, it will be caught with `vercel dev` too.

Anyway, I went ahead and updated this line so the nested errors are capture too.

TooTallNate (commented):
I'm noticing that we're inconsistent with using backticks vs. double-quotes when referencing property names in the error messages. Do you think we should normalize them?

styfle:
~~Sure, can you comment on the lines that are wrong?~~

Ok I see what you mean. I'll update where appropriate.

styfle:
@TooTallNate Ok take another look and let me know if you see anything I missed 👍

dav-is (approved):

TooTallNate (approved):

PR #4444, opened 2020-05-21
[build-utils] Fix Yarn 2 install argument order

SagnikPradhan:
Fix for Yarn Version 2 install error. Positioned the arguments array in proper order #4436

PR #4508, opened 2020-05-27
[build-utils] Remove `--ignore-engines` flag

styfle:
Follow up to #4444 that makes sure we run Next.js tests.

This `--ignore-engines` flag was originally added in https://github.com/vercel/now-builders/pull/463 as a workaround. We can remove it to make sure the version of Node selected will work with the dependencies.

Removing the flag also makes sure that Yarn 2 will work properly, see #4444.


TooTallNate (dismissed):

ijjk (dismissed):

TooTallNate:
Should this be considered a breaking change?

ijjk (approved):

styfle:
@TooTallNate Sure, I added the semver-major label

TooTallNate (approved):

PR #4514, opened 2020-05-28
[node] Pass in the `tsconfig.json` file to `startDevServer()`

TooTallNate:
So that if a `tsconfig.json` exists closer to the entrypoint file,
then that config file will be correctly used (rather than say, the
root-level `tsconfig.json` file, which may be specific to the frontend
configuration in Next.js for example).


packages/now-node/src/index.ts
```diff
@@ -430,6 +430,14 @@ export async function startDevServer(
   opts: StartDevServerOptions
 ): Promise<StartDevServerResult> {
   const { entrypoint, workPath, config, meta = {} } = opts;
+
+  // Find the `tsconfig.json` file closest to the entrypoint file
+  const projectTsConfig = await walkParentDirs({
+    base: workPath,
+    start: join(workPath, dirname(entrypoint)),
+    filename: 'tsconfig.json',
+  });
```
styfle, line 439: Let's add a test for this one. We already have one in `@now/node` but not in dev https://github.com/vercel/vercel/blob/be922796777a45cce5f075c7d73980e570058aef/packages/now-node/test/fixtures/18-nested-tsconfig/now.json

TooTallNate: f55560e

packages/now-cli/test/dev/integration.js
```diff
@@ -1347,3 +1347,15 @@ test(
...
   })
 );
+
+test(
+  '[vercel dev] Should work with nested `tsconfig.json` files',
+  testFixtureStdio(
+    'nested-tsconfig',
+    async testPath => {
+      await testPath(200, `/`, /Nested tsconfig.json test page/);
+      await testPath(200, `/api`, 'Nested `tsconfig.json` API endpoint');
```
styfle: How does this ensure the `api/tsconfig.json` file was used?

packages/now-cli/test/dev/integration.js
```diff
@@ -1347,3 +1347,15 @@ test(
...
+    },
+    { skipDeploy: true }
```
styfle: Why skipDeploy? Seems like it should also test production to make sure it works the same as dev?

packages/now-cli/test/dev/integration.js
TooTallNate: Because the root level `tsconfig.json` causes `ts-node` to have syntax errors (I believe due to `module: 'esnext'`). The fact that the API endpoint runs successfully is indicative that the proper tsconfig was used.

packages/now-cli/test/dev/integration.js
TooTallNate: Sure, updated in 5a3544164ceb65e0aa1f3a4d0a930bcba7b41e1e.

styfle (approved):

lucleray (approved):

PR #4531, opened 2020-05-29
[cli] Update zeit.ink to vercel.link shortener

styfle:
- Update `zeit.ink` to `vercel.link` shortener
- Support for `output.error()` action parameter

### DIFF
--- a/packages/now-cli/src/commands/deploy/latest.js
+++ b/packages/now-cli/src/commands/deploy/latest.js
@@ -578,7 +578,8 @@ export default async function main(
         deployment.message ||
           'An unexpected error occurred while deploying your project',
         null,
-        'https://vercel.link/help'
+        'https://vercel.link/help',
+        'Contact Support'
       );
       return 1;
     }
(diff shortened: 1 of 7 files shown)


packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -251,7 +251,9 @@ export default async function main(
   if (argv['--name']) {
     output.print(
       `${prependEmoji(
-        `The ${param('--name')} flag is deprecated (https://zeit.ink/1B)`,
+        `The ${param(
+          '--name'
+        )} flag is deprecated (https://vercel.link/name-flag)`,
```
TooTallNate: Can we call it "option" instead of "flag"? This would match the documentation in the link.

styfle: Done 👍

TooTallNate (approved):

lucleray (approved):

PR #4543, opened 2020-05-31
[cli] Throw an error if both `.vercel` and `.now` dirs exist

TooTallNate:


lucleray:
Is there a code path where we could create `.vercel` and `.now`? In other words, how did users end up with `.vercel` and `.now`?

styfle (approved):

TooTallNate:
@lucleray There only way that I can think of that this situation could happen with our tooling is if someone linked their project using a modern CLI (sees no `.now` dir, and creates `.vercel` dir). And _then_ the user linked their project again with an older CLI (before `.vercel` was supported, so it creates `.now` dir since it doesn't yet exist). And _then_ the user uses CLI after this PR and sees this error message.

I think a more likely scenario would be that the user copied the directory manually.

Either way, this particular error should be very rare (less so than conflicting `now.json`/`vercel.json` files, for example).

lucleray (approved):

PR #4547, opened 2020-06-01
[cli] Remove unnecessary build outputs

TooTallNate:
Upon investigation into the `dist` dir, it appears that `ncc` is bundling some assets that don't need to be there. This change is a quick band-aid fix to remove those assets, without addressing the underlying cause of _why_ they're being bundled, which requires further investigation.

Overall about 1mb of disk space is saved.


lucleray (dismissed):

styfle (commented):
You might try `declaration: false` in `./packages/now-cli/tsconfig.json`

packages/now-cli/scripts/build.ts
```diff
@@ -75,6 +75,17 @@ async function main() {
   const dest = join(dirRoot, 'dist/runtimes');
   await cpy('**/*', dest, { parents: true, cwd: runtimes });
 
+  // Band-aid to delete stuff that `ncc` bundles, but it shouldn't:
+
+  // TypeScript definition files from `@vercel/build-utils`
+  await remove(join(dirRoot, 'dist/dist'));
```
styfle:
```suggestion
  await remove(join(dirRoot, 'dist', 'dist'));
```

packages/now-cli/scripts/build.ts
```diff
@@ -75,6 +75,17 @@ async function main() {
...
+
+  // The Readme and `package.json` from "config-chain" module
+  await remove(join(dirRoot, 'dist/config-chain'));
```
styfle:
```suggestion
  await remove(join(dirRoot, 'dist', 'config-chain'));
```

packages/now-cli/scripts/build.ts
```diff
@@ -75,6 +75,17 @@ async function main() {
...
+
+  // A bunch of source `.ts` files from CLI's `util` directory
+  await remove(join(dirRoot, 'dist/util'));
```
styfle:
```suggestion
  await remove(join(dirRoot, 'dist', 'util'));
```

TooTallNate:
@styfle I applied your suggestions, but note that the change is a no-op (`path.join()` properly translates to forward slashes automatically).

TooTallNate:
> You might try `declaration: false` in `./packages/now-cli/tsconfig.json`

@styfle Didn't help.

styfle (approved):

lucleray (approved):

PR #4484, opened 2020-05-25
[next][routing-utils] Fix unnamed multi-match custom route

ijjk:
This adds a failing test for the regex from https://github.com/zeit/next.js/discussions/13347#discussion-4546 

It appears the reason the regex does not work when deployed on Vercel is due to us not indexing the un-named segments and instead we filter them out as noticed by @dav-is. 

This PR adds one approach to resolve this by accounting for the un-named indexes while not allowing them to be used in the destination still.

PR #4550, opened 2020-06-01
[routing-utils] Fix headers with content-security-policy URL

styfle:
Fixes a case when the header value contains a URL which was mistaken for a named segment.

https://sentry.io/organizations/zeithq/issues/1702692084/?project=1351065

The regression was introduced in PR #4484 where unnamed segments were implemented for `redirects` and `rewrites` but not handled properly in `headers`.


ijjk (dismissed):
LGTM although I think we need to still handle compiling segments if a URL is provided but since this should be less common it should be able to be done in a follow-up. 

Related PR in Next.js: https://github.com/vercel/next.js/pull/12031

dav-is (approved):

ijjk (approved):

PR #4559, opened 2020-06-02
[next] Handle restricted output names

ijjk:
This is one approach to handle a prototype collision occurring during output name checking which occurs when an output is named `constructor`, `hasOwnProperty`, or another prototype member name. A more ideal approach has also been opened to handle this outside of the builder to prevent this from having to be handled at the builder level


ijjk:
Closing for now as this is likely not the approach we want to take

PR #4562, opened 2020-06-02
[cli] Fetch environment variables from project if `.env` is not present

MaxLeiter:
Previously, users would have to run `vc env pull` to fetch cloud environment variables into `.env`. After this PR, if no `.env` or `.build.env` file is present, environment variables will be pulled by `vc dev` from your Vercel Environment Variables settings, no file necessary.

### DIFF
--- a/packages/now-cli/src/commands/dev/dev.ts
+++ b/packages/now-cli/src/commands/dev/dev.ts
@@ -93,5 +93,5 @@ export default async function dev(
   process.once('SIGINT', () => devServer.stop());
   process.once('SIGTERM', () => devServer.stop());
 
-  await devServer.start(client, link.project || undefined, ...listen);
+  await devServer.start(client, link.project, ...listen);
 }
--- a/packages/now-cli/src/commands/dev/dev.ts
+++ b/packages/now-cli/src/commands/dev/dev.ts
@@ -93,5 +93,5 @@ export default async function dev(
   process.once('SIGINT', () => devServer.stop());
   process.once('SIGTERM', () => devServer.stop());
 
-  await devServer.start(client, link.project, ...listen);
+  await devServer.start(client, link.project || undefined, ...listen);
 }
--- a/packages/now-cli/src/commands/dev/dev.ts
+++ b/packages/now-cli/src/commands/dev/dev.ts
@@ -9,7 +9,9 @@ import { getLinkedProject } from '../../util/projects/link';
 import { getFrameworks } from '../../util/get-frameworks';
 import { isSettingValue } from '../../util/is-setting-value';
 import { getCommandName } from '../../util/pkg-name';
-import { ProjectSettings } from '../../types';
+import { ProjectSettings, ProjectEnvTarget } from '../../types';
+import getDecryptedEnvRecords from '../../util/get-decrypted-env-records';
+import { Env } from '@vercel/build-utils';
 
 type Options = {
   '--debug'?: boolean;
@@ -54,6 +56,7 @@ export default async function dev(
   let devCommand: string | undefined;
   let frameworkSlug: string | undefined;
   let projectSettings: ProjectSettings | undefined;
+  let environmentVars: Env | undefined;
   if (link.status === 'linked') {
     const { project, org } = link;
     client.currentTeam = org.type === 'team' ? org.id : undefined;
@@ -80,6 +83,13 @@ export default async function dev(
     if (project.rootDirectory) {
       cwd = join(cwd, project.rootDirectory);
     }
+
+    environmentVars = await getDecryptedEnvRecords(
+      output,
+      client,
+      project,
+      ProjectEnvTarget.Development
+    );
   }
 
   const devServer = new DevServer(cwd, {
@@ -88,10 +98,11 @@ export default async function dev(
     devCommand,
     frameworkSlug,
     projectSettings,
+    environmentVars,
   });
 
   process.once('SIGINT', () => devServer.stop());
   process.once('SIGTERM', () => devServer.stop());
 
-  await devServer.start(client, link.project || undefined, ...listen);
+  await devServer.start(...listen);
 }
--- a/packages/now-cli/src/commands/dev/dev.ts
+++ b/packages/now-cli/src/commands/dev/dev.ts
@@ -93,5 +93,5 @@ export default async function dev(
   process.once('SIGINT', () => devServer.stop());
   process.once('SIGTERM', () => devServer.stop());
 
-  await devServer.start(...listen);
+  await devServer.start(client, link.project || undefined, ...listen);
 }
--- a/packages/now-cli/src/commands/env/pull.ts
+++ b/packages/now-cli/src/commands/env/pull.ts
@@ -4,8 +4,7 @@ import { Output } from '../../util/output';
 import promptBool from '../../util/prompt-bool';
 import Client from '../../util/client';
 import stamp from '../../util/output/stamp';
-import getEnvVariables from '../../util/env/get-env-records';
-import getDecryptedSecret from '../../util/env/get-decrypted-secret';
+import getDecryptedEnvRecords from '../../util/get-decrypted-env-records';
 import param from '../../util/output/param';
 import withSpinner from '../../util/with-spinner';
 import { join } from 'path';
@@ -13,6 +12,7 @@ import { promises, openSync, closeSync, readSync } from 'fs';
 import { emoji, prependEmoji } from '../../util/emoji';
 import { getCommandName } from '../../util/pkg-name';
 const { writeFile } = promises;
+import { Env } from '@vercel/build-utils';
 
 const CONTENTS_PREFIX = '# Created by Vercel CLI\n';
 
@@ -84,45 +84,21 @@ export default async function pull(
   );
   const pullStamp = stamp();
 
-  const records = await withSpinner('Downloading', async () => {
-    const dev = ProjectEnvTarget.Development;
-    const envs = await getEnvVariables(output, client, project.id, 4, dev);
-    const decryptedValues = await Promise.all(
-      envs.map(async env => {
-        try {
-          const value = await getDecryptedSecret(output, client, env.value);
-          return { value, found: true };
-        } catch (error) {
-          if (error && error.status === 404) {
-            return { value: '', found: false };
-          }
-          throw error;
-        }
-      })
-    );
-    const results: { key: string; value: string; found: boolean }[] = [];
-    for (let i = 0; i < decryptedValues.length; i++) {
-      const { key } = envs[i];
-      const { value, found } = decryptedValues[i];
-      results.push({ key, value, found });
-    }
-    return results;
-  });
+  const records: Env = await withSpinner(
+    'Downloading',
+    async () =>
+      await getDecryptedEnvRecords(
+        output,
+        client,
+        project,
+        ProjectEnvTarget.Development
+      )
+  );
 
   const contents =
     CONTENTS_PREFIX +
-    records
-      .filter(obj => {
-        if (!obj.found) {
-          output.print('');
-          output.warn(
-            `Unable to download variable ${obj.key} because associated secret was deleted`
-          );
-          return false;
-        }
-        return true;
-      })
-      .map(({ key, value }) => `${key}="${escapeValue(value)}"`)
+    Object.entries(records)
+      .map(([key, value]) => `${key}="${escapeValue(value)}"`)
       .join('\n') +
     '\n';
 
@@ -139,8 +115,10 @@ export default async function pull(
   return 0;
 }
 
-function escapeValue(value: string) {
+function escapeValue(value: string | undefined) {
   return value
-    .replace(new RegExp('\n', 'g'), '\\n') // combine newlines (unix) into one line
-    .replace(new RegExp('\r', 'g'), '\\r'); // combine newlines (windows) into one line
+    ? value
+        .replace(new RegExp('\n', 'g'), '\\n') // combine newlines (unix) into one line
+        .replace(new RegExp('\r', 'g'), '\\r') // combine newlines (windows) into one line
+    : '';
 }
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -140,6 +140,7 @@ export default class DevServer {
   private blockingBuildsPromise: Promise<void> | null;
   private updateBuildersPromise: Promise<void> | null;
   private updateBuildersTimeout: NodeJS.Timeout | undefined;
+  private startPromise: Promise<void> | null;
 
   private environmentVars: Env | undefined;
 
@@ -172,6 +173,7 @@ export default class DevServer {
     this.getNowConfigPromise = null;
     this.blockingBuildsPromise = null;
     this.updateBuildersPromise = null;
+    this.startPromise = null;
 
     this.watchAggregationId = null;
     this.watchAggregationEvents = [];
@@ -772,10 +774,18 @@ export default class DevServer {
     return Object.keys(files).filter(this.filter);
   }
 
+  start(...listenSpec: ListenSpec): void {
+    this.startPromise = this._start(...listenSpec);
+    this.startPromise.catch(err => {
+      this.stop();
+      throw err;
+    });
+  }
+
   /**
    * Launches the `vercel dev` server.
    */
-  async start(...listenSpec: ListenSpec): Promise<void> {
+  async _start(...listenSpec: ListenSpec): Promise<void> {
     if (!fs.existsSync(this.cwd)) {
       throw new Error(`${chalk.bold(this.cwd)} doesn't exist`);
     }
@@ -1232,6 +1242,8 @@ export default class DevServer {
     req: http.IncomingMessage,
     res: http.ServerResponse
   ) => {
+    await this.startPromise;
+
     let nowRequestId = generateRequestId(this.podId);
 
     if (this.stopping) {
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -83,10 +83,7 @@ import {
   HttpHeadersConfig,
   EnvConfigs,
 } from './types';
-import { ProjectEnvTarget, ProjectSettings, Project } from '../../types';
-
-import Client from '../../util/client';
-import getDecryptedEnvRecords from '../../util/get-decrypted-env-records';
+import { ProjectSettings } from '../../types';
 
 interface FSEvent {
   type: string;
@@ -144,13 +141,14 @@ export default class DevServer {
   private updateBuildersPromise: Promise<void> | null;
   private updateBuildersTimeout: NodeJS.Timeout | undefined;
 
-  private cachedEnvVars: Env | null;
+  private environmentVars: Env | undefined;
 
   constructor(cwd: string, options: DevServerOptions) {
     this.cwd = cwd;
     this.debug = options.debug;
     this.output = options.output;
     this.envConfigs = { buildEnv: {}, runEnv: {}, allEnv: {} };
+    this.environmentVars = options.environmentVars || {};
     this.files = {};
     this.address = '';
     this.devCommand = options.devCommand;
@@ -174,8 +172,6 @@ export default class DevServer {
     this.getNowConfigPromise = null;
     this.blockingBuildsPromise = null;
     this.updateBuildersPromise = null;
-
-    this.cachedEnvVars = null;
 
     this.watchAggregationId = null;
     this.watchAggregationEvents = [];
@@ -498,10 +494,7 @@ export default class DevServer {
       }
       return {
         ...this.validateEnvConfig(fileName, base || {}, env),
-        NOW_REGION: 'dev1',
-        NOW_URL: host,
-        VERCEL_REGION: 'dev1',
-        VERCEL_URL: host,
+        ...this.populateEnvConfig(this.environmentVars, host),
       };
     } catch (err) {
       if (err instanceof MissingDotenvVarsError) {
@@ -518,11 +511,11 @@ export default class DevServer {
     this.getNowConfigPromise = null;
   };
 
-  getNowConfig(client?: Client, project?: Project): Promise<NowConfig> {
+  getNowConfig(): Promise<NowConfig> {
     if (this.getNowConfigPromise) {
       return this.getNowConfigPromise;
     }
-    this.getNowConfigPromise = this._getNowConfig(client, project);
+    this.getNowConfigPromise = this._getNowConfig();
 
     // Clean up the promise once it has resolved
     const clear = this.clearNowConfigPromise;
@@ -531,7 +524,7 @@ export default class DevServer {
     return this.getNowConfigPromise;
   }
 
-  async _getNowConfig(client?: Client, project?: Project): Promise<NowConfig> {
+  async _getNowConfig(): Promise<NowConfig> {
     const configPath = getNowConfigPath(this.cwd);
 
     const [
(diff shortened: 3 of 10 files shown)


leo:
Think about it like this:

1. Load Development Environment Variables from the cloud, into memory.
2. If it exists, read `.env` into memory, make it overwrite the cloud Environment Variables and only apply it to runtime.
3. If it exists, read `.env.build` into memory, make it overwrite the cloud Environment Variables and only apply it to build time.

packages/now-cli/src/commands/dev/dev.ts
```diff
@@ -86,5 +87,5 @@ export default async function dev(
   process.once('SIGINT', () => devServer.stop());
   process.once('SIGTERM', () => devServer.stop());
 
-  await devServer.start(...listen);
+  await devServer.start(client, link.project, ...listen);
```
MaxLeiter: @TooTallNate The problem with this approach is now `DevServer` needs a reference to Client and the project

packages/now-cli/src/util/dev/server.ts
```diff
@@ -2216,3 +2243,45 @@ function hasNewRoutingProperties(nowConfig: NowConfig) {
     typeof nowConfig.trailingSlash !== undefined
   );
 }
+
+async function fetchAndDecryptEnvVars(
```
styfle: Can we reuse the code from `vc env pull` instead of implementing again here?

This seems to be copy/pasted.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -762,10 +771,28 @@ export default class DevServer {
...
     const nowConfigBuild = nowConfig.build || {};
-    const [runEnv, buildEnv] = await Promise.all([
-      this.getLocalEnv('.env', nowConfig.env),
-      this.getLocalEnv('.env.build', nowConfigBuild.env),
-    ]);
+    const localEnvVars = await this.getLocalEnv('.env', nowConfig.env);
+    const localBuildEnvVars = await this.getLocalEnv(
+      '.env.build',
+      nowConfigBuild.env
+    );
```
styfle: Any reason why these were changed from parallel to sequential reads?

MaxLeiter:
Addressed your comments @styfle

packages/now-cli/src/util/dev/server.ts
```diff
@@ -762,10 +769,28 @@ export default class DevServer {
...
+
+    const runEnv = Object.keys(localEnvVars).length
+      ? localEnvVars
+      : cloudEnvVariables;
+    const buildEnv = Object.keys(localBuildEnvVars).length
+      ? localBuildEnvVars
+      : cloudEnvVariables;
+
+    // if local .env/.env.build file exists, don't use cloud variables
     const allEnv = { ...buildEnv, ...runEnv };
```
styfle: I think you need to change this logic because the build/run env vars only makes sense if the files exist. If you are using the cloud, then it should be all or nothing.

Try something like this instead:

```js
let allEnv = { ...buildEnv, ...runEnv };
if (Object.keys(allEnv).length === 0) {
  allEnv = await getDecryptedEnvRecords(/*...*/);
}
```

styfle (commented):
We need to add an integration test for this new behavior in [./packages/now-cli/test/integration.js](https://github.com/vercel/vercel/blob/c80bb37e8d66eac83da40746cf1a15ba42d216d8/packages/now-cli/test/integration.js#L374).

TooTallNate:
@styfle It should be in `test/dev/integration.js` I think.

styfle:
@TooTallNate The dev tests are usually meant to create a deployment and test the paths between dev and production. But this test will be specific to the CLI behavior I think. That's why I linked to the previous `vercel env` test we already have. I suppose this test could be implemented in either file 🤷‍♂️

TooTallNate:
Fair enough. I don't really have a preference then.

packages/now-cli/test/integration.js
```diff
@@ -244,7 +249,7 @@ test.after.always(async () => {
   }
 
   // Make sure the token gets revoked
-  await execa(binaryPath, ['logout', ...defaultArgs]);
+  // await execa(binaryPath, ['logout', ...defaultArgs]);
```
TooTallNate: Don't commit this part, otherwise tokens will be leftover on the test users 😬

packages/now-cli/test/integration.js
```diff
@@ -602,6 +631,11 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
     t.is(exitCode, 0, formatOutput({ stderr, stdout }));
   }
 
+  // await nowEnvRemove();
+  // await nowEnvRemoveWithArgs();
+  // await nowEnvRemoveWithNameOnly();
+  // await nowEnvLsIsEmpty();
```
styfle: Did you forget to remove these comments?

packages/now-cli/test/integration.js
```diff
@@ -120,6 +120,11 @@ if (!process.env.CI) {
   globalDir = path.join(tmpDir.name, 'com.vercel.tests');
 
   defaultArgs.push('-Q', globalDir);
+
+  if (process.env.VERCEL_SCOPE) {
+    defaultArgs.push('--scope', process.env.VERCEL_SCOPE);
+  }
```
styfle: Hmm, I'm not sure this is a good idea since we have tests for `--scope` already.

See the test `try to deploy with non-existing team` for example. We don't want to pass two `--scope` options.

styfle (commented):
Did you intend to delete `examples/docusaurus-2/yarn.lock`? I think we need that file.

styfle (commented):
Theres a merge conflict but besides that it looks good 👍

packages/now-cli/test/integration.js
```diff
@@ -536,14 +537,38 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
     t.is(apiJson['VERCEL_URL'], host);
+  }
 
-    const homeUrl = `https://${host}`;
-    console.log({ homeUrl });
-    const homeRes = await fetch(homeUrl);
-    t.is(homeRes.status, 200, formatOutput({ stderr, stdout }));
-    const homeJson = await homeRes.json();
-    t.is(homeJson['MY_ENV_VAR'], 'MY_VALUE');
-    t.is(apiJson['VERCEL_URL'], host);
```
styfle, line 596: Please add this back. The existing `nowDeployWithVar()` function should not be modified.

MaxLeiter: Whoops, nice catch.

styfle (dismissed):
Looks great, thanks!

packages/now-cli/test/integration.js
```diff
@@ -546,6 +547,38 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+        localhost = /(https?:[^\s]+)/g.exec(chunk);
+        return true;
+      }
+      return false;
+    });
+
+    console.log('API URL: ' + localhost[0]);
+    const { host } = new URL(localhost[0]);
+
+    const apiUrl = `http://${host}/api/get-env`;
```
styfle:
```suggestion
    const apiUrl = `${localhost[0]}/api/get-env`;
```

styfle (dismissed):

packages/now-cli/test/integration.js
```diff
@@ -534,22 +535,51 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+    await waitForPrompt(vc, chunk => {
+      if (chunk.includes('Ready! Available at')) {
+        localhost = /(https?:[^\s]+)/g.exec(chunk);
+        return true;
+      }
+      return false;
+    });
+
+    const apiUrl = `${localhost[0]}/api/get-env`;
+    const apiRes = await fetch(apiUrl);
```
styfle, line 655: I only see one URL tested here - the API. Lets test the home page which is a static build to ensure the cloud assigns build env vars too.

You can probably copy past this code if you assign `host` https://github.com/vercel/vercel/blob/0eacbeae111dcc869afd848433d99f7aaa84d327/packages/now-cli/test/integration.js#L536-L550

styfle (dismissed):
Perfect 👌

packages/now-cli/test/integration.js
```diff
@@ -534,20 +535,55 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+    const apiJson = await apiRes.json();
+
+    t.is(apiJson['MY_ENV_VAR'], 'MY_VALUE');
+    t.is(apiJson['VERCEL_URL'], '');
+
+    const homeUrl = `${localhost[0]}`;
+    const homeRes = await fetch(homeUrl);
+    const homeJson = await homeRes.json();
+    t.is(homeJson['MY_ENV_VAR'], 'MY_VALUE');
+    t.is(homeJson['VERCEL_URL'], localhost[0]);
```
styfle:
```suggestion
    t.is(homeJson['VERCEL_URL'], '');
```

We don't populate the VERCEL_URL or any system env vars during development

packages/now-cli/test/integration.js
```diff
@@ -534,20 +535,55 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
     const apiUrl = `https://${host}/api/get-env`;
-    console.log({ apiUrl });
     const apiRes = await fetch(apiUrl);
     t.is(apiRes.status, 200, formatOutput({ stderr, stdout }));
     const apiJson = await apiRes.json();
     t.is(apiJson['MY_ENV_VAR'], 'MY_VALUE');
     t.is(apiJson['VERCEL_URL'], host);
 
     const homeUrl = `https://${host}`;
-    console.log({ homeUrl });
```
styfle: Any reason you're removing these logs?

They are helpful when the test fails and we need to see the deployment url

MaxLeiter: Thought they were accidentally left in

styfle (dismissed):

packages/now-cli/test/integration.js
```diff
@@ -542,12 +543,50 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+    console.log({ apiUrl });
```
styfle:
```suggestion
    console.log({ homeUrl });
```

styfle (dismissed):

lucleray (commented):
Can we add a test to verify that the behaviour when `now dev` is run and a `.env` file exists is correct?

packages/now-cli/src/util/get-decrypted-env-records.ts
```diff
@@ -0,0 +1,51 @@
...
+  client: Client,
+  project: null | Project,
+  apiVersion: string | number,
+  target: ProjectEnvTarget
+): Promise<Env> {
+  if (!project) {
+    return {};
+  }
+
+  const envs = await getEnvVariables(output, client, project.id, 4, target);
```
nkzawa, line 15:
```suggestion
  const envs = await getEnvVariables(output, client, project.id, apiVersion, target);
```
?

styfle:
~~@lucleray That test already exists, its just above the one added~~

I was mistaken, the existing test is for a production deployment with env vars, not a `vc dev` with a `.env` file.

styfle (dismissed):

packages/now-cli/src/commands/env/pull.ts
```diff
@@ -57,44 +59,29 @@ export default async function pull(
...
-      })
+  const records: Env = await withSpinner(
+    'Downloading',
+    async () =>
+      await getDecryptedEnvRecords(
+        output,
+        client,
+        project,
+        ProjectEnvTarget.Development
+      )
```
nkzawa, line 95:
```suggestion
    getDecryptedEnvRecords(
      output,
      client,
      project,
      ProjectEnvTarget.Development
    )
```

packages/now-cli/src/util/get-decrypted-env-records.ts
nkzawa: Can we make clear how this is solved?

packages/now-cli/src/util/get-decrypted-env-records.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!found) {
+      output.print('');
+      output.warn(
+        `Unable to download variable ${key} because associated secret was deleted`
+      );
+      continue;
+    }
+
+    results[key] = value ? value : '';
```
nkzawa, line 43: Just for confirmation, it looks like this does not allow duplicated keys even tho original code used to allow them. Is it no problem?

packages/now-cli/src/util/get-decrypted-env-records.ts
nkzawa: Oh ok, it does not receive the variable on argument anymore.

packages/now-cli/test/integration.js
nkzawa: ~Looks like this is not applied, why?~ actually it is, was looking old version.

packages/now-cli/src/commands/env/pull.ts
nkzawa: Actually, my suggestion is wrong since it takes function instead of a promise.

packages/now-cli/src/util/get-decrypted-env-records.ts
MaxLeiter: cc @styfle?

styfle: The API prevents multiple keys for the same `target` so this should be fine 

error-msg

packages/now-cli/src/commands/env/pull.ts
```diff
@@ -57,44 +59,29 @@ export default async function pull(
...
-          output.warn(
-            `Unable to download variable ${obj.key} because associated secret was deleted`
-          );
-          return false;
-        }
-        return true;
-      })
-      .map(({ key, value }) => `${key}="${escapeValue(value)}"`)
+    Object.keys(records)
+      .map(key => `${key}="${escapeValue(records[key]!)}"`)
```
styfle:
```suggestion
    Object.entries(records)
      .map(([key, value]) => `${key}="${escapeValue(value)}"`)
```

packages/now-cli/src/commands/dev/dev.ts
```diff
@@ -93,5 +93,5 @@ export default async function dev(
...
```
nkzawa: I think it's better to pass these arguments to the constructor of  `DevServer` since they're used on `getNowConfig` which is called from other methods as well. (tho it looks working fine because of cache).

packages/now-cli/test/integration.js
```diff
@@ -591,11 +592,86 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+    const apiRes = await fetch(apiUrl);
+
+    t.is(apiRes.status, 200);
+
+    const apiJson = await apiRes.json();
+
+    t.is(apiJson['MY_ENV_VAR'], 'MY_VALUE');
+    t.is(apiJson['VERCEL_URL'], '');
+
+    const homeUrl = `${localhost[0]}`;
```
nkzawa: nitpick but template literal is not necessary here?

nkzawa (dismissed):
a few nitpick and suggestion but otherwise LGTM

TooTallNate (dismissed):

MaxLeiter:
@TooTallNate @nkzawa sorry for dismissed review, had to address a merge conflict. The change since last review was adapting to the new `getNowConfig()` in `server.ts`

packages/now-cli/test/helpers/prepare.js
```diff
@@ -365,7 +365,7 @@ CMD ["node", "index.js"]`,
       'package.json': JSON.stringify({
         private: true,
         scripts: {
-          build: 'mkdir public && node print.js > public/index.json',
+          build: 'mkdir -p public && node print.js > public/index.json',
```
MaxLeiter, line 368: this is the change in this file

packages/now-cli/src/util/dev/server.ts
```diff
@@ -764,7 +803,40 @@ export default class DevServer {
     const { ig } = await getVercelIgnore(this.cwd);
     this.filter = ig.createFilter();
 
-    const nowConfig = await this.getNowConfig();
+    const devCommandPromise = this.runDevCommand();
+
+    let address: string | null = null;
+    while (typeof address !== 'string') {
```
MaxLeiter, line 803: this moved up so `address` is available when `VERCEL_URL` is set.

MaxLeiter: This also fixes a bug where if `VERCEL_URL` is set in your `.env` (like from `vc env pull`), it's stored and set to an empty string `VERCEL_URL=''`

Screen Shot 2020-07-20 at 12 51 26 AM

packages/now-cli/src/util/dev/server.ts
```diff
@@ -511,11 +518,11 @@ export default class DevServer {
     this.getNowConfigPromise = null;
   };
 
-  getNowConfig(): Promise<NowConfig> {
+  getNowConfig(client?: Client, project?: Project): Promise<NowConfig> {
```
styfle: Why is `client` optional?

Also why is `project` optional? Seems like we already error when no project is linked in `vc dev`.

MaxLeiter: So we don't pass them around each call to getNowConfig

MaxLeiter: They could be moved to ProjectSettings, or something else other than via `start()`?

packages/now-cli/src/util/dev/server.ts
MaxLeiter: @TooTallNate

packages/now-cli/src/util/dev/server.ts
```diff
@@ -491,10 +494,7 @@ export default class DevServer {
       }
       return {
         ...this.validateEnvConfig(fileName, base || {}, env),
-        NOW_REGION: 'dev1',
-        NOW_URL: host,
-        VERCEL_REGION: 'dev1',
-        VERCEL_URL: host,
+        ...this.populateEnvConfig(this.environmentVars, host),
```
MaxLeiter: I could get the `host` via `this.address` inside populateEnvConfig

packages/now-cli/src/util/get-decrypted-env-records.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export default async function getDecryptedEnvRecords(
+  output: Output,
+  client: Client,
+  project: null | Project,
+  target: ProjectEnvTarget
+): Promise<Env> {
+  if (!project) {
+    return {};
+  }
```
styfle: Similarly here for the return?

packages/now-cli/src/util/dev/types.ts
```diff
@@ -27,11 +27,12 @@ export interface DevServerOptions {
   devCommand?: string;
   frameworkSlug?: string;
   projectSettings?: ProjectSettings;
+  environmentVars?: Env;
 }
 
 export interface EnvConfigs {
   /**
-   * environment variables from `.env.build` file (deprecated)
```
styfle, line 34: I'm curious why deprecated was removed?

packages/now-cli/src/util/dev/server.ts
```diff
@@ -141,11 +141,14 @@ export default class DevServer {
...
   private updateBuildersTimeout: NodeJS.Timeout | undefined;
 
+  private environmentVars: Env | undefined;
+
   constructor(cwd: string, options: DevServerOptions) {
     this.cwd = cwd;
     this.debug = options.debug;
     this.output = options.output;
     this.envConfigs = { buildEnv: {}, runEnv: {}, allEnv: {} };
+    this.environmentVars = options.environmentVars || {};
```
styfle: Something seems wrong here. We need the `envConfigs` to `options.environmentVars` if the user didn't provide any files.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -764,7 +783,39 @@ export default class DevServer {
...
+        }
+      }
+    }
+
+    this.address = address
+      .replace('[::]', 'localhost')
+      .replace('127.0.0.1', 'localhost');
+
     const nowConfig = await this.getNowConfig();
+    const devCommandPromise = this.runDevCommand();
```
MaxLeiter, line 834: @TooTallNate moved the init to below starting the server, and left the await at the end

packages/now-cli/src/util/dev/types.ts
MaxLeiter: It's not deprecated. @TooTallNate and I discussed on Slack a while back

packages/now-cli/src/util/dev/server.ts
MaxLeiter: Happens in getNowConfig

packages/now-cli/src/util/dev/server.ts
```diff
@@ -630,13 +626,20 @@ export default class DevServer {
...
       this.getLocalEnv('.env', config.env),
       this.getLocalEnv('.env.build', config.build?.env),
     ]);
-    const allEnv = { ...buildEnv, ...runEnv };
-    this.envConfigs = { buildEnv, runEnv, allEnv };
 
+    let allEnv = { ...buildEnv, ...runEnv };
+
+    // If no .env/.build.env is present, fetch and use cloud environment variables
+    if (Object.keys(allEnv).length === 0) {
```
styfle, line 639: I think this will never be true because `NOW_REGION` is always added, right?

styfle: I would suggest adding `NOW_REGION` to `runEnv` after reading the files and reading the cloudEnv.

Something like this:

```js
let [runEnv, buildEnv] = await Promise.all([
  this.getLocalEnv('.env', config.env),
  this.getLocalEnv('.env.build', config.build?.env),
]);

if (Object.keys(buildEnv).length === 0 && Object.keys(runEnv).length === 0) {
  runEnv = buildEnv = this.environmentVars || {};
  if ('VERCEL_REGION' in runEnv) {
    runEnv['VERCEL_REGION'] = 'dev1';
  }
  if ('VERCEL_URL' in runEnv) {
    runEnv['VERCEL_URL'] = host;
  }
}

runEnv['NOW_REGION'] = 'dev1';
const allEnv = { ...buildEnv, ...runEnv };
this.envConfigs = { buildEnv, runEnv, allEnv };
```

styfle (dismissed):

packages/now-cli/src/util/dev/server.ts
styfle: Nevermind, ignore my comments

TooTallNate (dismissed):

TooTallNate (dismissed):

styfle (dismissed):

packages/now-cli/src/util/dev/server.ts
```diff
@@ -741,6 +746,26 @@ export default class DevServer {
...
+    if (!env) {
+      return {};
+    }
+
+    for (const name of Object.keys(env)) {
+      if (name === 'VERCEL_URL') {
+        const host = new URL(this.address).host;
+        env['VERCEL_URL'] = host;
+      } else if (name === 'VERCEL_REGION') {
+        env['VERCEL_REGION'] = 'dev1';
```
styfle, line 759: @TooTallNate Do we want to use `dev` instead of `dev1` like leo suggested in [ch4611](https://app.clubhouse.io/vercel/story/4611)? Once we ship this, I don't think we can change it.

TooTallNate: I mean I explained my opinion about it in the ch story, but ultimately I don't think it's my call. If we do change it to `dev` we'll need to update [the documentation](https://vercel.com/docs/v2/edge-network/regions#routing) as well.

styfle (dismissed):

TooTallNate (approved):

styfle (approved):

PR #4563, opened 2020-06-02
[cli][build-utils] Add support for custom 404.html for zero config deployments

styfle:
Next.js already has support for [customizing the 404 page](https://nextjs.org/docs/advanced-features/custom-error-page#customizing-the-404-page), but many other frameworks do not or they expect a 404.html in the output directory.

This PR adds support for rendering the a `404.html` page for all zero config deployments.

- Implements ch337
- Related to #3491


dav-is (dismissed):

TooTallNate (changes requested):
Dev tests are failing

styfle:
@TooTallNate I had to change a few tests to `skipDeploy: true` so that the tests will pass. Once this is merged I can release canary, update the Deployments API, and finally create another PR to remove that flag to ensure we are testing against both dev and prod.

dav-is (approved):

TooTallNate (approved):

AndyBitz (approved):

PR #4565, opened 2020-06-03
[all] Update TypeScript to v3.9.3 and regenerate `yarn.lock` file

TooTallNate:
To fix GitHub Dependabot security alerts. See: https://github.com/vercel/vercel/network/alerts

Screen Shot 2020-06-03 at 1 36 50 AM

Also updates "typescript" to `3.9.3` for all packages, which is necessary because a downstream dependency is using syntax that requires a newer version of TypeScript.

### DIFF
--- a/packages/frameworks/package.json
+++ b/packages/frameworks/package.json
@@ -12,6 +12,6 @@
     "ajv": "6.10.2",
     "jest": "24.9.0",
     "ts-jest": "24.1.0",
-    "typescript": "3.5.2"
+    "typescript": "3.9.3"
   }
 }
--- a/packages/now-build-utils/package.json
+++ b/packages/now-build-utils/package.json
@@ -45,7 +45,7 @@
     "node-fetch": "2.2.0",
     "semver": "6.1.1",
     "ts-jest": "24.1.0",
-    "typescript": "3.5.2",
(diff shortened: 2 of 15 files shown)


styfle (commented):
I think we’ll want to change the tsconfig.json files target from esnext to es2018 or es2019 so that we can take advantage of optional changing and other features available in TS. Otherwise we might break older versions of node.

TooTallNate:
@styfle Well we specify `engines >= 10` in the `package.json` and we test Node 10 and 12. So I think we're good? Which `tsconfig.json` file are you talking about specifically?

styfle:
@TooTallNate See how these are using `esnext` https://github.com/vercel/vercel/blob/8ec8affe29e49c27c437d6d74daa8a4f343fce94/packages/now-python/tsconfig.json#L16

This worked fine with the older TS because there was no feature of TS that wasn't implemented in Node 10. But now that there is a newer version of TS, there are some features so we need to downlevel by setting target.

styfle:
Ok I think we can merge because if there are bugs with downleveling, our Node 10 tests will catch it 👍

styfle (approved):

AndyBitz (approved):

PR #4570, opened 2020-06-03
[cli] Use `getArgs()` instead of `mri` to parse arguments in `vc projects`

TooTallNate:
* Ensures that `vc --debug projects ls` properly prints the Projects listing, rather than the usage help info.
 * `vc projects` (without a subcommand) shows the Projects listing (this is consistent with i.e. `vc domains`).
 * Returns with exit code `2` when the usage help info is printed (standard Unix convention).


styfle (commented):
Needs a CLI test

packages/now-cli/test/integration.js
```diff
@@ -2916,3 +2916,19 @@ test('reject conflicting `vercel.json` and `now.json` files', async t => {
     formatOutput({ stderr, stdout })
   );
 });
+
+test('`vc --debug project ls` should output the projects listing', async t => {
+  const { exitCode, stderr, stdout } = await execa(
+    binaryPath,
+    [...defaultArgs, '--debug', 'project', 'ls'],
```
styfle, line 2923: I'm not sure if it will make a difference, but usually defaultArgs is at the end

styfle (approved):

TooTallNate:
Added a test case.

packages/now-cli/test/integration.js
TooTallNate: Good point, but I don't think it makes a difference in this case (the issue was happening when `--debug` came before `project`). This test was copied from the test above.

AndyBitz (approved):

PR #4583, opened 2020-06-05
[cli] Update 502 error template for `vc dev`

TooTallNate:
* Fix the "Developer Documentation" link.
 * Remove the "If you're a visitor" section - doesn't make sense for `vc dev` since there are no "visitors".
 * Don't link to `_logs` since it's not supported in `vc dev`. Instead direct the user to look at their terminal window to see error logs.
 * Link to new GH issue for non-app error 502 (I don't think this code path ever happens in `vc dev`, but might as well make it correct in case we do in the future).

Screen Shot 2020-06-05 at 4 15 16 PM

### DIFF
--- a/packages/now-cli/src/util/dev/templates/error_502.jst
+++ b/packages/now-cli/src/util/dev/templates/error_502.jst
@@ -53,7 +53,7 @@
   <p>
     <ul>
       <li>
-        Please open a <a target="_blank" href="https://github.com/vercel/vercel/issues/new">GitHub issue</a> describing the problem you are experiencing with <code>vercel dev</code>.
+        Please open a <a target="_blank" href="https://github.com/vercel/vercel/issues/new/choose">GitHub issue</a> describing the problem you are experiencing with <code>vercel dev</code>.
       </li>
     </ul>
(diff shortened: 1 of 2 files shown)


packages/now-cli/src/util/dev/templates/error_502.jst
```diff
@@ -44,22 +44,16 @@
...
   </p>
   {{??}}
   <p>
     <ul>
       <li>
-        If you are a visitor, please try again later.
-      </li>
-      <li>
-        If you are the owner, no action is needed. Our engineers have been notified.
+        Please open a <a target="_blank" href="https://github.com/vercel/vercel/issues/new">GitHub issue</a> describing the problem you are experiencing with <code>vercel dev</code>.
```
styfle:
```suggestion
        Please open a <a target="_blank" href="https://github.com/vercel/vercel/issues/new/choose">GitHub issue</a> describing the problem you are experiencing with <code>vercel dev</code>.
```

styfle (approved):

AndyBitz (approved):

PR #4593, opened 2020-06-08
[ruby] Fix for UTF-8 responses in Ruby functions

jaredcwhite:
Apparently there's an unexpected string encoding issue with Net::HTTP, so this is a workaround. Further details:
https://bugs.ruby-lang.org/issues/15517


jaredcwhite:
Actually, now that I'm thinking about it, that response object is probably getting converted to JSON for the lambda processing, so it might need a forced encoding to UTF-8 no matter what, but that's beyond the scope of my local testing…

packages/now-ruby/now_init.rb
```diff
@@ -59,11 +59,17 @@ def webrick_handler(httpMethod, path, body, headers)
...
+  if res_headers["Content-Type"] && res_headers["Content-Type"].include?("charset=")
+    res_encoding = res_headers["Content-Type"].match(/charset=([^;]*)/)[1]
+  end
 
   {
     :statusCode => res.code.to_i,
-    :headers => res.each_capitalized.to_h,
-    :body => res.body,
+    :headers => res_headers,
+    :body => res.body.force_encoding(res_encoding),
```
styfle: Thanks for the PR!

Please add a test fixture to `./packages/now-ruby/test/fixtures` so we can confirm this fixes the issue 👍

packages/now-ruby/now_init.rb
```diff
@@ -59,11 +59,17 @@ def webrick_handler(httpMethod, path, body, headers)
 
   server.shutdown
   Thread.kill(th)
+  
+  res_headers = res.each_capitalized.to_h
+  res_encoding = "UTF-8"
+  if res_headers["Content-Type"] && res_headers["Content-Type"].include?("charset=")
+    res_encoding = res_headers["Content-Type"].match(/charset=([^;]*)/)[1]
```
styfle, line 68: Let's add a comment here linking to the upstream issue https://bugs.ruby-lang.org/issues/15517

jaredcwhite:
@styfle I added the comment and test fixture.

styfle:
@jaredcwhite Thanks!

I think there is still something wrong. Here's what I see when I deploy your test:

image


```

  | ______________________________
-- | --
  | \| gem:283148d7283148d7283148 ðŸ„ \|
  | ------------------------------
  | \   ^__^
  | \  (oo)\_______
  | (__)\       )\/\
  | \|\|----w \|
  | \|\|     \|\|
  |  


```

packages/now-ruby/now_init.rb
```diff
@@ -59,11 +59,19 @@ def webrick_handler(httpMethod, path, body, headers)
...
+  end
 
   {
     :statusCode => res.code.to_i,
-    :headers => res.each_capitalized.to_h,
-    :body => res.body,
+    :headers => res_headers,
+    :body => res.body.force_encoding(res_encoding),
```
styfle: I think the think we need to call `force_encoding` as well as `encode`.

This seems to work:

```ruby
res.body.force_encoding('UTF-8')
res.body = res.body.encode('UTF-8')
```

jaredcwhite: Seems like that would work when in the deployment context, yeah.

styfle (approved):
Thanks for this, great job! 🎉 

I pushed a couple changes and we can merge once CI passes 👍

PR #4596, opened 2020-06-08
[next] Add shared lambdas opt-out for functions config

ijjk:
As discussed this adds opting out of the shared lambdas optimization when a user adds a functions config in `now.json` or `vercel.json` since this could potentially be a breaking change. We plan to add new handling to still allow customizing this config for the combined lambdas that are created

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -226,7 +226,7 @@ export const build = async ({
 
   if (hasFunctionsConfig) {
     console.warn(
-      `WARNING: your application is being opted out of @vercel/next's optimized lambdas mode due to \`functions\` config. http://err.sh/vercel/vercel/next-functions-config-optimized-lambdas`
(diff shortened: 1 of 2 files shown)


Timer (dismissed):
LGTM if tests pass.

packages/now-next/src/index.ts
```diff
@@ -223,9 +224,17 @@ export const build = async ({
     }
   }
 
+  if (hasFunctionsConfig) {
+    console.warn(
+      `WARNING: your application is being opted out of @vercel/next's optimized lambdas mode due to \`functions\` config. http://err.sh/vercel/vercel/next-functions-config-optimized-lambdas`
```
TooTallNate: What do you think about putting the link on a new line?

```suggestion
      `WARNING: Your application is being opted out of "@vercel/next" optimized lambdas mode due to \`functions\` config.\nMore info: http://err.sh/vercel/vercel/next-functions-config-optimized-lambdas`
```

ijjk: Sounds good to me

TooTallNate (approved):

Timer (approved):

PR #4598, opened 2020-06-08
[cli] Show npm output when failing to install Runtimes in `vc dev`

TooTallNate:
If `npm install` doesn't exit with 0, then log the output to the
termial.

This was already happening with `--debug` due to `stdio: "inherit"`
when debug is enabled, so this change is really only relevant for the
non-debug case.

### Before

```
$ vc dev
Vercel CLI 19.0.2-canary.13 dev (beta) — https://vercel.com/feedback
Error! Command failed with exit code 1: npm install --save-exact --no-package-lock --no-audit --no-progress @now/build-utils@canary @vercel/build-utils@canary now-php@0.0.0
```

### After

Bad dependency:

Screen Shot 2020-06-08 at 12 22 00 PM

No `npm` installed:

Screen Shot 2020-06-08 at 12 21 16 PM

### DIFF
--- a/packages/now-cli/src/util/dev/builder-cache.ts
+++ b/packages/now-cli/src/util/dev/builder-cache.ts
@@ -3,12 +3,12 @@ import semver from 'semver';
 import npa from 'npm-package-arg';
 import pluralize from 'pluralize';
 import { basename, join } from 'path';
-import { PackageJson } from '@vercel/build-utils';
 import XDGAppPaths from 'xdg-app-paths';
 import { mkdirp, readJSON, writeJSON } from 'fs-extra';
+import { NowBuildError, PackageJson } from '@vercel/build-utils';
 import cliPkg from '../pkg';
 
-import { NoBuilderCacheError, NpmInstallError } from '../errors-ts';
+import { NoBuilderCacheError } from '../errors-ts';
 import { Output } from '../output';
 import { getDistTag } from '../get-dist-tag';
 
@@ -258,11 +258,14 @@ async function npmInstall(
       if (result.stderr) {
         console.error(result.stderr);
       }
-      const isEnoent = (result as any).code === 'ENOENT';
-      throw new NpmInstallError(
-        'Failed to install `vercel dev` dependencies.',
-        isEnoent
-      );
+      throw new NowBuildError({
+        message:
+          (result as any).code === 'ENOENT'
+            ? '`npm` is not installed'
+            : 'Failed to install `vercel dev` dependencies',
+        code: 'NPM_INSTALL_ERROR',
+        link: 'https://vercel.link/npm-install-error',
+      });
     }
   } finally {
     stopSpinner();
--- a/packages/now-cli/src/util/dev/builder-cache.ts
+++ b/packages/now-cli/src/util/dev/builder-cache.ts
@@ -232,21 +232,34 @@ async function npmInstall(
   output.debug(`Running npm install in ${cwd}`);
 
   try {
-    await execa(
-      'npm',
-      [
-        'install',
-        '--save-exact',
-        '--no-package-lock',
-        '--no-audit',
-        '--no-progress',
-        ...sortedPackages,
-      ],
-      {
-        cwd,
-        stdio: output.isDebugEnabled() ? 'inherit' : undefined,
+    const args = [
+      'install',
+      '--save-exact',
+      '--no-package-lock',
+      '--no-audit',
+      '--no-progress',
+    ];
+    if (process.stderr.isTTY) {
+      // Force colors in the npm child process
+      // https://docs.npmjs.com/misc/config#color
+      args.push('--color=always');
+    }
+    args.push(...sortedPackages);
+    const result = await execa('npm', args, {
+      cwd,
+      reject: false,
+      stdio: output.isDebugEnabled() ? 'inherit' : 'pipe',
+    });
+    if (result.failed) {
+      stopSpinner();
+      if (result.stdout) {
+        console.log(result.stdout);
       }
-    );
+      if (result.stderr) {
+        console.error(result.stderr);
+      }
+      throw result;
+    }
   } finally {
     stopSpinner();
   }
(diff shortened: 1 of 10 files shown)


packages/now-cli/src/util/dev/builder-cache.ts
```diff
@@ -232,21 +232,34 @@ async function npmInstall(
...
+    if (result.failed) {
+      stopSpinner();
+      if (result.stdout) {
+        console.log(result.stdout);
       }
-    );
+      if (result.stderr) {
+        console.error(result.stderr);
+      }
+      throw result;
```
styfle: Should this `throw new Error(result.stderr)` instead? I figure we should only throw Error objects and I'm not sure the type here.

TooTallNate: When `execa` fails with `reject: false`, it returns the Error object directly, with additional props tacked on (`stdout`, `stderr`, `exitCode`, etc.) so it is throwing an Error here.

However, I'm refactoring this part to use a custom Error subclass that'll include an HTTP "learn more" link.

packages/now-cli/src/util/errors-ts.ts
```diff
@@ -1236,3 +1236,26 @@ export class BuildError extends NowError<'BUILD_ERROR', {}> {
...
+    this.action = action;
+  }
+}
+
+export class NpmInstallError extends LinkableError {
+  constructor(message: string, isEnoent = false) {
+    if (isEnoent) {
+      message += ' Make sure `npm` is installed.';
+    }
+    super(message, 'https://vercel.link/npm-install-error');
```
TooTallNate: This link needs to be created.

packages/now-cli/src/util/errors-ts.ts
```diff
@@ -1236,3 +1236,26 @@ export class BuildError extends NowError<'BUILD_ERROR', {}> {
     });
   }
 }
+
+/**
+ * This Error subclass can be used by the `output.prettyError()` function.
+ */
+export class LinkableError extends Error {
```
styfle: I think you can use `import { NowBuildError } from '@vercel/build-utils'` instead of creating a new class

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -71,10 +73,21 @@ export default function createOutput({ debug: debugEnabled = false } = {}) {
...
-      print(`${action}: ${details}\n`);
+      print(`${chalk.bold(action)}: ${renderLink(details)}\n`);
     }
   }
 
+  function prettyError(err: Error) {
+    let str = err.message;
+    let link: string | null = null;
+    let action: string | undefined = undefined;
+    if (err instanceof LinkableError) {
```
styfle: Let's avoid using an `instanceof` because we should be able to do this for API Errors too.

packages/now-cli/src/util/errors-ts.ts
TooTallNate: Ok good call.

packages/now-cli/src/util/errors-ts.ts
TooTallNate: https://github.com/vercel/docs/pull/1894

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -120,8 +121,6 @@ export default function createOutput({ debug: debugEnabled = false } = {}) {
     return wait(message, delay);
   }
 
-  // This is pretty hacky, but since we control the version of Node.js
-  // being used because of `pkg` it's safe to do in this case.
```
styfle, line 111: This is no longer "hacky"? 😄

TooTallNate: It's still hacky, but the part about `pkg` is no longer relevant 😛

packages/now-cli/src/util/errors-ts.ts
TooTallNate: It's live: https://vercel.link/npm-install-error

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -64,15 +65,28 @@ export default function createOutput({ debug: debugEnabled = false } = {}) {
...
+  function prettyError(err: Error & { link?: string; action?: string }) {
+    let str = err.message;
+    let link: string | undefined = undefined;
+    let action: string | undefined = undefined;
+    if (err.link) {
+      link = err.link;
+    }
+    if (err.action) {
+      action = err.action;
     }
```
styfle: Are these two `if` statements necessary?

Seems like you could just `return error(str, undefined, err.link, err.action);` 🤔 

Or worst case scenario: `let link = err.link || undefined;`

packages/now-cli/src/util/errors-ts.ts
```diff
@@ -1236,3 +1237,16 @@ export class BuildError extends NowError<'BUILD_ERROR', {}> {
     });
   }
 }
+
+export class NpmInstallError extends NowBuildError {
```
styfle: I'm curious if we still need this pattern at all. Why a new class for each error? Are you planning to catch it somewhere? Since the `message` is passed in, it doesn't seem very reusable.

TooTallNate: a95312d

packages/now-routing-utils/src/index.ts
```diff
@@ -230,6 +230,7 @@ function createError(
   const errors = Array.isArray(allErrors) ? allErrors : [allErrors];
   const message = errors[0];
   const error: RouteApiError = {
+    name: 'RouteApiError',
```
styfle, line 233: What's the purpose of `name` here?

packages/now-cli/src/util/output/create-output.ts
TooTallNate: Good call. 85cb7ed

packages/now-routing-utils/src/index.ts
TooTallNate: It's a required property on `Error`, so needed to add it in here to be able to use it with `prettyError()`. Another option would be to use `Partial<Error>` in the prettyError() definition…

styfle (commented):
How about a test using `builds: [{"src": "index.js", "use": "@vercel/notreal"}]` to ensure this behavior works? You could use `skipDeploy: true` too.

styfle (approved):
🥇

AndyBitz (approved):

PR #4603, opened 2020-06-09
[cli] Invalidate build match for changes to `config` in builds

TooTallNate:
* Fixes changing a `config` value in the `builds` array (such as `helpers: false` for `@vercel/node`) and having it be reflected in `vc dev` without restarting the dev server.
* Re-validates the env vars configuration when a `.env` file changes or the `env` object is changed in `vercel.json` (same for the builds equivalents).
* Ensures that the `NODEJS_HELPERS` build env var is being properly checked in `startDevServer()`.
* Regenerated the `yarn.lock` file because yarn was erroring when trying to add the `fast-deep-equal` dependency.

### DIFF
--- a/packages/frameworks/package.json
+++ b/packages/frameworks/package.json
@@ -9,7 +9,7 @@
   "devDependencies": {
     "@types/jest": "24.0.22",
     "@types/node": "12.0.4",
-    "ajv": "6.10.2",
+    "ajv": "6.12.2",
     "jest": "24.9.0",
     "ts-jest": "24.1.0",
     "typescript": "3.9.3"
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -18,6 +18,7 @@ import directoryTemplate from 'serve-handler/src/directory';
 import getPort from 'get-port';
 import { ChildProcess } from 'child_process';
 import isPortReachable from 'is-port-reachable';
+import deepEqual from 'fast-deep-equal';
 import which from 'which';
 
 import { getVercelIgnore, fileNameSymbol } from '@vercel/client';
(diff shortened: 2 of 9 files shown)


packages/frameworks/package.json
```diff
@@ -9,7 +9,7 @@
   "devDependencies": {
     "@types/jest": "24.0.22",
     "@types/node": "12.0.4",
-    "ajv": "6.10.2",
+    "ajv": "6.12.2",
```
TooTallNate, line 12: This update was to remove a duplicated `ajv` dependency in `yarn.lock`.

packages/now-cli/test/dev/fixtures/node-helpers/vercel.json
```diff
@@ -0,0 +1 @@
+{ "builds": [{ "src": "index.js", "use": "@vercel/node@canary" }] }
```
styfle, line 1: Is this file necessary?

packages/now-cli/src/util/dev/server.ts
```diff
@@ -2261,3 +2272,11 @@ function hasNewRoutingProperties(nowConfig: NowConfig) {
     typeof nowConfig.trailingSlash !== undefined
   );
 }
+
+function buildMatchEquals(a?: BuildMatch, b?: BuildMatch): boolean {
+  if (!a || !b) return false;
+  if (a.src !== b.src) return false;
+  if (a.use !== b.use) return false;
+  if (!deepEqual(a.config || {}, b.config || {})) return false;
+  return true;
```
styfle, line 2281: Is there a reason why you are only comparing 3 properties (src, use, config) instead of all?

```ts
function buildMatchEquals(a?: BuildMatch, b?: BuildMatch)
  return deepEqual(a, b);
}
```

packages/now-cli/test/dev/fixtures/node-helpers/vercel.json
styfle: Nevermind, I see it is used in the test below 😄

styfle (approved):

packages/now-cli/src/util/dev/server.ts
styfle: I guess there are a lot of properties on `BuildMatch` and we only want to compare the properties on `BuildConfig` 👍

AndyBitz (approved):

PR #4605, opened 2020-06-09
[cli] Handle aliasWarning and bump to v13/now/deployments 

mglagola:
* Adds handling for `aliasWarning` that may exist on a deployment document
* Uses `getPreferredPreviewURL` which will try to find an alias/domain that is assigned to a deployment that has an SSL cert. Otherwise, `getPreferredPreviewURL` will fallback to the first alias (re-using this fallback from previous implementation).
* Bumps `/v12/now/deployments` url to `/v13/now/deployments`.
  * In v13, certs will be will not block alias assignment.


packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -706,6 +703,12 @@ export default async function main(
 
   return printDeploymentStatus(
     output,
+    new Client({
```
TooTallNate, line 706: Don't we already create a client before this? Can we reuse that instance?

TooTallNate (commented):
Code looks good. Would it be possible to add a test case for this behavior?

TooTallNate (approved):

AndyBitz (approved):

PR #4609, opened 2020-06-09
[next] Re-disable shared lambdas by default

ijjk:
As noticed, `now.json` isn't available during the build currently which makes feature detecting and opting out when `routes` are used not work currently so this re-disables the shared lambdas optimization by default while we investigate detecting this

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -70,7 +70,7 @@ import {
   syncEnvVars,
   validateEntrypoint,
 } from './utils';
-import findUp from 'find-up';
+// import findUp from 'find-up';
 
 interface BuildParamsMeta {
   isDev: boolean | undefined;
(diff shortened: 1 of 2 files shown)


Timer (dismissed):

timneutkens (dismissed):

Timer (approved):

timneutkens (approved):

packages/now-next/src/index.ts
```diff
@@ -70,7 +70,7 @@ import {
   syncEnvVars,
   validateEntrypoint,
 } from './utils';
-import findUp from 'find-up';
+// import findUp from 'find-up';
```
matheuss, line 73: Let's remove this?

Timer: We want to re-enable this code ASAP so figured it was just easier to leave commented out.

PR #4621, opened 2020-06-10
[node] Force `module: "commonjs"` in `startDevServer()`

TooTallNate:
This is a follow up to #4514 to handle the case where there is no `tsconfig.json` closer to the entrypoint. This is likely the case when `.js` files with ES Modules syntax are being used instead of `.ts`.

Fixes errors such as:

image


styfle (dismissed):
💯

lucleray (dismissed):

PR #4629, opened 2020-06-11
[cli] Invalidate the configuration if `vercel.json` is deleted

TooTallNate:
Before this, if the `vercel.json` file gets deleted while the dev server
is running, then it would still act like the file exists since it would
use the cached version.

Now it properly invalidates to an empty configuration if the
`vercel.json` file does not exist.


packages/now-cli/src/util/dev/server.ts
```diff
@@ -533,7 +533,7 @@ export default class DevServer {
     // The default empty `vercel.json` is used to serve all files as static
     // when no `vercel.json` is present
     let configPath = 'vercel.json';
-    let config: NowConfig = this.cachedNowConfig || {
+    let config: NowConfig = (canUseCache && this.cachedNowConfig) || {
```
styfle: I'm confused, wouldn't this line never evaluate to true `(canUseCache && this.cachedNowConfig)` because of line 527-529? It seems like it would always create a new object so you could remove the conditional here and always assign the new object.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -784,7 +784,6 @@ export default class DevServer {
       this.getLocalEnv('.env.build', nowConfigBuild.env),
     ]);
     const allEnv = { ...buildEnv, ...runEnv };
-    Object.assign(process.env, allEnv);
```
styfle, line 787: Why is this deleted? 🤔

TooTallNate: It's part of the bug. `vc dev`'s process.env should not be tainted. The project's configured env vars should be passed in to the child processes explicitly.

Consider that the "FOO" env var gets set in this parent process, and then the user deletes the `vercel.json` file so that "FOO" should no longer exist in the deployment, but it still does because it got assigned here in the parent process, which is wrong.

packages/now-cli/src/util/dev/server.ts
TooTallNate: The case is that `canUseCache` is set to false, so that first `if` branch isn't true, and then we want to _not_ use `cachedNowConfig` when it's set to false here on this line.

packages/now-cli/src/util/dev/server.ts
styfle: I see, nice find! 👍

packages/now-cli/src/util/dev/server.ts
styfle: But if the first `if` branch (line 527) is false, wouldn't this second branch (line 536) also be false?

Or is there something that can happen in between asynchronously to assign the cached now config?

TooTallNate: `cachedNowConfig` may exist, but `canUseCache` is set to false, in which case we want to force the empty object.

styfle: I'm suggesting doing this instead:

```js
let config: NowConfig = {
  version: 2,
  [fileNameSymbol]: configPath,
};
```

styfle:
```suggestion
    let config: NowConfig = canUseCache ? this.cachedNowConfig : {
```

Ok I think I understand now. Does this make more sense?

TooTallNate: Your other suggestion is better I think. See ed0b88b.

styfle (approved):
:shipit:

AndyBitz (approved):

PR #4643, opened 2020-06-12
[tests] Don't revoke `VERCEL_TOKEN` after tests run

MaxLeiter:
`VERCEL_TOKEN` will no longer be invalidated during tests

Story: https://app.clubhouse.io/vercel/story/2281

### DIFF
--- a/packages/now-cli/test/integration.js
+++ b/packages/now-cli/test/integration.js
@@ -243,8 +243,10 @@ test.after.always(async () => {
     loginApiServer.close();
   }
 
-  // Make sure the token gets revoked
-  await execa(binaryPath, ['logout', ...defaultArgs]);
+  // Make sure the token gets revoked unless it's passed in via environment
+  if (!process.env.VERCEL_TOKEN) {
+    await execa(binaryPath, ['logout', ...defaultArgs]);
+  }
 
   if (tmpDir) {
     // Remove config directory entirely


packages/now-cli/test/integration.js
```diff
@@ -243,8 +243,10 @@ test.after.always(async () => {
     loginApiServer.close();
   }
 
-  // Make sure the token gets revoked
-  await execa(binaryPath, ['logout', ...defaultArgs]);
+  // Make sure the token gets revoked unless it's passed in via environment
+  if (!process.env.VERCEL_TOKEN) {
```
TooTallNate, line 249: I'm pretty sure we set `VERCEL_TOKEN` in GitHub Actions as well, so this check isn't good. I think we need to check for `CI` instead:

```suggestion
  if (process.env.CI) {
```

styfle: We don't use `VERCEL_TOKEN` in CI, instead we use two other env vars:

https://github.com/vercel/vercel/blob/63cc9009c893612dcfb95f4d8c450523eab270fb/test/lib/deployment/now-deploy.js#L170-L180

The only reason for `VERCEL_TOKEN` is running locally.

TooTallNate: Ok, then disregard my suggestion. Code LGTM.

TooTallNate (approved):

styfle (approved):
Thanks!

Issue #3777, opened 2020-02-10
Now CLI 17.0.3 seems to ignore rewrites in dev


wiesson:
I have noticed that `now dev` (17.0.3) ignores the rewrites from the `now.json` config file. I have downgraded to 16.7.3 and it works as expected. I'm using the rewrites to proxy requests through our domain

```
{
  "version": 2,
  "builds": [
    {
      "src": "next.config.js",
      "use": "@now/next"
    }
  ],
  "rewrites": [
    {
      "source": "/api/v1/graphql",
      "destination": "https://route.to/our/backend/api/endpoint"
    }
  ]
}
```

Maybe this was deprecated or there is another breaking changes, but it's quite hard to see what is actually breaking change, the 17.0.0 release just says:

> Now CLI 17.0.0 is here! This includes a few breaking changes including Project Linking.

A few means, there are more breaking changes? :)

mirshko:
Just having this issue too.

Im also not able to get env vars in my project. This is a serious blocker for me.

They only work in production. Now Dev is ignoring it all.

wiesson:
@mirshko - regarding env vars -> https://github.com/zeit/now/issues/3758

glennreyes:
Redirects also seem to be ignored with 17.0.3.

willemliu:
For me it seems to not work with `now dev` and production as well. It simply feels like the rewrites, redirects, headers are completely ignored.
I asked this on the Zeit Spectrum as well. https://spectrum.chat/zeit/now/now-dev-v17-0-3-now-json-routes-give-404~d8c8ce9b-c655-4a2f-b593-968f2f090886

I need the rewrites part to work because I make use of next-offline which generates a service-worker in a particular place. So I need to rewrite `/service-worker.js` to something like `/_next/static/service-worker.js`.

Perhaps this bug is related to this one: https://github.com/zeit/now/issues/3539

wiesson:
I haven't had the time to debug the issue and I simply switched back to `16.7.3` to get things done. But I might look into it soon.

DarthRukian:
Still not working in 17.0.4 rewrites and headers (to set custom content-type) are not working in `now dev` .
Any chance to be fixed soon ?

Thanks

HP4k1h5:
Ive been having similar issues since very recently. [post](https://spectrum.chat/zeit/now/router-external-target-error~af8633d5-8c6f-4891-9445-249b94c9c569)
[...]

PR #3673, opened 2020-01-27
[now-cli] Use `devCommand` from project settings locally

lucleray:
Example:
image

The `devCommand` from project settings is used if:
- the project is linked (`.now` exists)
- `project.devCommand` is not `null`
- `project.devCommand` is null but `project.framework` is not `null`

When `devCommand` is used, these builders are filtered out:
- `@now/static-build`
- `@now/next`

PR #4644, opened 2020-06-12
[cli] Fix `vc dev` proxy to upstream frontend dev server

styfle:
This PR fixes a longstanding issue introduced in #3673 that prevents routing properties from applying to the framework's upstream dev server.

This mimic's the older proxy logic used in build matches here: https://github.com/vercel/vercel/blob/5035fa537f82181649d42778c60145d8890f3291/packages/now-cli/src/util/dev/server.ts#L1535-L1539

- Related to #3777
- Related to #4510

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -1503,16 +1503,15 @@ export default class DevServer {
     if (!match) {
       // If the dev command is started, then proxy to it
       if (this.devProcessPort) {
-        debug('Proxying to frontend dev server');
+        const upstream = `http://localhost:${this.devProcessPort}`;
+        debug(`Proxying to frontend dev server: ${upstream}`);
         this.setResponseHeaders(res, nowRequestId);
-        return proxyPass(
-          req,
-          res,
-          `http://localhost:${this.devProcessPort}`,
-          this,
-          nowRequestId,
-          false
-        );
+        const origUrl = url.parse(req.url || '/', true);
+        //delete origUrl.search;
+        origUrl.pathname = dest;
+        Object.assign(origUrl.query, uri_args);
+        req.url = url.format(origUrl);
+        return proxyPass(req, res, upstream, this, nowRequestId, false);
       }
 
       if (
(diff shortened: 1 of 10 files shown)


packages/now-cli/src/util/dev/server.ts
```diff
@@ -1503,16 +1503,15 @@ export default class DevServer {
...
-        return proxyPass(
-          req,
-          res,
-          `http://localhost:${this.devProcessPort}`,
-          this,
-          nowRequestId,
-          false
-        );
+        const origUrl = url.parse(req.url || '/', true);
+        //delete origUrl.search;
```
TooTallNate: How come this one is commented out? Let's delete if it's problematic?

styfle: I wasn't sure the purpose of this part. I noticed the search string is deleted in the build match here:

https://github.com/vercel/vercel/blob/5035fa537f82181649d42778c60145d8890f3291/packages/now-cli/src/util/dev/server.ts#L1535-L1539

Should we delete for the dev proxy too?

TooTallNate: I think the logic should be consistent. I think it is required so that `url.format()` respects the `query` object instead of the `search` string (which would be outdated if there are `uri_args` defined).

TooTallNate: Confirmed it's necessary:

```js
> var u = url.parse('/foo?bar=1', true);
> u.query.a='b'
> url.format(u)    // Bad, should have "a=b"
'/foo?bar=1'
> delete u.search
> url.format(u)
'/foo?bar=1&a=b'   // Good after delete `search`
```

TooTallNate:
```suggestion
        delete origUrl.search;
```

TooTallNate: We should probably add a test for this behavior as well.

styfle: Nice find!

AndyBitz (approved):

TooTallNate (approved):

PR #3308, opened 2019-11-13
[now-static-build] Add Gatsby optimizations back

lucleray:
- [x] Revert gatsby optimizations back
- [x] Add breaking test case
- [x] Fix the test case
- [x] Put gatsby redirects and caching headers behind a feature flag

The issue was the one described here: https://github.com/gatsbyjs/gatsby/issues/19150. In version 2.17.7, Gatsby enforced local plugins to have a `index.js` file in their root to resolve. The fix is to simply add an empty `index.js` file.

PR #4645, opened 2020-06-12
[static-build] Apply Gatsby default caching routes automatically

### LINKED ISSUE #3331: Default Gatsby configuration provides no cache options
Hello!

I have just deployed a static gatsby page using your service and basically none of the static content is cached.
Video `mp4` cache: `cache-control: s-maxage=0`
images etc
```
cache-control: public, max-age=0, must-revalidate
```

I see this MR probably breaks it - https://github.com/zeit/now/pull/3212

I will use for now https://github.com/universse/gatsby-plugin-zeit-now

pacocoursey:
CH-702
Fixes #3331

If the file generated by `gatsby-plugin-now` does not exist, we'll set up these routes automatically.


packages/now-static-build/src/frameworks.ts
```diff
@@ -42,26 +40,33 @@ const frameworkList: Framework[] = [
     dependency: 'gatsby',
     buildCommand: 'gatsby build',
     getOutputDirName: async () => 'public',
-    defaultRoutes: async (dirPrefix: string) => {
-      try {
-        const nowRoutesPath = join(
-          dirPrefix,
-          'public',
-          '__now_routes_g4t5bY.json'
```
styfle: I think we might need to keep this file for backwards compatibility.

@lucleray What do you think?

lucleray: It's not used by a lot of users, but I'd still keep it 👍

lucleray: It's used in:
https://www.npmjs.com/package/gatsby-plugin-now
https://www.npmjs.com/package/gatsby-plugin-zeit-now

packages/now-static-build/src/frameworks.ts
```diff
@@ -42,26 +40,33 @@ const frameworkList: Framework[] = [
...
-        const nowRoutes = JSON.parse(content);
-        try {
-          await unlinkPromise(nowRoutesPath);
-        } catch (err) {
-          // do nothing if deleting the file fails
-        }
-        return nowRoutes;
-      } catch (err) {
-        // if the file doesn't exist, we don't create routes
-        return [];
```
lucleray, line 62: I suggest returning the default routes you added here, and keep the old behavior if `__now_routes_g4t5bY.json` exists.

lucleray (changes requested):
Can you add a test for the headers? You might be able to copy the tests we had before: https://github.com/vercel/vercel/pull/3308/files#diff-83715bd09c6fbf83efb4fc7fac8f51f1R16-R58

lucleray (approved):
Thanks!

styfle (approved):
Thanks, nice work!

PR #4659, opened 2020-06-15
Fix to not render no token error, prompt login instead

nkzawa:
When `auth.json` exists but no `token` property exists in it, we currently render error like the following screenshot.
Screen Shot 2020-06-15 at 21 32 55

This PR fixes to prompt login instead.
Screen Shot 2020-06-15 at 21 28 09


lucleray (dismissed):

TooTallNate (commented):
Can you add an integration test for this?

nkzawa:
@TooTallNate done

lucleray (approved):

TooTallNate (approved):

PR #4662, opened 2020-06-15
[go] Implement `startDevServer()` function

TooTallNate:
Due to Go's peculiar package importing structure, there's unfortunately no way to _directly_ invoke a Go serverless file, so the approach for this `startDevServer()` function is:

 * Create a temp directory in `.vercel/cache`
 * Run the `analyze` Go program on the entrypoint file to determine the `functionName`
 * Copy the entrypoint file into the temp directory, while modifying it to use `package main`
 * Copy the `dev-server.go` file into the temp directory, replacing the placeholder function name with the `functionName` retrieved from the analysis step from before
 * Execute `go run $tmp_dir` to spawn the Go HTTP server on an ephemeral port
 * Return the Go process id and port number for `vercel dev` to proxy to
 * After the Go process has exited, delete the temp directory


packages/now-go/build.sh
```diff
@@ -1,4 +1,16 @@
...
+# Build with `ncc`
 ncc build index.ts -e @vercel/build-utils -e @now/build-utils -o dist
 ncc build install.ts -e @vercel/build-utils -e @now/build-utils -o dist/install
+
+# Move `install.js` to dist
 mv dist/install/index.js dist/install.js
 rm -rf dist/install
+
+# For some reason `ncc` isn't bundling this file
+cp dev-server.go dist
```
styfle: Should we try bumping ncc to the latest version?

packages/now-go/index.ts
```diff
@@ -436,4 +460,111 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
...
+    stdio: ['ignore', 'inherit', 'inherit', 'pipe'],
+  });
+
+  child.once('exit', async () => {
+    await remove(tmp); // Cleanup
+  });
+
+  const portPipe = child.stdio[3];
+  if (!isReadable(portPipe)) {
+    throw new Error('Not readable');
```
styfle:
```suggestion
    throw new Error('File descriptor 3 is not readable');
```

Can we improve this error message? Is there an action we can give the user like `run with sudo` or `change your system to X`?

packages/now-go/index.ts
```diff
@@ -436,4 +460,111 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
...
+  const patched = data.replace('__HANDLER_FUNC_NAME', functionName);
+
+  await writeFile(join(dest, 'vercel-dev-server-main.go'), patched);
+}
+
+export async function startDevServer(
+  opts: StartDevServerOptions
+): Promise<StartDevServerResult> {
+  const { entrypoint, workPath, meta = {} } = opts;
+  const { devCacheDir = join(workPath, '.vercel', 'cache') } = meta;
```
styfle, line 496:
```suggestion
  const { devCacheDir = join(workPath, '.now', 'cache') } = meta;
```

The only time this will be empty is on old versions of Now CLI so I think it should fallback to `.now`, right? Like the we do here:

https://github.com/vercel/vercel/blob/93a9e5bed36c2db7bb907035b5b372529272f4d5/packages/now-node/src/index.ts#L487

packages/now-go/build.sh
TooTallNate: I think this line isn't necessary anymore actually. I'll double-check.

packages/now-go/index.ts
TooTallNate: Not really, this is basically here to appease TypeScript. Based on the `stdio` configuration passed to `spawn()`, `child.stdio[3]` should _always_ be populated.

packages/now-go/index.ts
TooTallNate: Well, again, this default is really only here to appease TypeScript since `devCacheDir` is an optional property on the `Meta` object. In reality, `startDevServer()` functionality isn't even implemented in CLI until _after_ the `devCacheDir` was implemented, so practically, this default will never be used.

packages/now-go/build.sh
TooTallNate: Removed.

packages/now-go/index.ts
```diff
@@ -436,4 +460,111 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
...
+  const tmpPackage = join(tmp, entrypointDir);
+  await mkdirp(tmpPackage);
+
+  let goModAbsPathDir = '';
+  if (await pathExists(join(workPath, 'go.mod'))) {
+    goModAbsPathDir = workPath;
+  }
+  const analyzedRaw = await getAnalyzedEntrypoint(entrypoint, goModAbsPathDir);
+  if (!analyzedRaw) {
+    throw new Error(
```
styfle, line 525: Can this throw `NowBuildError` instead which has a `link` property.

packages/now-go/index.ts
```diff
@@ -436,4 +460,111 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
...
+  const { entrypoint, workPath, meta = {} } = opts;
+  const { devCacheDir = join(workPath, '.vercel', 'cache') } = meta;
+  const entrypointDir = dirname(entrypoint);
+
+  const tmp = join(
+    devCacheDir,
+    'go',
+    Math.random()
+      .toString(32)
+      .substring(2)
```
styfle, line 511: Why random instead of the entrypoint name?

TooTallNate: Because more than one HTTP request could be executed simultaneously, in which case we don't want a second HTTP request to be overwriting the files in the directory that is being read for the first HTTP request (potentially leading to reading an empty/incomplete file and a runtime error).

styfle (commented):
Looks good! Just needs dev integration tests so we'll probably need to install Go in the GitHub Actions

packages/now-go/index.ts
TooTallNate: It was copy/pasted from the `build()` function. I think we should do it in a follow up PR since it'll require a `docs` change as well.

packages/now-cli/test/dev/integration.js
```diff
@@ -1619,3 +1619,13 @@ test(
     await testPath(200, '/index.css', 'This is index.css');
   })
 );
+
+test(
+  '[vercel dev] Should support `*.go` API serverless functions',
+  testFixtureStdio('go', async testPath => {
+    await testPath(200, `/api`, 'This is the index page');
+    await testPath(200, `/api/another`, 'This is another page');
```
styfle, line 1629: Let's also add the `.go` extension just to be sure:

```js
await testPath(200, '/api/another.go', 'This is another page');
```

packages/now-go/go-helpers.ts
```diff
@@ -127,50 +126,35 @@ export async function downloadGo(
...
+  const isGoExist = await pathExists(join(dir, 'bin'));
+  if (!isGoExist) {
+    debug('Installing `go` v%s to %o for %s %s', version, dir, platform, arch);
+    const url = getGoUrl(version, platform, arch);
+    debug('Downloading `go` URL: %o', url);
+    const res = await fetch(url);
 
-    throw new Error(
-      `Your current ${stdout} doesn't support Go Modules. Please update.`
-    );
```
styfle, line 144: I don't see this error anymore:

```
Your current ${stdout} doesn't support Go Modules. Please update.
```

Don't we still need to throw if an old version of Go was found?

TooTallNate: If the system `go` binary is too old then this function will now fall back to downloading the requested `go` binary, basically the same behavior as if no system `go` binary was installed.

styfle (approved):
🐹 ⛳

AndyBitz (approved):

MaxLeiter (approved):

PR #4539, opened 2020-05-30
[frameworks][examples] Add RedwoodJS

lachlanjc:
Adding Redwood to the frameworks list, static builders file, & an example app for it.

Waiting on the creation of a Redwood builder to merge.

PR #4676, opened 2020-06-17
[redwoood] Add runtime `@vercel/redwood`

styfle:
### Changes
- Add new Runtime: `@vercel/redwood`
- Update `run.js` script to sort packages properly

### Related
- Related to #4539 
- Related to #4535
- Depends on https://github.com/apollographql/apollo-server/pull/4311


packages/redwood/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const apiDistPath = join(workPath, 'api', 'dist', 'functions');
+  const webDistPath = join(workPath, 'web', 'dist');
+  const lambdaOutputs: { [filePath: string]: LambdaType } = {};
+  const staticOutputs = await glob('**', webDistPath);
+
+  // Each file in the `functions` dir will become a lambda
+  const functionFiles = await glob('*.js', apiDistPath);
+
+  console.log({ functionFiles });
```
AndyBitz: Should we keep this line?

packages/redwood/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    );
+    const relativeEntrypoint = relative(workPath, absEntrypoint);
+    const awsLambdaHandler = getAWSLambdaHandler(relativeEntrypoint, 'handler');
+
+    console.log({
+      outputName,
+      absEntrypoint,
+      relativeEntrypoint,
+      awsLambdaHandler,
+    });
```
AndyBitz: Should we keep the `console.log`?

packages/redwood/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      }),
+    };
+
+    dependencies.forEach(fsPath => {
+      lambdaFiles[relative(workPath, fsPath)] = new FileFsRef({ fsPath });
+    });
+
+    console.log(
+      'adding entrypoint file ' + relative(workPath, fileFsRef.fsPath)
+    );
```
AndyBitz: Same as above?

AndyBitz (dismissed):
Code looks great 👍 

Just noticed a few `console.log` calls, not sure if you want to keep those @styfle

AndyBitz (dismissed):

AndyBitz (dismissed):

packages/redwood/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+  debug('Running build script...');
+  await spawnAsync('yarn', ['rw', 'build'], {
+    ...spawnOpts,
+    cwd: workPath,
+    prettyCommand: 'yarn rw build',
+  });
+
+  const {
+    buildCommand = 'yarn rw db up --no-db-client --auto-approve && yarn rw build',
+  } = config;
```
styfle, line 93: @AndyBitz Is `config.buildCommand` the correct way to get the project settings Build Command?

AndyBitz: Yes 👍

packages/redwood/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export async function prepareCache({
+  workPath,
+}: PrepareCacheOptions): Promise<Files> {
+  const cache = await glob('node_modules/**', workPath);
+  return cache;
+}
+
+export { shouldServe };
```
TooTallNate: I don't think the default `shouldServe()` will work in this case, as it's only really applicable to `version = 3` Runtimes. Did you test this with `vc dev`?

AndyBitz (dismissed):

utils/run.js
```diff
@@ -1,29 +1,37 @@
 const { execSync, spawn } = require('child_process');
 const { join, relative, sep } = require('path');
-const { readdirSync } = require('fs');
 
-if (
```
TooTallNate, line 5: How come this logic is removed?

styfle: It was only necessary when we would deploy forks during testing but we haven't done that for awhile since changing the token factory to the registration api.

I can add it back if you think its better to show green when someone is testing in their local fork?

TooTallNate: Ok makes sense.

packages/redwood/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@vercel/redwood",
+  "version": "0.0.1",
+  "main": "./dist/index.js",
+  "license": "MIT",
+  "homepage": "https://vercel.com/docs/runtimes#official-runtimes/redwood",
```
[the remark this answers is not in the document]
styfle: Hmm, maybe I should remove this since its unlikely to be documented here since `@vercel/static-build` isn't either

packages/redwood/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euo pipefail
+
+bridge_defs="$(dirname $(pwd))/now-node-bridge/src/bridge.ts"
+launcher_defs="$(dirname $(pwd))/now-node/src/launcher.ts"
+
+cp -v "$bridge_defs" src
+cp -v "$launcher_defs" src
+
+tsc
```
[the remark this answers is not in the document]
styfle: I'll try it out with ncc and see how it goes 👍

packages/redwood/src/index.ts
styfle: I haven't thought through `vc dev` yet because the redwood dev server provides serverless functions. I guess I'll have to add proxy support here to make sure everything goes through to the redwood dev server

packages/redwood/build.sh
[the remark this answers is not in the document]
styfle: So the size actually goes up with `ncc`, I'm not sure if we want to do this:

- Without `ncc`: 24K publish + [63MB dependencies](https://packagephobia.com/result?p=@netlify/zip-it-and-ship-it)
- With `ncc`: 117MB publish + 0 dependencies

It looks like this is because typescript is bundled twice 🤔

packages/redwood/src/index.ts
styfle: Ok I removed `shouldServe` and returned `routes` and `watch` when `meta.isDev` is true

packages/redwood/build.sh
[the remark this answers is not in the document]
TooTallNate: Omg!

packages/redwood/package.json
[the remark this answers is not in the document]
TooTallNate: Neither is Next.js (only the Serverless Function Runtimes are documented here).

TooTallNate (approved):

AndyBitz (approved):

PR #4678, opened 2020-06-18
[cli] Do not confirm `vc env pull` overwrite if created by CLI

nkzawa:
Append `# Created by Vercel CLI` to the head of `.env` file and automatically overwrite the file if it's there next time without confirmation.

https://app.clubhouse.io/vercel/story/316


lucleray (commented):
Can we add a test to check that `now env pull` doesn't overwrite a `.env` file created by the user?

lucleray (dismissed):

packages/now-cli/test/integration.js
```diff
@@ -529,12 +529,46 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+      reject: false,
+      cwd: target,
+    });
+
+    await waitForPrompt(vc, chunk =>
+      chunk.includes('Found existing file ".env". Do you want to overwrite?')
+    );
+    vc.stdin.write('y\n');
+
+    const { exitCode, stderr, stdout } = await vc;
```
lucleray, line 568: Somehow this line hangs 🤔

lucleray: It looks like it's a bug with `execa`

lucleray: You can use the same pattern like the one used for the test on line 1404, alternatively

lucleray (dismissed):

packages/now-cli/src/commands/env/pull.ts
```diff
@@ -35,10 +48,21 @@ export default async function pull(
...
   const [filename = '.env'] = args;
   const fullPath = join(process.cwd(), filename);
-  const exists = existsSync(fullPath);
   const skipConfirmation = opts['--yes'];
 
-  if (
+  let head;
+  try {
+    head = readHeadSync(fullPath, Buffer.byteLength(CONTENTS_PREFIX));
+  } catch (err) {
```
styfle: Can this try/catch be moved into the `readHeadSync` function?

nkzawa: I think it's better not to since it's not something you expect if you're used to `fs`. (if the file does not exist, should throw error with `ENOENT`)

styfle (dismissed):

MaxLeiter (dismissed):

packages/now-cli/src/commands/env/pull.ts
styfle: You could rename the function to something like `tryReadHead()`

lucleray (approved):

styfle (approved):

MaxLeiter (approved):

TooTallNate (approved):

PR #4682, opened 2020-06-18
[next] Update tests to run against latest canary

ijjk:
In the latest canary of Next.js pages are no longer nested under the `BUILD_ID` folder and instead are nested under a hash for the page bundle content. To prevent these tests from breaking too often from changes in canary this updates to locate the page bundle using the `buildManifest`. This also updates our latest SSG fixture to test against the latest canary to help ensure the feature doesn't break with a new canary release


packages/now-next/test/fixtures/16-base-path/next.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   generateBuildId() {
     return 'testing-build-id';
   },
+  // TODO: remove experimental after next canary release
```
Timer: Time for this to be dropped

ijjk: Updated

Timer (approved):

TooTallNate (approved):

PR #4694, opened 2020-06-22
[cli] Improve validation error message for vercel.json

styfle:
This PR improves the validation error message when the user has an invalid `vercel.json` file.

Previously, the error message did not account for nested properties so `{"foo": "will error"}` looked fine because it would mention there is an additional property `foo`. However, the error message for `{ "routes": [{ "foo": "will error" }] }` did not mention anything about `routes` when it explaining there was an additional property `foo`. This became more apparent as we added nested properties for `rewrites` and `redirects` (see tests in this PR).

This PR also adds suggestions for common mistakes such as `src` vs `source`.

### DIFF
--- a/packages/now-cli/package.json
+++ b/packages/now-cli/package.json
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "preinstall": "node ./scripts/preinstall.js",
-    "test-unit": "nyc ava test/unit.js test/dev-builder.unit.js test/dev-router.unit.js test/dev-server.unit.js --serial --fail-fast --verbose",
+    "test-unit": "nyc ava test/unit.js test/dev-builder.unit.js test/dev-router.unit.js test/dev-server.unit.js test/dev-validate.unit.js --serial --fail-fast --verbose",
     "test-integration-cli": "ava test/integration.js --serial --fail-fast --verbose",
     "test-integration-v1": "ava test/integration-v1.js --serial --fail-fast",
     "test-integration-dev": "ava test/dev/integration.js --serial --fail-fast --verbose",
--- a/packages/now-cli/src/commands/deploy/latest.js
+++ b/packages/now-cli/src/commands/deploy/latest.js
@@ -3,12 +3,12 @@ import bytes from 'bytes';
 import { join } from 'path';
 import { write as copy } from 'clipboardy';
 import chalk from 'chalk';
-import title from 'title';
 import { fileNameSymbol } from '@vercel/client';
 import Client from '../../util/client';
 import { handleError } from '../../util/error';
 import getArgs from '../../util/get-args';
 import toHumanPath from '../../util/humanize-path';
+import humanizeAjvError from '../../../src/util/humanize-ajv-error';
 import Now from '../../util';
 import stamp from '../../util/output/stamp.ts';
 import createDeploy from '../../util/deploy/create-deploy';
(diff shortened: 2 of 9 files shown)


packages/now-cli/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "preinstall": "node ./scripts/preinstall.js",
-    "test-unit": "nyc ava test/unit.js test/dev-builder.unit.js test/dev-router.unit.js test/dev-server.unit.js --serial --fail-fast --verbose",
+    "test-unit": "nyc ava test/unit.js test/dev-builder.unit.js test/dev-router.unit.js test/dev-server.unit.js test/dev-validate.unit.js --serial --fail-fast --verbose",
```
styfle, line 15: @TooTallNate I added a unit test file for this but I realized that we already have some integration tests for error messages like this. Should I move all my new tests to dev fixtures instead of unit tests?

TooTallNate: Good question… it makes sense to me to have comprehensive unit tests for many scenarios, but only a few integration tests to make sure the code path(s) to print the validation errors in the CLI are working as expected.

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -738,17 +738,15 @@ function handleCreateDeployError(output, error, localConfig) {
...
   }
   if (error instanceof SchemaValidationFailed) {
-    const { message, params, keyword, dataPath } = error.meta;
-
+    const ajvError = error.meta;
+    const fileName = localConfig[fileNameSymbol];
+    const humanError = humanizeAjvError(ajvError, fileName);
+    output.prettyError(humanError);
+    return 1;
+    /*
```
TooTallNate: How come this is being commented out / not deleted?

packages/now-cli/src/util/dev/server.ts
```diff
@@ -542,7 +532,7 @@ export default class DevServer {
       configPath = getNowConfigPath(this.cwd);
       this.output.debug(`Reading ${configPath}`);
       config = JSON.parse(await fs.readFile(configPath, 'utf8'));
-      config[fileNameSymbol] = configPath;
+      config[fileNameSymbol] = basename(configPath);
```
TooTallNate, line 535: This should probably be a path relative to `this.cwd`, I think for the case where a `--local-config` flag is passed (so it might not be at the root level).

packages/now-cli/src/commands/deploy/latest.js
styfle: Thanks I forgot to implement this part since there was no test. I just added a "suggestions" feature which should cover this use case.

packages/now-cli/src/util/dev/server.ts
styfle: Hmm, this error message is starting to become very verbose 🤔 

I found that we are already calling `basename()` for `vc deploy` so this is really a bug fix for `vc dev`.

https://github.com/vercel/vercel/blob/0986f4bcb6e6f63a756b45c153479893269052c4/packages/now-cli/src/util/config/files.ts#L131

After some thought, I think its best to use the file name here because thats what the variable is called, not file path. In the case when a user passes `--local-config`, then there error message will be directly below the path they used with the flag so there is no reason to print the path again. I think it is sufficient to be a simple file name and will look good when someone uses `vercel.prod.json` for example.

TooTallNate: Sounds good.

TooTallNate (approved):

MaxLeiter (approved):

PR #4697, opened 2020-06-22
[cli] Remove caching of `vercel.json` config in `vc dev`

TooTallNate:
Since the filesystem watcher may be slow, it's actually faster and more
reliable to simply re-read the `vercel.json` configuration for every
HTTP request. This also simplifies the logic as an added benefit.

Some `sleep()` calls are removed from relevant tests that were
previously necessary due to the lag in the filesystem watcher.

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -122,7 +122,6 @@ export default class DevServer {
   public address: string;
   public devCacheDir: string;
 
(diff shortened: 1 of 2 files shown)


styfle (approved):
Seems good if the tests pass! Less global state always good 👍 

I'm adding @MaxLeiter too since this could cause conflicts in the `vc dev` cloud env vars PR.

MaxLeiter (approved):

PR #4701, opened 2020-06-23
[cli][client] Add source prop when deploying

nkzawa:
Only for v2 deployments.

https://app.clubhouse.io/vercel/story/3190

We need to release this after API change.


packages/now-cli/src/util/index.js
```diff
@@ -145,6 +145,7 @@ export default class Now extends EventEmitter {
       regions,
       target: target || undefined,
       projectSettings,
+      source: 'cli',
```
styfle, line 148: This information is already in the the `User-Agent` header. I don't think we need a new property to detect the CLI, we just need the API to save the User-Agent or read it to see if it contains `vercel` or `now` https://github.com/vercel/vercel/blob/master/packages/now-cli/src/util/ua.ts

In fact, it might be better to store `vercel` or `now` so we can see the usage of each CLI.

nkzawa: This is because user-agent is not enough for other cases. I think it's better to not have special cases for handling CLI requests on API, not strong opinion tho.

styfle (approved):

packages/now-cli/src/util/index.js
styfle: Looks good now that the API is updated 👍

TooTallNate (approved):

PR #4704, opened 2020-06-23
[next] Update revalidate tests and correct shared lambdas revalidate

### LINKED ISSUE #14497: Revalidate fails with @vercel/next shared lambda mode
When leveraging the new shared lambdas mode in `@vercel/next` and `unstable_revalidate` in Next.js the revalidation encounters an error and no logs are returned. We need to resolve this and ensure tests capture this to prevent regression before landing stable

ijjk:
This adds more fine grained tests for our revalidate behavior to ensure pages are actually being updated instead of only checking the headers, it also corrects page lookups failing with shared lambdas enabled for dynamic pages using revalidate. 

Closes: https://github.com/vercel/next.js/issues/14497


TooTallNate (approved):

Timer (approved):
LGTM if tests are passing

PR #4705, opened 2020-06-23
[cli][build-utils] Refactor ajv error message

styfle:
This PR is a follow up to #4694 so we can use this same helper function in the API.

I tried to follow the existing error message format so that the API can use a different prefix than the CLI.

The API currently returns a 400 with `Invalid request` when a Git deploy fails validation:

api-comment


TooTallNate (approved):

packages/now-cli/test/dev-validate.unit.js
```diff
@@ -68,10 +68,7 @@ test('[dev-validate] should error with invalid routes array type', async t => {
     routes: { src: '/(.*)', dest: '/api/index.js' },
   };
   const error = validateConfig(config);
-  t.deepEqual(
-    error.message,
-    'Invalid vercel.json - property `routes` should be of type array.'
-  );
+  t.deepEqual(error.message, 'Invalid vercel.json - `routes` should be array.');
```
TooTallNate, line 71: All of these "should be" errors would read better to me if we prefixed "a"/"an" before the type. i.e. "should be _an_ array".

AndyBitz (approved):

PR #4714, opened 2020-06-24
[cli] Fix `vc dev` serving pure static dot files

styfle:
Fixes ch1233 and discussion #4475

### DIFF
--- a/packages/now-cli/test/dev/integration.js
+++ b/packages/now-cli/test/dev/integration.js
@@ -656,6 +656,7 @@ test(
   testFixtureStdio('public-and-api', async testPath => {
     await testPath(200, '/', 'This is the home page');
     await testPath(200, '/about.html', 'This is the about page');
+    await testPath(200, '/.well-known/humans.txt', 'We come in peace');
     await testPath(200, '/api/date', /current date/);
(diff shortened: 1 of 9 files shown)


TooTallNate (dismissed):

packages/now-cli/test/dev/integration.js
```diff
@@ -656,6 +656,7 @@ test(
   testFixtureStdio('public-and-api', async testPath => {
     await testPath(200, '/', 'This is the home page');
     await testPath(200, '/about.html', 'This is the about page');
+    await testPath(200, '/.well-known/humans.txt', 'We come in peace');
```
AndyBitz, line 659: Should we also add a test for e.g. `.env`?

styfle: I wanted to add a test for a real file that a user expects to be served like in #4475 so that's why its `.well-known/humans.txt`.

Do users want to serve the `.env` file? Also, what should the content-type be?

TooTallNate (dismissed):

AndyBitz (dismissed):

AndyBitz (approved):

TooTallNate (approved):

PR #4715, opened 2020-06-24
[cli] Use the configured "Output Directory" from project's settings in `vc dev`

TooTallNate:
Previously, when a project has a specific "Output Directory" configured in the project settings page, it was not being correctly considered when using `vercel dev`.

The main fix here is passing in the full project settings object (including the `outputDirectory`) to the `detectBuilders()` function.

Also cleaned up a few types and updated the error message that was previously being rendered to use a short link.


AndyBitz (approved):

packages/now-cli/test/helpers/prepare.js
```diff
@@ -500,12 +500,7 @@ CMD ["node", "index.js"]`,
       }),
     },
     'project-link': {
-      'pages/index.js': 'export default () => <div><h1>Now CLI test</h1></div>',
-      'package.json': JSON.stringify({
-        dependencies: {
-          gatsby: 'latest',
-        },
-      }),
```
styfle, line 509: I think you should leave gatbsy here to prove that the project settings will override the default framework settings, right?

TooTallNate: The problem was that gatsby has a default dev command, which we want to avoid so that the build command is run. I think it's fine to remove gatsby since the build command gets overwritten and executed in both prod and dev, which is what we're testing for here.

packages/now-cli/src/commands/dev/dev.ts
```diff
@@ -50,21 +51,26 @@ export default async function dev(
...
       devCommand = project.devCommand;
     } else if (project.framework) {
       const framework = frameworks.find(f => f.slug === project.framework);
 
       if (framework) {
-        frameworkSlug = framework.slug;
-        const defaults = framework.settings.devCommand;
+        if (framework.slug) {
+          frameworkSlug = framework.slug;
+        }
```
styfle, line 71: I'm curious, why add the `if` statement here?

Assigning `frameworkSlug = framework.slug` every time was causing an issue?

styfle (approved):

packages/now-cli/src/commands/dev/dev.ts
styfle: I see, its to convert null to undefined, seems fine 👍

PR #4728, opened 2020-06-26
[node] Set the TypeScript `target` in `vc dev`

TooTallNate:
Fixes: https://github.com/vercel/vercel/discussions/4724


packages/now-cli/test/dev/integration.js
```diff
@@ -1632,3 +1632,14 @@ test(
...
 );
+
+test(
+  '[vercel dev] Should set the `ts-node` "target" to match Node.js version',
+  testFixtureStdio(
+    'node-ts-node-target',
+    async testPath => {
+      await testPath(200, `/api`, '{"ok":true}');
+    },
+    { skipDeploy: true }
```
styfle: Why skip deploy? We should make sure it works correctly for both dev and prod?

styfle (commented):
Can we read the user’s tsconfig and use that target if available?

TooTallNate:
@styfle In this case I think we want to override actually - I don't see a good use-case for the user preferring a different value than what we are providing. Reason being that since we can check the Node.js version beforehand, we are providing the most accurate target based on the Node.js version, whereas the user is expected to have the _baseline_ "target" defined in their `tsconfig.json`, or an incompatible version listed such as "esnext" which might output syntax not compatible with the Node version being used.

styfle:
Imagine if the use set target to ES5. That means they would expect classes to not be available at runtime but if we bump to ES2018, the the classes will be preserved which might lead to unexpected errors at runtime.  

Furthermore, we don’t want dev to compile to a different target than prod.

PR #4748, opened 2020-06-30
[next] Skip Lambda Deployment for Static Sites

### LINKED ISSUE #14430: Vercel's Next.js builder should avoid uploading preview lambdas when it can
Currently, the Vercel Next.js builder will upload lambdas for preview mode even when we know it cannot be used.

In the case there are **no API routes**, we should skip the preview lambdas.

This should be extendable into the ISR lambda as well.

Timer:
This implements a deployment speed optimization by completely skipping lambdas in the event of a fully static site without API routes.

Without API routes, it's impossible to enter Preview Mode, so the Preview Mode lambdas are not necessary.

Fixes https://github.com/vercel/next.js/issues/14430

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -863,6 +863,7 @@ export const build = async ({
           [filePath: string]: FileFsRef;
         };
 
+    let canUsePreviewMode = false;
     let pseudoLayerBytes = 0;
     let apiPseudoLayerBytes = 0;
     const pseudoLayers: PseudoLayer[] = [];
@@ -890,6 +891,7 @@ export const build = async ({
       for (const page of allPagePaths) {
         if (isApiPage(page)) {
           apiPages.push(page);
+          canUsePreviewMode = true;
         } else {
           nonApiPages.push(page);
         }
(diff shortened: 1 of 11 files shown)


ijjk (dismissed):

packages/now-next/src/index.ts
```diff
@@ -1454,24 +1456,32 @@ export const build = async ({
...
-        bypassToken: prerenderManifest.bypassToken,
-      });
-      prerenders[outputPathData] = new Prerender({
-        expiration: initialRevalidate,
-        lambda,
-        fallback: jsonFsRef,
-        group: prerenderGroup,
-        bypassToken: prerenderManifest.bypassToken,
-      });
+      if (prerenders[outputPathPage] == null) {
```
ijjk, line 1467: Should this check be loose or can we just check for a truthy value like `if (prerenderes[outuptPathPage])`?

Timer: I think this if clause is fine as-is, really. All of them accomplish the same thing, the current one isn't "loose" imo. Only stricter would be `=== undefined`.

packages/now-next/test/fixtures/05-spr-support/pages/api/noop.js
```diff
@@ -0,0 +1,3 @@
+export default (_, res) => {
```
ijjk, line 1: An API route shouldn't be needed for revalidation support right?

Timer: It was specifically failing on the `/forever` route(s) that set revalidation to `false`. Our tests were asserting the `PRERENDER` header instead of `MISS` (new header).

Instead of updating the tests to check for `MISS` (which could let this regress into a lambda), I opted-into preview deployment so it used the Prerender type.

In the future, hits from L4 cache should respond different than `MISS` so we can appropriately detect it vs a lambda. This needs to be discussed with the CDN team.

ijjk (approved):

timneutkens (approved):

Timer:
Landing due to CI failing for unrelated reasons (Gatsby). It had passed full checks previously before syncing with `master`.

PR #4760, opened 2020-07-02
[cli] Reduce number of times `getNowConfig()` is invoked in `vercel dev`

TooTallNate:
When using the `vercel dev` command, the `getNowConfig()` function was
being invoked many times per HTTP-request. It should only be read _once_
per HTTP request as it may be an expensive function depending on the
project size.

This change reduces the number of times the function is called to once
per HTTP request, and passes around the resulting `nowConfig` object to
the functions that require it.


MaxLeiter (dismissed):

packages/now-cli/src/util/dev/server.ts
```diff
@@ -231,10 +229,9 @@ export default class DevServer {
     const nowConfig = await this.getNowConfig(false);
 
     // Update the env vars configuration
-    const nowConfigBuild = nowConfig.build || {};
     const [runEnv, buildEnv] = await Promise.all([
       this.getLocalEnv('.env', nowConfig.env),
-      this.getLocalEnv('.env.build', nowConfigBuild.env),
+      this.getLocalEnv('.env.build', nowConfig.build?.env),
```
TooTallNate, line 234: `prettier` was updated to v2 to support this `?` syntax.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -522,32 +520,17 @@ export default class DevServer {
...
-    let configPath = 'vercel.json';
-    let config: NowConfig = {
-      version: 2,
-      [fileNameSymbol]: configPath,
-    };
+    const configPath = getNowConfigPath(this.cwd);
 
-    try {
-      configPath = getNowConfigPath(this.cwd);
-      this.output.debug(`Reading ${configPath}`);
```
pacocoursey, line 537: Should we keep some of these debug messages?

pacocoursey (commented):
LGTM

packages/now-cli/src/util/dev/server.ts
TooTallNate: They're still there, moved into the `readJsonFile()` function.

MaxLeiter (approved):

pacocoursey (approved):

PR #4791, opened 2020-07-06
[next] Check all vercel.json files for routes when bailing shared lambdas

ijjk:
This updates the shared lambdas bail-out when `routes` are detected in `vercel.json` or `now.json` to check all `vercel.json` and `now.json` files in the project tree instead of just doing a find-up from the entry and checking the first result. 

Additional tests:

- `vercel.json` with `routes` basic [deploy](https://shared-lambda-tests-g1vjndti5.vercel.app/)
- `vercel.json` without `routes` and nested `vercel.json` with `routes` [deploy](https://shared-lambda-tests-butfkp8h2.vercel.app/)
- `vercel.json` with `routes` and nested `vercel.json` without `routes` [deploy](https://shared-lambda-tests-6mzzoh1gr.vercel.app/)
- `vercel.json` without routes and nested-non-up-tree `vercel.json` with `routes` [deploy](https://shared-lambda-tests-jdnqz744h.vercel.app/)


packages/now-next/src/index.ts
```diff
@@ -231,24 +231,32 @@ export const build = async ({
...
   let hasLegacyRoutes = false;
+  let possibleNowJsonPath = entryPath;
   const hasFunctionsConfig = !!config.functions;
 
-  if (nowJsonPath) {
-    const nowJsonData = JSON.parse(await readFile(nowJsonPath, 'utf8'));
+  while (possibleNowJsonPath.length >= workPath.length) {
+    const curFiles = await readdir(possibleNowJsonPath);
+    const nowJsonFile = curFiles.find(
+      (file) => file.endsWith('now.json') || file.endsWith('vercel.json')
```
Timer: Why the choice of endsWith here? Seems like it should be strict equality.

TooTallNate:
I don't really understand. Do we even support nested `vercel.json` files? The deployments aren't public so I can't see what the project file structure looks like.

packages/now-next/src/index.ts
ijjk: Was left over from the `find-up` implementation, updated

ijjk:
@TooTallNate this handles the edge case where users have a mono-repo set-up and might have a `now.json` or `vercel.json` nested inside of their project but it's not necessarily the one being used to deploy. Since we can't know which one is being used to deploy we check all of the ones up the tree and bail out if any of them have legacy routes that could break the shared lambda optimization instead of checking the first one which might not be being used and doesn't cause us to bail out when we need to

Timer (approved):

TooTallNate:
> Since we can't know which one is being used to deploy

Seems like this is the thing we should fix. Guessing is bad. Maybe we should expose the parsed `vercel.json` in the build options object.

Timer (changes requested):
I agree with @TooTallNate. We need to know the real config due to being able to use `now -A differentFileName.json`.

ijjk:
Closing this PR in favor of updating to provide the builder with the necessary info to know this without guessing. Opened new issue here https://github.com/vercel/next.js/issues/15175 for tracking

PR #4793, opened 2020-07-06
[cli] Make filesystem globbing faster in `vercel dev`

TooTallNate:
The current code uses `getAllProjectFiles()` which globs for every file
in the project, not considering the `.vercelignore` file and default
ignore list. For example, every file in the `node_modules` directory is
selected, just for `vercel dev` to manually ignore them afterwards,
which is very slow and wasteful. For a simple Next.js project, this
globbing was taking over 3 seconds on my machine 🤯!

Solution is to use the `staticFiles()` function which is the same as
`vercel dev` uses at boot-up time. This function properly considers the
ignore list and thus is much faster, and the manual filtering is no
longer necessary by `vercel dev`. For the same simple Next.js project,
this function takes on average 10 milliseconds.

Also did a bit of cleanup and removed the `getAllProjectFiles()`
function since it is no longer used anywhere.

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -49,7 +49,7 @@ import getNowConfigPath from '../config/local-path';
 import { MissingDotenvVarsError } from '../errors-ts';
 import cliPkg from '../pkg';
 import { getVercelDirectory } from '../projects/link';
-import { staticFiles as getFiles, getAllProjectFiles } from '../get-files';
+import { staticFiles as getFiles } from '../get-files';
 import { validateConfig } from './validate';
 import { devRouter, getRoutesTypes } from './router';
 import getMimeType from './mime-type';
@@ -89,6 +89,10 @@ interface FSEvent {
   type: string;
   path: string;
 }
+
+type WithFileNameSymbol<T> = T & {
+  [fileNameSymbol]: string;
+};
 
 function sortBuilders(buildA: Builder, buildB: Builder) {
   if (buildA && buildA.use && isOfficialRuntime('static-build', buildA.use)) {
@@ -292,7 +296,7 @@ export default class DevServer {
             filesRemovedArray
           ).catch((err: Error) => {
             this.output.warn(
-              `An error occurred while rebuilding ${match.src}:`
+              `An error occurred while rebuilding \`${match.src}\`:`
             );
             console.error(err.stack);
           });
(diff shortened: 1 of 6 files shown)


packages/now-cli/src/util/dev/server.ts
```diff
@@ -532,15 +536,6 @@ export default class DevServer {
       this.readJsonFile<NowConfig>(configPath),
     ]);
 
-    const allFiles = await getAllProjectFiles(this.cwd, this.output);
```
TooTallNate, line 535: This is the main fix.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -555,6 +550,11 @@ export default class DevServer {
       const featHandleMiss = true; // enable for zero config
       const { projectSettings, cleanUrls, trailingSlash } = config;
 
+      const opts = { output: this.output, isBuilds: true };
```
TooTallNate, line 554: This is also the main fix.

packages/now-cli/src/util/get-files.ts
```diff
@@ -1,7 +1,7 @@
-import { resolve, join } from 'path';
```
TooTallNate, line 1: Except for the removal of the `getAllProjectFiles()` function, the rest of the changes in this file are only cosmetic.

AndyBitz (approved):
🏅

AndyBitz (approved):

packages/now-cli/src/util/get-files.ts
```diff
@@ -415,24 +411,6 @@ export async function docker(
...
- * @return {Array} of {String}s with the found files
- */
-export async function getAllProjectFiles(cwd: string, { debug }: Output) {
-  // We need a slash at the end to remove it later on from the matched files
-  const current = join(resolve(cwd), '/');
-  debug(`Searching files inside of ${current}`);
-
-  const list = await glob('**', { cwd: current, absolute: true, nodir: true });
-
-  // We need to replace \ with / for windows
```
styfle, line 432: Does the new function work with Windows? I assume so but it’s good to test it out first to confirm.

styfle (approved):

packages/now-cli/src/util/get-files.ts
TooTallNate: It does. The `relative()` helper function that is used in `dev/server.ts` does this same replacement.

pacocoursey (approved):

MaxLeiter (approved):

PR #4807, opened 2020-07-07
[cli] Show helpful error when using unauthorized `.vercel` config

pacocoursey:
CH-3242

When a user has project configuration in `.vercel` that corresponds to a team that they are not a part of, we currently throw a 403 error. This adds special handling so that we can show a specific message when this happens.

I didn't want to update the return type of `getProjectByIdOrName` because we only want to handle this 403 in a special case, so I `throw` it instead.

Other changes are linter.


packages/now-cli/src/util/projects/link.ts
```diff
@@ -141,13 +141,25 @@ export async function getLinkedProject(
...
+  let org: Org | null = null;
+  let project: Project | ProjectNotFound | null = null;
   try {
     [org, project] = await Promise.all([
       getOrgById(client, link.orgId),
       getProjectByIdOrName(client, link.projectId, link.orgId),
     ]);
+  } catch (error) {
+    if (error instanceof ProjectUnauthorized) {
+      output.error(
```
TooTallNate: I think we want to call `spinner()` _before_ the `output.error()` otherwise part of the message may be cut off 🤔

packages/now-cli/src/util/errors-ts.ts
```diff
@@ -1179,6 +1179,16 @@ export class ProjectNotFound extends NowError<'PROJECT_NOT_FOUND', {}> {
   }
 }
 
+export class ProjectUnauthorized extends NowError<'PROJECT_UNAUTHORIZED', {}> {
```
TooTallNate: I _think_ we're moving away from these Error subclasses in favor of just instantiating a `NowBuildError()` instance (from `@vercel/build-utils`). This way, the `message` and `code` and `link` are all self-contained in the error instance. @styfle can confirm.

TooTallNate (commented):
An integration test would be appreciated 😺

packages/now-cli/test/helpers/prepare.js
```diff
@@ -519,6 +519,13 @@ CMD ["node", "index.js"]`,
       'vercel.json': getConfigFile(true),
       'now.json': getConfigFile(true),
     },
+    'unauthorized-vercel-config': {
+      // This project is under the testing-internal team
+      '.vercel/project.json': JSON.stringify({
+        orgId: 'team_JgimPl9u9uauL7E4MjMLt605',
+        projectId: 'QmRoBYhejkkmssotLZr8tWgewPdPcjYucYUNERFbhJrRNi',
+      }),
+    },
```
pacocoursey, line 528: Only added this test case, rest is lint.

packages/now-cli/test/integration.js
```diff
@@ -3077,3 +3067,24 @@ test('deploy gatsby twice and print cached directories', async t => {
...
+  );
+
+  t.is(exitCode, 1, formatOutput({ stderr, stdout }));
+  t.true(
+    stderr.includes(
+      'Could not retrieve Project Settings. To link your project again, run `rm -rf .vercel` and `vercel`.'
+    ),
+    formatOutput({ stderr, stdout })
+  );
+});
```
pacocoursey, line 3090: Only added this test case, rest is lint.

packages/now-cli/src/util/projects/link.ts
```diff
@@ -141,13 +141,27 @@ export async function getLinkedProject(
...
     [org, project] = await Promise.all([
       getOrgById(client, link.orgId),
       getProjectByIdOrName(client, link.projectId, link.orgId),
     ]);
+  } catch (err) {
+    if (err?.status === 403) {
+      spinner();
+      throw new NowBuildError({
+        message: `Could not retrieve Project Settings. To link your project again, run ${chalk.gray(
+          `\`rm -rf .vercel\``
```
TooTallNate: This won't work on Windows. We should probably render a different command when running under Windows.

TooTallNate (changes requested):

packages/now-cli/src/util/projects/link.ts
pacocoursey: Because of the chalk styling, you mean? Should we just remove the colors instead?

pacocoursey: I updated the message to be more generic:

> Could not retrieve Project Settings. To link your project, remove the .vercel directory and deploy again

TooTallNate (dismissed):

packages/now-cli/src/util/projects/link.ts
```diff
@@ -141,13 +141,25 @@ export async function getLinkedProject(
...
       getProjectByIdOrName(client, link.projectId, link.orgId),
     ]);
+  } catch (err) {
+    if (err?.status === 403) {
+      spinner();
+      throw new NowBuildError({
+        message: `Could not retrieve Project Settings. To link your project, remove the .vercel directory and deploy again.`,
+        code: 'PROJECT_UNAUTHORIZED',
+        link: 'https://vercel.link/cannot-load-project-settings',
+      });
```
styfle, line 158: I think the API should be updated instead since we can return these same 3 properties.

Then you can use `output.prettyError(err)` instead.

packages/now-cli/src/util/errors-ts.ts
styfle: Indeed we can use NowBuildError() for CLI and Build errors. However, this error is coming from the API so we should actually have the API return those 3 properties instead and then use `output.prettyError(err)`.

packages/now-cli/src/util/projects/link.ts
pacocoursey: I don't think so. The API is returning the correct response (403), we're simply adding a special handling in this case.

It would be wrong to add this handling in the API instead, where the response is generic.

styfle: I suppose the message itself is only relevant for the CLI, so it seems good 👍

styfle (dismissed):

packages/now-cli/src/util/projects/link.ts
styfle: However, the link https://vercel.link/cannot-load-project-settings doesnt seem to have any section in the FAQ https://vercel.com/docs/v2/platform/frequently-asked-questions#cannot-load-project-settings

Did you forget to update the docs?

pacocoursey: Docs have been merged!

packages/now-cli/test/helpers/prepare.js
styfle: @pacocoursey Can you run prettier once more on the files you changed in this PR? We reverted this parens feature so it should show fewer line changes.

pacocoursey:
Sorry for re-request, had to update the integration test after rewording the error message.

packages/now-cli/test/helpers/prepare.js
pacocoursey: Missed this sorry. Done now!

TooTallNate (approved):

styfle (approved):
Thanks!

PR #4808, opened 2020-07-07
[cli] Remove caching of `vercel.json` config in `vc dev`

TooTallNate:
This is a re-introduction of #4697, which was reverted for performance
resons. But now that #4760 and #4793 have landed, we can enable this
behavior again.


packages/now-cli/src/util/dev/server.ts
```diff
@@ -507,23 +497,24 @@ export default class DevServer {
...
-  }
+    this.getNowConfigPromise = this._getNowConfig();
 
-  async _getNowConfig(canUseCache: boolean = true): Promise<NowConfig> {
-    if (canUseCache && this.cachedNowConfig) {
-      return this.cachedNowConfig;
-    }
+    // Clean up the promise once it has resolved
+    const clear = this.clearNowConfigPromise;
+    this.getNowConfigPromise.then(clear, clear);
```
styfle: TIL about the second parameter to `then()`. I have been using `then().catch()` all these years 🤯

styfle: Is that different than `.finally()`? Seems like it was using `try/finally` before this PR.

styfle (dismissed):

MaxLeiter (dismissed):

packages/now-cli/src/util/dev/server.ts
TooTallNate: I always forget about it because it's relatively new (Node 10+), but `engines.node` is already set to 10 for CLI, so we could use `.finally()` here.

```suggestion
    this.getNowConfigPromise.finally(clear);
```

styfle (approved):

AndyBitz (approved):

PR #4825, opened 2020-07-09
[cli] Handle API Error `action` property

lucleray:
- Use "Learn More" by default
- Replace with `error.action` if present

Looks like this:
image


nkzawa (dismissed):

TooTallNate (approved):

styfle (approved):

styfle (commented):
@lucleray Oh this is missing one: https://github.com/vercel/vercel/blob/ac1f4cf789f3c3f7bcf09aee62b8a89c9699ad43/packages/now-cli/src/util/output/create-output.ts#L70

PR #4827, opened 2020-07-09
[cli] Handle 400 errors as user errors

styfle:
When a user performs an action such as trying to add a domain they don't own, the API returns a 400-499 response code.

Previously, we would save this error in Sentry but this PR changes the behavior so that the error is printed locally for the user but not saved in Sentry.

### DIFF
--- a/packages/now-cli/src/index.js
+++ b/packages/now-cli/src/index.js
@@ -46,6 +46,7 @@ import reportError from './util/report-error';
 import getConfig from './util/get-config';
 import * as ERRORS from './util/errors-ts';
 import { NowError } from './util/now-error';
+import { APIError } from './util/errors-ts.ts';
 import { SENTRY_DSN } from './util/constants.ts';
 import getUpdateCommand from './util/get-update-command';
 import { metrics, shouldCollectMetrics } from './util/metrics.ts';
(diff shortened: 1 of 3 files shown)


packages/now-cli/src/index.js
```diff
@@ -639,6 +654,12 @@ const main = async argv_ => {
       return 1;
     }
 
+    if (err instanceof APIError && 400 <= err.status && err.status <= 499) {
+      err.message = err.serverMessage;
+      output.prettyError(err);
+      return 1;
+    }
```
styfle, line 660: This is the relevant code block for this PR, the other changes are prettier formatting only

packages/now-cli/src/util/response-error.ts
```diff
@@ -20,12 +19,9 @@ export default async function responseError(
...
-    message = bodyError.message;
   }
 
-  if (message == null) {
-    message = fallbackMessage === null ? 'Response Error' : fallbackMessage;
-  }
+  const msg = bodyError?.message || fallbackMessage || 'Response Error';
 
-  return new APIError(message, res, bodyError);
+  return new APIError(msg, res, bodyError);
```
styfle, line 26: I simplified the logic in this function but it should be equivalent

packages/now-cli/src/index.js
```diff
@@ -190,8 +191,10 @@ const main = async argv_ => {
   } catch (err) {
     console.error(
       error(
-        `${'An unexpected error occurred while trying to find the ' +
-          'global directory: '}${err.message}`
+        `${
+          'An unexpected error occurred while trying to find the ' +
+          'global directory: '
+        }${err.message}`
```
TooTallNate:
```suggestion
        `An unexpected error occurred while trying to find the global directory: ${
          err.message
        }`
```

pacocoursey (approved):
LGTM

TooTallNate (approved):

PR #4828, opened 2020-07-09
[cli] Replace "More details" with "Learn More"

lucleray:
Follow up https://github.com/vercel/vercel/pull/4825. Replace more "More details" with "Learn More".

### DIFF
--- a/packages/now-build-utils/src/detect-builders.ts
+++ b/packages/now-build-utils/src/detect-builders.ts
@@ -496,7 +496,7 @@ function getMissingBuildScriptError() {
     code: 'missing_build_script',
     message:
       'Your `package.json` file is missing a `build` property inside the `scripts` property.' +
-      '\nMore details: https://vercel.com/docs/v2/platform/frequently-asked-questions#missing-build-script',
+      '\nLearn More: https://vercel.com/docs/v2/platform/frequently-asked-questions#missing-build-script',
   };
 }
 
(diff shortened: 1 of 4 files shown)


AndyBitz (approved):

packages/now-build-utils/src/detect-builders.ts
```diff
@@ -496,7 +496,7 @@ function getMissingBuildScriptError() {
     code: 'missing_build_script',
     message:
       'Your `package.json` file is missing a `build` property inside the `scripts` property.' +
-      '\nMore details: https://vercel.com/docs/v2/platform/frequently-asked-questions#missing-build-script',
+      '\nLearn More: https://vercel.com/docs/v2/platform/frequently-asked-questions#missing-build-script',
```
styfle, line 499: Can you use `link` and `action` here?

lucleray: I'm not sure 🤔 I'm trying to improve it a little step at a time, so I haven't looked at whether we properly handle the `error` object in all the cases here.

styfle (approved):

packages/now-build-utils/src/detect-builders.ts
styfle: Ok we can revisit in another PR

TooTallNate (approved):

PR #4829, opened 2020-07-09
[frameworks][api] Add sort order

styfle:
This PR adds an optional property called `sort` to each framework so that we can change the order returned in the API. 
The reason this is necessary is because the order of the original array determines the precedence of framework detection. So we need another way to indicate the order of templates/examples returned from the API.

In particular, we need "Next.js" to be first and "Other" to be last.

I also updated the deprecated `@now/node` usage to `@vercel/node` in the API.

### DIFF
--- a/api/frameworks.ts
+++ b/api/frameworks.ts
@@ -1,21 +1,27 @@
-import { NowRequest, NowResponse } from '@now/node';
+import { NowRequest, NowResponse } from '@vercel/node';
 import { withApiHandler } from './_lib/util/with-api-handler';
-import frameworkList, { Framework } from '../packages/frameworks';
+import _frameworks, { Framework } from '../packages/frameworks';
 
-const frameworks = (frameworkList as Framework[]).map(frameworkItem => {
-  const framework = {
-    ...frameworkItem,
-    detectors: undefined,
-  };
+const frameworks = (_frameworks as Framework[])
+  .sort(
+    (a, b) =>
+      (a.sort || Number.MAX_SAFE_INTEGER) - (b.sort || Number.MAX_SAFE_INTEGER)
+  )
+  .map(frameworkItem => {
+    const framework = {
+      ...frameworkItem,
+      detectors: undefined,
+      sort: undefined,
+    };
 
-  if (framework.logo) {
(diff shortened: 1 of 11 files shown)


api/frameworks.ts
```diff
@@ -1,21 +1,27 @@
...
+  .sort(
+    (a, b) =>
+      (a.sort || Number.MAX_SAFE_INTEGER) - (b.sort || Number.MAX_SAFE_INTEGER)
+  )
+  .map(frameworkItem => {
+    const framework = {
+      ...frameworkItem,
+      detectors: undefined,
+      sort: undefined,
+    };
```
styfle, line 15: Relevant code is here - sort ascending then map to remove the new property

lucleray (approved):

AndyBitz (approved):

api/package.json
```diff
@@ -9,16 +9,16 @@
   },
   "dependencies": {
     "@sentry/node": "5.11.1",
+    "@vercel/node": "1.7.1",
```
TooTallNate, line 12: How come this is moved to a regular dependency?

packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -1,7 +1,8 @@
 import Ajv from 'ajv';
-import path from 'path';
+import { join } from 'path';
 import { existsSync } from 'fs';
 import { Framework } from '../';
+const frameworkList = require('../frameworks.json') as Framework[];
 
 function isString(arg: any): arg is string {
```
TooTallNate, line 7: Not really related to this PR, but could use `util.isString()` here instead.

TooTallNate (approved):

api/package.json
styfle: Oh that was a mistake, I'll create another PR since this was just merged

packages/frameworks/test/frameworks.unit.test.ts
styfle: https://github.com/vercel/vercel/pull/4836

api/package.json
styfle: https://github.com/vercel/vercel/pull/4836

PR #4830, opened 2020-07-10
[go] Add filesystem write fallback for `startDevServer()`

TooTallNate:
This is for `@vercel/go` to work on Windows, which currently fails with
this error:

```
panic: write pipe: The handle is invalid.

goroutine 1 [running]:
main.main()
	C:/Users/Nathan/Code/casca/vercel-go-test/.vercel/cache/go/2flefmhra8o/api/vercel-dev-server-main.go:25 +0x1ec
exit status 2
```

So this fallback writes the port number to a temp file that
`startDevServer()` polls for in order to figure out the port number.

### DIFF
--- a/packages/now-go/dev-server.go
+++ b/packages/now-go/dev-server.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -18,11 +19,17 @@ func main() {
 	}
 
 	port := listener.Addr().(*net.TCPAddr).Port
+	portBytes := []byte(strconv.Itoa(port))
 
 	file := os.NewFile(3, "pipe")
-	_, err2 := file.Write([]byte(strconv.Itoa(port)))
+	_, err2 := file.Write(portBytes)
 	if err2 != nil {
-		panic(err2)
+		portFile := os.Getenv("VERCEL_DEV_PORT_FILE")
+		os.Unsetenv("VERCEL_DEV_PORT_FILE")
+		err3 := ioutil.WriteFile(portFile, portBytes, 0644)
+		if err3 != nil {
+			panic(err3)
+		}
 	}
 
 	panic(http.Serve(listener, handler))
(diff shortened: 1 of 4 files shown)


packages/now-go/index.ts
```diff
@@ -562,19 +567,56 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go`
...
       pid: child.pid,
     };
-  } else {
+  } else if (Array.isArray(result)) {
     // Got "exit" event from child process
     throw new Error(
       `Failed to start dev server for "${entrypointWithExt}" (code=${result[0]}, signal=${result[1]})`
     );
+  } else {
+    throw new Error('Unexpected error');
```
styfle:
```suggestion
    throw new Error(`Unexpected result type: ${typeof result}`);
```

packages/now-go/index.ts
```diff
@@ -562,19 +567,56 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go`
...
+  }
+}
+
+export interface CancelablePromise<T> extends Promise<T> {
+  cancel: () => void;
+}
```
styfle, line 593: I wish Node.js had [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/AbortController) 😖

styfle (dismissed):

packages/now-go/index.ts
TooTallNate: Deno 😉

packages/now-go/dev-server.go
```diff
@@ -12,17 +13,23 @@ func main() {
 	handler := http.HandlerFunc(__HANDLER_FUNC_NAME)
 
 	// https://stackoverflow.com/a/43425461/376773
-	listener, err := net.Listen("tcp", ":0")
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
```
TooTallNate, line 16: Binding only to localhost prevents a firewall prompt on Windows

AndyBitz (approved):

styfle (approved):

cascandaliato:
Thank you for fixing this issue so quickly.

@paulogdm gave me the link to this PR and I am amazed by how transparent and streamlined the entire process is, from support ticket to code reviews, comments, and CI/CD.

I fell in love with the workflow shown in this PR page.

PR #4841, opened 2020-07-12
[next] Don't trace pages for lambdas when not needed

ijjk:
When pages are fully static and don't leverage preview mode or revalidation we can skip tracing the pages for generating the lambdas since lambdas aren't needed in this case


timneutkens (dismissed):

Timer (dismissed):
LGTM if tests are passing.

packages/now-next/test/fixtures/29-ssg-all-static/pages/forever.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+// eslint-disable-next-line camelcase
+export async function getStaticProps() {
+  return {
+    props: {
+      world: 'world',
+      time: new Date().getTime(),
+    },
+    unstable_revalidate: false,
```
Timer:
```suggestion
    revalidate: false,
```

Timer (approved):
Fixed code for canary upgrade

timneutkens (approved):

PR #4850, opened 2020-07-14
[cli] Set Vercel proxy request headers in `startDevServer()`

TooTallNate:
The Vercel proxy request headers (`x-vercel-deployment-url` for example)
were not being sent to the dev server when the Runtime defines
`startDevServer()`. This is now fixed.

Also updates the header names to match production:

 - `now` -> `vercel`
 - 'x-now-trace', 'x-now-id', and 'x-now-log-id' are removed

Fixes https://github.com/vercel/vercel/discussions/4729.

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -174,7 +174,7 @@ export default class DevServer {
     this.watchAggregationEvents = [];
     this.watchAggregationTimeout = 500;
 
-    this.filter = (path) => Boolean(path);
+    this.filter = path => Boolean(path);
     this.podId = Math.random().toString(32).slice(-5);
 
     this.devServerPids = new Set();
@@ -213,8 +213,8 @@ export default class DevServer {
       }
     }
 
-    events = events.filter((event) =>
-      distPaths.every((distPath) => !event.path.startsWith(distPath))
+    events = events.filter(event =>
+      distPaths.every(distPath => !event.path.startsWith(distPath))
     );
 
     // First, update the `files` mapping of source files
@@ -370,7 +370,7 @@ export default class DevServer {
       this,
       fileList
     );
-    const sources = matches.map((m) => m.src);
+    const sources = matches.map(m => m.src);
 
     if (isInitial && fileList.length === 0) {
       this.output.warn('There are no files inside your deployment.');
@@ -543,7 +543,7 @@ export default class DevServer {
       const { projectSettings, cleanUrls, trailingSlash } = config;
 
       const opts = { output: this.output, isBuilds: true };
-      const files = (await getFiles(this.cwd, config, opts)).map((f) =>
+      const files = (await getFiles(this.cwd, config, opts)).map(f =>
         relative(this.cwd, f)
       );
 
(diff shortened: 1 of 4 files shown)


packages/now-cli/src/util/dev/server.ts
```diff
@@ -1104,23 +1104,24 @@ export default class DevServer {
...
-      'x-now-trace': 'dev1',
-      'x-now-deployment-url': host,
-      'x-now-id': nowRequestId,
-      'x-now-log-id': nowRequestId.split('-')[2],
-      'x-zeit-co-forwarded-for': ip,
+      'x-vercel-deployment-url': host,
+      'x-vercel-forwarded-for': ip,
+      'x-vercel-id': nowRequestId,
     };
+    if (xfwd) {
```
styfle, line 1119: Why do you need to conditionally add these 3 headers?

TooTallNate: Cause of `node-http-proxy` setting these headers as well (in the case of `startDevServer()`). Some of the `x-forwarded-*` headers are a comma separated list, so setting them here first causes the values to be reported twice (i.e. `x-forwarded-for: 127.0.0.1,127.0.0.1`).

styfle: Can we confirm this works correctly with non-node runtimes? For example, add a check to the existing python dev test.

TooTallNate: Is the idea to test a runtime that _doesn't_ implement `startDevServer()`?

styfle: Indeed, to make sure both return the correct headers 👍

TooTallNate: Added in 4c672594783268477e270b69444da7cf0c2b857b, though `startDevServer()` may end up being implemented in `@vercel/python` at some point.

styfle (dismissed):

packages/now-cli/src/util/dev/server.ts
styfle: You could also add a test to the `validateResponseHeaders()` function which is used in multiple places

TooTallNate: That's different though since it's testing for the final _response_ headers, rather than the request headers that the API endpoint receives.

packages/now-cli/test/dev/integration.js
```diff
@@ -153,6 +154,9 @@ async function testPath(
   if (typeof expectedText === 'string') {
     const actualText = await res.text();
     t.is(actualText.trim(), expectedText.trim(), msg);
+  } else if (typeof expectedText === 'function') {
```
TooTallNate, line 157: Added functionality to pass a validation function to `testPath()` for more complicated assertions.

styfle (approved):

packages/now-cli/src/util/dev/server.ts
styfle: I'm not so sure? See [line 153](https://github.com/vercel/vercel/pull/4850/files#diff-948deac20e5be5b5293a06cead4b7c55R153) which uses the same `res` object you passed to `expectedText()` a few lines below that.

PR #4851, opened 2020-07-14
Update next-legacy-routes-optimized-lambdas.md

brandonpittman:
This document doesn't clearly state that using redirects in `next.config.js` needs to be under `experimental`. Linking to a conversation is confusing and misleading.

### DIFF
--- a/errors/next-legacy-routes-optimized-lambdas.md
+++ b/errors/next-legacy-routes-optimized-lambdas.md
@@ -7,7 +7,7 @@ When legacy `routes` are added in `now.json` or `vercel.json`, they cause confli
 
 #### Possible Ways to Fix It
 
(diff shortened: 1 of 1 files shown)


ijjk (changes requested):
I'm not sure we want to change this since we have moved the rewrites/redirects feature outside of `experimental` in the latest canary of Next.js. Pointing to the RFC allows users to keep track of the state of the feature until we can update to point to the feature's documentation instead

Timer:
Thanks for the PR, but I agree with JJ here. These feature were already de-flagged from `experimental` on canary and are going to be the default very soon.

brandonpittman:
It doesn’t make sense to call something out as an error/deprecation until the newer version is in master. I wasted more time that I should’ve had to trying to make this work. It’s a bad experience for users.

So leave this as is, that’s fine, but in the future, I’d like to see Next.js take these sorts of cases into consideration a little more carefully.

PR #4854, opened 2020-07-14
[next] Separate unit tests and integration tests

styfle:
This PR will make sure the unit tests run in multiple versions of node but the integration tests will still run once.


TooTallNate (dismissed):

TooTallNate:
Are there other runtimes that we can do this to?

styfle:
@TooTallNate I think this was the only package that has unit tests that were not separated.

See others here: https://github.com/vercel/vercel/search?q=test-unit&unscoped_q=test-unit

PR #4733, opened 2020-06-28
[cli] Add test for VERCEL_REGION

styfle:
Related to https://github.com/vercel/og-image/issues/123

PR #4855, opened 2020-07-14
[cli] Set Vercel platform env vars for `startDevServer()` invocations

TooTallNate:
Ensures that `VERCEL_REGION` is set for API calls when the runtime
implements `startDevServer()`.

Also adds `VERCEL_URL` which was previously not defined.

### DIFF
--- a/packages/now-cli/test/dev/integration.js
+++ b/packages/now-cli/test/dev/integration.js
@@ -137,6 +137,7 @@ async function getPackedBuilderPath(builderDirName) {
 
 async function testPath(
   t,
+  isDev,
   origin,
   status,
   path,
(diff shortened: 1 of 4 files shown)


packages/now-cli/test/dev/integration.js
```diff
@@ -1656,16 +1656,22 @@ test(
...
       t.is(headers['x-vercel-deployment-url'], host);
       t.truthy(isIP(headers['x-real-ip']));
       t.truthy(isIP(headers['x-forwarded-for']));
       t.truthy(isIP(headers['x-vercel-forwarded-for']));
+
+      // Test that the API endpoint has the Vercel platform env vars defined
+      t.is(env.NOW_URL, host);
+      t.is(env.NOW_REGION, 'dev1');
+      t.is(env.VERCEL_URL, host);
+      t.is(env.VERCEL_REGION, 'dev1');
```
styfle: Can you also test python? 😄

styfle (commented):
I added a test in #4733 that compares both dev and production. Can we also make sure this is testing both?

styfle (dismissed):

PR #4863, opened 2020-07-16
[next][node][static-build] Execute "vercel-build" script if defined

### LINKED ISSUE #4758: "vercel-build" in package.json isn't recognized
There is an option to add `now-build` in package.json and it gets prioritized over `build` script.
I guess it should now be updated to also recognize `vercel-build`.

TooTallNate:
Fixes #4758.


packages/now-next/src/index.ts
```diff
@@ -329,49 +318,34 @@ export const build = async ({
...
+        'If you need to define a different build step, please create a `vercel-build` script in your `package.json` ' +
+        '(e.g. `{ "scripts": { "vercel-build": "npm run prepare && next build" } }`).'
     );
-    pkg.scripts = {
-      'now-build': 'next build',
-      ...(pkg.scripts || {}),
-    };
-    shouldRunScript = 'now-build';
-    await writePackageJson(entryPath, pkg);
+    buildCommand = 'next build';
```
TooTallNate, line 352: Setting `buildCommand` instead of `buildScriptName` here avoids the package.json rewriting logic.

packages/now-node/src/index.ts
```diff
@@ -361,10 +361,7 @@ export async function build({
     meta,
   });
 
-  debug('Running user script...');
-  const runScriptTime = Date.now();
-  await runPackageJsonScript(entrypointFsDirname, 'now-build', spawnOpts);
```
TooTallNate, line 366: @leo @styfle Previously, only `now-build` was considered in `@vercel/node` (`build` is never executed). It's not clear to me if this is intentional or not, but this PR updates to invoke `vercel-build`/`now-build`/`build`. Please advise on if this is the correct behavior.

packages/now-next/src/index.ts
```diff
@@ -329,49 +318,34 @@ export const build = async ({
...
-    shouldRunScript = 'build';
-  } else if (!pkg.scripts || !pkg.scripts['now-build']) {
-    debug(
+  }
+
+  const buildScriptName = getScriptName('build', pkg);
+  let { buildCommand } = config;
+
+  if (!buildScriptName && !buildCommand) {
+    console.log(
```
TooTallNate, line 347: Changed to `console.log()` instead of `debug()` because this seems user-facing. Let me know if it should be changed back.

packages/now-next/src/index.ts
```diff
@@ -329,49 +318,34 @@ export const build = async ({
...
-    }
-
     console.warn(
       "WARNING: your application is being deployed in @vercel/next's legacy mode. http://err.sh/vercel/vercel/now-next-legacy-mode"
     );
 
+    await Promise.all([
+      remove(path.join(entryPath, 'yarn.lock')),
+      remove(path.join(entryPath, 'package-lock.json')),
+    ]);
```
Timer, line 331: We didn't use to error if these failed, plus the debug message is gone. Should we use `allSettled` instead?

Even if `remove` handles the file being missing, not sure if we want to error for other edge cases. I'll leave this to your discretion.

packages/now-next/src/index.ts
Timer: This is good!

packages/now-node/src/index.ts
styfle: I think we will want to avoid this. Many users have a `build` script now that we introduced zero config and that script is not intended for serverless functions, so we might slow down existing deployments by changing this

packages/now-build-utils/src/fs/run-user-scripts.ts
```diff
@@ -349,20 +353,50 @@ export async function runPipInstall(
...
+  return null;
+}
+
+export type RunPackageJsonScriptOptions = GetScriptNameOptions & SpawnOptions;
+
 export async function runPackageJsonScript(
   destPath: string,
-  scriptName: string,
-  spawnOpts?: SpawnOptions
+  baseScriptName: string,
```
styfle: I think this is going to cause problems for Community Builders which might not want to try 3 different script names.

I think we should leave `runPackageJsonScript` unchanged and instead call `getScriptName()` from `@vercel/node` and `@vercel/next` to have more fine-grained control. For example `@vercel/node` shouldn't execute `build`.

TooTallNate: Ya I made it be opt-in now by setting `allowPlatformName: true` in the options object.

packages/now-node/src/index.ts
TooTallNate: Ya, makes sense. I reverted back to the current behavior.

packages/now-build-utils/src/fs/run-user-scripts.ts
styfle: Oh that is an interesting way to do it? Seems a little confusing as an API though. Its not clear what `allowBaseName` or `allowPlatformName` options do without reading the source code.

When we introduced Node version selection, we added a new function `getSpawnOptions()` that Community Runtimes had to explicitly call.

I think we should do the same thing here. Export a new function that an array of options and return the first one: 

```js
// "@vercel/next"
const script = getScriptName(pkgJson, ['vercel-build', 'now-build', 'build']);
if (script) {
  await runPackageJsonScript(dest, script, opts);
}
```


```js
// "@vercel/node"
const script = getScriptName(pkgJson, ['vercel-build', 'now-build']);
if (script) {
  await runPackageJsonScript(dest, script, opts);
}
```

I think this would reduce the changes to `runPackageJsonScript()` and also make `getScriptName()` very simple.

styfle (dismissed):
Beautiful 🤩

packages/now-next/src/index.ts
styfle: The docs for [`remove()`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/remove.md) - "Removes a file or directory. The directory can have contents. If the path does not exist, *silently does nothing*."

@Timer So I think this is identical 👍 

@TooTallNate Maybe you might win me over into using `fs-extra` now 😉

TooTallNate: `remove()` is silent if the file doesn't exist. I think throwing for other edge cases is fine, as they essentially should not happen. This code is under the `isLegacy` code path, so I don't think it's executed very often at this point.

packages/now-static-build/src/index.ts
```diff
@@ -264,7 +264,6 @@ export async function build({
   const pkg = getPkg(entrypoint, workPath);
 
   const devScript = pkg ? getScriptName(pkg, 'dev', config) : null;
```
styfle, line 266: Do we still need this local `getScriptName` function now that there is the one in build-utils?

TooTallNate: I'd prefer to leave it for now since the logic in this function is slightly different (changes behavior based on if it's zero-config or not).

styfle: That's unfortunate. We still ended up with 3 different usages, one for each runtime 🤦‍♂️

styfle (dismissed):

Timer (dismissed):
LGTM from Next.js side, thanks!

iinfin:
@TooTallNate thanks a lot for your work but it doesn't seem to work for me 🤔
PR is merged but is it not updated internally yet, am I missing something?

I reproduced it in a public repository and made the logs visible:

200807_135126_07Xxy4kNFk

- [now-build (cf4a81e)](https://github.com/iinfin/pop/commit/cf4a81ef0824820295a7226df85a4890b42ce6db)
  - [`252sbq5w4 build logs`](https://vercel.com/iinfin/pop/252sbq5w4)
- [vercel-build (6f87f63)](https://github.com/iinfin/pop/commit/6f87f63b07dfbfe7d6f62401fe9c1765e5bbf039)
  - [`je8s8r3fr build logs`](https://vercel.com/iinfin/pop/je8s8r3fr)

styfle:
@iinfin This PR is currently on the canary channel and will be available in the next stable release.

iinfin:
Got it @styfle, I'll wait for that thank you

PR #4869, opened 2020-07-16
[build-utils] Add `getScriptName()` function

TooTallNate:
And allow a set of possible script names to use in `runPackageJsonScript()`.

These are only the changes to `@vercel/build-utils` from #4863, since
we'll need to cut a canary with these changes in order for the CLI "dev"
tests to pass.


styfle (approved):
🥇

styfle:
Did you put "dev" in quotes because its really failing during the deployment and not the local dev test?

TooTallNate:
Good thought, but no, I was just referencing the GitHub action workflow name 😛 

The test case that I'm referring to ("output-is-source") doesn't actually use the `testFixtureStdio()` function (#4870 should also solve this issue, and somewhat makes this PR unnecessary, but still doesn't hurt to do separate commits).

styfle:
Do we still need a canary release now that this is merged?

PR #4870, opened 2020-07-16
[cli] Remove "npm pack" logic from `output-is-source` dev test

TooTallNate:
Now that `@vercel/static-build` is being symlinked to CLI due to the
yarn workspace, `vc dev` will always use the latest code for
zero-config, so this "npm pack" logic is no longer necessary.

The `vercel.json` also no longer gets created in this test so that
zero-config is used.


styfle (approved):
Thanks!

PR #4887, opened 2020-07-19
fix: avoid block user use global typescript

bluelovers:


styfle (commented):
What is this fixing? Can you add a test that fails without this PR?

TooTallNate:
What is this fixing exactly?

bluelovers:
```json
{
  "compilerOptions": {
    "target": "ES2019",
    "alwaysStrict": true
  }
}
```

```ts
let b;

export let a = b ??= true;
```

global install typescript@next
local install vercel

```json
  "scripts": {
    "test": "ts-node index.ts",
```

image

TooTallNate:
It's not going to work with a globally installed `typescript`. You need to install it as a local dependency instead (`yarn add --dev typescript`).

If you can create a repository that demonstrates your issue then we can take a look and perhaps send a PR to show you the fix.

PR #4894, opened 2020-07-20
[next] Add tests for trailing slash behavior

Timer:
This adds tests for Next.js' trailing slash support.


packages/now-next/test/fixtures/00-trailing-slash-add/now.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": 2,
+  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "probes": [
+    { "path": "/foo/", "status": 200, "mustContain": "foo page" },
+    {
+      "fetchOptions": { "redirect": "manual" },
+      "path": "/foo",
+      "status": 308,
+      "responseHeaders": {
+        "refresh": "/url=/foo/$/"
```
styfle, line 11: Why check `refresh` instead of `location`?

The `$` seems unexpected 🤔

packages/now-next/test/fixtures/00-trailing-slash-remove/next.config.js
```diff
@@ -0,0 +1 @@
+module.exports = { trailingSlash: false };
```
styfle, line 1: This test starts with `00-`. Did you mean for both to start with the same number?

styfle (commented):
Its missing tests for `/script.js` and `/script.js/` which behave differently because of the file extension

packages/now-next/test/fixtures/00-trailing-slash-add/now.json
Timer: We send both the location and refresh header, however, in tests, the `Location` header is absolute while the `Refresh` header is relative. 
Thus, it's easier to test the `Refresh` header as it has a `url=` anchor at the beginning. Otherwise, I'd need to know the deployment URL.

The `$` is expected as this is a regex.

packages/now-next/test/fixtures/00-trailing-slash-remove/next.config.js
Timer: Yes, it was intentional that these start with `00` so they're sorted to the top and run first before other tests (same thing we do with `00-shared-lambdas`).

packages/now-next/test/fixtures/00-trailing-slash-add/now.json
Timer:
```
      cache-control=s-maxage=0
      connection=close
      content-type=text/plain
      date=Mon, 20 Jul 2020 20:03:01 GMT
      location=https://test2020-4ookam6q1.vercel.app/foo
      refresh=0;url=/foo
      server=Vercel
      strict-transport-security=max-age=63072000; includeSubDomains; preload
      transfer-encoding=chunked
      x-robots-tag=noindex
      x-vercel-id=iad1::8tkzw-1595275381290-9df066c957d2
```

styfle: Strange, that must be a node-fetch feature then.

```
$ curl -I https://test2020-4ookam6q1.vercel.app/foo/
HTTP/2 308
date: Mon, 20 Jul 2020 22:26:43 GMT
content-type: text/plain
x-robots-tag: noindex
location: /foo
refresh: 0;url=/foo
server: Vercel
x-vercel-id: cle1::d5629-1595284003612-0f0c6325fef8
strict-transport-security: max-age=63072000; includeSubDomains; preload
cache-control: s-maxage=0
```

styfle (dismissed):

packages/now-next/test/fixtures/00-trailing-slash-add/now.json
Timer: An annoying feature at that!

styfle (approved):

Timer:
Landing this as Next.js E2E tests fail otherwise.

PR #4896, opened 2020-07-20
[cli] Use `update-notifier` instead of `update-check`

TooTallNate:
This makes the updating logic be asynchronous instead of synchronous, and as such will make boot-up of CLI be faster.

The actual update notification display is identical to previous, we are not using `update-notifier`'s default boxen rendering.

### DIFF
--- a/packages/now-cli/package.json
+++ b/packages/now-cli/package.json
@@ -68,7 +68,8 @@
     "@vercel/node": "1.7.3-canary.0",
     "@vercel/python": "1.2.2",
     "@vercel/ruby": "1.2.3",
-    "@vercel/static-build": "0.17.6-canary.1"
+    "@vercel/static-build": "0.17.6-canary.1",
+    "update-notifier": "4.1.0"
   },
   "devDependencies": {
     "@sentry/node": "5.5.0",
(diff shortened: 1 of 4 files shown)


packages/now-cli/package.json
```diff
@@ -68,7 +68,8 @@
     "@vercel/node": "1.7.3-canary.0",
     "@vercel/python": "1.2.2",
     "@vercel/ruby": "1.2.3",
-    "@vercel/static-build": "0.17.6-canary.1"
+    "@vercel/static-build": "0.17.6-canary.1",
+    "update-notifier": "4.1.0"
```
TooTallNate, line 72: `update-notifier` [is not compatible with `ncc`](https://github.com/vercel/ncc/issues/495), so it's a regular dependency rather than a devDependency.

(Also note that the `update-notifier-webpack` fork that is listed in that GH issue [is still not compatible with ncc](https://github.com/vercel/ncc/issues/495#issuecomment-661311495))

AndyBitz (approved):

styfle (approved):

packages/now-cli/package.json
styfle: We might see `npm i -g vercel` time go up since [`update-notifier`](https://packagephobia.com/result?p=update-notifier) adds 1 MB and 600+ files.

PR #4897, opened 2020-07-21
[cli] Add `vercel link` command

### LINKED ISSUE #3767: [now dev] Error! Your codebase isn’t linked to a project on ZEIT Now.
I just upgraded `now` and now I get the following on `now dev`:

```
Error! Your codebase isn’t linked to a project on ZEIT Now. Run `now` to link it.
```

Did I understand correctly that I have to deploy a project before I can develop it?

styfle:
- [x] Add `vercel link` command
- [x] Update `vercel dev` to automatically link 
- [x] Updated `vercel env` error message to mention `vercel link`
- [x] Test `vercel link`
- [x] Test `vercel dev`
- [x] Add `--confirm` flag

### Related

- Fixes #3767
- Fixes #4889 
- Fixes #3991 
- Fixes #4534 
- Related to #3989


packages/now-cli/src/util/projects/create-project.ts
```diff
@@ -0,0 +1,13 @@
+import Client from '../client';
+import { Project } from '../../types';
+
+export default async function createProject(
+  client: Client,
+  projectName: string
+) {
+  const project = await client.fetch<Project>('/v1/projects', {
```
AndyBitz, line 8: Is `v1` the latest version of this endpoint?

packages/now-cli/src/util/input/edit-project-settings.ts
```diff
@@ -4,32 +4,30 @@ import chalk from 'chalk';
 import { Output } from '../output';
 import { Framework } from '@vercel/frameworks';
 import { isSettingValue } from '../is-setting-value';
+import { ProjectSettings } from '../../types';
 
-export interface ProjectSettings {
+export interface PartialSettings {
```
AndyBitz: Could we rename this to `PartialProjectSettings`, I think just `PartialSettings` sounds a bit too generic.

packages/now-cli/src/util/link/setup-and-link.ts
```diff
@@ -0,0 +1,225 @@
...
+    return link;
+  }
+
+  if (force) {
+    const vercelDir = getVercelDirectory(path);
+    remove(vercelDir);
+  }
+
+  const shouldStartSetup =
+    autoConfirm ||
```
AndyBitz, line 74: `autoConfirm` is always `false`. Is that expected?

packages/now-cli/src/commands/dev/dev.ts
```diff
@@ -37,20 +37,21 @@ export default async function dev(
...
     getFrameworks(client),
   ]);
 
-  if (link.status === 'error') {
-    return link.exitCode;
+  if (link.status === 'not_linked' && !process.env.__VERCEL_SKIP_DEV_CMD) {
+    link = await setupAndLink(ctx, output, cwd, false, 'link');
+
+    if (link.status === 'not_linked') {
+      return 0;
```
AndyBitz, line 61: Can this happen? Should we show an error message?

packages/now-cli/src/util/projects/create-project.ts
styfle: I was following the documentation here https://vercel.com/docs/api#endpoints/projects/create-a-project

packages/now-cli/src/util/link/setup-and-link.ts
styfle: I think this is used with `--confirm` so maybe we should add `vc link --confirm` and `vc dev --confirm`? What do you think?

packages/now-cli/src/commands/dev/dev.ts
styfle: Yes, this happens when the user says "N" and it prints "Aborted"

packages/now-cli/src/util/input/edit-project-settings.ts
```diff
@@ -4,32 +4,30 @@ import chalk from 'chalk';
...
-) {
+): Promise<ProjectSettings> {
   // create new settings object, missing values will be filled with `null`
-  const settings: Partial<ProjectSettingsWithFramework> = {
-    ...projectSettings,
-  };
+  const settings: ProjectSettings = Object.assign(
+    { framework: null },
+    projectSettings
+  );
```
styfle, line 31: @AndyBitz I noticed this function was returning `Partial<ProjectSettingsWithFramework>` when it seems to populate all properties. What do you think about my solution here by assigning `framework` right at the beginning so we can return the complete `ProjectSettings` each time?

packages/now-cli/src/util/link/setup-and-link.ts
AndyBitz: That'd would make sense to me 👍

packages/now-cli/src/commands/dev/dev.ts
AndyBitz: Got it 👍 
Then maybe add a comment that mentions this, cause it seems like it's just exiting

packages/now-cli/src/util/input/edit-project-settings.ts
AndyBitz: I think this should be fine, since we set it to `null` anyways when it's falsy

packages/now-cli/src/commands/dev/dev.ts
styfle: I'm not sure if we want to return a 0 exit code but I couldn't find a convention, maybe @TooTallNate knows?

https://stackoverflow.com/q/53304662/266535

I also added comments.

packages/now-cli/src/util/link/setup-and-link.ts
styfle: I added `vc link --confirm` and `vc dev --confirm`.

packages/now-cli/src/util/input/edit-project-settings.ts
styfle: Renamed 👍

AndyBitz (dismissed):
Nice work :shipit:

packages/now-cli/src/commands/dev/dev.ts
TooTallNate: I'm not sure of any particular convention, but at least we should be consistent with other prompts in this CLI. My gut feeling is that `0` is correct since it's akin to pressing Ctrl+C.

MaxLeiter (dismissed):

TooTallNate (approved):

AndyBitz (approved):

simonhaenisch:
I quickly skimmed through the changes and it doesn't look like it's possible yet... would you consider adding the ability to specify an existing team scope and project with the command, e. g.

```
vercel link team/project
```

or is that not very straight-forward?

styfle:
@simonhaenisch No that was not implemented. That's a good suggestion, thanks!

PR #4908, opened 2020-07-22
[cli] print build error message instead of 'Build failed'

igorklopov:
The build pipeline exposes errors via https://github.com/vercel/vercel/blob/5b895744b134d7ee9ed042889fcbb4c5ce79b324/packages/now-cli/src/util/index.js#L336 so that we can show them to user. Also this change will make `lambda-with-invalid-runtime` test pass when new pipeline is in prod, by making this assert true https://github.com/vercel/vercel/blob/84158cfa7200ac85bde6c610c626d06e69c33f14/packages/now-cli/test/integration.js#L2510


lucleray (dismissed):
Code looks good 👍

TooTallNate (dismissed):

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -669,7 +669,7 @@ export default async function main(
     }
 
     if (err instanceof BuildError) {
-      output.error('Build failed');
+      error(err.message || 'Build failed');
```
styfle: Why `error()` instead of `output.error()`?

igorklopov: `error` is destructured here https://github.com/vercel/vercel/blob/d72d969971941d5448615668f9e6f4bc8fa5b645/packages/now-cli/src/commands/deploy/latest.js#L224
Anyways fixed 👌

lucleray (approved):

styfle (approved):
Thanks!

PR #4911, opened 2020-07-23
[cli] Fix devServer for routes rewritten to query params

jeantil:
Routes which use the `rewrites` field in now.json were incorrectly routed by `vercel dev` as demonstrated in the unit test added in this PR. 

given a path param rewrite : 
```
{
  "rewrites":[
   {"source": "/:path, "dest":"/foo"}
  ]
}
```
the rewrite is converted by vercel to the following route: 
```
{src: '^(?:/([^/]+?))$', dest: '/foo?path=$1', check: true}
```
However when running in dev, the server will attempt to resolve that with `devServer.hasFileSystem` which may fail if the target is a nextjs route. When failing, the des including the query param was assigned to the `reqPathname` instead of separating `path ` and `query`. The rest of the routing would end up encoding the query params to make the string a valid path and then it would attempt to find a route for that incorrectly encoded string.

This pr addresses this issue

### DIFF
--- a/packages/now-cli/src/util/dev/router.ts
+++ b/packages/now-cli/src/util/dev/router.ts
@@ -150,7 +150,9 @@ export async function devRouter(
               if (routeConfig.status && phase === 'miss') {
                 status = routeConfig.status;
               }
-              reqPathname = destPath;
+              const destParsed = url.parse(destPath, true);
+              query = Object.assign(destParsed.query, query);
+              reqPathname = destParsed.pathname || '/';
               continue;
             }
           }
--- a/packages/now-cli/test/dev-server.unit.js
+++ b/packages/now-cli/test/dev-server.unit.js
@@ -175,6 +175,18 @@ test(
     t.is(parsed.query['route-param'], 'b');
   })
 );
+test(
+  '[DevServer] Maintains query when builder defines routes through rewrites',
+  testFixture('vc-dev-next-rewrites', async (t, server) => {
+    const res = await fetch(`${server.address}/something?url-param=a`);
+    validateResponseHeaders(t, res);
+
+    const parsed = await res.json();
+
+    t.is(parsed.query['url-param'], 'a');
+    t.is(parsed.query['routeparam'], 'something');
+  })
+);
 
 test(
   '[DevServer] Allow `cache-control` to be overwritten',
--- a/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/.gitignore
+++ b/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/.gitignore
@@ -1,0 +1,4 @@
+.next
+yarn.lock
+
+.vercel
--- a/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/next.config.js
+++ b/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/next.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  target: 'serverless',
+};
--- a/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/package.json
+++ b/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/package.json
@@ -1,0 +1,11 @@
+{
+  "scripts": {
+    "build": "next build",
+    "dev": "next"
+  },
+  "dependencies": {
+    "next": "^9.4.0",
+    "react": "^16.7.0",
+    "react-dom": "^16.7.0"
+  }
+}
--- a/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/pages/api/index.js
+++ b/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/pages/api/index.js
@@ -1,0 +1,5 @@
+export default async function preview(req, res) {
+  res.statusCode = 200;
+  res.setHeader('Content-Type', 'application/json');
+  res.end(JSON.stringify({ query: req.query }));
+}
--- a/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/pages/index.js
+++ b/packages/now-cli/test/fixtures/unit/vc-dev-next-rewrites/pages/index.js
@@ -1,0 +1,11 @@
+import { withRouter } from 'next/router';
+
+function Index({ router }) {
+  const data = {
+    pathname: router.pathname,
+    query: router.query,
+  };
+  return <div>{JSON.stringify(data)}</div>;
+}
+
+export default withRouter(Index);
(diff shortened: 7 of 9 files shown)


packages/now-cli/test/dev-server.unit.js
```diff
@@ -175,6 +175,26 @@ test(
     t.is(parsed.query['route-param'], 'b');
   })
 );
+test(
+  '[DevServer] Maintains query when builder defines routes through rewrites',
+  testFixture('now-dev-next-rewrites', async (t, server) => {
```
styfle: `now-dev` should be `vc-dev` for this test fixture

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/now.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 2,
+  "name": "now-dev-next-rewrites",
```
styfle: Remove `name` since its deprecated

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/now.json
```diff
@@ -0,0 +1,11 @@
...
+  "builds": [{ "src": "package.json", "use": "@now/next@canary" }],
```
styfle:
```suggestion
  "builds": [{ "src": "package.json", "use": "@vercel/next@canary" }],
```

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "scripts": {
+    "build": "next build",
+    "dev": "next",
+    "now-build": "next build"
```
styfle: Remove `now-build`

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/package.json
```diff
@@ -0,0 +1,12 @@
...
+  },
+  "dependencies": {
+    "next": "^8.0.0",
```
styfle: Is this bug specific to Next 8? The last release of Next 8.x was over a year ago so we should probably use latest.

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/now.json
```diff
@@ -0,0 +1,11 @@
...
+  "rewrites": [
+    {
+      "source": "/:routeparam",
+      "destination": "/"
+    }
+  ]
+}
```
styfle: This file should be renamed from `now.json` to `vercel.json`

packages/now-cli/test/dev-server.unit.js
```diff
@@ -175,6 +175,26 @@ test(
...
+    const res = await fetch(`${server.address}/something?url-param=a`);
+    validateResponseHeaders(t, res);
+
+    const text = await res.text();
+
+    // Hacky way of getting the page payload from the response
+    // HTML since we don't have a HTML parser handy.
```
styfle: Why not use `pages/api` instead to return a proper json response? Or is this bug only affecting non-api endpoints?

packages/now-cli/src/util/dev/router.ts
```diff
@@ -150,7 +150,9 @@ export async function devRouter(
               if (routeConfig.status && phase === 'miss') {
                 status = routeConfig.status;
               }
-              reqPathname = destPath;
+              const destParsed = url.parse(destPath, true);
+              query = Object.assign(destParsed.query, query);
+              reqPathname = destParsed.pathname || '/';
```
styfle, line 155: @TooTallNate @ijjk @Timer I'm not sure if we should be hitting this code at all considering `@vercel/next` has its own dev server. What do you think?

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/package.json
jeantil: I copy pasted the `now-dev-next` test and only made very minor modifications to it to verify my specific condition. since `rewrites` and `routes` cannot be in the same file I couldn't evolve the existing test but I'll update my test to match the `now->vercel` rename.

packages/now-cli/test/fixtures/unit/now-dev-next-rewrites/now.json
jeantil: you may want to have a look at the other fixtures, quite a few of them are still referencing `@now` builders

packages/now-cli/test/dev-server.unit.js
jeantil: I hesitated but since there is not a single fixture named `vc-dev` in the unit tests I mistakenly assumed that consistency was better.

packages/now-cli/test/dev-server.unit.js
jeantil: Sure, as far as I'm aware the test affects any rewritten  route that targets next, and the failure occurs way before reaching next :) as for the rest of the comments you made on the test, it was cloned from another test which did that and I didn't need to change that specific part :)

packages/now-cli/src/util/dev/router.ts
jeantil: from my observations, in this configuration `serveProjectAsNowV2` goes through the `devRouter` multiple times with varying arguments. For each phase it can go through: 
- [`initial route resolution`](https://github.com/vercel/vercel/blob/3d4261300c3311b7e8faa7bd01ba0f72bccb1fb6/packages/now-cli/src/util/dev/server.ts#L1373) (which is pre-match and thus doesn't know the url will be handled by next)
- [`findBuildMatch`](https://github.com/vercel/vercel/blob/3d4261300c3311b7e8faa7bd01ba0f72bccb1fb6/packages/now-cli/src/util/dev/server.ts#L2108) which decides if it [`shouldServe`](https://github.com/vercel/vercel/blob/3d4261300c3311b7e8faa7bd01ba0f72bccb1fb6/packages/now-cli/src/util/dev/server.ts#L2146) 
- resolve effective proxied url 

since the `pathName` is corrupted in the initial route resolution `shouldServe` never gets the right information...and ends up resolving an incorrect path, it seems to understand it must proxy it to `next` but in the process what should be a query string ends up encoded as a escaped path segment  and the request eventually fails.

styfle: It seems like this case is already handled on line 181 below:

```js
const destParsed = url.parse(destPath, true);
Object.assign(destParsed.query, query);
```

Why is it needed here too? Is line 181 never executed?

styfle:
@jeantil Have you considered using Next.js 9.5 which supports `rewrites` in `next.config.js`?

https://nextjs.org/blog/next-9-5#support-for-rewrites-redirects-and-headers

PR #4913, opened 2020-07-23
[api] Fix api-framework cache header

juancampa:
## Problem

`api-frameworks` is not being cached by our edge when used via the fallback-to-front mechanism. That's because the headers are only applied to `/api/frameworks` but not `/api/v1/frameworks` which is the actual path being used.

## Solution

Also apply the caching headers to `/api/<version>/frameworks`


TooTallNate (approved):

styfle (approved):
Thanks!

Vichet97:
Does the stale-if-error work ?

PR #4915, opened 2020-07-23
[frameworks] Add autodetection for RedwoodJS

styfle:
- Supersedes #4539
- Related to #4676


packages/frameworks/frameworks.json
```diff
@@ -682,6 +682,33 @@
...
+          "matchContent": "\"(dev)?(d|D)ependencies\":\\s*{[^}]*\"@redwoodjs\\/core\":\\s*\".+?\"[^}]*}"
+        }
+      ]
+    },
+    "settings": {
+      "buildCommand": {
+        "value": "yarn rw db up --no-db-client --auto-approve && yarn rw build"
+      },
+      "devCommand": {
+        "value": "yarn rw dev"
```
AndyBitz, line 705: Don't we need to add `--port $PORT` or something to pass the port?

styfle: They don't support that flag unfortunately, they use a [redwood.toml](https://redwoodjs.com/docs/app-configuration-redwood-toml) file to set the port.

TooTallNate (approved):

packages/frameworks/frameworks.json
TooTallNate: My understanding is that this dev command won't be used, since the `@vercel/redwood` Runtime is special-cased. Is that correct?

styfle: It uses `config.devCommand` here:
https://github.com/vercel/vercel/blob/041e8cc601befc5e9e745daf7c098fed4914f889/packages/redwood/src/index.ts#L92

So it could be overridden if the user changes their project settings.

AndyBitz (approved):

PR #4916, opened 2020-07-23
[build-utils] Fix redwood `/api` detection

styfle:
Previously, the zero detection would see js files in the `/api` directory and attempt to use `@vercel/node`. However, this doesn't work with RedwoodJS which uses a different format for [`/api`](https://github.com/vercel/vercel/tree/master/examples/redwoodjs/api) so we must only use `@vercel/redwood` when RedwoodJS is detected.

### DIFF
--- a/packages/now-build-utils/src/detect-builders.ts
+++ b/packages/now-build-utils/src/detect-builders.ts
@@ -309,6 +309,13 @@ export async function detectBuilders(
     options
   );
 
+  if (frontendBuilder && framework === 'redwoodjs') {
+    // RedwoodJS uses the /api directory differently so we must
+    // clear any existing builders and only use `@vercel/redwood`.
+    builders.length = 0;
+    builders.push(frontendBuilder);
+  }
+
   return {
     warnings,
     builders: builders.length ? builders : null,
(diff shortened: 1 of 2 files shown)


packages/now-build-utils/src/detect-builders.ts
```diff
@@ -309,6 +309,13 @@ export async function detectBuilders(
     options
   );
 
+  if (frontendBuilder && framework === 'redwoodjs') {
+    // RedwoodJS uses the /api directory differently so we must
+    // clear any existing builders and only use `@vercel/redwood`.
+    builders.length = 0;
```
AndyBitz, line 315: TIL

packages/now-build-utils/src/detect-builders.ts
```diff
@@ -309,6 +309,13 @@ export async function detectBuilders(
...
+    builders.push(frontendBuilder);
```
TooTallNate, line 316: Do we need to print a warning here?

AndyBitz (approved):

TooTallNate (approved):
Code looks good.

packages/now-build-utils/src/detect-builders.ts
styfle: This is not really a warning, this is the expected behavior. I updated the PR description with more info.

packages/now-build-utils/src/detect-builders.ts
styfle: Its a bit of a hack but it avoids changing `const builders` to `let builders` 😅

PR #4917, opened 2020-07-24
[client] Export `buildFileTree()` and return the ignore list

TooTallNate:
The ignore list from this function will be used in production to
render a warning if there are files listed in the `.vercelignore` file
that are being committed to the project's git repo.


packages/now-client/src/utils/index.ts
```diff
@@ -81,17 +81,24 @@ export async function buildFileTree(
...
   debug: Debug
-): Promise<string[]> {
+): Promise<{ fileList: string[]; ignoreList: string[] }> {
+  const ignoreList: string[] = [];
   let fileList: string[];
-  let { ig } = await getVercelIgnore(path);
+  let { ig, ignores } = await getVercelIgnore(path);
 
-  debug(`Found ${ig.ignores.length} rules in .vercelignore`);
+  debug(`Found ${ignores.length} rules in .vercelignore`);
```
TooTallNate, line 89: This was previously always printing `1` because `ig.ignores` is a function that takes one paramater, so it coincidentally had a `length` property of `1`. Fixed to render the number of rules by checking the proper array.

packages/now-client/src/utils/index.ts
```diff
@@ -81,17 +81,24 @@ export async function buildFileTree(
...
   debug('Building file tree...');
 
   if (isDirectory && !Array.isArray(path)) {
     // Directory path
-    const cwd = process.cwd();
-    const ignores = (absPath: string) => ig.ignores(relative(cwd, absPath));
+    const ignores = (absPath: string) => {
+      const rel = relative(path, absPath);
```
styfle, line 95: Was this a bug before? It was previously relative to the `cwd`?

packages/now-client/src/utils/index.ts
```diff
@@ -81,17 +81,24 @@ export async function buildFileTree(
...
+      const ignored = ig.ignores(rel);
+      if (ignored) {
+        ignoreList.push(rel);
+      }
+      return ignored;
+    };
     fileList = await readdir(path, [ignores]);
     debug(`Read ${fileList.length} files in the specified directory`);
```
styfle:
```suggestion
    debug(`Found ${fileList.length} files in the specified directory`);
```

Just a nit pick here 😄

packages/now-client/src/utils/index.ts
```diff
@@ -81,17 +81,24 @@ export async function buildFileTree(
   path: string | string[],
   isDirectory: boolean,
   debug: Debug
```
styfle, line 83: Are we going to be able to use this `debug()` command during the build?

styfle (commented):
Let's add some tests (or update existing tests) to check for the new `ignoreList` export

packages/now-client/src/utils/index.ts
TooTallNate: Ya, seemed like a bug to me.

packages/now-client/src/utils/index.ts
TooTallNate: I was able to just pass `console.log` to it in a local script and typescript didn't complain, so I think we can pass it basically any function that accepts a string argument.

styfle (approved):
:shipit:

PR #4923, opened 2020-07-25
[build-utils] Allow `includeFiles` to be an array in ajv schema

TooTallNate:
`includeFiles` may be an array according to [the implementation](https://git.io/JJ49k).


styfle:
I believe @AndyBitz and @leo decided to only allow a string once `functions.includeFiles` was introduced since a glob can use multiple groups with curl brackets, no need for an array.

TooTallNate:
Meh. More inconsistencies between `builds` and `functions` 😕

PR #4935, opened 2020-07-27
[next] Update test fixtures for latest changes

ijjk:
This updates our Next.js test fixtures to run against canary. 

It looks like the mono-repo fixtures will need to be updated to run against canary separately since there may be a conflict with legacy routes and the trailing slash redirect


styfle (approved):
Thanks!

Timer (approved):

PR #4937, opened 2020-07-27
[cli][redwood] Update redwood port detection in `vc dev`

styfle:
Fixes redwood support for `vc dev`

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -1578,6 +1578,13 @@ export default class DevServer {
       debug(
         `Checking build result's ${buildResult.routes.length} \`routes\` to match ${newUrl}`
       );
+      for (const r of buildResult.routes) {
+        // This replace is necessary for `@vercel/redwood` but might be relevant
+        // for builders that wish to return routes and work with zero config.
+        if (r.dest) {
+          r.dest = r.dest.replace(/\$PORT/g, `${this.devProcessPort}`);
+        }
+      }
       const matchedRoute = await devRouter(
         newUrl,
         req.method,
--- a/packages/now-cli/test/dev/integration.js
+++ b/packages/now-cli/test/dev/integration.js
@@ -20,6 +20,8 @@ let port = 3000;
 const binaryPath = resolve(__dirname, `../../scripts/start.js`);
 const fixture = name => join('test', 'dev', 'fixtures', name);
 const fixtureAbsolute = name => join(__dirname, 'fixtures', name);
+const exampleAbsolute = name =>
+  join(__dirname, '..', '..', '..', '..', 'examples', name);
 
 let processCounter = 0;
 const processList = new Map();
@@ -127,9 +129,10 @@ async function testPath(
   path,
   expectedText,
   headers = {},
-  method = 'GET'
+  method = 'GET',
+  body = undefined
 ) {
-  const opts = { redirect: 'manual-dont-change', method };
+  const opts = { redirect: 'manual-dont-change', method, body };
   const url = `${origin}${path}`;
   const res = await fetch(url, opts);
   const msg = `Testing response from ${method} ${url}`;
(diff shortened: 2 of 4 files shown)


packages/redwood/src/index.ts
```diff
@@ -89,46 +70,10 @@ export async function build({
 
   const {
     buildCommand = 'yarn rw db up --no-db-client --auto-approve && yarn rw build',
-    devCommand = 'yarn rw dev',
```
styfle, line 92: I found that `vc dev` already runs the `devCommand` so it was being run twice.

I was able to remove the bulk of the dev logic here besides returning routes.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -1952,7 +1954,11 @@ export default class DevServer {
       `Running Dev Command ${chalk.cyan.bold(`“${devCommand}”`)}`
     );
 
-    const port = await getPort();
+    let port = await getPort();
+
+    if (this.frameworkSlug === 'redwoodjs') {
+      port = await getRedwoodPort(cwd);
+    }
```
styfle: @TooTallNate I'm open to ideas if you can think of a better way to handle this besides special case.

I created a PR upstream but it doesn't seem like we'll get it soon https://github.com/redwoodjs/redwood/pull/891

packages/now-cli/test/dev/integration.js
```diff
@@ -1560,6 +1565,21 @@ test(
...
+    async testPath => {
+      await testPath(200, '/', /<div id="redwood-app">/m);
+      await testPath(200, '/about', /<div id="redwood-app">/m);
+      const reqBody = '{"query":"{redwood{version}}"}';
+      const resBody = '{"data":{"redwood":{"version":"0.14.0"}}}';
+      await testPath(200, '/api/graphql', resBody, {}, 'POST', reqBody);
+    },
+    { isExample: true }
+  )
+);
```
styfle: I want to start testing examples because currently these are untested in PRs 😬 

This `isExample` option was a quick way to get us there.

styfle: These tests were copied from `probes` here: https://github.com/vercel/vercel/blob/74bb8a21ed710ccdc1546d474c2952181b58647f/packages/redwood/test/fixtures/01-create-redwood-app/vercel.json#L4-L19

.github/workflows/test-unit.yml
```diff
@@ -22,9 +22,11 @@ jobs:
       - uses: actions/checkout@v2
       - run: git fetch origin master --depth=10
       - run: git fetch origin ${{ github.ref }} --depth=10
-      - uses: actions/setup-node@v1
       - run: yarn install
       - run: yarn run build
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
```
styfle: For some reason, unit tests were failing during `ncc` so I used the same pattern as `test-integration-cli.yml` and `test-integration-dev.yml` and that seemed to fix it.

Its still a mystery why it ever worked before.

packages/now-cli/src/util/dev/server.ts
TooTallNate: I think it's fine considering that it's just temporary until upstream provides the proper functionality.

.github/workflows/publish.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
styfle: Should lead to faster checkouts https://github.com/actions/checkout/releases/tag/v2.0.0

.github/workflows/publish.yml
```diff
@@ -12,10 +12,10 @@ jobs:
...
     - uses: actions/setup-node@v1
       with:
-        node-version: 10
+        node-version: 12
```
styfle: We should publish with the latest LTS

.github/workflows/test-unit.yml
TooTallNate: Seems weird that we install node _after_ running `yarn`, right?

packages/now-cli/src/util/dev/server.ts
TooTallNate: Maybe avoid invoking `getPort()` if we're not going to use the value?

```suggestion
    const port = await (this.frameworkSlug === 'redwoodjs' ? getRedwoodPort(cwd) : getPort());
```

.github/workflows/test-unit.yml
styfle: Maybe we need to add `with` to every `run` step?

Take a look at the previous commit before this `with` was moved down here: https://github.com/vercel/vercel/runs/918955588#step:7:192

TooTallNate (dismissed):

TooTallNate (dismissed):

TooTallNate (approved):

AndyBitz (approved):

PR #4942, opened 2020-07-28
[routing-utils] Update header replacing to handle more cases

ijjk:
This adds handling for more cases while updating header values to make sure to escape any characters that could break compiling with `path-to-regexp`

x-ref: https://github.com/vercel/next.js/pull/15592


packages/now-routing-utils/test/superstatic.spec.js
```diff
@@ -534,12 +607,14 @@ test('convertHeaders', () => {
...
     ['404.html'],
     ['/blog/first-post', '/blog/another/one'],
+    ['/like/params/first', '/like/params/second'],
   ];
 
   const mustNotMatch = [
     ['hello/file.jpg', 'hello/font-css', 'dir/arial.font-css'],
     ['403.html', '500.html'],
     ['/blogg', '/random'],
+    ['/non-match', '/like/params'],
```
styfle: What about `/like/params/`? Should that match or not?

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -218,18 +218,42 @@ function replaceSegments(
...
-      return compiler(indexes);
-    })
-    .join('?');
+
+  for (const key of Object.keys(indexes)) {
+    if (value.includes(`:${key}`)) {
+      value = value
+        .replace(
+          new RegExp(`:${key}\\*`, 'g'),
+          `:${key}--ESCAPED_PARAM_ASTERISKS`
```
styfle: Nit pick: this is plural while the others are singular. Perhaps remove the `S`?

ijjk: Updated

packages/now-routing-utils/test/superstatic.spec.js
ijjk: It should not match since the param isn't optional, updated test to reflect this also

styfle (approved):
Beautiful, thanks! 🦄

Timer (approved):
LGTM, as we landed the same code in Next.js! Thanks!

styfle:
@ijjk Don't forget to update the API repo after publishing canary 👍

PR #4945, opened 2020-07-28
[next] Run monorepo fixtures against canary

ijjk:
This updates some monorepo test fixtures to run against the latest canary which highlights an issue with legacy `routes` being used for monorepo set-ups and the new trailing slash redirect in Next.js. 

We might want to add detecting of this case and either error or warn since it creates breaking behavior change


Timer (approved):

styfle:
@ijjk Was this just a 1-off test or do you plan to merge?

Timer:
This should be merged

styfle (approved):

styfle:
The test failed `Should opt-out of shared lambdas when routes are detected`

PR #4946, opened 2020-07-28
[tests] Update test fixture 10-nextjs-node

styfle:


packages/now-cli/src/util/get-files.ts
```diff
@@ -210,10 +210,6 @@ export async function staticFiles(
 
       const accepted = filter(relativePath);
 
-      if (!accepted) {
-        debug(`Ignoring ${file}`);
```
TooTallNate, line 214: ❤️

PR #4947, opened 2020-07-28
[node] Add `res.redirect()` helper method

markozxuu:
This pull request adds a method redirect, the behavior will be the same as that of Next.js with `pages/api`. This PR takes into account the change that is being made in next.js with redirect due to the [error that was reported](https://github.com/vercel/next.js/issues/15594) a few hours ago.


markozxuu:
I am working on adding unit tests for redirect

packages/now-node/src/helpers.ts
```diff
@@ -73,6 +73,18 @@ function status(res: NowResponse, statusCode: number): NowResponse {
   return res;
 }
 
+function redirect(
+  res: NowResponse,
+  url: string,
+  // justification: more explicit code for easy compression
+  // eslint-disable-next-line @typescript-eslint/no-inferrable-types
+  status: number = 307
```
TooTallNate:
```suggestion
  status = 307
```

markozxuu: I know that typescript has type inference implicitly but when declaring it explicitly, I think it helps a better compression of the code. People who are new to TS don't know they have implicit inference.

styfle: > I think it helps a better compression of the code

It doesn't, the types are erased after compilation.

> People who are new to TS don't know they have implicit inference.

I think its the opposite, JS users will know what [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) look like.

packages/now-node/src/helpers.ts
```diff
@@ -73,6 +73,18 @@ function status(res: NowResponse, statusCode: number): NowResponse {
   return res;
 }
 
+function redirect(
+  res: NowResponse,
+  url: string,
```
styfle: This needs to match the usage in Next.js

- `res.redirect(301, '/go')`
  - should have `res.statusCode === 301 && res.headers.location === '/go'`
- `res.redirect('/go')`
  -  should have `res.statusCode === 307 && res.headers.location === '/go`

lucleray: Agreed with @styfle, we want to match the usage in Next.js, which is based on Express. You can see the [related Express docs here](http://expressjs.com/en/4x/api.html#res.redirect).

packages/now-node/src/helpers.ts
markozxuu: I just made the changes suggested by @TooTallNate  @styfle

markozxuu:
@styfle @lucleray  The implementation I just added of `redirect` is the same as Next.js.  I just need to integrate the tests

markozxuu:
How can I create an endpoint to test `res.redirect` ? @styfle 

I need to do something similar to what I did with Next.js
https://github.com/vercel/next.js/pull/15844/files#diff-43c0d69b68a30ea5a008a2b6d1c4ad12R239

packages/now-node/test/helpers.test.js
```diff
@@ -307,6 +308,14 @@ describe('res.status', () => {
   });
 });
 
+describe('res.redirect', () => {
+  test('res.redirect() should redirect without statusCode', () => {
+    mockListener.mockImplementation((req, res) => {
+      res.redirect('example.com');
+    });
```
lucleray, line 356: This test is not testing the behaviour of `res.redirect`. Could you add tests to check that the header and status code is correctly added in the response?

lucleray (changes requested):

packages/now-node/test/helpers.test.js
lucleray: You can use something along these lines:
```
      const res = await fetchWithProxyReq(url);
      expect(res.status).toBe(...);
      expect(res.headers.get('location')).toBe(...);
```

packages/now-node/test/helpers.test.js
```diff
@@ -307,6 +308,47 @@ describe('res.status', () => {
...
+    expect(res.text()).toContain(
+      `Invalid redirect arguments. Please use a single argument URL, e.g. res.redirect('/destination') or use a status code and URL, e.g. res.redirect(307, '/destination').`
+    );
+  });
+  test('should show friendly error in case of passing null as first argument redirect', async () => {
+    mockListener.mockImplementation((req, res) => {
+      res.redirect(null);
+      res.end();
+    });
+    expect(res.text()).toContain(
```
styfle: This seems wrong, we shouldn't respond to the client with the error text. Instead it should throw. Is this assertion passing right now? 🤔

markozxuu: @styfle where are the errors stored ?

styfle (approved):
Looks good to me assuming the tests pass, thanks!

markozxuu:
All tests work : )

lucleray (approved):
Great job @markozxuu 🥇 🙏

styfle:
@markozxuu Thanks! Would you like to update the docs too?

https://vercel.com/docs/runtimes?query=helper#official-runtimes/node-js/node-js-request-and-response-objects/node-js-helpers

markozxuu:
Of course @styfle  ! I will be happy to update the documents :D

mosesoak:
Hey @markozxuu and folks, thanks for the feature add. 

[edit: removed incorrect statement! 🤦‍♂️]

markozxuu:
@mosesoak  Don't Worry. Happy to have contributed in this PR and I hope this feature is useful for your team 🙏 
If a bug happens, don't hesitate to report it.

PR #3439, opened 2019-12-16
[now-cli] Change style for warnings

AndyBitz:
Changes the warning style from Now CLI to make them more distinguishable.

#### Right now
image

#### With change
image

[PRODUCT-775]

[PRODUCT-775]: https://zeit.atlassian.net/browse/PRODUCT-775

PR #4959, opened 2020-07-31
[cli] Adjust output for recently changed domain commands

AndyBitz:
https://app.clubhouse.io/vercel/story/423

Adjust output for recently changed domain commands.

~Requires https://github.com/vercel/docs/pull/2037~

- Fix spinner for `domains inspect` and `ls` command to not jitter
- Adjust `domains inspect` output
- Add deprecation message for `domains verify`
- Adjust table and text for `domains ls`
- Fix bug in query string bug in `domains ls`
- Adjust `domains ls` table
- Remove a test, because we don't list project domains anymore


packages/now-cli/src/commands/domains/inspect.ts
```diff
@@ -108,46 +98,27 @@ export default async function inspect(
...
-    `    ${chalk.cyan('NS Verified At')}\t\t${formatDate(
-      domain.nsVerifiedAt
-    )}\n`
-  );
-  output.print(
-    `    ${chalk.cyan('TXT Verified At')}\t\t${formatDate(
-      domain.txtVerifiedAt
-    )}\n`
-  );
+  output.print(`    ${chalk.cyan('Edge Network')}\t\t${true}\n`);
```
styfle, line 107: This is always true?

AndyBitz: Yes, as far as I know

packages/now-cli/src/commands/domains/inspect.ts
```diff
@@ -208,8 +159,116 @@ export default async function inspect(
...
+    return 1;
+  }
+
+  if (domain instanceof DomainPermissionDenied) {
+    cancelWait();
+    output.error(
+      `You don't have access to the domain ${domainName} under ${chalk.bold(
+        contextName
+      )}`
+    );
```
styfle:
```suggestion
    output.prettyError(domain);
```

packages/now-cli/src/commands/domains/inspect.ts
```diff
@@ -208,8 +159,116 @@ export default async function inspect(
...
+    getDomainPrice(client, domainName, 'renewal')
+      .then(res => (res instanceof Error ? null : res.price))
+      .catch(() => null),
+  ]);
+
+  if (!domain || domain instanceof DomainNotFound) {
+    cancelWait();
+    output.error(
+      `Domain not found by "${domainName}" under ${chalk.bold(contextName)}`
+    );
```
styfle: Can we update `DomainNotFound` to include the `contextName` in the error message? Then we can use prettyError() here too?

packages/now-cli/src/commands/domains/inspect.ts
```diff
@@ -208,8 +159,116 @@ export default async function inspect(
...
+    output.log(`Run ${getCommandName(`domains ls`)} to see your domains.`);
+    return 1;
+  }
+
+  if (domain instanceof DomainPermissionDenied) {
+    cancelWait();
+    output.error(
+      `You don't have access to the domain ${domainName} under ${chalk.bold(
+        contextName
+      )}`
```
styfle:
```suggestion
    output.prettyError(domain);
```

packages/now-cli/src/commands/domains/verify.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!domainName) {
+    output.error(
+      `${getCommandName(`domains verify <domain>`)} expects one argument`
+    );
+
+    return 1;
+  }
+
+  output.error(
```
styfle: Let's use prettyError() here so you dont need to pass `undefined`

packages/now-cli/src/commands/domains/inspect.ts
AndyBitz: I don't think that's possible in this case. Because we have to style the message that we print with `chalk.bold`.

packages/now-cli/src/util/domains/get-domain-by-name.ts
```diff
@@ -33,7 +35,7 @@ async function getDomainByName(
     }
 
     throw error;
+  } finally {
+    cancelWait?.();
```
styfle, line 39: TIL you can use `?.` to invoke a function 🤯

packages/now-cli/src/util/domains/purchase-domain-if-available.ts
```diff
@@ -30,12 +30,19 @@ export default async function purchaseDomainIfAvailable(
     }
 
     output.debug(`Domain ${domain} is available to be purchased`);
-    const domainPrice = await getDomainPrice(client, domain);
-    cancelWait();
+
+    const domainPrice = await getDomainPrice(client, domain).finally(() => {
+      cancelWait();
```
styfle, line 35: I think you can remove the `cancelWait()` on line 78

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -52,6 +55,7 @@ export default function createOutput({ debug: debugEnabled = false } = {}) {
             right: 1,
           },
           borderColor: 'yellow',
+          ...options?.boxen,
```
styfle, line 51: Can we remove the "workaround" from #3439 by updating `boxen`?

packages/now-cli/src/commands/domains/inspect.ts
styfle: Why can't we use `chalk.bold` in the error class?

packages/now-cli/src/commands/domains/verify.ts
```diff
@@ -18,12 +18,16 @@ export default async function verify(
     return 1;
   }
 
-  output.error(
+  const error = new Error(
```
styfle: You can use `NowBuildError` if you dont want to create a custom error with `action` and `link` properties

packages/now-cli/src/util/output/create-output.ts
AndyBitz: Seems like it was already using the latest version, so I just removed the workaround.

packages/now-cli/src/commands/domains/inspect.ts
AndyBitz: I guess we can, but we're not doing it for any other error it seems.
Not sure if having an error message created with chalk makes any difference though.

AndyBitz: Done

styfle: Do a search for `under ${context}` in `errors-ts.ts` and there are many messages that print the context but not in bold:

https://github.com/vercel/vercel/blob/421be5d7386c6c7b4da9cf2daf5fa87738b55c0c/packages/now-cli/src/util/errors-ts.ts#L120

Let's decide if bold is necessary (probably a @leo decision) and either use it for all of these errors or none of these errors.

styfle (approved):

mglagola (approved):

packages/now-cli/src/commands/domains/inspect.ts
leo: This is how it loosk at the moment:

image

I think the domain needs to be bold too.

javivelasco (approved):

PR #4965, opened 2020-08-02
[cli] Fix `.gitignore` append during project link

kidonng:
~~On Windows~~, if `.gitignore` is using CRLF as EOL, the Vercel CLI will always add `.vercel` into `.gitignore` regardless of its existence. This is because the code is using LF (`\n`) to split the lines.

image

This PR should fix the problem, by detecting and using the EOL of `.gitignore` instead of hard-coded LF (`\n`).

EDIT: after submitting the PR, I realized it has nothing to do with the system line ending. I should check the file line ending instead.

### DIFF
--- a/packages/now-cli/src/util/projects/link.ts
+++ b/packages/now-cli/src/util/projects/link.ts
@@ -244,7 +244,7 @@ export async function linkFolderToProject(
     const gitIgnore = await readFile(gitIgnorePath)
       .then(buf => buf.toString())
       .catch(() => null);
-    const EOL = gitIgnore && gitIgnore.includes('\r\n') ? '\r\n' : '\n'
+    const EOL = gitIgnore && gitIgnore.includes('\r\n') ? '\r\n' : '\n';
 
     if (!gitIgnore || !gitIgnore.split(EOL).includes(VERCEL_DIR)) {
       await writeFile(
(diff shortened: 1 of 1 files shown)


packages/now-cli/src/util/projects/link.ts
```diff
@@ -244,11 +244,12 @@ export async function linkFolderToProject(
...
       .then(buf => buf.toString())
       .catch(() => null);
+    const EOL = gitIgnore && gitIgnore.includes('\r\n') ? '\r\n' : '\n';
 
-    if (!gitIgnore || !gitIgnore.split('\n').includes(VERCEL_DIR)) {
+    if (!gitIgnore || !gitIgnore.split(EOL).includes(VERCEL_DIR)) {
       await writeFile(
         gitIgnorePath,
-        gitIgnore ? `${gitIgnore}\n${VERCEL_DIR}` : VERCEL_DIR
+        gitIgnore ? `${gitIgnore}${EOL}${VERCEL_DIR}` : VERCEL_DIR
```
TooTallNate: While we're updating this, can we add the trailing newline as well? It's always bothered me that it's not there:

```suggestion
        gitIgnore ? `${gitIgnore}${EOL}${VERCEL_DIR}${EOL}` : `${VERCEL_DIR}${EOL}`
```

styfle (dismissed):
Thanks!

lucleray (approved):

styfle (approved):

TooTallNate:
Thank you!

PR #4979, opened 2020-08-04
[next] Run Prettier

Timer:
This PR strictly formats files so that their diffs don't noise up other PRs.


packages/now-next/src/index.ts
```diff
@@ -1,20 +1,3 @@
...
-  getLambdaOptionsFromFunction,
-  getNodeVersion,
-  getSpawnOptions,
-  getScriptName,
-  glob,
-  runNpmInstall,
-  runPackageJsonScript,
-  execCommand,
-  getNodeBinPath,
-} = buildUtils;
```
styfle, line 16: This is necessary for legacy support

Timer: Could you explain more? I just moved the code down a few lines—I didn't get rid of it.

styfle: :oh:

Did prettier re-order these?

Lets keep the `import buildUtils from './build-utils';` just above the deconstruction there.

Timer: This change was done by the auto import formatter, I don't think we should be trying to fight against it. It'll just get moved again in the future.

image

There's actually ESLint rules that ban JavaScript between `import`s, so I don't think we should move this back.

styfle (approved):

TooTallNate (approved):

PR #4981, opened 2020-08-04
[cli] Add `bundledDependencies` to package.json

styfle:
This PR adds the [`bundledDependencies`](https://docs.npmjs.com/files/package.json#bundleddependencies) property to `package.json` so that `npm publish` will include a copy of the `dependencies` in the tarball.

This should reduce the time it takes to run `npm install -g vercel` because it will require fewer http requests.

We will need to make sure that `bundledDependencies` matches `dependencies` in the future.

### DIFF
--- a/packages/now-cli/package.json
+++ b/packages/now-cli/package.json
@@ -72,6 +72,17 @@
     "@vercel/static-build": "0.17.7-canary.1",
     "update-notifier": "4.1.0"
   },
+  "bundledDependencies": [
+    "@vercel/build-utils",
+    "@vercel/go",
+    "@vercel/next",
+    "@vercel/node",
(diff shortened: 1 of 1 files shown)


timneutkens (approved):
💯 please let me know the difference in install time 👍

styfle:
I published to npm as `vercel@20.0.0-bundled.1` as tested the time for `npm install` and there is no difference. Both the old canary and new bundled install are around 11 seconds after trying 5 times.

```sh
time npm install vercel@20.0.0-canary.12 # before
time npm install vercel@20.0.0-bundled.1 # after
```

I don't see any benefit in using `bundledDependencies` so I'll close this PR.

PR #4984, opened 2020-08-05
[examples] Update Next.js Example

lfades:
Related to https://github.com/vercel/docs/pull/2047 - The Next.js docs page will mention features from 9.5, but the current starter is quite outdated.


Timer (approved):

TooTallNate (approved):

examples/nextjs/.gitignore
```diff
@@ -11,20 +11,24 @@
...
 yarn-error.log*
 
-# Environment Variables
-.env
-.env.build
+# local env files
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
styfle, line 31: This could probably use a wildcard to ignore all local files:

```
.env.local
.env.*.local
```

lfades: @styfle True and it's intentional. We made this explicit for now in create-next-app

PR #4999, opened 2020-08-06
[redwood] Fix file permissions and add support for `build` script

styfle:
Some files require execution privileges, such as Prisma, so we must preserve the file mode.

We also want redwood to behave the same as other frameworks and use `yarn build` if available.

### DIFF
--- a/packages/redwood/src/index.ts
+++ b/packages/redwood/src/index.ts
@@ -11,6 +11,7 @@ import {
   getNodeVersion,
   getSpawnOptions,
   runNpmInstall,
+  runPackageJsonScript,
   execCommand,
   FileBlob,
   FileFsRef,
@@ -54,10 +55,6 @@ export async function build({
     meta
   );
 
-  const {
-    buildCommand = 'yarn rw db up --no-db-client --auto-approve && yarn rw build',
-  } = config;
-
   if (meta.isDev) {
     debug('Detected @vercel/redwood dev, returning routes...');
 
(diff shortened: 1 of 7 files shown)


MaxLeiter (dismissed):

MaxLeiter:
Do we want to test this?

packages/redwood/src/index.ts
```diff
@@ -119,9 +120,10 @@ export async function build({
       }),
     };
 
-    dependencies.forEach(fsPath => {
-      lambdaFiles[relative(workPath, fsPath)] = new FileFsRef({ fsPath });
-    });
+    for (const fsPath of dependencies) {
+      const { mode } = statSync(fsPath);
+      lambdaFiles[relative(workPath, fsPath)] = new FileFsRef({ fsPath, mode });
```
TooTallNate:
```suggestion
      lambdaFiles[relative(workPath, fsPath)] = await FileFsRef.fromFsPath({ fsPath });
```

packages/redwood/src/index.ts
```diff
@@ -1,4 +1,5 @@
 import { join, dirname, relative, parse as parsePath, sep } from 'path';
+import { statSync } from 'fs';
```
TooTallNate: Can be removed with my suggestion [below](https://github.com/vercel/vercel/pull/4999/files#r466688471) (should be `lstat()` anyways 😉).

```suggestion
```

TooTallNate:
Probably should add a test…

packages/redwood/src/index.ts
styfle: Aha, I was wondering why the mode wasn't automatically set. This explains it 👍

TooTallNate (approved):

AndyBitz (approved):

PR #5008, opened 2020-08-07
Make timer type safe

jaredpalmer:
This may break other ts issues, but this is the correct way to type `.timer` AFAICT


styfle (commented):
This is failing to build in CI. Does it work with `yarn build` locally?

PR #5020, opened 2020-08-10
[redwood] Update default build command

styfle:
This PR updates the default build command per @thedavidprice 

It also uses `@vercel/frameworks` as a source of truth for the build command.

If the user has a `build` script in `package.json`, that will take precedence over the default.


packages/redwood/src/index.ts
```diff
@@ -77,25 +82,37 @@ export async function build({
...
+    .buildCommand;
+  const pkgPath = join(workPath, 'package.json');
+  const pkg = JSON.parse(readFileSync(pkgPath, 'utf8')) as PackageJson;
+  if (hasScript('vercel-build', pkg)) {
+    debug(`Executing "yarn vercel-build"`);
+    await runPackageJsonScript(workPath, 'vercel-build', spawnOpts);
+  } else if (hasScript('build', pkg)) {
+    debug(`Executing "yarn build"`);
+    await runPackageJsonScript(workPath, 'build', spawnOpts);
+  } else if (buildCommand) {
```
AndyBitz: I think this has to be the first in this list, so it's considered first before `vercel-build` and `build`.

packages/frameworks/frameworks.json
```diff
@@ -700,13 +700,13 @@
     },
     "settings": {
       "buildCommand": {
-        "value": "yarn rw db up --no-db-client --auto-approve && yarn rw build"
+        "value": "yarn rw build && yarn rw db up --no-db-client --auto-approve && yarn rw dataMigrate up"
```
styfle, line 703: This is the command recommended here: https://github.com/redwoodjs/create-redwood-app/blob/21960fe7cd9c41c1546d239c6950a6db4199b6f5/netlify.toml#L2

packages/redwood/src/index.ts
styfle: You're right, I fixed it!

AndyBitz (dismissed):

TooTallNate (approved):

AndyBitz (approved):

PR #5022, opened 2020-08-11
[client][cli] Add Learn More Option to Warnings

dav-is:
The deployment API might return a learn more link to address warnings, tips, or notices

CH-568

```
HTTP/2 200
...
x-vercel-warning-bad-error: Something went wrong
x-vercel-link-bad-error: https://vercel.link/bad-error

x-vercel-notice-low-funds: You are low on funds
x-vercel-link-low-funds: https://vercel.link/increase-funds
...
```

Should output as:
```
Warning: Something went wrong
Learn More: https://vercel.link/bad-error

Notice: You are low on funds
Learn More: https://vercel.link/increase-funds
```


TooTallNate:
Side note, but should we update `x-now` to `x-vercel`?

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +63,22 @@ async function* createDeployment(
       };
     }
 
+    const regex = /^x-now-(?:warning|notice|tip)-(.*)$/;
     for (const [name, value] of dpl.headers.entries()) {
       if (name.startsWith('x-now-warning-')) {
         debug('Deployment created with a warning:', value);
-        yield { type: 'warning', payload: value };
+        const info = dpl.headers[`x-now-info-${name.match(regex)[1]}`];
```
styfle:
```suggestion
        const info = dpl.headers[name.match(regex)[1]];
```

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +63,22 @@ async function* createDeployment(
...
+        yield { type: 'warning', payload: value, info };
       }
       if (name.startsWith('x-now-notice-')) {
         debug('Deployment created with a notice:', value);
-        yield { type: 'notice', payload: value };
+        const info = dpl.headers[`x-now-info-${name.match(regex)[1]}`];
```
styfle:
```suggestion
        const info = dpl.headers[name.match(regex)[1]];
```

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +63,22 @@ async function* createDeployment(
...
+        yield { type: 'notice', payload: value, info };
       }
       if (name.startsWith('x-now-tip-')) {
         debug('Deployment created with a tip:', value);
-        yield { type: 'tip', payload: value };
+        const info = dpl.headers[`x-now-info-${name.match(regex)[1]}`];
```
styfle:
```suggestion
        const info = dpl.headers[name.match(regex)[1]];
```

styfle (commented):
We need a test for this new behavior. Is there an existing API we no returns this information?

packages/now-client/src/deploy.ts
dav-is: This suggestion doesn't make any sense to me

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -162,7 +162,11 @@ const printDeploymentStatus = async (
         prependEmoji(
           `${chalk.dim(indication.payload)}`,
           emoji(indication.type)
-        ) + `\n`
+        ) +
+          `\n` +
+          indication.info
+          ? `Learn More: ${indication.info}`
```
styfle: For errors, we use two properties:
- `action: 'Learn More'`
- `link: https://vercel.link/foo`

Can we use `action` and `link` for warnings too?

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -162,7 +162,13 @@ const printDeploymentStatus = async (
         prependEmoji(
           `${chalk.dim(indication.payload)}`,
           emoji(indication.type)
-        ) + `\n`
+        ) +
+          `\n` +
+          indication.link
+          ? `${indication.action ? indication.action : 'Learn More: '}${
```
styfle:
```suggestion
          ? `${indication.action || 'Learn More'}: ${
```

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +64,31 @@ async function* createDeployment(
       };
     }
 
+    const regex = /^x-(?:vercel|now)-(?:warning|notice|tip)-(.*)$/;
```
styfle: I think you can utilize this regex to group match and remove all the `startsWith()` checks below. The group `(warning|notice|tip)` can be returned as `type`

styfle (dismissed):
Beautiful 🌈

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +64,19 @@ async function* createDeployment(
...
-        yield { type: 'warning', payload: value };
-      }
-      if (name.startsWith('x-now-notice-')) {
-        debug('Deployment created with a notice:', value);
-        yield { type: 'notice', payload: value };
-      }
-      if (name.startsWith('x-now-tip-')) {
-        debug('Deployment created with a tip:', value);
-        yield { type: 'tip', payload: value };
+    const indications = ['warning', 'notice', 'tip'];
```
styfle:
```suggestion
    const indications = new Set(['warning', 'notice', 'tip']);
```

packages/now-client/src/deploy.ts
```diff
@@ -59,18 +64,19 @@ async function* createDeployment(
...
+    const regex = /^x-(?:vercel|now)-(warning|notice|tip)-(.*)$/;
+    for (const [name, payload] of dpl.headers.entries()) {
+      const match = name.match(regex);
+      if (match) {
+        const [, type, identifier] = match;
+        const action = dpl.headers[`x-vercel-action-${identifier}`];
+        const link = dpl.headers[`x-vercel-link-${identifier}`];
+
+        if (indications.includes(type)) {
```
styfle:
```suggestion
        if (indications.has(type)) {
```

styfle:
I contacted GitHub support because our CI has been failing way too often. It sounds like they are rolling out changes that updates Git 2.28 and that might be why we see it failing during `git diff` 🤷

styfle:
@dav-is Is the API using these headers yet? We should add a test here to make sure it works correctly.

dav-is:
The deployments API is now adding a `x-vercel-link-` header. Here is how it looks now:

Screenshot-2020-08-13-16:21

styfle:
@dav-is Let's add an integration test similar to this one to make sure the learn more is printed: 
https://github.com/vercel/vercel/blob/828244293934439370463b050eec5514323c1c32/packages/now-cli/test/integration.js#L2292-L2299

Probably add a new fixture with the required next.js settings here: https://github.com/vercel/vercel/blob/828244293934439370463b050eec5514323c1c32/packages/now-cli/test/helpers/prepare.js#L258

packages/now-cli/test/helpers/prepare.js
```diff
@@ -241,6 +241,25 @@ module.exports = async function prepare(session) {
         },
       }),
     },
+    'zero-config-next-js-functions-warning': {
+      'pages/index.js':
+        'export default () => <div><h1>Now CLI test</h1><p>Zero-config + Next.js</p></div>',
```
styfle:
```suggestion
        'export default () => <div><h1>Vercel CLI test</h1><p>Zero-config + Next.js</p></div>',
```

packages/now-cli/test/helpers/prepare.js
```diff
@@ -241,6 +241,25 @@ module.exports = async function prepare(session) {
...
+      'vercel.json':
+        '{"version":2,"functions":{"pages/index.js":{"runtime": "@vercel/php@0.1.0"}}}',
+      'package.json': JSON.stringify({
+        name: 'zero-config-next-js-test',
```
styfle: You can remove `name` here

dav-is:
```suggestion

```

styfle (dismissed):
Wonderful, thanks! 🥇

styfle (approved):
Great, thanks! 🎉

TooTallNate (approved):

PR #5030, opened 2020-08-12
[python] Support {exclude,include}Files for minimizing monorepo bundle size

sisp:
This PR actually fixes a couple of related things in the `@vercel/python` package.

First and foremost, it adds proper support for setting `{exclude,include}Files` in the build config for building a Python function with `@vercel/python`. This is important for minimizing the AWS Lambda bundle size, especially when deploying to Vercel from a monorepo.

For example:

```json
{
  "builds": [
    {
      "use": "@vercel/python",
      "entrypoint": "api/py-func-a/index.py",
      "config": {
        "excludeFiles": "**",
        "includeFiles": "api/py-func-a/**"
      }
    },
    {
      "use": "@vercel/python",
      "entrypoint": "api/py-func-b/index.py",
      "config": {
        "excludeFiles": "**",
        "includeFiles": "api/py-func-b/**"
      }
    }
  ]
}
```

Currently, `@vercel/python` only supports `excludeFiles` which could be used with a negative pattern to achieve the same goal, but [`node-glob` has a bug](https://github.com/isaacs/node-glob/issues/409) which prevents this from working. Besides, the DX would not be as good with a negative ignore pattern. Having both `excludeFiles` and `includeFiles` is a lot more natural and works around the negative patterns bug in `node-glob`. The code for filtering files is borrowed from `@vercel/now-ruby`. I guess it could/should be refactored at some point to reduce duplicate code.

While at it, I believe installing dependencies in the work path is not good practice because there could be name collisions of package files/directories with existing user files in the work path. Therefore, I've changed the target directory of the installed packages to `now__pypackages` and dynamically add this directory to the search path in `now_init.py`/`now__handler__python.py`.

Another minor malpractice is to override the `PYTHONPATH` environment variable without resetting it later. This is currently done for adding the package directory to the search path in which `pipfile2req` is temporarily installed. I've fixed this, too.

And finally, I don't think naming the cache directory as the basename of the entrypoint when running `vercel` in dev mode is a good idea. In a monorepo, it's likely that the entrypoint file of multiple/all Python functions is `index.py` in some subdirectory. This would lead to name collisions in the cache root. Instead, I've changed this behavior to consider the entire path and replace path separators by `__`, so each function gets its own unique cache directory.

Sorry for making a couple of changes in a single PR, but they are all more or less related and crucial for using `@vercel/python` especially with monorepos.


sisp:
It looks like tests are only failing because this is a 3rd party PR:

> Failed to create test deployment. This is expected for 3rd-party Pull Requests. Please run tests locally.

packages/now-python/src/index.ts
```diff
@@ -166,16 +199,39 @@ export const build = async ({
...
-  const globOptions: GlobOptions = {
-    cwd: workPath,
-    ignore:
-      config && typeof config.excludeFiles === 'string'
-        ? config.excludeFiles
-        : 'node_modules/**',
-  };
+  const outputFiles = await glob('**', workPath);
+
+  // Static analysis is impossible with Python. Instead, provide `includeFiles`
```
styfle: That's not true--static analysis is possible with python and we would prefer to have that instead of include/exclude

packages/now-python/src/index.ts
```diff
@@ -116,9 +146,11 @@ export const build = async ({
 
     // Python needs to know where to look up all the packages we just installed.
     // We tell it to use the same location as used with `--target`
+    const pythonPath = process.env.PYTHONPATH;
     process.env.PYTHONPATH = tempDir;
     const convertCmd = join(tempDir, 'bin', 'pipfile2req');
     await pipenvConvert(convertCmd, pipfileLockDir);
+    process.env.PYTHONPATH = pythonPath;
```
styfle: Why set PYTHONPATH twice? I think the comment above needs to be updated

packages/now-python/now_init.py
```diff
@@ -1,11 +1,14 @@
...
+import os
 import sys
 import base64
 import json
 import inspect
 from importlib import util
 from http.server import BaseHTTPRequestHandler
 
+# Add the Python packages directory to the search path.
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '__NOW_PACKAGES_DIR'))
```
styfle: Does this work with python 3.6, 3.7, and 3.8?

packages/now-python/src/index.ts
sisp: If this were true, it would be even better. I'm not aware of any solution though. The highly dynamic nature of Python makes it difficult to statically trace all required dependencies. E.g. Pylint, which is one of the most sophisticated linters for Python, is finding it difficult to detect, e.g., some TensorFlow or OpenCV imports.

What's your envisioned approach for static analysis? The only library that I could find is [`treeshaker`](https://github.com/sclabs/treeshaker), but it's a very young project and I'd expect edge cases where it may fail. I haven't tried it though.

I think it's valuable to have include/exclude as a manual fallback in any case. When static analysis comes at some point, great, but include/exclude solves a real problem right now.

packages/now-python/src/index.ts
sisp: In order to reset `PYTHONPATH` to its original value again. It is only set to find `pipfile2req`, but after `await pipenvConvert(convertCmd, pipfileLockDir);` has been executed, any packges installed in `tempDir` should not be in the search path anymore for the rest of the script.

I just noticed that `process.env.PYTHONPATH = pythonPath;` is actually not correct. It should be:
```ts
if (pythonPath === undefined) {
  delete process.env.PYTHONPATH;
} else {
  process.env.PYTHONPATH = pythonPath;
}
```
I'll fix it once you agree with resetting `process.env.PYTHONPATH` in general.

packages/now-python/now_init.py
sisp: It does (also confirmed with a small test).

packages/now-python/src/index.ts
styfle: Python has the `ast` module which can analyze itself. https://docs.python.org/3/library/ast.html

I don't think we want to introduce `includeFiles` here because there is nothing additional to include. You're either including everything or you are excluding somethings.

Let's revert the `includeFiles` part of this PR.

packages/now-python/src/index.ts
styfle: Seems good 👍

packages/now-python/src/index.ts
```diff
@@ -42,7 +64,11 @@ export async function downloadFilesInWorkPath({
   if (meta.isDev) {
     // Old versions of the CLI don't assign this property
     const { devCacheDir = join(workPath, '.now', 'cache') } = meta;
-    const destCache = join(devCacheDir, basename(entrypoint, '.py'));
+    const entrypointCacheDir = join(
+      dirname(entrypoint),
+      basename(entrypoint, '.py')
+    ).replace(new RegExp(sep, 'g'), '__');
```
styfle: Why use `__` here?

packages/now-python/src/index.ts
sisp: > Python has the `ast` module which can analyze itself. https://docs.python.org/3/library/ast.html

I'm quite sure it only works for static imports. For instance, [Pylint has issues detecting TensorFlow imports](https://github.com/PyCQA/pylint/issues/2603). Static analysis may be sugar for cases where it works, but I don't think it'll be as reliable as treeshaking is in Javascript.

> I don't think we want to introduce `includeFiles` here because there is nothing additional to include. You're either including everything or you are excluding somethings.
>
> Let's revert the `includeFiles` part of this PR.

I beg to differ. How would you express

```json
{
  "excludeFiles": "**",
  "includeFiles": "api/py-func-a/**"
}
```

without both? The idea is to have a way of expressing something like `onlyFiles` with a glob that tells which files to include while excluding everything else. In a monorepo this is a common use case with a possibly large number of functions where each function is located in a directory `api/{function}`. As mentioned above, negative ignore patterns don't work because of a bug in `node-glob`.

packages/now-python/src/index.ts
sisp: No particular reason, just a delimiter that seemed unlikely to occur in the original file and directory names. Would you prefer another one?

packages/now-python/src/index.ts
sisp: Done

packages/now-python/src/index.ts
styfle: I mean, why was this changed? Was there a bug?

sisp: Before, it was
```ts
const destCache = join(devCacheDir, basename(entrypoint, '.py'));
```
and the problem is that there could be multiple APIs with the same entrypoint file name, e.g.:
```
api/py-func-a/index.py
api/py-func-b/index.py
```
Then, the cache directory is `{devCacheDir}/index` for both, which is a problem. Now, the full path is used as the cache directory name, in this example `{devCacheDir}/api__py-func-a__index` and `{devCacheDir}/api__py-func-b__index`, so the cache directories are unique.

styfle (commented):
We need integration tests for the new behavior.

See an example test fixture here and note the `probes` property: https://github.com/vercel/vercel/tree/master/packages/now-python/test/fixtures/25-exclude-files

sisp:
I've extended the `25-exclude-files` integration test to test `includeFiles` as well and added `26-exclude-dirs` to test excluding/including directories and the files inside them with recursive glob patterns, similar to my original example.

Thanks for referencing the existing tests and the hint regarding the `probes` property, this was very helpful!

packages/now-python/src/index.ts
styfle: @TooTallNate What do you think. Should we follow the same pattern here as `@vercel/node`?

packages/now-python/src/index.ts
styfle: Let's change the comment from `impossible` to `difficult`. I started work on a python equivalent to [@vercel/nft](https://github.com/vercel/nft) using `ast` and it it seems possible. Analyzing imported files is much easier than linting.

sisp: > Let's change the comment from `impossible` to `difficult`.

Agreed, done.

> I started work on a python equivalent to @vercel/nft using `ast` and it it seems possible.

Sounds great, looking forward to seeing it in the Python runtime some time. Again, I don't have hands-on experience with static import analysis in Python, but, e.g., [Nuitka](https://github.com/Nuitka/Nuitka) (a Python compiler) seems to have some difficulties compiling packages with C extensions, e.g. Pandas, NumPy (there is a plugin for Nuitka specifically for NumPy), scikit-learn, etc. I definitely encourage you to pursue this path - it would be a very nice feature to have -, I just suspect that it's non-trivial to get working reliably for non-trivial Python programs.

packages/now-python/src/index.ts
sisp: Any progress here?

sisp: @styfle @TooTallNate Do you require any further changes here?

sisp:
@styfle @TooTallNate Anything left to do here? Could you merge this PR?

packages/now-python/src/index.ts
```diff
@@ -29,6 +30,27 @@ async function pipenvConvert(cmd: string, srcDir: string) {
...
+    typeof configPatterns === 'string' ? [configPatterns] : configPatterns;
+
+  const patternPaths = await Promise.all(
+    patterns.map(async pattern => {
+      const files = await glob(pattern, workPath);
+      return Object.keys(files);
+    })
+  );
+
+  return ([] as string[]).concat(...patternPaths);
```
TooTallNate: Why not just `return patternPaths`?

packages/now-python/src/index.ts
TooTallNate: Why have the `.replace(new RegExp(sep, 'g'), '__')` part at all? Creating subdirectories makes sense to me, and I believe we do it like that in other Runtimes.

packages/now-python/src/index.ts
sisp: Good point, I've changed it as you suggested.

sisp:
@TooTallNate I've made the changes as you suggested.

sisp:
@TooTallNate Anything else needed to merge this PR?

sisp:
I've merged the latest state of the `main` branch into this one. @TooTallNate @styfle Any chance this PR gets merged soon?

sisp:
@vercel Any chance somebody could review this PR? It's been around for almost 2 years. :confused:

PR #5034, opened 2020-08-13
[build-utils][frameworks] Add support for runtime properties

styfle:
This PR adds two properties to `frameworks.json`:

1. `useRuntime` - this moves the special case for non `@vercel/static-build` frontends, so that any framework can do the same as Next.js and RedwoodJS
2. `ignoreRuntimes` - this allows a framework to opt out of api detection such as RedwoodJS which handle's its own `.js` extensions

This also fixes 2 bugs discovered during implementing the feature:

1. `test-unit.yml` was not testing Node 12, it was testing 10 for both runs
2. `sortFilesBySegmentCount()` was non-deterministic causing node 10 and 12 to sort differently

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -27,6 +27,8 @@ jobs:
       - run: git fetch origin ${{ github.ref }} --depth=100
       - run: git diff origin/master...HEAD --name-only
       - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
       - run: yarn install
       - run: yarn run build
       - run: yarn run lint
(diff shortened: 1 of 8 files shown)


.github/workflows/test-unit.yml
```diff
@@ -27,6 +27,8 @@ jobs:
       - run: git fetch origin ${{ github.ref }} --depth=100
       - run: git diff origin/master...HEAD --name-only
       - uses: actions/setup-node@v1
+        with:
```
TooTallNate, line 30: Do we need to add this on other workflow files?

styfle: Nope, the others already have it. This was the only workflow file missing `with`.

TooTallNate (approved):

AndyBitz (approved):

PR #5036, opened 2020-08-13
[cli][redwood] Update frontend framework detection in `vc dev`

styfle:
This PR uses the the new property added in PR #5034 to determine if a framework has a frontend runtime defined.

It also reverts a couple workarounds add in PR #4937 which was added for RedwoodJS which is no longer necessary since RedwoodJS defines a frontend framework the same way Next.js does.

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -20,6 +20,7 @@ import { ChildProcess } from 'child_process';
 import isPortReachable from 'is-port-reachable';
 import deepEqual from 'fast-deep-equal';
 import which from 'which';
+import npa from 'npm-package-arg';
 
 import { getVercelIgnore, fileNameSymbol } from '@vercel/client';
 import {
@@ -40,6 +41,7 @@ import {
   spawnCommand,
   isOfficialRuntime,
 } from '@vercel/build-utils';
+import _frameworks, { Framework } from '@vercel/frameworks';
 
 import link from '../output/link';
 import { Output } from '../output';
@@ -84,6 +86,11 @@ import {
   EnvConfigs,
 } from './types';
 import { ProjectSettings } from '../../types';
+
+const frameworkList = _frameworks as Framework[];
+const frontendRuntimeSet = new Set(
+  frameworkList.map(f => f.useRuntime?.use || '@vercel/static-build')
+);
 
 interface FSEvent {
   type: string;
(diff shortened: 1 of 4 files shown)


packages/now-cli/src/util/dev/server.ts
```diff
@@ -85,6 +87,11 @@ import {
 } from './types';
 import { ProjectSettings } from '../../types';
 
+const frameworkList = _frameworks as Framework[];
```
TooTallNate, line 90: Why is the cast necessary? Does `@vercel/frameworks` not properly set the type?

styfle: @TooTallNate Indeed, I have been saying that since its inception. Its source is written as JSON, not TS.

All the usages look like this:

https://github.com/vercel/vercel/blob/f1b426097ac1022fb42da0fb7e198287831dc197/api/frameworks.ts#L5

styfle: Oh I think we can actually solve this by defining a default export in `.d.ts` 🤔

packages/now-cli/src/util/dev/server.ts
```diff
@@ -2332,10 +2332,9 @@ async function checkForPort(
 }
 
 function filterFrontendBuilds(build: Builder) {
-  return (
-    !isOfficialRuntime('static-build', build.use) &&
-    !isOfficialRuntime('next', build.use)
-  );
+  const parsed = npa(build.use);
+  const { name } = parsed;
```
TooTallNate:
```suggestion
  const { name } = npa(build.use);
```

packages/redwood/src/index.ts
```diff
@@ -61,21 +61,9 @@ export async function build({
...
   if (meta.isDev) {
-    debug('Detected @vercel/redwood dev, returning routes...');
-
-    let srcBase = mountpoint.replace(/^\.\/?/, '');
-
-    if (srcBase.length > 0) {
-      srcBase = `/${srcBase}`;
-    }
-
+    console.log('WARN: Detected @vercel/redwood dev but this is not supported');
```
TooTallNate: Is this expected to ever happen?

packages/now-cli/src/util/dev/server.ts
styfle: Maybe not because we use named exports so we can use the CJS "default" export as far as I can see

https://github.com/vercel/vercel/blob/f1b426097ac1022fb42da0fb7e198287831dc197/packages/frameworks/index.d.ts

packages/redwood/src/index.ts
styfle: Nope, should I throw instead?

TooTallNate: Ya I think so.

packages/now-cli/src/util/dev/server.ts
styfle: Let's convert `@vercel/frameworks` to TS in another PR 👍

packages/redwood/src/index.ts
styfle: Done 👍

TooTallNate (approved):

MaxLeiter (approved):

PR #5040, opened 2020-08-14
[routing-utils] Update rewrite params query appending

ijjk:
This updates to not automatically append params to the query for rewrites if one or more of the params are already used in the destination's path. No other behavior is being changed and if the user still wants the params in the query after using them in the destination's path they can manually add them like with redirects.

x-ref: https://github.com/vercel/next.js/pull/16189

### DIFF
--- a/packages/now-routing-utils/src/superstatic.ts
+++ b/packages/now-routing-utils/src/superstatic.ts
@@ -5,7 +5,6 @@
 import { parse as parseUrl, format as formatUrl } from 'url';
 import { pathToRegexp, compile, Key } from 'path-to-regexp';
 import { Route, NowRedirect, NowRewrite, NowHeader } from './types';
-import { isString } from 'util';
 
 const UN_NAMED_SEGMENT = '__UN_NAMED_SEGMENT__';
 
@@ -81,7 +80,6 @@ export function convertRewrites(rewrites: NowRewrite[]): Route[] {
       const route: Route = { src, dest, check: true };
       return route;
     } catch (e) {
-      console.error(e);
       throw new Error(`Failed to parse rewrite: ${JSON.stringify(r)}`);
     }
   });
@@ -197,9 +195,11 @@ function replaceSegments(
         // params from the destination
       }
 
-      destParams = new Set([...pathnameKeys, ...hashKeys]
-        .map(key => key.name)
-        .filter(isString));
+      destParams = new Set(
+        [...pathnameKeys, ...hashKeys]
+          .map(key => key.name)
+          .filter(val => typeof val === 'string') as string[]
+      );
 
       pathname = safelyCompile(pathname, indexes);
       hash = hash ? safelyCompile(hash, indexes) : null;
@@ -214,7 +214,8 @@ function replaceSegments(
     }
 
     // We only add path segments to redirect queries if manually
-    // specified
+    // specified and only automatically add them for rewrites if one
+    // or more params aren't already used in the destination's path
     const paramKeys = Object.keys(indexes);
 
     if (!isRedirect && !paramKeys.some(param => destParams.has(param))) {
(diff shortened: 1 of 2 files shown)


dav-is (dismissed):
Nice

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -181,7 +183,24 @@ function replaceSegments(
...
+        pathToRegexp(pathname, pathnameKeys);
+        pathToRegexp(hash || '', hashKeys);
+      } catch (_) {
+        // this is not fatal so don't error when failing to parse the
+        // params from the destination
+      }
+
+      destParams = [...pathnameKeys, ...hashKeys]
+        .map(key => key.name)
+        .filter(isString);
```
styfle:
```suggestion
      destParams = new Set([...pathnameKeys, ...hashKeys]
        .map(key => key.name)
        .filter(isString));
```

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -196,10 +215,12 @@ function replaceSegments(
 
     // We only add path segments to redirect queries if manually
     // specified
-    if (!isRedirect) {
-      for (const [name, value] of Object.entries(indexes)) {
-        if (!(name in query) && name !== UN_NAMED_SEGMENT) {
-          query[name] = value;
+    const paramKeys = Object.keys(indexes);
+
+    if (!isRedirect && !paramKeys.some(param => destParams.includes(param))) {
```
styfle:
```suggestion
    if (!isRedirect && !paramKeys.some(param => destParams.has(param))) {
```

packages/now-routing-utils/src/superstatic.ts
styfle: Probably wont see much of a difference here but maybe depending on how many keys are used

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -196,10 +215,12 @@ function replaceSegments(
 
     // We only add path segments to redirect queries if manually
     // specified
```
styfle: Let's update the comment here

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -181,7 +183,24 @@ function replaceSegments(
       indexes[name] = toSegmentDest(index);
     });
 
+    let destParams: string[] = [];
```
styfle:
```suggestion
    let destParams = new Set<string>();
```

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -181,7 +183,24 @@ function replaceSegments(
...
+
     if (destination.includes(':') && segments.length > 0) {
+      const pathnameKeys: Key[] = [];
+      const hashKeys: Key[] = [];
+
+      try {
+        pathToRegexp(pathname, pathnameKeys);
+        pathToRegexp(hash || '', hashKeys);
```
styfle, line 192: Do we also need to consider other parts of the URL such as the query string?

packages/now-routing-utils/src/superstatic.ts
ijjk: Updated

styfle (approved):

packages/now-routing-utils/src/superstatic.ts
styfle: I guess thats already accounted for since we wont add duplicate query string keys

packages/now-routing-utils/src/superstatic.ts
styfle: Tested no Node 12 and there is no difference between Array.includes and Set.has when the items are less than 5 🤷‍♂️

packages/now-routing-utils/src/superstatic.ts
ijjk: It might be good to also opt-out of auto-appending if they're used in the destination's query at all 🤔 we could investigate adding this behavior in follow-up PRs

dav-is (approved):

PR #5051, opened 2020-08-17
[cli] Fix update message

styfle:
The update message was sometimes printed when there was no update.

This PR fixes the update check to ensure the that it is only printed when the latest version differs from the current version.

image


styfle:
@TooTallNate What do you think about dropping [`update-notifier`](https://packagephobia.com/result?p=update-notifier) (over 1 MB and 70+ transitive deps) in favor of running this instead in the background:

```sh
npm view vercel dist-tags --json
```

TooTallNate (approved):

PR #5011, opened 2020-08-07
[cli][client] (Major) Remove legacy Now CLI 1.0 code

MaxLeiter:
It was easier to make a new PR than re-base the old ones:

https://github.com/zeit/now/pull/4017
https://github.com/vercel/vercel/pull/4602

Story: https://app.clubhouse.io/vercel/story/425/remove-all-the-zeit-now-1-0-code-from-vercel-cli

PR #5052, opened 2020-08-17
[cli] Remove unused dependencies

styfle:
This PR is a follow up to removing v1 in PR #5011


packages/now-cli/src/index.js
```diff
@@ -13,7 +13,6 @@ try {
     process.exit(1);
   }
 }
-import 'core-js/modules/es7.symbol.async-iterator';
```
styfle, line 16: This was added 2 years ago, but the dependency no longer exists https://github.com/vercel/vercel/commit/64765b393a4be8d1b6e081d0d29283ec7790ef6f#diff-8adafe1e71c9fbe6d42d1ad18ad6e174R6

packages/now-cli/src/util/output/table.js
```diff
@@ -1,8 +1,7 @@
 import chalk from 'chalk';
-import printf from 'printf';
 
 const printLine = (data, sizes) =>
-  data.reduce((line, col, i) => line + printf(`%-${sizes[i]}s`, col), '');
+  data.reduce((line, col, i) => line + col.padEnd(sizes[i]), '');
```
styfle, line 4: Since Node 8, we can use [`padEnd()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd) to right pad a string.

AndyBitz (dismissed):

MaxLeiter (dismissed):

AndyBitz (approved):

packages/now-cli/package.json
```diff
@@ -174,8 +166,6 @@
     "tar-fs": "1.16.3",
     "test-listen": "1.1.0",
     "text-table": "0.2.0",
-    "then-sleep": "1.0.1",
```
TooTallNate, line 177: This was only used in tests?

TooTallNate (approved):

packages/now-cli/package.json
styfle: Yep

PR #5053, opened 2020-08-17
[routing-utils] Attempt normal path-to-regexp replacing before custom

ijjk:
This corrects behavior that was changed when switching to use `safelyCompile` which caused some `path-to-regexp` replacing behavior to be changed like not removing brackets `{}` around params which are removed in Next.js. This updates to first attempt to use the normal replacing behavior from `path-to-regexp` falling back to the safe compiling logic added when that fails. 

Additional tests have been added to ensure the brackets are stripped as expected when compiling with `path-to-regexp`


styfle (commented):
Do we need to add this behavior to `headers` too? 

Maybe this fallback should be added to the `safelyCompile()` function?

https://github.com/vercel/vercel/blob/b3958b3d12c1d6eee32cae90a186bc387381cead/packages/now-routing-utils/src/superstatic.ts#L99-L107

packages/now-routing-utils/src/superstatic.ts
```diff
@@ -201,13 +201,25 @@ function replaceSegments(
...
       );
 
-      pathname = safelyCompile(pathname, indexes);
-      hash = hash ? safelyCompile(hash, indexes) : null;
+      try {
+        // Attempt compiling normally with path-to-regexp first and fall back
+        // to safely compiling to handle edge cases if path-to-regexp compile
+        // fails
+        pathname = compile(pathname, { validate: false })(indexes);
+        hash = hash ? compile(hash, { validate: false })(indexes) : null;
```
styfle: Should the `pathname` and `hash` have separate try/catch instead so you can catch which one failed?

styfle:
Seems like the issue is here: https://github.com/vercel/vercel/blob/b4bd87658e8d7413968592618bd80a3d549eb884/packages/now-routing-utils/src/superstatic.ts#L270

packages/now-routing-utils/src/superstatic.ts
ijjk: Good point, moved the direct compiling attempt into `safelyCompile` behind a param so it can be used for non-header related values. Header values only use `safelyCompile` since they aren't expected to be URLs per-se and can have more edge cases so the directly compiling attempt is skipped for them

styfle (approved):
Looks good, thanks 👍

Timer (approved):
LGTM

PR #5064, opened 2020-08-20
[cli][build-utils] Use `handle: error` for api 404

styfle:
This PR does a few things:

- Consolidates build-utils to use both 404 routes in `handle: error` phase
- Fixes `vc dev` to process `handle: error` phase after the `handle: filesystem` phase
- Updates `/api` 404 route to be excluded for frameworks like RedwoodJS which use their own functions

### DIFF
--- a/packages/now-build-utils/src/detect-builders.ts
+++ b/packages/now-build-utils/src/detect-builders.ts
@@ -971,10 +971,10 @@ function getRouteResult(
       }
 
       rewriteRoutes.push(...dynamicRoutes);
-      rewriteRoutes.push({
-        src: '^/api(/.*)?$',
+      errorRoutes.push({
         status: 404,
-        continue: true,
+        src: '^/api(/.*)?$',
+        dest: '/_vc_no_match_',
       });
     } else {
       defaultRoutes.push(...apiRoutes);
--- a/packages/now-build-utils/src/detect-builders.ts
+++ b/packages/now-build-utils/src/detect-builders.ts
@@ -929,11 +929,10 @@ function getRouteResult(
   const redirectRoutes: Route[] = [];
   const rewriteRoutes: Route[] = [];
   const errorRoutes: Route[] = [];
-  const isNextjs =
-    frontendBuilder &&
-    ((frontendBuilder.use && frontendBuilder.use.startsWith('@vercel/next')) ||
-      (frontendBuilder.config &&
-        frontendBuilder.config.framework === 'nextjs'));
+  const framework = frontendBuilder?.config?.framework || '';
+  const use = frontendBuilder?.use || '';
+  const isNextjs = framework === 'nextjs' || use.startsWith('@vercel/next');
+  const ignoreRuntimes = slugToFramework.get(framework)?.ignoreRuntimes;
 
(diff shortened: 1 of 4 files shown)


packages/now-build-utils/src/detect-builders.ts
```diff
@@ -971,10 +971,10 @@ function getRouteResult(
...
 
       rewriteRoutes.push(...dynamicRoutes);
-      rewriteRoutes.push({
-        src: '^/api(/.*)?$',
+      errorRoutes.push({
         status: 404,
-        continue: true,
+        src: '^/api(/.*)?$',
+        dest: '/_vc_no_match_',
       });
```
styfle: @dav-is What are your thoughts on having two `handle: error` routes?

Can we handled this correctly in production today and use the default 404 page for `/api` and the custom `/404.html` page for all others?

```js
  [
      {
        status: 404,
        src: '^/api(/.*)?$',
        dest: '/_vc_no_match_',
      },
      {
        status: 404,
        src: '^/(?!.*api).*$',
        dest: '/404.html',
      },
  ]
```

dav-is: What is `/_vc_no_match`? You should be able to just leave off the `dest`

dav-is: `^/(?!.*api).*$` also matches `/docs/api` is that intended?

styfle: I thought we had to use `dest` so I used something that wouldn't match: `_vc_no_match`

I'll see if it works without the `dest`.

The two `src` matches here are supposed to be opposites: one matches `/api` and paths below, and the other matches everything besides `/api`. So yes, it is intended.

dav-is: `dest` is only required when changing the path. If it's in handle: error. It must not exist, there's no point in adding another filesystem check

How is `/test/api` and `/api` opposites? It's not in the API directory

dav-is: Oh sorry I meant to say, it **doesn't** match `/docs/api`, so it wouldn't see the custom 404 page

styfle: > Oh sorry I meant to say, it doesn't match /docs/api, so it wouldn't see the custom 404 page

Oh yes that is a problem then. Maybe the solution is to use a wildcard to match anything and the order should make it work correctly?

```js
  [
      {
        status: 404,
        src: '^/api(/.*)?$',
      },
      {
        status: 404,
        src: '^/.*$',
        dest: '/404.html',
      },
  ]
```

*Updated* to fix typo

dav-is: I think that'd work

dav-is: Remember the `.html` is conditional on `cleanUrls`

dav-is: Why is it `^/(.*)*$` and not `^/.*$`?

AndyBitz (dismissed):

AndyBitz (approved):

TooTallNate (approved):

PR #5070, opened 2020-08-21
[static-build] Fix `immutable` cache header for Vue assets

Ignigena:
We have started migrating our projects to use Vue CLI with the Vercel zero-config builders. One particularly bug we ran into was due to the `Cache-Control` headers automatically applied on these resources that [differ from the documentation](https://vercel.com/docs/edge-network/caching#static-files):

> By default we return a `Cache-Control` header containing `public, max-age=0, must-revalidate` to prevent clients (e.g. browsers) from caching the file locally.

Instead of these headers, we were seeing `immutable` on all our resources. This is a good assumption for a vanilla Vue CLI setup because by default a content hash is used on every chunk, even the entry, which will change as soon as the file contents change. However, this may not always be the case, particularly on Vue CLI projects with any modifications to the build process, custom plugins, etc. that may exclude one or more files from the hash.

In our case this was breaking things in dramatic ways every time we deployed (and we deploy quite often!). Since the resources were being sent with the `immutable` header, the client browser was persisting these in cache and subsequent deployments with potentially breaking changes caused errors that were only resolved by clearing the client browser cache.

Example:

```
dist
├── css
│   ├── app.db598b39.css
│   └── common.css
├── js
│   ├── app.ac201ece.js
│   ├── chunk-587c608b.7d4361a1.js
│   ├── chunk-vendors.00676d62.js
│   └── common.js
└── index.html
```

In the above example, `common.css` and `common.js` are not hashed, but still have an `immutable` header preventing them from being re-validated by the client browser when a new deployment modifies their content.

To solve for this, I've updated the RegExp for Vue CLI to only apply the `immutable` caching header on files which have a content hash. This should solve for both vanilla Vue CLI setups and more advanced ones in which there may exist both files with and without a content hash.


styfle (approved):
Perfect, thanks! 🎉

PR #5100, opened 2020-08-26
[cli] Improve startup perf - disable source maps and minification

guybedford:
Currently on my machine, `vercel --version` takes 850ms to run. With this change, that execution time is reduced to ~550ms.

The reason is that source maps seem to significantly slow down Node.js execution, compounded by the size of the source map, even if it is not used.

I believe the benefits of this perf improvement outweigh the direct source maps and minification benefits.

If it is felt that direct per-file source maps are necessary for internal errors, then reducing the source map size should be investigated further to improve the runtime performance here.

### DIFF
--- a/packages/now-cli/scripts/build.ts
+++ b/packages/now-cli/scripts/build.ts
@@ -49,16 +49,7 @@ async function main() {
   // Do the initial `ncc` build
   console.log();
   const src = join(dirRoot, 'src');
-  const args = [
-    '@zeit/ncc',
-    'build',
-    '--source-map',
-    '--external',
-    'update-notifier',
-  ];
-  if (!isDev) {
-    args.push('--minify');
-  }
+  const args = ['@zeit/ncc', 'build', '--external', 'update-notifier'];
   args.push(src);
   await execa('npx', args, { stdio: 'inherit' });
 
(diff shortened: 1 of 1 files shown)


styfle (commented):
Nice find! Let’s use —source-map when isDev is true so that we can attach a debugger.

styfle (approved):
@guybedford This is great! 🎉  I measured roughly 2x speed improvement and roughly 0x change in publish size.

@TooTallNate What do you think?

TooTallNate:
Fine with me. The source maps never were processed properly by Sentry so we're not really losing anything 😄

TooTallNate (approved):

guybedford:
The tests are failing here because the branch is on a third-party repo. @styfle perhaps you can reopen this as a local PR or otherwise further?

styfle:
Created PR #5101 and that looks like its running the tests here too, probably because the commits match 👍

TooTallNate:
> roughly 0x change in publish size

This seems strange right? I would expect the source maps to be large-ish.

styfle:
@TooTallNate This PR disabled minification which increased index.js by 2x, but disabled source maps which were about the same size as index.js was originally so its a net zero impact

styfle:
Its actually a bit smaller 👍 

- 20.0.1-canary.0 [install size](https://packagephobia.com/result?p=vercel@20.0.1-canary.0)
- 20.0.1-canary.1 [install size](https://packagephobia.com/result?p=vercel@20.0.1-canary.1)

TooTallNate:
Thanks @guybedford!

PR #5106, opened 2020-08-26
Fix missing `readyStateAt` error

rdev:
Ensure the API response without `readyStateAt` doesn't cause an error


joecohens:
Thank you @rdev! The fix was implemented on the API side.

PR #5113, opened 2020-08-27
[cli] Add support for the sourceFilesOutsideRootDirectory option

shuding:
If the value is set to `true` in the project settings, we don't cd into `rootDirectory`. 
We default it to `true` for new projects.

### DIFF
--- a/packages/now-cli/src/commands/deploy/latest.js
+++ b/packages/now-cli/src/commands/deploy/latest.js
@@ -276,6 +276,7 @@ export default async function main(
   let { org, project, status } = link;
   let newProjectName = null;
   let rootDirectory = project ? project.rootDirectory : null;
+  let sourceFilesOutsideRootDirectory = true;
 
   if (status === 'not_linked') {
     const shouldStartSetup =
@@ -333,6 +334,7 @@ export default async function main(
     } else {
       project = projectOrNewProjectName;
       rootDirectory = project.rootDirectory;
+      sourceFilesOutsideRootDirectory = project.sourceFilesOutsideRootDirectory;
 
       // we can already link the project
       await linkFolderToProject(
@@ -349,7 +351,13 @@ export default async function main(
     }
   }
 
-  const sourcePath = rootDirectory ? join(path, rootDirectory) : path;
+  // if we have `sourceFilesOutsideRootDirectory` set to `true`, we use the current path
+  // and upload the entire directory.
+  const sourcePath = rootDirectory
+    ? sourceFilesOutsideRootDirectory
+      ? path
+      : join(path, rootDirectory)
+    : path;
 
   if (
     rootDirectory &&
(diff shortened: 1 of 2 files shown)


packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -349,7 +351,13 @@ export default async function main(
...
   }
 
-  const sourcePath = rootDirectory ? join(path, rootDirectory) : path;
+  // if we have `sourceFilesOutsideRootDirectory` set to `true`, we use the current path
+  // and upload the entire directory.
+  const sourcePath = rootDirectory
+    ? sourceFilesOutsideRootDirectory
+      ? path
+      : join(path, rootDirectory)
+    : path;
```
shuding:
```suggestion
  const sourcePath = (rootDirectory && !sourceFilesOutsideRootDirectory) ? join(path, rootDirectory) : path;
```

TooTallNate (changes requested):
Code looks good, but we should add a test for this.

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -333,6 +334,7 @@ export default async function main(
     } else {
```
styfle, line 334: Do you need to set `project.sourceFilesOutsideRootDirectory = true` when the project is created? Or is that handled in the API?

styfle:
@shuding You can add an integration test [here](https://github.com/vercel/vercel/blob/master/packages/now-cli/test/integration.js) that creates a deployment with the CLI and then reads the deployment document to make sure the property was assigned as expected

packages/now-cli/src/commands/deploy/latest.js
styfle: Ignore that last comment

styfle (dismissed):

shuding:
@TooTallNate We found it difficult to test this scenario, because that option doesn't make any difference from the CLI aspect. The only difference is that we are uploading the entire dir, but the deployment doc is 100% the same.

The current `use rootDirectory from project when deploying` test should already cover this case.

TooTallNate (dismissed):

TooTallNate:
Seems the test is failing @shuding.

shuding:
@TooTallNate Yeah, we need https://github.com/vercel/api/pull/5371 to be rolled first to accept this new behavior (discussed with @leo and @styfle).

styfle:
I confirmed the tests pass when `let sourceFilesOutsideRootDirectory = false` so I will ship confirm `true` works with the new build pipeline 👍

packages/now-cli/src/commands/deploy/latest.js
```diff
@@ -523,6 +530,7 @@ export default async function main(
       deployStamp,
       target,
       skipAutoDetectionConfirmation: autoConfirm,
+      projectSettings: { sourceFilesOutsideRootDirectory },
```
styfle: We need this change in `vc link` too https://github.com/vercel/vercel/blob/0fc74feddcf0a5777ead14738171ad2dfd2e7077/packages/now-cli/src/util/link/setup-and-link.ts#L177-L186

styfle (approved):

TooTallNate (approved):

PR #5143, opened 2020-09-03
[node][next] Fix nft tracing for monorepos

styfle:
- Set `base: repoRootPath` so we can resolve files above the `workPath` for monorepos
- Bump nft to 0.9.2 to fix a corner case for `base`


packages/now-node/test/fixtures/04-tsx-resolve/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "strict": true,
     "esModuleInterop": true,
     "sourceMap": true,
+    "jsx": "react",
```
styfle, line 6: This is not necessary for this PR but TS was printing a warning and it confused me so I fixed it 😄 


> index.ts(1,17): error TS6142: Module './dep' was resolved to '/vercel/2df33c57/dep.tsx', but '--jsx' is not set.

TooTallNate, line 6: Why is this necessary?

packages/now-node/src/index.ts
```diff
@@ -74,6 +74,8 @@ function isPortInfo(v: any): v is PortInfo {
   return v && typeof v.port === 'number';
 }
 
+const baseDir = '/';
```
styfle: We could pass in the repository root directory and use that as the base, but technically Node.js will resolve all the way up to the system root as described here:

https://nodejs.org/docs/latest-v12.x/api/modules.html#modules_loading_from_node_modules_folders

packages/now-node/test/fixtures/04-tsx-resolve/tsconfig.json
styfle: https://github.com/vercel/vercel/pull/5143#discussion_r483290387

packages/now-next/src/index.ts
```diff
@@ -226,6 +227,7 @@ export const build = async ({
 
   // Limit for max size each lambda can be, 50 MB if no custom limit
   const lambdaCompressedByteLimit = config.maxLambdaSize || 50 * 1000 * 1000;
+  const baseDir = repoRootPath || '/'; // should fallback to workPath
```
styfle: I'm using `/` to see if any tests fail and it seems like they do so there must be some relative paths that are still wrong. @Timer @ijjk

ijjk (approved):

Timer (approved):

PR #5148, opened 2020-09-04
[cli] Fix color and indention in learn more link

dav-is:
CH-568

Before:
image

After:
Screenshot-2020-09-04-11:44

### DIFF
--- a/packages/now-cli/src/commands/deploy/latest.js
+++ b/packages/now-cli/src/commands/deploy/latest.js
@@ -157,17 +157,22 @@ const printDeploymentStatus = async (
   }
 
   if (indications) {
+    const indent = '    ';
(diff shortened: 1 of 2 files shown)


styfle (commented):
Let's update the CLI test "next unsupported functions config shows warning link" so that it checks for new new format

dav-is:
> Let's update the CLI test "next unsupported functions config shows warning link" so that it checks for new new format

This test doesn't use TTY, so this fix doesn't make any difference in its output

styfle (approved):
Thanks!

TooTallNate (approved):
The indentation part I think is inconsistent with the other times we render "Learn More" links, but code LGTM.

PR #5166, opened 2020-09-08
[cli] Update the default env file for Next.js projects

### LINKED ISSUE #16763: The default env file create by the Vercel CLI should match the Next.js recommended filename
Currently the default filename with `vercel env pull` is `.env` and Next.js recommends `.env.local` and makes sure this `.env` filename variant is already ignored in `.gitignore` in the templates. We should update the default for the CLI to match this so the additional filename isn't needed ([docs ref](https://nextjs.org/docs/basic-features/environment-variables#environment-variables-on-vercel))

ijjk:
Next.js recommends the env file containing sensitive values be named `.env.local` and not committed to the repo to prevent accidentally leaking the values. This updates the `vc env pull` command to default the env file name to `.env.local` to match this recommendation from Next.js. 

Closes: https://github.com/vercel/next.js/issues/16763

### DIFF
--- a/packages/now-cli/src/commands/env/pull.ts
+++ b/packages/now-cli/src/commands/env/pull.ts
@@ -56,7 +56,13 @@ export default async function pull(
     return 1;
   }
 
-  const [filename = '.env'] = args;
+  let [filename] = args;
+
+  if (!filename) {
+    // TODO: do we want to migrate this for non-Next.js projects too?
+    filename = project.framework === 'nextjs' ? '.env.local' : '.env';
+  }
+
   const fullPath = join(process.cwd(), filename);
   const skipConfirmation = opts['--yes'];
 


styfle (commented):
I don't think we can ship this as-is because `vc dev` doesn't read `.env.local`. Did you talk to @leo about this change?

ijjk:
The `.env.local` file should be loaded by Next.js and the default is only being changed for Next.js projects in this PR currently, [internal thread](https://vercel.slack.com/archives/CHZGRCJSD/p1598990458091800?thread_ts=1598990324.088800&cid=CHZGRCJSD)

styfle:
@ijjk This would break for monorepos that use Next.js and another framework in the same repo because the repo might be associated with multiple frameworks. So your dev env variables might not be "only" for Next.js.

Also consider `/api/foo.py` which could use env vars. That would need to be `.env`, not `.env.local`.

*Update*: I think we need to make `vc dev` read `.env.local` in addition to `.env` (the same way that `next dev` does).

PR #5168, opened 2020-09-09
[next] Fix config `functions` src file

styfle:
The [functions](https://vercel.com/docs/configuration#project/functions) property matches source files and it wasn't working for `index.js` or `.ts` files because `next build` converts the source file `pages/api/index.js` to the output `.next/server/pages/api.js`. So this PR fixes the logic to reverse the output page to source file.

### DIFF
--- a/packages/now-next/src/utils.ts
+++ b/packages/now-next/src/utils.ts
@@ -1000,11 +1000,38 @@ async function getSourceFilePathFromPage({
   workPath: string;
   page: string;
 }) {
+  let fsPath = path.join(workPath, 'pages', page);
   if (await usesSrcDirectory(workPath)) {
-    return path.join('src', 'pages', page);
+    fsPath = path.join(workPath, 'src', 'pages', page);
   }
 
-  return path.join('pages', page);
+  if (fs.existsSync(fsPath)) {
+    return path.relative(workPath, fsPath);
+  }
+
+  const extensionless = fsPath.slice(0, -3); // remove "".js"
+  fsPath = extensionless + '.ts';
+  if (fs.existsSync(fsPath)) {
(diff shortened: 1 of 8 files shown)


Timer (dismissed):

ijjk (dismissed):

packages/now-next/src/utils.ts
```diff
@@ -1000,11 +1000,38 @@ async function getSourceFilePathFromPage({
...
-    return path.join('src', 'pages', page);
+    fsPath = path.join(workPath, 'src', 'pages', page);
   }
 
-  return path.join('pages', page);
+  if (fs.existsSync(fsPath)) {
+    return path.relative(workPath, fsPath);
+  }
+
+  const extensionless = fsPath.slice(0, -3); // remove "".js"
```
TooTallNate:
```suggestion
  const extensionless = fsPath.slice(0, -3); // remove ".js"
```

ijjk (approved):

packages/now-next/test/fixtures/06-lambda-with-memory/src/pages/api/index.js
```diff
@@ -0,0 +1,3 @@
+export default function (req, res) {
+  res.end(`${process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE}`);
```
TooTallNate, line 2: You don't really need the backticks here I think, it's always going to be a string (or undefined, but that's ok for `res.end()` as well).

```suggestion
  res.end(process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE);
```

TooTallNate (approved):
Minor nit on the test fixtures, but otherwise LGTM!

PR #5179, opened 2020-09-10
[tests] Cache token when possible

styfle:
This PR solves an issue where tests would sometimes be rate limited when generating a token. This can be solved by caching the token like we do for E2E tests.

Example failure: https://github.com/vercel/vercel/runs/1098030228#step:11:9376

I also updated the `@vercel/node` tests to compare the error message, exactly like how we do it with `@vercel/static-build`.


lucleray (dismissed):
Code looks good 👍 Approving, provided the tests are passing.

TooTallNate (approved):

lucleray (approved):

PR #5220, opened 2020-09-22
[node] Only use the project's local "typescript" when it has compatible `ScriptTarget`

TooTallNate:
https://app.clubhouse.io/vercel/story/8304

### DIFF
--- a/packages/now-node/src/dev-server.ts
+++ b/packages/now-node/src/dev-server.ts
@@ -43,7 +43,7 @@ if (nodeMajor >= 14) {
 let compiler = resolveTypescript(process.cwd());
 if (compiler) {
   ts = requireTypescript(compiler);
-  if (target in ts.ScriptTarget) {
+  if (!(target in ts.ScriptTarget)) {
     ts = null;
   }
 }
--- a/packages/now-node/src/dev-server.ts
+++ b/packages/now-node/src/dev-server.ts
@@ -43,7 +43,7 @@ if (nodeMajor >= 14) {
 let compiler = resolveTypescript(process.cwd());
 if (compiler) {
   ts = requireTypescript(compiler);
-  if (typeof (ts.ScriptTarget as any)[target] === 'undefined') {
+  if (target in ts.ScriptTarget) {
     ts = null;
   }
 }
--- a/packages/now-node/src/dev-server.ts
+++ b/packages/now-node/src/dev-server.ts
@@ -32,7 +32,7 @@ try {
 
 // Otherwise fall back to using the copy that `@vercel/node` uses
 if (!compiler) {
-  compiler = resolveTypescript(__dirname);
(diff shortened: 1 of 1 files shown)


packages/now-node/src/dev-server.ts
```diff
@@ -8,17 +8,31 @@ if (!entrypoint) {
...
+import { join } from 'path';
 import { register } from 'ts-node';
 
-// Use the project's version of TypeScript if available,
-// otherwise fall back to using the copy that `@vercel/node` uses.
-let compiler: string;
-try {
-  compiler = require.resolve('typescript', {
-    paths: [process.cwd()],
+const resolveTypescript = (p: string): string =>
```
styfle: I think it would be better to put the try/catch in this function

packages/now-node/src/dev-server.ts
```diff
@@ -8,17 +8,31 @@ if (!entrypoint) {
...
-} catch (e) {
-  compiler = 'typescript';
+
+// Use the project's version of TypeScript if available
+let compiler = '';
+try {
+  const resolved = resolveTypescript(process.cwd());
+
+  // Only use the user's resolved "typescript" if it's
+  // located *within* the project root directory
```
styfle: What does "project root directory" mean here?
We have `workPath` and `repoRoot` in the `build()` function for monorepos so does this represent `workPath`?

styfle:
Any way we could add a test for this?

packages/now-node/src/dev-server.ts
TooTallNate: Ya, `workPath`.

styfle:
Per our discussion, lets change this to detect the resolved typescript version and fallback to our version shipped in `@vercel/node` if its too old. Probably something like `if (tsVersionMajor < 3) useFallback()`

TooTallNate:
@styfle I took it a step further and implemented _feature checking_ for the TypeScript version to check that it has compatible `"target"` for the Node.js version being used. If it does not, then finally it will fall back to the copy of TypeScript that `@vercel/node` has.

packages/now-node/src/dev-server.ts
```diff
@@ -8,33 +8,54 @@ if (!entrypoint) {
...
 } else if (nodeMajor >= 12) {
-  target = 'es2019';
+  target = 'ES2019';
+}
+
+// Use the project's version of Typescript if available and supports `target`
+let compiler = resolveTypescript(process.cwd());
+if (compiler) {
+  ts = requireTypescript(compiler);
+  if (target in ts.ScriptTarget) {
```
styfle:
```suggestion
  if (!(target in ts.ScriptTarget)) {
```

styfle (approved):
Perfect, thanks!

PR #5223, opened 2020-09-23
chore(examples): update nuxtjs/nuxt.config.js

pi0:
- update `nuxt.config.js` example to be consistent with latest CNA
- add target option

Remarks:
 - With server target, it would be probably better also using vercel builder for nuxt


mcsdevv (dismissed):
Thanks, @pi0!

J-Vin (approved):

PR #5258, opened 2020-10-05
[cli] Remove 1.0 path alias rules

styfle:
We don't support routing path aliases anymore, now that we've completely shut down the ZEIT Now 1.0 infrastructure. This PR removes path alias rules from Vercel CLI.


mglagola (approved):

AndyBitz (approved):

packages/now-cli/test/helpers/prepare.js
```diff
@@ -28,9 +28,6 @@ module.exports = async function prepare(session) {
       'first.png': getImageFile(session, { size: 30 }),
       'second.png': getImageFile(session, { size: 20 }),
     },
-    'single-dotfile': {
```
TooTallNate, line 31: What is this one? Why is it being removed?

TooTallNate (approved):

packages/now-cli/test/helpers/prepare.js
styfle: It was an unused fixture

PR #5298, opened 2020-10-16
[next] Add i18n support handling 

ijjk:
This adds handling for the new i18n support in Next.js. Currently this relies on fixes in https://github.com/vercel/next.js/pull/17926 for it to work properly and the `00-i18n-support` fixture should be updated to use canary after that PR is released. 

TODO: 

- [x] locale specific 404
- [x] enable locale redirects
- [x] additional tests


timneutkens (dismissed):

packages/now-next/package.json
```diff
@@ -29,7 +29,7 @@
     "@vercel/nft": "0.9.2",
     "async-sema": "3.0.1",
     "buffer-crc32": "0.2.13",
-    "escape-string-regexp": "3.0.0",
+    "escape-string-regexp": "2.0.0",
```
styfle, line 32: Why the older version?

ijjk: We need to match the version Next.js is using or else replacing won't work correctly, added a comment [here](https://github.com/vercel/vercel/pull/5298/files#diff-d1fee2ddc0872d679e1cf35b915460f018264a315f528b2cca0a1923b854bf05R22)

packages/now-next/src/index.ts
```diff
@@ -432,6 +439,26 @@ export const build = async ({
...
+        'utf8'
+      );
+      escapedBuildId = escapeStringRegexp(buildId);
+    } catch (err) {
+      console.error(
+        'BUILD_ID not found in ".next". The "package.json" "build" script did not run "next build"'
+      );
+      throw new NowBuildError({
+        code: 'NOW_NEXT_NO_BUILD_ID',
+        message: 'Missing BUILD_ID',
```
styfle:
```suggestion
        message: 'The BUILD_ID file was not found in the Output Directory. Did you forget to run "next build" in your Build Command?',
```

We need to make this actionable because its what the user sees.

I dont think we need the console.error() though.

packages/now-next/src/utils.ts
```diff
@@ -1066,6 +1075,37 @@ function isDirectory(path: string) {
...
+
+  return {
+    pathname,
+    detectedLocale,
+  };
+}
+
+export function addLocale(pathname: string, locale?: string) {
+  return locale ? `/${locale}${pathname}` : pathname;
+}
```
styfle, line 1114: Maybe you could add a parameter to this function for `routesManifest` so you dont need to repeat the same fallback logic in multiple places

ijjk: Updated to de-dupe the mentioned fallback logic

timneutkens (approved):

leo:
Merging this, as it'll be released on `canary` first regardless.

PR #5309, opened 2020-10-20
[next] Ensure root-most index GSP page is located correctly

ijjk:
This fixes root-most index GSP pages not being located correctly, this also uncovered the initial revalidate value not being stored to the `prerender-manifest` correctly causing an incorrect `Prerender` output to be built which is fixed in https://github.com/vercel/next.js/pull/18053

Note: tests won't pass until the above mentioned Next.js PR is landed to canary

### DIFF
--- a/packages/now-next/src/utils.ts
+++ b/packages/now-next/src/utils.ts
@@ -1110,7 +1110,9 @@ export function addLocaleOrDefault(
   if (!routesManifest?.i18n) return pathname;
   if (!locale) locale = routesManifest.i18n.defaultLocale;
 
-  return locale ? `/${locale}${pathname}` : pathname;
+  return locale
+    ? `/${locale}${pathname === '/index' ? '' : pathname}`
+    : pathname;
 }
 
 export {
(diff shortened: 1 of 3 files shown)


packages/now-next/src/utils.ts
```diff
@@ -1110,7 +1110,9 @@ export function addLocaleOrDefault(
   if (!routesManifest?.i18n) return pathname;
   if (!locale) locale = routesManifest.i18n.defaultLocale;
 
-  return locale ? `/${locale}${pathname}` : pathname;
+  return locale
+    ? `/${locale}${pathname === '/index' ? '' : pathname}`
+    : pathname;
```
styfle, line 1115: Does the second `pathname` (the fallback) also need to do the `/index` normalization?

timneutkens (approved):

packages/now-next/src/utils.ts
ijjk: No, the index only needs to be handled this ways with i18n, without i18n the file is located at `index.html`

styfle (approved):

PR #5331, opened 2020-10-26
[@vercel/next] Update default config

AndyBitz:
Updates the default config we generate for next to also consider the `experimental` property.

Story https://app.clubhouse.io/vercel/story/13391


styfle (dismissed):

Timer (changes requested):
If plugins are going to be used by our Platform they should be enabled by default in Next.js—not configured here. This would result in an "scary" console message and also doesn't include the necessary version checks.

AndyBitz:
Closing the PR to continue after the feature is not experimental anymore.

PR #5336, opened 2020-10-27
[next] Ensure index GSP data is available at correct path

ijjk:
This ensures the locale prefixed data route is available at the correct location 


Timer (approved):
lgtm

PR #5345, opened 2020-10-29
[cli] Add support for plaintext env variables

nkzawa:
https://app.clubhouse.io/vercel/story/13352
https://app.clubhouse.io/vercel/story/13743

Currently, we're trying to get decrypted value for plan env so that it fails with error `Unable to download variable ...`.

This PR adds support for plaintext env variables in `vc env pull`, `vc dev` and `vc env ls`.

`vc env ls` output looks like this:
image

The PR also removes `vc env ls --next` because the latest API version always returns all env variables (which has a hard limit of 100).


lucleray (dismissed):

TooTallNate:
Can we add a test?

nkzawa:
@TooTallNate How can I add a test in this case? It looks there is no test for `env pull` now and I don't know what the expected way is for testing something which involves API requests on this repo.

lucleray:
@nkzawa This is the existing test for `now env pull`: https://github.com/vercel/vercel/blob/df5aa1f10d3e9882e104f0ac996f6001403e0a70/packages/now-cli/test/integration.js#L535-L555

There's no way to create plaintext env via the CLI, but maybe you can do an api call to create it, and then check that it's correctly pulled?

nkzawa:
Added a test

lucleray (dismissed):

styfle (dismissed):

lucleray (approved):

packages/now-cli/src/util/env/get-env-records.ts
```diff
@@ -1,69 +1,24 @@
...
-  target?: ProjectEnvTarget,
-  next?: number
 ) {
   output.debug(
     `Fetching Environment Variables of project ${projectId} and target ${target}`
   );
   const query = new URLSearchParams();
-  if (apiVersion >= 5) {
-    query.set('limit', String(20));
-  }
```
styfle, line 50: What happened to `limit`?

styfle (approved):

packages/now-cli/src/util/env/get-env-records.ts
lucleray: v6 doesn't have pagination anymore  (it returns all env variables, which are capped to 100 env variables).

PR #5350, opened 2020-10-29
[next] Ensure revalidation works with i18n

ijjk:
This corrects some i18n pages being considered staticPages instead of prerender items which was breaking revalidate behavior. This also adds more verbose tests for i18n ensuring revalidation is working correctly

x-ref: https://github.com/vercel/next.js/discussions/18443
Fixes: https://github.com/vercel/next.js/issues/18503

PR #5354, opened 2020-10-30
[next] Fix root-most optional-catchall with i18n

### LINKED ISSUE #18503: [10.0.0][i18n][Vercel] a full-page reload occur on page transitions
# Bug report

## Describe the bug

I use the `i18n` feature. Following links using `next/link` does a full page reload. This happens only when an app is deployed to Vercel.

## To Reproduce

Repository: [next.js-bug-i18n-reload](https://github.com/rokinsky/next.js-bug-i18n-reload)

1. Deploy the app to Vercel
2. Go to the website, e.g. [https://next-js-bug-i18n-reload.vercel.app](https://next-js-bug-i18n-reload.vercel.app)
3. Follow the listed links
4. See full page reload (pay attention to favicon if you use Chrome)

## Expected behavior

No page reloading

## Screenshots

N/A

## System information

- OS: macOS
- Browser (if applies): chrome
- Version of Next.js: 10.0.0
- Version of Node.js: v10.22.1

## Additional context

Works great locally in dev and prod modes.

ijjk:
This is a follow-up to https://github.com/vercel/vercel/pull/5350 which ensures that root-most catch-all routes work correctly with i18n. An additional test suite has been added to ensure this works correctly. 

Additional tests have also been added to ensure dynamic SSG pages that were prerendered are revalidated correctly and all data routes return valid JSON

Closes: https://github.com/vercel/next.js/issues/18503

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -511,6 +511,7 @@ export const build = async ({
             const { i18n } = routesManifest;
 
             if (i18n) {
+              const origSrc = route.src;
               route.src = route.src.replace(
                 // we need to double escape the build ID here
                 // to replace it properly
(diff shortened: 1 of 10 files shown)


Timer (dismissed):

packages/now-next/src/index.ts
```diff
@@ -2170,10 +2186,22 @@ export const build = async ({
       { handle: 'filesystem' },
 
       // map pages to their lambda
-      ...pageLambdaRoutes,
+      ...pageLambdaRoutes.filter(route => {
+        // filter out any SSG pages as they are already present in output
+        if ('headers' in route) {
+          let page = route.headers?.['x-nextjs-page']!;
+          page = page === '/index' ? '/' : page;
```
styfle, line 2193: Is it okay to mutate `page` here or do you need a local var?

styfle (dismissed):

packages/now-next/src/index.ts
styfle: Nevermind, I thought page was global but realized it was from the header 🤦

ijjk: It's safe to change here, we're only using it for this check

styfle (approved):

PR #5359, opened 2020-10-31
[cli] Replace `util/prompt-bool` by `util/input/confirm`

lucleray:
One limitation with `util/prompt-bool` is that typing `yes` is considered "false":

This is confusing as a user, it looks like a bug happened to remove the environment variable and it's not clear that it was because I typed `yes`.

`util/input/confirm` correctly considers `yes` to be "true":

---

Side note: the prompts have small visual differences.

`util/prompt-bool`:
image

`util/input/confirm`:
image


styfle (approved):
Very nice, thanks!

PR #5366, opened 2020-11-02
[static-build] add Vercel Analytics Support for Gatsby

Timer:
This pull request adds Vercel Analytics support to the Gatsby framework.


packages/now-static-build/test/builds.test.js
```diff
@@ -0,0 +1,182 @@
...
+
+    expect(contents.some(name => name === 'gatsby-config.js')).toBeTruthy();
+    expect(
+      contents.some(
+        name => name === 'gatsby-config.js.__vercel_builder_backup__.js'
+      )
+    ).toBeTruthy();
+
+    expect(require(path.join(workPath, 'gatsby-config.js')))
+      .toMatchInlineSnapshot(`
```
styfle, line 132: Why toMatchInlineSnapshot instead of deepEqual?

Timer: Because snapshots were easier. 😄

packages/now-static-build/src/utils/gatsby.ts
```diff
@@ -0,0 +1,78 @@
...
+  const pkgJson: DeepWriteable<PackageJson> = (await readPackageJson(
+    dir
+  )) as DeepWriteable<PackageJson>;
+  if (!pkgJson.dependencies) {
+    pkgJson.dependencies = {};
+  }
+  if (!pkgJson.dependencies[gatsbyPluginPackageName]) {
+    pkgJson.dependencies[gatsbyPluginPackageName] = 'latest';
+
+    await writePackageJson(dir, pkgJson);
```
styfle, line 39: Is this going to cause a problem with yarn.lock or package-lock.json?

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -0,0 +1,40 @@
+import { PackageJson } from '@vercel/build-utils';
```
AndyBitz, line 1: Any specific reason for the underscore at the start of the file name? I don't think we have this convention anywhere else in this package.

packages/now-static-build/src/utils/gatsby.ts
Timer: No, we don't run the command with `--frozen-lockfile` or anything by default. It'll re-generate on install.

Timer: Also, this is feature gated strictly for Vercel Analytics users so we won't be mutating every install on our platform (so it's low-risk to ship).

packages/now-static-build/src/utils/gatsby.ts
```diff
@@ -0,0 +1,78 @@
...
+    await fs.rename(
+      gatsbyConfigPath,
+      gatsbyConfigPath + '.__vercel_builder_backup__.js'
+    );
+
+    await fs.writeFile(
+      gatsbyConfigPath,
+      `const userConfig = require("./gatsby-config.js.__vercel_builder_backup__.js");
+
+const vercelConfig = Object.assign({}, userConfig);
```
styfle, line 52: Looks like a shallow clone here but then `plugins` is mutated below.

What is the purpose of the clone?

styfle: Oh I see another clone with `vercelConfig.plugins = vercelConfig.plugins.slice()` so that should work then 👍 

You could also use `...` instead since this runs on Node 12

Timer: Plugins is sliced before we push into it. The shallow clone of both the array and full user object is to prevent mutation.

AndyBitz (approved):

packages/now-static-build/src/utils/_shared.ts
Timer: No specific reason other than its not meant to be imported from outside that folder. I could change it if we want, but I'd prefer not to re-trigger CI for that change.

packages/now-static-build/src/utils/gatsby.ts
Timer: Would be happy to tackle these nits in a follow up to not re-trigger CI 👍

packages/now-static-build/src/index.ts
```diff
@@ -327,6 +328,22 @@ export async function build({
       debug(
         `Detected ${framework.name} framework. Optimizing your deployment...`
       );
+
+      if (process.env.VERCEL_ANALYTICS_ID) {
+        const frameworkDirectory = path.join(
+          workPath,
+          path.dirname(entrypoint)
+        );
```
styfle, line 336: Should we also consider monorepos here where the package.json is at the repo root instead of the workpath?


```js
  const packagejson = await walkParentDirs({
    base: repoRoot,
    start: frameworkDirectory,
    filename: 'package.json',
  });
```

styfle (approved):

packages/now-static-build/src/index.ts
Timer: Good call. I'm not as familiar with this, could you possibly send the PR for this and add a test case?

packages/now-static-build/src/utils/gatsby.ts
styfle: > No, we don't run the command with --frozen-lockfile

@Timer We are going to start allowing the user to define custom install commands soon so this could happen in the future.

Timer: When we do that we might need to actually execute `yarn add` or `npm i` then!

PR #5372, opened 2020-11-04
[cli] (major) Update `vercel env`

lucleray:
Ref: https://app.clubhouse.io/vercel/story/14067

Add support for advanced env variables functionalities to `vercel env`.

Also bump the version of the API used to create env variables to v6.

This PR introduces the following breaking changes in `vercel env` and will require a new major version:
- `vc env add [name] [environment]` is now `vc env add [type] [name] [environment]`
- `cat <file> | vc env add [name] [environment]` is now `cat <file> | vc env add [type] [name] [environment]`
- `vc env add` does not create a secret automatically anymore, to create a secret, you need to run `vc secret add` and then `vc env add`


packages/now-cli/src/commands/env/index.ts
```diff
@@ -18,16 +18,17 @@ import ls from './ls';
 import rm from './rm';
 
 const help = () => {
-  const placeholder = getEnvTargetPlaceholder();
+  const typePlaceholder = getEnvTargetPlaceholder();
```
nkzawa: typo of `getEnvTypePlaceholder()`?

lucleray: Fixed in https://github.com/vercel/vercel/pull/5372/commits/f8180eddc9635def3bd89b4527e54c7d6065660b.

nkzawa (dismissed):

packages/now-cli/src/commands/env/add.ts
```diff
@@ -29,22 +32,25 @@ export default async function add(
   args: string[],
   output: Output
 ) {
+  // improve the way we show inquirer prompts
+  require('../../util/input/patch-inquirer');
```
styfle, line 35: If I remember correctly, this confused users because it removed the comments about `a` for all and `space` for selection in the multi-select.

Is it still necessary?

packages/now-cli/src/commands/env/add.ts
```diff
@@ -29,22 +32,25 @@ export default async function add(
...
     );
     return 1;
   }
 
-  if (stdInput && (!envName || !envTarget)) {
+  if (stdInput && (!envType || !envName || !envTarget)) {
     output.error(
       `Invalid number of arguments. Usage: ${getCommandName(
-        `env add <name> <target> < <file>`
+        `env add ${getEnvTypePlaceholder()} <name> <target> < <file>`
```
styfle, line 53:
```suggestion
        `env add ${getEnvTypePlaceholder()} <type> <name> <target> < <file>`
```

packages/now-cli/src/commands/env/add.ts
```diff
@@ -63,6 +69,33 @@ export default async function add(
...
+
+  if (!envType) {
+    const { inputEnvType } = (await inquirer.prompt({
+      name: 'inputEnvType',
+      type: 'list',
+      message: `Which type of Environment Variable do you want to add?`,
+      choices: [
+        { name: 'Plaintext', value: ProjectEnvType.Plaintext },
+        {
+          name: `Secret (can be created using ${code('vercel secret add')})`,
```
styfle:
```suggestion
          name: `Secret (can be created using ${getCommandName('secret add')})`,
```

packages/now-cli/src/commands/env/add.ts
```diff
@@ -98,15 +131,59 @@ export default async function add(
...
+      try {
+        const secret = await client.fetch<Secret>(
+          `/v2/now/secrets/${encodeURIComponent(secretName)}`
+        );
+
+        secretId = secret.uid;
+      } catch (error) {
+        if (error.status === 404) {
+          output.error(
+            `Please enter the name of an existing Secret (can be created with ${code(
```
styfle:
```suggestion
            `Please enter the name of an existing Secret (can be created with ${getCommandName(
```

packages/now-cli/src/util/output/capitalize.ts
```diff
@@ -0,0 +1,3 @@
+export default function capitalize(str: string) {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
styfle: We already have a dependency that does this called `title`.

See usage here: https://github.com/vercel/vercel/blob/1bfa310945831e9d1204122557bae16d0c6bfe5a/packages/now-cli/src/util/pkg-name.ts#L21-L24

packages/now-cli/src/util/env/system-env.ts
```diff
@@ -0,0 +1,32 @@
+export const SYSTEM_ENV_VALUES = [
```
styfle, line 1: Seems like it would be better to use an API endpoint instead of hard-coding these names so we could add more in the future.

packages/now-cli/src/commands/env/add.ts
```diff
@@ -98,15 +131,59 @@ export default async function add(
...
+              'vercel secret add'
```
styfle:
```suggestion
              'secret add'
```

packages/now-cli/src/commands/env/add.ts
```diff
@@ -98,15 +131,59 @@ export default async function add(
...
     const { inputValue } = await inquirer.prompt({
-      type: 'password',
+      type: 'input',
       name: 'inputValue',
       message: `What’s the value of ${envName}?`,
     });
+
     envValue = inputValue || '';
+  } else if (envType === ProjectEnvType.Secret) {
+    let secretId: string | null = null;
```
styfle, line 145:
```suggestion
    let secret: Secret | null = null;
```

Minor nit pick: I would change this to the secret object itself so there is one less variable needed. Then line 177 can assign uid to the envValue.

packages/now-cli/src/commands/env/add.ts
lucleray: I can add it back and keep the "custom styles".

Just tried it and it looks like this:
image

Alternatively, we can also just remove custom styles for the "checkbox" prompt type, or make other adjustments if needed.

lucleray: The reason I added it is because we also show this prompt:
image

And I want to make it consistent with the other prompts, like this one:
image

packages/now-cli/src/commands/env/add.ts
```diff
@@ -127,7 +204,16 @@ export default async function add(
   const addStamp = stamp();
   try {
     await withSpinner('Saving', () =>
-      addEnvRecord(output, client, project.id, envName, envValue, envTargets)
+      addEnvRecord(
+        output,
+        client,
+        project.id,
+        // @ts-ignore
```
styfle: Why the ts-ignore?

packages/now-cli/src/util/env/add-env-record.ts
```diff
@@ -1,60 +1,34 @@
...
-import { customAlphabet } from 'nanoid';
-import slugify from '@sindresorhus/slugify';
 
 export default async function addEnvRecord(
   output: Output,
   client: Client,
   projectId: string,
-  envName: string,
-  envValue: string | undefined,
+  type: 'plain' | 'secret' | 'system',
```
styfle:
```suggestion
  type: ProjectEnvType,
```

Should this use the Enum isntead?

packages/now-cli/test/integration.js
```diff
@@ -509,13 +539,20 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
         cwd: target,
       }
     );
 
+    await waitForPrompt(
+      now,
+      chunk =>
+        chunk.includes('What’s the value of') && chunk.includes('VERCEL_URL')
+    );
+    now.stdin.write(`\n`); // select VERCEL_URL
```
styfle, line 554: Why do we make the user select an empty value for system env vars? 

Previously, we would skip this step because we knew it was a system env var and the value was populated by the system.

packages/now-cli/src/util/env/add-env-record.ts
```diff
@@ -1,60 +1,34 @@
...
-  if (envValue) {
-    const secrets = await Promise.all(
-      targets.map(target =>
-        client.fetch<Secret>('/v2/now/secrets', {
-          method: 'POST',
-          body: JSON.stringify({
-            name: generateSecretName(envName, target),
-            value: envValue,
-            projectId: projectId,
-            decryptable: target === ProjectEnvTarget.Development,
```
styfle, line 30: Is `decryptable` handled in the secret API now?

packages/now-cli/test/integration.js
lucleray: It shows a list of system env vars and we select the first one: `VERCEL_URL`

packages/now-cli/src/commands/env/ls.ts
```diff
@@ -102,7 +91,9 @@ function getRow(env: ProjectEnvVariableV5) {
   return [
     chalk.bold(env.key),
     value,
-    env.target || '',
+    (Array.isArray(env.target) ? env.target : [env.target || ''])
+      .map(capitalize)
```
styfle: Its a little confusing that these are printed as titlecase in `vc env ls` but the `vc env add` only accepts lowercase.

Perhaps we should change the users input to lowercase or just remove this titlecase to avoid confusion?

packages/now-cli/src/commands/env/add.ts
styfle: I'm a little confused. Which of the 3 screenshots is _with_ the patch and which of the 3 is _without_ the patch?

packages/now-cli/src/commands/env/add.ts
lucleray: Removed in https://github.com/vercel/vercel/pull/5372/commits/3237f0fff7c5b1de0aef5268e5f7e95fdcfe37c7 👍

packages/now-cli/src/util/env/add-env-record.ts
lucleray: Fixed in https://github.com/vercel/vercel/pull/5372/commits/052ae16abbb2cc101bc0d7e113d271bffd012519 👍

packages/now-cli/src/commands/env/add.ts
lucleray: Updated in https://github.com/vercel/vercel/pull/5372/commits/0d191114447a377d4c3782c24a9e51ef095256a5.

packages/now-cli/src/commands/env/add.ts
lucleray: Updated in https://github.com/vercel/vercel/pull/5372/commits/0d191114447a377d4c3782c24a9e51ef095256a5.

packages/now-cli/src/util/output/capitalize.ts
lucleray: TIL! Updated in https://github.com/vercel/vercel/pull/5372/commits/06d0a6f8296b938642e4a5d2a562321708434a8f 👍

packages/now-cli/src/commands/env/add.ts
lucleray: Agreed, I updated it 👍

packages/now-cli/src/util/env/system-env.ts
lucleray: I agree, I created an internal ticket to track it: https://app.clubhouse.io/vercel/story/14475

packages/now-cli/src/commands/env/add.ts
lucleray: The 3 screenshots up there are _with_ the patch.

I pushed a commit to improve the "checkbox" type prompt here: https://github.com/vercel/vercel/pull/5372/commits/d396c15330f4c02c82b327c9c3446b5dac6e8149

Currently, this is how selecting the environment looks:


@styfle Does that lgty?

packages/now-cli/src/commands/env/ls.ts
lucleray: I followed the spec but I agree this is a compelling point to remove the titlecase in the list 👍

packages/now-cli/src/util/env/add-env-record.ts
lucleray: We're not allowing to create decryptable secrets anymore with advanced env variables. If you want to use a env value in development, it should be a plaintext env variable.

packages/now-cli/src/commands/env/ls.ts
styfle: I was thinking of someone who did `vc env ls | grep 'production' | cut -d ' ' | vc env rm` or something where they read `ls` output to send as input into `rm`.

Maybe its okay since the error message seems to use lowercase.

styfle (approved):

styfle:
Also remember to update all the docs/guides/etc that uses `vc env add` 👍

PR #5378, opened 2020-11-06
[cli] Show error when removing env variable from no target environment

lucleray:
Ref: https://app.clubhouse.io/vercel/story/14077

Show an error when removing an env variable but selecting no target environments to remove it from.

The error looks like this:
image

Also updates the UI of the prompts.


styfle (approved):
Nice 👍

PR #5390, opened 2020-11-09
[static-build] Add directory for Serverless Functions

AndyBitz:
#### Story
https://app.clubhouse.io/vercel/story/14652

#### Description
Allows the `@vercel/static-build` runtime to consume the `.vercel_build_output` directory to add Serverless Functions to the output.

It'll grep for all `index.js` files and make a Serverless Function out of them, it'll bundle all the files that are in the same directory as the index.js files in the same function.

At the moment, deploying https://github.com/pi0/nuxt-sls causes the front page to be a 404:
image

With this update SSR will work as expected:
image


packages/now-static-build/build.sh
```diff
@@ -6,3 +6,6 @@ rm -rf dist
 
 # Build with `ncc`
 ncc build src/index.ts -e @vercel/build-utils -e @now/build-utils -o dist
+
+bridge_defs="$(dirname $(pwd))/now-node/dist/bridge.js"
+cp -v "$bridge_defs" dist
```
styfle: Let's use the same pattern as the other runtimes:

```sh
bridge_defs="$(dirname $(pwd))/now-node-bridge/src/bridge.ts"

cp -v "$bridge_defs" src
```

AndyBitz: @styfle that means that we won't be able to use ncc for static-build, since we'd need to transform it to typescript, but not bundle it, since we just want to read it, would that be okay?

packages/now-static-build/src/index.ts
```diff
@@ -530,9 +537,16 @@ export async function build({
     );
     const spawnOpts = getSpawnOptions(meta, nodeVersion);
     await runShellScript(path.join(workPath, entrypoint), [], spawnOpts);
-    validateDistDir(distPath);
+
+    const extraOutputs = await readBuildOutputDirectory({ workPath });
```
styfle: Are you sure we want to enable this feature for legacy deployments?

Seems like it should only be enabled for `zeroConfig: true`.

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -1,6 +1,18 @@
-import { PackageJson } from '@vercel/build-utils';
 import { constants, PathLike, promises as fs } from 'fs';
-import * as path from 'path';
+import { FileBlob } from '@vercel/build-utils';
+import { Files } from '@vercel/build-utils';
+import { Lambda } from '@vercel/build-utils';
+import { PackageJson } from '@vercel/build-utils';
```
styfle: These 4 can be a single `import` statement

AndyBitz: True, I just don't like multiline imports 😛 
but I can change it

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -1,6 +1,18 @@
...
+import buildUtils from '../build-utils';
+import path from 'path';
+
+// @ts-ignore
+import { makeNowLauncher } from '@vercel/node/dist/launcher';
+
+const { createLambda } = buildUtils;
+const { debug } = buildUtils;
+const { getLatestNodeVersion } = buildUtils;
+const { glob } = buildUtils;
```
styfle: These 4 imports can be a single `import` statement

packages/now-static-build/src/index.ts
AndyBitz: I don't think it'll matter, but yea, no need to change the legacy path.

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -1,6 +1,18 @@
...
-import * as path from 'path';
+import { FileBlob } from '@vercel/build-utils';
+import { Files } from '@vercel/build-utils';
+import { Lambda } from '@vercel/build-utils';
+import { PackageJson } from '@vercel/build-utils';
+import buildUtils from '../build-utils';
+import path from 'path';
+
+// @ts-ignore
+import { makeNowLauncher } from '@vercel/node/dist/launcher';
```
styfle: Let's make sure this works with TS by following the pattern used with redwood:

```sh
# Copy shared dependencies
bridge_defs="$(dirname $(pwd))/now-node-bridge/src/bridge.ts"
launcher_defs="$(dirname $(pwd))/now-node/src/launcher.ts"

cp -v "$bridge_defs" src
cp -v "$launcher_defs" src
```

packages/now-static-build/test/fixtures/61-function-output-directory/now.json
```diff
@@ -0,0 +1,22 @@
+{
+  "version": 2,
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@now/static-build",
```
styfle:
```suggestion
      "use": "@vercel/static-build",
```

packages/now-static-build/test/fixtures/62-function-output-directory-with-static/now.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": 2,
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@now/static-build",
```
styfle:
```suggestion
      "use": "@vercel/static-build",
```

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -38,3 +50,75 @@ export async function writePackageJson(
...
+    'functions',
+    'node'
+  );
+  const nodeFunctionFiles = await glob('**/index.js', {
+    cwd: nodeFunctionPath,
+  });
+  const nodeBridgeData = await fs.readFile(path.join(__dirname, 'bridge.js'));
+
+  for (const fileName of Object.keys(nodeFunctionFiles)) {
+    const launcherFileName = '__now_launcher';
```
styfle:
```suggestion
    const launcherFileName = '___now_launcher';
```

nit: The underscores didnt match for these two files

packages/now-static-build/build.sh
styfle: @AndyBitz You need to add the missing `tsc` command like we do for node and next runtimes and it should work.

styfle (dismissed):

packages/now-static-build/src/utils/_shared.ts
```diff
@@ -38,3 +42,75 @@ export async function writePackageJson(
...
+    const launcherFiles: Files = {
+      [`${launcherFileName}.js`]: new FileBlob({
+        data: makeNowLauncher({
+          entrypointPath: `./index.js`,
+          bridgePath: `./${bridgeFileName}`,
+          helpersPath: '',
+          sourcemapSupportPath: '',
+          shouldAddHelpers: false,
+          shouldAddSourcemapSupport: false,
+          awsLambdaHandler: false,
```
styfle: Delete line 82

styfle (dismissed):

styfle (dismissed):

PR #5393, opened 2020-11-10
[static-build][next][node][redwood] Invoke `installCommand` from Project settings

TooTallNate:
Invokes the `installCommand` instead of yarn/npm if defined in the Project's settings for `@vercel/next`, `@vercel/redwood` and `@vercel/static-build` Runtimes.

Also removes the `--prefer-offline` flag by default for all of the above Runtimes, as well as `@vercel/node`.


packages/now-next/src/index.ts
```diff
@@ -376,8 +376,18 @@ export const build = async ({
     await writeNpmRc(entryPath, process.env.NPM_AUTH_TOKEN);
   }
 
-  console.log('Installing dependencies...');
-  await runNpmInstall(entryPath, [], spawnOpts, meta);
+  const installTime = Date.now();
```
styfle: This variable should be in the `else` block right?

packages/now-next/src/index.ts
```diff
@@ -376,8 +376,18 @@ export const build = async ({
...
+  if (typeof installCommand === 'string') {
+    console.log(`Running "install" command: \`${installCommand}\`...`);
+    await execCommand(installCommand, {
+      ...spawnOpts,
+      cwd: entryPath,
+    });
```
styfle: Let's also pass `prettyCommand: installCommand` here so errors look nice.

You'll probably need to change the type of the execCommand() function to `SpawnOptionsExtended`

packages/now-next/src/index.ts
TooTallNate: Ya, updated.

packages/now-node/src/index.ts
```diff
@@ -104,12 +104,7 @@ async function downloadInstallAndBundle({
   } else {
     const installTime = Date.now();
     console.log('Installing dependencies...');
-    await runNpmInstall(
-      entrypointFsDirname,
-      ['--prefer-offline'],
-      spawnOpts,
-      meta
-    );
+    await runNpmInstall(entrypointFsDirname, [], spawnOpts, meta);
```
styfle, line 107: This is missing the `config.installCommand`

packages/now-next/src/index.ts
TooTallNate: Seems that [`execCommand()` already passes `prettyCommand`](https://github.com/vercel/vercel/blob/e2db7c77347fe7698a76df8f31062b9c3b1490b4/packages/now-build-utils/src/fs/run-user-scripts.ts#L110).

packages/now-node/src/index.ts
TooTallNate: `installCommand` is for frontend frameworks, not serverless functions.

packages/now-static-build/src/index.ts
```diff
@@ -120,10 +120,10 @@ function getCommand(
     return null;
   }
 
-  const propName = name === 'build' ? 'buildCommand' : 'devCommand';
+  const propName = `${name}Command`;
```
styfle: I think you can avoid the any below if you do 
```suggestion
  const propName = `${name}Command` as typeof keyof config;
```

styfle (commented):
Looks good but needs tests

packages/now-static-build/src/index.ts
TooTallNate: Nice! That worked for the "config" object, but not for the "framework".

TooTallNate:
Tests added.

styfle (approved):

Timer (approved):
LGTM from Next.js angle

styfle (approved):

ijjk (approved):

TooTallNate:
E2E tests take too long and are flaky, but you can see in [this run](https://github.com/vercel/vercel/runs/1382722098) that the relevant tests (`now-next/test/fixtures/32-custom-install-command`, `now-static-build/test/fixtures/64-custom-install-command` and `redwood/test/fixtures/16-create-redwood-app`) passed. So merging.

PR #5396, opened 2020-11-10
[static-build] Support static and config directory in .vercel_build_output

AndyBitz:
#### Story
https://app.clubhouse.io/vercel/story/14739

#### Description
- Add support for the `static` directory in `.vercel_build_output`. If present, other static output is ignored.
- Add support for `config/routes.json` in `.vercel_build_output`. The routes will get appended to the framework routes.
- Prefix Serverless Functions from `.vercel_build_output` with `.vercel/functions`. Rewrites should be used if the functions need to be available under different paths.
- We only create Serverless Functions if they match `/functions/node/*/index.js`. We don't want to include subdirectories, we instead expect every folder to have one function, and all other files will be bundled with that Function. This prevents creating Functions from subdirectories that were actually just utils.
- Add logging in case anything from `.vercel_build_output` is used.

The tests have been adjusted to check for all new cases.


packages/now-static-build/src/index.ts
```diff
@@ -493,8 +493,34 @@ export async function build({
       const extraOutputs = await readBuildOutputDirectory({ workPath });
 
       // No need to verify the dist dir if there are other output files.
```
styfle: This comment is in the wrong place. It can probably be removed

AndyBitz: 👍 Moved it to the `validateDistDir` function. I'd rather keep it in case someone wonders why we ignore the validation when `functions` is present.

packages/now-static-build/src/index.ts
```diff
@@ -493,8 +493,34 @@ export async function build({
...
-      if (!Object.keys(extraOutputs.output).length) {
-        validateDistDir(distPath);
+      if (extraOutputs.staticFiles) {
+        output = Object.assign(
+          {},
+          extraOutputs.staticFiles,
+          extraOutputs.functions
+        );
+      } else {
```
styfle, line 501: Why is it if/else?

Shouldn't the same code in the `else` run when there are static and functions?

AndyBitz: @styfle no, if `extraOutputs.staticFiles` is present, we'll use the files from there as output instead of checking `public` etc. for files.

For functions it's different, `public` could be used together with the functions from the `.vercel_build_output` directory, since those won't overwrite static assets.

packages/now-static-build/src/utils/read-build-output.ts
```diff
@@ -0,0 +1,164 @@
...
+
+async function readRoutesConfig({
+  workPath,
+}: {
+  workPath: string;
+}): Promise<Route[]> {
+  const routesConfigPath = path.join(
+    workPath,
+    '.vercel_build_output',
+    'config',
```
styfle, line 144: I think this should be `_config` or something that would be less likely for the user to use.

Imagine a user who makes a `config/index.js` function.

styfle (approved):

packages/now-static-build/src/utils/read-build-output.ts
AndyBitz: @styfle this is the directory structure

```
.vercel_build_output
├── config
│   └── routes.json
├── functions
│   └── node
│       └── my-function
│           └── index.js
├── static
│   ├── index.html
│   └── about.html
```

They should either write to `config`, `functions`, or `static`, others will be ignored, so you have to explicitly put a file into functions to become a function.

AndyBitz: It is expected that frameworks (or whatever is doing the building), is creating files for `config` (at the moment only `routes.json`).

PR #5406, opened 2020-11-11
[build-utils] Include `installCommand` when empty string

TooTallNate:
An empty string for the Project's `installCommand` setting has different behavior than `null`, so properly provide the empty string to the Runtime.

### DIFF
--- a/packages/now-build-utils/test/unit.builds-and-routes-detector.test.ts
+++ b/packages/now-build-utils/test/unit.builds-and-routes-detector.test.ts
@@ -20,48 +20,6 @@ describe('Test `detectBuilders`', () => {
     expect(builders![0].src).not.toBe('now.json');
   });
 
-  it('should select "installCommand"', async () => {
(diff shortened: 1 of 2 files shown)


packages/now-build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -20,6 +20,48 @@ describe('Test `detectBuilders`', () => {
     expect(builders![0].src).not.toBe('now.json');
   });
 
+  it('should select "installCommand"', async () => {
```
styfle: Move these tests into the second `describe()` since we're going to remove this one where `featHandleMiss=false`.

All the production deployments use `featHandleMiss=true` so we should only add new tests there.

TooTallNate: Done.

styfle (approved):

PR #5425, opened 2020-11-15
Add support for reading the `public` directory from `.next`

leo:
When support for the `public` directory was added to the Next.js Builder, we forgot to add support for making it read `public` from within `.next` if it's present there (we're already doing that for `static`).

In order to solve [this story](https://app.clubhouse.io/vercel/story/15096/support-reading-the-next-js-public-directory-from-the-next-directory) and make [NX](https://nx.dev/) work great on Vercel, we have to fix this.

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -1928,6 +1928,10 @@ export async function build({
     '**',
     path.join(entryPath, outputDirectory, 'static')
   );
+  const nextPublicFiles = await glob(
+    '**',
+    path.join(entryPath, outputDirectory, 'public')
+  );
   const staticFolderFiles = await glob('**', path.join(entryPath, 'static'));
   const publicFolderFiles = await glob('**', path.join(entryPath, 'public'));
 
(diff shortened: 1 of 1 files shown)


Timer (changes requested):
Next.js does not emit `public` files into the output directory (it doesn't for `static` either).

The top-level `static` directory (akin to `public`) does **not** contain the same contents as the `<outputDirectory>/static` version. These are two completely different sets of files (despite their same name).

This behavior would have to go through an RFC process to be built-into Next.js before we'd do it at the builder level.

---

It's not even clear to me how the `public` files would ever end up in there? Is NX placing them there? Next.js would never respect that location when running `next start`, so I'm not sure this would even be the desired behavior.

leo:
Makes sense.

I thought that was the case after having built [this Git repository](https://github.com/academiares/minimal-example-nx-next) locally (powered by [NX](https://nx.dev/)):

image

After building [this Git repository](https://github.com/vercel/next-site), however, I'm seeing this:

image

So it seems like you're right.

The reason why I didn't try this myself before was that I didn't think [NX](https://nx.dev/) would output the `public` folder twice – because they're already placing it **next to** `.next`:

image

Both of these have the same content.

Would you be fine with just supporting `public` if it's placed next to `.next` then? Personally, I also find it a bit weird that I'm asking Next.js to build my application and then it just gives me some of the assets that are necessary to serve it. On the other hand, I also understand that they're not actually "built" – they would just be copied.

Since it's possible to configure a custom destination for `.next` in Next.js, I think it would make sense to output everything there that is needed to serve the application. I think there are plans to potentially deprecate that option, but that would break tools like NX, right? What's your take?

ijjk:
Closing in favor of https://github.com/vercel/vercel/pull/5526

PR #5428, opened 2020-11-16
[cli] Add tests to compare dev/prd image optimization

styfle:
Add E2E test for to test image optimization against `vc dev` as well as a prod deployment.

#### Tests

#### Code Review

- [x] Issue from task tracker has a link to this PR (CH13104)


juancampa (commented):
@styfle does `vc dev` return `last-modified`? That's something that people have asked about, imgix returns it and we simply forward it

styfle:
@juancampa No it does not. The `next dev` implementation uses `etag` instead (per @Timer) which is why it was omitted from these tests. See https://github.com/vercel/next.js/pull/18986

juancampa:
Got it, that's an inconsistency that we might want to fix later

juancampa (dismissed):

Timer:
@juancampa I think the inconsistency is inherent to the fact that Next.js is optimizing the image itself (thus has the Etag readily at its disposal) verses on Vercel where we proxy Imgix. I don't see the need to be consistent here.

juancampa:
I could see users being confused by etags appearing in their `vc dev` experience, but not in prod. But yeah, for now I think it's okay to accept that there are some (non fundamental) differences between the two environments

juancampa (approved):

Timer (approved):

packages/now-cli/test/dev/fixtures/30-next-image-optimization/vercel.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": 2,
+  "build": {
+    "env": {
+      "FORCE_BUILDER_TAG": "canary"
```
TooTallNate, line 5: Why is this necessary?

TooTallNate (approved):

packages/now-cli/test/dev/fixtures/30-next-image-optimization/vercel.json
styfle: Its not necessary but it will help us catch differences in `@vercel/next` canary before they go to stable

PR #5432, opened 2020-11-17
[build-utils] Remove `continue: true` from 404 route

AndyBitz:
#### Pull Request
Removes the `continue: true` from the 404 API route, so it'll be sorted at the end of the routing phase. Otherwise matching routes that will come afterwards get the 404 status applied, which is not correct if it exists.

#### Story
https://app.clubhouse.io/vercel/story/15143


styfle (commented):
This is only updating build utils but you need to update the CLI as well so it works correctly in vc dev, right?

I suspect we'll see dev tests fail without that.

styfle (approved):

dav-is (approved):

PR #5434, opened 2020-11-17
[cli] Expose system envs when autoExposeSystemEnvs is enabled on vc dev

nkzawa:
Ref: https://app.clubhouse.io/vercel/story/15112

We added a property called `autoExposeSystemEnvs` to projects. If that property is `true`, we automatically expose system env variables such as `VERCEL=1`, `VERCEL_ENV=<production | preview>`, ... to the runtime and build time.

This PR makes sure we mirror this behavior when running `vc dev` locally.

#### Tests

#### Code Review

### DIFF
--- a/packages/now-cli/src/util/dev/builder.ts
+++ b/packages/now-cli/src/util/dev/builder.ts
@@ -66,9 +66,9 @@ async function createBuildProcess(
   let PATH = `${dirname(process.execPath)}${delimiter}${process.env.PATH}`;
 
   const env: Env = {
-    ...systemEnvs.buildEnv,
     ...process.env,
     PATH,
+    ...systemEnvs.buildEnv,
     ...envConfigs.allEnv,
   };
 
--- a/packages/now-cli/src/util/dev/builder.ts
+++ b/packages/now-cli/src/util/dev/builder.ts
@@ -35,6 +35,7 @@ import {
   BuildResultV3,
   BuilderOutputs,
   EnvConfigs,
+  SystemEnvs,
 } from './types';
 import { normalizeRoutes } from '@vercel/routing-utils';
 import getUpdateCommand from '../get-update-command';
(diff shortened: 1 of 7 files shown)


packages/now-cli/src/util/dev/builder.ts
```diff
@@ -64,6 +66,7 @@ async function createBuildProcess(
   let PATH = `${dirname(process.execPath)}${delimiter}${process.env.PATH}`;
 
   const env: Env = {
+    ...systemEnvs.buildEnv,
     ...process.env,
```
styfle, line 69: Do we want the local process.env to overwrite system env vars? Seems like it should be the other way around.

packages/now-cli/src/util/dev/server.ts
```diff
@@ -1997,6 +2011,7 @@ export default class DevServer {
       // Most frameworks use `chalk`/`supports-color` so we enable it anyway.
       FORCE_COLOR: process.stdout.isTTY ? '1' : '0',
       ...(this.frameworkSlug === 'create-react-app' ? { BROWSER: 'none' } : {}),
+      ...this.systemEnvs.runEnv,
       ...process.env,
```
styfle, line 2014: Same question here

packages/now-cli/test/integration.js
```diff
@@ -762,6 +763,71 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
     t.is(homeJson['VERCEL_URL'], localhostNoProtocol);
     t.is(homeJson['MY_STDIN_VAR'], '{"expect":"quotes"}');
 
+    // system env vars are not automatically exposed
+    t.is(apiJson['VERCEL'], undefined);
+    t.is(homeJson['VERCEL'], undefined);
+
+    vc.kill('SIGTERM', { forceKillAfterTimeout: 2000 });
```
styfle, line 770: Why kill here?

packages/now-cli/src/util/dev/builder.ts
lucleray: Agreed, I changed the order 👍

styfle (approved):

TooTallNate (approved):

PR #5477, opened 2020-11-25
[routing-utils] More accurate schema

javivelasco:
This PR completes the routes schema to be more accurate for the provided type.
This will allow us to generate the type dynamically from the JSON Schema


packages/now-routing-utils/src/schemas.ts
```diff
@@ -2,94 +2,126 @@
...
+			  enum: [
+				"error",
+				"filesystem",
+				"hit",
+				"miss",
+				"resource",
+				"rewrite",
+			  ]
+			},
+			src: {
```
dav-is: `handle` should be the only property here

dav-is (commented):
Looks good otherwise. How come tests are failing?

styfle (changes requested):
The tests are failing because a route with the `handle` property cannot have other properties (see the TS types for an example)

packages/now-routing-utils/src/schemas.ts
```diff
@@ -2,94 +2,115 @@
...
+			},
+			status: {
+			  type: 'integer',
+			  minimum: 100,
+			  maximum: 999,
+			},
+			locale: {
+			  type: 'object',
+			  additionalProperties: false,
+			  required: ['redirect', 'cookie'],
```
styfle: @ijjk Are the redirect/cookie props required on the locale?

styfle (commented):
Not sure about locale either

packages/now-routing-utils/src/schemas.ts
ijjk: `cookie` is optional as far as I know but `redirect` is required for the `locale` route

javivelasco: I've updated it so that all properties are optional but it requires at least one to be provided. That way it is matching all of the tests so everything should be fine.

I had to change the type assertions for errors though because the schema changed and the returned errors are not the same. The only "problem" is that error messages would be different maybe not that intuitive but that's a price to pay to have a strict validation. Before this updated the contents of some objects could include anything.

styfle (approved):
Looks good once all the tests pass, thanks 👍

ijjk (approved):

PR #5526, opened 2020-12-07
[next] Add public dir check and update tests

### LINKED ISSUE #19239: Verify Next.js reads  `public` directory from next to the `.next` directory
# Bug report

## Describe the bug

When you run `next build abc/` and then `next start abc/`, we need to verify that the `public` directory (and likely `tsconfig.json`) is read from the `abc/` directory.

This verification extends into the Vercel builder.

### LINKED ISSUE #19599: Nested public folder is ignored by the Next.js builder
# Bug report

## Describe the bug

If there's a nested `public` folder, like `public/public/icon.png`, `icon.png` should be available under `/public/icon.png`. But in the case of Vercel that's currently not the case and it will be under `/icon.png`.

## To Reproduce

Add a file under `public/public` and deploy to Vercel. Locally it works fine with `next dev` and `next start`.

## System information

- Version of Next.js: 10.0.1

ijjk:
This updates out public directory check to fallback to checking the same directory level as the configured `outputDirectory`, fixes public files output names to not un-necessarily remove the `/public` prefix as the files aren't prefixed with this by default, and also updates our i18n revalidate tests to be stabilized by checking for revalidations more than just once and failing. 

x-ref: https://github.com/vercel/vercel/discussions/5512#discussioncomment-147163
Fixes: https://github.com/vercel/next.js/issues/19599
Closes: https://github.com/vercel/next.js/issues/19239


packages/now-next/test/fixtures/00-i18n-support-root-catchall/additional.js
```diff
@@ -121,14 +187,27 @@ module.exports = function (ctx) {
...
+        throw new Error(`Invalid status code ${res.status}`);
+      }
+      const $ = cheerio.load(await res.text());
+      const props = JSON.parse($('#props').text());
+
+      if (isNaN(props.random)) {
+        throw new Error(`Invalid random value ${props.random}`);
+      }
+      return initialRandom !== props.random ? 'success' : 'failed';
+    }, 'success');
```
styfle: Can this be a shared function like `checkRandomChanged()` to avoid copy/pasta?

test/lib/run-build-lambda.js
```diff
@@ -11,13 +12,21 @@ function runAnalyze(wrapper, context) {
 async function runBuildLambda(inputPath) {
   const inputFiles = await glob('**', inputPath);
   const nowJsonRef = inputFiles['vercel.json'] || inputFiles['now.json'];
-  expect(nowJsonRef).toBeDefined();
+
+  if (typeof expect !== 'undefined') {
```
styfle, line 16: Is `expect` undefined in some cases? 🤔

ijjk: Added this to allow requiring this util outside of tests

Timer (dismissed):
Seems fine as it's non-breaking for the condition of 2x `public/` folders. 

The sudden inclusion of the `public/` folder is prob a welcomed bugfix, even tho "breaking" in some edge cases.

I'd really like to see this done via a manifest emitted from Next.js tho—however, we can do that in the future. This is because I'm worried that we're diverging from how Next.js itself works, meaning we have a discrepancy that may start biting users.

TooTallNate (dismissed):

packages/now-next/test/fixtures/00-i18n-support-root-catchall/additional.js
ijjk: Added a function to de-dupe this a bit

styfle (approved):

TooTallNate (approved):

Timer (approved):

PR #3851, opened 2020-02-27
[now-python] Fix wrong unquote in `payload['path']` error

chentschel:
Wrong unquote in payload['path'], makes urls with url encoded space, to throw error. Fixing and proposing this PR As stated in https://spectrum.chat/zeit/now/double-url-encoding~1f95aed5-b226-4976-8681-47e5608be8d7?m=MTU3MTc0ODk5OTUyOA==

Issue #525, opened 2017-05-09
What are those things I keep hearing about FROZEN states?


vladejs:
What's the default policy to FREEZE a deployment? It doesn't make any sense to me to be paying for a service that shutdown my apps randomly, especially in my case, because I use `Meteor`, which apps  are known to take some time to boot up.

I linked a `ping` service to my `now` apps and I'm constantly receiving emails about my apps going down and up.

How I'm suppose to be taken seriously by my users if my apps are almost always failing?

Do `now` service isn't suitable for Meteor apps? I have the understanding that a Meteor app is just a NodeJS app, so what's the matter with this?

The only reason I see this service useful is for apps like Facebook, which are constantly been visited.
In my case, I have apps which are used 3 times per week, Monday, Thursday and Saturday, for example.

am I wrong?

jamo:
The apps shouldn't be put to sleep randomly, rather, after a certain period of inactivity:
- https://zeit.co/docs/deployment-types/docker#builds
- https://zeit.co/docs/deployment-types/node#deployment-inactivity

You may also control it with `now scale`; scaling an deployment to 1 instance will disable the sleep. Keep in mind, however, that there is a limit of how many instances can be running simultaneously.

jamo:
Also, as long as the deployment works with the convention of supporting `npm install` `npm run build` and `npm run start` it works as an npm deployment. 
See https://zeit.co/now#how-do-i-run-my-build-process and https://zeit.co/docs/deployment-types/node

vladejs:
Thanks for the heads up @jamo . One question arises though.

Which Plans are allowed to use `now scale` command? 

Currently I am using Premium plan.

Another question. Is now service suitable for production Meteor apps? Is it better than Galaxy?

jamo:
`now scale` to a constant number is allowed for all users, regardless of the plan. While remaining within the instance limits.

Also `now scale <url> 0 1` is also allowed for users regardless of the plan as that's the default setting and isn't really "auto scaling" as intended in the new pricing.

I don't have experience with Meteor, so I cannot really say.
However, I can tell that we have lots of meteor apps deployed with Dockerfiles with both  https://github.com/jkrup/meteor-now and with custom dockerfiles.
Due to its diversion from the 'normal npm workflow" it's not currently natively supported as our NPM deployments.

vladejs:
[...]

PR #5598, opened 2020-12-23
[python] Fix error for BaseHTTPRequestHandler dealing url beyond ascii after unquoting

### LINKED ISSUE #5580: Python bug: error occurs when url including special charactors visited
The server cannot deal with url with special charactors
when the serverless function is written in python using handler that inherits from the BaseHTTPRequestHandler class.

For example, 
https://test.nicelee.vercel.app/api/date goes right, while
https://test.nicelee.vercel.app/api/date?test=%E5%93%88%E5%93%88 goes wrong, or
https://test.nicelee.vercel.app/api/date?test=哈哈 goes wrong

The serverless function follows here.
```
from http.server import BaseHTTPRequestHandler, HTTPServer
from datetime import datetime

count = 0

class handler(BaseHTTPRequestHandler):

  def do_GET(self):
    global count
    count = count + 1
    self.send_response(200)
    self.send_header('Content-type', 'text/plain')
    self.end_headers()
    self.wfile.write(str(datetime.now().strftime('%Y-%m-%d %H:%M:%S\n')).encode())
    self.wfile.write(str(count).encode())
    return
```

nICEnnnnnnnLee:
[python] Fix error for BaseHTTPRequestHandler dealing url beyond ascii after unquoting (#5580)

### Related Issues

> Fixes #5580

#### Tests

#### Code Review

### DIFF
--- a/packages/now-python/now_init.py
+++ b/packages/now-python/now_init.py
@@ -46,8 +46,7 @@ if 'handler' in __now_variables or 'Handler' in __now_variables:
         _thread.start_new_thread(server.handle_request, ())
 
         payload = json.loads(event['body'])
-        path = unquote(payload['path'])
-        path = path.replace(' ', '%20')
+        path = payload['path']
         headers = payload['headers']
         method = payload['method']
         encoding = payload.get('encoding')


styfle (changes requested):
This breaks the existing test fixture `10-url-params-http-handler`.

Also, we would need a new test fixture to verify it fixes the broken case.

nICEnnnnnnnLee:
> 
> 
> This breaks the existing test fixture `10-url-params-http-handler`.
> 
> Also, we would need a new test fixture to verify it fixes the broken case.

Well, I have to admit [#525](https://github.com/vercel/now-builders/pull/525) and #3851 is reasonalbe and this PR is not quite a perfect solution.  
But absoulutely there's a bug here. 
It will be somehow much more complicated if not dealt this way. In this perspective, I think test like [follows](https://github.com/vercel/vercel/pull/3851/files/c7f3ceac18cb89032dbec85792b5d9a578e1fa6c#diff-e1d641e3ad9ae23197604d00c4380e319f21839c9a60fd46866263a0bdb8eada) can be accepted.
```
    { "path": "/?hello=/", "mustContain": "path=/?hello=%2F" },
    { "path": "/测试", "mustContain": "path=/%E6%B5%8B%E8%AF%95" },
    { "path": "/another?hello=/", "mustContain": "path=/another?hello=%2F" }
```

styfle:
Please take a look at PR #5739 to see if this solves your use case, thanks!

PR #5614, opened 2020-12-29
[next] share cache between `nft` runs

Timer:
This should result in faster Lambda tracing, as it avoids duplicate work.

This is a performance optimization that doesn't need additional tests. The existing tests should catch if this introduces a bug.

---

### Related Issues

N/A

#### Tests

#### Code Review


styfle (approved):
Sounds good as long as the file system is the same between the two traces 👍

ijjk (approved):

PR #5618, opened 2020-12-31
[next] Ensure 404 is static with _app.GIP and 404 GSP

### LINKED ISSUE #19849: SSG doesn't work for 404 error pages in Vercel environment when using getInitialProps on _app
# Bug report
SSG doesn't work for 404 error pages in Vercel environment when using getInitialProps on _app

## Describe the bug
Since there is still no support for [getStaticProps on _app](https://github.com/vercel/next.js/discussions/10949), I'm forced to keep using getInitialProps on _app to fetch common data for shared components (header, footer, seo, ...). Up until this point, I had no problems with this approach since I'm using getStaticProps to fetch data in all of my pages, and this **does not** opt-out of SSG. So only downside is that my website cannot use automatic static optimization.

However, today I noticed that when I go to page that does not exists and I get redirected to a 404 error page, I see changes in header that were made after the build. What is even more strange is that even when I clear all of my cache, on all subsequent requests to this page, changes in header are no longer reflected. **But** they still are reflected on 404 pages that i didn't access yet (but also only on the first access). The way it works closely resembles how Incremental Static Generation works.

It is important to note that I'm returning an empty getStaticProps in this page also, since I want the 404 error page to be also able to utilize SSG (since I'm using getInitialProps on _app and this is not automatic). (See screenshots below)

When seeing the build log, I get a full dot next to the 404 page meaning the page is statically pre-generated.

When running `npm run build` and `npm run start` on localhost, this issue does not occur. Not even when exporting to serverless environment (target: 'serverless' in next.config). But once I push the project to Vercel, the problem arises. From what I have discovered during my research, this bug is exclusive to Vercel environment.

## To Reproduce

1. Run getInitialProps on _app
2. Create a custom 404 error page with empty getStaticProps
3. Push to Vercel
4. Go to a page that does not exist

## Expected behavior

I expect the page to be statically generated since that is what I'm seeing from the build output.

## Screenshots
Custom 404:
Screenshot 2020-12-04 at 19 41 13
Build output:
Screenshot 2020-12-04 at 19 46 03

## System information

- OS: macOS 11.0.1
- Browser: Chrome 87.0.4280.88
- Version of Next.js: 10.0.3
- Version of Node.js: 12.14.1
- Deployment: Vercel

## Additional context

During my research I also discovered that getStaticProps is not passing data to custom 404 component. I don't know if that's intentional. I didn't find it mentioned anywhere in the documentation though.

ijjk:
This ensures when generating a static 404 using `getStaticProps` in `/404.js` that the resulting file isn't a `Prerender` output unless `revalidate` is used otherwise the `/404.js` `getStaticProps` will be unexpectedly called during runtime for each 404 path. 
 
### Related Issues

Fixes: https://github.com/vercel/next.js/issues/19849

#### Tests

#### Code Review

### DIFF
--- a/packages/now-next/src/index.ts
+++ b/packages/now-next/src/index.ts
@@ -1831,7 +1831,11 @@ export async function build({
           });
         }
 
-        if (!canUsePreviewMode) {
+        // If revalidate isn't enabled we force the /404 route to be static
+        // to match next start behavior otherwise getStaticProps would be
+        // recalled for each 404 URL path since Prerender is cached based
+        // on the URL path
+        if (!canUsePreviewMode || (hasPages404 && routeKey === '/404')) {
           htmlFsRef.contentType = htmlContentType;
           prerenders[outputPathPage] = htmlFsRef;
           prerenders[outputPathData] = jsonFsRef;
--- a/packages/now-next/test/integration/gip-gsp-404/next.config.js
+++ b/packages/now-next/test/integration/gip-gsp-404/next.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  generateBuildId() {
+    return 'testing-build-id';
+  },
+};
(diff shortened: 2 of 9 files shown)


Timer (approved):
LGTM if tests pass

styfle (approved):

PR #5621, opened 2021-01-02
[node] Correction of missing types

zce:
`query` & `cookie` may be undefined.

### Related Issues

none

#### Tests

#### Code Review


TooTallNate:
@zce Thank you for the PR. However, I'm not sure that this is the case. From [looking at the code](https://github.com/vercel/vercel/blob/971481ba51b60112e2c8b4d9764df6ab9954d9a4/packages/now-node/src/helpers.ts#L49-L69), it seems to me that those properties should always be at least an empty object.

Can you clarify in which situation `query` and/or `cookie` would be undefined?

PR #5634, opened 2021-01-04
[node-bridge] Move `launcher.ts` to `@vercel/node-bridge`

TooTallNate:
The `launcher.ts` file is more closely aligned to the node bridge package, as it's used by other runtimes as well.

### DIFF
--- a/packages/now-node/build.sh
+++ b/packages/now-node/build.sh
@@ -1,29 +1,33 @@
 #!/bin/bash
 set -euo pipefail
 
-bridge_defs="$(dirname $(pwd))/now-node-bridge/src/bridge.ts"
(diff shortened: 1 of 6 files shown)


packages/now-node/build.sh
```diff
@@ -16,7 +16,6 @@ ln -sf symlinked-asset test/fixtures/11-symlinks/symlink
 
 # use types.d.ts as the main types export
 mv dist/types.d.ts dist/types
-rm dist/*.d.ts
```
styfle, line 19: I think this was intentional, we only want the `dist/types.d.ts` to be public.

styfle (approved):

PR #5641, opened 2021-01-05
Stop publishing legacy `@now` scoped packages

TooTallNate:
These `@now` scoped packages have been published and deprecated with the following warning since May 2020:

```
DEPRECATED ⚠️  - "@now/node" is deprecated and will stop receiving updates on December 31, 2020. Please use "@vercel/node" instead.
```

Since it is now 2021, the time has come to stop publishing.

### Related Issues

[[ch3378]]


styfle (approved):
▲

utils/publish.sh
```diff
@@ -5,24 +5,16 @@ set -euo pipefail
...
-echo "__dirname: ${__dirname}"
-
 if [ -z "$NPM_TOKEN" ]; then
   echo "NPM_TOKEN not found. Did you forget to assign the GitHub Action secret?"
   exit 1
 fi
 
 echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
 
-if [ ! -e ~/.npmrc ]; then
```
TooTallNate, line 18: Removed this block because there's no possibility that `~/.npmrc` does not exist, since it's being written to in the line above.

utils/publish.sh
```diff
@@ -32,19 +24,12 @@ fi
 
 if [[ "$tag" =~ -canary ]]; then
   echo "Publishing canary release"
-  npm_tag="--npm-tag canary"
+  dist_tag="--dist-tag canary"
```
TooTallNate, line 27: The lerna flag was renamed to `--dist-tag`, so this fixes a deprecation warning.

Timer (approved):

PR #5684, opened 2021-01-12
[cli] Formatting fixes for `startDevServer()` and npm error messages

TooTallNate:
* Adds error link when `startDevServer()` throws ENOENT
 * Make `npm` not installed error formatting consistent with `startDevServer()` version
 * Fixes "double error message printing" for non-ENOENT `startDevServer()` errors, as seen in the screenshots below

### Before

Clipboard 2021-10-01 at 4 12 43 PM

### After

Screen Shot 2021-01-11 at 4 21 10 PM

### DIFF
--- a/packages/now-cli/src/util/dev/server.ts
+++ b/packages/now-cli/src/util/dev/server.ts
@@ -887,7 +887,7 @@ export default class DevServer {
         })
         .catch(err => {
           this.updateBuildersPromise = null;
-          this.output.error(`Failed to update builders: ${err.message}`);
+          this.output.prettyError(err);
           this.output.debug(err.stack);
         });
     }, ms('30s'));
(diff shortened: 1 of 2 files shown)


AndyBitz (approved):

packages/now-cli/src/util/dev/server.ts
```diff
@@ -1683,17 +1683,15 @@ export default class DevServer {
...
-            chalk.red
-          );
-          log(`Please ensure that ${cmd(err.path)} is properly installed`);
-        } else {
-          error(`Failed to start "${builderPkg.name}" dev server: ${err}`);
+          err.message = `Command not found: ${chalk.cyan(
+            err.path,
+            ...err.spawnargs
+          )}\nPlease ensure that ${cmd(err.path)} is properly installed`;
+          err.link = 'https://vercel.link/command-not-found';
```
styfle, line 1690: This URL is 404

TooTallNate: Live now.

PR #5687, opened 2021-01-12
[cli] Use `cmd()` formatting for "npm not installed" message

TooTallNate:
For consistency with the `startDevServer()` ENOENT error message.

### DIFF
--- a/packages/now-cli/src/util/dev/builder-cache.ts
+++ b/packages/now-cli/src/util/dev/builder-cache.ts
@@ -264,7 +264,7 @@ async function npmInstall(
           (result as any).code === 'ENOENT'
             ? `Command not found: ${chalk.cyan(
                 'npm'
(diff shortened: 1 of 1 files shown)


AndyBitz (dismissed):

packages/now-cli/src/util/dev/builder-cache.ts
```diff
@@ -264,7 +264,7 @@ async function npmInstall(
           (result as any).code === 'ENOENT'
             ? `Command not found: ${chalk.cyan(
```
styfle, line 266:
```suggestion
            ? `Command not found: ${cmd(
```

TooTallNate: Negative, see [here](https://github.com/vercel/vercel/blob/master/packages/now-cli/src/util/dev/server.ts#L1686-L1689).

styfle (approved):

PR #5690, opened 2021-01-13
[cli] Respect the `directoryListing` Project setting in `vc dev`

TooTallNate:
Implements the ["Directory Listing" Project setting](https://vercel.com/changelog/listing-the-content-of-directories-can-now-be-toggled) in `vercel dev`.

The Dev integration test file has been updated to allow specifying custom Project settings that will be `PATCH`'d to the project document before running `vc deploy` / `vc dev`.

### DIFF
--- a/packages/now-cli/test/dev/integration.js
+++ b/packages/now-cli/test/dev/integration.js
@@ -232,7 +232,7 @@ async function testFixture(directory, opts = {}, args = []) {
 function testFixtureStdio(
   directory,
   fn,
-  { expectedCode = 0, skipDeploy, isExample } = {}
+  { expectedCode = 0, skipDeploy, isExample, projectSettings } = {}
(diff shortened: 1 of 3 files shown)


packages/now-cli/test/dev/integration.js
```diff
@@ -622,14 +649,6 @@ test(
     });
   })
 );
-/*
```
TooTallNate, line 625: I'm not sure why this was commented out, but removed since it's dead code.

lucleray (approved):
Thanks!

packages/now-cli/test/dev/integration.js
styfle: It might be a test that is inconsistent between production and development

styfle (approved):

PR #5697, opened 2021-01-14
[cli] Implement generic announcement system

anatrajkovska:
This PR improves the CLI in a way that if the API from any request returns a response which contains a header with the prefix `x-vercel-warning-*`, `x-vercel-notice-*` or `x-vercel-tip-*`, it will print out the message to the output to let the user know.

This new feature also supports additional headers which improve the message, so that a link can be printed out for more information. Those are `x-vercel-link-*`, where we specify a link, and `x-vercel-action`, where we specify the text before the link.

Here's an example for a warning message which warns the user if they add a DNS record for a domain which does not use Vercel Nameservers:

Screenshot 2021-01-18 at 17 49 14

The API request from above which creates a DNS record has the following HTTP response headers set:

```
x-vercel-warning-create-domain-record: Your Domain isn't using Vercel nameservers. In order for the changes to your DNS Records to get applied, however, it needs to.
x-vercel-link-create-domain-record: https://vercel.link/configure-vercel-nameservers
```

If `x-vercel-action` is not specified, it defaults to `Learn More`.

#### Tests

#### Code Review


packages/now-cli/src/util/print-indications.ts
```diff
@@ -0,0 +1,30 @@
...
+  const regex = /^x-(?:vercel|now)-(warning|notice|tip)-(.*)$/;
+
+  for (const [name, payload] of res.headers) {
+    const match = name.match(regex);
+    if (match) {
+      const [, type, identifier] = match;
+      const action = res.headers.get(`x-vercel-action-${identifier}`);
+      const link = res.headers.get(`x-vercel-link-${identifier}`);
+      if (indications.has(type)) {
+        const indent = process.stdout.isTTY ? '    ' : ''; // if using emojis
```
TooTallNate: `output.print()` outputs to stderr, so we should check `stderr.isTTY`.

```suggestion
        const indent = process.stderr.isTTY ? '    ' : ''; // if using emojis
```

packages/now-cli/src/util/print-indications.ts
```diff
@@ -0,0 +1,30 @@
...
+        const newline = '\n';
+        const message =
+          prependEmoji(chalk.dim(payload), emoji(type as EmojiLabel)) + newline;
+        let finalLink = '';
+        if (link)
+          finalLink =
+            indent + chalk.dim(`${action || 'Learn More'}: ${link}`) + newline;
```
TooTallNate: The indent on the link doesn't look great IMO. I'd prefer aligned to the left. Also, I'd suggest using `src/util/output/link` to style the link.

TooTallNate (dismissed):
Would be ideal to add a test for this.

TooTallNate (approved):

packages/now-cli/src/util/print-indications.ts
```diff
@@ -0,0 +1,32 @@
+import chalk from 'chalk';
+import { Response } from 'node-fetch';
+import { emoji, EmojiLabel, prependEmoji } from './emoji';
+import createOutput from './output';
+import linkStyle from './output/link';
+
+export default function printIndications(res: Response) {
+  const _output = createOutput();
+  const indications = new Set(['warning', 'notice', 'tip']);
+  const regex = /^x-(?:vercel|now)-(warning|notice|tip)-(.*)$/;
```
styfle, line 10: Do we want to keep supporting `now` here? I think we could only check for `x-vercel` here.

styfle (approved):

packages/now-cli/src/util/print-indications.ts
anatrajkovska: I left it in case we still have any API endpoints which return `x-now`.

PR #5701, opened 2021-01-16
[cli] Fix `cleanUrls: true` when using an output directory

TooTallNate:
This is essentially a refactor of the `@vercel/static` builder that simplifies the code a lot and fixes a bug where `cleanUrls: true` was not working correctly when using an `outputDirectory`.

### DIFF
--- a/packages/now-cli/src/util/dev/builder.ts
+++ b/packages/now-cli/src/util/dev/builder.ts
@@ -280,7 +280,6 @@ export async function executeBuild(
       path = extensionless;
     }
 
-    delete output[path];
     output[path] = value;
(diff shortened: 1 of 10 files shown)


packages/now-cli/src/util/dev/builder.ts
```diff
@@ -280,7 +280,6 @@ export async function executeBuild(
       path = extensionless;
     }
 
-    delete output[path];
```
TooTallNate, line 283: This line was not doing anything since it's being re-assigned on the line below.

styfle (approved):

PR #5717, opened 2021-01-19
[cli] Make `Output` class a singleton and have reusable `Spinner`

TooTallNate:
This is a refactor of the `ora` Spinner usage to make the preferred usage be via the `output.spinner()` and `output.stopSpinner()` functions.

* The `Output` instance has a local spinner instance that can be updated via multiple calls to `output.spinner()`.
* `output.print()` and friends call `output.stopSpinner()` implicitly, so `output.stopSpinner()` doesn't need to be called unless it is desired to immediately remove the spinner before rendering the next text.
* Because there's meant to be a shared `spinner` for the `output` instance, it means that the same `output` instance needs to be used throughout the CLI lifecycle. Therefore, the `createOutput()` function now returns a singleton so that it's guaranteed to be the same instance.

https://user-images.githubusercontent.com/71256/105009979-0670a200-59f0-11eb-9853-232f92eae536.mov

https://user-images.githubusercontent.com/71256/105010034-17211800-59f0-11eb-918d-b15f457e9640.mov

In the "before" video above, note how there's a brief second where the spinner is cleared, but the deployment URL has not yet been printed. In the "after" video, this is fixed and the spinner is only cleared once the CLI is ready to render the deployment URL.

### DIFF
--- a/packages/now-cli/src/util/deploy/process-deployment.ts
+++ b/packages/now-cli/src/util/deploy/process-deployment.ts
@@ -6,7 +6,7 @@ import {
   DeploymentOptions,
   NowClientOptions,
 } from '@vercel/client';
-import { Output, StopSpinner } from '../output';
+import { Output } from '../output';
 // @ts-ignore
 import Now from '../../util';
 import { NowConfig } from '../dev/types';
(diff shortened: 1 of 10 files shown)


packages/now-cli/src/util/deploy/process-deployment.ts
```diff
@@ -100,11 +100,7 @@ export default async function processDeployment({
     skipAutoDetectionConfirmation,
   };
 
-  let queuedSpinner = null;
```
TooTallNate, line 103: The spinner logic in this file gets simplified drastically, due to the new `output.spinner()` behavior and `output.stopSpinner()` function.

mglagola (approved):

styfle (approved):

PR #5718, opened 2021-01-19
[cli] Show login prompt for `vercel dev`

TooTallNate:
`vercel dev` requires the user to be logged in ever since the project needs to be queried from the database.

**Before:**

Screen Shot 2021-01-18 at 7 58 56 PM

**After:**

Screen Shot 2021-01-18 at 7 59 01 PM


lucleray (dismissed):

styfle (dismissed):
Nice catch!

PR #5722, opened 2021-01-19
[cli] Re-use a single `Output` instance throughout the codebase

TooTallNate:
Previously every command implementation was invoking `createOutput()` individually, causing multiple instances of `Output` to be created, which was unnecessary, and causes issues with the shared `spinner` instance introduced by #5717.

Now, the `Output` instance created in `src/index.js` is passed down to the sub-command being executed and therefore re-used.

### DIFF
--- a/packages/now-cli/src/commands/inspect.js
+++ b/packages/now-cli/src/commands/inspect.js
@@ -58,7 +58,12 @@ export default async function main(ctx) {
     return 2;
   }
 
-  const { apiUrl, output } = ctx;
+  const {
+    apiUrl,
+    output,
+    authConfig: { token },
+    config,
+  } = ctx;
   const debugEnabled = argv['--debug'];
   const { print, log, error } = output;
 
@@ -71,16 +76,13 @@ export default async function main(ctx) {
     return 1;
   }
 
-  const {
-    authConfig: { token },
-    config,
-  } = ctx;
   const { currentTeam } = config;
   const client = new Client({
     apiUrl,
     token,
     currentTeam,
     debug: debugEnabled,
+    output,
   });
   let contextName = null;
 
@@ -95,7 +97,13 @@ export default async function main(ctx) {
     throw err;
   }
 
-  const now = new Now({ apiUrl, token, debug: debugEnabled, currentTeam });
+  const now = new Now({
+    apiUrl,
+    token,
+    debug: debugEnabled,
+    currentTeam,
+    output,
+  });
 
   // resolve the deployment, since we might have been given an alias
   const depFetchStart = Date.now();
--- a/packages/now-cli/src/commands/inspect.js
+++ b/packages/now-cli/src/commands/inspect.js
@@ -3,7 +3,6 @@ import getArgs from '../util/get-args';
 import buildsList from '../util/output/builds';
 import routesList from '../util/output/routes';
 import indent from '../util/output/indent';
-import createOutput from '../util/output';
 import Now from '../util';
 import logo from '../util/output/logo';
 import elapsed from '../util/output/elapsed.ts';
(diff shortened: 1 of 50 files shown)


packages/now-cli/src/commands/inspect.js
```diff
@@ -101,14 +99,13 @@ export default async function main(ctx) {
...
   // resolve the deployment, since we might have been given an alias
   const depFetchStart = Date.now();
-  const cancelWait = output.spinner(
+  output.spinner(
     `Fetching deployment "${deploymentIdOrHost}" in ${chalk.bold(contextName)}`
   );
 
   try {
     deployment = await now.findDeployment(deploymentIdOrHost);
   } catch (err) {
```
styfle, line 116: Do you need to stop this spinner here?

packages/now-cli/src/commands/list.js
```diff
@@ -78,11 +77,15 @@ export default async function main(ctx) {
...
+    authConfig: { token },
+    output,
+    apiUrl,
+    config,
+  } = ctx;
 
-  const { print, log, error, note, debug } = createOutput({
-    debug: debugEnabled,
-  });
+  const debugEnabled = argv['--debug'];
```
styfle, line 86: I think you can remove `debugEnabled` since its the same as `debug` below.

packages/now-cli/src/commands/logs.js
```diff
@@ -117,18 +120,13 @@ export default async function main(ctx) {
   }
 
   debug = argv.debug;
```
styfle, line 121: This debug is assigned twice. `debugEnabled` and `debug`

packages/now-cli/src/commands/secrets.js
```diff
@@ -104,9 +103,9 @@ const main = async ctx => {
 
   const {
     authConfig: { token },
+    output,
     config: { currentTeam },
   } = ctx;
-  const output = createOutput({ debug });
   const client = new Client({ apiUrl, token, currentTeam, debug });
```
styfle: Is the Client() constructor missing `output`?

packages/now-cli/src/util/client.ts
```diff
@@ -33,20 +33,22 @@ export default class Client extends EventEmitter {
     forceNew = false,
     withCache = false,
     debug = false,
+    output = createOutput({ debug }),
```
styfle, line 36: Its probably better not to create output here, right? Then all the callers need to pass in the `ctx.output`?

packages/now-cli/src/util/link/setup-and-link.ts
```diff
@@ -39,9 +37,10 @@ export default async function setupAndLink(
 ): Promise<ProjectLinkResult> {
   const {
     authConfig: { token },
+    apiUrl,
+    output,
     config,
   } = ctx;
-  const { apiUrl } = ctx;
   const debug = output.isDebugEnabled();
```
styfle, line 44: I think we can remove a bunch of debug args parsing too if we utilize this `output.isDebugEnabled()` everywhere

packages/now-cli/src/commands/inspect.js
TooTallNate: No, because calling `error()` below implicitly clears the spinner.

packages/now-cli/src/commands/list.js
TooTallNate: The `debug` below is a logger function, not the boolean. Though I agree these should be cleaned up to utilize `output.isDebugEnabled()`, but probably in a separate PR. This one is big enough already.

packages/now-cli/src/util/client.ts
TooTallNate: Mostly addressed in https://github.com/vercel/vercel/pull/5722/commits/fe84eb790dcd49f1f010915eb93ed794a93008b7. There's probably some missing but we can address those in a follow-up. Ideally `Now` and `Client` classes would enforce an `output` being provided and remove the `debug` field in favor of `output.isDebugEnabled()`.

packages/now-cli/src/util/link/setup-and-link.ts
TooTallNate: Agreed, but should be a follow-up PR IMO.

styfle (approved):
💅

PR #5628, opened 2021-01-03
[now-python] Fix qs encoding in http-handler

laiso:
### Issues

Fixes: error occurs when URL including special characters visited #5580 
PR: https://github.com/vercel/vercel/pull/5598
@nICEnnnnnnnLee @styfle

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [x] Issue from task tracker has a link to this PR

## Description

In my application, the request succeeded when I gave the URL argument English, and crashed when I gave it Japanese.

```
$ curl "https://test2020-1z9mjtcwa.vercel.app/handler?hello=world" 

path: /handler query: {'hello': ['world']}

$ curl "https://test2020-1z9mjtcwa.vercel.app/handler?hello=%E4%B8%96%E7%95%8C"

An error occurred with this application.

NO_RESPONSE_FROM_FUNCTION
```

```
> https://vercel.com/laiso/test2020/1z9mjtcwa/functions

20:54:16:08
'ascii' codec can't encode characters in position 19-20: ordinal not in range(128): UnicodeEncodeError
Traceback (most recent call last):
  File "/var/task/now__handler__python.py", line 64, in now_handler
    conn.request(method, path, headers=headers, body=request_body)
  File "/var/lang/lib/python3.6/http/client.py", line 1287, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/var/lang/lib/python3.6/http/client.py", line 1298, in _send_request
    self.putrequest(method, url, **skips)
  File "/var/lang/lib/python3.6/http/client.py", line 1136, in putrequest
    self._output(self._encode_request(request))
  File "/var/lang/lib/python3.6/http/client.py", line 1216, in _encode_request
    return request.encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode characters in position 19-20: ordinal not in range(128)
```

This issue occurs because giving a query string decoded with `unquote ()` to `HTTPConnection.request ()` will recognize it as an invalid URL encode string.

` path = unquote(payload['path'])`

but I don't really understand why `unquote()` was needed. However, it has passed existing tests to protect compatibility.

## about how to fix

First I did a print debug and looked at the raw data for the path in the payload.

When opening `/?hello=世界` in the browser, `/?hello=%E4%B8%96%E7%95%8C` was passed in the payload.
    
[...]

PR #5739, opened 2021-01-21
[python] Fix request URL path encoding for basic, WSGI and ASGI interfaces

TooTallNate:
`@vercel/python` has been incorrectly "unquoting" the incoming request URL for all Python interfaces (basic, WSGI, ASGI). The proper behavior is to pass along the request URL to the serverless function handler unmodified. This is proven by the additional test that is added in this PR, which actually fires up the corresponding Python server via i.e. `python3 api/asgi.py`, and executes the test probes defined in the `test/fixtures/00-request-path/vercel.json` file. Adding this test ensures that the request URL behavior of `@vercel/python` matches the behavior as compared to executing via `python` directly.


packages/now-python/test/fixtures/13-wsgi-url-params-encoding/now.json
```diff
@@ -1,15 +0,0 @@
-{
-  "version": 2,
-  "builds": [{ "src": "*.py", "use": "@vercel/python" }],
-  "routes": [{ "src": "/%E6%82%A8%E5%A5%BD", "dest": "custom.py" }],
-  "probes": [
-    {
-      "path": "/?%E6%82%A8%E5%A5%BD=/",
-      "mustContain": "path:  query: {'您好': '/'}"
```
TooTallNate, line 8: These tests were actually wrong, as compared against other languages like Node.js and against actual Python server implementations (as shown in the new test case in this PR).

packages/now-python/test/fixtures/00-request-path/vercel.json
```diff
@@ -0,0 +1,59 @@
+{
+  "version": 2,
+  "builds": [{ "src": "api/*.py", "use": "@vercel/python" }],
+  "probes": [
```
TooTallNate, line 4: These probes are the _fixed_ versions from the two removed fixtures in this PR (`10-` and `13-`).

packages/now-python/test/fixtures/00-request-path/vercel.json
```diff
@@ -0,0 +1,59 @@
...
+    {
+      "path": "/api/python.py?query=hello%20%40world",
+      "mustContain": "query=hello%20%40world"
+    },
+    { "path": "/api/python.py?hello=%2F", "mustContain": "?hello=%2F" },
```
styfle, line 9: What about `/api/python.py?hello=/` ? What happens then?

TooTallNate: Diverging behavior, stemming from the proxy:

* Proxy will URL-encode the `/` if it's after the `?` before passing it on, so the serverless function would receive `/api/python.py?hello=%2F`. I can't think of anything wrong with doing this, but at the same time I feel like proxy should do as little modifications to the request as possible, because the handling of the original URL may be opinionated (redirecting double slashes is another example where I think proxy gets in the way and prevents certain use cases).
* `fetch()` doesn't automatically URL-encode the `/`, so the Python server receives the original `/api/python.py?hello=/`

styfle:
@TooTallNate Will this close #5628 and #5598?

TooTallNate:
@styfle Looks like it

styfle (dismissed):
Looks great! 

Let's figure out how to notify existing python user's of this change since it could break existing behavior.

TooTallNate:
Added logic to disable `@vercel/python` stable release in the `utils/changelog.js` file.

styfle (approved):

AndyBitz (approved):

PR #5740, opened 2021-01-22
[frameworks] Convert to TypeScript

TooTallNate:
This PR converts the `frameworks.json` file to TypeScript, and extends the values with the detection logic from `@vercel/static-build`, so that it's publicly editable. You also don't need to do the type casting downstream anymore.

As a consequence, it also makes Zola a 1st-class framework, as it was previously missing from the `frameworks.json` file, but present in the static-build frameworks. An example has been included based on their "Getting Started" tutorial.

CH-3808
CH-18771


packages/frameworks/package.json
```diff
@@ -1,14 +1,22 @@
 {
   "name": "@vercel/frameworks",
   "version": "0.2.0",
-  "main": "frameworks.json",
+  "main": "./dist/frameworks.js",
+  "types": "./dist/frameworks.d.ts",
   "license": "UNLICENSED",
```
styfle, line 9: You should add `"files": [ "dist" ]` to avoid publishing unnecessary files to npm

packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -54,7 +53,6 @@ const Schema = {
   items: {
     type: 'object',
     required: ['name', 'slug', 'logo', 'description', 'settings'],
```
styfle: Let's add the additional required props to this test

utils/run.js
```diff
@@ -3,8 +3,8 @@ const { join, relative, sep } = require('path');
 
 // The order matters because we must build dependencies first
 const allPackages = [
-  'frameworks',
   'now-routing-utils',
+  'frameworks',
```
styfle, line 7: Why the change of order here?

styfle: Now that I think about it, maybe frameworks should be built after build-utils so it can import `readConfigFile`

packages/frameworks/src/frameworks.ts
```diff
@@ -0,0 +1,1605 @@
...
+      outputDirectory: {
+        value: 'build',
+      },
+    },
+    devCommand: 'bundle exec middleman server -p $PORT',
+    buildCommand: 'bundle exec middleman build',
+    getOutputDirName: async () => 'build',
+  },
+  {
+    name: 'RedwoodJS',
```
styfle: Any reason why you changed the order of RedwoodJS and Docusaurus 1?

packages/frameworks/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import { Route } from '@vercel/routing-utils/dist/types';
```
styfle:
```suggestion
import { Route } from '@vercel/routing-utils';
```

utils/run.js
TooTallNate: Cause `frameworks` requires the `Route` type from `routing-utils` (see src/types.ts).

packages/frameworks/src/types.ts
TooTallNate: I did it like this so that downstream users that only load the types (i.e. front only needs `Framework` type) don't have to load all of routing-utils into memory.

packages/frameworks/src/frameworks.ts
TooTallNate: Accidental, I'll fix.

packages/frameworks/test/frameworks.unit.test.ts
TooTallNate: I'll add the string ones, but there's no way to test functions with `ajv` so maybe we can manually check for that after ajv.

packages/frameworks/src/types.ts
styfle: Won't it be stripped at build time so it wont be emitted in the bundle js in front?

Also you could try: `import type { Route } from '@vercel/routing-utils';`

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

utils/run.js
styfle: Maybe we should remove `readConfigFile` from build-utils and anything relying on it can import from frameworks instead?

TooTallNate: That's what this PR is doing 😛

packages/frameworks/src/types.ts
TooTallNate: Oh ya I think you're right, that's a good point.

styfle (changes requested):
The api is failing with the following:

```
ERROR	Cannot find module '../packages/frameworks'
Require stack:
- /var/task/api/frameworks.ts
```

Also there are build errors: https://vercel.com/vercel/now/f25ig4pt5

We should enable `noEmitOnError: true` in tsconfig.json

TooTallNate:
Nice catch

api/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "API for the vercel/vercel repo",
   "main": "index.js",
   "scripts": {
-    "build": "yarn --cwd .. && node ../utils/run.js build all"
+    "vercel-build": "yarn --cwd .. && node ../utils/run.js build all"
```
styfle, line 8: I think the API should probably strip the new properties so it remains unchanged. Compare your PR deployment to https://vercel.com/api/v1/frameworks and you'll see the diff.

packages/frameworks/src/frameworks.ts
```diff
@@ -0,0 +1,1605 @@
...
+        continue: true,
+      },
+      {
+        src: '/(.*)',
+        dest: '/index.html',
+      },
+    ],
+  },
+  {
+    name: 'Ember',
```
```suggestion
    name: 'Ember.js',
```

Was this intentionally changed?

packages/frameworks/src/frameworks.ts
```diff
@@ -0,0 +1,1605 @@
...
+      },
+      {
+        src: '.*',
+        status: 404,
+        dest: '404.html',
+      },
+    ],
+  },
+  {
+    name: 'Docusaurus 1.0',
```
```suggestion
    name: 'Docusaurus 1',
```

Was this intentionally changed?

api/package.json
TooTallNate: 8830dd9f559685a9b42156629dd2a90d5495391c

packages/frameworks/src/frameworks.ts
TooTallNate: Updated, but to answer your question, these were the values from `static-build`.

packages/frameworks/src/frameworks.ts
```diff
@@ -0,0 +1,1605 @@
...
+      {
+        src: '/(.*)',
+        headers: { 'cache-control': 's-maxage=0' },
+        dest: '/index.html',
+      },
+    ],
+  },
+  {
+    name: 'Create React App (ejected)',
+    slug: 'create-react-app',
```
styfle: Why are there two entries for `create-react-app`? I think you can remove this one.

styfle (dismissed):
Looks good, thanks! 🏅

styfle (approved):

AndyBitz (approved):

PR #5762, opened 2021-01-28
[cli] Use `output.stopSpinner()` interface consistently

TooTallNate:
There was an issue with `output.spinner()` spinner if the old-style interface was used (invoking the return value to stop the spinner), which caused the internal spinner instance in `output` to be defined, but in a "stopped" state. Then later the `text` gets updated to a different value but the spinner is still stopped and nothing gets rendered.

The fix is to ensure that `output.stopSpinner()` is called rather than directly invoking the cancel function. In fact, the return value of `output.spinner()` has been removed completely.


AndyBitz (dismissed):

AndyBitz (dismissed):

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -111,31 +112,29 @@ function _createOutput({ debug: debugEnabled = false }: OutputOptions = {}) {
...
     if (spinner) {
       spinner.text = message;
     } else {
       spinner = wait(message, delay);
     }
-    return spinner;
   }
 
   function stopSpinner() {
+    if (debugEnabled && spinnerMessage) {
```
styfle, line 129: Is there a case where `spinnerMessage` has a value but `spinner` does not?

I would think this function could be simplified to something like:

```js
function stopSpinner() {
    if (spinner) {
      spinner();
      spinner = null;
      debug(`Spinner stopped (${spinnerMessage})`);
      spinnerMessage = '';
    }
}
```

packages/now-cli/src/util/output/create-output.ts
```diff
@@ -22,14 +22,15 @@ export default function createOutput(opts?: OutputOptions) {
...
+  let spinnerMessage = '';
   let spinner: StopSpinner | null = null;
 
   function isDebugEnabled() {
     return debugEnabled;
   }
 
   function print(str: string) {
-    stopSpinner();
+    if (spinnerMessage) stopSpinner();
```
styfle: Why conditionally call `stopSpinner()`? Seems like you could handle that in the `stopSpinner()` function.

packages/now-cli/src/util/output/create-output.ts
TooTallNate: It's cause when debug is enabled, then `spinner` is never set.

packages/now-cli/src/util/output/create-output.ts
TooTallNate: This one probably isn't necessary anymore.

styfle (approved):

packages/now-cli/src/util/output/create-output.ts
styfle: How about this:

```js
function stopSpinner() {
    debug(`Spinner stopped (${spinnerMessage})`);
    spinnerMessage = '';
    if (spinner) {
      spinner();
      spinner = null;
    }
}
```

AndyBitz (approved):

PR #5768, opened 2021-01-29
[build-utils] bypass multistream, better await

igorklopov:
### Related Issues

Better call await in async functions than pass the stream through multistream. Also i suspect multistream approach doesn't work well when copying numerous files.

#### Tests

#### Code Review


packages/now-build-utils/src/fs/download.ts
```diff
@@ -25,7 +25,7 @@ async function downloadFile(file: File, fsPath: string): Promise<FileFsRef> {
     await symlink(target, fsPath);
     return FileFsRef.fromFsPath({ mode, fsPath });
   } else {
-    const stream = file.toStream();
+    const stream = ('toStreamAsync' in file) ? await file.toStreamAsync() : file.toStream();
```
lucleray: You can also try:
```suggestion
    const stream = file instanceof FileFsRef ? await file.toStreamAsync() : file.toStream();
```

If `toStreamAsync` only exist for `FileFsRef`

styfle (dismissed):

packages/now-build-utils/src/fs/download.ts
igorklopov: Let's do!

lucleray (dismissed):

EndangeredMassa:
Updated to latest `main`.

@lucleray @styfle @TooTallNate is this something we want to add?

cb1kenobi (approved):

PR #5778, opened 2021-02-03
[cli] Add `vercel bisect` command

TooTallNate:
This PR adds a new `vercel bisect` command which is inspired by `git bisect`. This command takes two deployment URLs as its input (one "bad" deployment that exhibits a bug, and one "good" deployment where the bug does not occur) and steps through the deployments using a binary search. The command is useful for finding a deployment (and commit if the project is connected to a Git repo) where a bug has been introduced.

Screen Shot 2022-01-26 at 12 46 29 PM


packages/now-cli/src/commands/bisect/index.ts
```diff
@@ -0,0 +1,323 @@
...
+    if (next) {
+      query.set('until', String(next));
+    }
+
+    const res = await client.fetch<{
+      deployments: Deployment[];
+      pagination: PaginationOptions;
+    }>(`/v6/deployments?${query}`);
+    deployments = deployments.concat(res.deployments);
+    next = res.pagination.next;
```
styfle: Do we need to sleep() here in this loop to avoid rate limiting?

packages/now-cli/src/commands/bisect/index.ts
```diff
@@ -0,0 +1,323 @@
...
+  let deployments: Deployment[] = [];
+  do {
+    const query = new URLSearchParams();
+    query.set('projectId', project.id);
+    query.set('target', 'production');
+    query.set('limit', '100');
+    query.set('state', 'READY');
+    if (next) {
+      query.set('until', String(next));
+    }
```
styfle: Seems like your implementation is going to fetch all deployments for the project. It should probably stop once the original "good" and "bad" deployments are both fetched (along with everything in between of course)

packages/now-cli/src/commands/bisect/index.ts
TooTallNate: I think this should be done in the `Client` class. I'll take a look into implementing.

packages/now-cli/src/commands/bisect/index.ts
TooTallNate: Yup, adding the ranging logic today.

packages/now-cli/src/commands/bisect/index.ts
TooTallNate: Added `sleep(100)`

packages/cli/src/commands/bisect/index.ts
```diff
@@ -0,0 +1,411 @@
...
+        // Script does not exist or is not executable, so exit
+        output.prettyError(proc);
+        return 1;
+      }
+      const { exitCode } = proc;
+      let color: Chalk;
+      if (exitCode === 0) {
+        color = chalk.green;
+        action = 'good';
+      } else if (exitCode === 125) {
```
styfle, line 309: Do we need to document this exit code?

packages/cli/src/commands/bisect/index.ts
```diff
@@ -0,0 +1,411 @@
...
+
+  if (badDeployment.createdAt < goodDeployment.createdAt) {
+    output.error(`Good deployment must be older than the Bad deployment`);
+    return 1;
+  }
+
+  if (badDeployment.target !== goodDeployment.target) {
+    output.error(
+      `Bad deployment target "${badDeployment.target}" does not match good deployment target "${goodDeployment.target}"`
+    );
```
styfle, line 205: We might want to use `target || 'preview'` since `target` is undefined for preview deployments

packages/cli/src/commands/bisect/index.ts
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  if (badDeployment.createdAt < goodDeployment.createdAt) {
+    output.error(`Good deployment must be older than the Bad deployment`);
+    return 1;
+  }
+
+  if (badDeployment.target !== goodDeployment.target) {
+    output.error(
+      `Bad deployment target "${badDeployment.target}" does not match good deployment target "${goodDeployment.target}"`
```
styfle: Pretty sure `target` is undefined for preview deployments

```suggestion
      `Bad deployment target "${badDeployment.target || 'preview'}" does not match good deployment target "${goodDeployment.target || 'preview'}"`
```

packages/cli/src/commands/bisect/index.ts
TooTallNate: Yes

packages/cli/src/types.ts
```diff
@@ -133,6 +133,13 @@ export type Deployment = {
   created: number;
   createdAt: number;
   creator: { uid: string; username: string };
+  target: string | null;
```
styfle, line 132:
```suggestion
  target: string | null | undefined;
```

styfle (approved):

PR #5811, opened 2021-02-10
[build-utils][cli] Fix frameworks usage

TooTallNate:
Fixes unit test failures like: https://github.com/vercel/vercel/runs/1867209532#step:11:191 from PR #5740


styfle (approved):
Very nice, we can finally remove the underscore workaround 😌

AndyBitz (approved):

PR #5814, opened 2021-02-10
Update Codecov package and add Codecov badge to README

thomasrockhu:
Tom from Codecov here. Looks like this repository wasn't sending coverage reports to Codecov for quite awhile. I updated the package to do so and added the Codecov badge.

#### Tests

#### Code Review

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 
 [![CI Status](https://badgen.net/github/checks/vercel/vercel?label=CI)](https://github.com/vercel/vercel/actions?workflow=CI)
 [![Join the community on GitHub Discussions](https://badgen.net/badge/join%20the%20discussion/on%20github/black?icon=github)](https://github.com/vercel/vercel/discussions)
+[![codecov](https://codecov.io/gh/vercel/vercel/branch/master/graph/badge.svg)](https://codecov.io/gh/vercel/vercel)
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


TooTallNate:
So all you had to do was update the package?

thomasrockhu:
@TooTallNate, seems like it. It could be a combination of fixes we made to the uploader and fixes we made recently. I would be inclined to believe it's the uploader, though.

README.md
```diff
@@ -8,6 +8,7 @@
 
 [![CI Status](https://badgen.net/github/checks/vercel/vercel?label=CI)](https://github.com/vercel/vercel/actions?workflow=CI)
 [![Join the community on GitHub Discussions](https://badgen.net/badge/join%20the%20discussion/on%20github/black?icon=github)](https://github.com/vercel/vercel/discussions)
+[![codecov](https://codecov.io/gh/vercel/vercel/branch/master/graph/badge.svg)](https://codecov.io/gh/vercel/vercel)
```
styfle, line 11: Let's not add a badge until we can get this number higher

styfle (commented):
Thanks for the PR!

Does this take into account the monorepo paths?

I would expect `packages/now-cli/src/util/dev/router.ts` as the path, not `src/util/dev/router.ts` which is what its reporting.

Ideally, we would add code coverage to all the packages in this monorepo.

PR #5818, opened 2021-02-10
[frameworks] Add test case to ensure all "demo" URLs are public Vercel deployments

TooTallNate:
All demo URLs must be public for the template import flow, so this test case ensures that we don't accidentally publish a demo URL that is not public.


styfle (approved):
Great, thanks!

AndyBitz (approved):
I think we have a test that does the same thing in Checkly, but probably good to have it here before it's deployed.

pacocoursey (approved):

PR #5552, opened 2020-12-10
[build-utils][static-build][next] Force `YARN_NODE_LINKER="node-modules"` env var when executing `yarn`

TooTallNate:
### Related Issues

* https://vercel.com/knowledge/does-vercel-support-yarn-2
* https://github.com/vercel/vercel/discussions/4223
* https://github.com/vercel/vercel/discussions/4910
* https://github.com/vercel/vercel/issues/5136
* https://github.com/vercel/vercel/issues/5280

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #5825, opened 2021-02-11
[build-utils] Replace cross-spawn with execa

frencojobs:
### Related Issues

There is an error that happens when using the [vercel-dev-runtime](https://github.com/TooTallNate/vercel-dev-runtime) as a runtime on the deployed functions. 

errer

It seems that the error comes from the `runPackageJsonScript` function. It only happens on the deployed functions and can't be reproduced locally with the `vercel dev` command.

It seems to be caused by the [cross-spawn](https://github.com/moxystudio/node-cross-spawn/issues/104) package which is used to execute the scripts from `package.json`. Even though I couldn't find an exact relation between the mentioned issue and the error, the error can be easily fixed by switching to the [execa](https://github.com/sindresorhus/execa) package.

This PR replaces all usages of `cross-spawn` with `execa`.

#### Tests

#### Code Review

### DIFF
--- a/packages/now-build-utils/package.json
+++ b/packages/now-build-utils/package.json
@@ -19,7 +19,6 @@
   "devDependencies": {
     "@iarna/toml": "2.2.3",
     "@types/async-retry": "^1.2.1",
-    "@types/cross-spawn": "6.0.0",
     "@types/end-of-stream": "^1.4.0",
     "@types/fs-extra": "^5.0.5",
     "@types/glob": "^7.1.1",
@@ -35,8 +34,8 @@
     "async-retry": "1.2.3",
     "async-sema": "2.1.4",
     "boxen": "4.2.0",
-    "cross-spawn": "6.0.5",
     "end-of-stream": "1.4.1",
+    "execa": "5.0.0",
     "fs-extra": "7.0.0",
     "glob": "7.1.3",
     "into-stream": "5.0.0",
--- a/packages/now-build-utils/src/fs/run-user-scripts.ts
+++ b/packages/now-build-utils/src/fs/run-user-scripts.ts
@@ -2,15 +2,14 @@ import assert from 'assert';
 import fs from 'fs-extra';
 import path from 'path';
 import debug from '../debug';
-import spawn from 'cross-spawn';
-import { SpawnOptions } from 'child_process';
+import execa from 'execa';
 import { deprecate } from 'util';
 import { cpus } from 'os';
 import { NowBuildError } from '../errors';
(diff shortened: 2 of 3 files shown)


TooTallNate:
Interesting. I'm curious how you are testing. Do you have an idea about why the `runNpmInstall()` works, but then immediately afterwards the `runPackageJsonScript()` is failing? That's the part that I haven't figure out.

frencojobs:
Ok, so I did a little more debugging and I think I found the culprit.

This line of code overrides the environment variables of the spawning process, which would basically remove the `PATH` variable, therefore, removing the installed path of `yarn`.
```ts
const env: typeof process.env = { ...spawnOpts?.env };
// ...
await spawnAsync('yarn', ['run', scriptName], {
    ...spawnOpts,
    env, // <- this
    cwd: destPath,
    prettyCommand,
  });
```

To test the scenario, I passed the `PATH` variable from the vercel-dev-runtime like this,
```ts
await runPackageJsonScript(workPath, 'build', {
  env: {
    PATH: process.env.PATH,
  },
});
```
and it worked. Here is the code https://github.com/frencojobs/vercel-dev-runtime/blob/fab5484fd2c1d00c8f48577ac56b85b3ec65d30c/index.js#L21. It's not using the execa version though. [Here](https://vercel.com/frenco/vercel-dart/dih2gvpdy) is a succeeded log from a deployment using it.

The reason why it worked with execa is that execa uses this function to add all required variables before passing them to the process spawner.

```ts
const getEnv = ({env: envOption, extendEnv, preferLocal, localDir, execPath}) => {
  // `extendEnv` option is `true` by default in execa
  const env = extendEnv ? {...process.env, ...envOption} : envOption;
	if (preferLocal) {
	  return npmRunPath.env({env, cwd: localDir, execPath});
	}
  return env;
};
```

So, we can also just merge that env object with `process.env` before passing it to the `cross-spawn`, instead of using execa I guess. But if there's no specific reason to use cross-spawn instead of execa, execa seems like a better option IMO.

What do you think?

styfle:
I think the issue was introduced in #5552

We should probably change this line:

```ts
const env: typeof process.env = { ...spawnOpts?.env };
```

to this:

```ts
const env: typeof process.env = { ...process.env, ...spawnOpts?.env };
```

TooTallNate:
Hey @frencojobs, thanks for the PR btw. I think #5841 is a better solution in this case. Let me know what you think.

frencojobs:
Cool, thanks. I just love that it works now.

PR #5828, opened 2021-02-11
[frameworks] Restore correct order of frameworks for detection

TooTallNate:
Follow up to #5740 to restore the original order of the frameworks.


styfle (approved):
Thanks!

PR #5839, opened 2021-02-12
[frameworks] Add `defaultVersion` field

TooTallNate:
This will be used on the docs page to render the default version that is included in the build image.

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1385,7 +1385,7 @@ export const frameworks = [
 
       return (config && config.publishDir) || 'public';
     },
-    defaultVersion: 'v0.58.2',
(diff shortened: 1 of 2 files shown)


AndyBitz (dismissed):

packages/frameworks/src/frameworks.ts
```diff
@@ -1523,6 +1524,7 @@ export const frameworks = [
     devCommand: 'zola serve --port $PORT',
     buildCommand: 'zola build',
     getOutputDirName: async () => 'public',
+    defaultVersion: 'v0.13.0',
```
styfle: Let's drop the `v` prefix since we can add it in the docs if we want but its hard to remove it.

Also the env var doesn't accept the `v` prefix.

styfle (approved):
💅

PR #5840, opened 2021-02-12
[tests] Add `--network-timeout 1000000` to yarn install

TooTallNate:
Seems like this might help with the networking issues in CI. See https://github.com/yarnpkg/yarn/issues/4890#issuecomment-358179301


AndyBitz (approved):

styfle (approved):
🤞

PR #5841, opened 2021-02-12
[build-utils] Inherit `process.env` when `runPackageJsonScript()` is invoking yarn

TooTallNate:
This is a more proper fix for #5825.

Closes #5825.

### DIFF
--- a/packages/now-build-utils/test/unit.test.js
+++ b/packages/now-build-utils/test/unit.test.js
@@ -8,6 +8,7 @@ const {
   getNodeVersion,
   getLatestNodeVersion,
   getDiscontinuedNodeVersions,
+  runNpmInstall,
   runPackageJsonScript,
 } = require('../dist');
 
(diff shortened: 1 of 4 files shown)


styfle (dismissed):

packages/now-build-utils/test/unit.test.js
```diff
@@ -255,3 +256,13 @@ it('should support require by path for legacy builders', () => {
   expect(FileRef2).toBe(index.FileRef);
   expect(Lambda2).toBe(index.Lambda);
 });
+
+it('should have correct $PATH when running `runPackageJsonScript()` with yarn', async () => {
+  const fixture = path.join(__dirname, 'fixtures/19-yarn-v2');
```
styfle:
```suggestion
  const fixture = path.join(__dirname, 'fixtures', '19-yarn-v2');
```

TooTallNate: Doesn't really make a difference. Node.js normalizes the slashes, even on Windows.

styfle (dismissed):

AndyBitz (dismissed):

PR #5873, opened 2021-02-20
[go] Add Go version selection via `go.mod`

styfle:
#### Tests

#### Code Review


packages/now-go/go-helpers.ts
```diff
@@ -1,17 +1,18 @@
...
+const archMap = new Map([
+  ['x64', 'amd64'],
+  ['x86', '386'],
+]);
 const platformMap = new Map([['win32', 'windows']]);
-
-// Location where the `go` binary will be installed after `postinstall`
-const GO_DIR = join(__dirname, 'go');
-const GO_BIN = join(GO_DIR, 'bin', 'go');
+export const cachDir = '.vercel-go-cache';
```
TooTallNate: Typo (needs to be updated in all places)?

```suggestion
export const cacheDir = '.vercel-go-cache';
```

packages/now-go/go-helpers.ts
```diff
@@ -156,5 +166,20 @@ export async function downloadGo(
...
     });
   }
-  return createGo(dir, platform, arch);
+  return createGo(workPath, dir, platform, arch);
+}
+
+async function parseGoVersion(modulePath: string): Promise<string> {
+  let version = '1.16';
+  const file = join(modulePath, 'go.mod');
+  if (!existsSync(file)) {
```
TooTallNate: I'd prefer to `try/catch` the `readFile()` function and print the `debug()` line when `err.code === 'ENOENT'`.

packages/now-go/go-helpers.ts
```diff
@@ -156,5 +166,20 @@ export async function downloadGo(
...
+    debug(`go.mod not found: ${file}`);
+    return version;
+  }
+  const content = await readFile(file, 'utf8');
+  const match = /^go (\d+\.\d+)\.?$/gm.exec(content);
+  if (match && match[1]) {
```
TooTallNate:
```suggestion
  if (match?[1]) {
```

packages/now-go/index.ts
```diff
@@ -136,8 +142,12 @@ Learn more: https://github.com/golang/go/wiki/Modules
     if (fileName in downloadedFiles) {
       goModAbsPathDir = dirname(downloadedFiles[fileName].fsPath);
       debug(`Found ${fileName} file in "${goModAbsPathDir}"`);
+    } else if ('api/go.mod' in downloadedFiles) {
```
TooTallNate, line 145: What if I have `api/foo/go.mod`? I feel like we should be using `scanParentDirs()` or whatever that function is called to iterate up and find the `go.mod` file.

styfle: @TooTallNate We don't allow that behavior and I don't think Go does either. This was really just a feature for users to separate backend and frontend (backend being everything in the api directory).

TooTallNate (dismissed):

packages/now-go/go-helpers.ts
```diff
@@ -156,5 +163,26 @@ export async function downloadGo(
...
+    if (major === 1 && minor > 12) {
+      return `${major}.${minor}`;
+    } else {
+      debug(`Failed to parse version or its too old: ${file}`);
+    }
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      debug(`File not found: ${file}`);
+    }
+    throw err;
```
TooTallNate:
```suggestion
    } else {
      throw err;
    }
```

TooTallNate (changes requested):

TooTallNate (approved):

PR #5905, opened 2021-02-25
[frameworks] add snowpack-react

coetry:
This PR adds framework detection for the Snowpack React template

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1590,7 +1590,7 @@ export const frameworks = [
         placeholder: '`yarn install` or `npm install`',
       },
       buildCommand: {
-        placeholder: '`npm run build` or `react-scripts build`',
+        placeholder: '`npm run build` or `snowpack build`',
       },
       devCommand: {
         value: 'snowpack dev',
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1561,6 +1561,64 @@ export const frameworks = [
     buildCommand: null,
     getOutputDirName: async () => 'public',
   },
+  {
+    name: 'React Snowpack',
+    slug: 'react-snowpack',
+    demo: 'https://react-snowpack-delta.vercel.app',
+    logo:
+      'https://raw.githubusercontent.com/vercel/vercel/master/packages/frameworks/logos/react.svg',
+    tagline:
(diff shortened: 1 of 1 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -1561,6 +1561,64 @@ export const frameworks = [
...
+      },
+    },
+    dependency: '@snowpack/plugin-react-refresh',
+    devCommand: 'snowpack dev',
+    buildCommand: 'snowpack build',
+    getOutputDirName: async () => 'build',
+    defaultRoutes: [
+      {
+        src: '/static/(.*)',
+        headers: { 'cache-control': 's-maxage=31536000, immutable' },
```
styfle, line 1609: Is this documented?

packages/frameworks/src/frameworks.ts
```diff
@@ -1561,6 +1561,64 @@ export const frameworks = [
...
+        continue: true,
+      },
+      {
+        handle: 'filesystem',
+      },
+      {
+        src: '/(.*)',
+        headers: { 'cache-control': 's-maxage=0' },
```
styfle, line 1617: What is the purpose of this header? We already purge static files for each deployment so I don't think its necessary.

packages/frameworks/src/frameworks.ts
```diff
@@ -1561,6 +1561,64 @@ export const frameworks = [
     buildCommand: null,
     getOutputDirName: async () => 'public',
   },
+  {
```
styfle, line 1564: The order of this file matters, so I would move Snowpack above "Other"

packages/frameworks/src/frameworks.ts
```diff
@@ -1561,6 +1561,64 @@ export const frameworks = [
...
+    name: 'React Snowpack',
+    slug: 'react-snowpack',
+    demo: 'https://react-snowpack-delta.vercel.app',
+    logo:
+      'https://raw.githubusercontent.com/vercel/vercel/master/packages/frameworks/logos/react.svg',
+    tagline:
+      'Unlike traditional React application tooling, Snowpack saves you from getting bogged down with complex bundler setups and configuration files.',
+    description: 'A React app, bootstrapped with create-snowpack-app',
```
styfle, line 1572: It seems like `create-snowpack-app` doesn't work by itself but it requires a template. Which template did you use? Do we need other framework detection entries for each template, for example Vue?

EndangeredMassa:
@coetry this PR is pretty old at this point. If you'd like to re-open it after addressing the feedback, feel free!

PR #5919, opened 2021-03-01
[routing-utils] Add has route field to schema

ijjk:
### Related Issues

x-ref: https://github.com/vercel/next.js/pull/22341
x-ref: https://github.com/vercel/next.js/issues/22345

#### Tests

#### Code Review


packages/now-routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,27 @@
+const hasSchema = {
+  type: 'array',
+  maxItems: 1024,
```
dav-is: I think this should probably be `12` or something around that

styfle: Yeah or maybe 16 if you want to keep it base 2. 

1024 seems too large if you only have 3 keys to choose from.

ijjk: Updated to `16`

styfle (dismissed):

dav-is (dismissed):

dav-is (dismissed):

packages/routing-utils/src/types.ts
```diff
@@ -67,17 +72,32 @@ export interface VercelConfig {
...
+  }>;
 }
 
 export interface Header {
   source: string;
+  has?: Array<{
+    type: 'header' | 'cookie' | 'query';
+    key: string;
+    value?: string;
+  }>;
```
styfle: If these are all the same, then it might make more sense to create a new type

```ts
export interface HasKeyValue {
    type: 'header' | 'cookie' | 'query';
    key: string;
    value?: string;
  }
```

styfle: Or maybe just

```ts
export type Has = Array<{
    type: 'header' | 'cookie' | 'query';
    key: string;
    value?: string;
  }>
```

ijjk: Updated to share the type 👍

styfle (approved):

dav-is (approved):

PR #5931, opened 2021-03-05
[tests] Install Go in GH Actions Workflows

styfle:
We need to make CI explicit about the lowest common denominator for Go Version like we do with the Node.js Version.

This will add about 10 seconds to CI which is negligible to the 8 min+ CI currently.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,10 +12,13 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: actions/setup-node@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '1.13.15'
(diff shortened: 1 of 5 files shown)


AndyBitz (approved):

.github/workflows/publish.yml
```diff
@@ -12,10 +12,13 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: actions/setup-node@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '1.13.15'
```
TooTallNate, line 17: Is it needed for "Publish"?

styfle: Yes, its needed for `yarn build`

TooTallNate (approved):

ernestd (commented):
lgtm

PR #5957, opened 2021-03-09
[cli] Add SAML Single Sign-On to `vercel login`

TooTallNate:
Refactors the `vercel login` command to be more modular and adds SAML Single Sign-On functionality, where the user's web browser is automatically opened to the identity provider's authentication page, before finally being redirected back to a server that CLI is running on localhost in order to retrieve the final auth token.

### DIFF
--- a/packages/cli/src/util/login/email.ts
+++ b/packages/cli/src/util/login/email.ts
@@ -3,7 +3,6 @@ import { stringify as stringifyQuery } from 'querystring';
 import fetch from 'node-fetch';
 import sleep from '../sleep';
 import ua from '../ua';
-import ok from '../output/ok';
 import error from '../output/error';
 import highlight from '../output/highlight';
 import eraseLines from '../output/erase-lines';
@@ -79,7 +78,6 @@ export default async function doEmailLogin(
   }
 
   // Clear up `Sending email` success message
-  //output.print(eraseLines(possibleAddress ? 1 : 2));
   output.print(eraseLines(1));
 
   output.print(
@@ -109,6 +107,6 @@ export default async function doEmailLogin(
     }
   }
 
-  output.log(ok('Email confirmed'));
+  output.success('Email confirmed');
   return token;
 }
--- a/packages/cli/src/util/login/email.ts
+++ b/packages/cli/src/util/login/email.ts
@@ -1,18 +1,12 @@
 import ms from 'ms';
-import chalk from 'chalk';
 import { stringify as stringifyQuery } from 'querystring';
 import fetch from 'node-fetch';
 import sleep from '../sleep';
 import ua from '../ua';
-import hp from '../humanize-path';
-import { getCommandName } from '../pkg-name';
 import ok from '../output/ok';
 import error from '../output/error';
 import highlight from '../output/highlight';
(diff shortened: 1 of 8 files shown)


packages/cli/src/util/login/email.ts
```diff
@@ -0,0 +1,114 @@
...
+    const data = await executeLogin(apiUrl, email);
+    verificationToken = data.token;
+    securityCode = data.securityCode;
+  } catch (err) {
+    output.error(err.message);
+    return 1;
+  }
+
+  // Clear up `Sending email` success message
+  //output.print(eraseLines(possibleAddress ? 1 : 2));
```
styfle: Is `possibleAddress` still used?

packages/cli/src/util/login/sso.ts
```diff
@@ -0,0 +1,102 @@
...
+          );
+          res.end();
+        });
+        server.once('error', reject);
+      }),
+      open(url.href),
+    ]);
+
+    const rawQuery = req.url!.slice(req.url!.indexOf('?') + 1);
+    const query = new URLSearchParams(rawQuery);
```
styfle:
```suggestion
    const query = new URL(req.url || '/', 'http://localhost').searchParams;
```

Nit pick: use url parsing instead of string slicing

packages/cli/src/util/login/sso.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (!verifyRes.ok) {
+      output.error(
+        `Unexpected ${verifyRes.status} status code from verify API`
+      );
+      output.debug(await verifyRes.text());
+      return 1;
+    }
+
+    output.success(
+      `Successfully logged in to "${slug}" team as ${chalk.bold(
```
styfle:
```suggestion
      `Successfully logged into "${slug}" team as ${chalk.bold(
```

Is this going to print two login messages?
This one and `You are now logged in. In order to deploy something...`?

TooTallNate: I'm not so sure about the suggestion: https://helpcenter.veeam.com/docs/styleguide/tw/log_in_to.html

And ya, there's two success messages. Will figure out the best way to simplify that.

packages/cli/src/util/login/sso.ts
```diff
@@ -0,0 +1,101 @@
+import http from 'http';
+import open from 'open';
+import chalk from 'chalk';
+import fetch from 'node-fetch';
+import { hostname } from 'os';
+import { URL, URLSearchParams } from 'url';
```
styfle:
```suggestion
import { URL } from 'url';
```

packages/cli/src/util/login/email.ts
TooTallNate: Removed.

packages/cli/src/util/login/sso.ts
TooTallNate: Changed the message to just say SAML was completed. Doesn't mention "login" specifically, since that happens on the next line.

styfle (dismissed):

styfle (approved):

PR #6007, opened 2021-03-18
[frameworks] Add envPrefix property

styfle:
We need this logic in frameworks for static-build as well as docs.


packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -136,6 +137,7 @@ const Schema = {
       cachePattern: { type: 'string' },
       buildCommand: { type: ['string', 'null'] },
       devCommand: { type: ['string', 'null'] },
+      defaultVersion: { type: 'string' },
```
styfle, line 140: This was missing from a previous PR #5839 , cc @TooTallNate

AndyBitz (approved):

PR #6014, opened 2021-03-19
[frameworks] Fix "svelte" detection and Dev Command

styfle:
- We were matching any project using [sirv](https://www.npmjs.com/package/sirv) which was too broad.
- Hot reloading didn't work because default dev command was incorrect.

This PR updates the framework detection and default dev command to match the example:

https://github.com/vercel/vercel/blob/c2e7be80e869137d3dd482e017c9e65a1431225a/examples/svelte/package.json#L6

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -855,14 +855,14 @@ export const frameworks = [
         placeholder: '`npm run build` or `rollup -c`',
       },
       devCommand: {
-        value: 'sirv public --single --dev --port $PORT',
+        value: 'rollup -c -w',
       },
       outputDirectory: {
         value: 'public',
       },
     },
     dependency: 'sirv-cli',
-    devCommand: 'sirv public --single --dev --port $PORT',
+    devCommand: 'rollup -c -w',
     buildCommand: 'rollup -c',
     getOutputDirName: async () => 'public',
     defaultRoutes: [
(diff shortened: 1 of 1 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -855,14 +855,14 @@ export const frameworks = [
...
       },
       devCommand: {
-        value: 'sirv public --single --dev --port $PORT',
+        value: 'rollup -c -w',
       },
       outputDirectory: {
         value: 'public',
       },
     },
     dependency: 'sirv-cli',
```
AndyBitz, line 869: I guess this doesn't make sense either then, since `sirv-cli` doesn't seem to be specific for Svelte

AndyBitz (dismissed):

packages/frameworks/src/frameworks.ts
styfle: Do we need to change line 846 to something else? Its currently matching `sirv-cli` as a dependency

https://github.com/vercel/vercel/blob/c2e7be80e869137d3dd482e017c9e65a1431225a/packages/frameworks/src/frameworks.ts#L846

AndyBitz: I think it would make the most sense to add svelte as dependency there as well, to ensure both are matched

AndyBitz: Updated the check, I guess we can leave this one for now, since it's only used as a fallback at build time

AndyBitz (approved):

TooTallNate (approved):

PR #6016, opened 2021-03-20
[cli] Refactor to merge `NowContext` into `Client`

TooTallNate:
This is a rather large refactor of the CLI codebase that merges the `NowContext` object into the `Client` class. There's a few reasons to do this:

1. Nearly every sub-command was creating its own `Client` instance, leading to a lot of duplicated code. It makes more sense to create one instance and pass it to every sub-command, which deletes a lot of code as well.
2. There's a lot of overlap between `NowContext` and the props on the `Client` class, so it was a relatively easy transition.
3. The main reason is so the same `authConfig` is used throughout the program lifecycle. This paves the way for updating the token mid-flight when i.e. a SAML token has expired. In a follow-up PR, CLI will handle that scenario gracefully by re-triggering the SAML auth flow.


packages/cli/src/commands/deploy/latest.js
```diff
@@ -295,7 +287,7 @@ export default async function main(
         output,
         'Which scope do you want to deploy to?',
         client,
-        ctx.config.currentTeam,
+        client.config.currentTeam,
```
styfle: `selectOrg()` already has a client so can we remove `currentTeam` parameter here?

packages/cli/src/commands/env/index.ts
```diff
@@ -129,7 +120,7 @@ export default async function main(ctx: NowContext) {
     return 1;
   } else {
     const { project, org } = link;
-    client.currentTeam = org.type === 'team' ? org.id : undefined;
+    config.currentTeam = org.type === 'team' ? org.id : undefined;
```
styfle, line 123: Whats the difference between `client.currentTeam` and `client.config.currentTeam`?

packages/cli/src/commands/teams/invite.js
```diff
@@ -163,57 +146,52 @@ export default async function ({
...
+        output.log(`${chalk.cyan(chars.tick)} ${sentEmailPrefix}${email}`);
         if (hasError) {
           hasError = false;
-          process.stdout.write(eraseLines(emails.length + 2));
-          console.log(
-            info(
-              introMsg ||
-                `Inviting team members to ${chalk.bold(currentTeam.name)}`
-            )
+          process.stderr.write(eraseLines(emails.length + 2));
```
styfle, line 164: Do you need a `output.stopSpinner()` here before erasing lines?

packages/cli/src/util/client.ts
```diff
@@ -98,45 +91,43 @@ export default class Client extends EventEmitter {
...
 
-        return res.headers.get('content-type').includes('application/json')
-          ? res.json()
-          : res;
+      if (opts.json === false) {
+        return res;
       }
-      const error = await responseError(res);
-      if (res.status >= 400 && res.status < 500) {
-        return bail(error);
```
styfle, line 131: Why did you remove the `bail()` here?

packages/cli/src/commands/env/index.ts
TooTallNate: `client.currentTeam` no longer exists, instead it takes the entire "global config" object as `client.config`.

packages/cli/src/commands/teams/add.js
```diff
@@ -144,12 +144,8 @@ export default async function({ apiUrl, token, teams, config }) {
 
   stopSpinner();
 
-  await invite({
-    teams,
-    args: [],
-    token,
-    apiUrl,
-    config,
+  client.argv._ = [];
```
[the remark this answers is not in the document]
TooTallNate: Ya you're right, this is incorrect. Updated.

packages/cli/src/commands/teams/invite.js
TooTallNate: Not in this case because `output.log()` on line 161 stops the spinner.

packages/cli/src/commands/teams/invite.js
```diff
@@ -104,14 +90,14 @@ export default async function ({
...
 
-  console.log(
-    info(introMsg || `Inviting team members to ${chalk.bold(currentTeam.name)}`)
+  output.log(
+    introMsg || `Inviting team members to ${chalk.bold(currentTeam.name)}`
   );
 
-  if (args.length > 0) {
-    for (const email of args) {
+  if (argv._.length > 0) {
```
styfle, line 97: This length is always zero now, right? Since its hardcoded by the caller?

styfle: Oh nevermind, I see there are two callers

styfle (approved):

packages/cli/src/commands/teams/invite.js
TooTallNate: The hard-coding is when you do `vc teams add` (which finishes by allowing you to invite team members). But you can still do `vc teams invite xxx@example.com` in which case the length would not be zero.

PR #6023, opened 2021-03-22
[routing-utils] Add schema descriptions to routing schema

dav-is:
Our automated documentation requires descriptions in the schema

#### Tests

#### Code Review

### DIFF
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -1,4 +1,4 @@
-const hasSchema = {
+export const hasSchema = {
   description: 'An array of requirements that are needed to match',
   type: 'array',
   maxItems: 16,
@@ -48,7 +48,7 @@ const hasSchema = {
       },
     ],
   },
-};
+} as const;
 
 /**
  * An ajv schema for the routes array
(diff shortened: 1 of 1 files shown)


packages/routing-utils/src/schemas.ts
```diff
@@ -207,16 +233,21 @@ export const redirectsSchema = {
...
+          'A pattern that matches each incoming pathname (excluding querystring)',
         type: 'string',
         maxLength: 4096,
       },
       headers: {
+        description:
+          'An array of key/value pairs representing each response header.',
         type: 'array',
         maxItems: 1024,
         items: {
```
styfle, line 253: Missing description on `key` and `value` below

styfle (dismissed):

ijjk (dismissed):

packages/routing-utils/src/schemas.ts
dav-is: Copied from here: https://github.com/vercel/api/blob/4ec9ff002e8d866754ace4aaee8c74da3c9e010d/utils/deployments-util/vercel-json-schema.ts#L237

TooTallNate (changes requested):
We need to update these to use `as const` at the end of each object. This is for our API docs generator to pick up the proper types.

TooTallNate (approved):

ijjk (approved):

PR #6049, opened 2021-03-30
[cli] (Major) Update `vc env` command

styfle:
### Changes

- Add support for Git Branch
- Use a single env type: "encrypted"
- Remove unnecessary questions during `vc env add`

#### Tests

#### Code Review


packages/cli/src/commands/env/ls.ts
```diff
@@ -1,15 +1,11 @@
...
-  Project,
-  ProjectEnvVariable,
-  ProjectEnvType,
-} from '../../types';
+import { Project, ProjectEnvVariable, ProjectEnvType } from '../../types';
 import Client from '../../util/client';
 import formatTable from '../../util/format-table';
-import getEnvVariables from '../../util/env/get-env-records';
+import getEnvRecords from '../../util/env/get-env-records';
+import formatEnvTarget from '../../util/env/format-env-target';
```
ernestd, line 8: Can we use aliases for the base folders?

ernestd (dismissed):

packages/cli/src/commands/env/ls.ts
styfle: I don't think `ncc` supports that TS feature

packages/cli/src/types.ts
```diff
@@ -208,22 +208,21 @@ export enum ProjectEnvTarget {
 export enum ProjectEnvType {
   Plaintext = 'plain',
   Secret = 'secret',
+  Encrypted = 'encrypted',
   System = 'system',
 }
 
 export interface ProjectEnvVariable {
+  id: string;
```
styfle, line 217: I confirmed with @lucleray that all env vars now have an ID thanks to a previous migration

packages/cli/src/commands/env/add.ts
```diff
@@ -219,17 +126,32 @@ export default async function add(
...
+    !envGitBranch &&
+    envTargets.length === 1 &&
+    envTargets[0] === ProjectEnvTarget.Preview
+  ) {
+    const { inputValue } = await inquirer.prompt({
+      type: 'input',
+      name: 'inputValue',
+      message: `Add ${envName} to which Git Branch? (leave empty for all Prevew branches)?`,
+    });
+    envGitBranch = inputValue || '';
```
TooTallNate, line 140: Is the `|| ''` necessary? I figure the prompt would return empty string already.

styfle: I have no idea since the prompt returns `any` so just doing this to be safe (we do it the same on the other prompts)

packages/cli/src/util/env/add-env-record.ts
```diff
@@ -13,26 +12,22 @@ export default async function addEnvRecord(
...
-  await client.fetch<ProjectEnvVariableV5>(urlProject, {
+  const body: Omit<ProjectEnvVariable, 'id'> = {
+    type,
+    key,
+    value,
+    target: targets,
+    gitBranch: gitBranch || undefined,
+  };
+  const url = `/v7/projects/${projectId}/env`;
+  await client.fetch<ProjectEnvVariable>(url, {
```
TooTallNate: I think we don't need the type since we aren't using the return value?

```suggestion
  await client.fetch(url, {
```

TooTallNate (approved):

ernestd (approved):

packages/cli/src/commands/env/add.ts
```diff
@@ -219,17 +126,32 @@ export default async function add(
...
+  if (
+    !stdInput &&
+    !envGitBranch &&
+    envTargets.length === 1 &&
+    envTargets[0] === ProjectEnvTarget.Preview
+  ) {
+    const { inputValue } = await inquirer.prompt({
+      type: 'input',
+      name: 'inputValue',
+      message: `Add ${envName} to which Git Branch? (leave empty for all Prevew branches)?`,
```
lucleray:
```suggestion
      message: `Add ${envName} to which Git Branch? (leave empty for all Preview branches)?`,
```

lucleray (approved):

PR #6059, opened 2021-04-01
Ensure has segments are replaced in destination

ijjk:
This ensures named matches from has items are updated in the destination e.g. this makes sure we convert `/:username` -> `/$username` when `{ type: 'header', key: 'x-username', value: '(?<username>.*)' }` is used.

### Related Issues

x-ref: https://github.com/vercel/vercel/pull/5919

#### Tests

#### Code Review


packages/routing-utils/src/superstatic.ts
```diff
@@ -174,28 +182,60 @@ export function sourceToRegex(
   return { src: r.source, segments };
 }
 
+function collectHasSegments(has?: HasField) {
+  const hasSegments = new Set<string>();
+
+  for (const hasItem of has || []) {
+    if (!hasItem.value && (hasItem as any).key) {
```
styfle: Can you avoid `as any` here by doing `key in hasItem`?

packages/routing-utils/src/superstatic.ts
```diff
@@ -174,28 +182,60 @@ export function sourceToRegex(
...
   destination: string,
   isRedirect?: boolean
 ): string {
   const parsedDestination = parseUrl(destination, true);
-  delete parsedDestination.href;
-  delete parsedDestination.path;
-  delete parsedDestination.search;
+  delete (parsedDestination as any).href;
+  delete (parsedDestination as any).path;
+  delete (parsedDestination as any).search;
```
styfle, line 223: What changed here 🤔

ijjk: This was showing an error since we're using delete on non-optional fields so casted it to any to remove.
 
Screen Shot 2021-04-01 at 7 10 19 PM

packages/routing-utils/src/superstatic.ts
ijjk: Updated 👍

packages/routing-utils/src/superstatic.ts
styfle: Oh, I'm surprised the build didn't fail before 🤔

packages/routing-utils/src/regexr/expression-lexer.js
```diff
@@ -0,0 +1,952 @@
+/*
```
styfle: Where did this come from? Is it copy/pasta? And if so, why?

ijjk: It's from regexr.com specifically this repo https://github.com/gskinner/regexr it's copied/pasted since it's not available through npm and we only need this specific lexer to gather the named groups from the regex correctly. 

Attempting to gather them without a lexer seems to be un-reliable and the same lexer is being added in Next.js' handling here https://github.com/vercel/next.js/pull/23626

styfle: > it's not available through npm 

@ijjk What about `yarn add --dev https://github.com/gskinner/regexr` so that its added to the lock file?

ijjk: We would need to add a build step if we pull in the source code directly since it's using import/export syntax. We only need the expression lexer so seems better to only pull that in here instead of adding an additional dependency/build step.

I could create a separate repo with just the lexer with the source added here if that's preferable.

styfle: I'm just thinking long term, how do we handle updating this code in case it changes upstream (think bugs, vulnerabilities, etc)?

Forking and publishing to npm won't really solve that problem. But I guess it will solve the problem of shared code between `vercel/vercel` and `vercel/next.js` if we need to modify regexr.

> since it's using import/export syntax

Node.js 12.17.0+ supports ESM without a flag so we could make this work without a build step I believe (not sure how TS will mix ESM with CJS though 🤔 )

dav-is (approved):

packages/routing-utils/src/regexr/expression-lexer.js
ijjk: I updated to use a shared repo here https://github.com/ijjk/regexr-lexer that has a built version of the lexer/profiles included that is added to the lockfile. 

Do we want to ncc the routing-utils files as well or do we need to keep the same output structure?

packages/routing-utils/src/superstatic.ts
```diff
@@ -5,6 +5,11 @@
 import { parse as parseUrl, format as formatUrl } from 'url';
 import { pathToRegexp, compile, Key } from 'path-to-regexp';
 import { Route, Redirect, Rewrite, Header } from './types';
+import { HasField } from '.';
```
styfle: HasField can be moved up to the `./types` import above

packages/routing-utils/src/superstatic.ts
```diff
@@ -174,28 +186,60 @@ export function sourceToRegex(
...
   const namedSegments = segments.filter(name => name !== UN_NAMED_SEGMENT);
 
-  if (namedSegments.length > 0) {
+  if (namedSegments.length > 0 || hasItemSegments.length > 0) {
     const indexes: { [k: string]: string } = {};
     segments.forEach((name, index) => {
       indexes[name] = toSegmentDest(index);
     });
 
+    // 'has' matches override 'source' matches
```
styfle, line 237: I dont see a test for the override source use case, can you add one?

styfle: Oh I see it now: `another/$1/$username/$pathname/$another/$host`

styfle (approved):

packages/routing-utils/package.json
```diff
@@ -14,13 +14,14 @@
   },
   "license": "MIT",
   "scripts": {
-    "prepublishOnly": "tsc",
+    "prepublishOnly": "yarn build",
```
styfle: Can this be reverted to `tsc` now?

packages/routing-utils/src/regexr/expression-lexer.js
styfle: We don't need to use `ncc` here because that will happen downstream, either CLI or API

packages/routing-utils/src/superstatic.ts
ijjk: Updated

packages/routing-utils/package.json
ijjk: Yeah reverted this change

PR #6081, opened 2021-04-08
[cli] Update `vc env` messages

styfle:
This PR addresses feedback from @pacocoursey

### DIFF
--- a/packages/cli/src/commands/env/add.ts
+++ b/packages/cli/src/commands/env/add.ts
@@ -135,7 +135,7 @@ export default async function add(
     const { inputValue } = await inquirer.prompt({
       type: 'input',
       name: 'inputValue',
-      message: `Add ${envName} to which Git Branch? (leave empty for all Preview branches)?`,
+      message: `Add ${envName} to which Git branch? (leave empty for all Preview branches)?`,
     });
     envGitBranch = inputValue || '';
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/env/add.ts
```diff
@@ -135,7 +135,7 @@ export default async function add(
     const { inputValue } = await inquirer.prompt({
       type: 'input',
       name: 'inputValue',
-      message: `Add ${envName} to which Git Branch? (leave empty for all Preview branches)?`,
+      message: `Add ${envName} to which Git branch? (leave empty for all Preview branches)?`,
```
pacocoursey, line 138:
```suggestion
      message: `Add ${envName} to which Git branch? (leave empty for none)?`,
```

Optional, but this wording also covers the cases of no branches (CLI deployments)

pacocoursey (approved):

TooTallNate (approved):

packages/cli/src/commands/env/add.ts
styfle: "none" is not really the right answer. We show "All Preview Branches" in front so I'm going to keep this as is.

Remember that the CLI can add env vars and still do git push after to deploy.

PR #6108, opened 2021-04-13
[cli] Replace "mri" with "arg"

styfle:
Replace a few usages of [mri](https://www.npmjs.com/package/mri) with [arg](https://www.npmjs.com/package/arg) which will ensure that unknown flags will throw instead of skip silently.

This also bumps `arg` to version 5.0.0 which has support for TS types.


mglagola (approved):

packages/cli/package.json
```diff
@@ -107,7 +107,7 @@
     "alpha-sort": "2.0.1",
     "ansi-escapes": "3.0.0",
     "ansi-regex": "3.0.0",
-    "arg": "2.0.0",
+    "arg": "5.0.0",
```
AndyBitz, line 110: Can we remove `mri` from the package.json, or is it used somewhere else?

styfle: @AndyBitz `mri` is still used for `vc secrets` because there is a weird corner case where a secret's value starts with `-` and we can't use `arg`. However, we'll probably deprecate and remove secrets soon.

AndyBitz (approved):

TooTallNate (approved):

PR #6115, opened 2021-04-15
Deprecate Docusaurus v1 in favor of v2 (and update example).

leerob:
The Docusaurus team highly recommends Docusaurus v2 over Docusaurus v1. Let's deprecate v1 so developers don't need to make that decision when starting projects. 

- Removed the `docusaurus` example
- Renamed `docusaurus-2` to `docusaurus`
- Updated code to the latest Docusaurus 2 example

I also updated the frameworks list to consolidate to just one version of Docusaurus. Is this okay or will it have issues that I'm unaware of? I'm not sure if it's used for more than setting up new projects.


styfle:
Why is version 2 still alpha?? 

>Docusaurus v2 is still alpha (since mid-2019) but already quite stable.

https://docusaurus.io/docs

versions

leerob:
Great question. I am not sure. Going to close this for now.

PR #6121, opened 2021-04-16
[node][node-bridge] Add support for ESM

styfle:
### `@vercel/node-bridge`
- [x] Accept both ESM or CJS entrypoint in launcher via `await import('./entrypoint.js')`
- [x] Change launcher from TS to JS due to https://github.com/microsoft/TypeScript/issues/43329
- [x] Use `tsc` to type check JS

### `@vercel/node`
- [x] Disable babel transpilation for `.mjs`
- [x] Disable babel transpilation when `type: "module"` in `package.json`
- [x] Add E2E tests for deployment
- [x] Add test for `vc dev` (follow up PR #6385)
- [x] Add support for `.mjs` with zero config so you can use `/api/users.mjs` (follow up PR #6386)


markozxuu:
Awesome job @styfle !

packages/node/src/dev-server.ts
```diff
@@ -81,8 +81,9 @@ register({
 
 import { createServer, Server, IncomingMessage, ServerResponse } from 'http';
 import { Readable } from 'stream';
-import { Bridge } from './bridge';
-import { getNowLauncher } from './launcher';
+import type { Bridge } from '@vercel/node-bridge/bridge';
+// @ts-ignore - copied to the `dist` output as-is
```
TooTallNate, line 85: Are we able to remove the `@ts-ignore` If we copy over the `launcher.d.ts` file?

TooTallNate (approved):

packages/node/src/dev-server.ts
styfle: It gets tricky because we need to copy over multiple type files.

I'll see if this can be improved in a follow up PR 👍

PR #6142, opened 2021-04-23
Ensure segments from only `has` are replaced

ijjk:
This ensures that when segments are coming from only `has` items we still replace them correctly as they currently don't get replaced if the `source` doesn't have any segments as well. 

### Related Issues

x-ref: https://vercel.slack.com/archives/CLDDX2Y0G/p1619061783470000

#### Tests

#### Code Review

### DIFF
--- a/packages/routing-utils/test/superstatic.spec.js
+++ b/packages/routing-utils/test/superstatic.spec.js
@@ -200,7 +200,19 @@ test('convertRedirects', () => {
     },
     {
       source: '/hello/:first',
-      destination: '/another/:first/:username/:pathname/:another/:host',
+      destination: '/another/:a/:b',
+      has: [
+        {
+          type: 'host',
+          value: '(?<a>.*)\\.(?<b>.*)',
+        },
+      ],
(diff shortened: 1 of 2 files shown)


packages/routing-utils/test/superstatic.spec.js
```diff
@@ -577,7 +642,18 @@ test('convertRewrites', () => {
     },
     {
       check: true,
-      dest: '/another/$1/$username/$pathname/$another/$host',
+      dest: '/another/$a/$b',
```
styfle, line 645: Should this have a querystring with `?first=:first` appended?

styfle (approved):

packages/routing-utils/test/superstatic.spec.js
ijjk: We don't automatically append the params from the source if at least one segment is being used in the `destination` currently so since the `has` segments are being used in the `destination` we don't auto append the `source` segment

dav-is (approved):

PR #6168, opened 2021-04-29
[frameworks] Add placeholder to frameworks' dev command

lucleray:
### Related Issues

Add a more readable `placeholder` to dev commands that contain a value with something like `--port $PORT`.

The goal is to show a readable value for the dev command on the success page of the import flow when one of the framework is imported.

Ref: https://app.clubhouse.io/vercel/story/21632

#### Tests

#### Code Review


packages/frameworks/src/frameworks.ts
```diff
@@ -1383,6 +1402,7 @@ export const frameworks = [
       },
       devCommand: {
         value: 'hugo server -D -w -p $PORT',
+        placeholder: 'hugo server -D',
```
raunofreiberg, line 1405:
```suggestion
        placeholder: 'hugo server -D -w',
```

Should we keep `-w`?

raunofreiberg (approved):

packages/frameworks/src/frameworks.ts
lucleray: I'm not sure 🤔 Here in the quickstart it's not present so I didn't put it: https://gohugo.io/getting-started/quick-start/#step-5-start-the-hugo-server

AndyBitz (approved):

PR #6185, opened 2021-05-05
[cli] Refactor `vc login` command

TooTallNate:
Refactors the CLI `vc login` command with the following features:

* Adds GitHub, GitLab and Bitbucket as login options
* Uses a "list" input to select which login method to use (same list as `vc init`)
* Support connecting SAML Profile to Vercel user during login

Screen Shot 2021-05-05 at 11 02 06 AM
Screen Shot 2021-05-05 at 11 03 13 AM
Screen Shot 2021-05-05 at 11 03 51 AM

### DIFF
--- a/packages/cli/src/commands/login.ts
+++ b/packages/cli/src/commands/login.ts
@@ -1,18 +1,18 @@
-import inquirer from 'inquirer';
 import { validate as validateEmail } from 'email-validator';
 import chalk from 'chalk';
 import hp from '../util/humanize-path';
 import getArgs from '../util/get-args';
-import error from '../util/output/error';
 import handleError from '../util/handle-error';
 import logo from '../util/output/logo';
+import prompt from '../util/login/prompt';
 import doSsoLogin from '../util/login/sso';
-import doEmailLogin from '../util/login/email';
+import doEmailLogin from '../util/login//email';
 import { prependEmoji, emoji } from '../util/emoji';
 import { getCommandName, getPkgName } from '../util/pkg-name';
 import getGlobalPathConfig from '../util/config/global-path';
 import { writeToAuthConfigFile, writeToConfigFile } from '../util/config/files';
 import Client from '../util/client';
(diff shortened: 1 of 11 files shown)


packages/cli/src/util/login/oauth.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const loginError = query.get('loginError');
+    if (loginError) {
+      const err = JSON.parse(loginError);
+      output.prettyError(err);
+      return 1;
+    }
+
+    //console.log(query);
+
```
bl-ue: Did you mean to leave this in?
```suggestion
```

bl-ue (changes requested):

packages/cli/src/commands/login.ts
```diff
@@ -94,21 +73,25 @@ export default async function login(client: Client): Promise<number> {
...
-    result = await doEmailLogin(input, { output, apiUrl });
-  } else if (isValidSlug) {
-    result = await doSsoLogin(input, { output, apiUrl });
+  const params: LoginParams = { output, apiUrl };
+
+  if (input) {
+    // Email or Team slug was provided via command line
+    if (validateEmail(input)) {
+      result = await doEmailLogin(input, params);
+    } else if (isValidSlug(input)) {
```
styfle: Lets simplify this with `!input.includes('@')`

packages/cli/src/util/login/bitbucket.ts
```diff
@@ -0,0 +1,11 @@
+import { URL } from 'url';
+import { LoginParams } from './types';
+import doOauthLogin from './oauth';
+
+export default function doBitbucketLogin(params: LoginParams) {
+  const url = new URL(
+    '/api/registration/bitbucket/connect',
+    'https://vercel.com'
```
styfle, line 10: Lets leave a comment here mentioning why we cant use `apiUrl` like we do for others

styfle (approved):

PR #6217, opened 2021-05-10
[cli] Recommend `vc env` instead of `vc secrets`

styfle:
We no longer need `vercel secrets` command. The former use cases can be solved with `vercel env`.

This adds a note recommending to use `vc env` as an alternative.

Screen Shot 2021-08-11 at 1 38 18 PM
Screen Shot 2021-08-11 at 1 41 06 PM


TooTallNate (approved):

packages/cli/src/commands/secrets.js
```diff
@@ -253,7 +258,7 @@ async function run({ output, token, contextName, currentTeam, client }) {
   }
 
   if (subcommand === 'rename') {
-    output.warn(
+    output(
```
styfle:
```suggestion
    output.note(
```

nkzawa (approved):

packages/cli/src/commands/secrets.js
```diff
@@ -199,6 +211,11 @@ async function run({ output, token, contextName, currentTeam, client }) {
   }
 
   if (subcommand === 'rm' || subcommand === 'remove') {
+    output.note(
+      `The ${getCommandName(
+        'env rm'
```
AndyBitz, line 211: This makes it sound like the commands are interchangeable, but I don't think it's possible to delete secrets with env. So this might be a bit confusing

AndyBitz (approved):

leerob (approved):
Do you think we need to add a link explaining _why_ they're recommended?

markozxuu:
> Do you think we need to add a link explaining _why_ they're recommended?

I seem a good option, add a link to explain the because `vc env` is better instead of the `vc secret` 👍🏼

ktcarter (approved):

packages/cli/src/commands/secrets.js
AndyBitz: Maybe it'd make sense to mention that `vc secrets` is a deprecated flow in general and that `vc env` is recommended.

markozxuu: To attach a link would be a good idea @AndyBitz

anatrajkovska:
Is this PR still relevant?

styfle:
@anatrajkovska Yes, this is still needed. Someone needs to take on this [roadmap item](https://www.notion.so/vercel/Deprecate-Secrets-37cba3e0ff73467cb3ff86f345db4980) and make a formal announcement so users are aware and we need to come up with a sunset date.

PR #6257, opened 2021-05-19
[cli] Refactor `vc switch` command

TooTallNate:
Refactors the `vc switch` command to more closely match how the team picker works on the Vercel dashboard:

 * Converts to TypeScript
 * Adds separator between user scope and team scopes
 * Sorts teams by name
 * No longer places the current scope at the top of the list (but the current scope is still selected by default)

Screen Shot 2021-05-20 at 11 18 12 AM


packages/cli/src/commands/teams/switch.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  output.spinner('Fetching teams information');
+  const [user, teams] = await Promise.all([getUser(client), getTeams(client)]);
+
+  const currentTeam = personalScopeSelected
+    ? undefined
+    : teams.find(team => team.id === config.currentTeam);
+
+  if (!personalScopeSelected && !currentTeam) {
+    output.error(`You are not a part of the current team anymore`);
```
AndyBitz:
```suggestion
    output.error(`You are not a part of the current team anymore.`);
```

packages/cli/src/commands/teams/switch.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  if (!desiredSlug) {
+    const teamChoices = teams
+      .slice(0)
+      .sort((a, b) => {
+        return a.name.toLowerCase().localeCompare(b.name.toLowerCase());
+      })
+      .map(({ id, slug, name }) => {
+        let title = `${slug} (${name})`;
+        const selected = id === currentTeam?.id;
```
AndyBitz: If we show the slug first in the list, wouldn't it make sense to also sort by it instead of by name?

AndyBitz (approved):

packages/cli/src/commands/teams/switch.ts
TooTallNate: I think we should show name first. I'll swap.

PR #6263, opened 2021-05-21
[cli] Add SAML reauthentication logic when using different scope

TooTallNate:
When the API returns a SAML error, then show the proper reauthentication prompt depending on the scope being requested:

Team with SAML enforced, shows only SSO login option:

Screen Shot 2021-05-24 at 1 50 29 PM

Team with SAML enabled, but not enforced, prompts with all login methods:

Screen Shot 2021-05-24 at 1 50 36 PM

Team without SAML enabled, or User scope, shows prompt with SSO option removed:

Screen Shot 2021-05-24 at 1 50 44 PM

[ch21964]

### DIFF
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -121,7 +121,6 @@ export default class Client extends EventEmitter {
 
       if (!res.ok) {
         const error = await responseError(res);
-        console.log(error);
 
         if (isSAMLError(error)) {
           // A SAML error means the token is expired, or is not
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -5,19 +5,14 @@ import { VercelConfig } from '@vercel/client';
 import retry, { RetryFunction, Options as RetryOptions } from 'async-retry';
 import fetch, { BodyInit, Headers, RequestInit, Response } from 'node-fetch';
 import ua from './ua';
-import { APIError } from './errors-ts';
 import { Output } from './output/create-output';
 import responseError from './response-error';
 import printIndications from './print-indications';
 import reauthenticate from './login/reauthenticate';
+import { SAMLError } from './login/types';
 import { writeToAuthConfigFile } from './config/files';
 import { AuthConfig, GlobalConfig, JSONObject } from '../types';
 import { sharedPromise } from './promise';
-
-interface SAMLError extends APIError {
-  saml: true;
-  teamId: string | null;
-}
 
 const isSAMLError = (v: any): v is SAMLError => {
   return v && v.saml;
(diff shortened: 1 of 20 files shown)


AndyBitz (approved):

packages/cli/src/util/client.ts
```diff
@@ -78,34 +85,30 @@ export default class Client extends EventEmitter {
       }
 
       _url = `${apiUrl}${parsedUrl.pathname}?${query}`;
-
-      delete opts.useCurrentTeam;
-      delete opts.accountId;
     }
 
-    if (opts.json !== false && opts.body && typeof opts.body === 'object') {
```
styfle, line 86: Do we no longer support `opts.json: false`?

TooTallNate: It's supported for making the return value be `res` instead of the parsed response body.

But for the request body it's not necessary since the `isJSONObject()` check is in place. The [two](https://github.com/vercel/vercel/blob/bf3a1fc832f0d131772f1728950c20c9ce2b6275/packages/cli/src/util/dns/import-zonefile.ts#L31) [places](https://github.com/vercel/vercel/blob/930063638c4ddeb5ed475269a5133d39e040a182/packages/cli/src/util/events.ts) where `json: false` is currently being used are not passing in an object body anyways.

packages/cli/src/util/client.ts
```diff
@@ -140,19 +135,31 @@ export default class Client extends EventEmitter {
...
-        ? res.json()
-        : res;
+      return contentType.includes('application/json') ? res.json() : res;
     }, opts.retry);
   }
 
-  _onRetry(error: Error) {
-    this.output.debug(`Retrying: ${error}\n${error.stack}`);
-  }
+  reauthenticate = sharedPromise(async function (
```
styfle, line 147: This sharedPromise() is an interesting solution. I wonder if it can be solved with `async-sema` of 1 since we already have that

TooTallNate: Perhaps. I kinda like the current approach though since it doesn't need any dependencies.

packages/cli/src/util/login/reauthenticate.ts
```diff
@@ -0,0 +1,30 @@
...
+  let result: string | number = 1;
+  if (error.teamId && error.enforced) {
+    // If team has SAML enforced then trigger the SSO login directly
+    params.output.log(
+      `You must re-authenticate with SAML to use ${highlight(
+        error.scope
+      )} scope.`
+    );
+    if (await confirm(`Log in with SAML?`, true)) {
+      result = await doSsoLogin(params, error.teamId);
```
styfle, line 20: Is it okay if `error.teamId` is undefined?

TooTallNate: It's not going to be. See the `if` branch on line 12.

styfle (approved):

PR #6272, opened 2021-05-24
[python] Fix installRequirement to use explicit version

styfle:
This PR pins the exact version of dependencies to avoid breaking changes.

- [Werkzeug](https://pypi.org/project/Werkzeug/#history)
- [pipfile-requirements](https://pypi.org/project/pipfile-requirements/#history)

### DIFF
--- a/packages/python/src/index.ts
+++ b/packages/python/src/index.ts
@@ -86,6 +86,7 @@ export const build = async ({
 
   await installRequirement({
     dependency: 'werkzeug',
+    version: '1.0.1',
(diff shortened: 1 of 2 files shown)


AndyBitz (approved):

coolreader18:
This breaks Flask 2.0 - it depends on werkzeug 2, which doesn't actually get installed when installing from requirements.txt b/c werkzeug 1 was already installed. Should specifically `installRequirementsFile` pass `--upgrade`?

PR #6276, opened 2021-05-24
[cli] Fix `vc logout` retry

styfle:
This PR fixes flaky networks like GH Actions https://github.com/vercel/vercel/runs/2658182610#step:11:213

We moved the API call before the config file write so that we can utilize `client.fetch()` which already does retry.

### DIFF
--- a/packages/cli/src/commands/logout.ts
+++ b/packages/cli/src/commands/logout.ts
@@ -1,5 +1,4 @@
 import chalk from 'chalk';
-import fetch from 'node-fetch';
 import logo from '../util/output/logo';
 // @ts-ignore
 import { handleError } from '../util/error';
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/logout.ts
```diff
@@ -76,25 +89,10 @@ export default async function main(client: Client): Promise<number> {
...
-
-  if (res.status === 403) {
-    output.debug('Token is invalid so it cannot be revoked');
-  } else if (res.status !== 200) {
-    const err = await res.json();
-    output.error('Failed to revoke token');
-    output.debug(err ? err.message : '');
-    return 1;
+    output.debug(err?.message ?? '');
+    exitCode = 1;
```
TooTallNate, line 91: I think we want to return early still here, otherwise "Logged out!" success message is gonna be printed.

TooTallNate (changes requested):

packages/cli/src/commands/logout.ts
styfle: @TooTallNate We can't return early because we need to make sure that the file IO happens even if the API is down so that we don't ever prevent users from logging out (I assume thats why it was reversed before)

styfle: @TooTallNate I see what you mean now. I updated the "Logged out" message so it will not print unless the exit code is `0`

TooTallNate (approved):

PR #6277, opened 2021-05-24
[cli] Set `tokenName` query param for Git provider login methods

TooTallNate:
We're not currently setting the `tokenName` when logging in via a Git provider, so the name becomes the default "Website" one which is confusing / incorrect.


AndyBitz (approved):

pacocoursey (approved):

styfle (approved):
Thanks 👍

PR #6280, opened 2021-05-25
[cli] Reauthenticate scopes with limited access in `vc switch`

TooTallNate:
In the `vc switch` command, if your current access token results in "limited" Team information being returned, then show a lock emoji next to the team/user name in the select input. When a locked scope is selected, then pre-emptively prompt the user to re-authenticate using a valid login method in relation to the desired scope.

https://user-images.githubusercontent.com/71256/119441172-87abae80-bcda-11eb-801a-cb6837bae353.mov

[ch21964]


AndyBitz (approved):

packages/cli/src/types.ts
```diff
@@ -62,18 +62,26 @@ export type User = {
...
   billing: Billing;
   created: string;
   creatorId: string;
   membership: { uid: string; role: 'MEMBER' | 'OWNER'; created: number };
   name: string;
   slug: string;
-};
+  limited?: boolean;
+  saml?: {
+    enforced: boolean;
```
styfle, line 79: Should `enforced` be optional too?

packages/cli/src/util/client.ts
```diff
@@ -151,7 +152,9 @@ export default class Client extends EventEmitter {
     const result = await reauthenticate(this, error);
 
     if (typeof result === 'number') {
-      this.output.prettyError(error);
+      if (error instanceof APIError) {
+        this.output.prettyError(error);
+      }
```
styfle, line 162: Do we need to print a generic error message if its not `APIError`? Otherwise it will exit without telling the user anything?

packages/cli/src/types.ts
TooTallNate: No, it's always present if the `saml` object exists.

packages/cli/src/util/client.ts
TooTallNate: Good idea.

styfle (approved):

PR #6299, opened 2021-05-28
[build-utils] fix glob result path for windows paths

sokra:
### Related Issues

This fixes serving static files from subdirectories on windows

#### Tests

#### Code Review


packages/build-utils/src/fs/glob.ts
```diff
@@ -62,7 +62,7 @@ export default async function glob(
         finalPath = path.join(mountpoint, finalPath);
       }
 
-      results[finalPath] = new FileFsRef({ mode: stat.mode, fsPath });
+      results[finalPath.replace(/\\/g, '/')] = new FileFsRef({ mode: stat.mode, fsPath });
```
AndyBitz, line 63: It seems to be possible to create folders with a `\` in the name on MacOS and Linux, so this might have unexpected side effects in those cases, maybe it should only be applied in case the OS is Win?

AndyBitz (changes requested):

packages/build-utils/src/fs/glob.ts
sokra: There is something similar a few lines higher:

```js
const fsPath = path.join(options.cwd!, relativePath).replace(/\\/g, '/');
```

So I guess folders with `\` in the name would break in many many places anyway...

And I think you can't have `\` in urls?

styfle: Lets add a test to confirm this works as expected (we run unit tests on windows/mac/linux)

styfle:
> Currently the integration tests doesn't run on windows?

Indeed, these are typically deployments so its linux only.

However, you can add a unit test which will run on windows/mac/linux.

styfle (approved):

cb1kenobi:
I just tested a static app with nested pages on Windows and it works as expected. Is there a specific reproduction case that would indicate this is still an issue?

styfle:
Closing since there is no reproduction or test

PR #6322, opened 2021-06-04
[cli] Fix `vc logout` command when using Team scope

TooTallNate:
Fixes logout command not working when switched to a Team scope:

```
$ vc login
$ vc switch $some_team
$ vc logout
Failed during logout
```

### DIFF
--- a/packages/cli/src/commands/logout.ts
+++ b/packages/cli/src/commands/logout.ts
@@ -1,6 +1,5 @@
 import chalk from 'chalk';
 import logo from '../util/output/logo';
-// @ts-ignore
 import { handleError } from '../util/error';
 import { writeToConfigFile, writeToAuthConfigFile } from '../util/config/files';
 import getArgs from '../util/get-args';
@@ -62,6 +61,7 @@ export default async function main(client: Client): Promise<number> {
   try {
     await client.fetch(`/v3/user/tokens/current`, {
       method: 'DELETE',
+      useCurrentTeam: false,
     });
   } catch (err) {
     if (err.status === 403) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/types.ts
```diff
@@ -16,7 +16,7 @@ export interface JSONObject {
 }
 
 export interface AuthConfig {
-  token: string;
+  token?: string;
```
TooTallNate, line 19: VS Code was complaining (though the version of TypeScript we use to build doesn't seem to mind).

packages/cli/src/commands/logout.ts
```diff
@@ -1,6 +1,5 @@
 import chalk from 'chalk';
 import logo from '../util/output/logo';
-// @ts-ignore
```
TooTallNate, line 3: Apparently not necessary 🤷

AndyBitz (approved):

packages/cli/src/commands/logout.ts
markozxuu: Right, the comment `@ts-ignore` is intended to ignore TS errors (**not recommend**), If what I code is well, not necessary that comment : -)

packages/cli/src/types.ts
styfle: @TooTallNate You can tell VS Code to use the same version as the build by clicking the version in the bottom right of VS Code > Select TypeScript Version > Use Workspace Version.

https://stackoverflow.com/a/59076553/266535

styfle (approved):

packages/cli/src/types.ts
TooTallNate: Oh, nice TIL.

PR #6325, opened 2021-06-04
[cli] Convert `vc alias` to TS

markozxuu:
This PR focuses on rewrite all commands from `alias` to typescript. I have made some minor changes, to make the code easier to read.

- The code was simplified the file `rs`
- Remove unnecessary instances of `Now`, is better to use the `Client` class

## Minor changes 

- feat: Remove parameter generic, "null" for the fetch
- feat(cli): Add `PaginationOptions` type to `pagination`

#### Tests

#### Code Review


packages/cli/src/util/alias/find-alias-by-alias-or-id.ts
```diff
@@ -1,21 +1,15 @@
...
   );
 }
-
-function getSafeAlias(alias: string) {
-  return alias
-    .replace(/^https:\/\//i, '')
-    .replace(/^\.+/, '')
-    .replace(/\.+$/, '')
-    .toLowerCase();
-}
```
styfle: Any reason why this function was moved? It seems like this function is only used in this file.

packages/cli/src/commands/alias/ls.ts
```diff
@@ -73,21 +72,20 @@ export default async function ls(client, opts, args) {
...
-      ['source', 'url', 'age'].map(h => chalk.gray(h)),
+      ['source', 'url', 'age'].map(header => chalk.gray(header)),
       ...aliases.map(a => [
         // for legacy reasons, we might have situations
         // where the deployment was deleted and the alias
         // not collected appropriately, and we need to handle it
         a.deployment && a.deployment.url ? a.deployment.url : chalk.gray('–'),
         a.alias,
-        ms(Date.now() - new Date(a.createdAt)),
+        ms(Date.now() - new Date(a.created).getTime()),
```
styfle: This should remain `createdAt`

packages/cli/src/commands/alias/rm.ts
```diff
@@ -61,7 +60,10 @@ export default async function rm(client, opts, args) {
     return 1;
   }
 
-  const alias = await findAliasByAliasOrId(output, now, aliasOrId);
+  const alias = await findAliasByAliasOrId(output, client, aliasOrId);
+
+  const { alias: aliasUrl } = alias;
```
styfle: This is going to cause bugs because `alias` can be falsy as seen on the line below. Let's revert this renaming.

markozxuu: Done ! Change made 👍

packages/cli/src/util/alias/find-alias-by-alias-or-id.ts
markozxuu: The intention was reduced file size. Making more modular.

packages/cli/src/commands/alias/ls.ts
markozxuu: The problem is `createdAt` does not exist in the `Alias` type. TS complained earlier because createdAt doesn't exist. What we can do are 2 things.

1.- Create `createdAt` property in the `Alias` type
2.- We could rename `created` to` createdAt `function scope

markozxuu: What do you think @styfle  ? 🤔

styfle: It sounds like the type is wrong. It should have `createdAt` and thats what the code was using before

packages/cli/src/util/alias/find-alias-by-alias-or-id.ts
styfle: Let's revert it back to how it was

markozxuu: Ok 👍

styfle (approved):
Great work, thanks! 🎉

packages/cli/src/util/alias/find-alias-by-alias-or-id.ts
```diff
@@ -1,17 +1,17 @@
...
+
 import Client from '../client';
 
 export default async function findAliasByAliasOrId(
   output: Output,
   client: Client,
   aliasOrId: string
 ) {
-  return client.fetch<Alias | null>(
+  return client.fetch<Alias>(
```
styfle:
```suggestion
  return client.fetch<Alias | null>(
```

I think this can return null if the alias was not found

styfle: Oh nevermind, if we get a 200 it should never be null 👍

styfle: Yeah I checked and the early version of this API returns 204 status with null body when not found

markozxuu: Thanks for checking this :)

PR #6328, opened 2021-06-05
[test] Fix corrupt bmp test

styfle:
The previous test was using a corrupt bmp file so it was a coincidence that next dev performed the same as prod.

We will disable this test until `next dev` supports converting bpm to webp.


jhoch (approved):

jhoch (approved):

styfle:
Hmm, on second thought maybe the corrupt bmp was intentional to trigger the bypass 🤔 

https://github.com/vercel/vercel/blob/6eea26c39edd8ac911c9a74096b584854a280bf4/packages/cli/test/dev/integration.js#L1713

With the updated bmp, we're getting `next dev` bypass and the deployment is actually converting to webp.

styfle:
Since the test was relying on undefined behavior, I updated the accept header to ensure it bypasses to serve the original bmp image.

ijjk (approved):

PR #6359, opened 2021-06-14
[cli] Fix pathing for passthrough urls on vite

t3dotgg:
I am not sure if the node `url.format` is necessary, but it was breaking here. Transformed paths of style `/src/test.svg?import` to `/src/test.svg?import=`, which is a breaking url modification for vite

I have requested that vite is modified to handle these query params better in the future. Not sure how y'all feel about using this fix for now https://twitter.com/t3dotgg/status/1404287709393276936

### Related Issues

> Fixes https://github.com/vercel/vercel/discussions/6356 , https://github.com/vercel/vercel/discussions/6145 , https://github.com/vitejs/vite/issues/3781

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1621,7 +1621,7 @@ export default class DevServer {
         delete origUrl.search;
         origUrl.pathname = dest;
         Object.assign(origUrl.query, uri_args);
-        req.url = url.format(origUrl);
+        req.url = origUrl.path;
         return proxyPass(req, res, upstream, this, nowRequestId, false);
       }
 


t3dotgg:
Just finished up a blog post about Vite + Vercel. Might be worth a read if y'all want to streamline the process a little more :)

https://t3.gg/blog/posts/vite-vercel

styfle (commented):
Thanks for the PR!

Is there a test fixture we can add to [./cli/test/dev/integration.js](https://github.com/vercel/vercel/blob/main/packages/cli/test/dev/integration.js) so we won't regress?

TooTallNate:
Thanks for the PR! We definitely want to support Vite on Vercel, including with `vc dev`.

FYI, the reason that `url.format()` is used here is so that we can append query string parameters that get defined in i.e. the `routes`/`rewrites` configurations of the Project, so this PR as-is would break that.

But I understand that the difference between `?import` and `?import=` is breaking Vite. Perhaps we'll need to find an alternative query-string formatting function that preserves the missing `=` sign, but ultimately we _do_ want to keep the appending of query string parameters intact.

t3dotgg:
Really appreciate the feedback, thank you both @styfle and @TooTallNate !

Sadly I'm unable to get the repo to build or run tests on my machine (Apple Silicon Mac). Guessing it's some weird golang ARM issue, but it's made it very hard to test my changes :(

I validated this solution by modifying the `vercel` cli JS directly and confirmed that it worked.  Knowing that this is breaking for other flows, 100% agree we need a better formatter or a patch on the Vite side.

Not sure what the best path is from here. I'm running a patch-package on my end for now. Happy to investigate any other formatters or solutions you think of :)

Ty again!

Build error from my machine for reference:
image

t3dotgg:
Beginning to feel more and more like Vite is the right place to fix this. `field=value` is not mandated per RFC spec, but it is a common enough assumption to leak into node's url package

Made a PR adding a value on vite's side: https://github.com/vitejs/vite/pull/3805

TooTallNate:
@TheoBr [This](https://github.com/vercel/vercel/pull/6361) is related to the package failing to build for you.

I agree that a fix in Vite seems appropriate in this case. But also still wouldn't mind having `vc dev` be more "transparent" and preserve the original URL structure. Possibly we can use the [`qs`](https://github.com/ljharb/qs#handling-of-null-values) module instead of the core `url` module, since it appears to have support for no equal sign via `null`.

styfle:
How about using `new URL()`?

TooTallNate:
@styfle Good idea, but appears to have the same issue as the `url` module, since once you modify the `searchParams`, the `URL` loses the `=` context as well:

```js
> u = new URL('http://a.com/p?import')

// good
> u.href
'http://a.com/p?import'

// modify
> u.searchParams.set('a', 'b')

// bad
> u.href
'http://a.com/p?import=&a=b'
```

t3dotgg:
> @styfle Good idea, but appears to have the same issue as the `url` module, since once you modify the `searchParams`, the `URL` loses the `=` context as well:
> 
> ```js
> > u = new URL('http://a.com/p?import')
> 
> // good
> > u.href
> 'http://a.com/p?import'
> 
> // modify
> > u.searchParams.set('a', 'b')
> 
> // bad
> > u.href
> 'http://a.com/p?import=&a=b'
> ```

Good catch. Vite often has additional query parameters after the `import`, so IMO it is a change that should be implemented there. Shared your comment on my PR - thank you again :)

t3dotgg:
Approved on Vite side, going to close this :) ty again!

https://github.com/vitejs/vite/pull/3805

leerob:
Btw, @TheoBr https://vercel.com/changelog/vite-projects-can-now-be-deployed-with-zero-configuration 😄

PR #6361, opened 2021-06-15
[cgi] Remove `@vercel/cgi` Runtime

TooTallNate:
This Runtime is very old, outdated, unmaintained, and has never been documented.

Additionally, it no longer compiles with the latest version of Go. So rather than fixing it, let's just remove it since we don't want to invest any more time into it.


styfle (approved):

t3dotgg:
Confirmed this fixes my build issues - ty!! 🎉

PR #6362, opened 2021-06-15
Remove `utils/go`

TooTallNate:
This code now lives in its own repository, so the code in this repo is not being used.

https://github.com/vercel/go-bridge


styfle (approved):
Thanks!

AndyBitz (approved):

PR #6382, opened 2021-06-22
[cli] Convert `vc deploy` to TS

markozxuu:
This PR focuses on rewrite all commands from `deploy` to TypeScript. The `deploy` command is one of the most important that can be found in everything that is the CLI, so I took the task of passing it to typescript so that it is much safer regarding the types 😄 

#### Tests

#### Code Review


styfle:
The build is failing with a TS error:

```
TS2345: Argument of type 'VercelConfig | null' is not assignable to parameter of type 'VercelConfig'.
```

styfle (changes requested):
Let's avoid using `any` so we can use the proper types

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -53,10 +53,12 @@ import validatePaths, {
...
 import { readLocalConfig } from '../../util/config/files';
-import { getCommandName } from '../../util/pkg-name.ts';
-import { getPreferredPreviewURL } from '../../util/deploy/get-preferred-preview-url.ts';
+import { getCommandName } from '../../util/pkg-name';
+import { getPreferredPreviewURL } from '../../util/deploy/get-preferred-preview-url';
+import { Output } from '../../util/output';
+import Client from '../../util/client';
 
-const addProcessEnv = async (log, env) => {
+const addProcessEnv = async (log: (str: string) => void, env: any) => {
```
markozxuu: @styfle The `env` parameter it is complicated of type because it changes type during code execution 🤔

TooTallNate: You can probably use `typeof process.env` instead of `any`.

markozxuu: Thank You, @TooTallNate ! `typeof process.env` I was the correct type 👍

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -553,12 +553,8 @@ export default async function main(
       client
     );
 
-    if (
-      deployment instanceof Error &&
-      deployment.code === 'missing_project_settings'
-    ) {
+    if (deployment.code === 'missing_project_settings') {
```
markozxuu, line 563: @TooTallNate @styfle  Remove `deploymente instanceof Error &&`, the current check should be enough 👍

packages/cli/src/commands/deploy/args.ts
```diff
@@ -98,9 +99,9 @@ export const args = {
   '--with-cache': Boolean,
   '--public': Boolean,
   '--no-clipboard': Boolean,
-  '--env': [String],
-  '--build-env': [String],
-  '--meta': [String],
+  '--env': String,
+  '--build-env': String,
+  '--meta': String,
```
styfle: These need to remain arrays because the user can pass the flag multiple times.

For example:

```sh
vercel --env TAG=CANARY --env VERSION=1
```

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -575,15 +587,16 @@ export default async function main(
     }
 
     if (deployment instanceof NotDomainOwner) {
-      output.error(deployment);
+      // Return a message error deployment
```
styfle: Let's remove the comment

packages/cli/src/util/deploy/create-deploy.ts
```diff
@@ -84,7 +89,7 @@ export default async function createDeploy(
     if (error.code === 'cert_missing') {
       const result = await generateCertForDeploy(
         output,
-        now,
+        client!,
```
styfle: This might be a bug. What about if `client` is undefined?

packages/cli/src/util/index.js
```diff
@@ -28,6 +29,7 @@ export default class Now extends EventEmitter {
   }) {
     super();
 
+    this.url = url;
```
styfle, line 32: How was this working before?? 🤔

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -85,19 +90,26 @@ const addProcessEnv = async (log, env) => {
...
     aliasError,
     target,
     indications,
     url: deploymentUrl,
+  }: {
+    readyState: string;
+    alias: string[];
+    aliasError: Error;
+    target: string;
+    indications: any;
```
styfle, line 107: I think this should be `indications?: { type: string, payload: string, link?: string, action:?: string }[]`

packages/cli/src/commands/deploy/index.ts
```diff
@@ -42,9 +43,13 @@ export default async client => {
...
   if (!localConfig || localConfig instanceof Error) {
-    localConfig = readLocalConfig(paths[0]);
+    const maybeConfig = readLocalConfig(paths[0]);
+    if (maybeConfig) {
+      localConfig = maybeConfig;
+    }
   }
-  const stats = {};
+
+  const stats: any = {};
```
styfle: This shouldn't be `any`. Although it might be removed in PR #6388 soon.

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -176,7 +189,7 @@ const printDeploymentStatus = async (
 };
 
 // Converts `env` Arrays, Strings and Objects into env Objects.
-const parseEnv = env => {
+const parseEnv = (env: any | object) => {
```
styfle, line 192: I think the type here should be `env: Env | undefined` which you can find in `@vercel/build-utils`

markozxuu: @styfle I get the following error. I think that is expect an array 🤔 
Screen Shot 2021-06-23 at 12 13 06

packages/cli/src/util/deploy/create-deploy.ts
markozxuu: I don't have idea, this error is generated when the file transform to TypeScript 🤔

packages/cli/src/util/index.js
markozxuu: The entire `deployment` object was passed, I don't know exactly what property it took for this to work. What I did was create a new property for that case 👍

packages/cli/src/commands/deploy/latest.ts
styfle: Perhaps its `Env | string | undefined` then?

packages/cli/src/util/deploy/create-deploy.ts
styfle: So what about `if (!client) throw new Error ('Expected client but found undefined')`

markozxuu: I agree, this solution It seems to work @styfle. I think that approach is the most appropriate for checking In If case `client` is that to `undefined`  👍

packages/cli/src/commands/deploy/latest.ts
markozxuu: @styfle It seems not working. Too, I tried with this variation `Env | string | string[] | undefined`

markozxuu: What can we do in this case ? 🤔

packages/cli/src/commands/deploy/index.ts
markozxuu: I see that the PR is merged, but how can I solve this now in my PR ? cc @TooTallNate  @styfle

packages/client/src/types.ts
```diff
@@ -102,6 +102,7 @@ export const fileNameSymbol = Symbol('fileName');
 export interface VercelConfig {
   [fileNameSymbol]?: string;
   name?: string;
+  meta?: string;
```
styfle: I think meta should be `Dictionary<string>` or `string[]`

packages/cli/src/commands/deploy/index.ts
```diff
@@ -20,7 +21,31 @@ export default async client => {
   let argv = null;
 
   try {
-    argv = getArgs(client.argv.slice(2), args);
+    argv = getArgs(client.argv.slice(2), {
```
styfle, line 19: Instead of inlining this code, you can append `as const` after the `args` object on line 121.

packages/client/src/types.ts
markozxuu: Done, Thank You. Opt for `string[]`

packages/cli/src/commands/deploy/index.ts
markozxuu: For some reason that not understand, When I opt for your solution this happens @styfle 

Screen Shot 2021-06-28 at 9 53 29

Log error

`
Argument of type '{ '--force': BooleanConstructor; '--with-cache': BooleanConstructor; '--public': BooleanConstructor; '--no-clipboard': BooleanConstructor; ... 15 more ...; '--target': StringConstructor; }' is not assignable to parameter of type 'Spec'.
  Property ''--env'' is incompatible with index signature.
    Type 'StringConstructor[]' is not assignable to type 'string | Handler | [Handler]'.
      Type 'StringConstructor[]' is not assignable to type '[Handler]'.
        Target requires 1 element(s) but source may have fewer.ts(2345)
`

markozxuu:
Close PR for I make priority to this #6413

PR #6388, opened 2021-06-23
[cli] Make `vc deploy` not load current "scope"

TooTallNate:
`vc deploy` doesn't care about the current "scope" that the CLI has selected, since it uses the `.vercel` directory to determine the ownerId and projectId.

Therefore, it should not be fetching the team details of the selected scope during `vc deploy`, since it's possible that the token might not have access to the current scope and returns a 403, even though that API call is unnecessary for the `vc deploy` command to complete.


styfle (approved):
Looks good but the test are failing

TooTallNate:
Tests failing was related to an npm outage last night, I believe.

PR #6396, opened 2021-06-24
[frameworks] Add missing cachePattern to Next.js

styfle:
Follow up to #6395

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -104,6 +104,7 @@ export const frameworks = [
     devCommand: 'next dev --port $PORT',
     buildCommand: 'next build',
     getOutputDirName: async () => 'public',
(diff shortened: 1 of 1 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -104,6 +104,7 @@ export const frameworks = [
     devCommand: 'next dev --port $PORT',
     buildCommand: 'next build',
     getOutputDirName: async () => 'public',
+    cachePattern: '.next/cache/**',
```
ijjk, line 107: Isn't this handled in the runtime in `prepareCache` already or do we need it here as well?

styfle: It is but we need a way to document it and this pattern like we have for the other frameworks here

ijjk (approved):

PR #6397, opened 2021-06-24
Update to TypeScript 4.3.4

TooTallNate:
Also updates eslint and related packages to work with the new TypeScript version.


AndyBitz (approved):

packages/cli/src/util/event-listener-to-generator.ts
```diff
@@ -1,15 +0,0 @@
-import { EventEmitter } from 'events';
```
TooTallNate, line 1: File wasn't being used anywhere, and was causing a build error on latest TypeScript.

ijjk (approved):

styfle (approved):
Thanks!

styfle:
@TooTallNate Looks like eslint needs to be updated as well since CI is complaining

styfle (approved):

PR #6412, opened 2021-06-29
[cli] Use client-generated `verifyToken` during OAuth login

TooTallNate:
This PR simplifies the OAuth login logic by removing the localhost HTTP server that waits for the user to complete the OAuth login in the web browser. The localhost server technique does not work when the CLI is in fact not runnning on localhost (i.e. in headless environments such as Docker or SSH). So instead, the CLI will pre-generate the verify token that it will poll against to verify completion of the OAuth login. Once the provided `verifyToken` is "confirmed" by Vercel, then the CLI will exchange the verify token for the actual auth token.

This also removes the `ssoUserId` linking logic, since that will happen all browser-side with this change.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -66,6 +66,7 @@
     "@vercel/node": "1.11.2-canary.2",
     "@vercel/python": "2.0.4",
     "@vercel/ruby": "1.2.6",
+    "uid-promise": "1.1.0",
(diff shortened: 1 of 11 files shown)


TooTallNate:
We're not gonna go with this client-side generated verifyToken route. See https://github.com/vercel/vercel/pull/6455 for an alternative solution.

PR #6413, opened 2021-06-29
[cli] Convert `vc deploy` to TypeScript

TooTallNate:
This is a continuation of https://github.com/vercel/vercel/pull/6382 (thank you @markozxuu!) with merge conflicts addressed and pushed to a local branch, so that CI tests run.


markozxuu:
With `test-unit` this did not happen. Any idea of why this now these failing ? cc @styfle

packages/cli/src/commands/deploy/args.ts
```diff
@@ -1,13 +1,14 @@
...
-import { getPkgName } from '../../util/pkg-name.ts';
+import { getPkgName } from '../../util/pkg-name';
 
 export const help = () => `
   ${chalk.bold(`${logo} ${getPkgName()}`)} [options] <command | path>
 
   ${chalk.dim('Commands:')}
 
     ${chalk.dim('Basic')}
+
```
styfle: Was this newline intentional?

packages/cli/src/commands/deploy/index.ts
```diff
@@ -79,5 +105,30 @@ export default async client => {
...
+    '-b': '--build-env',
+    '-C': '--no-clipboard',
+    '-m': '--meta',
+    '-c': '--confirm',
+
+    // deprecated
+    '--name': String,
+    '-n': '--name',
+    '--target': String,
+  });
```
styfle: Is there anyway to avoid duplicating these args?

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -176,7 +189,7 @@ const printDeploymentStatus = async (
 };
 
 // Converts `env` Arrays, Strings and Objects into env Objects.
-const parseEnv = env => {
+const parseEnv = (env: any | object) => {
```
styfle:
```suggestion
const parseEnv = (env?: string[] | { [key: string]: string }) => {
```

packages/cli/src/commands/deploy/latest.ts
```diff
@@ -500,13 +519,13 @@ export default async function main(client, paths, localConfig, args) {
...
 
-  const currentTeam = org.type === 'team' ? org.id : undefined;
+  const currentTeam = org?.type === 'team' ? org.id : undefined;
   const now = new Now({ apiUrl, token, debug: debugEnabled, currentTeam });
   let deployStamp = stamp();
   let deployment = null;
 
   try {
-    const createArgs = {
+    const createArgs: any = {
```
styfle, line 519: I think we could add a type here and use that in createDeploy() too

packages/cli/src/util/deploy/create-deploy.ts
```diff
@@ -3,17 +3,23 @@ import * as ERRORS_TS from '../errors-ts';
...
-) {
+  output: Output,
+  now: Now,
+  contextName: string,
+  paths: string[],
+  createArgs: any,
+  org: Org | null,
+  isSettingUpProject: boolean,
+  cwd?: string,
+  client?: Client
```
styfle: Maybe we shouldn't allow `createDeploy()` without a `client`?

packages/cli/src/util/deploy/create-deploy.ts
```diff
@@ -82,15 +88,21 @@ export default async function createDeploy(
 
     // If the cert is missing we try to generate a new one and the retry
     if (error.code === 'cert_missing') {
+      if (!client) {
+        throw new Error('Expected client but found undefined');
+      }
```
styfle: This seems wrong to throw here because we already got an error from the server

packages/cli/src/commands/deploy/latest.ts
markozxuu: @styfle  This solution is not working. I tried it in my previous PR

packages/cli/src/commands/deploy/args.ts
markozxuu: Create a speration between the words `Commands` and `Bacis` 👍

TooTallNate: I removed it.

packages/cli/src/commands/deploy/index.ts
TooTallNate: Ya, done.

packages/cli/src/commands/deploy/latest.ts
TooTallNate: Added proper types here.

packages/cli/src/commands/deploy/latest.ts
TooTallNate: I spent some time trying to properly type this, but the auto-generated types from JS `now.create()` ended up being incompatible. I think I'm going to punt on this for now. We can address in a separate PR.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -39,7 +65,7 @@ export default async client => {
     paths = [process.cwd()];
   }
 
-  let { localConfig } = client;
+  let localConfig: VercelConfig | null = client.localConfig;
```
nkzawa, line 68: I wonder why not fix the type of `Client.localConfig` if it can be `null`.

nkzawa (approved):

styfle (approved):
Looks good! Great work @TooTallNate and @markozxuu 🎉

PR #6440, opened 2021-07-06
[cli][client] Handle deployment `aliasWarning` links

mglagola:
### Related Issues

Handles deployment `aliasWarning` links in cli.

#### Tests

#### Code Review

### DIFF
--- a/packages/client/src/check-deployment-status.ts
+++ b/packages/client/src/check-deployment-status.ts
@@ -18,7 +18,7 @@ import {
 
 interface DeploymentStatus {
   type: DeploymentEventType;
-  payload: Deployment | DeploymentBuild[];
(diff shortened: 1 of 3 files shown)


styfle (commented):
This is breaking the existing warnings

```
  ✖ ensure we render a warning for deployments with no files 
```

Also we need a new test for the new behavior

packages/client/src/check-deployment-status.ts
```diff
@@ -95,7 +95,7 @@ export async function* checkDeploymentStatus(
       ) {
         yield {
           type: 'warning',
```
styfle, line 97: Let's change the type here to `alias-warning` so that way we don't have two shapes for the same type

PR #6452, opened 2021-07-08
[cli] Convert `vc projects` to TypeScript

markozxuu:
Yesterday, I spoke with @TooTallNate about of why the command `projects` not appears in the option to the `help` command. We conclude that the team very probably forgot to add it.  This PR focuses on the covert `vc projects` to TypeScript and also added the command to the options of the help :)

#### Tests

#### Code Review


packages/cli/src/commands/deploy/args.ts
```diff
@@ -16,6 +16,7 @@ export const help = () => `
       env                              Manages the Environment Variables for your current Project
       init                 [example]   Initialize an example project
       ls | list            [app]       Lists deployments
+      projects                         Show all projects in the selected team/user
```
styfle: This should go under Advanced just above `domains` and the description should be `Manages your Projects`

styfle: Also this should go under Advanced just about `domains`

packages/cli/src/commands/deploy/args.ts
```diff
@@ -27,6 +27,7 @@ export const help = () => `
 
       rm | remove          [id]        Removes a deployment
       domains              [name]      Manages your domain names
+      proejcts                         Manages your Projects
```
styfle:
```suggestion
      projects                         Manages your Projects
```

styfle (approved):
Thanks!

TooTallNate (approved):

PR #6455, opened 2021-07-09
[cli] Add "out-of-band" mode to `vc login`

TooTallNate:
Adds an "out-of-band" login mode to the `vc login` command, for use in headless environments like inside a Docker container or an SSH session. In these situations, spawning a localhost HTTP server won't work, because the HTTP server will not be accessible to the web browser that is doing the authentication.

In "out-of-band" mode, the login URL is simply presented to the user, and they must visit the login URL in their web browser and complete the authentication. The callback URL is a page on front that shows the verification token to the user and instructs them to copy + paste the token back into the CLI, at which point the login can complete.

Docker/SSH sessions are attempted to be automatically detected, and "out-of-band" mode is enabled implicitly. But there is also an `--oob` flag in case that detection logic were to fail (or if you'd like to run OOB mode locally).

This PR also adds shortcut flags for the Git provider login mechanisms, i.e. `vc login --github`.

https://user-images.githubusercontent.com/71256/125129304-1f5f3380-e0b4-11eb-8139-d1109811bd86.mov

### DIFF
--- a/packages/cli/src/commands/login.ts
+++ b/packages/cli/src/commands/login.ts
@@ -5,8 +5,11 @@ import getArgs from '../util/get-args';
 import handleError from '../util/handle-error';
 import logo from '../util/output/logo';
 import prompt from '../util/login/prompt';
-import doSsoLogin from '../util/login/sso';
+import doSamlLogin from '../util/login/saml';
 import doEmailLogin from '../util/login/email';
+import doGithubLogin from '../util/login/github';
+import doGitlabLogin from '../util/login/gitlab';
+import doBitbucketLogin from '../util/login/bitbucket';
 import { prependEmoji, emoji } from '../util/emoji';
 import { getCommandName, getPkgName } from '../util/pkg-name';
 import getGlobalPathConfig from '../util/config/global-path';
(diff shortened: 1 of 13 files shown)


packages/cli/src/util/login/oauth.ts
```diff
@@ -14,14 +15,57 @@ export default async function doOauthLogin(
...
+  if (typeof result === 'number') {
+    return result;
+  }
 
+  if ('verificationToken' in result) {
+    const { output } = client;
+    output.spinner('Verifying authentication token');
+    result = await verify(
+      client,
+      'nate@vercel.com',
```
zach-jayz: leftover placeholder?

zach-jayz (approved):
LGTM

packages/cli/src/util/login/oauth.ts
TooTallNate: Yup. Removed.

pacocoursey (approved):

packages/cli/src/commands/login.ts
```diff
@@ -41,6 +44,10 @@ const help = () => {
   ${chalk.gray('–')} Log in using a specific team "slug" for SAML Single Sign-On
 
     ${chalk.cyan(`$ ${getPkgName()} login acme`)}
+
+  ${chalk.gray('–')} Log in using GitHub in "out-of-band" mode
```
styfle, line 48: Is there precedent for this terminology?

How about `--headless` instead?

TooTallNate: I copied the terminology from [Twitter](https://developer.twitter.com/en/docs/authentication/oauth-1-0a/pin-based-oauth)'s solution to this same problem.

styfle (approved):
Looks good!

Don't forget to update the docs https://vercel.com/docs/cli

PR #6460, opened 2021-07-11
[cli] Convert the rest the commands to TS

markozxuu:
This PR focuses on convert the rest of the commands located in `cli/src` to TypeScript, in this case there **3 files** still being working with JavaScript. Of course, in this scenario I'm going to exclude the commands that have their own folder, e.g. `src/deploy`. I focus for now in the individual files 👍🏼 

**Update**: The `secret` command will be sent next PR

### 📋 Commands

- [x] list
- [x] remove
- [ ] secret 

#### Tests

#### Code Review


packages/cli/src/commands/list.ts
```diff
@@ -230,23 +238,21 @@ export default async function main(client) {
...
           // flatten since the previous step returns a nested
           // array of the deployment and (optionally) its instances
-          .reduce((ac, c) => ac.concat(c), [])
-          .filter(
-            app == null
-              ? // if an app wasn't supplied to filter by,
-                // we only want to render one deployment per app
-                filterUniqueApps()
-              : () => true
+          .flat()
```
markozxuu, line 247: @TooTallNate  @styfle  Here is no necessary use the `reduce` for getting flatten the array. The `flat` method is the **best option** for this scenario also designed to flatten any level nesting of a `array` 😄

packages/cli/src/commands/list.ts
```diff
@@ -230,23 +238,21 @@ export default async function main(client) {
...
+          .filter(app =>
+            // if an app wasn't supplied to filter by,
+            // we only want to render one deployment per app
+            app === null ? filterUniqueApps() : () => true
           ),
       ],
       {
-        align: ['l', 'l', 'r', 'l', 'b'],
+        align: ['l', 'l', 'r', 'l'],
```
markozxuu: @TooTallNate  @styfle  Removed the `b` element of the array. I check and this not affect to align of the table. Also, I was looking for, but there is no `b` in the alignment parameters.

 [text-table API](https://github.com/substack/text-table)

TooTallNate: Considering that there's 5 columns in the table, I think we should use `l` or `r` as the value, instead of removing the `b` completely.

markozxuu: I seem a good idea. Thank You, Nate 👍🏼

packages/cli/src/commands/remove.ts
```diff
@@ -61,8 +64,8 @@ const help = () => {
 `);
 };
 
-export default async function main(client) {
-  let argv;
+export default async function main(client: Client) {
+  let argv: any;
```
styfle: Let's avoid `any`

packages/cli/src/commands/remove.ts
```diff
@@ -129,33 +132,33 @@ export default async function main(client) {
       .join(' ')} in ${chalk.bold(contextName)}`
   );
 
-  let aliases;
-  let projects;
-  let deployments;
+  let aliases: any;
```
styfle: Let's avoid `any`

packages/cli/src/commands/remove.ts
```diff
@@ -129,33 +132,33 @@ export default async function main(client) {
...
+    const searchFilter = (d: Deployment) =>
       ids.some(
         id =>
           d &&
           !(d instanceof NowError) &&
           (d.uid === id || d.name === id || d.url === normalizeURL(id))
       );
 
-    const [deploymentList, projectList] = await Promise.all([
+    const [deploymentList, projectList] = await Promise.all<any>([
```
styfle, line 152: Let's avoid `any`

packages/cli/src/commands/remove.ts
```diff
@@ -129,33 +132,33 @@ export default async function main(client) {
...
       ),
       Promise.all(
         ids.map(async idOrName => getProjectByIdOrName(client, idOrName))
       ),
     ]);
 
-    deployments = deploymentList.filter(searchFilter);
-    projects = projectList.filter(searchFilter);
+    deployments = deploymentList.filter((d: Deployment) => searchFilter(d));
+    projects = projectList.filter((d: Deployment) => searchFilter(d));
```
styfle: These types will no longer be necessary if you remove the `any` above

packages/cli/src/commands/remove.ts
```diff
@@ -171,7 +174,7 @@ export default async function main(client) {
 
       projectDeployments
         .slice(0, 201)
-        .map(pDeployments => deployments.push(...pDeployments));
+        .map((pDeployments: any) => deployments.push(...pDeployments));
```
styfle: Let's add the correct type to `projectDeployments` so you don't need any

packages/cli/src/commands/remove.ts
```diff
@@ -129,37 +135,38 @@ export default async function main(client) {
...
       ),
       Promise.all(
         ids.map(async idOrName => getProjectByIdOrName(client, idOrName))
       ),
     ]);
 
-    deployments = deploymentList.filter(searchFilter);
-    projects = projectList.filter(searchFilter);
+    deployments = deploymentList
+      .filter((d): d is any => d && 'name' in d)
```
markozxuu: @styfle @TooTallNate  The only way that could be working

packages/cli/src/commands/remove.ts
markozxuu: Done 👍🏼

packages/cli/src/commands/remove.ts
markozxuu: Done 👍🏼

packages/cli/src/commands/remove.ts
markozxuu: 👍🏼

packages/cli/src/commands/remove.ts
markozxuu: Fix 👍🏼

packages/cli/src/commands/remove.ts
markozxuu: The previous solution `DeploymentPartial` no working. In this case, the "best solution" was to opt for `any`

packages/cli/src/commands/remove.ts
```diff
@@ -77,6 +81,8 @@ export default async function main(client) {
     return 1;
   }
 
+  const safe = argv['--safe'];
```
styfle: Let's move this down to line 94/95

packages/cli/src/commands/remove.ts
```diff
@@ -110,7 +116,7 @@ export default async function main(client) {
     return 1;
   }
 
-  let contextName = null;
+  let contextName: string | null = null;
```
styfle, line 119:
```suggestion
  let contextName;
```

I think you can let TS infer the type here and the early return will let it work. Then you can remove the `!` on 147

packages/cli/src/commands/remove.ts
markozxuu: Ok, no problem :)

packages/cli/src/commands/remove.ts
markozxuu: No working this solution
Error:
> Variable 'contextName' implicitly has type 'any' in some locations where its type cannot be determined.ts(7034)

styfle (approved):
Thanks!

TooTallNate (approved):

PR #6477, opened 2021-07-14
[examples][frameworks] Add support for Astro

stramel:
This adds the framework [Astro](https://astro.build/) to the list of frameworks. We're hoping to improve upon this template after getting the initial one added and we find ways to optimize it further.

Example was generated using `npm init astro`

/cc @fredkschott @matthewp @natemoo-re @drwpow


packages/frameworks/src/frameworks.ts
```diff
@@ -1581,6 +1581,59 @@ export const frameworks = [
...
+      },
+      outputDirectory: {
+        value: 'dist',
+      },
+    },
+    dependency: 'astro',
+    devCommand: 'astro dev',
+    buildCommand: 'astro build',
+    getOutputDirName: async () => 'dist',
+    cachePattern: '.snowpack/**',
```
stramel: @fredkschott can you verify this is a cache folder?

packages/frameworks/src/frameworks.ts
```diff
@@ -1581,6 +1581,59 @@ export const frameworks = [
...
+    defaultRoutes: [
+      {
+        src: '^/_astro/(.*)\\.css$',
```
stramel, line 1657: @fredkschott can you verify that this will be okay? I believe `_astro`'s `.css` files are all hashed but I'm definitely not a power user.

FredKSchott: This looks fine for now. JS files are bundled but filenames aren't hashed yet, they will be by v1.0.

packages/frameworks/src/frameworks.ts
FredKSchott: I'm not sure what this config is for, but I know that Vite has a similar cache folder and they don't have a `cachePattern` key in their config.

stramel: Okay, I noticed it in the `.gitignore` wasn't sure where it came from but I'll remove it for now. The `cachePattern` is to optimize build caches I believe. 

> A glob string of files to cache for future deployments

FredKSchott:
If it helps to get this looked at, Github published some stats on popularity & growth as a part of merging our syntax into Github.com code syntax highlighting: https://github.com/github/linguist/pull/5460#issuecomment-882443477

stramel:
@styfle @leerob Could I bug either of you for a review? I noticed you had reviewed similar PRs for this repo.

examples/astro/README.md
```diff
@@ -0,0 +1,40 @@
+# Welcome to [Astro](https://astro.build)
```
leerob, line 1: Could this README follow the pattern of the other examples here? It can keep everything, just needs to mention deploying to Vercel and have the deploy button.

examples/astro/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@example/starter",
+  "version": "0.0.1",
```
leerob:
```suggestion
```

examples/astro/public/style/home.css
```diff
@@ -0,0 +1,52 @@
...
+    font-size: 1em;
+    font-weight: 500;
+}
+header img {
+    width: 2em;
+    height: 2.667em;
+}
+
+h2 {
+    font-weight: 500;
```
leerob: The formatting on this file looks off. We probably need to add Prettier to this repo, but in the meantime, could you run Prettier over these files?

packages/frameworks/src/frameworks.ts
```diff
@@ -1618,6 +1618,58 @@ export const frameworks = [
...
+  {
+    name: 'Astro',
+    slug: 'astro',
+    demo: 'https://astro.examples.vercel.com',
+    logo: 'https://raw.githubusercontent.com/vercel/vercel/main/packages/frameworks/logos/astro.svg',
+    tagline:
+      'Astro is a next-generation static site generator that mixes lightning-fast performance with modern developer experience.',
+    description: 'An Astro site, using the default starter kit.',
+    website: 'https://astro.build',
+    envPrefix: 'SNOWPACK_PUBLIC_',
```
leerob, line 1627: Should this be `ASTRO_`?

leerob (commented):
I can deploy the example to `astro.examples.vercel.com` once we're ready to go!

packages/frameworks/src/frameworks.ts
stramel: I believe this is still `SNOWPACK_PUBLIC_` since internally it is using `snowpack`. I don't believe this has been aliased yet. @fredkschott can you confirm?

leerob: Sorry, I should have been more clear. This is used for Vercel.

https://vercel.com/docs/environment-variables#system-environment-variables

CleanShot 2021-07-26 at 09 57 37@2x

stramel: Sorry, I think I'm missing the point of this property. The comment on the Framework typing says it is used to inline values into the browser bundle.  

```ts
  /**
   * The environment variable prefix used to inline values into the browser bundle.
   * @example "NEXT_PUBLIC_"
   */
  envPrefix?: string;
```

This would be how the developer could utilize the vercel env vars in their builds, correct? If I wanted to use `VERCEL_ENV` in my browser build I would need to use `NEXT_PUBLIC_VERCEL_ENV` for next? If so, I believe it would be `SNOWPACK_PUBLIC_VERCEL_ENV`?

styfle: Yep, that sounds right. Are there Astro docs that show this usage?

packages/frameworks/logos/astro.svg
```diff
@@ -0,0 +1 @@
...31.1c6.8 0 10.2 0 13.1 1.1A16 16 0 01131 19z" clip-rule="evenodd"/><path fill="#FF5D01" fill-rule="evenodd" d="M136.2 180.2c-7.1 6-21.4 10.2-37.8 10.2-20.2 0-37-6.3-41.5-14.7-1.6 4.8-2 10.4-2 14 0 0-1 17.3 11 29.3 0-6.2 5.1-11.3 11.4-11.3 10.7 0 10.7 9.4 10.7 17v.6c0 11.6 7 21.5 17 25.7a23.3 23.3 0 01-2.3-10.2c0-11 6.5-15.2 14-20 6-3.7 12.6-8 17.2-16.4a31 31 0 003.8-14.8c0-3.3-.6-6.5-1.5-9.4z" clip-rule="evenodd"/><path fill="url(#paint1_linear)" fill-rule="evenodd" d="M136.2 180.2c-7.1 6-21.4 10.2-37.8 10.2-20.2 0-37-6.3-41.5-14.7-1.6 4.8-2 10.4-2 14 0 0-1 17.3 11 29.3 0-6.2 5.1-11.3 11.4-11.3 10.7 0 10.7 9.4 10.7 17v.6c0 11.6 7 21.5 17 25.7a23.3 23.3 0 01-2.3-10.2c0-11 6.5-15.2 14-20 6-3.7 12.6-8 17.2-16.4a31 31 0 003.8-14.8c0-3.3-.6-6.5-1.5-9.4z" clip-rule="evenodd"/><defs><linearGradient id="paint0_linear" x1="144.6" x2="95.8" y1="5.4" y2="173.4" gradientUnits="userSpaceOnUse"><stop stop-color="#000014"/><stop offset="1" stop-color="#150426"/></linearGradient><linearGradient id="paint1_linear" x1="168.3" x2="126.1" y1="130.5" y2="219" gradientUnits="userSpaceOnUse"><stop stop-color="#FF1639"/><stop offset="1" stop-color="#FF1639" stop-opacity="0"/></linearGradient></defs></svg>
```
styfle: I think this image needs to be a square, otherwise it will get cropped when we turn it into a circle.

For example 
image

packages/frameworks/src/frameworks.ts
```diff
@@ -1618,6 +1618,58 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `astro build`',
+      },
+      devCommand: {
+        value: 'astro dev',
```
styfle: Does `astro dev` read the `$PORT` env var already? If not, then it will need to be passed in a flag such as `astro dev --port $PORT`.

stramel: > Does astro dev read the $PORT env var already?
It does not. I have filed an issue to discuss adding one, https://github.com/snowpackjs/astro/issues/872

Would this need to be added to both `settings.devCommand.value` and `devCommand` properties?

styfle: Yes

styfle (approved):

packages/frameworks/logos/astro.svg
stramel: I have made it square and re-centered it. Please let me know if the way I did so will be a problem.

packages/frameworks/src/frameworks.ts
stramel: @styfle I don't believe there is anything in the docs currently. I know the team just launched the docs site recently. I will file an issue to add it to the docs. In the meantime, you can see that the docs site itself utilizes the `SNOWPACK_PUBLIC_` prefix. 

https://github.com/snowpackjs/astro/blob/main/docs/README.md
https://github.com/snowpackjs/astro/blob/23735c53cc59e489cf1d0f528a97c2ce535d0a21/docs/src/components/AvatarList.astro#L9

Related issue: https://github.com/snowpackjs/astro/issues/873

styfle: Ok lets wait to see what happens with the upstream issue.

I'm also wondering if we should wait for JS files to be hashed so we can cache them per [this comment](https://github.com/vercel/vercel/pull/6477#discussion_r669937050)

stramel: Sounds good! I will also open a RFC around additional file hashing in the production builds to keep the conversation going.

leerob (approved):
@styfle if this looks good to you, I can deploy the example!

styfle (changes requested):
I think we should wait until Astro 1.0 is released (or at least a RC) before supporting zero config, otherwise we'll have to have two presets to avoid breaking changes:

- Astro 0.x
- Astro 1.x

The reason is, we can't safely add caching for JS files so it will have to be versioned.
See  [this comment](https://github.com/vercel/vercel/pull/6477#discussion_r669937050).

Furthermore, it sounds like the SNOWPACK env prefix might change to ASTRO.
See [this comment](https://github.com/vercel/vercel/pull/6477#discussion_r676915702).

Putting this on hold until both of those can be addressed by the Astro team.

FredKSchott:
Not opposed to that at all, makes sense if this multi-version support would be difficult to manage. Once we reach 1.0 I don't expect larger changes to the degree that your deployment defaults would need to change.

TooTallNate:
Superseded by #7747.

PR #6482, opened 2021-07-16
[frameworks][examples] Add support for zero-config SvelteKit

leerob:
While we already have support for [SvelteKit](https://sveltekit-delta.vercel.app) on Vercel with the [File System API](https://vercel.com/docs/more/adding-your-framework), this makes the Svelte logo show up and adds an example. I also updated the Svelte example to match their default template and guide the user to use SvelteKit if they want functions.

https://sveltekit.examples.vercel.com


packages/frameworks/src/frameworks.ts
```diff
@@ -1090,44 +1127,6 @@ export const frameworks = [
       },
     ],
   },
-  {
-    name: 'Sapper',
-    slug: 'sapper',
```
styfle: I don't think we can safely remove from this list, otherwise `front` will break for existing projects when fetching the logo and other metadata.

leerob: Should we just revert this or do you think we need the whole example back?

AndyBitz (approved):

AndyBitz:
We have a test that ensures that the example Deployments are public, so they can be inspected by anyone.

You'll need to make sure that the Project for `sveltekit.examples.vercel.com` has "Logs and Source Protection" disabled in the Project Settings.

> Updated it myself, so the tests pass now.

styfle (approved):

styfle:
Looks like you updated the basic svelte example. Did you deploy that one too? Need to make sure any changes to examples are deployed so the screenshot is correct.

leerob:
Done!

packages/frameworks/src/frameworks.ts
```diff
@@ -875,6 +875,43 @@ export const frameworks = [
...
+      'SvelteKit is a framework for building web applications of all sizes.',
+    description:
+      'A SvelteKit app optimized to work for serverless.',
+    website: 'https://kit.svelte.dev',
+    detectors: {
+      every: [
+        {
+          path: 'package.json',
+          matchContent:
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"@sveltejs/kit":\\s*".+?"[^}]*}',
```
```suggestion
            '"(dev)?(d|D)ependencies":\\s*{[^}]*"@sveltejs\\/kit":\\s*".+?"[^}]*}',
```

packages/frameworks/src/frameworks.ts
```diff
@@ -875,6 +875,43 @@ export const frameworks = [
...
+        placeholder: '`npm run build` or `svelte-kit build`',
+      },
+      devCommand: {
+        placeholder: 'svelte-kit dev',
+      },
+      outputDirectory: {
+        placeholder: 'public',
+      },
+    },
+    devCommand: 'svelte-kit dev',
```
styfle: Does it accept the `$PORT` env var already? If not then we need to do something like `svelte-kit dev --port $PORT`

leerob: I can't find anything in the docs about this, this is all I found: https://github.com/sveltejs/kit/blob/073fc3b5ff7500087b9ae5da8f635430b4d96267/packages/kit/src/cli.js#L55

styfle: Let's pass `--port` here and in the placeholder too

https://github.com/sveltejs/kit/blob/073fc3b5ff7500087b9ae5da8f635430b4d96267/packages/kit/src/cli.js#L75

styfle: You can also verify this by running `vc dev` and modify your projects Dev Command to see if it works

packages/frameworks/src/frameworks.ts
```diff
@@ -875,6 +875,43 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `svelte-kit build`',
+      },
+      devCommand: {
+        placeholder: 'svelte-kit dev',
```
leerob:
```suggestion
        placeholder: 'svelte-kit dev --port $PORT',
```

packages/frameworks/src/frameworks.ts
```diff
@@ -875,6 +875,43 @@ export const frameworks = [
...
+      },
+      outputDirectory: {
+        placeholder: 'public',
+      },
+    },
+    devCommand: 'svelte-kit dev',
```
leerob:
```suggestion
    devCommand: 'svelte-kit dev --port $PORT',
```

styfle (approved):
Thanks!

mcsdevv (approved):

PR #6510, opened 2021-07-22
[cli] Add request counter to `--debug` logs

TooTallNate:
Follow up to #6508. Adds a request "counter" that increments for every API request, matching how `front` does it.

Also tweaks the `output.time()` output a bit by rendering the timestamp in gray, and with `[]` brackets. If the time diff is < 1000ms then the full milliseconds will be printed intstead of the seconds.

Screen Shot 2021-07-22 at 2 45 39 PM


AndyBitz (approved):

styfle (approved):
💅

PR #6532, opened 2021-07-30
[build-utils] Invoke npm 7 when lockfile version 2 is detected

TooTallNate:
As the title suggests, this PR makes it such that npm 7 will be invoked when there is a `package-lock.json` file with `lockfileVersion` 2 or greater, by prepending the the `$PATH` a directory within the build container where npm 7 is located.

The test fixture `20-npm-7` is also enabled for E2E testing, to ensure that the proper npm version is enabled in the production build.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -329,9 +329,9 @@ export async function runNpmInstall(
       .filter(a => a !== '--prefer-offline')
       .concat(['install', '--no-audit', '--unsafe-perm']);
 
-    if (lockfileVersion === 2) {
+    if (typeof lockfileVersion === 'number' && lockfileVersion >= 2) {
       // Ensure that npm 7 is at the beginning of the `$PATH`
-      env.PATH = `/node16/bin:${env.PATH}`;
+      env.PATH = `/node16/bin-npm7:${env.PATH}`;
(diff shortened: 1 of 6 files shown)


AndyBitz (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -285,32 +307,34 @@ export async function runNpmInstall(
...
   let commandArgs: string[];
 
   if (cliType === 'npm') {
     opts.prettyCommand = 'npm install';
-    command = 'npm';
     commandArgs = args
       .filter(a => a !== '--prefer-offline')
       .concat(['install', '--no-audit', '--unsafe-perm']);
+
+    if (typeof lockfileVersion === 'number' && lockfileVersion >= 2) {
```
styfle, line 332:
```suggestion
    if (typeof lockfileVersion === 'number' && lockfileVersion === 3) {
```

Shouldn't this be version 3 instead since 2 is backwards compatible?

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

styfle: I guess npm 6 only creates version 1 so we can use npm7 for version 2 and newer

styfle (approved):

PR #6554, opened 2021-08-04
[client][frameworks][api] Update `readdir()` call with `withFileTypes`

styfle:
This PR avoids running unnecessary code such as stat().isDirectory()

See docs:

- [fs.readdir()](https://nodejs.org/api/fs.html#fs_fspromises_readdir_path_options)
- [fs.Dirent](https://nodejs.org/api/fs.html#fs_class_fs_dirent)


packages/client/src/create-deployment.ts
```diff
@@ -50,8 +50,6 @@ export default function buildCreateDeployment() {
     clientOptions.isDirectory =
       !Array.isArray(path) && lstatSync(path).isDirectory();
 
-    let rootFiles: string[];
```
TooTallNate, line 53: This was just not being used?

TooTallNate (approved):

packages/client/src/create-deployment.ts
styfle: Yep, it was just used for that one log which was useless.

The root files were read again in `let { fileList } = await buildFileTree()` just below

TooTallNate:
Thanks!

PR #6574, opened 2021-08-10
[api] Fix `GET /api/examples/list` output

TooTallNate:
A regression from #6554 caused the return value to contain a nested object with a `name` property for the `name` key of the response in the list.


styfle:
Nice catch, thanks!

PR #6576, opened 2021-08-10
[cli] Remove `withSpinner()` helper

TooTallNate:
It's not really necessary and doesn't use the common `Output` instance.


styfle (approved):
🧹

PR #6638, opened 2021-08-26
[cli] Rewrite Unit tests to TypeScript + Jest

TooTallNate:
* Rewrites the CLI unit tests to be TypeScript and use Jest (consistent with the unit tests in the other packages in this repo).
* The file structure of the new test files mirrors that of the `src` directory, so that it's easy to find the relevant tests.
* Code coverage is also properly set up through Jest - you can already see a big increase in coverage from this PR.
* Adds a mock API server framework, with the intention of making it possible to write granular tests of the CLI commands. Using mocks also has the benefit of not requiring `VERCEL_TOKEN` env var to be set, so 3rd party PRs will be able to have their tests run. Ultimately this will also help with test coverage, since we will be writing unit tests that test the commands' code directly.
* Converts `Output` into a proper class (which is subclassed for the mocks).


packages/cli/test/mocks/client.ts
```diff
@@ -0,0 +1,109 @@
...
+      apiUrl: '',
+      authConfig: {},
+      output: new Output(),
+      config: {},
+      localConfig: {},
+    });
+    this.mockOutput = jest.fn();
+
+    this.app = express();
+    this.app.use(express.json());
```
styfle, line 33: Why use `express` here? We don't really use it anywhere else as far as I know?

packages/cli/src/util/get-user.ts
```diff
@@ -2,11 +2,7 @@ import Client from './client';
 import { User } from '../types';
 import { APIError, InvalidToken, MissingUser } from './errors-ts';
 
-let user: User | undefined;
-
 export default async function getUser(client: Client) {
-  if (user) return user;
```
styfle, line 8: This used to be cached?

packages/cli/src/util/output/create-output.ts
```diff
@@ -68,112 +77,99 @@ function _createOutput({ debug: debugEnabled = false }: OutputOptions = {}) {
...
-    dim,
-    time,
-    note,
-    spinner: setSpinner,
-    stopSpinner,
   };
 }
+
+export default function createOutput(opts?: OutputOptions) {
+  return new Output(opts);
```
styfle, line 174: I noticed this is no longer a singleton.

What happens when `createOutput()` is called more than once?

packages/cli/test/mocks/client.ts
TooTallNate: Mainly just for the decent router capabilities. Open to suggestions if you'd prefer a different framework.

packages/cli/src/util/get-user.ts
TooTallNate: It did, but for not really any reason. This function is invoked exactly once for most commands (and if it's > 1 then the command should be updated to not do that). Also the caching was getting in the way of the unit tests getting unexpectedly cached results 😛

packages/cli/src/util/output/create-output.ts
TooTallNate: The codebase has been cleaned up at this point to not create many output instances, just a single one gets created in the entrypoint file. The singleton requirement was more of a temporary band-aid while cleaning that part up. I think it's ok to remove that requirement at this point.

styfle (approved):

AndyBitz (approved):

PR #6651, opened 2021-08-31
[build-utils] Add `allowQuery` to Lambda

styfle:
This adds the `allowQuery` prop to the Lambda class

### DIFF
--- a/packages/build-utils/src/prerender.ts
+++ b/packages/build-utils/src/prerender.ts
@@ -66,11 +66,18 @@ export class Prerender {
     }
     this.fallback = fallback;
 
-    if (allowQuery && !Array.isArray(allowQuery)) {
-      throw new Error(
-        'The `allowQuery` argument for `Prerender` needs to be a string Array.'
-      );
+    if (!allowQuery) {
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/prerender.ts
```diff
@@ -62,5 +65,12 @@ export class Prerender {
       );
     }
     this.fallback = fallback;
+
+    if (allowQuery && !Array.isArray(allowQuery)) {
+      throw new Error(
+        'The `allowQuery` argument for `Prerender` needs to be a string Array.'
```
AndyBitz: Should we do `allowQuery.some(x => typeof x !== 'string')` to make sure it's really all strings?

AndyBitz (approved):

ijjk (approved):

packages/build-utils/src/prerender.ts
styfle: Good point, done!

AndyBitz (approved):

dav-is (approved):

ijjk (approved):

PR #6652, opened 2021-08-31
[cli] Remove support for single file deployments

TooTallNate:
Deploying a single file has printed a deprecation warning for a long time. Let's finally remove that behavior.


errors/no-single-file-deployments.md
```diff
@@ -0,0 +1,9 @@
+# No Single File Deployments
+
+#### Why This Error Occurred
+
+You attempted to create a Vercel deployment where the input is a file, rather than a directory. Previously this was allowed, however this behavior has been removed as of Vercel CLI vX.X.X because it exposed a potential security risk if the user accidentally created a deployment from a sensitive file.
```
styfle: Let's remember to add the version here. Probably semver major?

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -188,7 +187,7 @@ export default async function setupAndLink(
         [sourcePath],
         createArgs,
         org,
-        !isFile,
+        true,
```
styfle, line 181: Should this param be removed also?

packages/cli/src/util/validate-paths.ts
```diff
@@ -78,14 +74,12 @@ export default async function validatePaths(
...
-    output.print(
-      `${prependEmoji(
-        'Deploying files with Vercel is deprecated (https://vercel.link/faq-deploy-file)',
-        emoji('warning')
-      )}\n`
-    );
+  if (!pathStat.isDirectory()) {
+    output.prettyError({
+      message: 'Support for single file deployments has been removed.',
+      link: 'https://err.sh/vercel/no-single-file-deployments',
```
styfle: I don't think we use err.sh anymore. This should be vercel.link

packages/cli/src/util/link/setup-and-link.ts
TooTallNate: No, it's still used dynamically in the `deploy` command.

errors/no-single-file-deployments.md
TooTallNate: Tentatively set it to 24.0.0 since that will be the next major release.

packages/cli/src/util/validate-paths.ts
TooTallNate: K, done. Right now it's just pointing to the error markdown page on this PR, like how err.sh would have done it. Let me know if you think we should link to somewhere else.

styfle (approved):

PR #6683, opened 2021-09-06
Use pip --upgrade when installing python function dependencies

coolreader18:
Passes `--upgrade` to `pip install -r` so that werkzeug 2.0 can be installed if that's a dependency of the function.

### Related Issues

> Related to https://github.com/vercel/vercel/pull/6272#issuecomment-913226139

#### Tests

I don't think there's a good way to test this, but let me know if I'm wrong - it seems like the fixture tests just check that the runtime can build them.

#### Code Review

### DIFF
--- a/packages/python/src/install.ts
+++ b/packages/python/src/install.ts
@@ -86,6 +86,10 @@ interface InstallRequirementArg {
   args?: string[];
 }
 
+// note that any internal dependency that vc_init.py requires that's installed
+// with this function can get overriden by a newer version from requirements.txt,
+// so vc_init should do runtime version checks to be compatible with any recent
+// version of its dependencies
 export async function installRequirement({
   dependency,
   version,
@@ -120,5 +124,5 @@ export async function installRequirementsFile({
     debug(`Skipping requirements file installation, already installed`);
     return;
   }
-  await pipInstall(workPath, ['-r', filePath, ...args]);
+  await pipInstall(workPath, ['--upgrade', '-r', filePath, ...args]);
 }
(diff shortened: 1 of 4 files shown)


coolreader18:
Oh, also - as far as I can tell, `vc_init.py` works fine with werkzeug 2.0; none of the breaking changes affect what it's using.

styfle (commented):
Thanks!

Please add a test fixture so we can ensure this works with Flask 2.

Take a look at this one for an example:
https://github.com/vercel/vercel/tree/cfae7ec3c291d1802fb6eec6642a0ebf792fb71d/packages/python/test/fixtures/00-request-path

Then you can run the integration tests with:

```
yarn install
yarn build
cd ./packages/python
export VERCEL_TOKEN='YourTokenHere'
yarn test-integration-once
```

Note this will make real deployments so feel free to create a test account for running these.

coolreader18:
Added! I wasn't able to get tests working locally; every one errored with `Cannot find module '/vercel/35fd446a/.build-utils/.builder/node_modules/@vercel/python/dist/index.js'` (which was happening before as well), but hopefully it should just work in CI?

styfle:
You probably need to run `yarn build` at the root of the repo so all packages are built at least once

I'll try running CI here: #6688

styfle (approved):
Great work, thanks! 🎉

styfle:
You can try this out on canary with `FORCE_RUNTIME_TAG=canary` env var for zero config or `@vercel/python@canary` if you're using the legacy `builds` property.

PR #6700, opened 2021-09-09
[all] Bump ncc to 0.31.1

styfle:
https://github.com/vercel/ncc/releases/tag/0.31.0

https://github.com/vercel/ncc/releases/tag/0.31.1

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -587,19 +587,92 @@ const main = async () => {
     }
   }
 
-  if (!targetCommand) {
-    const sub = param(subcommand);
-    console.error(error(`The ${sub} subcommand does not exist`));
-    return 1;
-  }
-
   const metric = metrics();
   let exitCode;
   const eventCategory = 'Exit Code';
 
   try {
     const start = Date.now();
-    const full = require(`./commands/${targetCommand}`).default;
+    let full: any;
+    switch (targetCommand) {
+      case 'alias':
+        full = await import('./commands/alias');
+        break;
+      case 'billing':
+        full = await import('./commands/billing');
+        break;
+      case 'certs':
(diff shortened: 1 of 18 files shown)


AndyBitz (approved):

packages/cli/src/index.ts
```diff
@@ -587,20 +587,93 @@ const main = async () => {
...
+        break;
+    }
+
+    if (!func || !targetCommand) {
+      const sub = param(subcommand);
+      output.error(`The ${sub} subcommand does not exist`);
+      return 1;
+    }
+
+    if (func.default) {
```
TooTallNate, line 672: I believe this is always going to be true.

packages/cli/src/index.ts
```diff
@@ -587,20 +587,93 @@ const main = async () => {
...
-
   const metric = metrics();
   let exitCode;
   const eventCategory = 'Exit Code';
 
   try {
     const start = Date.now();
-    const full = require(`./commands/${targetCommand}`).default;
-    exitCode = await full(client);
+    let func: any;
```
TooTallNate, line 596:
```suggestion
    let func: (c: Client) => Promise<number>;
```

TooTallNate (approved):

packages/cli/src/index.ts
styfle: Some of the commands are TS and some are JS (notable `secrets.js`) so this might be different

packages/cli/src/index.ts
styfle: This was failing for me which is why I left it as `any`.

I'll try again in another PR and see if we can get this more strict.

Another option is to move the imports to the top because they're already bundled into a single file via ncc.

PR #6713, opened 2021-09-13
[routing-utils] Fix host segment replacing

ijjk:
This fixes segments inside of the host of the destination failing to parse with `url.parse` due to the unexpected colon by escaping the colon for segments and then unescaping them after we have parsed the URL. 

### Related Issues

Fixes: https://vercel.slack.com/archives/CLDDX2Y0G/p1631496861015500

#### Tests

#### Code Review

### DIFF
--- a/packages/routing-utils/src/superstatic.ts
+++ b/packages/routing-utils/src/superstatic.ts
@@ -211,91 +211,111 @@ export function collectHasSegments(has?: HasField) {
   return [...hasSegments];
 }
 
+const escapeSegment = (str: string, segmentName: string) =>
+  str.replace(new RegExp(`:${segmentName}`, 'g'), `__ESC_COLON_${segmentName}`);
+
+const unescapeSegments = (str: string) => str.replace(/__ESC_COLON_/gi, ':');
+
 function replaceSegments(
   segments: string[],
   hasItemSegments: string[],
   destination: string,
   isRedirect?: boolean
 ): string {
-  const parsedDestination = parseUrl(destination, true);
+  const namedSegments = segments.filter(name => name !== UN_NAMED_SEGMENT);
+  const canNeedReplacing =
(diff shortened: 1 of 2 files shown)


packages/routing-utils/test/superstatic.spec.js
```diff
@@ -564,7 +584,7 @@ test('convertRewrites', () => {
     { src: '^\\/some\\/old\\/path$', dest: '/some/new/path', check: true },
     {
       src: '^\\/proxy(?:\\/(.*))$',
-      dest: 'https://www.firebase.com',
+      dest: 'https://www.firebase.com/',
```
styfle, line 588: Can you add a test where the input has a trailing slash? Just want to make sure it doesn’t double up.

packages/routing-utils/src/superstatic.ts
```diff
@@ -211,91 +211,111 @@ export function collectHasSegments(has?: HasField) {
...
+  destination = formatUrl({
+    ...rest,
+    hostname,
+    pathname,
+    query,
+    hash,
+  });
+
+  // url.format() escapes the dollar sign but it must be preserved for now-proxy
+  return destination.replace(/%24/g, '$');
```
styfle, line 318: Does this replace need to happen at the bottom? I want to avoid replacing user input

packages/routing-utils/test/superstatic.spec.js
ijjk: Added

packages/routing-utils/src/superstatic.ts
ijjk: This is where this was happening previously, I don't think we can do this any earlier since it needs come after the `url.format`

styfle (approved):

PR #6742, opened 2021-09-20
[routing-utils] Allow passing internal params to convertRewrites

### LINKED ISSUE #27563: Rewrites don't automatically pass path params in query with i18n enabled
When leveraging rewrites in `next.config.js` rewrites are meant to automatically pass params from the `source` in the `destination`'s query unless a param is already used in the destination. This doesn't work correctly when `i18n` is used as well unless `locale: false` is set on the rewrite since we are detecting the internal `locale` path param as a user path param causing the `destination` query to not be updated 

Example: 
```js
// next.config.js
module.exports = {
  i18n: {
    defaultLocale: 'en'
    locales: ['en', 'fr']
  },
  rewrites() {
    return [
      // path should be available in the query for the below rewrite 
      // but is not when the above i18n config is enabled
      { source: '/old-blog/:post', destination: '/blog/new-location' }, 
    ]
  }
}
```

ijjk:
This adds an argument to allow passing internal param names that should be ignored when considering whether params should be auto-added to a rewrite's destination query. After adding this we should be able to resolve https://github.com/vercel/next.js/issues/27563 in the runtime where `convertRewrites` is called. 

This matches Next.js' handling for internal params which can be seen [here](https://github.com/vercel/next.js/blob/e90825ad88c6f8191bed1dcc43e03465afba32ed/packages/next/shared/lib/router/utils/prepare-destination.ts#L203)

### Related Issues

x-ref: https://github.com/vercel/next.js/issues/27563

#### Tests

#### Code Review

### DIFF
--- a/packages/routing-utils/test/superstatic.spec.js
+++ b/packages/routing-utils/test/superstatic.spec.js
@@ -471,115 +471,122 @@ test('convertRedirects', () => {
 });
 
 test('convertRewrites', () => {
-  const actual = convertRewrites([
-    { source: '/some/old/path', destination: '/some/new/path' },
-    { source: '/proxy/(.*)', destination: 'https://www.firebase.com' },
-    { source: '/proxy/(.*)', destination: 'https://www.firebase.com/' },
-    {
-      source: '/proxy-regex/([a-zA-Z]{1,})',
-      destination: 'https://firebase.com/$1',
-    },
-    {
-      source: '/proxy-port/([a-zA-Z]{1,})',
-      destination: 'https://firebase.com:8080/$1',
-    },
-    { source: '/projects/:id/edit', destination: '/projects.html' },
-    {
-      source: '/users/:id',
-      destination: '/api/user?identifier=:id&version=v2',
-    },
-    {
-      source: '/:file/:id',
-      destination: '/:file/get?identifier=:id',
-    },
-    {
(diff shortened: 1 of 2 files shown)


dav-is (approved):

packages/routing-utils/test/superstatic.spec.js
```diff
@@ -755,6 +767,7 @@ test('convertRewrites', () => {
     ['/hello/world'],
     ['/hello/world'],
     ['/array-query-string/10/email'],
+    ['/en/hello'],
```
styfle, line 770: Should `/en/hello/` (with a trailing slash) match or not?

It would be good to add it to the `mustMatch` or `mustNotMatch` array either here in line 770 or line 796

ijjk: It should not in this case but in Next.js itself we add a [wrapper](https://github.com/vercel/next.js/blob/e90825ad88c6f8191bed1dcc43e03465afba32ed/packages/next/lib/load-custom-routes.ts#L82) to handle matching trailing slashes automatically

ijjk: Added additional check in https://github.com/vercel/vercel/pull/6742/commits/e63682b5dafccc3a97e730ad327f01309ba024d9

styfle (approved):
Thanks!

PR #6787, opened 2021-09-29
[go] Add support for version 1.17

StefMa:
### Related Issues

I noticed that I can not build with Go 1.17.
I got an error:
> Warning: Unknown Go version /vercel/path1/go.mod

And it used 1.16 instead.

After getting in contact with your support (via your website), Dom Eccleston send me a link to this `go-helper.ts` file.
I guess this change will now support Go 1.17 (as well as 1.17.1) and more patch versions of 1.15 and 1.16 ("kind of support". When I see it correctly, you ignore patch version and install major versions only, right?).

This will also use 1.17 as default now.
So probably the docs needs to be updated too.
But I don't know where I can find them 🙃 

TBH, this doesn't really scale, right? 
I guess you should think about a better (more scalable) approach, than hardcoding Go versions and update them on each release 😅 . 

#### Tests

- [ ] ~The code changed/added as part of this PR has been covered with tests~
- [ ] ~All tests pass locally with `yarn test-unit`~

#### Code Review

- [ ] ~Issue from task tracker has a link to this PR~


TooTallNate (approved):

StefMa:
Any news here?

packages/go/go-helpers.ts
```diff
@@ -6,8 +6,9 @@ import { join } from 'path';
 import stringArgv from 'string-argv';
 import { debug } from '@vercel/build-utils';
 const versionMap = new Map([
-  ['1.16', '1.16'],
-  ['1.15', '1.15.8'],
+  ['1.17', '1.17.1'],
```
styfle:
```suggestion
  ['1.17', '1.17.3'],
```

packages/go/go-helpers.ts
```diff
@@ -6,8 +6,9 @@ import { join } from 'path';
...
+  ['1.16', '1.16.8'],
```
styfle:
```suggestion
  ['1.16', '1.16.10'],
```

packages/go/go-helpers.ts
```diff
@@ -6,8 +6,9 @@ import { join } from 'path';
...
+  ['1.15', '1.15.15'],
```
styfle:
```suggestion
  ['1.15', '1.15.8'],
```

Let's leave 1.15 alone since its considered "archived" according to https://golang.org/dl/#stable

styfle (commented):
Thanks! Looks like there are some new versions out now

styfle:
Running tests in PR #7002 and then we can ship canary once it passes

StefMa:
Yay 🎉 🥳

PR #6821, opened 2021-10-11
Add 'VITE_' prefix to SvelteKit framework

domeccleston:
### 📋 Checklist

Since SvelteKit uses Vite under the hood, it needs the _VITE environment variable prefix. This adds it.

Example which reproduces the Svelte Kit example with undefined environment variables:
-  [Repo](https://github.com/domeccleston/env-var-vercel-sveltekit)
- [Preview](https://env-var-vercel-sveltekit-domeccleston.vercel.app/)

#### Tests

Currently I get an error when trying to run the tests: `  Module ts-jest in the transform option was not found.`

#### Code Review


leerob:
You might want to check with @styfle, I believe changing the env var preset is considered a breaking change here. Did SvelteKit change this under the hood?

packages/frameworks/src/frameworks.ts
```diff
@@ -881,6 +881,7 @@ export const frameworks = [
       'SvelteKit is a framework for building web applications of all sizes.',
     description: 'A SvelteKit app optimized to work for serverless.',
     website: 'https://kit.svelte.dev',
+    envPrefix: 'VITE_',
```
styfle, line 1108: is this documented by sveltkit?

Also, are there plans to use SVELTE_ prefix instead? 

Is the current VITE_ prefix documented?

raymclee:
Is it the reason I cannot get SVELTEKIT_VERCEL_URL on vercel?

packages/frameworks/src/frameworks.ts
styfle: Looks like its documented here https://kit.svelte.dev/faq#env-vars

styfle (approved):
Thanks!

Rich-Harris (approved):

TooTallNate (approved):

TooTallNate:
Thanks @domeccleston!

devunt:
This should be updated to `PUBLIC_` prefix, according to the newly released SvelteKit v1.0 docs [1]

[1] https://kit.svelte.dev/docs/configuration#env

Rich-Harris:
We actually don't want a prefix here, because it would cause the values to be available via `$env/static/public` and `$env/dynamic/public`, when they only need to be available on the server (via `$env/static/private` and `$env/dynamic/private`).

In particular, adding them to public env will mean they're included in every rendered HTML page (because that's how `$env/dynamic/public` is populated), which is undesirable.

PR #6822, opened 2021-10-11
[cli] Add `vercel pull` command

jaredpalmer:
Add `vercel pull` command. Closes https://github.com/vercel/runtimes/issues/142. 

CleanShot 2021-10-19 at 08 50 07

It does the following:
- Pull the `.env` file (it calls `pull` internally)
- Adds the following to `.vercel/project.json` (which will then be used by `vercel build`):
   ```tsx
   interface ProjectLink {
        projectId: string
        orgId: string
        settings: {
          buildCommand: string | null,
          devCommand:  string | null,
          directoryListing:  string | null,
          outputDirectory:  string | null,
          rootDirectory:  string | null,
          framework: string | null,
        }
    }
   ```
This PR also deprecates `vc env pull` with a warning:

CleanShot 2021-10-19 at 08 50 21

### 📋 Checklist

- [x] Help output
- [x] Flags (need to combine flags from `env pull`)

#### Tests

#### Code Review


packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (link.status === 'error') {
+    return link.exitCode;
+  } else if (link.status === 'not_linked') {
+    output.error(
+      `Your codebase isn’t linked to a project on Vercel. Run ${getCommandName(
+        'link'
+      )} to begin.`
+    );
+    return 1;
+  } else {
```
TooTallNate: This `else` block doesn't seem necessary, since the previous two do a `return`. Prefer to keep the meat of the code at the "top-level".

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+  const cwd = argv._[1] || process.cwd();
+  debug = argv['--debug'];
+  yes = argv['--yes'];
+  const link = await getLinkedProject(client);
+  if (link.status === 'error') {
+    return link.exitCode;
+  } else if (link.status === 'not_linked') {
+    output.error(
+      `Your codebase isn’t linked to a project on Vercel. Run ${getCommandName(
```
TooTallNate: Let's do the linking inline here, like how `vc dev` does it.

packages/cli/src/commands/pull.ts
jaredpalmer: Agree. I copied that out of `pull` good looks

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,99 @@
...
+      return 0;
+    }
+  }
+
+  if (link.status === 'error') {
+    return link.exitCode;
+  }
+
+  if (link.status !== 'linked') {
+    // This branching is needed or we need another TS guard to handle this
```
jaredpalmer: See comment @TooTallNate . 

Lmk if you'd rather use a `!` beneath and remove this logic branch

TooTallNate: LGTM as-is.

packages/cli/package.json
```diff
@@ -133,6 +133,7 @@
     "load-json-file": "3.0.0",
     "mime-types": "2.1.24",
     "minimatch": "3.0.4",
+    "mock-stdin": "1.0.0",
```
TooTallNate: Seems like this isn't being used? Let's remove if not.

packages/cli/package.json
```diff
@@ -155,6 +156,7 @@
     "test-listen": "1.1.0",
     "text-table": "0.2.0",
     "title": "3.4.1",
+    "tmp": "0.2.1",
     "tmp-promise": "1.0.3",
```
styfle: We already have `tmp-promise`, can that be used instead of `tmp`?

jaredpalmer: will give it a shot

packages/cli/src/commands/deploy/args.ts
```diff
@@ -21,6 +21,7 @@ export const help = () => `
       link                             Link local directory to a Vercel Project
       login                [email]     Logs into your account or creates a new one
       logout                           Logs out of your account
+      pull                 [filename]  Pull the Environment Variables and settings for your current Project
```
styfle:
```suggestion
      pull                 [filename]  Pull your Project Settings from the cloud
```

I think env vars are included in Project Settings 

image

packages/cli/test/mocks/project.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const defaultProject = {
+  id: 'foo',
+  name: 'cli',
+  accountId: 'K4amb7K9dAt5R2vBJWF32bmY',
+  createdAt: 1555413045188,
+  updatedAt: 1555413045188,
+  env: [
+    {
+      type: 'secret',
```
styfle: We don't really use `type=secret` anymore. Probably should use `type=encrypted`

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+    -Q ${chalk.bold.underline('DIR')}, --global-config=${chalk.bold.underline(
+    'DIR'
+  )}    Path to the global ${'`.vercel`'} directory
+    -d, --debug                    Debug mode [off]
+
+  ${chalk.dim('Examples:')}
+
+  ${chalk.gray(
+    '–'
+  )} Pull the latest Environment Variables and settings for your Project
```
styfle:
```suggestion
  )} Pull the latest Project Settings from the cloud
```

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+import { getPkgName } from '../util/pkg-name';
+import {
+  getLinkedProject,
+  VERCEL_DIR,
+  VERCEL_DIR_PROJECT,
+} from '../util/projects/link';
+import pull from './env/pull';
+
+const help = () => {
+  // @todo help output
```
styfle: What's let todo here?

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+  )} Pull the latest Environment Variables and settings for your Project
+
+    ${chalk.cyan(`$ ${getPkgName()} pull`)}
+`);
+};
+export default async function main(client: Client) {
+  let argv;
+  let debug;
+  let yes;
+  let link: ProjectLinkResult;
```
styfle: Generally we don't define variables before their used. I think you can move all of these down except for `argv`

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+  const { project, org } = link;
+
+  const result = await pull(
+    client,
+    project,
+    { '--yes': yes, '--debug': debug },
+    [join(cwd, '.env')],
+    client.output
+  );
+  if (result != 0) {
```
styfle:
```suggestion
  if (result !== 0) {
```

packages/cli/src/commands/pull.ts
jaredpalmer: Verify that all options are correctly documented

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+    return link.exitCode;
+  }
+
+  const { project, org } = link;
+
+  const result = await pull(
+    client,
+    project,
+    { '--yes': yes, '--debug': debug },
+    [join(cwd, '.env')],
```
styfle: Maybe we shouldn't write to the `.env` file and instead write env vars to the `project.json` file

That way everything is in one place and we can avoid having to solve the `.env` vs `.env.local` debate that we haven't really solved yet.

cc @TooTallNate

jaredpalmer: Yeah definitely some asymmetry because `vc dev` which seems to just load env vars into memory and yet this serializes to .env. I guess my only problem with project.json is that (most) users don't usually open the `.vercel` directory, but they are likely way more familiar with editing stuff in `.env`.

styfle: Hmm, I see your point. There's the scenario when system env vars have blank values and then users typically want to edit them.

The problem with this PR as-is would be a user who already ran `vercel env pull .env.local` today like many Next.js users because thats how we document it https://nextjs.org/docs/basic-features/environment-variables#environment-variables-on-vercel

They might want to run `vercel pull` but then they would end up with an unused `.env` file. Or actually I think it would override the `.env.local` (I can't remember which takes precedence).

jaredpalmer: IIRC in our initial kick off call, @leo mentioned that `vc pull` command would ultimately deprecate `vc env pull`, but not to deprecate it yet. 

I believe this PR should work w/ `vercel pull .env.local`

jaredpalmer: @styfle @leo update: the current implementation does not work w/ `vercel pull .env.local` only `vc pull <directory>`. The conflict is that `link` expects a directory as argument and `pull` expects a filename. 

Should we allow users to customize both? and if so, how?

```sh
vc pull . .env.local # order matters
vc pull . --filename=".env.local" # a flag for one of them
vc pull --cwd="." --filename=".env.local" # a flag for both of them
```

or some other thing? Let me know

styfle: I think I like the last option. It doesn't really follow our conventions but I always thought it was strange to take a directory as a parameter anyway.

I would also call the second param `--env`

```sh
vc pull --cwd="." --env=".env.local"
```

Thoughts @TooTallNate @leo ?

TooTallNate: I think `--cwd` is inconsistent with `vc deploy` / `vc link` / etc. Should be the first argument.

`--env` sounds good to me, or I was also thinking `--envfile`.

packages/cli/test/commands/pull.test.ts
```diff
@@ -14,13 +14,13 @@ describe('pull', () => {
...
-      const cwd = setupFixture('now-dev-next');
+      const cwd = setupFixture('now-pull-next');
       useUser();
       useTeams();
       useProject({
         ...defaultProject,
-        id: 'now-dev-next',
-        name: 'now-dev-next',
+        id: 'now-pull-next',
+        name: 'now-pull-next',
```
styfle: We shouldn't add new tests with `now-` prefix

packages/cli/test/fixtures/unit/now-pull-next/next.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  target: 'serverless',
+};
```
styfle: We shouldn't add new tests with `serverless`

packages/cli/test/fixtures/unit/now-pull-next/vercel.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 2,
+  "name": "now-dev-next",
```
styfle: the name here doesn't match the directory

packages/cli/src/commands/pull.ts
jaredpalmer: Added `--env` flag and test

packages/cli/test/fixtures/unit/now-pull-next/next.config.js
jaredpalmer: yep, just copied from that one

packages/cli/src/commands/env/index.ts
```diff
@@ -139,6 +140,14 @@ export default async function main(client: Client) {
...
         return rm(client, project, argv, args, output);
       case 'pull':
+        output.warn(
+          `${getCommandName(
+            'env pull'
+          )} will be deprecated in future releases. Run ${getCommandName(
+            'pull'
+          )} instead.`
+        );
+        await writeProjectSettings(process.cwd(), project, org);
```
jaredpalmer: @TooTallNate @styfle I am deprecating `vc env pull` with warning but also making it equivalent to `vc pull` here with the `writeProjectSettings()` call. Just want to confirm this is what we want?

packages/cli/src/util/projects/write-project-settings.ts
```diff
@@ -0,0 +1,28 @@
...
+import { join } from 'path';
+
+// writeProjectSettings writes the project configuration to `vercel/project.json`
+// that is needed for `vercel build` and `vercel dev` commands
+export async function writeProjectSettings(
+  cwd: string,
+  project: Project,
+  org: Org
+) {
+  return await writeFile(
```
jaredpalmer, line 13: @TooTallNate @styfle what should error handling look like here? As it stands now, it will explode the calling command if the `writeFile` fails. Let me know what you think UX should be.

packages/cli/src/commands/deploy/args.ts
```diff
@@ -21,6 +21,7 @@ export const help = () => `
       link                             Link local directory to a Vercel Project
       login                [email]     Logs into your account or creates a new one
       logout                           Logs out of your account
+      pull                 [filename]  Pull your Project Settings from the cloud
```
TooTallNate: Deploy uses `[path]`. `link` doesn't specify a sub-argument at all. We should probably at least be consistent here. How about:

```suggestion
      link                 [path]      Link local directory to a Vercel Project
      login                [email]     Logs into your account or creates a new one
      logout                           Logs out of your account
      pull                 [path]      Pull your Project Settings from the cloud
```

packages/cli/src/commands/env/pull.ts
```diff
@@ -56,8 +56,12 @@ export default async function pull(
     return 1;
   }
 
+  // handle relative or absolute filename
   const [filename = '.env'] = args;
-  const fullPath = join(process.cwd(), filename);
+  let fullPath = isAbsolute(filename)
+    ? filename
+    : join(process.cwd(), filename);
+
```
TooTallNate: We can use `path.resolve()` here since it already handles cwd when given a relative path:

```suggestion
  const fullPath = resolve(filename);
```

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,91 @@
...
+ ${chalk.dim('Options:')}
+
+    -h, --help                     Output usage information
+    -A ${chalk.bold.underline('FILE')}, --local-config=${chalk.bold.underline(
+    'FILE'
+  )}   Path to the local ${'`vercel.json`'} file
+    -Q ${chalk.bold.underline('DIR')}, --global-config=${chalk.bold.underline(
+    'DIR'
+  )}    Path to the global ${'`.vercel`'} directory
+    -d, --debug                    Debug mode [off]
```
TooTallNate, line 32: Need to document `--env` and `--yes`.

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,91 @@
...
+import setupAndLink from '../util/link/setup-and-link';
+import logo from '../util/output/logo';
+import { getPkgName } from '../util/pkg-name';
+import { getLinkedProject } from '../util/projects/link';
+import { writeProjectSettings } from '../util/projects/write-project-settings';
+import pull from './env/pull';
+
+const help = () => {
+  return console.log(`
+  ${chalk.bold(`${logo} ${getPkgName()} pull`)} [filename]
```
TooTallNate:
```suggestion
  ${chalk.bold(`${logo} ${getPkgName()} pull`)} [path]
```

packages/cli/src/util/projects/write-project-settings.ts
```diff
@@ -0,0 +1,28 @@
+import { writeFile } from 'fs-extra';
+import { Org, Project } from '../../types';
+import { VERCEL_DIR, VERCEL_DIR_PROJECT } from './link';
+import { join } from 'path';
+
+// writeProjectSettings writes the project configuration to `vercel/project.json`
```
TooTallNate:
```suggestion
// Write the project configuration to `.vercel/project.json`
```

packages/cli/src/util/projects/write-project-settings.ts
TooTallNate: As far as the code here is concerned, I don't think we should do anything extra for error handling.

If we want to make filesystem errors more "pretty" then I think that logic would belong centralized in `src/index.ts` for all commands to benefit from.

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,91 @@
...
+    [join(cwd, env)],
+    client.output
+  );
+  if (result !== 0) {
+    // an error happened
+    return result;
+  }
+
+  await writeProjectSettings(cwd, project, org);
+
```
TooTallNate, line 103: We should probably log a success message here (unless `writeProjectSettings()` already does that.

packages/cli/test/commands/pull.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { defaultProject, useProject } from '../mocks/project';
+import { useTeams } from '../mocks/team';
+import { useUser } from '../mocks/user';
+import fs from 'fs-extra';
+import path from 'path';
+
+describe('pull', () => {
+  afterAll(() => {
+    cleanupFixtures();
+  });
```
TooTallNate: I think it's possible to move this code into `setup-fixture.ts` so that tests don't need to copy/paste this block every time.

packages/cli/src/commands/env/index.ts
TooTallNate: The warning is good. But curious why we want to change the behavior. Seems kind of unexpected for a command that is specifically about "env".

packages/cli/src/commands/env/index.ts
```diff
@@ -139,6 +140,14 @@ export default async function main(client: Client) {
       case 'rm':
         return rm(client, project, argv, args, output);
       case 'pull':
+        output.warn(
+          `${getCommandName(
+            'env pull'
+          )} will be deprecated in future releases. Run ${getCommandName(
```
styfle:
```suggestion
          )} is deprecated and will be removed in future releases. Run ${getCommandName(
```

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (link.status === 'error') {
+    return link.exitCode;
+  }
+
+  const { project, org } = link;
+  const result = await pull(
+    client,
+    project,
+    { '--yes': yes, '--debug': debug },
```
styfle:
```suggestion
    argv,
```

Would it make sense to pass `argv` here?

styfle (approved):
Thanks! 🎉 

Remember to update the docs to add deprecation notes as well as the new command here: https://vercel.com/docs/cli

packages/cli/test/commands/pull.test.ts
jaredpalmer: Makes sense.

packages/cli/src/commands/env/index.ts
jaredpalmer: Okay will remove this, but keep warning.

TooTallNate (approved):

PR #6834, opened 2021-10-13
[build-utils] Add support for PNPM

jaredpalmer:
Add support for `pnpm` as well as new fixtures for `pnpm` with and without workspaces

#### Tests

- [x] All tests pass locally with `yarn test-unit && yarn test-integration-once` 

#### Code Review


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -348,6 +358,13 @@ export async function runNpmInstall(
       env.PATH = `/node16/bin-npm7:${env.PATH}`;
       console.log('Detected `package-lock.json` generated by npm 7...');
     }
+  } else if (cliType === 'pnpm') {
+    // PNPM's install command is similar to NPM's but without the audit nonsense
+    // @see options https://pnpm.io/cli/install
+    opts.prettyCommand = 'pnpm install';
+    commandArgs = args
+      .filter(a => a !== '--prefer-offline')
+      .concat(['install', '--unsafe-perm']);
```
styfle, line 360: Did you add `pnpm` to the build image? It needs to be a global for this to work

jaredpalmer: Yes I did.

jaredpalmer: It has not merged yet though. It is in lhr1

styfle: The build image needs to be merged and tagged before build-utils can ship

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -348,6 +358,14 @@ export async function runNpmInstall(
       env.PATH = `/node16/bin-npm7:${env.PATH}`;
       console.log('Detected `package-lock.json` generated by npm 7...');
     }
+  } else if (cliType === 'pnpm') {
+    await execCommand('pnpm config set store-dir .pnpm-store');
```
styfle: Does this config need to be executed each time? Perhaps this should happen once in the build image? Or could we use an env var instead?

packages/build-utils/test/fixtures/22-pnpm/vercel.json
```diff
@@ -0,0 +1 @@
+{}
```
styfle: This file can be deleted, right?

packages/build-utils/test/integration.test.ts
```diff
@@ -31,6 +31,7 @@ const skipFixtures: string[] = [
   '07-zero-config-jekyll',
   '08-zero-config-middleman',
   '21-npm-workspaces',
+  '23-pnpm-workspaces',
```
styfle, line 34: If we're not deploying, then how do we know it works correctly?

The unit tests don't actually run `pnpm install` so I think we need a test to verify that it works properly with the given flags.

packages/build-utils/src/fs/run-user-scripts.ts
Ethan-Arrowood: as far as I can tell this is a per-project configuration. `pnpm config` is actually an alias to `npm config` if I understand their docs correctly. Thus, I believe you are correct and that this command only needs to be executed once.

> Perhaps this should happen once in the build image

So should this be executed via docker? Or is there another place in the code base for it?

> Or could we use an env var instead? 

We could create an env var for the path so users can override, but we should default it to `.pnpm-store`. @jaredpalmer mentioned an idea of putting it inside of `node_modules` potentially too

packages/build-utils/test/fixtures/22-pnpm/vercel.json
Ethan-Arrowood: Unfortunately no, I was getting an undefined error here: https://github.com/vercel/vercel/blob/2a24210b7d3e1412bb6877358bb483c892ad3f72/test/lib/deployment/test-deployment.js#L57-L61 and the test was failing.

Would you prefer a change in that file?

packages/build-utils/test/integration.test.ts
Ethan-Arrowood: > If we're not deploying, then how do we know it works correctly?

Ah that must be my misunderstanding. Do these integration tests not execute pnpm install when they are deployed?

> The unit tests don't actually run pnpm install so I think we need a test to verify that it works properly with the given flags.

I'll see if I can write some unit tests that execute `pnpm install`

styfle: Look at the `20-npm-7` test fixture for an example text that actually deploys.

Note the `probes` are the parts that actually test the deployment req/res.

https://github.com/vercel/vercel/blob/6ef2c16d631c28e6282df473faefa44165031ef5/packages/build-utils/test/fixtures/20-npm-7/vercel.json#L1-L5

packages/build-utils/src/fs/run-user-scripts.ts
styfle: > this is a per-project configuration

Is it per-project or per-machine? Where is the config stored?

Ethan-Arrowood: Ahh good point, per-machine

Ethan-Arrowood: > Where is the config stored?

it has the same behavior as `.npmrc`. If a global one exists, it uses that, otherwise it looks for one in the project scope/directory

packages/build-utils/test/integration.test.ts
Ethan-Arrowood: fixed this part in latest commit

packages/build-utils/src/fs/run-user-scripts.ts
Ethan-Arrowood: Is the resolution here a change in the build-container Docker file? cc @jaredpalmer

jaredpalmer: pnpm is per-machine is by default IIRC, but I'm not personally familiar enough with our build container and infra to understand security implications of ~/.pnpm-store. Looping in @gudmundur @igorklopov for their opinions.

I think for now, to unblock and unlock ARR, we can do this per-project even though it's obviously suboptimal from what is possible

styfle: From my understanding, the default store is in the home directory at `~/.pnpm-store` and the command you're running would put the store in the local project directory ` ./.pnpm-store`.

https://pnpm.io/npmrc#store-dir

This behavior is actually worse than default because you're creating a new store for every project and the store isn't cached so you're guaranteeing a new store every time.

What is the goal of `pnpm config set store-dir .pnpm-store`? Can it be removed?

packages/build-utils/test/fixtures/22-pnpm/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "22-pnpm",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "mkdir -p public && (printf \"pnpm version: \" && pnpm -v) > public/index.txt"
```
styfle, line 7:
```suggestion
    "build": "mkdir -p public && printf \"pnpm version: \" && pnpm -v > public/index.txt"
```

The paranthesis aren't needed here

packages/build-utils/test/fixtures/22-pnpm/vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "version": 2,
+  "builds": [{ "src": "package.json", "use": "@vercel/static-build" }],
+  "probes": [{ "path": "/", "mustContain": "pnpm version: 6.32.1" }]
```
styfle:
```suggestion
  "probes": [{ "path": "/", "mustContain": "pnpm version: 6" }]
```

Lets test only major since we might be bumping the minor and patch version frequently

packages/build-utils/src/fs/run-user-scripts.ts
igorklopov: We can place it somewhere here depending on what it is going to contain https://github.com/vercel/api/blob/412e6f6814e85ad8045d25a9e1736e791d0e228e/scripts/build-container/src/utils/npm-utils.ts#L252. Either we run `pnpm config set store-dir .pnpm-store` here or add a corresponding config line or even leave it default to `~/.pnpm-store`.

igorklopov: Per-machine equals per-deployment so i don't see any security issues, given `~/.pnpm-store` is empty at start.

styfle (approved):

PR #6838, opened 2021-10-13
[cli] Use getters for `Now` props that mirror `Client`

TooTallNate:
This will prevent any issues where the `Now` instance is out-of-sync with the `Client` instance, for example, during a re-auth where a new auth token is issued. Also reduces a bit of code which is nice.

### DIFF
--- a/packages/cli/src/commands/billing/index.js
+++ b/packages/cli/src/commands/billing/index.js
@@ -51,8 +51,6 @@ const help = () => {
 };
 
 let argv;
-let debug;
-let apiUrl;
 let subcommand;
 
 export default async client => {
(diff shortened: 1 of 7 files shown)


AndyBitz (approved):

packages/cli/src/commands/billing/index.js
```diff
@@ -76,17 +72,13 @@ export default async client => {
 
   const {
     output,
-    authConfig: { token },
     config: { currentTeam },
   } = client;
 
   const start = new Date();
   const creditCards = new NowCreditCards({
```
styfle, line 79: Nice catch with this one inheriting from `Now` class.

We might want to convert `NowCreditCards` to TS in the future 👍

styfle (approved):

packages/cli/src/commands/billing/index.js
TooTallNate: Yup, definitely. Actually it should just be turned into a couple helper functions that take `Client` as a param.

PR #6861, opened 2021-10-18
[cli] Add `vercel build`

jaredpalmer:
Add new `vercel build` command

### 📋 Checklist

- [x] Pretty print configuration/settings/environment for observability
- [x] Pretty print FS API output for observabilty
- [x] Detect and run build plugins

#### Tests

#### Code Review


packages/cli/package.json
```diff
@@ -76,6 +79,7 @@
     "@types/node": "11.11.0",
     "@types/node-fetch": "2.5.10",
     "@types/npm-package-arg": "6.1.0",
+    "@types/p-limit": "2.2.0",
```
TooTallNate: Remove?

```suggestion
```

packages/cli/src/commands/build.ts
```diff
@@ -0,0 +1,608 @@
...
+      cwd: cwd,
+    });
+  } else {
+    result = await runPackageJsonScript(
+      client,
+      cwd,
+      ['vercel-build', 'now-build', 'build'],
+      spawnOpts
+    );
+  }
```
AndyBitz: In the case that `framework` is `null`, and no `package.json` exists, can we still set `result` to `true` to avoid failing because no `package.json` exists?

jaredpalmer: Yep

packages/frameworks/src/frameworks.ts
```diff
@@ -323,6 +344,52 @@ export const frameworks = [
 
       return base;
     },
+    defaultHeaders: [
+      {
+        source: '^/[^./]+\\.[0-9a-f]{8}\\.(css|js)$',
+        regex: '^/[^./]+\\.[0-9a-f]{8}\\.(css|js)$',
+        headers: { 'cache-control': 'max-age=31536000, immutable' },
+        continue: true,
+      },
```
AndyBitz: That's not quite correct, take a look at this example: https://github.com/curated-tests/new-file-system-api/blob/main/projects/all-routes/.output/routes-manifest.json#L18-L27

`headers` is an array of objects that have a `key` and `value` property.

AndyBitz: `continue` will also not be required (or supported).

jaredpalmer: The link doesn't work

AndyBitz: @jaredpalmer invited you

jaredpalmer: @AndyBitz review now

AndyBitz (approved):

AndyBitz (approved):

PR #6866, opened 2021-10-19
[cli] Add CLI login unit test

TooTallNate:
This has just been sitting in my working tree for a while, might as well push it up.

### DIFF
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -100,7 +100,7 @@ export default class Client extends EventEmitter {
     let body;
     if (isJSONObject(opts.body)) {
       body = JSON.stringify(opts.body);
-      headers.set('content-type', 'application/json; charset=utf8');
+      headers.set('content-type', 'application/json; charset=utf-8');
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/client.ts
```diff
@@ -100,7 +100,7 @@ export default class Client extends EventEmitter {
     let body;
     if (isJSONObject(opts.body)) {
       body = JSON.stringify(opts.body);
-      headers.set('content-type', 'application/json; charset=utf8');
+      headers.set('content-type', 'application/json; charset=utf-8');
```
TooTallNate, line 103: Needed because express' `body-parser` is strict about this (which we use for the API test mocks - this is the first test that does a `POST` apparently).

AndyBitz (approved):

packages/cli/src/util/client.ts
styfle: *shakes fist at express*

old-man

styfle (approved):

PR #6874, opened 2021-10-21
[node] Create `vercel-plugin-node`

TooTallNate:
This PR adds the `vercel-plugin-node` CLI Plugin.

Originally this replaced the `@vercel/node runtime`, but we'll keep both for now instead.


packages/node/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@vercel/node",
-  "version": "1.12.2-canary.4",
+  "name": "vercel-plugin-node",
```
styfle: I think we should leave `@vercel/node` as-is and instead introduce a new package `vercel-plugin-node`.

packages/cli/package.json
```diff
@@ -48,7 +48,9 @@
     "@vercel/node": "1.12.2-canary.4",
     "@vercel/python": "2.0.6-canary.4",
     "@vercel/ruby": "1.2.8-canary.3",
-    "update-notifier": "4.1.0"
+    "update-notifier": "4.1.0",
+    "vercel-plugin-middleware": "0.0.0-canary.5",
+    "vercel-plugin-node": "1.12.2-plugin.6"
```
styfle: This should be `-canary`

PR #6880, opened 2021-10-22
Add support for "_middleware.{js,ts}" to `vercel dev`

gdborton:
### Related Issues

This adds support for running _middleware.{js,ts} in the `vc dev` command.

#### Tests

#### Code Review


packages/middleware/src/index.ts
```diff
@@ -37,25 +49,26 @@ async function getMiddlewareFile() {
   return middlewareFiles[0];
 }
 
-export async function build() {
-  const middlewareFile = await getMiddlewareFile();
+export async function build({ workPath }: BuildOptions) {
```
TooTallNate: I was trying to avoid having to specify `workPath` explicitly since it should be able to be inferred from cwd. But let's leave it since it'll make writing tests easier (makes the function more "pure").

Using `BuildOptions` interface though isn't quite correct, since that's the interface for the Runtimes' `build()` function, not the Plugins' `build()` function. So far, this `workPath` is the only prop that needs to be passed in to the Plugin build fn.

packages/middleware/src/index.ts
```diff
@@ -73,14 +86,249 @@ export async function build() {
...
+    !result.response.headers.has('Location')
+  ) {
+    result.response.headers.set('x-middleware-refresh', '1');
+  }
+
+  result.response.headers.delete('x-middleware-next');
+
+  // @ts-ignore
+  for (const [key, value] of result.response.headers.entries()) {
+    if (key !== 'content-encoding') {
```
TooTallNate: How come `content-encoding` is special-cased here?

packages/middleware/src/index.ts
```diff
@@ -73,14 +86,249 @@ export async function build() {
...
+    !result.response.headers.has('x-middleware-rewrite') &&
+    !result.response.headers.has('x-middleware-next') &&
+    !result.response.headers.has('Location')
+  ) {
+    result.response.headers.set('x-middleware-refresh', '1');
+  }
+
+  result.response.headers.delete('x-middleware-next');
+
+  // @ts-ignore
```
TooTallNate: Is the ts-ignore because of `entries()` not existing/being iterable on Headers? I think we need to add `dom.iterable` to "lib" in tsconfig.

packages/middleware/src/index.ts
gdborton: Good question, this was pulled from next.js:
https://github.com/vercel/next.js/blob/canary/packages/next/server/next-server.ts#L1175

Looks like there was an update to the for loop as well that'll need to be pulled in.

packages/middleware/src/index.ts
gdborton: Yup
```
Property 'entries' does not exist on type 'Headers'.
```

adding dom.iterable fixed.

packages/cli/src/commands/build.ts
```diff
@@ -281,24 +281,12 @@ export default async function main(client: Client) {
...
       cwd,
       ['vercel-build', 'now-build', 'build'],
       spawnOpts
     );
-  } else {
-    // no package.json exists and no build command present
-    result = true;
-  }
-
-  if (!result) {
```
TooTallNate, line 295: We removed this error message because it's possible you have a `package.json` and no build script defined (i.e. I want to have deps in my middleware or API file).

PR #6889, opened 2021-10-26
[cli][client] Add step for checks

AndyBitz:
### Related Issues
https://github.com/vercel/runtimes/issues/201

Shows an error message in case the checks failed and a spinner while they're running.

It won't show any new information when there are no checks, if they're skipped, or if they all succeed, since it wouldn't really fit into the `Inspect, Production/Preview` view.

For the error we also don't need an extra link, because the inspect link above will be the only relevant one.

#### Tests

#### Code Review


packages/cli/src/commands/deploy/index.ts
```diff
@@ -529,6 +530,20 @@ export default async (client: Client) => {
...
+      const { checks } = await getDeploymentChecks(client, deployment.id);
+      const counters = new Map<string, number>();
+      checks.forEach(c => {
+        counters.set(c.conclusion, (counters.get(c.conclusion) ?? 0) + 1);
+      });
+
+      const counterList = Array.from(counters)
+        .map(([name, no]) => `${no} ${name}`)
+        .join(', ');
+      error(`Running Checks: ${counterList}`);
```
chriswdmr: Why do we `error` here? (Sorry I'm not familiar with the CLI)

AndyBitz: Because the checks failed and the domains won't get assigned.

It wouldn't make sense to wait for someone to skip the checks, because that might never happen.
And showing just the preview domain for something that might be a production Deployment would also not make much sense.

And above the error they'll still see the URL that takes them to the Deployment view on the Dashboard, where they can check why exactly the tests failed.

AndyBitz: Though, maybe it'd make sense to add a flag to the CLI to skip the checks, and not rely on the Dashboard for it?

AndyBitz: Thinking more about it, maybe that's not a good idea, since it'd make it easier to publish a broken Deployment, clicking the Dashboard would at least force someone to see which checks failed.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -186,6 +186,15 @@ export default async function processDeployment({
         output.spinner('Completing', 0);
       }
 
+      if (event.type === 'checks-running') {
+        output.spinner('Running Checks', 0);
+      }
+
+      if (event.type === 'checks-conclusion-failed') {
+        output.stopSpinner();
+        return event.payload;
```
styfle, line 202: Is the payload an error here?

AndyBitz: No, it's the Deployment. Since even if the checks failed, the Deployment can still succeed, but the aliases won't get assigned.

packages/client/src/check-deployment-status.ts
```diff
@@ -88,6 +88,42 @@ export async function* checkDeploymentStatus(
       yield { type: 'ready', payload: deploymentUpdate };
     }
 
+    if (deploymentUpdate.checksState !== undefined) {
+      if (deploymentUpdate.checksState === 'completed') {
+        finishedEvents.add('checks-completed');
```
styfle, line 96: You're adding to the finishedEvents set but never reading from it.

Seems like it should either check for existence or just remove these lines since the values aren't used

AndyBitz: Do you know what `finishedEvents` is used for? I just did what all other events do, I though that it's a way to just keep track of which events were already yielded.

styfle: I don't recall, but looking at the code it seems to be used to avoid yielding duplicate events

AndyBitz: In that case it'd be expected to add it to `finishedEvents`, no?

AndyBitz: Oh, I see

AndyBitz: Updated

packages/cli/src/commands/deploy/index.ts
styfle:
```suggestion
      output.error(`Running Checks: ${counterList}`);
```

styfle (approved):

styfle (approved):

TooTallNate (approved):

PR #6896, opened 2021-10-27
[cli] Fix tracing for `vc build`

AndyBitz:
### Related Issues
Ensures the `input` and `output` path are correct for `vercel build` in `.nft.json` files.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -63,7 +63,7 @@ const help = () => {
   ${chalk.gray('–')} Build the project
 
     ${chalk.cyan(`$ ${getPkgName()} build`)}
-    ${chalk.cyan(`$ ${getPkgName()} build --cwd ./path-to-project`)}    
+    ${chalk.cyan(`$ ${getPkgName()} build --cwd ./path-to-project`)}
 `);
 };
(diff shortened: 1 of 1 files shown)


jaredpalmer (approved):

packages/cli/src/commands/build.ts
```diff
@@ -395,27 +395,25 @@ export default async function main(client: Client) {
     for (let f of files) {
       client.output.debug(`Processing ${f}:`);
       const json = await fs.readJson(f);
-      const newFilesList: Array<{ input: string; output: string }> = [];
+      const newFilesList: (string | { input: string; output: string })[] = [];
```
jaredpalmer, line 398: in some versions of next.js, i was seeing this as `string`? maybe that's been changed?

jaredpalmer: Nvm I see this is for `newFilesList`. Got confused for a moment

AndyBitz: Next.js always outputs as `files: string[]`, but the File System API also supports `{ input: string; output: string; }[]`, so we can leave the entries that are unaffected as they are.

PR #6898, opened 2021-10-27
[cli] Improve tracing in vc build

jaredpalmer:
#### Tests

#### Code Review


packages/cli/src/commands/build.ts
```diff
@@ -606,3 +627,53 @@ async function glob(pattern: string, options: GlobOptions): Promise<string[]> {
...
 }
+
+interface NftFile {
+  version: number;
+  files: (string | { input: string; output: string })[];
+}
+
+// resolveNftToOutput takes nft file and moves all of its traces files
+// into the specified directory + "inputs", (renaming them to there hash + ext) and
+// subsequently updating the original nft file accordingly.
```
TooTallNate:
```suggestion
// subsequently updating the original nft file accordingly. This is done
// to make the `.output` directory be self-contained, so that it works
// properly with `vc --prebuilt`.
```

TooTallNate (approved):
Code looks good to me. Let's just add some comments explaining why we're special-casing Next.js and explain the 0 NFT files special-case as well.

leo:
> Which versions of Next.js does this actually work for?

@jaredpalmer Let's do it based on whether NFT files are present, IMO. We don't have the framework version available in the Project Settings and also regardless, I think that's a better idea. No need to read `package.json`.

packages/cli/src/commands/build.ts
```diff
@@ -382,44 +383,156 @@ export default async function main(client: Client) {
...
+          smartCopy(
+            client,
+            f,
+            f.replace(
+              join(OUTPUT_DIR, 'static'),
+              join(OUTPUT_DIR, 'static', '_next', 'static')
+            )
+          )
+        )
+      );
```
AndyBitz: @jaredpalmer This should happen before we move the other files from `public` and `static` into it, right? Otherwise we'd copy the already copied files?

AndyBitz (approved):

PR #6909, opened 2021-10-29
add Yarn (berry) as a framework option

meirish:
We recently moved to using Yarn 3 for our application and are seeing an issue where when there's a cache, the build errors and we have to redeploy without the cache. Turns out this is because yarn monorepos uner Yarn 3 don't have a single top-level `node_modules` directory, but possibly have nested `node_modules` directories in the different packages directories ([more info here](https://github.com/yarnpkg/berry/issues/3270#issuecomment-897579867)). 

Based on the discussion here: 
https://github.com/yarnpkg/berry/discussions/2621#discussioncomment-505872
only caching `.yarn/cache` is the recommended way to do this.

This PR adds that ability via a `Yarn (berry)` framework option.
 
I wasn't sure if you would want this as a framework preset, or if the default caching mechanism in Vercel should change. Easy enough to go whichever way you decide.

Thanks! 

#### Tests

#### Code Review


meirish:
Looks like there's a new API to do this natively in vercel: https://vercel.com/docs/file-system-api#configuration/build

thanks folks!

PR #6932, opened 2021-11-02
Bundle middleware deps

gdborton:
### Related Issues

This PR is meant to address 2 issues.

The first is unbundled dependencies in the middleware-plugin (fixed by using ncc vs tsc for building, and by unmarking the plugins as external).

The second is deprecation warnings on install, fixed by replacing globby with glob, and removing the external querystring lib in favor of the node built in.

#### Tests

#### Code Review

### DIFF
--- a/packages/middleware/build.js
+++ b/packages/middleware/build.js
@@ -4,14 +4,17 @@ const execa = require('execa');
 const { join } = require('path');
 
 async function main() {
+  const srcDir = join(__dirname, 'src');
   const outDir = join(__dirname, 'dist');
 
   // Start fresh
   await fs.remove(outDir);
 
-  await execa('tsc', [], {
-    stdio: 'inherit',
-  });
+  await execa(
+    'ncc',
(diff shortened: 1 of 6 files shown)


packages/middleware/build.js
```diff
@@ -4,14 +4,17 @@ const execa = require('execa');
...
 async function main() {
+  const srcDir = join(__dirname, 'src');
   const outDir = join(__dirname, 'dist');
 
   // Start fresh
   await fs.remove(outDir);
 
-  await execa('tsc', [], {
-    stdio: 'inherit',
-  });
```
styfle, line 14: Do we export any types from this package?

That might be a reason to keep `tsc` but I'm guessing its safe to remove

styfle (approved):

packages/middleware/build.js
gdborton: Nope, the previous tsc command was actually changed to by me when the websandbox was a separate package. Didn't have enough context into how this was built at the time.

AndyBitz (approved):

PR #6934, opened 2021-11-02
[cli] Render an error page when middleware explodes.

gdborton:
### Related Issues

This adds error handling for middleware in `vc dev`

Local looks a little different than production:
Local
image

Production
image

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1421,6 +1421,16 @@ export default class DevServer {
     const middlewareResult = await runDevMiddleware(req, res, this.cwd);
 
     if (middlewareResult) {
+      if (middlewareResult.error) {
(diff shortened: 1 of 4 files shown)


styfle (commented):
Please add a test so we don't regress

gdborton:
@styfle I added an integration test.  I'm not sure how it's generated, but I noticed that the response in the test has the error code that I'm passing to sendError. This isn't the same response as what I'd see in the browser, but I didn't look too deeply into it as it seems be validating what I was trying to test.

styfle (approved):
Thanks!

styfle:
@gdborton Its probably a different response because the `Accept` header isn't provided by the test but it is by your browser.

TooTallNate (approved):

PR #6937, opened 2021-11-03
Remove --cwd option from build in favor of existing patterns.

gdborton:
### Related Issues

I noticed while working on `vc dev` that the vc build command expects a `--cwd` flag for setting the working directory. Other commands already exist that expect it to be the final argument passed to the command.  This is especially apparent when `vc build` runs `vc pull` as it passes along the same options.

This change consolidates the apis to be the same.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -44,7 +44,7 @@ const sema = new Sema(16, {
 
 const help = () => {
   return console.log(`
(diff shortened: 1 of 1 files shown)


TooTallNate:
I believe that `--cwd` is that pattern that we'd like to migrate to, it's just been implemented in `vc build` first since it just got added recently.

gdborton:
@TooTallNate so should the other commands be updated then?

PR #6941, opened 2021-11-03
[cli] Run middleware as a plugin instead of running directly

gdborton:
### Related Issues

This removes the explicit call to `vercel-plugin-middleware`'s `runDevMiddleware` in favor of loading and running from plugins.

There is a slight deviation from the plugin api as outlined here:
https://github.com/vercel/runtimes/issues/179#issuecomment-946699178

I've added a function called `runDevMiddleware` that's detected as a type of plugin for running middleware in dev.  This is needed to ensure that the middleware is run BEFORE any other routing logic.

~I made an assumption that only one plugin at a time will be used for middleware logic, although I'm not sure if this is correct.~

#### Tests

#### Code Review


packages/cli/package.json
```diff
@@ -48,7 +48,9 @@
     "@vercel/node": "1.12.2-canary.4",
     "@vercel/python": "2.0.6-canary.4",
     "@vercel/ruby": "1.2.8-canary.3",
-    "update-notifier": "4.1.0"
+    "update-notifier": "4.1.0",
+    "vercel-plugin-middleware": "0.0.0-canary.5",
```
gdborton: These were added back to dependencies from devDependencies as they're never `import`ed directly and so were not getting bundled.

Also, the loading plugins currently matches against a project's package.json for dependencies/devDependencies and the CLI's package.json for dependencies.

packages/cli/src/util/plugins.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw error;
+    }
+  }
+
+  /**
+   * This is might be wrong... should we support more than one plugin defining runDevMiddleware?
+   * Should the middleware plugin ALWAYS be added? What about when a different framework defines
+   * a runDevMiddleware plugin?
+   */
+  if (devMiddlewarePlugins.length > 1) {
```
gdborton: @leo I'm not sure about the assumption that I'm making here.

gdborton: I'm thinking that if there is a plugin defining `runDevMiddleware` in the project's package.json, we should override the one we're providing in `vercel-plugin-middleware` as they likely don't want both.

packages/cli/src/util/plugins.ts
```diff
@@ -0,0 +1,91 @@
+import code from '../util/output/code';
+import { getColorForPkgName } from '../util/output/color-name-cache';
+import cliPkgJson from '../util/pkg';
+import { scanParentDirs } from '@vercel/build-utils';
+
+const VERCEL_PLUGIN_PREFIX = 'vercel-plugin-';
+
+export async function loadCliPlugins(
+  cwd: string,
+  logError: (errorMessage: string) => void = console.error
```
styfle: This should probably accept the `Output` object instead of a function

gdborton: Oh, I missed that it was even a thing in the server. I'll update.

packages/cli/package.json
```diff
@@ -48,7 +48,9 @@
...
+    "vercel-plugin-node": "1.12.2-plugin.6"
```
styfle: Why is this version so strange? Usually we use `-canary` suffix?

styfle: Its also 81 MB which seems wrong 

https://packagephobia.com/result?p=vercel-plugin-node@1.12.2-plugin.6

gdborton: That's a good question, looks like it was originally done here:
https://github.com/vercel/vercel/pull/6861

@jaredpalmer

packages/cli/src/util/dev/server.ts
```diff
@@ -1351,6 +1350,25 @@ export default class DevServer {
...
 
+  runDevMiddleware = async (
+    req: http.IncomingMessage,
+    res: http.ServerResponse
+  ) => {
+    const { devMiddlewarePlugins } = await loadCliPlugins(
+      this.cwd,
+      this.output
+    );
+    if (devMiddlewarePlugins.length) {
```
styfle: If we're only expecting one, then `loadCliPlugins` should probably only return one or zero.

`devMiddlewarePlugins` => `devMiddlewarePlugin`

packages/cli/src/util/plugins.ts
leo: I think it would be beneficial if multiple CLI Plugins are able to provide Middleware at the same time.

One of the big benefits of Edge Functions is that we're planning to let people run them "on top of each other", which allows for plug-and-play setups the world hasn't seen before – like one-click installation of Sentry through us automatically adding it as another Middleware, and since Middlewares can capture each other's responses, we can capture the crash stack trace of another middleware. Same for logs, etc.

gdborton: @leo Plugins would still be able to run multiple middlewares if needed, a plugin's `runDevMiddleware` would run them sequentially.  For example if next defined one, and had a filesystem like below, it'd run two middlewares for a request to `/nested` before returning the result:

```
pages/_middleware.ts
pages/nested/_middleware.ts
pages/nested/index.ts
```

BUT we probably shouldn't run `_middleware.ts` (if added to project root) via the `vercel-plugin-middleware`.

leo: Right, what I mean is this:

> I think it would be beneficial if multiple CLI Plugins are able to provide Middleware at the same time.

Not that one CLI Plugin would be able to run multiple Middlewares at the same time.

Basically, I'm thinking that, we don't yet know how Integrations like Sentry would be one-click-enabled, but if multiple CLI Plugins could provide Middleware, the user could provide some in `_middleware.js` and then we could auto-install `vercel-plugin-sentry`, which provides another one that runs on top, for example.

styfle (approved):

packages/cli/src/util/plugins.ts
gdborton: @leo I've updated to allow for more than one again, and the dev server now runs them sequentially until it finds a result w/ `finished: true` do you have an example of a middleware intercepting another middleware's response?

I adapted from next.js's implementation originally and doesn't look like multiple middlewares there are ever passed the result of a previous middleware... maybe that wasn't the best place to be looking?

https://github.com/vercel/next.js/blob/canary/packages/next/server/next-server.ts#L671

They also only seem to run multiple matching middlewares if the `x-middleware-next` header is present:
https://github.com/vercel/next.js/blob/canary/packages/next/server/next-server.ts#L687

Took a look through here for inspiration, but all the examples seem to use a single _middleware:
https://github.com/vercel/examples/tree/main/edge-functions

Just trying to verify correct behavior.

packages/cli/package.json
TooTallNate: I've changed it back to `canary` in the plugin-node PR, so next publish will use that tag.

packages/cli/src/util/plugins.ts
leo: What I mean when I say Middleware running on top of each other is them receiving each other's responses, not them invoking each other programmatically and receiving their return values.

Is `finished: true` needed? Can't we just run the Middleware of all the CLI Plugins and the one at which it stops is the last one? Because keep in mind that CLI Plugins aren't aware of each other, so they don't know what will be run on top or below them, so they never know whether they are last. Does that make sense?

In my mind, there basically wouldn't be a difference between a Middleware running on top of a Deployment directly, or on top of Middleware that runs on top of a Deployment. Both could be the case and the Middlewares can decide whether they want to handle both, or just one by looking at the shape of the response of what they're targeting. Things like crash stack traces or logs will be delivered via the response body or meta information (like headers). Logs are already delivered via headers for Lambdas today.

leo: Note that I'm not saying we need to make sure that things like logs and crash stack traces are already properly passed around – all I'm saying is that we want to make sure that Middlewares of different CLI Plugins should be able to run on top of each other and communicate via the requests, just like how they communicate with their underlying Deployment if there is no other Middleware underneath them.

gdborton: It's less about knowing whether you're the last middleware to be run, and more about whether the response has been completed.  If a middleware looks like this, I wouldn't expect any other middleware to be run:

```js
export default () => {
  return new Response('hey')
}
```

packages/cli/package.json
styfle: @TooTallNate What about the 81MB? Seems really large, doesn't seem right 🤔

styfle: Oh I guess its 69.7MB now https://packagephobia.com/result?p=vercel-plugin-node@1.12.2-canary.4

That's basically the same as `@vercel/node` https://packagephobia.com/result?p=@vercel/node@1.12.1

So we can revisit that in a different PR 👍

packages/cli/src/util/plugins.ts
AndyBitz: The current implementation would only continue with the next middleware if `x-middleware-next` is present as header, so your Response will indeed not continue, unless you'll add `new Response('hey', { headers: { 'x-middleware-next': '1' }})`.

AndyBitz (approved):

packages/cli/src/util/plugins.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  let pluginCount = 0;
+  const preBuildPlugins = [];
+  const buildPlugins = [];
+  const devServerPlugins = [];
+  const devMiddlewarePlugins = [];
+  const deps = new Set(
+    [
+      ...Object.keys(packageJson?.dependencies || {}),
+      ...Object.keys(packageJson?.devDependencies || {}),
```
styfle, line 20: Probably change the order of deps vs devDeps since I think deps should be able to override devDeps?

styfle (approved):

PR #6942, opened 2021-11-04
[build-utils] Add `convertRuntimeToPlugin()`

styfle:
Closes https://github.com/vercel/runtimes/issues/235.


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    const entrypoints = await glob(`api/**/*${ext}`, opts);
+    const dir = join(workPath, OUTPUT_DIR, 'inputs', 'runtime-temp');
+    await fs.ensureDir(dir);
+    for (const entrypoint of Object.keys(entrypoints)) {
+      const { output } = await buildRuntime({
+        files,
+        entrypoint,
+        workPath,
+        config: {}, // TODO: what should config be?
+        meta: {}, // TODO: what should meta be?
```
AndyBitz: Not sure about `meta`, but `config` should probably be what we use for `zero-config`.

styfle: Added in 5b9cc08

packages/build-utils/test/fixtures/20-npm-7/vercel.json
```diff
@@ -1,5 +1,5 @@
 {
   "version": 2,
   "builds": [{ "src": "package.json", "use": "@vercel/static-build" }],
-  "probes": [{ "path": "/", "mustContain": "npm version: 7" }]
+  "probes": [{ "path": "/", "mustContain": "npm version: 8" }]
```
styfle, line 4: @TooTallNate Looks like [Node.js 16](https://github.com/nodejs/node/releases/tag/v16.12.0) actually bumped a npm major. I don't think this will cause any problems but wanted to point it out.

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -0,0 +1,85 @@
...
+          throw new Error(`Unknown file type: ${file.type}`);
+        }
+      });
+
+      const nft = join(
+        workPath,
+        OUTPUT_DIR,
+        'server',
+        'pages',
+        entrypoint.replace(ext, '.nft.json')
```
AndyBitz: The extension must not be replaced, it just appends `.nft.json`, like this: `hello.js.nft.json`.

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -0,0 +1,85 @@
...
+    const files = await glob('**', opts);
+    const entrypoints = await glob(`api/**/*${ext}`, opts);
+    const dir = join(workPath, OUTPUT_DIR, 'inputs', 'runtime-temp');
+    await fs.ensureDir(dir);
+    for (const entrypoint of Object.keys(entrypoints)) {
+      const { output } = await buildRuntime({
+        files,
+        entrypoint,
+        workPath,
+        // TODO: What about includeFiles/excludeFiles?
```
AndyBitz: I don't think we can consider it in this case.

With the new system we'd expect the build process to take care of that, while the old system expected it to be there upfront.

Maybe we'd need to read `vercel.json` 🤔

packages/build-utils/src/convert-runtime-to-plugin.ts
styfle: Fixed in 101b292

packages/build-utils/src/convert-runtime-to-plugin.ts
styfle: I just realized there won't even be a vercel.json config to read because it used to read `functions` or `builds` but those won't exist for plugins.

And I think the plan is for legacy builders/runtimes to provide a new npm package so users would have to opt into the new plugin and thus we could change the way this works: perhaps the plugin could listen for environment variables?

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    const build = await convertRuntimeToPlugin(buildRuntime, '.py');
+    await build({ workPath });
+    const apiDir = join(workPath, '.output', 'server', 'pages', 'api');
+
+    const indexJson = await readFile(join(apiDir, 'index.py.nft.json'), 'utf8');
+    expect(JSON.parse(indexJson)).toMatchObject({
+      version: 1,
+      files: [
+        {
+          input: '../../../../inputs/runtime-temp/api/index.py',
```
TooTallNate: So this is the entrypoint file right? It already gets copied into the output lambda at `.output/server/pages/api/index.py`. So are we copying it again here because the default output path is not suitable?

styfle: This is the part I'm not sure about.

I didn't create a `.output/server/pages/api/index.py` file but I guess I could?

TooTallNate: The FS API already does that is what I'm saying. If you have `.output/server/pages/api/index.py` as the entrypoint in the output files, the FS API will include that entrypoint file at `.output/server/pages/api/index.py` _within the lambda_. This is because it's what Next.js expects, but IMO not a good default for serverless functions (where I'd expect the same path as the entrypoint source file i.e. `api/index.py`, not the output path).

So as a net result I believe the `index.py` entrypoint file will be included in the Lambda _twice_ (once at `.output/server/pages/api/index.py` as done implicitly by FS API), and again here in the nft file where it will be placed at `api/index.py`.

styfle: I replaced the need for nft.json in 1f03667

packages/build-utils/src/convert-runtime-to-plugin.ts
styfle: I updated the code to read `vercel.json` and handle functions in f0d8d3d

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
styfle: @TooTallNate I spoke to Andy and learned that we need to continue using nft.json so I added it back in c5599d9 along with the proper entrypoint file `.output/server/pages/api/index.py`

AndyBitz (approved):

PR #6952, opened 2021-11-05
[routing-utils] Add caseSensitive field to routes schema

ijjk:
### Related Issues

x-ref: https://github.com/vercel/customer-issues/issues/34

#### Tests

#### Code Review


packages/routing-utils/src/schemas.ts
```diff
@@ -95,6 +95,9 @@ export const routesSchema = {
               maxLength: 32,
             },
           },
+          caseSensitive: {
+            type: 'boolean',
+          },
```
styfle, line 100: Is this referring to the `src` or `dest` being case sensitive? I assume src but not sure if there is a spec for this new feature. Also I don't see tests for confirming it works?

ijjk: Tests were added in the linked proxy PR as we can't test the routing logic here, a test to ensure this validates against the schema was added in this PR though. This is for `src` being case-sensitive It's not exactly a new feature but a different way to toggle an existing feature as we currently treat `src` as case-insensitive when only routes are used but treat it as case-sensitive when `rewrites`, `redirects`, or `headers` are present on the deployment. 

This allows us to treat specific routes as case-sensitive in Next.js so we can gradually migrate to case-sensitive routes.

styfle (approved):

dav-is (approved):

PR #6953, opened 2021-11-05
[node] Fix launcher ESM on Windows dev

styfle:
### Overview

This PR solves an issue where Windows users with ESM dependencies would see the following error during `vc dev`:

```
[ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader.
```

### Details

The root cause, described in the [nodejs/node#31710](https://github.com/nodejs/node/issues/31710), is that `import()` doesn't accept file paths, it accepts URLs. It unexpectedly works on macOS/Linux because absolute file paths look like URLs, but Windows absolute paths do not look like URLs and thus fail. Thankfully, Node.js ships with a `pathToFileURL()` util for this exact purpose.

The reason why I didn't use `pathToFileURL()` for relative paths is because it doesn't seem to account for `__dirname`, meaning the path is resolved relative to the cwd and not the current file.

- Fixes https://github.com/vercel/customer-issues/issues/9
- Fixes https://github.com/vercel/customer-issues/issues/36
- Related to https://github.com/nodejs/node/issues/31710


packages/node/src/index.ts
```diff
@@ -460,6 +460,7 @@ export async function startDevServer(
     entrypoint.endsWith('.mjs') ||
     (pkg.type === 'module' && entrypoint.endsWith('.js'));
 
+  console.log({ isEsm, entrypoint });
```
styfle: arbitrary change to `@vercel/node` to make sure we run the tests...I'll remove after CI completes

AndyBitz (approved):

packages/node/src/index.ts
styfle: Ok CI is passing in 2edecba ✅ 

I reverted this in 433c8db 👍

PR #6960, opened 2021-11-08
[cli] Prevent recursion of `vercel build` and `vercel dev`

msimulcik:
This PR prevents recursive invocation of `vercel build` and `vercel dev` through the development command in the project settings and the dev script in `package.json`.

### Related Issues

https://github.com/vercel/runtimes/issues/237

#### Tests

#### Code Review


packages/cli/src/util/dev/server.ts
```diff
@@ -2161,7 +2161,10 @@ export default class DevServer {
       process.stdout.write(data.replace(proxyPort, devPort));
     });
 
-    p.on('exit', () => {
+    p.on('exit', (code: number) => {
+      if (code > 0) {
+        process.exit(code);
+      }
```
msimulcik, line 2167: This is needed for process to exit when the dev command fails. What was the intent behind leaving the process running when dev command exited? Because in that case it would only time out on non-existing dev port anyway.

packages/cli/src/commands/dev/index.ts
```diff
@@ -48,6 +48,22 @@ const help = () => {
...
+      )} script in ${cmd('package.json')}`
+    );
+    client.output.error(
+      `Learn More: http://err.sh/vercel/recursive-invocation`
+    );
+    return 1;
+  } else {
+    process.env.__VERCEL_DEV_RUNNING = '1';
+  }
+
```
msimulcik: This covers only the case when recursion happens in the dev command. The dev script recursion remains covered few lines below because to get rid of it we would need a change in the runtimes which is the only place which calls the dev script during `vercel dev`. Here https://github.com/vercel/runtimes/blob/a853c94419fb0ec934a8b0fb5dd84e0f016c281f/packages/static-build/src/index.ts#L520 we would need to exit the process when the child exits with an error. @AndyBitz mentioned that we are soon getting rid of that part of the builder so it's not necessary.

packages/cli/src/commands/dev/index.ts
```diff
@@ -48,6 +48,25 @@ const help = () => {
 };
 
 export default async function main(client: Client) {
+  client.output.log(
+    `__VERCEL_DEV_RUNNING = ${process.env.__VERCEL_DEV_RUNNING}`
+  );
```
styfle: Did you intend to print this for the user?

packages/cli/src/commands/dev/index.ts
```diff
@@ -48,6 +48,25 @@ const help = () => {
...
+  if (process.env.__VERCEL_DEV_RUNNING) {
+    client.output.error(
+      `${cmd(
+        `${getPkgName()} dev`
+      )} must not recursively invoke itself. Check the Development Command in the Project Settings or the ${cmd(
+        'dev'
+      )} script in ${cmd('package.json')}`
+    );
+    client.output.error(
+      `Learn More: http://err.sh/vercel/recursive-invocation`
```
styfle: This link is 404

packages/cli/src/commands/dev/index.ts
msimulcik: No, that's only for debugging purposes. Newly added test is failing on CI but passing locally.

packages/cli/src/util/dev/server.ts
styfle: I _think_ that listening for the exit event doesnt actually prevent it from exiting.

Maybe @lucleray or @TooTallNate can provide some insight here since it was added in https://github.com/vercel/vercel/pull/3673

packages/cli/test/integration.js
```diff
@@ -2238,6 +2299,92 @@ test('fail `now dev` dev script without now.json', async t => {
   );
 });
 
+// eslint-disable-next-line jest/no-focused-tests
+test.only('[vercel dev] fails when development commad calls vercel dev recursively', async t => {
```
styfle:
```suggestion
test('[vercel dev] fails when development command calls vercel dev recursively', async t => {
```

styfle (commented):
Lets make sure all tests are passing

packages/cli/src/util/dev/server.ts
msimulcik: The goal here is not to prevent it from exiting but to exit the parent process too when the child process exits

errors/recursive-invocation.md
```diff
@@ -0,0 +1,9 @@
+# Recursive invocation of the command
+
+#### Why This Error Occurred
+
+The Development/Build Command of the Project Settings or the `dev`/`build` script in `package.json` file of the used project invokes `vercel dev`/`vercel build` recursively.
+
+#### Possible Ways to Fix It
+
+Adjust the command/script to match what your framework uses to begin development mode, e.g. `next` for Next.js or `gatsby develop` for Gatsby.
```
msimulcik: @leo please check the content

leo (changes requested):
I tweaked the content of the error, but please move it to the right place:

The [Errors Reference](https://vercel.com/docs/errors). The place you added it to is legacy.

Maybe you can also add a note for others that ensures nobody will add more errors to the `vercel/vercel` repo directly.

msimulcik:
> I tweaked the content of the error, but please move it to the right place:
> 
> The [Errors Reference](https://vercel.com/docs/errors). The place you added it to is legacy.
> 
> Maybe you can also add a note for others that ensures nobody will add more errors to the `vercel/vercel` repo directly.

@leo I created a PR https://github.com/vercel/front/pull/11729 with the content you provided. I also fixed other places in this repo linking to the legacy errors.

leo (changes requested):
Sounds good! Approved the other PR. Make sure to use `vercel.link` (Bitly), though, like all the other errors.

msimulcik:
@styfle tests are now passing. @leo links are updated

packages/cli/src/commands/build.ts
```diff
@@ -78,6 +78,22 @@ const fields: {
...
+  if (process.env.__VERCEL_BUILD_RUNNING) {
+    client.output.error(
+      `${cmd(
+        `${getPkgName()} build`
+      )} must not recursively invoke itself. Check the Build Command in the Project Settings or the ${cmd(
+        'build'
+      )} script in ${cmd('package.json')}`
+    );
+    client.output.error(
+      `Learn More: https://vercel.link/recursive-invocation-of-commands`
```
styfle, line 90: Is there a docs PR to update this link? The section doesn't exist right now

msimulcik: Yes, https://github.com/vercel/front/pull/11729

leo (commented):
Link looks great now!

leo (approved):
Seems like I have to approve in order for GitHub to remove my request.

I haven't reviewed the code, but the error link looks good now. Just make sure it nicely resolves to the docs section in the other PR.

styfle (approved):
Thanks!

PR #6961, opened 2021-11-08
Add plugin packages for go/python/ruby

styfle:
Closes https://github.com/vercel/runtimes/issues/235


AndyBitz (approved):

AndyBitz:
I thought those will be added to https://github.com/vercel-community?

styfle:
I think that we should move them after properly converting. Otherwise we’ll have to move the depended builders too which is not safe for community to publish due to the org/scope on npm.

PR #6963, opened 2021-11-08
Fix middleware build result. Return middleware next when invalid result.

gdborton:
### Related Issues

This fixes an issue in canary builds where if the user's middleware doesn't return anything, the request errors.

Now we'll return a response that indicates that we've not handled the request.

#### Tests

#### Code Review

### DIFF
--- a/packages/middleware/src/entries.js
+++ b/packages/middleware/src/entries.js
@@ -10,6 +10,15 @@ _ENTRIES['middleware_pages/_middleware'] = {
         response: result,
       };
     }
-    return result;
+
+    return {
+      promise: Promise.resolve(),
+      waitUntil: Promise.resolve(),
+      response: new Response(null, {
+        headers: {
+          'x-middleware-next': 1,
+        },
+      }),
+    };
   },
 };
--- a/packages/middleware/src/entries.js
+++ b/packages/middleware/src/entries.js
@@ -3,22 +3,16 @@ _ENTRIES = typeof _ENTRIES === 'undefined' ? {} : _ENTRIES;
 _ENTRIES['middleware_pages/_middleware'] = {
   default: async function (ev) {
     const result = await middleware.default(ev.request, ev);
-    if (result instanceof Response) {
-      return {
-        promise: Promise.resolve(),
-        waitUntil: Promise.resolve(),
-        response: result,
-      };
-    }
-
     return {
       promise: Promise.resolve(),
       waitUntil: Promise.resolve(),
-      response: new Response(null, {
-        headers: {
-          'x-middleware-next': 1,
-        },
-      }),
+      response:
+        result ||
+        new Response(null, {
+          headers: {
+            'x-middleware-next': 1,
+          },
+        }),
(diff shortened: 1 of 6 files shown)


styfle (commented):
Can you add a test to ensure this is working?

packages/middleware/src/entries.js
```diff
@@ -10,6 +10,15 @@ _ENTRIES['middleware_pages/_middleware'] = {
         response: result,
       };
     }
-    return result;
+
+    return {
+      promise: Promise.resolve(),
+      waitUntil: Promise.resolve(),
+      response: new Response(null, {
```
gdborton: @AndyBitz is this assumption correct? Whenever middleware returns something that isn't a `Response` we essentially call next?

packages/middleware/src/entries.js
```diff
@@ -10,6 +10,15 @@ _ENTRIES['middleware_pages/_middleware'] = {
         response: result,
       };
     }
-    return result;
+
```
styfle: What about if the user does `return null`? Do we handle that differently than `return undefined`?

gdborton: The platform always expects a result like `{ waitUntil, response, promise }` I'm not aware of any special convention for returning null from a middleware.

packages/middleware/src/entries.js
AndyBitz: Yes, but we should check what Next.js does to make sure we don't create any differences in that behaviour

gdborton:
> Can you add a test to ensure this is working?

Yep!, but whether or not it "works" depends mostly on whether the result matches what the platform expects.

packages/middleware/test/build.test.ts
```diff
@@ -17,8 +25,30 @@ describe('build()', () => {
...
+    expect(String(handledResponse.response.body)).toEqual('Hi from the edge!');
+    expect(
+      (handledResponse.response as Response).headers.get('x-middleware-next')
+    ).toEqual(null);
+    expect(unhandledResponse.response.body).toEqual(null);
     expect(
-      await fsp.stat(join(fixture, '.output/server/pages/_middleware.js'))
-    ).toBeTruthy();
+      (unhandledResponse.response as Response).headers.get('x-middleware-next')
+    ).toEqual('1');
```
gdborton, line 55: @styfle Updated w/ a unit test that ensures returning nothing from userland middleware gets wrapped with a response w/ `x-middleware-next` set.

styfle (approved):
Great work, thanks! 🎉

packages/middleware/src/entries.js
gdborton: @AndyBitz I checked next.js and here's what they're doing - https://github.com/vercel/next.js/blob/canary/packages/next/server/web/adapter.ts#L35

No checking for the result to be an instance of Response, if it's truthy it's returned. I've updated this PR to do the same.

AndyBitz (approved):

PR #6967, opened 2021-11-09
Fix "undefined" DNS record types when running "dns rm"

unitof:
I noticed `vercel dns rm [recordID]` was showing `undefined` each time where the DNS Record Type should display:
Screen Shot 2021-11-08 at 10 13 31 PM

Did some digging and it _appears_ Vercel API v5 renames the `type` property of a DNS Record to `recordType`, so made a quick change to reflect that.

(`vercel dns ls [domain]` still uses API v4, so it still works as expected.)

Let me know if I should create an Issue!

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/dns/rm.ts
+++ b/packages/cli/src/commands/dns/rm.ts
@@ -103,7 +103,7 @@ function getDeleteTableRow(domainName: string, record: DNSRecord) {
   return [
     record.id,
     chalk.bold(
-      `${recordName} ${record.type} ${record.value} ${record.mxPriority || ''}`
+      `${recordName} ${record.recordType} ${record.value} ${record.mxPriority || ''}`
     ),
     chalk.gray(
       `${ms(Date.now() - new Date(Number(record.createdAt)).getTime())} ago`


unitof:
Though, maybe the issue is that the v5 API should revert from `record.recordType` back to `record.type`? Looks like v5 has been in use since May 2020 and I think this bug is newer? https://github.com/vercel/vercel/blame/986b4c0b1a2790b9d9623abed22e5aa133a31986/packages/cli/src/util/dns/get-dns-record-by-id.ts#L7-L8

packages/cli/src/commands/dns/rm.ts
```diff
@@ -103,7 +103,7 @@ function getDeleteTableRow(domainName: string, record: DNSRecord) {
   return [
     record.id,
     chalk.bold(
-      `${recordName} ${record.type} ${record.value} ${record.mxPriority || ''}`
+      `${recordName} ${record.recordType} ${record.value} ${record.mxPriority || ''}`
```
styfle, line 106: Thanks for the PR!

However, this will fail to build because the interface doesn't have `recordType` property:

https://github.com/vercel/vercel/blob/main/packages/cli/src/types.ts#L152-L166

Furthermore, there is no `recordType` property documented for Domain or DNS

https://vercel.com/docs/rest-api#endpoints/dns

https://vercel.com/docs/rest-api#endpoints/domains

I'll defer to @mglagola

unitof: Righto, makes sense! That's why I thought maybe something is up with the `v5` endpoint? It's definitely returning an attribute named `recordType` instead of the expected `type`:

```
curl "https://api.vercel.com/v5/domains/records/rec_45aa1cd08e614b7fce9acecd" \
     -H 'Authorization: Bearer [testingToken]'
```
Returns:
```
{"createdAt":1631766102686,"creator":"z9nQ0UbzQf52vfYODE78vvVk","domain":"jacobford.com","id":"rec_45aa1cd08e614b7fce9acecd","name":"waterboa","recordType":"CNAME","value":"starman.fathomdns.com.","ttl":60}%
```

mglagola: @unitof Thanks for finding this. This was an api bug - the endpoint responds with `type` now.

unitof: Confirmed! Thanks @mglagola.

styfle:
Yep, that was the fix. The API has been reverted 👍

PR #6969, opened 2021-11-09
[cli] Fix resolving of config paths

msimulcik:
This PR fixes functions for resolving local and global config paths. The issue was that `--local-config` resp. `--global-config` arguments were silently ignored when other than common args were also used  (e.g. `--confirm`). It caused that CLI integration tests were failing locally.

#### Tests

#### Code Review


styfle (approved):
Nice find, thanks! 🎉

AndyBitz (approved):

PR #6971, opened 2021-11-09
[build-utils][node] Add `updateFunctionsManifest` function

styfle:
- Fixes https://github.com/vercel/runtimes/issues/242
- Fixes https://github.com/vercel/runtimes/issues/234

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -2,7 +2,7 @@ import fs from 'fs-extra';
 import { join, dirname, relative } from 'path';
 import glob from './fs/glob';
 import { normalizePath } from './fs/normalize-path';
-import { FILES_SYMBOL, Lambda } from './lambda';
+import { FILES_SYMBOL, getLambdaOptionsFromFunction, Lambda } from './lambda';
 import type FileBlob from './file-blob';
 import type { BuilderFunctions, BuildOptions, Files } from './types';
 import minimatch from 'minimatch';
@@ -21,8 +21,8 @@ export function convertRuntimeToPlugin(
     const files = await glob('**', opts);
     delete files['vercel.json']; // Builders/Runtimes didn't have vercel.json
     const entrypoints = await glob(`api/**/*${ext}`, opts);
-    const functionsManifest: { [key: string]: any } = {};
-    const functions = await readVercelConfigFunctions(workPath);
+    const pages: { [key: string]: any } = {};
+    const { functions = {} } = await readVercelConfig(workPath);
     const traceDir = join(workPath, '.output', 'runtime-traced-files');
     await fs.ensureDir(traceDir);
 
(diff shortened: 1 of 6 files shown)


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -44,11 +44,11 @@ export function convertRuntimeToPlugin(
...
       });
 
-      functionsManifest[entrypoint] = {
+      pages[entrypoint] = {
         handler: output.handler,
         runtime: output.runtime,
-        memory: config.memory || output.memory,
-        maxDuration: config.maxDuration || output.maxDuration,
+        memory: output.memory,
+        maxDuration: output.maxDuration,
```
styfle, line 51: Some of this logic was moved into `updateFunctionsManifest()` so it can be shared with `vercel-plugin-node`

packages/plugin-node/src/index.ts
```diff
@@ -482,27 +483,13 @@ export async function buildEntrypoint({
...
-  } catch (_err) {
-    // ignore...
-  }
-  if (!functionsManifest.version) functionsManifest.version = 1;
-  if (!functionsManifest.pages) functionsManifest.pages = {};
-  functionsManifest.pages[entrypointWithoutExtIndex] = {
-    handler: `${getFileName(LAUNCHER_FILENAME).slice(0, -3)}.launcher`,
-    runtime: nodeVersion.runtime,
+  const pages = {
+    [entrypointWithoutExtIndex]: {
```
styfle, line 487: @AndyBitz @TooTallNate I kept this as without extension but I'm wondering if it should keep the extension. Thats how we document the `functions` key (it should match source files)

https://vercel.com/docs/cli#project-configuration/functions

msimulcik (approved):

packages/plugin-node/src/index.ts
leo: Great question! I think this also depends on the version that is returned from the Runtime, as the Build Step behaves differently for Runtimes than it does for Builders.

Can you confirm that the right thing to do is here, @AndyBitz?

AndyBitz: Files that go in the `pages` directory keep their extension.

Inside of the `functions-manifest.json` file, they are then referenced by their filename, **with the extension**.

Example of the `functions-manifest.json`:

```json
{
  "version": 1,
  "pages": {
    "api/date.js": {
      "memory": 128,
      "regions": ["gru1"]
    },
    "my-api.js": {
      "memory": 128,
      "maxDuration": 3,
      "regions": ["hkg1"]
    },
    "my-go-api.bin": {
      "memory": 1024,
      "runtime": "go1.x",
      "handler": "my-go-api.bin"
    },
    "my-isr-page.js": {
      "memory": 128
    },
    "my-isr-image.svg.js": {
      "memory": 3008
    }
  }
}
```

styfle: Thanks! Looks like this was fixed in #6988

PR #6984, opened 2021-11-12
[frameworks][examples] Add solidstart template

ryansolid:
This PR adds support for Solid framework with a default Solid client rendered template and a SolidStart SSR template and corresponding example projects.

I think I've put most things together, but a lot of copy and paste. Let me know if I'm missing anything.

#### Tests

#### Code Review


packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,94 @@ export const frameworks = [
...
+  {
+    name: 'SolidStart',
+    slug: 'solidstart',
+    demo: 'https://solidstart.examples.vercel.com',
+    logo: 'https://raw.githubusercontent.com/vercel/vercel/main/packages/frameworks/logos/solid.svg',
+    tagline:
+      'A Reactive JavaScript Library for Building User Interfaces.',
+    description: 'A Solid app, created with SolidStart.',
+    website: 'https://solidjs.com',
+    envPrefix: 'VITE_',
```
styfle, line 590: Should this be `SOLID_` since this framework isn't using vite?

ryansolid: It also uses Vite which brings up an interesting question if the detection rules are adequate since Vite and Solid will be present for both templates. Solid Start just buries the Vite commands and uses the programmatic APIs instead for setting defaults and better control but the environment and bundling is all still Vite.

Which actually I guess also brings a question about the dependencies. Since I have Vite as a peer dependency for the solid-start CLI.. which probably means I need to make it explicit for the build scripts to work? It works fine if someone builds locally because it's all installed but unclear what happens with the straight repo linking approach that would build with the CLI.

styfle: It should evaluate in order so the first match is used for the framework.

However, I'm wondering if we really need both here. 

Why are there two options if they both have the same dependencies? Shouldn't the framework using a single build command and then provide config to change internals. Like for Next.js we use `next.config.js` to let the user swap Babel for SWC for example.

ryansolid: They don't completely have the same dependencies just they both use Solid and Vite. One is just standard Vite template set up for client rendered SPAs, with all Vites standard things like index.html in the root of the project.

The other tries to take a more opinionated approach and has a completely different entry setup, basically setup for SSR(it hydrates from the Document root, something that you can do with a client side render), has a other features like built in file-based routing, head components etc, and has a CLI wrapper over Vite to do custom builds for different adapter platforms and honestly is a work in progress still.

It might be easiest to think of it as the difference between Svelte and SvelteKit or Vue and Nuxt. We just happen to use Vite as the bundler for both. 

I guess my question is around dependencies. The project will have their own dependencies and the build script should install them, but what does the dependency field in the framework.ts do? What needs to go there? I took it to mean to install the CLI. And for the basic Vite template I put `vite` there. But SolidStart has peer deps that it expects to come in as part of the repo since it is more than just the CLI but also the runtime/tooling parts. Do those need to go in the dependency list as well? Or is that for something else?

examples/solid/README.md
```diff
@@ -0,0 +1,30 @@
+# Solid Example
```
styfle: Missing the logo here.

See example/angular/README.md or examples/gatsby/README.md for the correct format

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,94 @@ export const frameworks = [
...
+        placeholder: '`npm run build` or `vite build`',
+      },
+      devCommand: {
+        placeholder: '`npm start` or `vite`',
+      },
+      outputDirectory: {
+        value: 'dist',
+      },
+    },
+    dependency: 'vite',
```
```suggestion
    dependency: undefined,
```

This is a legacy property before `detectors` existed so we can't rely on `vite` alone

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,94 @@ export const frameworks = [
...
+      },
+    },
+    dependency: 'solid-start',
+    devCommand: 'solid-start dev',
+    buildCommand: 'solid-start build',
+    getFsOutputDir: async () => 'dist',
+    getOutputDirName: async () => 'dist',
+  },
+  {
+    name: 'Solid',
```
styfle: I just realized `Solid` is nearly identical to the current `Vite` entry we already have. Let's remove this entry in favor of the existing `Vite`.

https://github.com/vercel/vercel/blob/c4fc060030490a90906ab764972f4406541ea0d6/packages/frameworks/src/frameworks.ts#L1913

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,94 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `vite build`',
+      },
+      devCommand: {
+        placeholder: '`npm start` or `vite`',
```
```suggestion
        value: 'vite',
```

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,94 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `solid-start build`',
+      },
+      devCommand: {
+        placeholder: '`npm start` or `solid-start dev`',
```
```suggestion
        value: 'solid-start dev',
```

packages/frameworks/src/frameworks.ts
styfle: > One is just standard Vite template set up for client rendered SPAs, with all Vites standard things like index.html in the root of the project.

In that case, we don't need an entry for it since we already have Vite.

styfle: > what does the dependency field in the framework.ts do

Its legacy before `detectors` existed.

ryansolid: > In that case, we don't need an entry for it since we already have Vite.

Ok sure I can remove the entry.  Vercel recognizing the project as Vite is fine from a detection standpoint. Should I keep/remove the example? I wasn't sure if people used this to find templates, where they are like "Hey I want to start a SolidJS project" and choose it.

packages/frameworks/src/frameworks.ts
ryansolid: I've removed it now.

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,49 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `solid-start build`',
+      },
+      devCommand: {
+        value: '`solid-start dev`',
```
```suggestion
        value: 'solid-start dev',
```

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,49 @@ export const frameworks = [
...
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"solid-start":\\s*".+?"[^}]*}',
+        },
+      ],
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install` or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `solid-start build`',
```
styfle, line 610:
```suggestion
        placeholder: '`npm run build` or `solid-start build`',
        value: 'solid-start build',
```

These have been updated to always require `value` so  its easier to understand

packages/frameworks/src/frameworks.ts
```diff
@@ -554,6 +554,49 @@ export const frameworks = [
...
+      devCommand: {
+        value: '`solid-start dev`',
+      },
+      outputDirectory: {
+        value: '.output',
+      },
+    },
+    dependency: 'solid-start',
+    devCommand: 'solid-start dev',
+    buildCommand: 'solid-start build',
```
```suggestion
```

These can all be removed, I double checked

styfle (changes requested):
We need a `yarn.lock` file committed to the example as well

styfle (commented):
Let's also remove the "Solid" example since it doesn't have a framework preset anymore.

We just need the SolidStart example 👍

ryansolid:
> Let's also remove the "Solid" example since it doesn't have a framework preset anymore.
> We just need the SolidStart example 👍

Sure. It does feel odd to me that the much more common(CSR) and the currently recommended way to use Solid wouldn't be part of this (SolidStart is still really new and only does SSR/SSG-based apps). But from entries perspective, I can see it making sense.

packages/frameworks/src/frameworks.ts
```diff
@@ -546,6 +546,47 @@ export const frameworks = [
...
+      },
+      buildCommand: {
+        placeholder: '`npm run build` or `solid-start build`',
+        value: 'solid-start build',
+      },
+      devCommand: {
+        value: 'solid-start dev',
+      },
+      outputDirectory: {
+        value: '.output',
```
styfle, line 617: I think the output directory here must be the static output, because I tried deploying and this doesn't work:

https://solidstart-1ovuv4mzd-now-examples.vercel.app/

cc @AndyBitz @TooTallNate

AndyBitz: Interesting, it seems like we need a way to support adding a framework to this list, that supports the File System API by default, unless this directory is the static directory.

styfle:
```suggestion
        value: '.output/static',
```

I confirmed, this should be static only

packages/frameworks/src/frameworks.ts
```diff
@@ -546,6 +546,47 @@ export const frameworks = [
...
+      },
+    },
+    getFsOutputDir: async () => '.output',
+    getOutputDirName: async () => '.output',
```
styfle, line 588:
```suggestion
    getOutputDirName: async () => '.output/static',
```

packages/frameworks/src/frameworks.ts
styfle: I tried deploying the example code here with Output Directory set to `.output/static` but that doesn't work because there is no index.html there either.

Maybe we need an indicator here to say that this framework uses the File System API?

ryansolid: Any updates on this? Anything I can do to help?

styfle: We'll make sure that `solidstart` is always enabled to the File System API beta.

Also, this PR will reduce the confusion for new frameworks: #7124

packages/frameworks/src/frameworks.ts
```diff
@@ -546,6 +546,47 @@ export const frameworks = [
...
+        value: 'solid-start build',
+      },
+      devCommand: {
+        value: 'solid-start dev',
+      },
+      outputDirectory: {
+        value: '.output',
+      },
+    },
+    getFsOutputDir: async () => '.output',
```
```suggestion
```

This can be removed now 👍

packages/frameworks/src/frameworks.ts
styfle: We're going to keep `outputDirectory` here pointing at the value of the static output, not the `.output`

packages/frameworks/src/frameworks.ts
ryansolid: Alright done and I updated the example to the latest with updated lock file. Is everything where it needs to be now?

Let me know when you would be about to merge. SolidStart is still in active development so it would be nice to make sure the deps are up to date when we do go to merge.

styfle (approved):

PR #7005, opened 2021-11-15
[cli] Change `vc build` to use stdout

styfle:
Since `output.log()` and `output.print()` go to stderr, the logs looked confusing

image

This PR changes the `vc build` output to use stdout instead of stderr.

- Fixes https://github.com/vercel/runtimes/issues/243

PR #7017, opened 2021-11-16
[cli] Fix spinner output outside of TTY

msimulcik:
This PR changes output of the spinner outside of TTY in a way that it only prints the provided message.

### Related Issues

Fixes https://github.com/vercel/runtimes/issues/267

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/deploy/process-deployment.ts
+++ b/packages/cli/src/util/deploy/process-deployment.ts
@@ -164,7 +164,7 @@ export default async function processDeployment({
         printInspectUrl(output, event.payload.inspectorUrl, deployStamp);
 
         if (quiet) {
-          process.stdout.write(`https://${event.payload.url}`);
+          process.stdout.write(`https://${event.payload.url}\n`);
         }
 
         output.spinner(
--- a/packages/cli/src/util/deploy/process-deployment.ts
+++ b/packages/cli/src/util/deploy/process-deployment.ts
@@ -164,7 +164,7 @@ export default async function processDeployment({
         printInspectUrl(output, event.payload.inspectorUrl, deployStamp);
 
         if (quiet) {
-          process.stdout.write(`https://${event.payload.url}\n`);
+          process.stdout.write(`https://${event.payload.url}`);
         }
 
         output.spinner(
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/output/create-output.ts
```diff
@@ -125,10 +125,14 @@ export class Output {
...
-      this._spinner.text = message;
+    if (this.isTTY) {
+      if (this._spinner) {
+        this._spinner.text = message;
+      } else {
+        this._spinner = wait(message, delay);
+      }
     } else {
-      this._spinner = wait(message, delay);
+      this.print(`${message}\n`);
```
styfle, line 135: I wonder if this should use `console.log()` instead.

- Related https://github.com/vercel/vercel/pull/7005

msimulcik: Make sense, going to change it

styfle (approved):

msimulcik:
@styfle please take on more look at the changes in tests

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -164,7 +164,7 @@ export default async function processDeployment({
         printInspectUrl(output, event.payload.inspectorUrl, deployStamp);
 
         if (quiet) {
-          process.stdout.write(`https://${event.payload.url}`);
+          process.stdout.write(`https://${event.payload.url}\n`);
```
styfle: I don't think this is a good idea. The deployment only prints the url to stdout. It shouldn't print anything else to stdout.

packages/cli/src/util/output/create-output.ts
```diff
@@ -125,10 +125,14 @@ export class Output {
...
-      this._spinner.text = message;
+    if (this.isTTY) {
+      if (this._spinner) {
+        this._spinner.text = message;
+      } else {
+        this._spinner = wait(message, delay);
+      }
     } else {
-      this._spinner = wait(message, delay);
+      console.log(message);
```
styfle: Since this function is used by `vc deploy`, I think we need to change it back to print() which is stderr 😢

msimulcik:
@styfle reverted back to `print()`

PR #7018, opened 2021-11-16
[cli] Fix update command to favor npm

styfle:
This PR fixes an issue where the wrong update command is show for npm 7 or npm 8.

This will favor npm if we can't be certain `vercel` was previously installed with yarn.


gdborton (approved):
LGTM

PR #7029, opened 2021-11-19
[frameworks] Fix Parcel default output directory to dist

samcx:
### Related Issues

The Parcel template via the Project Creating Flow returns a build error because the default output directory for the Parcel **Framework Preset** is searching for `public` instead of `dist`. The default behavior of Parcel's CLI I believe is `dist`.

#### Tests

#### Code Review

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1946,7 +1946,7 @@ export const frameworks = [
         value: 'parcel',
       },
       outputDirectory: {
-        value: 'dist'
+        value: 'dist',
       },
     },
     dependency: 'parcel',
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1978,6 +1978,7 @@ export const frameworks = [
       },
       outputDirectory: {
         placeholder: 'dist',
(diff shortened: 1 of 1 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -1978,6 +1978,7 @@ export const frameworks = [
       },
       outputDirectory: {
         placeholder: 'dist',
```
styfle, line 1980: Nice find!

You can remove `placeholder`, its either `placeholder` or `value` but not both

packages/frameworks/src/frameworks.ts
```diff
@@ -1978,6 +1978,7 @@ export const frameworks = [
...
+        value: 'dist'
```
```suggestion
        value: 'dist'
```

Nice find!

You can remove `placeholder`, its either `placeholder` or `value` but not both

styfle (approved):

AndyBitz (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -1946,7 +1946,7 @@ export const frameworks = [
         value: 'parcel',
       },
       outputDirectory: {
-        placeholder: 'dist',
+        value: 'dist'
```
```suggestion
        value: 'dist',
```

PR #7033, opened 2021-11-19
[tests] Make test more reliable for function timeout

styfle:
This will help flaky tests pass the first time

For example: https://github.com/vercel/vercel/runs/4268488876?check_suite_focus=true#step:11:1721

### DIFF
--- a/packages/cli/test/integration.js
+++ b/packages/cli/test/integration.js
@@ -2683,7 +2683,7 @@ test('deploy a Lambda with 3 seconds of maxDuration', async t => {
   const url = new URL(output.stdout);
 
   // Should time out
-  url.pathname = '/api/wait-for/4';
+  url.pathname = '/api/wait-for/5';
   const response1 = await fetch(url.href);
   t.is(
     response1.status,
@@ -2692,7 +2692,7 @@ test('deploy a Lambda with 3 seconds of maxDuration', async t => {
   );
 
   // Should not time out
-  url.pathname = '/api/wait-for/2';
+  url.pathname = '/api/wait-for/1';
   const response2 = await fetch(url.href);
   t.is(
     response2.status,


packages/cli/test/integration.js
```diff
@@ -2683,7 +2683,7 @@ test('deploy a Lambda with 3 seconds of maxDuration', async t => {
   const url = new URL(output.stdout);
 
   // Should time out
-  url.pathname = '/api/wait-for/4';
+  url.pathname = '/api/wait-for/5';
```
gdborton, line 2686: any reason this couldn't be 100000?

packages/cli/test/integration.js
```diff
@@ -2692,7 +2692,7 @@ test('deploy a Lambda with 3 seconds of maxDuration', async t => {
   );
 
   // Should not time out
-  url.pathname = '/api/wait-for/2';
+  url.pathname = '/api/wait-for/1';
```
gdborton, line 2695: same for this, but 0?

gdborton (approved):

packages/cli/test/integration.js
styfle: Because that would make CI slower, no need to wait much more than 3 seconds.

This test is called "deploy a Lambda with 3 seconds of maxDuration" so we just need to prove it will timeout

packages/cli/test/integration.js
styfle: Probably best to test some duration to make sure it doesn't timeout prematurely 

This test is called "deploy a Lambda with 3 seconds of maxDuration"

packages/cli/test/integration.js
gdborton: I guess I was assuming that this would timeout after 3 seconds no matter how long the route actually takes.

styfle: Oh good point. Thats probably why this assertion wasn't flaky, it was the other one

PR #7034, opened 2021-11-20
[cli] Fix metrics crash on some platforms

### LINKED ISSUE #6811: Vercel crash in some docker environement
Hi,

Running the Vercel CLI in some docker environments will crash with the following:

```
SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
    at new SystemError (node:internal/errors:233:5)
    at new NodeError (node:internal/errors:336:7)
    at Object.userInfo (node:os:347:11)
    at Object.metrics (/usr/local/lib/node_modules/vercel/dist/index.js:239808:23)
    at Object.86137 (/usr/local/lib/node_modules/vercel/dist/index.js:240111:26)
```

This is because the docker image might not have /etc/passwd mounted or referencing the current user, which crashes node.

The culprit seems to be this line: https://github.com/vercel/vercel/blob/bf3a1fc832f0d131772f1728950c20c9ce2b6275/packages/cli/src/util/metrics.ts#L20

A potential fix would be try/catching this and using a default value

styfle:
- Fixes #6811

### DIFF
--- a/packages/cli/src/util/metrics.ts
+++ b/packages/cli/src/util/metrics.ts
@@ -1,6 +1,5 @@
 import crypto from 'crypto';
 import ua from 'universal-analytics';
-import { platform, release, userInfo } from 'os';
 import { getPlatformEnv } from '@vercel/build-utils';
 
 import userAgent from './ua-browser';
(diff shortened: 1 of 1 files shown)


packages/cli/src/util/metrics.ts
```diff
@@ -16,10 +15,15 @@ export const shouldCollectMetrics =
...
+    typeof config.token === 'string'
+      ? config.token
+      : process.platform + process.arch;
+  const salt =
+    (process.env.USER || '') +
+    (process.env.LANG || '') +
+    (process.env.SHELL || '');
   const hash = crypto
-    .pbkdf2Sync(token, salt, 1000, 64, 'sha512')
+    .pbkdf2Sync(token, salt, 100, 64, 'sha512')
```
styfle, line 26: I reduced the iterations here for speed because this doesn't need to be cryptographically secure (its being truncated to 24 chars anyway)

AndyBitz (approved):

PR #7047, opened 2021-11-22
[cli] vc build - add symlinking of directories

msimulcik:
This PR switches from symlinking individual files to symlinking directories when possible. The performance gain is noticable only when there are many files which linking can be skipped by linking the directories. When tested on `front` package in `vercel/front` the performance remains the same:
Screenshot 2021-11-22 at 16 52 45
But when tested by running `vc build` in the root of `vercel/front` which contains roughly 22k files the difference is significant:
Screenshot 2021-11-22 at 21 34 24

### Related Issues

https://github.com/vercel/runtimes/issues/267

#### Tests

#### Code Review


packages/cli/src/commands/build.ts
```diff
@@ -316,55 +316,76 @@ export default async function main(client: Client) {
...
-      cwd,
-      absolute: true,
-    });
+
+    // copy files that are going to be mutated later in this script
+    const files = await glob(
+      join(
+        relative(cwd, distDir),
+        '**/{*.nft.json,middleware-manifest.json,required-server-files.json}'
+      ),
```
styfle, line 338: I think this is going to fail on Windows. The docs say to only use forward slashes

https://github.com/isaacs/node-glob#windows

Instead, the `cwd` param at the bottom should be changed

AndyBitz: If the `cwd` is changed, then the result changes as well, because `middleware-manifest.json` is not in `.next/` as the other files, it's inside `.next/server/`.

Wouldn't it make more sense to instead use something like `path.posix` for this?

AndyBitz:
We've just reverted a PR that switched from hard-links to symlinks, since symlinks cause other issues after the build process.

I'm not sure what will happen with directories, so we need to test it.
If it causes issues, we can also look into hard-linking directories if that's possible.

styfle:
Ideally, I think we should have only 1 symlink

    ln -s .next .output

Then we only need to resolve one symlink after running vercel build

leo:
Thank you for the effort you've put into this!

I'm closing it in favor of https://github.com/vercel/vercel/pull/7143 though, which removed the concept of having to copy anything entirely.

PR #7053, opened 2021-11-22
[python] Add warning for deprecated version

styfle:
- Print warning when python 3.6 is selected
- Print warning when invalid python version is selected


packages/python/src/index.ts
```diff
@@ -114,10 +114,24 @@ export const build = async ({
...
+          console.warn(
+            `Warning: Python version "${version}" detected in Pipfile.lock is deprecated and will reach End-Of-Life December 2021.`
+          );
+        } else if (version === '3.9') {
+          pipPath = 'pip3.9';
+          pythonPath = 'python3.9';
+          pythonRuntime = 'python3.9';
+        } else {
+          console.warn(
+            `Warning: Invalid Python version "${version}" detected in Pipfile.lock will be ignored. http://vercel.link/python-version`
```
TooTallNate, line 132: Should we print which version will be used instead?

TooTallNate (approved):

packages/python/src/index.ts
styfle: Perhaps. We're not doing that for Node.js right now. I think we just put all other details in the docs link.

https://github.com/vercel/vercel/blob/2721b3449dd3ccd213f9bdd558e61eeba3ba710c/packages/build-utils/src/fs/node-version.ts#L57-L63

styfle: Oh I guess for Node.js we error. Do you think we should error for Python too? I'm kind of worried about breaking existing deployments.

Issue #6975, opened 2021-11-10
`vc build` is mutating `.next` .nft.json files in Next.js > 12


jaredpalmer:
`vc build` is mutating the `*.nft.json` files in `.next` in place when moving files to `.output`. This seems like an observability issue and possibly a caching issue

### Running `next build`

![CleanShot 2021-11-10 at 08 21 15@2x](https://user-images.githubusercontent.com/4060187/141120623-dcc8567e-f960-4ecf-bb36-6465d193a1aa.png)

### Running `vc build`
![CleanShot 2021-11-10 at 08 17 13@2x](https://user-images.githubusercontent.com/4060187/141120555-02f803da-c3d0-475a-bc7a-199c0fbe8aae.png)

AndyBitz:
I guess this happens because we don't really copy the files, we just link them?
In that case we should probably copy all files that we modify in this process instead of linking them, which should be all `nft.json` files and `middleware-manifest.json`.

jaredpalmer:
but why are we writing the changes back to `.next/`? These changes should only be in the .output

styfle:
@jaredpalmer Probably because of the hardlinking?

jaredpalmer:
@styfle ahh duhhhh! my b. @AndyBitz you are correct. We should do exactly that

giancarlosisasi:
hi @jaredpalmer, could be this related to the fact that we need to redeploy (without the 'Include Build Cache' option) our apps to see the changes in our CSS files? 🤔 
The javascript files are updated in each deploy but our CSS files are only updated when we re-deploy the app 😢 The content of the CSS file built locally is different from the content that is built and deployed in vercel. We just noticed this today 🤔

jaredpalmer:
@GiancarlosIO yes that would make a lot of sense

PR #7026, opened 2021-11-18
[cli] Switch from hardlinks to symlinks in vc build

jaredpalmer:
- Safelist manipulated .nft.json and middleware-manifest.json files so that they are always copied. This ensures that .next directory is never altered by `vc build`
- Switch from `fs.createLink` to `fs.createSymlink`

When removing calling build plugins: `fs.createSymlink` was ~2x faster when copying 70k files. 

CleanShot 2021-11-18 at 10 55 22@2x

With build plugins, symlink is roughly 30%-40% faster.

### Related Issues

> Fixes #6975, #267

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [x] Issue from task tracker has a link to this PR

PR #7054, opened 2021-11-22
Revert "[cli] Switch from hardlinks to symlinks in vc build"

AndyBitz:
Reverts vercel/vercel#7026

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -678,16 +678,28 @@ export async function runPackageJsonScript(
 
 async function linkOrCopy(existingPath: string, newPath: string) {
   try {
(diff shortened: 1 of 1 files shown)


styfle (approved):

packages/cli/src/commands/build.ts
```diff
@@ -678,28 +678,16 @@ export async function runPackageJsonScript(
 
 async function linkOrCopy(existingPath: string, newPath: string) {
   try {
-    if (
-      newPath.endsWith('.nft.json') ||
```
jaredpalmer: @AndyBitz we actually want to keep this conditional copying logic here. It fixes #6975

AndyBitz: Ah, good point

PR #7059, opened 2021-11-23
[cli] Pass vercel.json to plugins during `vercel build`

styfle:
- Related to https://github.com/vercel/runtimes/issues/281

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -16,13 +16,20 @@ export function convertRuntimeToPlugin(
   buildRuntime: (options: BuildOptions) => Promise<{ output: Lambda }>,
   ext: string
 ) {
-  return async function build({ workPath }: { workPath: string }) {
+  // This `build()` signature should match `plugin.build()` signature in `vercel build`.
+  return async function build({
+    vercelConfig,
+    workPath,
+  }: {
+    vercelConfig: { functions?: BuilderFunctions; regions?: string[] };
+    workPath: string;
+  }) {
     const opts = { cwd: workPath };
     const files = await glob('**', opts);
     delete files['vercel.json']; // Builders/Runtimes didn't have vercel.json
     const entrypoints = await glob(`api/**/*${ext}`, opts);
     const pages: { [key: string]: any } = {};
-    const { functions = {} } = await readVercelConfig(workPath);
+    const { functions = {} } = vercelConfig;
     const traceDir = join(workPath, '.output', 'runtime-traced-files');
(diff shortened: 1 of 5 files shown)


packages/plugin-node/src/index.ts
```diff
@@ -407,16 +413,24 @@ export async function build({ workPath }: { workPath: string }) {
...
 }
 
 export async function buildEntrypoint({
+  vercelConfig,
   workPath,
   entrypoint,
   config,
   installedPaths,
 }: {
+  vercelConfig: any;
```
AndyBitz: Can we re-use the type from the `client` package?

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -124,22 +131,17 @@ async function readJson(filePath: string): Promise<{ [key: string]: any }> {
...
  * If `.output/functions-manifest.json` exists, append to the pages
  * property. Otherwise write a new file. This will also read `vercel.json`
  * and apply relevant `functions` property config.
  */
 export async function updateFunctionsManifest({
+  vercelConfig,
   workPath,
   pages,
 }: {
+  vercelConfig: { functions?: BuilderFunctions; regions?: string[] };
```
AndyBitz, line 144: Same here, can we use the actual type, or should it be a just a subset?

packages/plugin-node/src/index.ts
styfle: Done in f04411e

packages/build-utils/src/convert-runtime-to-plugin.ts
styfle: We don't need all the props, just these two are used in the conversion function here.

Also I don't think we can depend on `@vercel/client` from `@vercel/build-utils` since build-utils builds before client so that would make a cyclical dependency

styfle: I think in the future, we could try to merge `routing-utils` and `build-utils` into a single package and then move the VercelConfig type there instead of `client`

AndyBitz: Makes sense. However, we still need the `trailingSlash` and `cleanUrls` property to ensure we can output the correct routes for those settings.

AndyBitz: Unless you want to add that in a follow-up PR?

AndyBitz (approved):
Seems good, but we still need to consider `trailingSlash` and `cleanUrls`, but I guess it makes sense to add that in a follow up PR

PR #7061, opened 2021-11-24
Added Remix to the list of frameworks

leo:
### Related Pull Requests

This change adds [Remix](https://remix.run) to the list of frameworks and allows for https://github.com/remix-run/remix/pull/588 to be merged.

The example used in this PR right here is the same one that will be generated by `create-remix` after the PR above.

In this PR, the only relevant file is this one: `packages/frameworks/src/frameworks.ts`.

The demo can be found here: https://remix.examples.vercel.com

#### Tests

#### Code Review


packages/frameworks/src/frameworks.ts
```diff
@@ -17,6 +17,73 @@ const { readdir, readFile, unlink } = promises;
...
+    defaultHeaders: [
+      {
+        source: '^/build/(.*)$',
+        regex: '^/build/(.*)$',
+        headers: [
+          { key: 'cache-control', value: 'public, max-age=31536000, immutable' },
+        ],
+      },
+    ]
+  },
```
styfle: Lets move this entry down so remix doesn't show up as the first item in the template list.

We also have a `sort` property that should handle the sorting for front.

leo: > Lets move this down so it doesn't show up as the first item in the template list.

That's not how it works, though (I had a similar thought)? Blitz.js is first, yet it's not first there.

Let me check the property you mentioned.

AndyBitz (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -17,6 +17,73 @@ const { readdir, readFile, unlink } = promises;
...
+        src: '^/build/(.*)$',
+        headers: { 'cache-control': 'public, max-age=31536000, immutable' },
+        continue: true,
+      },
+      {
+        handle: 'filesystem',
+      },
+      {
+        src: '/(.*)',
+        dest: '/api',
```
styfle: This doesn't seem right. Is there always a `/api/index.js` file for every remix app?

packages/frameworks/src/frameworks.ts
leo: Updated the sorting to match our [Getting Started Docs](https://vercel.com/docs/get-started)!

styfle: Oh right the order in the array here is the order of detection.

So if you have both remix and next then it would favor remix since its above

packages/frameworks/src/frameworks.ts
leo: Yup! Remix is designed edge-first, so they have a rendering fallback for all routes.

On Vercel, they're gonna be Lambdas for now, until we expose Edge Functions as a user-facing primitive that is not Middleware (I discussed this with them on a call earlier).

packages/frameworks/src/frameworks.ts
leo: Ah! Moved it down a bit for correctness!

examples/remix/remix.config.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @type {import('@remix-run/dev/config').AppConfig}
+ */
+module.exports = {
+  appDirectory: "app",
+  browserBuildDirectory: "public/build",
+  publicPath: "/build/",
```
styfle, line 7: Seems like this is configurable so it should probably read the config file here to determine the default routes like we do for angular and some other frameworks

leo: That's a very good point! Agreed.

I'd like to open a separate PR for that, though, because there are currently a lot of people deploying Remix (it just went GA earlier this week), and it doesn't work well on Vercel.

It's probably gonna take some time to get that config reading logic to be reliable. I'll also have to generate the routes, not just the output directory value.

styfle:
@leo The demo deployment needs to be public

https://github.com/vercel/vercel/runs/4314064130?check_suite_focus=true#step:12:35

leo:
@styfle Fixed!

leo:
I'll try moving this to making use of `_middleware.js` before we merge it.

styfle (approved):

leo:
Nevermind. I won't use `_middleware.js` for now, because the respective Vercel CLI changes aren't on stable yet.

PR #7062, opened 2021-11-24
[build-utils] Add support for writing routes-manifest.json

styfle:
- Related https://github.com/vercel/runtimes/issues/281


packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -80,14 +119,54 @@ describe('convert-runtime-to-plugin', () => {
...
+        {
+          source: "^/api/(.+)(?:\\.(?:py))$",
+          destination: "/api/db/[id]?id=$1",
+          regex: "^/api/(.+)(?:\\.(?:py))$"
+        }
+        */
+      ],
+      dynamicRoutes: [
+        {
+          page: '/api/project/[aid]/[bid]/index',
```
styfle, line 138: @AndyBitz Does the page need to strip `/index` here?

AndyBitz: Good question. I think it doesn't matter, because `/index` and `/` should be equal

AndyBitz: ~Ah, I think it'd be best to remove it, as I just checked what Next.js does:~ 

```
  "dynamicRoutes": [
    {
      "page": "/[nice]",
      "regex": "^/([^/]+?)(?:/)?$",
      "routeKeys": {
        "nice": "nice"
      },
      "namedRegex": "^/(?<nice>[^/]+?)(?:/)?$"
    }
  ]
```
 for a `pages/[nice]/index.js`

AndyBitz: Not relevant because Next.js will not create an `index` file for it, it'll make `[nice]` the file itself, so yea, we can leave it as it is.

AndyBitz (approved):

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -80,14 +119,54 @@ describe('convert-runtime-to-plugin', () => {
       },
     });
 
+    const routesManifest = JSON.parse(output['routes-manifest.json']);
+    expect(routesManifest).toMatchObject({
+      version: 3,
+      pages404: true,
+      redirects: [],
+      rewrites: [
+        /* TODO: `handle: miss`
```
styfle, line 128: Currently, zero config routes have a `handle: miss` phase thats used to add a rewrite so that both the extension and the extensionless api file work (unless `cleanUrls: true` which will use a redirect instead).

However, the current `routes-manifest.json` doesn't support `handle: miss` routes, so we'll need to provide a new prop in the manifest if we want to keep the same behavior.

PR #7064, opened 2021-11-25
Updated Remix template and ensured correct headers

leo:
### Related Issues

Following up to https://github.com/remix-run/remix/pull/631, this updates the default Remix Template to the latest version generated by `create-remix` and ensures that the caching headers are working correctly.

#### Tests

#### Code Review


packages/frameworks/src/frameworks.ts
```diff
@@ -251,8 +251,8 @@ export const frameworks = [
     ],
     defaultHeaders: [
       {
-        source: '^/build/(.*)$',
-        regex: '^/build/(.*)$',
+        source: '/build/(.*)',
+        regex: '/build/(.*)',
```
AndyBitz, line 255:
```suggestion
        source: '/build/(.*)',
        regex: '^/build/(.*)$',
```

This should stay for the `regex`, but it's not required for `source`. Or was their a specific reason for why you removed it?

leo: Yes, I removed it because it caused a double regex in the final output, even though `routes-manifest.json` ended up being correct:
 
CleanShot 2021-11-25 at 11 25 01@2x

CleanShot 2021-11-25 at 11 48 15@2x

CleanShot 2021-11-25 at 11 27 41@2x

So either the File System API (not `vercel build`) is behaving incorrectly, or the file you commented on wasn't correct before my change.

AndyBitz (approved):

packages/frameworks/src/frameworks.ts
AndyBitz: I see, I'll investigate later on about this one.

PR #6988, opened 2021-11-12
[vercel-plugin-node] Use `.nft.json` files instead of directories.

AndyBitz:
### Related Issues

~Depends on https://github.com/vercel/fun/pull/73 for tests.~

We don't want to use directories for `pages`, as they cause issues with nesting.
This PR makes sure that we use `nft.json` files and put the related files into the `inputs` folder.

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #7084, opened 2021-11-30
Don't install Ruby or Python dependencies unnecessarily

leo:
### Related Issues

This closes https://github.com/vercel/runtimes/issues/252.

### Notes

- The changes for `vercel-plugin-node` were already handled in [this issue](https://github.com/vercel/vercel/pull/6988).
- `vercel-plugin-go` is not touched, because the File System API doesn't natively install dependencies for Go.
- `vercel-plugin-python` continues to run `installRequirement` because the `werkzeug` and `pipfile-requirements` dependencies are only needed for API Routes, and not for all Python dependency installations in general, so the File System API doesn't install those natively.

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -52,6 +52,9 @@ export function convertRuntimeToPlugin(
           includeFiles: config.includeFiles,
           excludeFiles: config.excludeFiles,
         },
+        meta: {
+          avoidTopLevelInstall: true,
+        },
       });
 
       pages[entrypoint] = {
--- a/packages/build-utils/src/types.ts
+++ b/packages/build-utils/src/types.ts
@@ -58,6 +58,7 @@ export interface Meta {
   filesRemoved?: string[];
   env?: Env;
   buildEnv?: Env;
+  avoidTopLevelInstall?: boolean;
 }
 
 export interface AnalyzeOptions {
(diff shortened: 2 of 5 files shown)


AndyBitz (approved):

styfle:
> vercel-plugin-go is not touched, because the File System API doesn't natively install dependencies for Go.

@leo Are you sure? I think the `go mod tidy` command installs dependencies.

leo:
@styfle I asked @AndyBitz and he said we don't do anything for it in the FS API. Is he wrong?

styfle:
I don't really understand the purpose of this PR but it sounds like you don't want to install dependencies. I'm not sure how its related to the FS API but just pointing out that `@vercel/go` does install dependencies in a similar way to `@vercel/python`.

AndyBitz:
See [this file](https://github.com/vercel/runtimes/blob/48923dc6563176f3493fb282cbe4887e3d4828ad/packages/file-system-api/src/install-dependencies.ts). It contains all the install code for the File System API, it was taken from `static-build`. There is no `go` related code in it.

leo:
The purpose of this PR is resolving the issue that was linked at the top, which is https://github.com/vercel/runtimes/issues/252. It talks about double installations of dependencies happening because the Install Command (that's what I mean with "File System API" here, I should've been more clear) already installs dependencies for Node.js, Ruby, and Python.

It doesn't do that for Go, so there's no point in trying to avoid a double install for it.

styfle:
Oh I understand now. In that case you're correct, static-build didn't attempt to install go mod dependencies 👍

PR #7087, opened 2021-11-30
[middleware] Define env vars when building `_middleware.js` with esbuild

### LINKED ISSUE #7086: `_middleware.js` does not have access to process.env when run without Next.js
When compiling the `_middleware.js` file with esbuild you need to pass the `process.env` variables to define or those will be undefined

https://github.com/vercel/vercel/blob/be1c78e72f1bbb23fbbcecf569049211cdf8bd52/packages/middleware/src/index.ts#L66

remorses:
### Related Issues

> Fixes #7086
> Related to #7086

#### Tests

#### Code Review


packages/middleware/test/build.test.ts
```diff
@@ -54,4 +54,29 @@ describe('build()', () => {
...
+    await build({
+      workPath: fixture,
+    });
+    // env var should be inlined in the output
+    delete process.env.ENV_VAR_SHOULD_BE_DEFINED;
+
+    const outputFile = join(fixture, '.output/server/pages/_middleware.js');
+    expect(await fsp.stat(outputFile)).toBeTruthy();
+
+    require(outputFile);
```
TooTallNate, line 79: Rather than polluting the global scope, can we use the `vm` module instead?

remorses: I copied what had already been done in the other test here
https://github.com/vercel/vercel/blob/c42efe6ecdebd914dd961f89f7d9acafb4dad500/packages/middleware/test/build.test.ts#L34

I think it would be more appropriate to migrate to `vm` in another PR if you want

remorses: Do you agree?

TooTallNate: Ya sorry, I see other tests doing that as well now. So disregard my comment here in this case.

gdborton (approved):
Looks good to me, thanks for adding a test!

styfle (approved):
Thanks!

TooTallNate:
Looks like the [unit tests are failing on Windows](https://github.com/vercel/vercel/runs/4414155272) due to having env var names with parenthesis `()`. We should be using bracket notation when defining the env vars object to remedy this.

packages/middleware/src/index.ts
```diff
@@ -53,6 +53,18 @@ async function getMiddlewareFile(workingDirectory: string) {
   return middlewareFiles[0];
 }
 
+function makeDefineObject() {
+  const defineObject: Record<string, string> = {};
+  for (const envVar in process.env) {
+    if (process.env[envVar] != null) {
+      defineObject[`process.env.${envVar}`] = JSON.stringify(
```
TooTallNate:
```suggestion
      defineObject[`process.env[${JSON.stringify(envVar)}]`] = JSON.stringify(
```

TooTallNate:
Hmmm, bracket notation didn't seem to work either. We may need to use `inject` instead.

remorses:
I added a plugin that polyfills the process global together with environment variables, that should work even for env variables like `ProgramFiles(x86)`

packages/middleware/src/esbuild-plugins.ts
```diff
@@ -0,0 +1,58 @@
+import path from 'path';
+import * as esbuild from 'esbuild';
+
+const processInjectFile = `
+const env = {};
+
+// envOverride is passed by esbuild plugin
+if (typeof envOverride === 'object') {
```
styfle: Lets perform this check prior to injecting the file. 

That way, this file can be simplified to `const env = envOverride;` (unless we need to clone it)

remorses: Fixed, another good reason to remove that check is that this way the process variable gets tree shaken by esbuild when not accessed

javivelasco (approved):

TooTallNate:
Re-running tests now.

TooTallNate:
Thank you @remorses!

PR #7088, opened 2021-11-30
Corrected dependency installation systems

leo:
### Related Issues

This is a follow-up change to https://github.com/vercel/vercel/pull/7084 and makes the last PR work by adding a missing `return` statement to the logic for preventing the `bundler` installation.

Furthermore, it prevents `.output` and `.vercel` from being provided to Runtimes and also ensures that all other files listed in `.vercelignore` or `.nowignore` are excluded — just like in the build container.

Lastly, it also ensures that the directory names in `.output/inputs` for `vercel-plugin-go`, `vercel-plugin-ruby`, and `vercel-plugin-python` are the same format, and create one input directory per CLI Plugin, not one for all, so that they cannot overwrite each other. `vercel-plugin-node` was also adjusted to share a similar format, but not yet use one directory for the entire CLI Plugin, as that will be handled in https://github.com/vercel/runtimes/issues/294.

The changes will make [this example](https://github.com/curated-tests/testing-file-system-api/tree/master/api-routes-ruby) work just fine.

#### Tests

#### Code Review


packages/plugin-node/src/index.ts
```diff
@@ -428,10 +428,8 @@ export async function buildEntrypoint({
   installedPaths?: Set<string>;
 }) {
   // Unique hash that will be used as directory name for `.output`.
-  const entrypointHash = crypto
-    .createHash('sha256')
-    .update(entrypoint)
-    .digest('hex');
+  // We're using the same format as for our other CLI Plugins.
+  const entrypointHash = getInputHash(`api/**/*.js`);
```
leo: This means we're using a single directory for all files that `vercel-plugin-node` handles.

I don't see an issue with that.

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -19,10 +20,19 @@ export function convertRuntimeToPlugin(
...
     const opts = { cwd: workPath };
     const files = await glob('**', opts);
+
     delete files['vercel.json']; // Builders/Runtimes didn't have vercel.json
-    const entrypoints = await glob(`api/**/*${ext}`, opts);
+
+    const entrypointPattern = `api/**/*${ext}`;
+    const entrypoints = await glob(entrypointPattern, opts);
     const pages: { [key: string]: any } = {};
-    const traceDir = join(workPath, '.output', 'runtime-traced-files');
```
leo, line 25: Thanks to this change, we're no longer using a single directory for all CLI Plugins.

Instead, just like with `vercel-plugin-node`, one per CLI Plugin is used. I don't see an issue with that.

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -50,6 +60,15 @@ export function convertRuntimeToPlugin(
...
       const lambdaFiles: Files = output[FILES_SYMBOL];
 
+      // `.output` was already created by the Build Command, so we have
+      // to ensure its contents don't get bundled into the Lambda. Similarily,
+      // we don't want to bundle anything from `.vercel` either.
+      for (const file in lambdaFiles) {
+        if (file.startsWith('.output') || file.startsWith('.vercel')) {
+          delete lambdaFiles[file];
+        }
+      }
```
AndyBitz: I guess in this case we can ignore them in the initial glob already, no? If we iterate over all files again it could slow down quite a bit, if we do it at the time we glob, we don't have to iterate again.

packages/ruby/install-ruby.ts
```diff
@@ -61,18 +61,27 @@ function getRubyPath(meta: Meta, gemfileContents: string) {
...
 // the absolute path to it
 export async function installBundler(meta: Meta, gemfileContents: string) {
+  const { gemHome, rubyPath, gemPath, vendorPath, runtime } = getRubyPath(
+    meta,
+    gemfileContents
+  );
+
   // If the new File System API is used (`avoidTopLevelInstall`), the Install Command
   // will have already installed the dependencies, so we don't need to do it again.
   if (meta.avoidTopLevelInstall) {
```
AndyBitz, line 71: Don't we need to check if it's the correct path? Did I oversee that in  the previous PR?

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Great point!

I thought the glob would be inside the Runtime, but it's actually outside, so I'll do that.

packages/ruby/install-ruby.ts
leo: No, because this function installs `bundler` (which is used for `bundle install`) and that CLI is already available ahead of time, because it's used for the `Gemfile` installation in the File System API.

What might be is that that `bundler` is located elsewhere and this function has to be adjusted accordingly, but I don't think so, and I'll verify that in the Deployment once it's published.

What might also be is that our `Gemfile` installation in the File System API doesn't support installing custom versions of `bundler`, which I think this function does, but I'll verify that too.

packages/build-utils/src/get-ignore-filter.ts
```diff
@@ -0,0 +1,81 @@
+import path from 'path';
```
leo, line 1: This entire file is 100% copy paste from [here](https://github.com/vercel/api/blob/31aba6fe60a5aa93990d36a7cc15d92c4766e659/scripts/build-container/src/utils/purpose-obtain.ts#L310). I'd like to make the Build Container load this file afterwards.

packages/plugin-go/tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "noUnusedParameters": true,
     "outDir": "dist",
     "strict": true,
-    "target": "esnext"
+    "target": "esnext",
+    "resolveJsonModule": true
```
AndyBitz, line 16: @TooTallNate I'm not sure if you were the one who told me about it before, but was there a reason for why we didn't add `resolveJsonModule` to the `vercel` repo before, did it have some side effects?

AndyBitz (approved):

packages/plugin-go/tsconfig.json
TooTallNate: No not that I'm aware of. I think we just tent to not add this option until/unless a `.json` file actually is being imported.

styfle: It only works when `module: "commonjs"` so I think we didn't want to use it if we ever moved our code to real modules ESM. Hopefully TS would error but not sure.

The other alternative is to use an actual `require()` so its clear it only works with require and not import.

PR #7094, opened 2021-12-01
[cli] Use "127.0.0.1" instead of "localhost" in `vc dev`

TooTallNate:
Node.js doesn't like when a hostname resolves to an IPv6 address (https://stackoverflow.com/a/15244890/376773) so use the IPv4 localhost IP address instead. Specifically this fixes vc dev on Node.js 17 which now prefers IPv6 by default.

Slack thread: https://vercel.slack.com/archives/C01A2M9R8RZ/p1638330248263400


AndyBitz (approved):

styfle (approved):

jensmeindertsma:
Thanks for fixing this Nate 😊

PR #7095, opened 2021-12-01
[build-utils] ignore symlinked directories in glob

igorklopov:
Currently `glob` npm package has a flaw that allows following symlinked directories even if `follow` is not true. An example is pattern `**/node_modules/**`. Glob returns `node_modules/link-dir/node_modules/b.txt` among the results, tho `link-dir` is a symlink directory. In general this issue causes build cache extraction fail due to duplicate files (captured from both real dir and symlinked dir).

#### Tests

#### Code Review


packages/build-utils/src/fs/glob.ts
```diff
@@ -39,9 +39,30 @@ export default async function glob(
 
   options.symlinks = {};
   options.statCache = {};
+  options.cache = {};
   options.stat = true;
   options.dot = true;
 
+  await vanillaGlob(pattern, options); // populates the caches
```
AndyBitz, line 46: Isn't it pretty inefficient if we call `glob` twice with the same pattern?

igorklopov: I don't think so because the second run reuses the caches almost entirely

igorklopov: Actually double-checking...

igorklopov: Unfortunately glob still makes lstat calls even with full cache, so the second run is not in-memory only. Not enough "score" to continue here. The solution is going to be too hacky here. Will implement in another place of the infra.

PR #7111, opened 2021-12-02
[cli] Ignore `required-server-files.json` if it does not exist

AndyBitz:
### Related Issues
This file might not exist if the build doesn't emit anything that would require it.


leo (approved):
Seems to be the same code, just wrapped into an `if` statement!

PR #7114, opened 2021-12-02
Filter CLI Plugin output

leo:
### Related Issues

This change completes https://github.com/vercel/vercel/pull/7088.

We were already filtering the list of files that get passed to Legacy Runtimes, but since that list of files is only used when deploying (locally it just uses whatever exists on the file system), we have to apply the same filtering to the output too.

`vercel-plugin-node` doesn't have this problem, because it only includes what it has traced, whereas `vercel-plugin-go`, `vercel-plugin-ruby`, and `vercel-plugin-python` just need to include everything because they're not tracing.

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -13,10 +13,18 @@ import { getIgnoreFilter } from '.';
 // Builders/Runtimes didn't have `vercel.json` or `now.json`.
 const ignoredPaths = ['.output', '.vercel', 'vercel.json', 'now.json'];
 
-const shouldIgnorePath = (file: string, ignoreFilter: any) => {
+const shouldIgnorePath = (
+  file: string,
+  ignoreFilter: any,
+  ignoreFile: boolean
+) => {
   const isNative = ignoredPaths.some(item => {
     return file.startsWith(item);
   });
+
+  if (!ignoreFile) {
+    return isNative;
+  }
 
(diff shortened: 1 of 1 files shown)


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -89,6 +94,16 @@ export function convertRuntimeToPlugin(
...
 
+      // When deploying, the `files` that are passed to the Legacy Runtimes already
+      // have certain files that are ignored stripped, but locally, that list of
+      // files isn't used by the Legacy Runtimes, so we need to apply the filters
+      // to the outputs that they are returning instead.
+      for (const file in lambdaFiles) {
+        if (shouldIgnorePath(file, ignoreFilter)) {
+          delete files[file];
+        }
+      }
```
AndyBitz: Could it be possible, that `.vercelignore` contains some build output, that the user just doesn't want to upload, but should be part of the lambda, since it's created at build time?

AndyBitz (approved):

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Yes, that's a good point. I think we only want to exclude the native ones then.

Done in https://github.com/vercel/vercel/pull/7114/commits/406ef8877e6e48622ddae4cd2c99643d5c8b0d18!

leo:
Unit tests worked on Ubuntu!

PR #7117, opened 2021-12-02
Fixed Go, Ruby, and Python CLI Plugin output generation

leo:
### Related Issues

This contributes to resolving https://github.com/vercel/runtimes/issues/294 by aligning all CLI Plugins within this repo with how the File System API expects them to work, and resolves a number of bugs as part of that.

At the moment, API Routes overwrite each other's launcher files, so if you have more than one API Route per CLI Plugin, that would cause a failure when deployed. That's fixed now.

`vercel-plugin-node` wasn't yet updated, because it requires changes that are a little different and more complicated, because it requires multiple launcher files (a bridge and a launcher) per entrypoint. A PR for it will follow shortly.

Before this change, CLI Plugins would treat the entrypoint files (like `api/test.js`) as the Lambda's entrypoint file too, which is wrong, and it now has been changed to being the launcher file, with all the other files placed inside `inputs`.

CleanShot 2021-12-02 at 17 49 35@2x

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -1,5 +1,5 @@
 import fs from 'fs-extra';
-import { join, dirname, parse } from 'path';
+import { join, dirname, parse, relative } from 'path';
 import glob from './fs/glob';
 import { normalizePath } from './fs/normalize-path';
 import { FILES_SYMBOL, Lambda } from './lambda';
@@ -88,7 +88,9 @@ export function convertRuntimeToPlugin(
     const pages: { [key: string]: any } = {};
     const pluginName = packageName.replace('vercel-plugin-', '');
 
-    const traceDirInputs = join(
+    const traceDir = join(
+      workPath,
+      `.output`,
       `inputs`,
       // Legacy Runtimes can only provide API Routes, so that's
       // why we can use this prefix for all of them. Here, we have to
@@ -96,8 +98,6 @@ export function convertRuntimeToPlugin(
       // need to be able to easily inspect the output.
       `api-routes-${pluginName}`
     );
-
-    const traceDir = join(workPath, `.output`, traceDirInputs);
 
     await fs.ensureDir(traceDir);
 
@@ -186,7 +186,10 @@ export function convertRuntimeToPlugin(
 
       await Promise.all(toRemove);
 
-      const tracedFiles: string[] = [];
+      const tracedFiles: {
+        absolutePath: string;
+        relativePath: string;
+      }[] = [];
 
       Object.entries(lambdaFiles).forEach(async ([relPath, file]) => {
         const newPath = join(traceDir, relPath);
@@ -196,7 +199,7 @@ export function convertRuntimeToPlugin(
           return;
         }
 
-        tracedFiles.push(relPath);
+        tracedFiles.push({ absolutePath: newPath, relativePath: relPath });
 
         if (file.fsPath) {
           await linkOrCopy(file.fsPath, newPath);
@@ -218,12 +221,10 @@ export function convertRuntimeToPlugin(
 
       const json = JSON.stringify({
         version: 1,
-        files: tracedFiles.map(file => {
-          return {
-            input: normalizePath(join('../../..', traceDirInputs, file)),
-            output: normalizePath(file),
-          };
-        }),
+        files: tracedFiles.map(file => ({
+          input: normalizePath(relative(nft, file.absolutePath)),
+          output: normalizePath(file.relativePath),
+        })),
       });
 
       await fs.ensureDir(dirname(nft));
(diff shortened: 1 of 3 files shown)


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -90,8 +113,28 @@ export function convertRuntimeToPlugin(
...
+      // so we don't want to pollute this space unnecessarily. That means we have to clean
+      // up files that were created by the build, which is done further below.
+      const sourceFilesAfterBuild = await getSourceFiles(
+        workPath,
+        ignoreFilter
+      );
+
+      // Further down, we will need the filename of the Lambda handler
+      // for placing it inside `server/pages/api`, but because Legacy Runtimes
+      // don't expose the that filename directly, we have to construct it
```
```suggestion
      // don't expose the filename directly, we have to construct it
```

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -141,12 +214,15 @@ export function convertRuntimeToPlugin(
...
+
       const json = JSON.stringify({
         version: 1,
-        files: tracedFiles.map(f => ({
-          input: normalizePath(relative(nft, f.absolutePath)),
-          output: normalizePath(f.relativePath),
-        })),
+        files: tracedFiles.map(file => {
+          return {
+            input: normalizePath(join('../../..', traceDirInputs, file)),
```
AndyBitz: I don't think this will work.

The reason why it was relative before, is because the file can be nested, e.g. `/api/users/info.py`, in that case it'd need to be `../../../../`, that's why `relative` was used.

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -68,12 +81,12 @@ describe('convert-runtime-to-plugin', () => {
...
           api: {
-            'index.py': expect.stringContaining('index'),
+            'index.py': expect.stringContaining('handler'),
             'index.py.nft.json': expect.stringContaining('{'),
             users: {
-              'get.py': expect.stringContaining('get'),
+              'get.py': expect.stringContaining('handler'),
               'get.py.nft.json': expect.stringContaining('{'),
-              'post.py': expect.stringContaining('post'),
+              'post.py': expect.stringContaining('handler'),
```
styfle, line 89: Is this change necessary? Looks like a bug to me since its not reading the contents

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Oh, right, I missed that.

Unfortunately `relative` adds one level too much, but I'll pick a different solution then!

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -29,6 +29,24 @@ const shouldIgnorePath = (
...
 
+const getSourceFiles = async (workPath: string, ignoreFilter: any) => {
+  const list = await glob('**', {
+    cwd: workPath,
+  });
+
+  // We're not passing this as an `ignore` filter to the `glob` function above,
+  // so that we can re-use exactly the same `getIgnoreFilter` method that the
+  // Build Step uses (literally the same code). Note that this exclusion only applies
+  // when deploying. Locally, another exclusion further below is needed.
```
styfle: What does "futher below" mean? Seems like there is nothing below this loop

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
leo: Yes, because the files placed in `server/pages/api` should be the actual Lambda entrypoints, and therefore the launcher files, not the request handlers that the user provided in their source code (those are separate files invoked by the launcher, depending on what Legacy Runtime is used).

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Great point. I moved that comment, but didn't update it.

Updated it in https://github.com/vercel/vercel/pull/7117/commits/7fbf4668cc74168d33be2fe74b76ad1def2ed91b!

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Seems like my other changes made `relative` work properly again! 🤔 

Fixed in https://github.com/vercel/vercel/pull/7117/commits/ca890767bc574df2a58bb48fd0dc6ecc5a096d5b and https://github.com/vercel/vercel/pull/7117/commits/7c4baeaafaf41609f47c97a09f5e9647fd8b89ee.

AndyBitz (approved):
Would probably be useful to have one e2e test for this that deploys the .output of a test of this. We'll definitely need to check if the handler logic will continue to work in prod

leo:
Agreed! I can add some shortly.

leo:
Everything except for unit tests on Windows seems to be working! Unit tests on Ubuntu and macOS are working.

PR #7118, opened 2021-12-02
Make middleware plugin output code in strict mode

gdborton:
### Related Issues

This updates the output middleware to run in strict mode. Added related dev/middleware build tests to ensure no regressions. I've also built and deployed a small test project locally.

Fixes https://github.com/vercel/runtimes/issues/289

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
+++ b/packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
@@ -2,5 +2,5 @@ export default function (req) {
   const isStrict = (function () {
     return !this;
   })();
-  return new Response('is strict mode? (next.js) ' + (isStrict ? 'yes' : 'no'));
+  return new Response('is strict mode? ' + (isStrict ? 'yes' : 'no'));
 }
--- a/packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
+++ b/packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
@@ -1,0 +1,6 @@
+export default function (req) {
+  const isStrict = (function () {
+    return !this;
+  })();
+  return new Response('is strict mode? (next.js) ' + (isStrict ? 'yes' : 'no'));
+}
--- a/packages/middleware/src/index.ts
+++ b/packages/middleware/src/index.ts
@@ -22,7 +22,8 @@ const SUPPORTED_EXTENSIONS = ['.js', '.ts'];
 
 // File name of the `entries.js` file that gets copied into the
 // project directory. Use a name that is unlikely to conflict.
-const ENTRIES_NAME = '___vc_entries.js';
+const TMP_ENTRIES_NAME = '.output/server/pages/___vc_entries.js';
+const TMP_MIDDLEWARE_BUNDLE = '.output/server/pages/_temp_middleware.js';
 
 async function getMiddlewareFile(workingDirectory: string) {
   // Only the root-level `_middleware.*` files are considered.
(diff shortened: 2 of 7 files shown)


packages/middleware/src/websandbox/sandbox/sandbox.ts
```diff
@@ -114,6 +114,7 @@ export async function run(params: {
     const content = readFileSync(params.path, 'utf-8');
     const esBuildResult = esbuild.transformSync(content, {
       format: 'cjs',
+      banner: '"use strict";',
```
gdborton, line 117: In `vc dev` all middleware ultimately runs through the sandbox, which does not bundle, but instead transforms imported code as it is imported. Therefore, adding `'use strict';` to the top of each imported file is the same as adding the build's final singular `'use strict';` added to the top of the file.

packages/middleware/src/index.ts
```diff
@@ -52,24 +53,35 @@ async function getMiddlewareFile(workingDirectory: string) {
...
   if (!middlewareFile) return;
 
   console.log('Compiling middleware file: %j', middlewareFile);
 
-  // Create `_ENTRIES` wrapper
-  await fsp.copyFile(join(__dirname, 'entries.js'), entriesPath);
-
   // Build
   try {
+    await esbuild.build({
```
gdborton, line 65: Going to add a code comment here explaining what's going on. 

Two builds happen here, because esbuild doesn't offer a way to add a banner to individual input files, and the entries wrapper relies on running in non-strict mode to access the `ENTRIES` global.

To work around this, we bundle the middleware directly and add `'use strict';` to make the entire bundle run in strict mode. We then bundle a second time, adding the global ENTRIES wrapper and preserving the `'use strict'` for the entire scope of the original bundle.

packages/middleware/src/index.ts
```diff
@@ -52,24 +53,44 @@ async function getMiddlewareFile(workingDirectory: string) {
...
-  // Build
+  /**
+   * Two builds happen here, because esbuild doesn't offer a way to add a banner
+   * to individual input files, and the entries wrapper relies on running in
+   * non-strict mode to access the ENTRIES global.
+   *
+   * To work around this, we bundle the middleware directly and add
+   * 'use strict'; to make the entire bundle run in strict mode. We then bundle
+   * a second time, adding the global ENTRIES wrapper and preserving the
+   * 'use strict' for the entire scope of the original bundle.
```
gdborton, line 72: Open to alternative approaches here.  I hit a snag when trying to bundle w/ swc and @leo recommended sticking w/ esbuild for now if it means we can quickly address the linked issue.

AndyBitz (approved):

packages/middleware/src/index.ts
leo: Exactly, we definitely should switch to SWC long-term, but to make it work, we can go this route already.

leo (approved):

packages/middleware/src/index.ts
gdborton: To be clear, looking for feedback (if any) on the approach of double bundling here as described in the comment.

leo: Maybe @TooTallNate or @styfle have a suggestion?

styfle: Can we avoid accessing the "ENTRIES global"?

Perhaps we can do the same method we use in dev which doesn't actually bundle.

gdborton: @styfle that's a good question... my understanding was that the runtime expects it to be available to hook up the middleware. Is there another option?

packages/middleware/test/build.test.ts
```diff
@@ -54,4 +61,14 @@ describe('build()', () => {
       (unhandledResponse.response as Response).headers.get('x-middleware-next')
     ).toEqual('1');
   });
+
+  it('should create a middlware that runs in strict mode', async () => {
```
styfle:
```suggestion
  it('should create a middleware that runs in strict mode', async () => {
```

packages/middleware/src/index.ts
styfle: We could swap `esbuild` for `@vercel/nft` to avoid bundling files. This is how we do it for node lambdas: https://github.com/vercel/vercel/blob/adb284519ad2b424651d420cd2f46b511173044f/packages/node/src/index.ts#L193

gdborton: I'm going to look into `@vercel/nft` approach to ensure it works in prod

gdborton: Slack conversations indicated that this wouldn't be possible as production runtime doesn't support import/export. I believe that bundle to a single file is mandatory.

styfle (approved):

packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
```diff
@@ -0,0 +1,6 @@
+export default function (req) {
+  const isStrict = (function () {
+    return !this;
+  })();
+  return new Response('is strict mode? (next.js) ' + (isStrict ? 'yes' : 'no'));
```
TooTallNate: Why `(next.js)`?

TooTallNate (approved):
One comment with the text fixture. But tested with a deployment and with dev and strict mode is indeed enabled. Nice one @gdborton 👍

packages/cli/test/fixtures/unit/edge-middleware-strict/_middleware.ts
gdborton: No reason, copy/pasta'd from the original report that strict mode wasn't in place. I'll remove rq to make sure no one thinks it's important in the future.

PR #7124, opened 2021-12-03
[frameworks][cli] Remove duplicate `getFsOutputDir()` definitions

styfle:
This PR will make it easier to add new frameworks since it doesn't need a duplicate function that is mostly unused.

Currently, the only frameworks using `getFsOutputDir()` are the following:

- `Next.js`
- `Blitz.js`
- `Other`

There were a couple others using `.output` but that is the default value anyway so those were removed.

- Related to #6984


AndyBitz (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -650,7 +617,6 @@ export const frameworks = [
       },
     },
     dependency: '@dojo/cli',
-    getFsOutputDir: async () => 'output/dist',
     getOutputDirName: async () => join('output', 'dist'),
```
TooTallNate, line 620: Unrelated, but why is this one using `path.join()`?

packages/frameworks/src/frameworks.ts
```diff
@@ -849,7 +813,6 @@ export const frameworks = [
       },
     },
     dependency: '@scullyio/init',
-    getFsOutputDir: async () => 'dist',
```
TooTallNate, line 852: This one has a different value than `getOutputDirName()`. Maybe this one needs to stay?

packages/frameworks/src/frameworks.ts
```diff
@@ -1666,7 +1617,6 @@ export const frameworks = [
       },
     },
     dependency: 'nuxt',
-    getFsOutputDir: async () => '.output',
```
TooTallNate, line 1669: Ditto.

packages/frameworks/src/frameworks.ts
styfle: Probably so it could work on Windows if we ever ran this function in `vc dev` for example

TooTallNate:
> There were a couple others using `.output` but that is the default value anyway so those were removed.

I'm not following. If `getFsOutputDir()` isn't defined then won't we fall back to `getOutputDir()` which wouldn't be `.output`?

packages/frameworks/src/frameworks.ts
styfle: @jaredpalmer Do you know why this one was different?

packages/frameworks/src/frameworks.ts
styfle: Since `.output` is the default value, it can be omitted like the PR desc mentions

styfle: If the framework emits `.output` then it will never call the `getFsOutputDir()` function because of this `if` statement:

https://github.com/vercel/vercel/blob/9c1b7dca31c8e51e8eda6a1aa6801dd72cb8e57d/packages/cli/src/commands/build.ts#L336

packages/frameworks/src/frameworks.ts
TooTallNate: Looks like it's because `dist/static` exists, so it already _sorta_ conforms to the FS API if we consider `dist` as the output dir.

packages/frameworks/src/frameworks.ts
TooTallNate: Shouldn't really make a difference I think. Windows file paths support backslashes.

packages/frameworks/src/frameworks.ts
TooTallNate: I see, thanks. Disregard this comment then.

PR #7127, opened 2021-12-03
Fixed Legacy Runtime output generation

leo:
As you can see in [this Deployment](https://vercel.com/curated-tests/api-routes-python/CMy1HBhNULg8U7f7ZXiMGmTcTg5k), the utility wrapped around Legacy Runtimes to convert them into CLI Plugins previously had an issue where dependencies of Serverless Functions would not be available.

This error was caused by the functionality within the utility that automatically cleans up the `workPath` after the Legacy Runtime has handled the build, so that the user's working directory won't remain polluted, and only `.output` is the result that remains at the end.

The changes in this PR ensure that the `workPath` remains clean, but the files that the Legacy Runtimes temporarily create in that directory are still made available in the final `.output` result.

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -101,6 +101,8 @@ export function convertRuntimeToPlugin(
 
     await fs.ensureDir(traceDir);
 
+    let newFilesRuntime: Set<string> = new Set();
+
     for (const entrypoint of Object.keys(entrypoints)) {
       const { output } = await buildRuntime({
         files: sourceFilesPreBuild,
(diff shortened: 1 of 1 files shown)


AndyBitz (approved):

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -229,8 +300,39 @@ export function convertRuntimeToPlugin(
...
+        ...newFilesEntrypoint,
+        ...newDirectoriesEntrypoint,
+      ]);
     }
 
+    // Instead of of waiting for all of the linking to be done for every
+    // entrypoint before processing the next one, we immediately handle all
+    // of them one after the other, while then waiting for the linking
+    // to finish right here, before we clean up newly created files below.
+    await Promise.all(linkersRuntime);
```
styfle, line 318: Do you need a semaphore to prevent unbounded File IO?

Seems like this will fail when lambdaFiles is in the thousands and the file descriptor limit is reached.

leo: I think you wrote the code for this, correct? I'm only adding a `Promise.all` to wait until all of them are finished.

Previously, the program would continue, even if all the linking wasn't done yet (which was fine before).

leo: Regardless, I do agree that it's potentially a good idea to add, but I think it's unrelated to the change in this PR and can be done in a separate one.

styfle (approved):

styfle:
How about updating the tests?

leo:
> How about updating the tests?

The current tests aren't affected by this change, but I'm happy to add new ones that assert that the output created by the CLI Plugins remains correct and we don't regress.

I'll do this once I've asserted that this change actually works in production, though, which requires one more PR for the `handlers` property of the resulting `functions-manifest.json` that I'll create shortly.

leo:
Merging because the relevant unit tests have passed and we're still not yet at a stage where the three CLI Plugins are working in the first place, so we need to move swiftly until they do — then we can tread more carefully.

PR #7129, opened 2021-12-03
Fixed Lambda handler for compiled languages

leo:
Interpreted languages deployed to AWS Lambda use the following format for the `handler` of a Lambda:

```
<file-name-without-extension>.<method-name>
```

Compiled languages, on the other hand, use this format:

```
<file-name>
```

The latter previously wasn't supported at all, but after this change, it will be.

The logic does not rely on actually knowing whether a compiled or interpreted language is used. Instead, it just checks if the `handler` exists as-is on the file system, and if it does, it's considered a compiled language. If that's not the case, the first format is assumed and used instead.

After this change, the error shown on the "Errors" tab of [this Deployment](https://vercel.com/curated-tests/api-routes-go/HofQuEBJSr4Zb8DnnjypsZCVHqCV/functions) should disappear.

It works with `vercel-plugin-go`, `vercel-plugin-ruby`, and `vercel-plugin-python`.

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -1,5 +1,5 @@
 import fs from 'fs-extra';
-import { join, dirname, parse, relative } from 'path';
+import { join, parse, relative, dirname, basename, extname } from 'path';
 import glob from './fs/glob';
 import { normalizePath } from './fs/normalize-path';
 import { FILES_SYMBOL, Lambda } from './lambda';
@@ -128,15 +128,6 @@ export function convertRuntimeToPlugin(
         ignoreFilter
       );
 
-      // Further down, we will need the filename of the Lambda handler
-      // for placing it inside `server/pages/api`, but because Legacy Runtimes
-      // don't expose the filename directly, we have to construct it
-      // from the handler name, and then find the matching file further below,
-      // because we don't yet know its extension here.
-      const handler = output.handler;
-      const handlerMethod = handler.split('.').reverse()[0];
-      const handlerFileName = handler.replace(`.${handlerMethod}`, '');
-
       // @ts-ignore This symbol is a private API
       const lambdaFiles: Files = output[FILES_SYMBOL];
 
@@ -150,9 +141,24 @@ export function convertRuntimeToPlugin(
         }
       }
 
-      const handlerFilePath = Object.keys(lambdaFiles).find(item => {
-        return parse(item).name === handlerFileName;
+      let handlerFilePath = Object.keys(lambdaFiles).find(item => {
+        return parse(item).name === output.handler;
       });
+
+      const { handler } = output;
+      const handlerMethod = handler.split('.').reverse()[0];
+      const handlerFileName = handler.replace(`.${handlerMethod}`, '');
+
+      // For compiled languages, the launcher file for the Lambda generated
+      // by the Legacy Runtime matches the `handler` defined for it, but for
+      // interpreted languages, the `handler` consists of the launcher file name
+      // without an extension, plus the name of the method inside of that file
+      // that should be invoked, so we have to construct the file path explicitly.
+      if (!handlerFilePath) {
+        handlerFilePath = Object.keys(lambdaFiles).find(item => {
+          return parse(item).name === handlerFileName;
+        });
+      }
 
       const handlerFileOrigin = lambdaFiles[handlerFilePath || ''].fsPath;
 
@@ -162,7 +168,9 @@ export function convertRuntimeToPlugin(
         );
       }
 
-      const entry = join(workPath, '.output', 'server', 'pages', entrypoint);
+      const entryRoot = join(workPath, '.output', 'server', 'pages');
+      const entryBase = basename(entrypoint).replace(ext, extname(handler));
+      const entry = join(entryRoot, dirname(entrypoint), entryBase);
 
       // We never want to link here, only copy, because the launcher
       // file often has the same name for every entrypoint, which means that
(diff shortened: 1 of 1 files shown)


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -150,19 +141,38 @@ export function convertRuntimeToPlugin(
...
       }
 
-      const handlerFilePath = Object.keys(lambdaFiles).find(item => {
-        return parse(item).name === handlerFileName;
+      let handlerFilePath = Object.keys(lambdaFiles).find(item => {
+        return parse(item).name === output.handler;
       });
 
+      const { handler } = output;
+      const handlerMethod = handler.split('.').reverse()[0];
```
```suggestion
      const handlerMethod = handler.split('.').pop();
```

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -150,19 +141,38 @@ export function convertRuntimeToPlugin(
...
+
+      // For compiled languages, the launcher file for the Lambda generated
+      // by the Legacy Runtime matches the `handler` defined for it, but for
+      // interpreted languages, the `handler` consists of the launcher file name
+      // without an extension, plus the name of the method inside of that file
+      // that should be invoked, so we have to construct the file path explicitly.
+      if (!handlerFilePath) {
+        handlerFilePath = Object.keys(lambdaFiles).find(item => {
+          return parse(item).name === handlerFileName;
+        });
```
AndyBitz: Instead of using `Object.keys` to iterate over all the files again, couldn't you do `lambdaFiles[completeHandlerFileName]`, where `completeHandlerFileName` is also including the path to the handler file?

AndyBitz: Because if we just compare the name, then it's probably pretty easy to also get false-positives, e.g. if there are two files that are called `test`, it could return for either, no?

AndyBitz: Hmm, but I guess `completeHandlerFileName` is what you're trying to figure out? But shouldn't there be a specific path that we can check since we know the source file name?

leo: `handlerFileName` doesn't have an extension, and we don't know which one it has before we find the actual file that's associated with it, so we couldn't construct `completeHandlerFileName`.

We could make the assumption that the `ext` that Runtime authors pass to the utility always matches the handler's extension, but I'm not sure if that's safe. I don't think so, because it would mean assuming that the outputs have the same extensions as the inputs.

leo: **TLDR:** We don't know the full file name (including the extension) of the handler on the file system, unless we make the assumption mentioned above.

AndyBitz: I assume it'd be something like this:

```
lambdaFiles[join(entryDir, entryBase, handlerFileName)];
```

Or without the `entryDir`.

leo: `handlerFileName` does not contain the extension. The `handler` that Runtimes return for their `Lambda` types does not contain the extension, so we don't know it before iterating through `lambdaFiles`.

AndyBitz: > handlerFileName doesn't have an extension, and we don't know which one it has before we find the actual file that's associated with it, so we couldn't construct completeHandlerFileName.

I see

AndyBitz: But just comparing the file name won't be enough.

We'll have to include the path, and the path should be known, right?

leo: If you think it's safe to rely on `ext` we can do it — I don't think it's safe, but I also don't know of any examples where it would break.

AndyBitz: Though, if we don't know the extension, it could still be a false-positive if there are two files with a different extension in the same directory.

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -305,12 +309,14 @@ export function convertRuntimeToPlugin(
         ...newDirectoriesEntrypoint,
       ]);
 
-      const apiRouteHandler = `${parse(entry).name}.${handlerMethod}`;
-
       // Add an entry that will later on be added to the `functions-manifest.json`
       // file that is placed inside of the `.output` directory.
-      pages[entrypoint] = {
-        handler: apiRouteHandler,
+      pages[join(entryDir, entryBase)] = {
```
styfle: `join()` here is going to use the windows backslash on windows and I think we want to use `/`

https://github.com/vercel/vercel/runs/4413823572?check_suite_focus=true

styfle: I think you can use `path.posix.join()` instead

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Agreed that we probably want to rely on `ext`, otherwise, we'll have the problem of two files with the same name (without an extension) causing a problem. I'll fix that.

AndyBitz: If the filename is anything other than `vc___` or something similar, then it's definitely not safe to rely on the filename either. Probably less than the extension.

leo: I also just noticed that this doesn't support `.exe` handlers because those are missing the extension on the left side, but they have their extension on the right side.

leo: Fixed!

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Fixed!

styfle (approved):
Looks good but it should add a new test, perhaps for Go, to prove it works correctly

leo:
Sounds great — I'll add some tests for all three shortly, once they work in production.

AndyBitz (approved):

PR #7133, opened 2021-12-04
Correctly position dependencies for CLI Plugins

leo:
Previously, CLI Plugins would try to mount the user-provided request handler (e.g. `api/test.rb`) at the same position inside the Lambda at which the launcher was located (e.g. `api/test.rb`), which would cause the launcher to be overwritten.

With this change, all the destination mounting points for NFT input files are becoming relative to the `.output/server/pages/api` directory instead of `.output/server/pages`, so they should no longer overwrite the launcher, and instead be loaded from the launcher, like normal.

This PR might have two problems:

- If imports in Ruby/Python/etc are relative to `cwd` and not the file from which the import is executed (which I doubt), this would fail.
- We might have to replace `api` with the exact sub folder of the API Route within `.output/server/pages/api` if the current change doesn't yet work for nested paths. Although that also means repeating all of the other dependencies (not just the user-provided request handler) in a different location for every single API Route.

The two above will be tested after this PR was merged, as there currently isn't a way to test `vercel-plugin-go`, `vercel-plugin-python`, and `vercel-plugin-ruby` without publishing a canary, because they don't bundle `@vercel/build-utils`, which was the package that was just updated.

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
+++ b/packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
@@ -114,35 +114,35 @@ describe('convert-runtime-to-plugin', () => {
       files: [
         {
           input: `../../../inputs/api-routes-python/api/db/[id].py`,
-          output: 'api/db/[id].py',
+          output: '.output/server/pages/api/api/db/[id].py',
         },
         {
           input: `../../../inputs/api-routes-python/api/index.py`,
-          output: 'api/index.py',
+          output: '.output/server/pages/api/api/index.py',
         },
         {
           input: `../../../inputs/api-routes-python/api/project/[aid]/[bid]/index.py`,
-          output: 'api/project/[aid]/[bid]/index.py',
+          output: '.output/server/pages/api/api/project/[aid]/[bid]/index.py',
         },
         {
           input: `../../../inputs/api-routes-python/api/users/get.py`,
-          output: 'api/users/get.py',
+          output: '.output/server/pages/api/api/users/get.py',
         },
         {
           input: `../../../inputs/api-routes-python/api/users/post.py`,
-          output: 'api/users/post.py',
+          output: '.output/server/pages/api/api/users/post.py',
         },
         {
           input: `../../../inputs/api-routes-python/file.txt`,
-          output: 'file.txt',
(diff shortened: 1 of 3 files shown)


AndyBitz (approved):

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -114,35 +114,35 @@ describe('convert-runtime-to-plugin', () => {
       files: [
         {
           input: `../../../inputs/api-routes-python/api/db/[id].py`,
-          output: 'api/db/[id].py',
+          output: '.output/server/pages/api/api/db/[id].py',
```
styfle, line 117: It looks like the `output` prop is now relative to the `workPath` dir.

Should we also make `input` prop relative to the `workPath` dir?

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
```diff
@@ -114,35 +114,35 @@ describe('convert-runtime-to-plugin', () => {
...
         },
         {
           input: `../../../inputs/api-routes-python/api/project/[aid]/[bid]/index.py`,
-          output: 'api/project/[aid]/[bid]/index.py',
+          output: '.output/server/pages/api/api/project/[aid]/[bid]/index.py',
         },
         {
           input: `../../../inputs/api-routes-python/api/users/get.py`,
-          output: 'api/users/get.py',
+          output: '.output/server/pages/api/api/users/get.py',
```
styfle, line 129: Something seems wrong here. Did you intend to have `api` nested in `api` like this?

leo: Yes! All the dependencies are placed next to the launcher.

packages/build-utils/test/unit.convert-runtime-to-plugin.test.ts
leo: Not sure I understand — `output` is not relative, it's the full output path.

leo: The `input` property has to be relative to the NFT file.

PR #7141, opened 2021-12-06
Support nested API Routes and fix handler for CLI Plugins

leo:
We have identified that the `handler` for Lambdas does not support a dot-preceded path for Ruby, Python, and probably other languages, so we're adjusting the File System API to change `.output` inside the Lambda to something else, which requires version `2` of `functions-manifest.json`.

Furthermore, we're also bumping the `.nft.json` files to version `2`, which allows `output` to be relative to the NFT file itself, so that, inside the Lambda, the behavior mentioned at the top can be applied by the File System API.

As a nice side effect, this will also support nested API Routes, because it'll place all the dependencies next to every API Route, meaning that the launcher will have access to all of them (bundling multiple API Routes or Pages into the same Lambda currently doesn't work for non-Next.js anyways, because of https://github.com/vercel/runtimes/issues/305).

#### Tests

#### Code Review


AndyBitz (approved):

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -290,12 +290,12 @@ export function convertRuntimeToPlugin(
       const nft = `${entry}.nft.json`;
 
       const json = JSON.stringify({
-        version: 1,
+        version: 2,
```
styfle, line 293: The tests need to be updated

leo: Done!

styfle (approved):
Thanks!

TooTallNate (approved):

styfle (approved):

PR #7144, opened 2021-12-06
[client] Resolve .nft.json files when `vc deploy --prebuilt`

styfle:
This ensures that using `vc deploy --prebuilt` will also upload any files that `.output/**/.nft.json` points to and also handle the Root Directory correctly since `vc build` emits `rootdir/.output`.

- Related to https://github.com/vercel/runtimes/issues/304.


packages/client/src/utils/hashes.ts
```diff
@@ -39,34 +38,65 @@ export const mapToObject = (
...
+): Promise<string[]> {
+  const semaphore = new Sema(100);
+  const existingFiles = Array.from(hashedFiles.values());
+  const resolvedFiles = new Set<string>();
+
+  await Promise.all(
+    existingFiles.map(async file => {
+      await semaphore.acquire();
+      const fileName = file.names[0];
+      if (fileName.startsWith('.output') && fileName.endsWith('.nft.json')) {
```
styfle: Seems to be a bug here since `fileName` is an absolute path so it always starts with `/`

packages/client/src/utils/hashes.ts
```diff
@@ -39,34 +38,65 @@ export const mapToObject = (
...
+      const fsPath = file.names[0];
+      if (fsPath.endsWith('.nft.json')) {
+        const json = file.data.toString('utf8');
+        const { version, files } = JSON.parse(json) as {
+          version: number;
+          files: string[];
+        };
+        if (version === 1 || version === 2) {
+          for (let f of files) {
+            resolvedFiles.add(join(dirname(fsPath), f));
```
AndyBitz: If you include `version: 1`, then you also have to consider that `files` will be `(string | { input: string; output: string; })[]`.

However, that should be easy to support by just doing: `typeof f === 'string' ? f : f.input`;

packages/client/src/utils/index.ts
```diff
@@ -119,37 +122,50 @@ export async function buildFileTree(
...
+      '.env.local',
+      '.env.*.local',
+      '.venv',
+      'npm-debug.log',
+      'config.gypi',
+      'node_modules',
+      '__pycache__',
+      'venv',
+      'CVS',
+      `${outputDir}`,
```
AndyBitz: I think we still want to ignore all `.output` directories.

For example, if I have a structure that looks like this:
```
.
└── packages
    ├── project-a
    │   └── .output
    └── project-b
        └── .output
```

Then I don't think I want to have `project-a/.output` when I deploy `project-b`, right?

packages/client/src/utils/hashes.ts
styfle: As far as I know, Next.js uses version 1 and its a string. But I'll add support for the input/output object too just in case other frameworks are using it 👍

packages/client/src/utils/index.ts
styfle: Good catch! Fixed in c92a88f

AndyBitz (approved):

packages/client/src/utils/hashes.ts
```diff
@@ -39,34 +38,66 @@ export const mapToObject = (
...
+    existingFiles.map(async file => {
+      await semaphore.acquire();
+      const fsPath = file.names[0];
+      if (fsPath.endsWith('.nft.json')) {
+        const json = file.data.toString('utf8');
+        const { version, files } = JSON.parse(json) as {
+          version: number;
+          files: string[] | { input: string; output: string }[];
+        };
+        if (version === 1 || version === 2) {
```
TooTallNate, line 89: What if an invalid version is supplied?

styfle: Then it logs an error in the `else` statement

TooTallNate (approved):

PR #7148, opened 2021-12-07
Adjust import statements inside Runtime launchers

leo:
This contributes to https://github.com/vercel/runtimes/issues/294 by ensuring that every launcher file used with an interpreted language like Ruby or Python imports the user-provided request handler from the source files, instead of expecting it to be copied right next to the launcher inside `.output`.

The path is already located in the respective NFT files, so all that was left is replacing the import statements inside the laucher files to load the user-provided request handler from the source files instead of `.output`.

#### Examples

**Python:**

CleanShot 2021-12-07 at 14 37 13@2x

**Ruby:**

CleanShot 2021-12-07 at 14 37 56@2x

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -228,6 +228,8 @@ export function convertRuntimeToPlugin(
 
             handlerContent = replaceAt(
               handlerContent,
+              // We'd like to add one character to the index, to consider
+              // that the import path is wrapped in quotes.
               match.index + 1,
               path,
               join(locationPrefix, path)
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -202,7 +202,7 @@ export function convertRuntimeToPlugin(
           // This is the full entrypoint path, like `./api/test.py`
           `./${entrypoint}`,
           // This is the entrypoint path without extension, like `api/test`
-          entrypoint.replace(ext, ''),
+          entrypoint.slice(0, -ext.length),
         ];
 
         // Generate a list of regular expressions that we can use for
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -236,7 +236,9 @@ export function convertRuntimeToPlugin(
               newPath
             );
 
-            debug(`Replaced "${oldPath}" inside "${entry}" with "${newPath}"`);
+            debug(
+              `Replaced "${oldPath}" inside "${entry}" with "${newPath}" to ensure correct import of user-provided request handler`
+            );
 
             replacedMatch = match;
           }
--- a/packages/build-utils/src/convert-runtime-to-plugin.ts
+++ b/packages/build-utils/src/convert-runtime-to-plugin.ts
@@ -191,6 +191,9 @@ export function convertRuntimeToPlugin(
       if (handlerHasImport) {
         const { fsPath } = handlerFile;
         const encoding = 'utf-8';
+
+        // This is the true location of the user-provided request handler in the
+        // source files, so that's what we will use as a import path in the launcher.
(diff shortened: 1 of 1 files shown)


packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -179,6 +180,81 @@ export function convertRuntimeToPlugin(
...
+      // it, so we have to adjust the import statement inside the launcher to point
+      // to the respective source file. Previously, Legacy Runtimes simply expected
+      // the user-provided request-handler to be copied right next to the launcher,
+      // but with the new File System API, files won't be moved around unnecessarily.
+      if (handlerHasImport) {
+        const { fsPath } = handlerFile;
+        const encoding = 'utf-8';
+
+        // This is the true directory of the user-provided request handler in the
+        // source files, so that's what we will use as a import path in the launcher.
```
```suggestion
        // source files, so that's what we will use as an import path in the launcher.
```

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -179,6 +180,81 @@ export function convertRuntimeToPlugin(
...
+        const locationPrefix = relative(entry, outputPath);
+
+        let handlerContent = await fs.readFile(fsPath, encoding);
+
+        const importPaths = [
+          // This is the full entrypoint path, like `./api/test.py`
+          `./${entrypoint}`,
+          // This is the entrypoint path without extension, like `api/test`
+          entrypoint.replace(ext, ''),
```
styfle:
```suggestion
          entrypoint.slice(0, -ext.length),
```

`replace()` will fail if there are multiple extensions since it replaces the first occurrence

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -179,6 +180,81 @@ export function convertRuntimeToPlugin(
...
+        ];
+
+        // Generate a list of regular expressions that we can use for
+        // finding matches, but only allow matches if the import path is
+        // wrapped inside single (') or double quotes (").
+        const patterns = importPaths.map(path => {
+          // eslint-disable-next-line no-useless-escape
+          return new RegExp(`('|")${path.replace(/\./g, '\\.')}('|")`, 'g');
```
styfle: Let's add some tests so we can refactor this implementation in the futuer

packages/build-utils/src/convert-runtime-to-plugin.ts
```diff
@@ -179,6 +180,81 @@ export function convertRuntimeToPlugin(
...
+        });
+
+        let replacedMatch = null;
+
+        for (const pattern of patterns) {
+          const matches = handlerContent.matchAll(pattern);
```
styfle: Can we simplify this loop to something like:

```js
for (const pattern of patterns) {
  handlerContent = handlerContent.replace(pattern, '$1')
}
```

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Right, but that's never the case?

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: As mentioned before, the output currently isn't yet stable and working, so I will defer adding tests until it's working, because the API is still changing quite dramatically, as we've seen yesterday yet again.

AndyBitz (approved):

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: I'm fine with going this route anyways, though.

packages/build-utils/src/convert-runtime-to-plugin.ts
leo: Done in https://github.com/vercel/vercel/pull/7148/commits/ab47a8db4b206ee58a43624c0522e4cf4257711b!

styfle (approved):

leo:
Only the CLI Plugins were modified, and their tests seem to have passed!

PR #7150, opened 2021-12-08
[client] Fix duplicate files when analyzing `nft.json`

styfle:
This PR fixes a regression from #7144 where a duplicate file was added if `nft.json` referenced an existing file.

I also refactored the prepared files logic to avoid cloning the array in every loop iteration.

Review [without whitespace](https://github.com/vercel/vercel/pull/7150/files?diff=split&w=1) will make it easier to understand.

- Related to https://github.com/vercel/runtimes/issues/304

### DIFF
--- a/packages/client/src/utils/hashes.ts
+++ b/packages/client/src/utils/hashes.ts
@@ -58,7 +58,9 @@ export async function hashes(
       const entry = map.get(h);
 
       if (entry) {
-        entry.names.push(name);
+        if (entry.names[0] !== name) {
+          entry.names.push(name);
+        }
       } else {
         map.set(h, { names: [name], data, mode });
       }
--- a/packages/client/src/utils/index.ts
+++ b/packages/client/src/utils/index.ts
@@ -266,39 +266,31 @@ export const prepareFiles = (
   files: Map<string, DeploymentFile>,
   clientOptions: VercelClientOptions
 ): PreparedFile[] => {
-  const preparedFiles = [...files.keys()].reduce(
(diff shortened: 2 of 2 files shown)


packages/client/src/utils/hashes.ts
```diff
@@ -58,7 +58,9 @@ export async function hashes(
       const entry = map.get(h);
 
       if (entry) {
-        entry.names.push(name);
+        if (entry.names[0] !== name) {
```
TooTallNate, line 61: Is the conflicting entry always going to be at index 0? Why not use a Set?

styfle: There's a lot of existing logic using `names[0]` so changing to a Set would complicate things.

https://github.com/vercel/vercel/blob/695f3a921220310a924c0af7dffed4eee306e060/packages/client/src/upload.ts#L88

I'm not even sure what the use case is for multiple names. Do you?

TooTallNate: I think it's so that you can have `foo.txt` and `bar.txt` that have the same file contents (same hash) but are two distinct files when re-populated in the build.

styfle: Oh right, the name is the file path so there could be multiple files that has to the same contents.

I think its fine to leave it as an array since the duplication is only caused by my previous PR #7144 which introduced a second call to hashes()

TooTallNate (approved):
Code seems good, but we should have/update tests for the `hashes()` function.

PR #7159, opened 2021-12-09
[client] Use a Set to prevent duplicates in `hashes()` function

TooTallNate:
Follow-up to #7150 to use a Set instead of assuming the conflicting entry is at index 0, which is not always the case.


styfle (approved):
It would be great to have a test 🙂

PR #7173, opened 2021-12-10
[WIP] Adjust Node.js CLI Plugin to use correct output

leo:
This pull request aligns the output of `vercel-plugin-node` with our other CLI Plugins.

The only change missing is one that @AndyBitz suggested, which changes the code path that writes the launcher file to write a JS file that imports the actual launcher file from the `.output/inputs` directory (only for the `.js` ext).

#### Tests

#### Code Review


TooTallNate:
So what happens to the differences that plugin-node has already introduced? For example parsing the static config with the static-config module?

leo:
I suggest adding that to the utility and only running it for the `.js` extension at the moment.

TooTallNate:
It's still not very clear to me what this PR is accomplishing. What does "aligns the output of `vercel-plugin-node` with our other CLI Plugins" mean?

TooTallNate:
> I suggest adding that to the utility and only running it for the `.js` extension at the moment.

In that case it should probably go into `@vercel/node` itself? Seems hacky to special-case a file extension in the general purpose utility module.

PR #7174, opened 2021-12-10
[build-utils] Add exclusion condition for vercel cli version

styfle:
Fixes https://github.com/vercel/runtimes/issues/295

### DIFF
--- a/packages/build-utils/src/detect-file-system-api.ts
+++ b/packages/build-utils/src/detect-file-system-api.ts
@@ -17,7 +17,8 @@ const enableFileSystemApiFrameworks = new Set(['solidstart']);
 
 /**
  * If the Deployment can be built with the new File System API,
- * we'll return the new Builder here, otherwise return `null`.
+ * return the new Builder. Otherwise an "Exclusion Condition"
+ * was hit so return `null` builder with a `reason` for exclusion.
  */
 export async function detectFileSystemAPI({
   files,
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/detect-file-system-api.ts
```diff
@@ -168,24 +169,44 @@ export async function detectFileSystemAPI({
         reason: `Detected Next.js with Output Directory \`${projectSettings.outputDirectory}\` override. Please change it back to the default.`,
       };
     }
-    const versionRange = deps['next'];
-    if (!versionRange) {
+    const nextVersion = deps['next'];
```
AndyBitz, line 172: I think calling it `nextVersion` might be a bit confusing, since it's a range when it comes from `package.json`.

Might make sense to call it `nextVersionRange`.

AndyBitz (approved):

packages/build-utils/src/detect-file-system-api.ts
styfle: I see what you mean.

I think of a range as the `engines` field, not the `dependencies` field because a range can be `>= 12.1.3 <= 14.0.0`.

I left the todo comment so that we can come back to resolving the actual installed version using the lockfile (we won't be able to use require()/resolve() unfortunately since this code executes before `npm install`)

PR #7175, opened 2021-12-10
[ci] Speed up GH publish skip

styfle:
This is following a similar pattern to Next.js workflows so that the publish step is skipped and reduces the time it takes GH Actions to run since we don't need to build to know that it should be skipped.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,21 +12,40 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/setup-go@v2
+    - name: Check Release
+      id: check-release
+      run: |
+        tag="$(git describe --tags --exact-match 2> /dev/null || :)"
+        if [[ -z "$tag" ]];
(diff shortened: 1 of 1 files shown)


AndyBitz (approved):

.github/workflows/publish.yml
```diff
@@ -12,21 +12,39 @@ jobs:
...
+    - name: Check Release
+      id: check-release
+      run: |
+        tag="$(git describe --tags --exact-match 2> /dev/null || :)"
+        if [[ -z "$tag" ]];
+          then
+            echo "::set-output name=IS_RELEASE::false"
+          else
+            echo "::set-output name=IS_RELEASE::true"
+        fi
```
styfle, line 26: This script is an excerpt from the existing publish script here:

https://github.com/vercel/vercel/blob/ecb44812263a03803157ad258c2f0f3cd0fc83ef/utils/publish.sh#L18-L23

ijjk (approved):

PR #7257, opened 2022-01-02
[cli] replace deprecated `--no-install` option of `npx` with `--no`

kidonng:
Quoting from [npm docs](https://docs.npmjs.com/cli/v8/commands/npx#:~:text=The-,%2D%2Dno%2Dinstall,-option%20is%20deprecated):

> The `--no-install` option is deprecated, and will be converted to `--no`.

### Related Issues

N/A

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -2156,7 +2156,7 @@ export default class DevServer {
       .catch(() => false);
 
     if (isNpxAvailable) {
-      command = `npx --no-install ${command}`;
+      command = `npx --no ${command}`;
     } else {
       const isYarnAvailable = await which('yarn')
         .then(() => true)


packages/cli/src/util/dev/server.ts
```diff
@@ -2156,7 +2156,7 @@ export default class DevServer {
       .catch(() => false);
 
     if (isNpxAvailable) {
-      command = `npx --no-install ${command}`;
+      command = `npx --no ${command}`;
```
styfle, line 2159: Seems like we need to support both flags so it works with old and new versions of npx?

kidonng: Yeah I forgot that 😅
`--no` is available since npm 7, which is bundled since Node 15. That means this change can only be made when Node 14 retires, which is 2023-04-30...

I will close this, but perhaps leave a note or so.

styfle: Do we know if `--no-install` is just deprecated or is it completely removed in newer versions of npm?

kidonng: It's just deprecated. I don't think npm is going to remove it any time soon.

PR #7307, opened 2022-01-13
[cli] Bump cert by id endpoint to v6

mglagola:
Bumps cert by id endpoint to v6 since v5 can have different response structure depending on team vs user. 

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/certs/get-cert-by-id.ts
+++ b/packages/cli/src/util/certs/get-cert-by-id.ts
@@ -4,7 +4,7 @@ import * as ERRORS from '../errors-ts';
 
 export default async function getCertById(client: Client, id: string) {
   try {
-    return await client.fetch<Cert>(`/v5/now/certs/${id}`);
+    return await client.fetch<Cert>(`/v6/now/certs/${id}`);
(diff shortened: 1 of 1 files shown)


styfle (approved):

packages/cli/src/util/certs/get-cert-by-id.ts
```diff
@@ -4,7 +4,7 @@ import * as ERRORS from '../errors-ts';
 
 export default async function getCertById(client: Client, id: string) {
   try {
-    return await client.fetch<Cert>(`/v5/now/certs/${id}`);
+    return await client.fetch<Cert>(`/v6/now/certs/${id}`);
```
styfle, line 7: Just to be clear `Cert` is the correct type still? It doesn't need to be modified?

mglagola: Yeah @styfle - v5 either returns `{ cert:  {...} }` or `{ ...cert }` 🥴 

v6 is standardized to `{...cert}`

TooTallNate (approved):

PR #7313, opened 2022-01-18
Update example to Eleventy 1.0.

leerob:
- Updates 11ty template to match their [base blog](https://github.com/11ty/eleventy-base-blog) example and use 1.0
- Updates all references to `.now.sh` to use their Vercel URL replacements
- Updates a few old references to ZEIT to Vercel
- Removes the custom build example, as it's no longer needed

The updated example has been deployed at https://eleventy.examples.vercel.com/.


leerob:
Just an example update, going to merge forward and can retroactively fix any comments applied. Going to keep trying to patch up some old examples here 👍

zachleat:
Hey @leerob—noticed you stripped out the license and author credit here—can we get that restored? Thanks

leerob:
Hey! Definitely not intentional, I believe I just copy-pasta'ed over the base blog - but I might have messed something up. Will fix!

Edit: Ahhh yep, I see what happened. I was trying to make the `package.json` as small as possible (so it's essentially only scripts) but good shout. Fixing!

PR #7326, opened 2022-01-19
[frameworks] Amend Hugo default `buildCommand` to exclude drafts enabled flag

RizBizKits:
### Related Issues

Reported in https://github.com/vercel/community/discussions/194

Having the `-D` flag results in all draft content to be published (docs reference [here](https://gohugo.io/getting-started/usage/)).

#### Tests

#### Code Review


AndyBitz (dismissed):

EndangeredMassa (dismissed):

TooTallNate (dismissed):

cb1kenobi:
I agree this makes sense. We don't want drafts published for production builds. However, isn't this technically a breaking change? Perhaps we should detect if the build command has _not_ been overridden and log a console message that draft pages are not being built?

EndangeredMassa:
We talked about this internally and decided to pair this change with a migration in our database for existing Hugo projects that have no custom build command. It will set their build command to the previous default.

cb1kenobi (dismissed):

cb1kenobi:
I don't think we ever did the agreed migration for existing users that rely on the drafts being published. We will need to discuss this and determine the migration steps. Also, we'll need a changeset for this PR.

cb1kenobi (approved):

PR #7350, opened 2022-01-22
[frameworks] Add Sanity Studio

kmelve:
This PR adds the default config and detector for a Sanity Studio project. The Sanity Studio is a Single Page Application, will always include a `sanity.json` config file in its root folder, and needs the `@sanity/cli` to build and run its development server.

#### Tests

#### Code Review


packages/frameworks/src/frameworks.ts
```diff
@@ -2006,6 +2006,51 @@ export const frameworks = [
...
+  {
+    name: 'Sanity',
+    slug: 'sanity',
+    demo: 'https://www.sanity.io/get-started',
+    logo: 'https://cdn.sanity.io/images/3do82whm/next/cd02ce807fbd324cbda82a77cd726c10ff6074d1-128x128.svg',
+    tagline:
+      'The structured content platform.',
+    description: 'A Sanity Studio',
+    website: 'https://www.sanity.io',
+    sort: 2,
```
TooTallNate: We can drop this one.

```suggestion
```

packages/frameworks/src/frameworks.ts
```diff
@@ -2006,6 +2006,51 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'Sanity',
+    slug: 'sanity',
+    demo: 'https://www.sanity.io/get-started',
+    logo: 'https://cdn.sanity.io/images/3do82whm/next/cd02ce807fbd324cbda82a77cd726c10ff6074d1-128x128.svg',
```
TooTallNate: Can you copy this file into the `frameworks/logos` directory and have it reference the GitHub URL like the other ones do?

packages/frameworks/src/frameworks.ts
```diff
@@ -2006,6 +2006,51 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'Sanity',
+    slug: 'sanity',
+    demo: 'https://www.sanity.io/get-started',
```
TooTallNate: This should point to a demo deployment, but don't worry about that for now - we will take care of it.

packages/frameworks/src/frameworks.ts
```diff
@@ -2006,6 +2006,51 @@ export const frameworks = [
...
+    detectors: {
+      every: [
+        {
+          path: 'sanity.json',
+        },
+      ],
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`sanity install`',
```
TooTallNate: `sanity` CLI won't be present in the build container at the time the install command is executed, so let's leave this like the others:

```suggestion
        placeholder: '`yarn install` or `npm install`',
```

packages/frameworks/src/frameworks.ts
kmelve: That's a tricky one since the studio is a fully authenticated application. Not sure what makes sense as a demo tbh.

packages/frameworks/src/frameworks.ts
kmelve: Done!

packages/frameworks/src/frameworks.ts
kmelve: For the record: The proposal is to add a Sanity Studio to `/examples` that can be deployed through Vercel's project creation and require the Sanity Integration. Then it will mint a project and deploy with the correct environment variables that will make the deployed studio work.

kmelve:
Any progress on this @TooTallNate?

examples/sanity-studio/.env.template
```diff
@@ -0,0 +1,3 @@
+# Run `vercel env pull` to generate a .env file from yur Vercel project
```
styfle:
```suggestion
# Run `vercel env pull` to generate a .env file from your Vercel project
```

styfle: Does `sanity start` work with both `.env` and `.env.local`?

If it works with `.env.local` then we should suggest `vercel env pull .env.local` instead

TooTallNate (approved):

AndyBitz (approved):

styfle (approved):

PR #7386, opened 2022-02-01
[build-utils] Add Builder TypeScript types

TooTallNate:
Adds formal type interfaces for Builders into `@vercel/build-utils`, and updates the serverless functions runtimes to utilize them. This provides type safety for the args/return values in i.e. the `build()` functions, where previously they were not using any explicit return type.

Also removes `watch` since this is no longer being used.


AndyBitz (approved):

packages/go/index.ts
```diff
@@ -456,20 +454,10 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
     environment: {},
   });
 
-  const watch = parsedAnalyzed.watch;
-  let watchSub: string[] = [];
-  // if `entrypoint` located in subdirectory
-  // we will need to concat it with return watch array
-  if (entrypointArr.length > 1) {
-    entrypointArr.pop();
```
styfle: This mutates `entrypointArr`. Is it okay to remove it? I can see `entrypointArr` is used below. 🤔

packages/build-utils/src/types.ts
```diff
@@ -368,3 +369,49 @@ export interface ProjectSettings {
...
   gitForkProtection?: boolean;
 }
+
+export interface FrameworkBuilder {
+  version: 2;
+  build: FrameworkBuild;
+  prepareCache?: PrepareCache;
+}
+
+export interface FunctionBuilder {
```
styfle: Should this be called `Runtime` or do we want to rebrand this in the docs?

styfle: Lets just call this `BuilderV3` to match the `BuildResultV3` return type?

packages/build-utils/src/types.ts
```diff
@@ -368,3 +369,49 @@ export interface ProjectSettings {
...
+export interface Images {
+  domains: string[];
+  sizes: number[];
+  minimumCacheTTL?: number;
+  formats?: ImageFormat[];
+}
+
+export interface BuildResultV2 {
+  // TODO: needs routing-utils for `Route` type
+  //routes: Route[];
```
styfle:
```
yarn turbo run build --graph=graph.html 
```

image

I think we're going to get into circular dependencies, so it might be best to create a new package with just types? For the time being you can just use `routes: any[]` and we can fix it in a future PR.

TooTallNate: Updated to use `any[]` for now.

packages/go/index.ts
TooTallNate: I don't see it being used for anything outside of the "watch" part which is being removed anyways, so pretty sure this is safe.

styfle (approved):
Code looks good but CI is failing

PR #7387, opened 2022-02-02
[cli] enhance changelog accuracy and format

EndangeredMassa:
This PR updates the `yarn changelog` format to make it easier to produce an accurate and organized changelog since the last release. Primary changes are:

- refactored existing logic to use more modules and functions to make it easier for me to follow
- parse the git log into commit objects for easier processing
- adds tests to the changelog generation that can be invoked with `yarn test` in the root

The flow is now:

- parse git log into commit objects
- filter out "Publish" commits
- filter out "Revert ..." commits
- filter out the commits those "Revert ..." commits reverted
- group commits into areas
- serialize the changelog

Note that "Revert ..." commits can:

- revert multiple other commits
- revert commits IN the current changelog entry
- revert commits BEFORE the current changelog entry

If a "Revert ..." commit reverts a commit before the current changelog, nothing is done with either the "Revert ..." commit or the reverted commits. The resulting changelog will include all of those commits, leaving the changelog author to decide what exactly to do. This decision was made because at least some of those commits were already in a published changelog. So, the latest changelog entry should probably say that the original commit is being undone.

---

Example of a commit object:

```
{
  areas: ['cli'],
  hash: '073f353fcf1944633bb43119c8ffcff46eea0480',
  message: 'This reverts commit 17fd88e044a807adf4ee6ed662cdb7c7556e912d.',
  revertsHashes: ['17fd88e044a807adf4ee6ed662cdb7c7556e912d'],
  subject: '[cli] Add "outDir" to `tsconfig.json` (#6566) [Nathan Rajlich]',
}
```

Example changelog files:

- [changelog-with-reverts.txt](https://github.com/vercel/vercel/files/7989404/changelog-with-reverts.txt)
- [changelog-with-publish-commits.txt](https://github.com/vercel/vercel/files/7989405/changelog-with-publish-commits.txt)
- [changelog-complete.txt](https://github.com/vercel/vercel/files/7989406/changelog-complete.txt)

---

### Follow-up Discussions

1. Should we automate this?
2. What tool should we use? These are currently used in some form:
    https://github.com/semantic-release/semantic-release
    https://github.com/conventional-changelog/standard-version
    https://github.com/vercel/release
    https://github.com/vercel/next.js/blob/canary/release.js
3. Should it be standard across all projects?
4. Should all commits be required to go through a PR? Can we enforce that? 
5. Should we have a convention that all PR titles use area tags (like `[cli][frameworks] ...`)? Can we enforce that? How does this interact with github labels?
6. Should we disallow manual merges?


utils/changelog/filter.js
```diff
@@ -0,0 +1,41 @@
...
+  );
+
+  let hashesToRemove = [];
+  revertCommits.forEach(revertCommit => {
+    hashesToRemove = hashesToRemove.concat(revertCommit.revertsHashes);
+  });
+
+  return filteredLogCommits.filter(
+    commit => !hashesToRemove.includes(commit.hash)
+  );
```
EndangeredMassa: I need to update this for the case where reverted commits were not found in the current changelog.

package.json
```diff
@@ -56,7 +56,8 @@
     "build": "turbo run build",
     "vercel-build": "mkdir -p public && echo '<a href=\"https://vercel.com/import\">Import</a>' > public/output.html",
     "pre-commit": "lint-staged",
-    "test-unit": "node utils/run.js test-unit",
+    "test": "jest",
+    "test-unit": "jest && node utils/run.js test-unit",
```
styfle: Is running `jest` going to pickup too many tests? Seems like we should only run unit tests instead of all here

package.json
```diff
@@ -147,5 +148,12 @@
         }
       }
     ]
+  },
+  "jest": {
+    "verbose": false,
+    "testEnvironment": "node",
+    "testMatch": [
+      "<rootDir>/test/**/*.test.js"
```
styfle: Oh I see, this is only running the root test dir. I would prefer this be added to `test-unit` instead of overriding the global

EndangeredMassa: @styfle I rearranged this to remove the global jest config. What do you think?

styfle (commented):
Can we sort the areas so the less important ones are at the bottom such as ci/tests/examples?

EndangeredMassa:
> Can we sort the areas so the less important ones are at the bottom such as ci/tests/examples?

@styfle definitely! I'd like to get this merged to have a solid base that we can iterate on. I'm writing an RFC to propose how we can manage CLI releases, which will affect how we want to write changelogs.

utils/changelog.js
```diff
@@ -48,8 +72,23 @@ async function main() {
     pkgs.add('vercel');
   }
 
-  const pub = Array.from(pkgs).join(',');
+  return pkgs;
+}
+
+// git log --pretty=format:"- %s [%an]" `git show-ref -s 'now@16.7.3'`...HEAD | grep -v '\- Publish '
```
styfle, line 78: This comment is no longer accurate, right?

styfle (approved):

TooTallNate (approved):

utils/changelog.js
EndangeredMassa: https://github.com/vercel/vercel/pull/7512

PR #7390, opened 2022-02-03
[cli] Remove initial iteration of `vc build` and Plugins

TooTallNate:
The `vercel build` command is being restructured, so we'll remove it for now so that this initial iteration isn't included as part of the stable CLI release.


packages/cli/src/commands/init/init.ts
```diff
@@ -28,6 +28,8 @@ type Example = {
 
 const EXAMPLE_API = 'https://now-example-files.zeit.sh';
 
+export const isBeta = true;
```
styfle: Why is `vercel init` beta?

TooTallNate: Whoops, didn't mean to commit that.

```suggestion
```

styfle (approved):

javivelasco (approved):

gdborton (approved):

styfle (commented):
Lets also remove the `.output` change to `.gitignore`

styfle (approved):

EndangeredMassa (approved):
There are a couple of references to `vercel build` in comments, but I don't think it's blocking.

- packages/cli/src/util/projects/project-settings.ts
- packages/build-utils/src/detect-file-system-api.ts

PR #7417, opened 2022-02-10
[build-utils] De-couple zip file creation from `Lambda` class

TooTallNate:
Makes the `Lambda` class itself just a reference of the files that are needed to produce a zip file, and adds a `createZip()` function that does the actual zip file creation. This will allow the zip file creation to be done in the Vercel build container instead of directly in the Builder itself.

* Adds `files` property to `Lambda` class
* Adds `createZip()` function to `Lambda` class
* Deprecates the `createLambda()` function (`new Lambda()` should be used instead)
* Deprecates the `zipBuffer` property (only populated when `createLambda()` is used)
* Removes the private `FILES_SYMBOL` symbol, since `files` is the same thing

### DIFF
--- a/packages/build-utils/src/lambda.ts
+++ b/packages/build-utils/src/lambda.ts
@@ -39,7 +39,7 @@ export class Lambda {
   public allowQuery?: string[];
   public regions?: string[];
   /**
-   * @deprecated
+   * @deprecated Use `await lambda.createZip()` instead.
    */
   public zipBuffer?: Buffer;
 
--- a/packages/build-utils/src/lambda.ts
+++ b/packages/build-utils/src/lambda.ts
@@ -13,17 +13,6 @@ interface Environment {
 }
 
 interface LambdaOptions {
-  zipBuffer: Buffer;
-  handler: string;
-  runtime: string;
-  memory?: number;
-  maxDuration?: number;
-  environment: Environment;
-  allowQuery?: string[];
-  regions?: string[];
-}
-
-interface CreateLambdaOptions {
   files: Files;
   handler: string;
   runtime: string;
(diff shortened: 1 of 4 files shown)


packages/build-utils/src/lambda.ts
```diff
@@ -39,31 +28,61 @@ interface GetLambdaOptionsFromFunctionOptions {
   config?: Pick<Config, 'functions'>;
 }
 
-export const FILES_SYMBOL = Symbol('files');
```
styfle, line 42: We can't remove this until we remove the usage in PR #7390

packages/build-utils/src/lambda.ts
```diff
@@ -39,31 +28,61 @@ interface GetLambdaOptionsFromFunctionOptions {
...
+  public files: Files;
   public handler: string;
   public runtime: string;
   public memory?: number;
   public maxDuration?: number;
   public environment: Environment;
   public allowQuery?: string[];
   public regions?: string[];
+  /**
+   * @deprecated
```
styfle: Should we make a suggestion of what to use instead here in the comment?

packages/build-utils/src/lambda.ts
styfle: If you want to merge this PR before 7390, then you can update the existing function to read `lambda.files` instead of `lambda[FILES_SYMBOL]`

TooTallNate: This PR is also updating the usage to use `lambda.files` instead, so not a blocker.

packages/build-utils/src/lambda.ts
TooTallNate: Sure thing. Added.

styfle (approved):

PR #7422, opened 2022-02-10
[build-utils] Retain support for passing in a `zipBuffer` directly to `new Lambda()`

TooTallNate:


styfle (approved):
Ideally, we should add some tests for this usage

PR #7423, opened 2022-02-11
[build-utils] Add `NodejsLambda` class

TooTallNate:
When an instance of this class is returned in a Builder's `output`, it is a signal to the build system that it needs to add additional files to the final Lambda before creating the zip file.


packages/build-utils/src/lambda.ts
```diff
@@ -12,7 +12,7 @@ interface Environment {
   [key: string]: string;
 }
 
-interface LambdaOptions {
+interface LambdaOptions<LauncherConfig = any> {
```
styfle: Do we need `= any` here? I think passing in a generic should be sufficient. Or are you thinking it might be a breaking change 🤔

TooTallNate: Switched to using a subclass instead of a generic.

styfle (approved):

styfle (approved):

PR #7436, opened 2022-02-14
[node] Use `NodejsLambda` class

TooTallNate:
Updates `@vercel/node` to utilize the `NodejsLambda` class so that it doesn't need to explicitly bundle in the launcher/bridge files into the build result. This reduces the complexity of the Builder and is a cleaner separation of concerns. The `helpers.test.ts` file has been moved to `@vercel/node-bridge` so it's being removed in this PR as well.


packages/node/test/helpers.test.js
```diff
@@ -1,842 +0,0 @@
-const fetch = require('node-fetch');
-const listen = require('test-listen');
-const qs = require('querystring');
-
-const { createServerWithHelpers } = require('../dist/helpers');
```
styfle, line 5: Lets make sure these tests are moved to the bridge package so we don't lose them

styfle (approved):

packages/node/src/index.ts
```diff
@@ -342,18 +333,14 @@ export * from './types';
...
   workPath,
   repoRootPath,
   config = {},
   meta = {},
-}: BuildOptions) {
-  const shouldAddHelpers = !(
-    config.helpers === false || process.env.NODEJS_HELPERS === '0'
-  );
-
+}) => {
```
EndangeredMassa, line 343: Should the argument still be typed `BuildOptions`?

EndangeredMassa (approved):

packages/node/test/helpers.test.js
TooTallNate: Already happened in #7451.

packages/node/src/index.ts
TooTallNate: It is still typed as such due to the `const build: BuildV3` type.

EndangeredMassa: Ahh, so it is. I saw that it was `any` locally because I had to `yarn build` to get the updated types across.

PR #7455, opened 2022-02-18
Use IPv4 local address for port checking

### LINKED ISSUE #7306: vc dev issues with node 17; unresponsive site and/or 502 on api routes (ipv4 / ipv6 issue)
**TL;DR;**

The issue is caused by changes in DNS resolution order in node 17.  The quickest short-term *workaround* is to remove `::1` from your `hosts` file (or otherwise change your system to resolve `localhost` to 127.0.0.1 before ::1), but watch this issue for a better solution.

**Full Explanation**

This issue affects `create-react-app` and presumably any other framework which defaults to `ipv4` and where `vc dev` waits for a port to become reachable.

Node 17 has a potentially breaking change where [`--dns-result-order` now defaults to `verbatim`](https://github.com/nodejs/node/commit/1b2749ecbe) (instead of `ip4first`).  Depending on system setup, this means `localhost` will now resolve to `::1` (ipv6) instead of `127.0.0.1` (ipv4).

Since create-react-app [currently defaults to ipv4](https://github.com/facebook/create-react-app/issues/11302), what happens in this case is:

1. `create-react-app` still opens your browser to `localhost` (127.0.0.1) on some random port
2. Browser requests to e.g. `localhost:3000` hang (with little hint why on the console)
3. After 5m (i.e. 300s), `vc dev` fails with, e.g.

```
Error! Detecting port 33409 timed out after 300000ms
```

This is because `vc dev`, using `is-port-reachable`, is testing on `localhost` which in this case resolves to `::1`, while `create-react-app` is only listening on `127.0.0.1`, so it will never be reachable (for `vc dev`).

The easiest workaround is to modify your system to resolve the `ipv4` address first, or, failing that, to remove `::1` from your `hosts` file.  Obviously this is not a good long-term solution.

You can force `create-react-app` to bind to an ipv6 with the `HOST` environment variable:

```
$ HOST=::1 vc dev
```

This allows the dev server to fully load and proxy requests to `create-react-app`, but API routes fail since the serverless funcs don't listen for them on `::1`, i.e.

```
Error! Failed to complete request to /api/serverless: Error: connect ECONNREFUSED ::1:35855
```

That's perhaps a different issue though :)

If relevant, I'm currently on Vercel CLI 23.1.3-canary.72.

A little surprised I'm the first one to report this, but maybe my system is unique :)

hguandl:
### Related Issues

> Fixes #7306

#### Tests

#### Code Review


styfle:
Looks like I can't push changes to your branch, so this should fix it: https://github.com/vercel/vercel/pull/7956

PR #7498, opened 2022-02-28
[build-utils] Make `files` in Lambda be optional

TooTallNate:
This ensures better backwards compatability in the case that `new
Lambda()` was passing in `zipBuffer` explicitly.

### DIFF
--- a/packages/build-utils/src/lambda.ts
+++ b/packages/build-utils/src/lambda.ts
@@ -28,6 +28,9 @@ export interface LambdaOptionsWithFiles extends LambdaOptionsBase {
   files: Files;
 }
 
+/**
+ * @deprecated Use `LambdaOptionsWithFiles` instead.
+ */
 export interface LambdaOptionsWithZipBuffer extends LambdaOptionsBase {
   /**
    * @deprecated Use `files` property instead.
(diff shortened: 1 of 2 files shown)


ijjk (approved):

packages/build-utils/src/lambda.ts
```diff
@@ -47,19 +55,18 @@ export class Lambda {
...
+      runtime,
+      maxDuration,
+      memory,
+      environment = {},
+      allowQuery,
+      regions,
+    } = opts;
+    if ('files' in opts) {
+      assert(typeof opts.files === 'object', '"files" must be an object');
     }
```
styfle, line 73: Should we also check zipBuffer too if it exists?

packages/build-utils/src/lambda.ts
```diff
@@ -12,19 +12,27 @@ interface Environment {
...
+}
+
+export interface LambdaOptionsWithFiles extends LambdaOptionsBase {
+  files: Files;
+}
+
+export interface LambdaOptionsWithZipBuffer extends LambdaOptionsBase {
   /**
    * @deprecated Use `files` property instead.
    */
```
styfle, line 37: Should we move the `LambdaOptionsWithZipBuffer` to be deprecated instead of the property?

packages/build-utils/src/lambda.ts
TooTallNate: Added.

packages/build-utils/src/lambda.ts
TooTallNate: Added.

styfle (approved):

EndangeredMassa (approved):

PR #7510, opened 2022-03-02
[build-utils] Add `EdgeFunction` output type

Schniz:
Add EdgeFunction type to produce as an output of a builder.
Still unused though.

#### Tests

#### Code Review


packages/routing-utils/src/types.ts
```diff
@@ -21,7 +21,39 @@ export type HasField = Array<
...
+       * A middleware index in the `middleware` key under the build result
+       */
+      middleware?: never;
+      /**
+       * A middleware key within the `output` key under the build result
+       */
+      middlewareKey?: never;
+    };
+
+export type Source = MiddlewareIndexed & {
```
dav-is: Does this mean it will look like:
```
{ src: '/.*', middleware: 0 }
```
or
```
{ src: '/.*', middlewareKey: '/_middleware' }
```

packages/routing-utils/src/types.ts
```diff
@@ -21,7 +21,39 @@ export type HasField = Array<
...
+    }
+  | {
+      /**
+       * A middleware index in the `middleware` key under the build result
+       */
+      middleware?: never;
+      /**
+       * A middleware key within the `output` key under the build result
+       */
+      middlewareKey: string;
```
dav-is: I think `middlewareAt` would be a better name

e.g.
```
{ src: '/blog/.*', middlewareAt: '/blog/_middleware' }
```

packages/routing-utils/src/types.ts
Schniz: yeah, but never both

Schniz: ahh, mm. it should be the same name like the output object. So probably yeah.

packages/routing-utils/src/types.ts
```diff
@@ -21,7 +21,39 @@ export type HasField = Array<
     }
 >;
 
-export type Source = {
+type MiddlewareIndexed =
+  | {
+      /**
+       * A middleware index in the `middleware` key under the build result
```
dav-is: Also, is there a way we can avoid duplicating this text 3 times?

dav-is (approved):

packages/routing-utils/src/types.ts
Schniz: there is, but it feels too hacky and "deep-typescripty" to make it work.
image
but fun to write nonetheless

[(playground link)](https://www.typescriptlang.org/play?#code/C4TwDgpgBAcg9gOwgeQGYB4AqA+KBeKAbygG0BrCEKASwSgpDlSkwF0B+ALiiQDcIATlAC+AKFGhIUZEjRZcBQqKikGNOgyYtW3YuXX1KOluVYioAMliIUGZAFtqwLABpDIbNlHDylLW3FJaBkUAXhZDBx8aQj5KAAfa1icQPBoACEAQwFoRWUoAHoAKiKoRwATcoAbCAB3bOhqAGcoAHM4OHKoIoL8iuq6hu4EAFd7ACNBfOLS-pr6nJoWyeBgQW7elTnBnIBBYG4m4AFaVu9UqQBZakr5hoBhRFRqVuiQ5DCbOSyc7ABucQFApQAACwCaAFoIAAPSAAY2AUIEAjgAlEcMQRyg0IAjNxrrcdhBHghnq88lsbgMFhB9twAET0lx9Kl3HLcHHMsTozHAbEAJnxrKJJLJ0SUlMJNLpUEZXPEGIQWOhAGYhVKHk8XuKWRr2VB+d4gA)

packages/build-utils/src/edge-function.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * WebAssembly bindings that the user code requires.
+   */
+  wasmBindings: EdgeFunctionWasmBinding[];
+};
+
+/**
+ * A helper to generate an EdgeFunction
+ */
+export function createEdgeFunction(
```
styfle: I think we're going to stop using `createLambda()` so I dont see a reason to add a helper for `createEdgeFunction()`.

@TooTallNate thoughts?

packages/build-utils/src/edge-function.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * The actual WASM binary to bind.
+   */
+  file: File;
+};
+
+export type EdgeFunction = {
+  type: 'EdgeFunction';
+  /**
+   * A display name for the edge function
```
styfle:
```suggestion
   * A display name for the edge function.
```

packages/build-utils/src/edge-function.ts
```diff
@@ -0,0 +1,61 @@
...
+  script: File;
+
+  /**
+   * The source map for the user code.
+   */
+  scriptSourceMap?: File;
+
+  /**
+   * The environment variables in use for the user code, to be
+   * bound into the edge function
```
styfle:
```suggestion
   * assigned to the edge function.
```

packages/routing-utils/src/types.ts
```diff
@@ -21,7 +21,39 @@ export type HasField = Array<
...
+       */
+      middleware?: never;
+      /**
+       * A middleware key within the `output` key under the build result
+       */
+      middlewareAt?: never;
+    };
```
styfle: It might be good to add tests for this to make sure it works as you expect with the 3 variations

packages/routing-utils/src/types.ts
styfle: Usually `At` suffix means a timestamp such as `createdAt: number` or `updatedAt: number`.

I think `middlewareKey` or `middlewarePath` is better.

dav-is: `middlewarePath` seems good too

packages/build-utils/src/edge-function.ts
TooTallNate: Agreed.

packages/routing-utils/src/types.ts
Schniz: 👍

AndyBitz (approved):

packages/build-utils/src/edge-function.ts
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * The list of files to be included in the edge function bundle.
+   */
+  files: Files;
+
+  /**
+   * The environment variables in use for the user code, to be
+   * assigned to the edge function.
+   */
+  envVarsInUse: string[];
```
TooTallNate: Can this one be detected by the post-processing logic? Or is it expected that the framework / Builder does static analysis to determine the values here?

Schniz: We can allow builders to force more env vars. Like in Next.js, where they already do that (and the compiled output is harder to statically analyze for these things)

packages/build-utils/src/edge-function.ts
```diff
@@ -0,0 +1,57 @@
+import type { Files } from './types';
+
+/**
+ * A WASM binding
+ */
+export type EdgeFunctionWasmBinding = {
```
TooTallNate: `type` vs. `interface` vs. `class`? `Lambda` for example is a class.

packages/build-utils/src/edge-function.ts
Schniz: Related:
* https://github.com/vercel/edge-functions/issues/39

i like the module syntax because it means we can treat it just like wasm

packages/build-utils/src/edge-function.ts
styfle: Would you prefer a class so we can do `new EdgeFunctionWasmBinding({ name, pathInFiles })`?

TooTallNate: I don't have a huge preference. For consistency `class` would make sense, but also only seems necessary if we plan on including instance functions.

TooTallNate (approved):
LGTM. Should we add this new output type to the `output` property of `BuildResultV2` in types.ts or should that happen in a separate PR?

packages/build-utils/src/edge-function.ts
Schniz: I am going to remove `EdgeFunctionWasmBinding` anyway because it will be inferred from imports.
Builders don't need to think about it--they can import a WASM file and that should just work, which makes the interface much simpler

But `new EdgeFunction` might make sense because all the API is based on classes, so I'll migrate it to that. :)

PR #7516, opened 2022-03-03
[cli] fail when local config is missing

EndangeredMassa:
Currently, when the path is not found in a `vc deploy -A SOME_PATH` command, the fallback config is used, which is unexpected behavior. This PR changes that to throw an error that says the path was not found.

---

Card: https://linear.app/vercel/issue/BUI-32/vc-deploy-a-should-fail-when-path-not-found


TooTallNate (approved):

styfle (approved):
Thanks!

PR #7530, opened 2022-03-08
[build-utils] streamline File types for better type narrowing

EndangeredMassa:
When narrowing a type based on `File`, the typescript compiler can't tell what the options are for `type`, which results in not narrowing.

Example:

```
if (file.type === 'FileFsRef') {
  file; // still a "File"
}
```

After this PR, we get narrowed `File` types:

```
if (file.type === 'FileFsRef') {
  file; // now a FileFsRef
}
```

### DIFF
--- a/packages/build-utils/src/types.ts
+++ b/packages/build-utils/src/types.ts
@@ -1,5 +1,6 @@
 import type FileRef from './file-ref';
 import type FileFsRef from './file-fs-ref';
+import type FileBlob from './file-blob';
 import type { Lambda } from './lambda';
 import type { Prerender } from './prerender';
 import type { EdgeFunction } from './edge-function';
@@ -8,16 +9,13 @@ export interface Env {
   [name: string]: string | undefined;
 }
 
-export interface File {
+export type File = FileRef | FileFsRef | FileBlob;
+export interface FileBase {
   type: string;
   mode: number;
   contentType?: string;
   toStream: () => NodeJS.ReadableStream;
   toStreamAsync?: () => Promise<NodeJS.ReadableStream>;
-  /**
-   * The absolute path to the file in the filesystem
-   */
-  fsPath?: string;
 }
 
 export interface Files {
@@ -399,7 +397,7 @@ export interface BuildResultV2 {
   routes?: any[];
   images?: Images;
   output: {
-    [key: string]: File | Lambda | Prerender | EdgeFunction;
+    [key: string]: FileBase | Lambda | Prerender | EdgeFunction;
   };
   wildcard?: Array<{
     domain: string;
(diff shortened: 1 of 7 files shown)


packages/cli/src/util/dev/builder.ts
```diff
@@ -288,7 +288,7 @@ export async function executeBuild(
   // Convert the JSON-ified output map back into their corresponding `File`
   // subclass type instances.
   for (const name of Object.keys(output)) {
-    const obj = output[name] as File;
+    const obj = output[name] as File | Lambda;
```
EndangeredMassa, line 291: This is a bit of a hammer to make the types work. This chunk of code should probably be refactored later to use the newer `File` types. (That's what @TooTallNate tells me, anyway!)

packages/build-utils/src/types.ts
```diff
@@ -8,16 +9,13 @@ export interface Env {
   [name: string]: string | undefined;
 }
 
-export interface File {
+export type File = FileRef | FileFsRef | FileBlob;
```
styfle, line 12: Should this include `Lambda`?

I think that was the purpose of the `toStream()` and `toStreamAsync()` props here.

EndangeredMassa: There are parts of the code that treat `Lambda` and `File*` the same, but they don't have properties in common. We think this should be separate. You can see some of the impact of this here: https://github.com/vercel/vercel/pull/7530#discussion_r821997250

packages/cli/src/util/dev/builder.ts
styfle: What about the `Files` type. I think thats like a list of `File` objects but it is sometimes uses as the "Output" (which would include `Lambda`).

packages/build-utils/src/types.ts
```diff
@@ -8,16 +9,13 @@ export interface Env {
...
   contentType?: string;
   toStream: () => NodeJS.ReadableStream;
   toStreamAsync?: () => Promise<NodeJS.ReadableStream>;
-  /**
-   * The absolute path to the file in the filesystem
-   */
-  fsPath?: string;
 }
 
 export interface Files {
```
styfle, line 21: Since `Files` is unchanged, this PR is effectively removing `Lambda` from the output however I think we often use `File` to also include `Lambda`

TooTallNate: "output" is defined on `BuildResultV2` and still includes `Lambda` as a possible type. `Files` has never allowed for Lambda (and is typically only used as the "input" to Builders, which makes sense that `Lambda` wouldn't be a possibility in that case).

styfle (approved):

TooTallNate (approved):

PR #7533, opened 2022-03-09
[client] preserve symlinks during deploy upload

EndangeredMassa:
Ensure that deploy uploads preserve symlinks.

---

Card: https://linear.app/vercel/issue/BUI-23/preserve-symlinks-in-vc-deploy


packages/client/src/utils/readdir-recursive.ts
```diff
@@ -0,0 +1,80 @@
+// based on https://github.com/jergason/recursive-readdir
```
styfle: Lets mention the MIT license here and also mention what was changed.

Can we send a PR upstream to fix it?

packages/client/tests/unit.utils.test.ts
```diff
@@ -38,6 +38,21 @@ describe('buildFileTree()', () => {
...
+    const cwd = fixture('symlinks');
+    const { fileList } = await buildFileTree(cwd, { isDirectory: true }, noop);
+
+    const expectedFileList = toAbsolutePaths(cwd, [
+      'folder-link',
+      'folder/text.txt',
+      'index.txt',
+      'index-link.txt',
+    ]);
+    expect(normalizeWindowsPaths(expectedFileList).sort()).toEqual(
```
styfle, line 52: Should we also assert the `mode` of the file to ensure the symlink is preserved?

packages/client/src/utils/readdir-recursive.ts
EndangeredMassa: Adding license and change info.

---

The project appears to be unmaintained. There have been no merged PRs in years. I could submit a PR, but it seems like it would go unnoticed.

It's also not necessarily a bug in the original code. They may want it to follow symlinks and not list directories as separate entries. The [readme](https://github.com/jergason/recursive-readdir#readme) calls out the behavior, but doesn't mention explicitly if that's a limitation or the intent.

packages/client/tests/unit.utils.test.ts
EndangeredMassa: added assertions

TooTallNate (approved):

PR #7534, opened 2022-03-09
[build-utils] Preserve symlinks for `FileRef` and `FileBlob` types in `download()`

EndangeredMassa:
Ensure we preserve symlinks for `FileRef` and `FileBlob` files in the build-utils `download()` function.

---

Card: https://linear.app/vercel/issue/BUI-23/preserve-symlinks-in-vc-deploy

### DIFF
--- a/packages/build-utils/test/unit.test.ts
+++ b/packages/build-utils/test/unit.test.ts
@@ -14,6 +14,7 @@ import {
   runNpmInstall,
   runPackageJsonScript,
   scanParentDirs,
+  FileBlob,
 } from '../src';
 
 async function expectBuilderError(promise: Promise<any>, pattern: string) {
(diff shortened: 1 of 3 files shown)


packages/build-utils/test/unit.test.ts
```diff
@@ -71,6 +72,53 @@ it('should re-create symlinks properly', async () => {
...
+  const outDir = path.join(__dirname, 'symlinks-out');
+  await fs.remove(outDir);
+
+  const files2 = await download(files, outDir);
+  assert.equal(Object.keys(files2).length, 4);
+
+  const [linkStat, linkDirStat, aStat] = await Promise.all([
+    fs.lstat(path.join(outDir, 'link.txt')),
+    fs.lstat(path.join(outDir, 'link-dir')),
+    fs.lstat(path.join(outDir, 'a.txt')),
```
styfle, line 123: Lets also check that `dir` is a directory

packages/build-utils/test/unit.test.ts
```diff
@@ -71,6 +72,53 @@ it('should re-create symlinks properly', async () => {
   assert(aStat.isFile());
 });
 
+it('should re-create FileRef symlinks properly', async () => {
```
TooTallNate:
```suggestion
it('should re-create FileBlob symlinks properly', async () => {
```

packages/build-utils/test/unit.test.ts
EndangeredMassa: Updated assertions to include that and symlink target values

TooTallNate (approved):

PR #7543, opened 2022-03-10
pnpm pnp symlinkn integration test

Ethan-Arrowood:
#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #7548, opened 2022-03-11
Revert "pnpm pnp symlinkn integration test"

Ethan-Arrowood:
Reverts vercel/vercel#7543

Noticed while working on my next integration test that not only does this test not actually probe the logs (`logsMustContain` is invalid, it should be `logMustContain`), but also it is not actually redeploying. I don't think `testDeployment` supports redeployment.


TooTallNate (approved):

styfle (approved):

packages/build-utils/test/integration.test.ts
```diff
@@ -221,51 +220,3 @@ it('Test `detectBuilders` with `index` files', async () => {
...
-  };
-
-  await fs.writeFile(
-    path.join(fixture, 'vercel.json'),
-    JSON.stringify(config2, null, 2)
-  );
-
-  const deployment2 = await testDeployment(
-    { builderUrl, buildUtilsUrl },
-    fixture
```
styfle, line 267: We would probably need to preserve the projectId from the previous deployment to actually test the build cache.

PR #4254, opened 2020-05-04
[cli][node] Re-apply @TooTallNate's `vc dev` updates

TooTallNate:
This reverts commit 099bc6dbf6b3e5cb0cc02acc44786cf87bdc4c60 (#4231).

PR #7549, opened 2022-03-11
[node] Consolidate dev and prd tsconfig.json

styfle:
In a previous PR (#4254), the behavior of `vercel dev` was added separately without matching how deployments work.

This PR consolidates the `tsconfig.json` options for `vercel` deployments as well as `vercel dev` so they work the same.

Fixes https://github.com/vercel/customer-issues/issues/349


EndangeredMassa (approved):
This looks good to me, but I think someone else with more context should review it as well.

AndyBitz (approved):

PR #7552, opened 2022-03-12
[go] Use `new Lambda()`

TooTallNate:
`createLambda()` is deprecated.


styfle (approved):

TooTallNate:
Something is going on with these go tests that we need to address first. Will retry this change once that's sorted out.

PR #7553, opened 2022-03-12
[build-utils] Remove `as` cast in `Lambda`

TooTallNate:
`Files` can now be properly narrowed by checking "type", so no need for the `as` cast here anymore.


AndyBitz (approved):

styfle (approved):
Wonderful, thanks!

PR #7555, opened 2022-03-12
[client] Use `.vercel/output` for --prebuilt

TooTallNate:
`.vercel/output` will now be used for the `vc build` output directory, rather than `.output`.

Also removes the `.nft.json` file tracing since it's no longer relevant in the latest `vc build` iteration.


packages/client/src/utils/index.ts
```diff
@@ -126,18 +126,22 @@ export async function getVercelIgnore(
...
+    const outputDir = posix.join(rootDirectory || '', '.vercel/output');
+    ignores = ['*'];
     const parts = outputDir.split('/');
     parts.forEach((_, i) => {
       const level = parts.slice(0, i + 1).join('/');
       ignores.push(`!${level}`);
     });
     ignores.push(`!${outputDir}/**`);
+
+    // TODO: remove - `package.json` should not be required
```
EndangeredMassa: Should this be resolved?

EndangeredMassa (approved):

packages/client/src/utils/index.ts
styfle: I dont understand the purpose here 🤔

packages/client/src/utils/index.ts
```diff
@@ -126,18 +126,22 @@ export async function getVercelIgnore(
   prebuilt?: boolean,
   rootDirectory?: string
 ): Promise<{ ig: Ignore; ignores: string[] }> {
-  let ignores: string[] = [];
-
-  const outputDir = posix.join(rootDirectory || '', '.output');
+  const ig = ignore();
```
styfle, line 129: Whats the difference between the `ig` object here vs the `ignores` array?

TooTallNate: I'm not sure entirely but I didn't change anything related to the response here.

packages/client/src/utils/index.ts
```diff
@@ -126,18 +126,19 @@ export async function getVercelIgnore(
...
+  let ignores: string[];
 
   if (prebuilt) {
-    ignores.push('*');
+    const outputDir = posix.join(rootDirectory || '', '.vercel/output');
```
styfle, line 133:
```suggestion
    const outputDir = posix.join(rootDirectory || '', '.vercel', 'output');
```

styfle (approved):

packages/client/src/utils/index.ts
TooTallNate: Same thing really 🤷

styfle: Don't we need it to support Windows?

styfle: Now that I think about it, the `outputDir.split('/')` on line 135 wont work on windows

TooTallNate: It's using `posix.join()`

PR #7558, opened 2022-03-13
[build-utils] Export * from `Lambda` and `NodejsLambda`

TooTallNate:
To have access to the "options" types.


EndangeredMassa (approved):
This also appears to export `createZip` from `labda.ts`. Is that intentional too?

AndyBitz (approved):

styfle (commented):
Are we worried about accidentally exporting anything?

Remember that build-utils is not versioned so once we export something, we can't easily remove it without possibly breaking Community Runtimes.

TooTallNate:
Ok good call, let's keep it explicit then.

PR #7568, opened 2022-03-15
[cli] Extend `vc pull` to pull "preview" and "production" env files as well

EndangeredMassa:
In order to know what the proper environment variables are when running `vc build` and `vc deploy --prebuilt`, we need to pull down the environment variables for that target. @TooTallNate and I decided it would probably work best if we pulled down `development`, `preview`, and `production` when `vc pull` was invoked.

We don't want to clog up the project root and the user doesn't need to care about these files. So, we're stuffing them into `.vercel`.

---

This PR extends the `vc pull` command to:

- pull down preview and production environment files, in addition to development
- names all files `.env.${target}.local`
- writes those files to `.vercel`
- refactors the command logic to make it easier for me (and hopefully others) to follow

This PR does not break existing behavior. You will still get `.env` written with the development values to your root. You can still change that location with the `--env` option.

The new `.vercel/.env.${target}.local` files will be used in follow-up PRs. After all related work is done, we can remove the root level `.env` file that's created.

---

Card: https://linear.app/vercel/issue/BUI-36/vc-pull-should-pull-environment-variables-for-all-targets


packages/cli/src/commands/pull.ts
```diff
@@ -77,32 +81,93 @@ export default async function main(client: Client) {
...
+}
+
+export default async function main(client: Client) {
+  const { argv, argResultCode } = parseArgs(client);
+  if (argResultCode !== 0) {
+    return argResultCode;
+  }
+  if (!argv) {
+    throw new Error('argResultCode was 0, but `argv` did not exist');
+  }
```
EndangeredMassa, line 145: I was trying to find a way to type this function such that it either has a non-zero exit code OR has an `argResultCode` value. That way, I wouldn't have this extra check for `argResultCode` missing that will essentially never happen. (Same goes for the other result code functions below.)

I could rely on throwing an exception in non-zero cases, making the function signature simpler to manage, but then the consuming code has to wrap in try-catch. We don't get away from extra code that's noise to the reader most of the time.

If anyone can think of a better way to type these, let me know.

packages/cli/src/commands/pull.ts
```diff
@@ -77,32 +81,93 @@ export default async function main(client: Client) {
...
-  const { project, org } = link;
+  return { link, linkResultCode: 0 };
+}
 
-  client.config.currentTeam = org.type === 'team' ? org.id : undefined;
+async function pullAllEnvFiles(
+  envFileRoot: string,
+  client: Client,
+  project: Project,
+  argv: any,
```
EndangeredMassa: I'd like to get rid of this `any`, but the dynamic type `argv` can't be referenced in static code. 

I could manually create a matching type, but that would be brittle. The inferred from usage type only gives me the subset that I currently use, but I'd like to have the full type here:

```
argv: Partial<{ '--debug': boolean; '--yes': boolean; }>
```

Is there any way to get that type into this function?

packages/cli/test/commands/pull.test.ts
```diff
@@ -31,8 +72,8 @@ describe('pull', () => {
       id: 'vercel-pull-next',
       name: 'vercel-pull-next',
     });
-    const expectedEnvFilename = '.env.vercel';
-    client.setArgv('pull', '--yes', `--env=${expectedEnvFilename}`, cwd);
+    const expectedEnvFilename = '.env.vercel.development.local';
+    client.setArgv('pull', '--yes', '--env=.env.vercel', cwd);
```
EndangeredMassa, line 76: Does the `--env` option make sense to include anymore? It seems like it was an option put in place to allow users to move the file, but we're about to stuff all of these into the `.vercel` directory.

packages/cli/test/mocks/project.ts
```diff
@@ -84,6 +95,12 @@ export function useProject(project = defaultProject) {
     res.json(project);
   });
   client.scenario.get(`/v7/projects/${project.id}/env`, (_req, res) => {
+    const target = _req.query.target;
+    if (typeof target === 'string') {
+      const targetEnvs = envs.filter(env => env.target.includes(target));
+      res.json({ envs: targetEnvs });
+    }
```
EndangeredMassa, line 102: This allows for requests to vary per target.

packages/cli/src/commands/pull.ts
```diff
@@ -77,32 +81,93 @@ export default async function main(client: Client) {
...
+  cwd: string
+) {
+  const devEnvFile = `${envFileRoot}.development.local`;
+  const pullDevPromise = envPull(
+    client,
+    project,
+    ProjectEnvTarget.Development,
+    argv,
+    [join(cwd, devEnvFile)],
```
EndangeredMassa, line 108: @TooTallNate this should be `    [join(cwd, '.vercel', devEnvFile)],`, right?

packages/cli/src/commands/pull.ts
jaredpalmer: Other than inlining the function definition into scope and using `typeof`, I don't think so unless you can abstract and pass a generic to share in both places

EndangeredMassa: I extracted the arg parsing into a separate function that can return that dynamic type, then used ` ReturnType<typeof processArgs>` to get a reference to it here. Seems to work!

jaredpalmer: Brilliant

PR #7577, opened 2022-03-16
[cli] Add support for `vc --cwd <dir>`

EndangeredMassa:
`vc`'s default command is `deploy`, which can lead to ambiguous cli invocations when running `vc dir-or-command` like:

```
$  vc list
Vercel CLI 23.1.2
Error! The supplied argument "list" is ambiguous.
If you wish to deploy the subdirectory "list", first run "cd list".
```

when run in a directory that contains a subdirectory "list". This conflict will happen with any current and future commands, like `vc build`.

In order to make sure the CLI can be invoked either way, this PR deprecates the default behavior. Going forward, a user would see the following.

**Conflicting Command, Run Command**

```bash
$ vc list
# warning: Did you mean to deploy the subdirectory "list"? Use `vc --cwd list` instead.
# ... runs the `list` command
```

**Conflicting Command, Deploy Directory**

```bash
$ vc --cwd list
# ... deploy as normal
```

---

Card: https://linear.app/vercel/issue/BUI-33/prevent-ambiguous-vc-command-oror-dir


packages/cli/test/integration.js
```diff
@@ -378,6 +392,91 @@ test('login', async t => {
...
+  t.regex(
+    stderr || '',
+    /Did you mean to deploy the subdirectory "list"\? Use `vc --cwd list` instead./
+  );
+
+  const listHeader = /project +latest deployment +state +age +username/;
+  t.regex(stdout || '', listHeader); // ensure `list` command still ran
+});
+
+test('default command should work with --cwd option', async t => {
```
EndangeredMassa, line 470: I believe I had this test set up to link properly at one point, but I can't get it. Can someone who better understands how linking works help be get this test set up?

You can see the test failure here: https://github.com/vercel/vercel/runs/5573632410?check_suite_focus=true

packages/cli/src/util/projects/get-project-by-id-or-name.ts
```diff
@@ -9,7 +9,7 @@ export default async function getProjectByNameOrId(
 ) {
   try {
     const project = await client.fetch<Project>(
-      `/projects/${encodeURIComponent(projectNameOrId)}`,
+      `/v8/projects/${encodeURIComponent(projectNameOrId)}`,
```
EndangeredMassa, line 12: This was necessary to avoid collisions with v1 routes. In the tests added here specifically, the path `/projects/list` would hit a specific route that listed all projects instead of retrieving a project named "list".

packages/cli/test/integration.js
```diff
@@ -378,6 +393,90 @@ test('login', async t => {
...
+      '--cwd',
+      target,
+      ...defaultArgs,
+    ],
+    {
+      cwd: projectDir,
+    }
+  );
+
+  t.is(exitCode, 0, formatOutput({ stderr, stdout }));
```
styfle, line 489: Lets also assert `stderr` is empty here

packages/cli/test/integration.js
```diff
@@ -1088,30 +1187,6 @@ test('output the version', async t => {
   t.is(version, pkg.version);
 });
 
-test('should error with suggestion for secrets subcommand', async t => {
```
styfle, line 1091: Why is this test being removed?

packages/cli/test/mocks/project.ts
```diff
@@ -77,10 +77,10 @@ export const defaultProject = {
...
 export function useProject(project = defaultProject) {
-  client.scenario.get(`/projects/${project.name}`, (_req, res) => {
+  client.scenario.get(`/v8/projects/${project.name}`, (_req, res) => {
     res.json(project);
   });
-  client.scenario.get(`/projects/${project.id}`, (_req, res) => {
+  client.scenario.get(`/v8/projects/${project.id}`, (_req, res) => {
     res.json(project);
   });
   client.scenario.get(`/v7/projects/${project.id}/env`, (_req, res) => {
```
styfle: Should this one also be `v8`?

packages/cli/src/index.ts
```diff
@@ -390,41 +390,32 @@ const main = async () => {
       const fileType = lstatSync(targetPath).isDirectory()
         ? 'subdirectory'
         : 'file';
```
styfle: Do we still support deploying a single file like this? I thought it was removed recently and you can only deploy a directory?

packages/cli/src/index.ts
```diff
@@ -390,41 +390,32 @@ const main = async () => {
...
+        );
+      }
+
       subcommand = 'deploy';
     }
   } else {
     debug('user supplied no target, defaulting to deploy');
+    output.warn(
+      'calling `vc` without a command or the `--cwd` option is deprecated, use `vc --cwd directory` or `vc deploy` instead'
+    );
```
styfle: Do we want to warn in the general case like this since there is no conflict?

Seems like its one of the nice things to be able to type two letters to deploy: `vc`

packages/cli/test/integration.js
EndangeredMassa: The suggestion logic for alternate singular/plural was removed. It was only used when there was a conflict between a subdirectory and a command. We have a better way to resolve that now, which is to use the flag if you meant to deploy or let the command run, which is the new behavior.

I thought about moving this logic to the failure cause where the first argument was neither a command or a subdirectory, but this logic was very narrow. It checked plural/singular variants by adding or removing a trailing `s`, which would miss things like `alias` -> `aliases` because it would check `aliass`. 

So, I removed it instead.

packages/cli/src/index.ts
EndangeredMassa: I agree that is nice. I think it would be fine to leave the `vc` invocation alone. I called it out in the description under "Alternatives".

@TooTallNate what do you think?

packages/cli/src/index.ts
EndangeredMassa: Ah yes. I'll remove that.

@TooTallNate you were telling about some quirk of this decision, but I forget what it was. Do you see issue here?

packages/cli/test/integration.js
EndangeredMassa: That reminds me: the `subdirectory-secret` fixture is now no longer used. I'll delete that too.

packages/cli/test/integration.js
EndangeredMassa: `stderr` is used for a bunch of non-error logs. Here's an example from this test:

```
Vercel CLI 24.0.1
Retrieving project…
Deploying endangeredmassa/list
Inspect: https://vercel.com/endangeredmassa/list/6BP3c8mUz536cMPcW8RecTrN3Evh [862ms]
Queued
Building
Completing
Preview: https://list-endangeredmassa.vercel.app [copied to clipboard] [5s]
To deploy to production (list-rosy.vercel.app), run `vercel --prod`
```

I added an assertion that the warning doesn't show up, at least.

packages/cli/src/index.ts
TooTallNate: Correct. There's probably a few places in the codebase that we can simplify now because of that. Perhaps not relevant for this PR though.

packages/cli/src/index.ts
TooTallNate: `vc` by itself is fine but `vc <some-dir>` should print a warning, otherwise someone might rely on that in i.e. their CI and that makes it hard to add new subcommands (`vc build` to deploy the static "build" directory for example).

EndangeredMassa: I removed the deprecation warning

TooTallNate: Let's update the PR description as well since it says the bare `vc` version prints a warning which has now been changed.

packages/cli/test/integration.js
styfle: Maybe I'm missing it. Which line is the assertion? Did you forget to git push?

EndangeredMassa: Sorry, I added a `notRegex` for this, but a later change removed it. We're no longer showing that warning at all, which makes testing for it not showing up here useless.

Now stderr will only have the example I pasted above. Is there anything useful we can assert here given the latest state of the code?

styfle (approved):
Thanks!

packages/cli/src/index.ts
styfle: @EndangeredMassa The PR title and description are still outdated

EndangeredMassa: I think the title is still correct because it's talking about `vc <dir>`, not `vc`. I updated the description.

styfle: How about `Add support for vc --cwd <dir>` since the warning doesn't say anything about deprecation

EndangeredMassa: That is better. Updated.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -82,6 +82,7 @@ export default async (client: Client) => {
       '--prebuilt': Boolean,
       '--prod': Boolean,
       '--confirm': Boolean,
+      '--cwd': String,
```
styfle: Can we make this a global flag so it works with other subcommands?

I think we can simplify this to be `process.chdir(cwd)` before any subcommands run.

EndangeredMassa: That sounds good to me. Updated.

packages/client/package.json
```diff
@@ -50,6 +50,7 @@
     "minimatch": "5.0.1",
     "ms": "2.1.2",
     "node-fetch": "2.6.1",
+    "recursive-readdir": "2.2.2",
```
EndangeredMassa: I don't know what shifted, but `client` seems to have an indirect dependency on `recursive-readdir`. When I run the cli tests locally, I'd get a failure saying this package (`client`) couldn't find `recursive-readdir`. I made sure I had yarn install and project built. The only thing that resolved it was to add the dependency here.

Any suggestions on how to dig deeper are most welcome.

EndangeredMassa: When trying to reproduce one last time to get the exact error message, it's building fine. Maybe I was in an inconsistent local state?

packages/cli/src/commands/deploy/args.ts
```diff
@@ -42,6 +42,7 @@ export const help = () => `
 
     -h, --help                     Output usage information
     -v, --version                  Output the version number
+    --cwd                          Current working directory
```
EndangeredMassa, line 45: This seems to be where top-level args are described in `help`. Is this the right place?

styfle: I think so.

We'll also want to update the "global options" section of the docs once this lands https://vercel.com/docs/cli#introduction/global-options

packages/cli/src/index.ts
```diff
@@ -387,33 +392,10 @@ const main = async () => {
...
-            `\nIf you wish to deploy the ${fileType} ${param(
-              targetOrSubcommand
-            )}, first run "cd ${targetOrSubcommand}". ` +
-            (alternative
-              ? `\nIf you wish to use the subcommand ${param(
-                  targetOrSubcommand
-                )}, use ${param(alternative)} instead.`
-              : '')
-        )
+      output.warn(
```
styfle, line 395: Now that I think about it,  we might still need to check if `--cwd` was used before warning because it should never warn in that case, right?

Lets test `vc list --cwd <dir>` when the dir contains another dir called `list`. I suspect that will warn when it shouldn't.

packages/cli/src/commands/deploy/args.ts
EndangeredMassa: Good call. I made a card for that: https://linear.app/vercel/issue/BUI-58/document-global-cli-option-cwd

packages/cli/src/index.ts
EndangeredMassa: You are right. I added a test for that and it did still warn.

I just added a check for the arg being used. Now it does not warn in that case.

TooTallNate (approved):

PR #7584, opened 2022-03-18
[static-build] Don't fail CRA builds on warnings

leerob:
Don't fail the build on warnings from Create React App. Node.js will load 'false' as a string, not a boolean, so it's truthy still. This is to ensure we don't accidentally break other packages that check if process.env.CI is true somewhere.
    
- https://github.com/facebook/create-react-app/issues/2453
- https://github.com/facebook/create-react-app/pull/2501
- https://github.com/vercel/community/discussions/30

### DIFF
--- a/packages/static-build/src/index.ts
+++ b/packages/static-build/src/index.ts
@@ -389,12 +389,9 @@ export const build: BuildV2 = async ({
     */
     if (framework && framework.slug === 'create-react-app') {
       if (!spawnOpts.env) {
-        spawnOpts.env = {
-          CI: 'false',
-        };
-      } else {
-        spawnOpts.env.CI = 'false';
+        spawnOpts.env = {};
       }
+      spawnOpts.env.CI = 'false';
     }
 
     if (meta.isDev) {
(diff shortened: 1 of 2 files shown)


styfle (commented):
Please add a test fixture so we can confirm this works as expected

leerob:
Looks like there was an existing fixture with `CI=false` set, so I removed that from there and the test/build should still pass. Does that work, or should we add another test? And would it be part of integration test suite, or the builds test suite (in static-build)?

Timer (approved):
Looks legit for the test because there is a warning that's triggered:
https://github.com/vercel/vercel/blob/70468db557409161e1d5e37a972256779f35f398/packages/static-build/test/fixtures/12-create-react-app/src/App.js#L1

packages/static-build/src/index.ts
```diff
@@ -378,6 +378,25 @@ export const build: BuildV2 = async ({
...
+    https://github.com/vercel/community/discussions/30
+    */
+    if (framework && framework.slug === 'create-react-app') {
+      if (!spawnOpts.env) {
+        spawnOpts.env = {
+          CI: 'false',
+        };
+      } else {
+        spawnOpts.env.CI = 'false';
+      }
```
styfle:
```suggestion
      if (!spawnOpts.env) {
        spawnOpts.env = {};
      }
      spawnOpts.env.CI = 'false';
```

This might read better

styfle (approved):
Thanks!

PR #7592, opened 2022-03-21
[redwood] Add default rewrites, cleanUrls and nested API function support

dac09:
Working with @TooTallNate to update the Vercel Redwood builder, in preparation for the Redwood 1.0 launch.

**- Support for nested api functions in dist folder**
This is essentially a change to the glob to pick up functions 1 level deep in the dist folder. In prior version of Redwood, we had a workaround specific to Vercel that we would like to take out. See https://github.com/redwoodjs/redwood/pull/4813

**- ZeroConfig redirects**
Redwood now generates a 200.html file on every build, which is the "fallback" page for when a html file cannot be matched with the path. This PR adds support for using this 200.html file, out of the box for Redwood projects on Vercel

This is to make it easy for users to deploy a partly prerendered project - without having to mess with the vercel.json file.

### 📋 Checklist

- [x] Validate that the change is backwards compatible
- [x] Add extra tests to the builder (need a bit of guidance)
- [x] Address why the routing config in the builder does not have the desired affect

#### Tests

#### Code Review


packages/redwood/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vercel/redwood",
-  "version": "0.5.2-canary.3",
+  "version": "1.0.0-canary.2",
```
dac09: I will make this 1.0.0 when everything is ready to go, unless advised otherwise :)

examples/remix/vercel.json
```diff
@@ -0,0 +1,7 @@
+{
+  "build": {
+    "env": {
+      "ENABLE_FILE_SYSTEM_API": "1"
+    }
+  }
+}
```
styfle: Something seems wrong. A lot of files that are not related to redwood have been changed 🤔

dac09:
I believe these are from the commits that were added by you guys? I can rebase only my changes though if it helps.

styfle:
Yes please rebase. This should also fix the merge conflicts reported by GitHub below.

dac09:
@styfle done :)

packages/redwood/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vercel/redwood",
-  "version": "0.6.0",
+  "version": "1.0.0-canary.17",
```
dac09: Please let me know if you'd like me to update this version to 1.0.0 or something your CI will handle

TooTallNate: Let's leave it untouched for now. The release process bumps the version, but we'll be sure to do a major bump after we merge this.

```suggestion
  "version": "0.6.0",
```

packages/redwood/src/index.ts
```diff
@@ -150,10 +160,47 @@ export const build: BuildV2 = async ({
...
+      staticOutputs[fileName] = fileFsRef;
+    } else {
+      const fileNameWithoutExtension = pathBasename(fileName, '.html');
+
+      const pathWithoutHtmlExtension = join(
+        parsedPath.dir,
+        fileNameWithoutExtension
+      );
+
+      renameSync(fileFsRef.fsPath, pathWithoutHtmlExtension);
```
TooTallNate: Is there a need to rename the file on the actual filesystem?

TooTallNate:
This is looking great! Could you please add a test dir to `pacakges/redwood/test/fixtures` that relies on 1. one level deep API functions and 2. the stripped HTML extension?

packages/redwood/src/index.ts
```diff
@@ -237,9 +284,30 @@ export const build: BuildV2 = async ({
...
+  // Note that in builder post-processing, we remove the .html extension
+  const fallbackHtmlPage = existsSync(join(webDistPath, '200'))
+    ? '/200'
+    : '/index';
+
+  const defaultRoutesConfig = getTransformedRoutes({
+    nowConfig: {
+      // this makes sure we send back 200.html for unprerendered pages
+      rewrites: [{ source: '/(.*)', destination: fallbackHtmlPage }],
+      // @NOTE leave clean urls so vercel redirects are generated
```
TooTallNate: Remove?

```suggestion
```

packages/redwood/src/index.ts
styfle: It shouldn't need to be renamed. The `staticOutputs` should map the name to the file so it can be renamed by changing the key.

In fact, removing this rename line should work.

packages/redwood/src/index.ts
```diff
@@ -150,10 +160,47 @@ export const build: BuildV2 = async ({
...
-  const functionFiles = await glob('*.js', apiDistPath);
+  // Also supports nested functions like:
+  // ├── functions
+  // │   ├── bazinga
+  // │   │   ├── bazinga.js
+  // │   ├── graphql.js
+
+  const functionFiles = {
+    ...(await glob('*.js', apiDistPath)), // top-level
+    ...(await glob('*/*.js', apiDistPath)), // one-level deep
```
styfle, line 201: Can you use a single `**/*.js` instead of two globs?

packages/redwood/src/index.ts
dac09: Misunderstood what Nate suggested. I'll try removing the rename!

packages/redwood/src/index.ts
dac09: This is intentional, suggested by Nate. We only want to go two levels deep - because that is the "most compatible" option  e.g. Netlify behaves the same way, and I'm also changing our api server to also adopt this behaviour. Previously it was one level deep, because of Vercel - now we're good everywhere :) 

In Redwood typically we use the "deep" option 
```
export const findApiDistFunctions = (cwd: string = getPaths().api.base) => {
  return fg.sync('dist/functions/**/*.{ts,js}', {
    cwd,
    deep: 2, // 👈 We don't support deeply nested api functions, to maximise compatibility with deployment providers
    absolute: true,
  })
}
```
But the glob func from buildTools doesn't support this, which is why theres the two globs!

packages/redwood/src/index.ts
dac09: So... if I remove the rename I'm getting this error

```
🔍  Inspect: https://vercel.com/redwoodjs/prerender-redirects/8Ku5ZWg9hojbdv1JiVDkyfhjf6pP [2s]
Error! ENOENT: no such file or directory, open '/vercel/path1/web/dist/200'
Error! Check your logs at https://prerender-redirects-8406jwmyu-redwoodjs.vercel.app/_logs
```

```
Failed to process build result for "200". Data: {"type":"FileFsRef"}.
[Error: ENOENT: no such file or directory, open '/vercel/path1/web/dist/200'] {
```

Welcome to view logs - nothing private in the repo

packages/redwood/src/index.ts
```diff
@@ -150,10 +160,47 @@ export const build: BuildV2 = async ({
...
+      const fileNameWithoutExtension = pathBasename(fileName, '.html');
+
+      const pathWithoutHtmlExtension = join(
+        parsedPath.dir,
+        fileNameWithoutExtension
+      );
+
+      renameSync(fileFsRef.fsPath, pathWithoutHtmlExtension);
+      fileFsRef.contentType = 'text/html; charset=utf-8';
+      fileFsRef.fsPath = pathWithoutHtmlExtension;
```
TooTallNate:
```suggestion
      fileFsRef.contentType = 'text/html; charset=utf-8';
```

packages/redwood/src/index.ts
TooTallNate: It's because you're still also updating the `fsPath`. Keep that pointing to the `.html`-suffixed file like so: https://github.com/vercel/vercel/pull/7592/files#r836676763

packages/redwood/src/index.ts
```diff
@@ -237,9 +284,30 @@ export const build: BuildV2 = async ({
     lambdaOutputs[outputName] = lambda;
   }
 
+  // Older versions of redwood did not create 200.html automatically
+  // From v0.50.0+ 200.html is always generated as part of web build
+  // Note that in builder post-processing, we remove the .html extension
+  const fallbackHtmlPage = existsSync(join(webDistPath, '200'))
```
TooTallNate: Since we wouldn't be renaming the source file on the filesystem:

```suggestion
  const fallbackHtmlPage = existsSync(join(webDistPath, '200.html'))
```

dac09:
Thanks @TooTallNate and @styfle

- [x] Added new fixture with nested functions and prerendered homepage (where cleanUrls are normally used)
- [x] Removed `renameSync` to let the builder do its thing
- [x] Cleaned up as per suggestions

Note: re-nested functions, I will be removing the workaround we had for vercel _after_ this builder goes live so that changes are non-breaking. I will come back and update the redwood version in this fixture after that!

PR #7599, opened 2022-03-21
[api] Convert more API endpoints to static

TooTallNate:
This is a continuation of #7596, which converts the following endpoints to static files:

* `/api/frameworks`
* `/api/examples/download/[segment]`

The endpoint `/api/examples/info` remains as a Serverless Function, since it changes its response based on the provided query parameter.

This PR also removes the dependence on `ENABLE_FILE_SYSTEM_API` and works with `vc build`.


utils/vercel-build/build.ts
```diff
@@ -0,0 +1,64 @@
...
+const EXAMPLES_DIR = join(ROOT_DIR, 'examples');
+const PUBLIC_DIR = join(ROOT_DIR, 'public');
+
+async function main() {
+  // Start fresh
+  await fs.remove(PUBLIC_DIR);
+  await fs.mkdirp(PUBLIC_DIR);
+
+  // Output `frameworks.json`
+  const frameworks = (_frameworks as Framework[])
```
styfle, line 21: This cast is no longer necessary

.vercelignore
```diff
@@ -1,17 +1,19 @@
...
 !yarn.lock
 !package.json
 
 # api
-!api/
+!api
 !api/**
 
 # packages
-!packages/
```
styfle, line 15: I think the slash is necessary since there was a but in Vercel CLI negated patterns 

did you try deploying from the CLI?

utils/vercel-build/build.ts
TooTallNate: Screen Shot 2022-03-21 at 3 23 44 PM

Not so sure about that.

.vercelignore
TooTallNate: CLI deploy seems fine. Might have been a bug that was since resolved (but removing the slash was necessary in this case for `vc build` to build the proper files selection 🤷).

utils/vercel-build/build.ts
styfle: Oh because sort() is mutating the array 😬

utils/vercel-build/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
styfle, line 14: This can be es2018 or even es2020 I believe.

Might want to also remove all the comments

api/package.json
```diff
@@ -5,11 +5,10 @@
   "description": "API for the vercel/vercel repo",
   "main": "index.js",
   "scripts": {
-    "vercel-build": "node ../utils/run.js build all"
+    "vercel-build": "yarn --cwd .. build"
```
styfle, line 8: Do we still need this API? I thought this PR was making all the endpoints static?

utils/vercel-build/build.ts
```diff
@@ -0,0 +1,97 @@
...
+  const list = [...existingExamples, ...oldExamples];
+  await fs.writeJSON(join(PUBLIC_DIR, 'list.json'), list, {
+    spaces: 2,
+  });
+  console.log(`Wrote "list.json"`);
+
+  // Output `list-all.json`
+  const examplesList = getExampleList();
+  await fs.writeJSON(join(PUBLIC_DIR, 'list-all.json'), examplesList, {
+    spaces: 2,
```
styfle, line 71: I dont think we were previously using pretty spaces at all, it was just the minified json.

PR #7600, opened 2022-03-21
Downgrade Hugo Version again.

leerob:
I didn't realize we had a hardcoded version in the build container.

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1772,7 +1772,7 @@ export const frameworks = [
 
       return (config && config.publishDir) || 'public';
     },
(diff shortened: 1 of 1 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -1772,7 +1772,7 @@ export const frameworks = [
 
       return (config && config.publishDir) || 'public';
     },
-    defaultVersion: '0.64.0',
+    defaultVersion: '0.92.2',
```
leerob:
```suggestion
    defaultVersion: '0.58.2',
```

leerob:
Force merging - this is still broken in prod currently.

PR #7602, opened 2022-03-21
[examples] add typescript as dev dependency for nextjs example

jescalan:
Resolves a warning that typescript was needed as a peer dependency

### DIFF
--- a/examples/nextjs/package.json
+++ b/examples/nextjs/package.json
@@ -12,7 +12,8 @@
     "react-dom": "17.0.2"
   },
   "devDependencies": {
-    "eslint": "8.9.0",
-    "eslint-config-next": "12.1.0"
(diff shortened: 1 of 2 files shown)


styfle (commented):
Has this been fixed in `create-next-app`?

jescalan:
Never mind this kind of seems like a 🐰 🕳️ - the addition of the dev dependency created a chain of odd unrelated errors on build, and since this example doesn't use typescript it shouldn't be necessary in the first place. Closing.

Edit: never mind, re-opening and re-running tests seemed to resolve the issues?

TooTallNate:
> since this example doesn't use typescript it shouldn't be necessary in the first place

I'm curious about this part as well.

PR #7608, opened 2022-03-23
Set `ENABLE_FILE_SYSTEM_API=1` in vercel.json

TooTallNate:
Temporarily set this in `vercel.json` since the current way the Vercel deployment is structured requires it to be set.

Can remove this once #7599 is merged.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -46,5 +46,10 @@
         }
       ]
     }
-  ]
+  ],
+  "build": {
+    "env": {
+      "ENABLE_FILE_SYSTEM_API": "1"
+    }
+  }
 }


vercel.json
```diff
@@ -46,5 +46,10 @@
         }
       ]
     }
-  ]
+  ],
+  "build": {
```
styfle, line 50: This is already on line 21

styfle (changes requested):
This is a duplicate key. Maybe your other branch was outdated?

vercel.json
TooTallNate: 🤦

TooTallNate:
Lol good catch. True dat.

PR #7615, opened 2022-03-24
[client] Don't consider `rootDirectory` for --prebuilt buildFileTree()

TooTallNate:
`vc build` outputs to the root-level `.vercel` directory, so no need to consider `rootDirectory` of the Project Settings when collecting this file list.

### DIFF
--- a/packages/client/tests/unit.utils.test.ts
+++ b/packages/client/tests/unit.utils.test.ts
@@ -127,49 +127,4 @@ describe('buildFileTree()', () => {
       normalizeWindowsPaths(ignoreList).sort()
     );
   });
-
-  it('should find root files but ignore all .output files when prebuilt=false and rootDirectory=root', async () => {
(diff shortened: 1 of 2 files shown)


packages/client/tests/unit.utils.test.ts
```diff
@@ -127,49 +127,4 @@ describe('buildFileTree()', () => {
       normalizeWindowsPaths(ignoreList).sort()
     );
   });
-
-  it('should find root files but ignore all .output files when prebuilt=false and rootDirectory=root', async () => {
```
styfle, line 131: I think we need to keep a test with `rootDirectory` assigned, to make sure this change works as expected.

styfle: Oh I see, this was a unit test so the prop was removed 😆

styfle (approved):

PR #7616, opened 2022-03-24
[client] Drop "/now" prefix from API endpoints

TooTallNate:
This prefix is not necessary for the API endpoints so remove.

Also updated a reference of `NowConfig` to use `VercelConfig` instead.


styfle (approved):
Lets make sure the tests pass, because this didn't work last time I tried it

PR #7619, opened 2022-03-24
[cli] Allow projectSettings override in vercel.json

Ethan-Arrowood:
This PR adds the required changes to the CLI to support overridable `vercel.json` `projectSettings` values. This PR is linked with changes on the API side too.

- The core changes are in `packages/cli/src/commands/deploy/index.ts`
- An unused code path was removed from `packages/client/src/create-deployment.ts`, `packages/cli/src/util/deploy/process-deployment.ts` and `packages/cli/src/util/index.ts`.

This will only work for *existing* deployment (not *new* ones). This is implemented by checking the `status` of the deployment (is it linked or not). An error is thrown if the user attempts to send overrided settings to a new project. 

A warning is outputted when overrided settings are discovered and shares what settings are being overridden.

#### Tests

#### Code Review


packages/cli/src/util/deploy/create-deploy.ts
```diff
@@ -19,7 +19,7 @@ export default async function createDeploy(
   cwd?: string
 ): Promise<any | DeploymentError> {
   try {
-    return await now.create(paths, createArgs, org, isSettingUpProject, cwd);
+    return now.create(paths, createArgs, org, isSettingUpProject, cwd);
```
TooTallNate: We still want `await` here otherwise a thrown error won't be caught in the try/catch here.

Ethan-Arrowood: i feel like the async/await is the wrong pattern here, but i don't think this is a high volume promise call so there is little point trying to optimize. If it makes the error handling easier than no need to change either 😄

packages/cli/src/commands/deploy/index.ts
```diff
@@ -460,7 +460,10 @@ export default async (client: Client) => {
 
     if (!localConfig.builds || localConfig.builds.length === 0) {
       // Only add projectSettings for zero config deployments
-      createArgs.projectSettings = { sourceFilesOutsideRootDirectory };
+      createArgs.projectSettings = {
+        ...localConfig.projectSettings,
```
Ethan-Arrowood, line 472: this is the line that applies the projectSettings to the deployment

packages/cli/src/commands/deploy/index.ts
```diff
@@ -321,7 +321,7 @@ export default async (client: Client) => {
       `${prependEmoji(
         `The ${code('name')} property in ${highlight(
           localConfig[fileNameSymbol]!
-        )} is deprecated (https://vercel.link/name-prop)`,
+        )} is deprecated (https://vercel.link/name-prop). Use the settings UI or projectSettings.projectName instead.`,
```
styfle: I don't think we want to encourage projectName either

packages/client/src/create-deployment.ts
```diff
@@ -76,29 +72,6 @@ export default function buildCreateDeployment() {
...
-      // If the user did not provide a config file, use the one in the root directory.
-      const relativePaths = fileList.map(f => relative(cwd, f));
-      const hasVercelConfig = relativePaths.includes('vercel.json');
-      const hasNowConfig = relativePaths.includes('now.json');
-
-      if (hasVercelConfig) {
-        if (hasNowConfig) {
-          throw new DeploymentError({
-            code: 'conflicting_config',
-            message:
```
styfle, line 90: Why was this removed?

Ethan-Arrowood: `configPath` and `nowConfig` are never used throughout the function.

Ethan-Arrowood: so this whole block is useless; and thus the `nowConfig` function parameter is similarly useless.

styfle: Strange because there is a test for this error message here

https://github.com/vercel/vercel/blob/b2bfae6e2e15237103607ab0f783617a0af1e45e/packages/cli/test/integration.js#L3210-L3229

Ethan-Arrowood: Yes because the if condition will still run when the config is omitted, but since that config is never used in this block I think the error message is pointless too. I will remove that test as well

packages/cli/src/commands/deploy/index.ts
```diff
@@ -153,6 +153,27 @@ export default async (client: Client) => {
...
+        'framework',
+        'name',
+        'nodeVersion',
+        'commandForIgnoringBuildStep',
+        'rootDirectory',
+        'sourceFilesOutsideRootDirectory',
+      ];
+      for (const key of Object.keys(localConfig.projectSettings)) {
+        if (!allowedProperties.includes(key)) {
+          output.error(`Disallowed property ${key} found in "projectSettings"`);
```
```suggestion
          output.error(`Unknown property ${key} found in "projectSettings"`);
```

We should consider this unknown just like if there was any other unknown property.

What happens when you do `{ projectSettings: { "foo": "bar" } }`?

Ethan-Arrowood: I think you'll eventually hit an API error when this object is validated by the API using the schema

styfle: Lets remove this and validate in the API instead of the CLI

packages/client/src/create-deployment.ts
styfle: @Ethan-Arrowood We need to keep the test. We must throw when both now.json exist and vercel.json exist. I suspect the error might be thrown elsewhere too which is why the test still passed.

Ethan-Arrowood: Upon review, this error is thrown from elsewhere in the code base. The test passes with no issue.

packages/cli/src/commands/deploy/index.ts
Ethan-Arrowood: So the create-deployment api route will need to use a different schema than update-project. Okay ✅

packages/cli/test/integration.js
```diff
@@ -3580,3 +3580,76 @@ test('[vc link] should support the `--project` flag', async t => {
...
+      stdout: deployment.stdout,
+    })
+  );
+
+  const page = await fetch(deployment.stdout);
+  const text = await page.text();
+  t.is(text, `Hello, World\n`);
+});
+
+test('invalid vercel.json projectSettings should result in a deployment error', async t => {
```
Ethan-Arrowood: This test won't pass until API is updated. This gives me good reason to focus on merging API changes first and then we can wrap up these CLI changes.

packages/build-utils/src/types.ts
```diff
@@ -320,17 +320,29 @@ export interface BuilderFunctions {
...
   rootDirectory?: string | null;
-  createdAt?: number;
-  autoExposeSystemEnvs?: boolean;
+  serverlessFunctionRegion?: string | null;
   sourceFilesOutsideRootDirectory?: boolean;
-  directoryListing?: boolean;
-  gitForkProtection?: boolean;
+  ssoProtection?: {
+    deploymentType: 'all' | 'preview';
+  };
```
styfle: Do we need two types now? One to represent overridable project settings and one to represent all setting on the project?

packages/cli/src/commands/deploy/index.ts
styfle: You might also need to add this to the `missing_project_settings` error handling.

You can test this by making two deployments to the same project. The first one will return early and enter the `missing_project_settings` branch and ask the use to edit the settings if they wish to override. If the `vercel.json` has settings then those should settings automatically populate instead of asking the user again. Then the second deployment should not hit the `missing_project_settings` branch and accept the deployment on the first api call.

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,79 @@ test('[vc link] should support the `--project` flag', async t => {
...
+  // Pass a disallowed command
+  vercelConfig['projectSettings'] = {
+    autoExposeSystemEnvs: true,
+  };
+
+  await writeFile(vercelJsonPath, JSON.stringify(vercelConfig));
+
+  await testDeployment();
+
+  await writeFile(vercelJsonPath, vercelJsonOriginal);
```
styfle: We also need a test that creates a new deployment without custom project settings, then modifies the project settings for a second deployment via vercel.json

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,79 @@ test('[vc link] should support the `--project` flag', async t => {
...
+  async function testDeployment() {
+    try {
+      await execa(binaryPath, [
+        directory,
+        ...defaultArgs,
+        '--public',
+        '--confirm',
+      ]);
+      t.fail('Deployment should fail');
+    } catch ({ exitCode, stderr, stdout }) {
```
styfle: You can do this without try/catch by passing `reject: false` option to execa (see other tests). That will ensure the test fails if the exit code was successful.

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,79 @@ test('[vc link] should support the `--project` flag', async t => {
...
+      );
+    }
+  }
+
+  const vercelJsonPath = path.join(directory, 'vercel.json');
+  const vercelJsonOriginal = await readFile(vercelJsonPath, 'utf8');
+  const vercelConfig = JSON.parse(vercelJsonOriginal);
+
+  // Pass a command that does not exist
+  vercelConfig['projectSettings'] = {
```
styfle:
```suggestion
  vercelConfig.projectSettings = {
```

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,79 @@ test('[vc link] should support the `--project` flag', async t => {
...
+    command_does_not_exist: 'echo "foo bar"',
+  };
+
+  await writeFile(vercelJsonPath, JSON.stringify(vercelConfig));
+
+  await testDeployment();
+
+  // Pass a disallowed command
+  vercelConfig['projectSettings'] = {
```
styfle:
```suggestion
  vercelConfig.projectSettings = {
```

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,79 @@ test('[vc link] should support the `--project` flag', async t => {
...
 });
+
+test('vercel.json projectSettings should be applied to deployment', async t => {
+  const directory = path.join(__dirname, 'fixtures/unit/vercel-json-overrides');
+
+  const deployment = await execa(binaryPath, [
+    directory,
+    ...defaultArgs,
+    '--public',
+    '--confirm',
```
styfle: We need one more test that doesn't use `--confirm`.

In that case, I think it should utilize vercel.json to pre-populate answers to the questions and only ask the questions that vercel.json didn't answer.

This would match how it works for the https://vercel.com/new page when importing a new git repo.

packages/cli/src/commands/deploy/index.ts
Ethan-Arrowood: Wouldn't the settings from `vercel.json` be applied on the first deployment?

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,140 @@ test('[vc link] should support the `--project` flag', async t => {
...
+
+  const page = await fetch(deployment.stdout);
+  const text = await page.text();
+  t.is(text, `Hello, World\n`);
+});
+
+test('invalid vercel.json projectSettings should result in a deployment error', async t => {
+  const directory = path.join(
+    __dirname,
+    'fixtures/unit/vercel-json-overrides-error'
```
styfle: All of these tests should be moved from `fixtures/unit` to `fixtures/integration`

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,140 @@ test('[vc link] should support the `--project` flag', async t => {
...
+      ...defaultArgs,
+      '--public',
+    ]);
+
+    return deployment;
+  }
+
+  // Deploy without projectSettings
+  await writeFile(vercelJsonPath, JSON.stringify({}));
+  assertDeployment(await deploy(), 0);
```
styfle: You also need to visit the deployment to see if the output directory was assigned properly. It should fetch the root index file (you could use index.txt instead of index.html to make this a simple content comparison)

packages/cli/src/commands/deploy/index.ts
styfle: It depends if the new project should be created with the settings from vercel.json

packages/cli/test/fixtures/unit/vercel-json-overrides-error/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "vercel-json-overrides-error"
+}
```
styfle: Why is this one an error? I think you can probably reuse the same test fixture for the successful test and error test.

packages/cli/test/integration.js
```diff
@@ -3675,3 +3675,140 @@ test('[vc link] should support the `--project` flag', async t => {
...
+      })
+    );
+  }
+
+  async function deploy() {
+    const deployment = await execa(binaryPath, [
+      directory,
+      ...defaultArgs,
+      '--public',
+    ]);
```
styfle: We also need to test a single setting such as 

```
{
  projectSettings: {
    outputDirectory: "foo"
  }
}
```

Make sure it overrides one but allows defaults of the others

packages/cli/src/commands/deploy/index.ts
```diff
@@ -195,6 +195,16 @@ export default async (client: Client) => {
   let sourceFilesOutsideRootDirectory: boolean | undefined = true;
 
   if (status === 'not_linked') {
+    if (
+      localConfig &&
+      Object.prototype.hasOwnProperty.call(localConfig, 'projectSettings')
+    ) {
```
```suggestion
    if (localConfig?.projectSettings) {
```

packages/cli/src/commands/deploy/index.ts
styfle: Now that we have an error above for new projects, I think this should be fine.

packages/cli/src/commands/deploy/index.ts
Ethan-Arrowood: Lol i forget that this syntax exists sometimes 😅

styfle (approved):
Thanks!

packages/cli/src/commands/deploy/index.ts
```diff
@@ -195,6 +195,13 @@ export default async (client: Client) => {
   let sourceFilesOutsideRootDirectory: boolean | undefined = true;
 
   if (status === 'not_linked') {
+    if (localConfig?.projectSettings) {
+      output.error(
+        'New projects cannot contain `projectSettings` in vercel.json'
```
TooTallNate: This error message is a bit confusing. Why can't they? Perhaps we should have a "More info" link that explains why this is the case and what the user should do about it.

Ethan-Arrowood: Ill add a link to the incoming projectSettings section and then expand on the reasoning there

TooTallNate (approved):

styfle (approved):

TooTallNate (approved):

Issue #7621, opened 2022-03-25
When using GO Runtime with vercel dev, accessing the api get an error


await-ovo:
**Describe the bug**

When the go api function refers to variables of other api, it will get an error use vercel dev, but deploying to the preview environment is ok.

**Expected behavior**

Should work fine when a api use variables from other api.

**To Reproduce**

This is the reposity to reproduce: https://github.com/await-ovo/go-rutime-demo

1. `npm install -g vercel`
2. `git clone https://github.com/await-ovo/go-rutime-demo.git`
3. `cd go-runtime-demo`
4. `vercel dev`
5. open `http://localhost:3000/api/json` in browser.

Then you can see the error in terminal as below:

<img width="1238" alt="relative-error" src="https://user-images.githubusercontent.com/41503212/160086059-055f8dd5-7c0d-4c6b-845a-afd86a27b8eb.png">

After I debugged  log the preview deployment  and the `@vercel/go`code,  I found that using the full `fsPath` for the entrypoint when analyzing the entrypoint code can avoids this problem:

<img width="1282" alt="image" src="https://user-images.githubusercontent.com/41503212/160090446-5ee51336-e57c-44b2-b867-ff2fdd1271c2.png">

After re-dev, the above error disappeared, but a new one appeared:

![image](https://user-images.githubusercontent.com/41503212/160090144-7aa8eecb-398f-49a6-8a0d-a4de4bfa075c.png)

The `api/json.go` uses the `struct Circle` exported from the file `api/hello.go`:

 
<img width="660" alt="api-json" src="https://user-images.githubusercontent.com/41503212/160091217-5638de28-ef9a-424f-b809-de1f9144d9a1.png">

I was wondering if it is possible to copy the relevant files returned by the `getAnalyzedEntrypoint` function to the cache directory along with entrypoint. I tried it locally and all the api's inside my demo work fine:
<img width="1194" alt="image" src="https://user-images.githubusercontent.com/41503212/160095515-d3129043-3cae-43fb-9751-d5b4c338918e.png">

I tried to raise a pr to solve this problem, thanks for review. :)

> I haven't run integration tests locally yet because I keep getting problems with token setting errors

PR #7622, opened 2022-03-25
 [go] fix vercel dev error when a api use variables from other file (#7621)

await-ovo:
### Related Issues

> Related to #7621 

#### Tests

#### Code Review


styfle (commented):
Thanks for the PR!

Do you have an example that fails today so it can be added as a test fixture?

await-ovo:
yes，i have create a demo to reproduce in the related issue

await-ovo:
> Thanks for the PR!
> 
> Do you have an example that fails today so it can be added as a test fixture?



Thanks for the review, I've added the test fixture now(`/packages/go/test/fixtures/21-use-variable-form-other-api`), and I've changed the pattern in `util/analyze.go` to ignore the files under the `.now` and `.vercel` directories.
image


I've also executed `yarn test-integration-once` locally, and **everything else worked except for the `fixtures/01-cowsay` fixture, but I'm confused because I'm getting the same error on the main branch**.
Pasted Graphic 3


I think the code changes in this pr will only affect dev, I'm not sure where the problem is yet. I need your help, thanks!

packages/go/index.ts
```diff
@@ -524,7 +538,7 @@ export async function startDevServer(
   }
   const analyzedRaw = await getAnalyzedEntrypoint(
     workPath,
-    entrypointWithExt,
+    (files[entrypointWithExt] as FileFsRef).fsPath,
```
styfle: I think using `files` here is incorrect. Instead we should keep the call site the same and if we need the absolute path, the `getAnalyzedEntrypoint()` can do `join(workPath, entrypoint)` instead.

packages/go/index.ts
```diff
@@ -536,7 +550,16 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go`
   const analyzed: Analyzed = JSON.parse(analyzedRaw);
 
   await Promise.all([
-    copyEntrypoint(entrypointWithExt, tmpPackage),
+    copyEntrypoint(
+      [
+        relative(
+          goModAbsPathDir,
+          (files[entrypointWithExt] as FileFsRef).fsPath
```
styfle: Same here 

```suggestion
          join(workPath, entrypointWithExt)
```

packages/go/index.ts
```diff
@@ -479,13 +482,24 @@ function isReadable(v: any): v is Readable {
   return v && v.readable === true;
 }
 
-async function copyEntrypoint(entrypoint: string, dest: string): Promise<void> {
-  const data = await readFile(entrypoint, 'utf8');
+async function copyEntrypoint(files: string[], dest: string): Promise<void> {
+  for (const file of files) {
+    if (!existsSync(file)) {
```
styfle: Why would the file not exist?

packages/go/index.ts
await-ovo: @styfle  Hi, I'm using `files` to get the absolute path here by following the usage in the build function. 

The build function calls `getAnalyzedEntrypoint()` with the argument `filePath` which is the absolute path
>https://github.com/vercel/vercel/blob/main/packages/go/index.ts#L147

```
  analyzed = await getAnalyzedEntrypoint(
      workPath,
      downloadedFiles[entrypoint].fsPath,
      goModAbsPathDir
    );
```

Is it better to pass the absolute path when calling the function than to join inside the `getAnalyzedEntrypoint` function?

packages/go/index.ts
await-ovo: Ok, I tried it and it works fine, I will use `path.join` instead of `files`.

packages/go/index.ts
await-ovo: This may have something to do with `analyze.go`, when the directory structure is like this:
```
.
├── README.md
├── api
│   ├── date.go
│   ├── hello.go
│   ├── json.go
│   └── myinfo.go
├── go.mod
└── util
    └── util.go
```
When accessing `/api/json` in a browser, `analyzed.watch` in  the result is like this:

```
[ 'json.go', 'api/hello.go',  'util/util.go']
```

When copying the entry and watch files, I had to ignore `json.go` which does not exist relative to `goModAbsPathDir`.

Since `analyze.go` is used by both build and dev, it works fine when building for production, so I didn't modify `analyze.go` here. Instead, I added a check for the existence of the file when copying.

packages/go/index.ts
styfle: I think it’s best to join inside the func since there are two callers so this will consolidate

packages/go/index.ts
styfle: Hmm that seems like a bug if it returned json.go instead of api/json.go

Perhaps the working directory is wrong?

await-ovo: @styfle  I changed the `analyze.go` code a bit, so that the `entrypoint` path returned in `analyzed.watch` array works.

Also, I changed `/go/test/fixtures/08-include-files/another.go`, so that the `yarn test-integration-once` does not report any errors.

packages/go/index.ts
```diff
@@ -479,13 +480,20 @@ function isReadable(v: any): v is Readable {
...
 
-  // Modify package to `package main`
-  const patched = data.replace(/\bpackage\W+\S+\b/, 'package main');
+      // Modify package to `package main`
+      const patched = data.replace(/\bpackage\W+\S+\b/, 'package main');
 
-  await writeFile(join(dest, 'entrypoint.go'), patched);
+      await outputFile(join(dest, file), patched);
+    } catch (err) {
+      debug(`copy entrypoint ${file} error`, err);
```
styfle, line 493: When would an error happen?

packages/go/index.ts
```diff
@@ -479,13 +480,20 @@ function isReadable(v: any): v is Readable {
   return v && v.readable === true;
 }
 
-async function copyEntrypoint(entrypoint: string, dest: string): Promise<void> {
-  const data = await readFile(entrypoint, 'utf8');
+async function copyEntrypoint(files: string[], dest: string): Promise<void> {
+  for (const file of files) {
```
styfle, line 484: Can we change this function back so the copyEntrypoint() function can be invoked multiple times? I think line 547 can be changed to a `files.map()` so you run all the copy files concurrently.

packages/go/test/fixtures/08-include-files/another.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // Handler function
-func Handler(w http.ResponseWriter, r *http.Request) {
+func AnotherHandler(w http.ResponseWriter, r *http.Request) {
```
styfle, line 10: Let's not change the existing tests. Otherwise we might introduce regressions.

EndangeredMassa:
Closing for now, but we are tracking this internally. We'll come back to this when we next focus on Go improvements.

PR #7620, opened 2022-03-25
[tests] Add `turbo.json` and bump to turbo 1.2.2

TooTallNate:
Ran `npx @turbo/codemod create-turbo-config` like the warnings have been telling us to.

PR #7623, opened 2022-03-25
[tests] Bump `turbo` and allow `turbo.json` in `.vercelignore`

jaredpalmer:
- Fixes builds on vercel by allowing `turbo.json` in the build container
- Upgrades to turbo@1.1.10
- Closes #7620

### DIFF
--- a/.vercelignore
+++ b/.vercelignore
@@ -6,6 +6,7 @@
 !.yarnrc
 !yarn.lock
 !package.json
+!turbo.json
 
 # api
 !api/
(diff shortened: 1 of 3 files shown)


styfle (approved):
Thanks!

.vercelignore
```diff
@@ -6,6 +6,7 @@
 !.yarnrc
 !yarn.lock
 !package.json
+!turbo.json
```
TooTallNate, line 9: We use `turbo run build` in the Vercel deployment. Should we be excluding this file?

TooTallNate (changes requested):
I don't see `turbo.json` being added in this PR, so is https://github.com/vercel/vercel/pull/7620 still necessary or no?

PR #7624, opened 2022-03-25
[cli] Add `--environment` to `vc pull`

EndangeredMassa:
In order to allow local development with preview and production environment variables, this PR adds a `--environment` flag to the `vc pull` command. This is useful for working offline.

The file is placed under `.vercel`. This new location will make it easier to NOT accidentally commit the environment variables. A follow-up PR will make that file be used by `vc build` and another follow-up will make it be used by `vc dev`.

Note: `--environment` is a top-level arg, but it's deprecated there.

---

**Example: Defaults**

```
$ vc pull
```

Output:

- `.env`: contains "development" values
- `.vercel/.env.development.local`: contains "development" values

---

**Example: with --env-file**

```
$ vc pull --env-file=rando.env
```

Output:

- `rando.env`: contains "development" values
- `.vercel/.env.development.local`: contains "development" values

---

**Example: with --environment**

```
$ vc pull --environment=preview
```

Output:

- `.env`: contains "preview" values
- `.vercel/.env.preview.local`: contains "preview" values

---

**Example: with --env-file and --environment**

```
$ vc pull --environment=production --env-file=rando.env
```

Output:

- `rando.env`: contains "production" values
- `.vercel/.env.production.local`: contains "production" values

---

**Example: with --env**

```
$ vc pull --env=production
// boxed
WARN! --env deprecated: please use --env-file instead
// end box
```

---

Card: https://linear.app/vercel/issue/BUI-70/add-vc-pull-environment-for-selecting-target-environment
Previous PR: https://github.com/vercel/vercel/pull/7568


packages/cli/src/commands/pull.ts
```diff
@@ -31,6 +38,7 @@ const help = () => {
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
     --env [filename]               The file to write Development Environment Variables to [.env]
+    --target ${getEnvTargetPlaceholder()}
```
styfle: I'm not sure about `--target`. Should this be `--environment` to match the `vc env add [name] [environment] [gitbranch]` name?

EndangeredMassa: I think that `--environment` will be confusing next to `--env`. I went with `--target` here because that what `deploy` used, but I see now that's deprecated. I'm definitely open to naming ideas for this.

EndangeredMassa: Maybe we can deprecate `--env` in favor of `--env-file` to make it more distinct from `--environment`?

styfle: I think I like the second option better, otherwise we would need to rewrite all the docs and `vc env --help`.

Notice that all the examples use "Environment" 

image


https://vercel.com/docs/cli#commands/env

EndangeredMassa: That makes sense. I updated this PR to deprecate `--env` for `--env-file` and renamed `--target` to `--environment`.

packages/cli/src/commands/pull.ts
```diff
@@ -30,7 +37,8 @@ const help = () => {
     'DIR'
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
-    --env [filename]               The file to write Development Environment Variables to [.env]
+    --env-file [filename]               The file to write Development Environment Variables to [.env]
+    --environment ${getEnvTargetPlaceholder()}
```
styfle: Do we have a default if `--environment` isn't provided? It should be listed in brackets

packages/cli/test/commands/pull.test.ts
```diff
@@ -41,5 +55,78 @@ describe('pull', () => {
     expect(exitCode).toEqual(0);
     expect(actualEnv).toBeTruthy();
     expect(raw.includes('# Created by Vercel CLI')).toBeTruthy();
+
+    // --env flag does not affect nested (under .vercel) files
+    const nestedEnvFileExists = await fs.pathExists(
+      path.join(cwd, '.vercel', '.env.development.local')
+    );
+    expect(nestedEnvFileExists).toBeTruthy();
```
styfle: Should this test compare the contents instead of checking existence? Its not clear what "does not affect" means here.

packages/cli/src/commands/pull.ts
EndangeredMassa: We do! Should the options use brackets too? This is what the options section looks like after adding the default value:

```
-A FILE, --local-config=FILE   Path to the local `vercel.json` file
-Q DIR, --global-config=DIR    Path to the global `.vercel` directory
--env-file [filename]          The file to write Development Environment Variables to [.env]
--environment <production | preview | development> Deployment environment [development]
-y, --yes                      Skip the confirmation prompt
```

The result of `getEnvTargetPlaceholder` is `<production | preview | development>`, but should it be something like `[production | preview | development]` instead? I looked at other help output and don't see another example of an enum option like this. I did see at least one example (certs command) where a placeholder value used angle brackets (`<cn>`).

packages/cli/test/commands/pull.test.ts
EndangeredMassa: I updated the comment to say that the file name is not affected. I also tested the contents to make sure it at least has the env file header line.

packages/cli/src/commands/pull.ts
styfle: Looks like `vc env --help` doesn't list the possible enum options in the initial commands at the top, but it does list them in the examples at the bottom, such as

```
  – Add a new variable for a specific Environment

      $ vercel env add <name> <production | preview | development>
      $ vercel env add DB_PASS production
```

packages/cli/src/commands/pull.ts
```diff
@@ -30,7 +37,8 @@ const help = () => {
     'DIR'
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
-    --env [filename]               The file to write Development Environment Variables to [.env]
+    --env-file [filename]          The file to write Development Environment Variables to [.env]
+    --environment ${getEnvTargetPlaceholder()} Deployment environment [development]
```
styfle:
```suggestion
    --environment [environment] Deployment environment [development]
```

This one can be changed to just a name. Then lets add examples below with the target placeholder expanded like in `vc env --help`

EndangeredMassa: I like it. Now it looks like this:

```
  ▲ vercel pull [path]

 Options:

    -A FILE, --local-config=FILE   Path to the local `vercel.json` file
    -Q DIR, --global-config=DIR    Path to the global `.vercel` directory
    --env-file [filename]          The file to write Development Environment Variables to [.env]
    --environment [environment]    Deployment environment [development]
    -y, --yes                      Skip the confirmation prompt

  Examples:

  – Pull the latest Project Settings from the cloud

    $ vercel pull
    $ vercel pull ./path-to-project
    $ vercel pull --env .env.local
    $ vercel pull ./path-to-project --env .env.local

  – Pull specific environment's Project Settings from the cloud

    $ vercel pull --environment=<production | preview | development>
```

styfle (approved):
Thanks!

packages/cli/src/commands/pull.ts
```diff
@@ -41,32 +49,54 @@ const help = () => {
...
+
+    ${chalk.cyan(
+      `$ ${getPkgName()} pull --environment=${getEnvTargetPlaceholder()}`
+    )}
 `);
 };
-export default async function main(client: Client) {
-  let argv;
+
+// ${}
```
styfle: What is this comment?

EndangeredMassa: No idea. I'll remove it.

styfle (approved):

packages/cli/src/commands/pull.ts
```diff
@@ -30,7 +37,8 @@ const help = () => {
     'DIR'
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
-    --env [filename]               The file to write Development Environment Variables to [.env]
+    --env-file [filename]          The file to write Development Environment Variables to [.env]
```
EndangeredMassa, line 40: The line for `--env-file` need to be removed. It no longer applies. I'll open a new PR for that.

EndangeredMassa: https://github.com/vercel/vercel/pull/7647

PR #7653, opened 2022-04-07
[cli] trim env file path output

EndangeredMassa:
The log lines for what files were written when running `vc pull` were inconsistent. One was a full path, the other was a relative path. This PR makes them both relative.

Before:

```
$  vc pull
✅  Created /Users/nrajlich/nextjs/.vercel/.env.development.local file [103ms]
✅  Downloaded project settings to .vercel/project.json [0ms]
```

After:

```
$  vc pull
Overwriting existing .vercel/.env.development.local file
Downloading Development Environment Variables for Project vercel-demo-nextjs
✅  Updated .vercel/.env.development.local file [1s]
✅  Downloaded project settings to .vercel/project.json [0ms]
```

### DIFF
--- a/packages/cli/src/commands/env/pull.ts
+++ b/packages/cli/src/commands/env/pull.ts
@@ -48,7 +48,8 @@ export default async function pull(
   environment: ProjectEnvTarget,
   opts: Partial<Options>,
   args: string[],
-  output: Output
+  output: Output,
+  cwd: string = process.cwd()
 ) {
   if (args.length > 1) {
     output.error(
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/env/pull.ts
```diff
@@ -48,7 +48,8 @@ export default async function pull(
   environment: ProjectEnvTarget,
   opts: Partial<Options>,
   args: string[],
-  output: Output
+  output: Output,
+  cwd: string = process.cwd()
```
styfle, line 52:
```suggestion
  cwd: string,
```
This shouldn't be optional. We should provide the `--cwd` global flag here, right?

EndangeredMassa: The global `--cwd` flag sets the process's current directory before the command functions are invoked. https://github.com/vercel/vercel/blob/e40eecafc97dbe120f4c95f30b7b9aaf621e4b81/packages/cli/src/index.ts#L139-L142

I needed an override here for tests only.

styfle (approved):

TooTallNate (approved):

PR #7656, opened 2022-04-08
[cli] set environment properly in `vc pull`

EndangeredMassa:
This PR sets environment properly in `vc pull`. The system env vars were not respecting the target `--environment`.

Adds tests for related `vc env` logic.


packages/cli/src/commands/env/pull.ts
```diff
@@ -98,7 +99,8 @@ export default async function pull(
   const records = exposeSystemEnvs(
     projectEnvs,
     systemEnvValues,
-    project.autoExposeSystemEnvs
+    project.autoExposeSystemEnvs,
+    environment
```
EndangeredMassa, line 104: This is the fix for the `--environment` not being respected

packages/cli/src/commands/pull.ts
```diff
@@ -69,19 +68,14 @@ function processArgs(client: Client) {
...
-    return argv;
-  } catch (err) {
-    handleError(err);
-    return 1;
+  if (argv['--help']) {
+    help();
+    return 2;
   }
+
+  return argv;
```
EndangeredMassa, line 78: This try-catch was changing how errors were handled. We should let errors bubble up to the CLI entry point where they are handled consistently.

packages/cli/test/mocks/client.ts
```diff
@@ -109,6 +122,6 @@ beforeEach(() => {
   client.reset();
 });
 
-afterAll(() => {
-  client.stopMockServer();
+afterAll(async () => {
+  await client.stopMockServer();
```
EndangeredMassa, line 126: This works as written before, but the mock server was still listening after the test suite completed for some amount of time. When I was using `yarn test --detectOpenHandles`, this mock server listener was listed as an open handle when it's not realllly going to stay open for much longer after the test runner completes. It caused me to spend time digging into this when my real problem was something else.

This change makes the test suite wait for the mock server to be shut down.

packages/cli/test/mocks/team.ts
```diff
@@ -14,7 +15,7 @@ export function useTeams() {
   ];
 
   for (let team of teams) {
-    client.scenario.get(`/v1/team/${team.id}`, (_req, res) => {
+    client.scenario.get(`/teams/${team.id}`, (_req, res) => {
```
EndangeredMassa, line 18: The previous path `/v1/team` doesn't exist. I think this was written, but never invoked. The new path `/teams/:id` is invoked by the new tests.

styfle: I'll defer to @TooTallNate here

packages/cli/src/commands/env/pull.ts
```diff
@@ -48,7 +48,8 @@ export default async function pull(
   environment: ProjectEnvTarget,
   opts: Partial<Options>,
   args: string[],
-  output: Output
+  output: Output,
+  cwd: string = process.cwd()
```
styfle:
```suggestion
  cwd: string,
```

This can be a required param now, right?

EndangeredMassa: Good call, yeah

packages/cli/src/commands/env/pull.ts
```diff
@@ -81,7 +81,7 @@ export default async function pull(
   }
 
   output.print(
-    `Downloading Development Environment Variables for Project ${chalk.bold(
+    `Downloading ${environment} Environment Variables for Project ${chalk.bold(
```
TooTallNate: Note that this will be lowercase target (i.e. "development" rather then "Development") since it's the enum value. I think we either need to capitalize to have the same format or add quotations around the enum value.

packages/cli/test/mocks/team.ts
TooTallNate: Ya, confirmed.

packages/cli/test/integration.js
```diff
@@ -974,11 +977,12 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
-    t.true(lines.has('VERCEL_URL=""'));
-    t.true(lines.has('VERCEL_ENV="development"'));
-    t.true(lines.has('VERCEL_GIT_PROVIDER=""'));
-    t.true(lines.has('VERCEL_GIT_REPO_SLUG=""'));
+
+    t.true(lines.has('VERCEL="1"'), 'VERCEL');
+    t.true(lines.has('VERCEL_URL=""'), 'VERCEL_URL');
+    t.true(lines.has('VERCEL_ENV="development"'), 'VERCEL_ENV');
+    t.true(lines.has('VERCEL_GIT_PROVIDER=""'), 'VERCEL_GIT_PROVIDER');
+    t.true(lines.has('VERCEL_GIT_REPO_SLUG=""'), 'VERCEL_GIT_REPO_SLUG');
```
EndangeredMassa, line 985: All of these changes were made to make the assertion failure messages clearer.

packages/cli/package.json
```diff
@@ -51,6 +51,7 @@
     "update-notifier": "4.1.0"
   },
   "devDependencies": {
+    "@alex_neo/jest-expect-message": "1.0.5",
```
EndangeredMassa, line 54: This allows for a custom message to appear when a test fails. You can specify the custom message as the second argument to `expect`:

```
expect(exitCode, client.outputBuffer).toEqual(0);
```

packages/cli/test/commands/env.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      useUser();
+      useTeams('team_dummy');
+      useProject({
+        ...defaultProject,
+        id: 'vercel-env-pull',
+        name: 'vercel-env-pull',
+      });
+      client.setArgv('env', 'pull', '--yes', '--cwd', cwd);
+      const exitCode = await env(client);
+      expect(exitCode, client.outputBuffer).toEqual(0);
```
EndangeredMassa, line 23: I added `client.outputBuffer` as a "Custom Message" to failures on these so we'd see what the buffer contained when they fail. Here's an example:

```
$  yarn test test/commands/env.test.ts test/commands/pull.test.ts
yarn run v1.22.15
$ jest test/commands/env.test.ts test/commands/pull.test.ts
 PASS  test/commands/pull.test.ts
 FAIL  test/commands/env.test.ts
  ● env › pull › should handle pulling

    Custom message:
      Downloading "development" Environment Variables for Project vercel-env-pull
    Created .env file [6ms]


    expect(received).toEqual(expected) // deep equality

    Expected: 999
    Received: 0

      21 |       client.setArgv('env', 'pull', '--yes', '--cwd', cwd);
      22 |       const exitCode = await env(client);
    > 23 |       expect(exitCode, client.outputBuffer).toEqual(999);
         |                                             ^
      24 |
      25 |       const rawDevEnv = await fs.readFile(path.join(cwd, '.env'));
      26 |

      at Object.<anonymous> (test/commands/env.test.ts:23:45)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 5 passed, 6 total
Snapshots:   0 total
Time:        1.261 s, estimated 2 s
Ran all test suites matching /test\/commands\/env.test.ts|test\/commands\/pull.test.ts/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/cli/test/integration.js
```diff
@@ -45,6 +45,12 @@ function fixture(name) {
   return directory;
 }
 
+const URL_REGEX = /(https?:\/\/[^ ]*)/;
+function extractUrl(input) {
+  let match = input.match(URL_REGEX);
+  return match?.[1];
+}
```
EndangeredMassa: This allows the tests that rely on stdout containing a URL to still pass when you are trying to add debugging logs for other tests. When these tests fail because of those extra logs, it's essentially a false negative that wastes time.

If we want tests to fail when we introduce superfluous output in general, we should tackle that as an explicit effort.

EndangeredMassa: I'll remove these and we can look at this more later.

packages/cli/test/commands/pull.test.ts
```diff
@@ -19,60 +19,12 @@ describe('pull', () => {
     });
     client.setArgv('pull', '--yes', cwd);
     const exitCode = await pull(client);
-    expect(exitCode).toEqual(0);
+    expect(exitCode, client.outputBuffer).toEqual(0);
```
styfle, line 22: Can you pass 2 params to expect() ? 🤔

styfle: Oh I see your other comment now 😄

packages/cli/package.json
```diff
@@ -176,10 +178,13 @@
...
       }
     },
+    "setupFilesAfterEnv": [
+      "@alex_neo/jest-expect-message"
+    ],
     "verbose": false,
     "testEnvironment": "node",
     "testMatch": [
-      "<rootDir>/test/**/*.test.ts"
+      "<rootDir>/test/commands/pull.test.ts"
```
styfle: This seems like its going to skip tests

packages/cli/package.json
styfle: Does jest not have this built in? Could we use a matcher instead? https://stackoverflow.com/a/45375363/266535

packages/cli/package.json
EndangeredMassa: Oh, that was for debugging. Thanks for catching that. I need to remove it.

packages/cli/src/commands/env/pull.ts
```diff
@@ -81,7 +81,7 @@ export default async function pull(
   }
 
   output.print(
-    `Downloading Development Environment Variables for Project ${chalk.bold(
+    `Downloading "${environment}" Environment Variables for Project ${chalk.bold(
```
styfle, line 84: Do we need to use `title(environment)` here so the first letter becomes uppercase?

packages/cli/test/commands/pull.test.ts
EndangeredMassa: Only with: https://github.com/vercel/vercel/pull/7656#discussion_r848607417

packages/cli/test/fixtures/unit/vercel-env-pull/vercel.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 2,
+  "name": "vercel-env-next",
+  "builds": [{ "src": "package.json", "use": "@vercel/next@canary" }],
```
styfle: Can `builds` be removed here?

packages/cli/test/integration.js
styfle: Was it not showing the line number before?

packages/cli/package.json
EndangeredMassa: It looks like an older version of jest allowed exactly this, but now we need a plugin.

We could make custom matches, but (1) the syntax isn't as clear and (2) we'd have to make one per type of match we'd want to perform. I think we may only need one type of match, but if we did need more, the plugin in this PR already accounts for all built-in matches.

We could use a custom matcher like:

```
expect.extend({
  toBeEqualWithMessage(received, expected, message) {
    return {
      message: () => `Expected ${received} to equal ${expected}. Context: ${message}`,
      pass: received === expected
    };
  }
});

// ...

expect(3).toBeEqualWithMessage(4, 'player count');
```

packages/cli/src/commands/env/pull.ts
EndangeredMassa: We could, but I like that the environment is quoted to make it more obvious that it could have been something else.

packages/cli/test/mocks/client.ts
styfle: Nice find!

packages/cli/package.json
styfle: In the future, you can use `jest -t 'some test name'` or more likely `yarn test-unit -t 'some test name'` to run a specific test instead of changing config

EndangeredMassa:
I figured out the weird test failure issue. I had a local `.vercel/project.json` file in my fixture that was gitignored. So, my tests would pass locally, but not in CI.

packages/cli/package.json
EndangeredMassa: I was doing that locally, but I wanted to run only this test file in CI to see if there was some test pollution

packages/cli/test/fixtures/unit/vercel-env-pull/.gitignore
```diff
@@ -0,0 +1,3 @@
+.next
+yarn.lock
+.vercel
```
styfle: Maybe this should be more specific since you intended to include `.vercel/project.json`?

styfle (approved):

packages/cli/test/fixtures/unit/vercel-env-pull/.gitignore
EndangeredMassa: I removed the line. The updated PR matches the intent better.

TooTallNate (approved):

PR #7658, opened 2022-04-08
[build-utils] Add `repoRootPath` to `PrepareCacheOptions`

TooTallNate:
`repoRootPath` is provided to the `prepareCache()` function and is the same value as provided to the `build()` function.

Also remove `cachePath` since it's no longer used.


styfle (approved):
Thanks!

AndyBitz (approved):

Ethan-Arrowood (approved):
Sweet!

PR #7668, opened 2022-04-11
[docs] Fix links to docs

styfle:
These links were all outdated


errors/missing-env-file.md
```diff
@@ -2,7 +2,7 @@
 
 #### Why This Error Occurred
 
-You ran `vercel dev` inside a project that contains a `vercel.json` file with `env` or `build.env` properties that use [Vercel Secrets](https://vercel.com/docs/v2/build-step#environment-variables).
+You ran `vercel dev` inside a project that contains a `vercel.json` file with `env` or `build.env` properties that use [Vercel Secrets](https://vercel.com/docs/concepts/projects/environment-variables).
```
EndangeredMassa, line 5: The URL https://vercel.com/docs/v2/build-step#environment-variables redirects to https://vercel.com/docs/concepts/deployments/build-step#environment-variables not https://vercel.com/docs/concepts/projects/environment-variables. It looks like the `concepts/projects/environment-variables` is correct. Should the redirect be updated? Same for the link below in this file.

Should the link still be labeled with "Vercel Secrets"? The word "secret" doesn't seem to be how we refer to these throughout.

EndangeredMassa (approved):

errors/missing-env-file.md
styfle: it looks like someone created a tiny section at https://vercel.com/docs/concepts/deployments/build-step#environment-variables which is simply a link to https://vercel.com/docs/concepts/projects/environment-variables

Seems like it could have been a redirect instead.

styfle: Yeah so this is confusing because we don't have any docs on Vercel Secrets anymore. Not sure what to link to here. Maybe it should just say to use Environment Variables instead? We never officially deprecated Secrets 🤔

EndangeredMassa: I would change it to "Environment Variables", but happy to defer to whomever can make these decisions.

AndyBitz (approved):

errors/missing-env-file.md
styfle: Looks like this error really is for secrets, not all env vars. You can see it checks for “@“ here 

https://github.com/vercel/vercel/blob/da893e7c573ec35308e542287108a6a80f9896a6/packages/cli/src/util/dev/server.ts#L763

That makes me think we need a new section in the docs for secrets, or a proper deprecation.

TooTallNate (approved):

PR #7669, opened 2022-04-11
[static-build] Add support for Build Output v3 detection

TooTallNate:
Adds detection logic for a framework / build script outputting Build Output v3
format to the filesystem. In this case, `static-build` will simply stop processing
after the Build Command since deserialization happens in the build-container side
of things (this is different compared to the v1 output which gets handled in this
Builder. The reason for that is because the v3 output matches what `vc build`
outputs vs. v1 which is a different format).


packages/build-utils/src/types.ts
```diff
@@ -356,19 +356,42 @@ export interface Images {
...
+/**
+ * If a Builder ends up creating filesystem outputs conforming to
+ * the Build Output API, then the Builder should return this type.
+ */
+export interface BuildResultBuildOutput {
+  /**
+   * Version number of the Build Output API that was created.
+   * @example 3
+   */
+  buildOutputVersion: number;
```
EndangeredMassa: Should we make this a literal type that currently only accepts `3`?

packages/static-build/test/build.test.ts
```diff
@@ -0,0 +1,54 @@
+import path from 'path';
```
EndangeredMassa, line 1: This file name is confusing next to `builds.test.ts`: https://github.com/vercel/vercel/tree/update/static-build-build-output-v3/packages/static-build/test

Are these the best names for these files?

packages/static-build/package.json
```diff
@@ -14,13 +14,24 @@
   },
   "scripts": {
     "build": "node build",
-    "test-unit": "jest --env node --verbose --runInBand --bail test/unit.test.js",
+    "test-unit": "jest --env node --verbose --bail test/build.test.ts test/prepare-cache.test.ts",
```
EndangeredMassa, line 17: Should we make `test-unit` grab all tests under a `unit` directory instead? I don't want to forget to add a new file to this list and wonder why my test wasn't being run.

packages/static-build/src/index.ts
```diff
@@ -637,6 +632,30 @@ export const build: BuildV2 = async ({
...
+
+      if (buildOutputPath) {
+        // Ensure that `vercel build` is being used for this Deployment
+        if (!meta.cliVersion) {
+          let buildCommandName: string;
+          if (buildCommand) buildCommandName = `"${buildCommand}"`;
+          else if (framework) buildCommandName = framework.name;
+          else buildCommandName = 'the "build" script';
+          throw new Error(
+            `Detected Build Output v3 from ${buildCommandName}, but this Deployment is not using \`vercel build\`.\nPlease set the \`ENABLE_VC_BUILD=1\` environment variable. More info: ____________`
```
EndangeredMassa: Will we link this to a card somewhere so we don't forget to update it? I think we should remove the `More info` part until we have that link ready.

EndangeredMassa (approved):

packages/static-build/test/build.test.ts
TooTallNate: `builds.test.js` isn't even used apparently. It looks like it's unit tests though so probably should be merged into this new file. I'd prefer to do it in another PR.

packages/static-build/package.json
TooTallNate: Perhaps, but it won't work currently cause of `integration.test.js` which shouldn't be run by `test-unit`.

EndangeredMassa (approved):

packages/static-build/package.json
styfle: How about `*.unit.test.ts` instead?

packages/static-build/src/utils/build-output-v3.ts
```diff
@@ -0,0 +1,23 @@
+import { join } from 'path';
+import { promises as fs } from 'fs';
+
+export const BUILD_OUTPUT_DIR = '.vercel/output';
```
styfle: Is this export used?

packages/static-build/src/utils/build-output-v3.ts
```diff
@@ -0,0 +1,23 @@
...
+ * `config.json` file was created by the framework / build script,
+ * or `undefined` if the framework did not create the v3 output.
+ */
+export async function getBuildOutputDirectory(
+  path: string
+): Promise<string | undefined> {
+  try {
+    const outputDir = join(path, BUILD_OUTPUT_DIR);
+    const configPath = join(outputDir, 'config.json');
+    await fs.stat(configPath);
```
styfle, line 17: Do we need to check that stat returns `isDirectory()` true?

packages/static-build/test/build-fixtures/09-build-output-v3/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
styfle: Is this lock file necessary?

packages/static-build/src/utils/build-output-v3.ts
TooTallNate: It's stat'ing a file (`config.json`). The reason it doesn't check for the directory itself is because in `vc build` the skeleton of the directory has already been created before the Builders are run, but the `config.json` doesn't get created until the end, so it's a better thing to check.

packages/static-build/test/build-fixtures/09-build-output-v3/yarn.lock
TooTallNate: No.

packages/static-build/src/utils/build-output-v3.ts
TooTallNate: No, I'll remove the `export`.

PR #7670, opened 2022-04-11
[node][redwood] update `@vercel/nft` to 0.18.1

Ethan-Arrowood:
### Related Issues

Updates the @vercel/nft dependency across the repo to v0.18.1.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -92,7 +92,6 @@
     "@vercel/fetch-retry": "5.0.3",
     "@vercel/frameworks": "0.7.1",
     "@vercel/ncc": "0.24.0",
-    "@vercel/nft": "0.18.1",
     "@zeit/fun": "0.11.2",
     "@zeit/source-map-support": "0.6.2",
     "ajv": "6.12.2",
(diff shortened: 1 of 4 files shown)


TooTallNate (approved):

packages/cli/package.json
```diff
@@ -92,7 +92,7 @@
     "@vercel/fetch-retry": "5.0.3",
     "@vercel/frameworks": "0.7.1",
     "@vercel/ncc": "0.24.0",
-    "@vercel/nft": "0.17.5",
+    "@vercel/nft": "0.18.1",
```
TooTallNate: We're not actually using nft in CLI itself (I think this was leftover from the v2 vc build).

```suggestion
```

styfle (approved):

packages/cli/package.json
Ethan-Arrowood: Should the lock file change as well? I tried running `yarn install` from `vercel/packages/cli` with this change and its not changing the lock file

TooTallNate: I wouldn't expect it to change since nft is still being referenced by these other packages.

PR #7671, opened 2022-04-12
[build-utils] Only allow `runNpmInstall()` to run once per `package.json`

TooTallNate:
Adds a best-effort optimization to only run `npm install` once per `pacakge.json` file. This will save a lot of time in the single-sandbox build world (i.e. `vc build`).


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -319,55 +329,84 @@ export async function walkParentDirs({
   return null;
 }
 
+function isSet<T>(v: any): v is Set<T> {
+  return v?.constructor.name === 'Set';
```
styfle:
```suggestion
  return v?.constructor?.name === 'Set';
```

packages/build-utils/test/unit.test.ts
```diff
@@ -413,3 +414,22 @@ it('should detect pnpm Workspaces', async () => {
...
+
+it('should only invoke `runNpmInstall()` once per `package.json` file', async () => {
+  const meta: Meta = {};
+  const fixture = path.join(__dirname, 'fixtures', '02-zero-config-api');
+  const apiDir = path.join(fixture, 'api');
+  const [run1, run2, run3] = await Promise.all([
+    runNpmInstall(apiDir, [], undefined, meta),
+    runNpmInstall(apiDir, [], undefined, meta),
+    runNpmInstall(fixture, [], undefined, meta),
+  ]);
```
styfle, line 443: Does it work if invoked sequentially (as opposed to concurrently)?

I think we need a test with sequential access too.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -319,55 +329,84 @@ export async function walkParentDirs({
...
+      destPath
+    );
+
+    // Only allow `runNpmInstall()` to run once per `package.json`
+    // when doing a default install (no additional args)
+    if (meta && packageJsonPath && args.length === 0) {
+      if (typeof meta.runNpmInstallSet === 'undefined') {
+        meta.runNpmInstallSet = new Set<string>();
+      }
+      if (isSet<string>(meta.runNpmInstallSet)) {
```
EndangeredMassa, line 362: When would `meta.runNpmInstallSet` not be a `Set<string>`? In that case, should we log a warning or something?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -319,55 +329,84 @@ export async function walkParentDirs({
...
+      commandArgs = ['install', ...args];
+    }
 
-  return spawnAsync(cliType, commandArgs, opts);
+    if (process.env.NPM_ONLY_PRODUCTION) {
+      commandArgs.push('--production');
+    }
+
+    await spawnAsync(cliType, commandArgs, opts);
+    return true;
```
EndangeredMassa, line 406: I didn't realize that `finally` blocks would still run after `return` statements are executed. Huh.

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
styfle: I don't think it can happen since the user can't set `meta`.

This whole block could probably be simplified to:


```ts
if (meta && packageJsonPath && args.length === 0) {
  if (!isSet<string>(meta.runNpmInstallSet)) {
    meta.runNpmInstallSet = new Set<string>();
  }

  if (meta.runNpmInstallSet.has(packageJsonPath)) {
    return false;
  }

  meta.runNpmInstallSet.add(packageJsonPath);
}
```

TooTallNate: It would not be set the first time that `runNpmInstall()` gets run, since it gets initialized here.

Unfortunately the `isSet()` if check is still necessary. I tried with latest version of TypeScript too. Bad TypeScript!

packages/build-utils/test/unit.test.ts
TooTallNate: Yes. Added a test.

styfle (approved):

PR #7672, opened 2022-04-12
[build-utils] Move "Installing dependencies..." log to `runNpmInstall()`

TooTallNate:
Follow-up to #7671. Since `runNpmInstall()` might now be de-duped, only print "Installing dependencies..." when the dependencies are actually being installed. This avoids printing the log message unnecessarily when the command won't actually be run, and also removes some duplication in the Builders' code.


styfle (approved):
Might need to change in `@vercel/next` too

EndangeredMassa (approved):

PR #7677, opened 2022-04-14
use require().default so ts-eager understands them

### LINKED ISSUE #7676: cli: `npx ts-eager src` no longer working
Scenario: Trying to work on `packages/cli` locally.

The README of `packages/cli` mentions it's possible to run the `cli` using `npx ts-eager src` here https://github.com/vercel/vercel/tree/main/packages/cli#local-development.

This seems to have stopped working. Depending on which command I run, I get a slightly different error.

```
$ npx ts-eager src whoami
Vercel CLI 24.1.1-canary.1 — https://vercel.com/feedback
Error! Cannot find module '/x/vercel/vercel/packages/cli/src/commands/whoami' imported from /x/vercel/vercel/packages/cli/src/index.ts

$ npx ts-eager src init  
Vercel CLI 24.1.1-canary.1 — https://vercel.com/feedback
Error! Directory import '/x/vercel/vercel/packages/cli/src/commands/init' is not supported resolving ES modules imported from /x/vercel/vercel/packages/cli/src/index.ts
```

I guess the different error is due to the respective command either being a directory (`packages/cli/src/commands/init`) or a file (`packages/cli/src/commands/whoami.ts`).

dferber90:
Switches `await import()` to `require().default` so that ts-eager understands them.

### Related Issues

closes #7676

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -585,73 +585,73 @@ const main = async () => {
     let func: any;
     switch (targetCommand) {
       case 'alias':
-        func = await import('./commands/alias');
+        func = require('./commands/alias').default;
         break;
       case 'billing':
-        func = await import('./commands/billing');
+        func = require('./commands/billing').default;
         break;
       case 'bisect':
-        func = await import('./commands/bisect');
+        func = require('./commands/bisect').default;
         break;
       case 'certs':
-        func = await import('./commands/certs');
+        func = require('./commands/certs').default;
         break;
       case 'deploy':
-        func = await import('./commands/deploy');
+        func = require('./commands/deploy').default;
         break;
       case 'dev':
-        func = await import('./commands/dev');
+        func = require('./commands/dev').default;
         break;
       case 'dns':
-        func = await import('./commands/dns');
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -585,73 +585,73 @@ const main = async () => {
...
+        func = require('./commands/teams').default;
         break;
       case 'update':
-        func = await import('./commands/update');
+        func = require('./commands/update').default;
         break;
       case 'whoami':
-        func = await import('./commands/whoami');
+        func = require('./commands/whoami').default;
         break;
```
dferber90, line 655: There is another `await import()` in https://github.com/vercel/vercel/blob/1df65eaac1fbe132b82de8e613aeed82b748f4ef/packages/node-bridge/launcher.js#L69-L71 but I'm not sure if that needs changing as it has a `useRequire` check.

TooTallNate: That one should stay as-is. Note that it's a `.js` file so doesn't get compiled by TypeScript.

TooTallNate:
Thanks!

PR #7686, opened 2022-04-15
[examples] Remove "amp" example

EndangeredMassa:
The `amp` example does not deploy properly, `amp` itself is not a support framework, and @leerob said we can remove this.


TooTallNate (approved):

styfle (approved):
Seems fine since I don’t see “amp” listed in frameworks.ts 

It’s just strange that it was ever added as an example. 

@AndyBitz do you remember the history here?

AndyBitz:
> Seems fine since I don’t see “amp” listed in frameworks.ts
> 
> It’s just strange that it was ever added as an example.
> 
> @AndyBitz do you remember the history here?

Can't recall anything about it

AndyBitz (approved):

PR #7690, opened 2022-04-18
[static-build] Support subset of Build Output API v2

EndangeredMassa:
We want to handle Build Output API v2 directories similar to how we handle v1. This PR adds another code branch where we parse some of the results of a v2 directory and return the appropriate build outputs.

This is not a complete v2 implementation. It's just enough to make sure some frameworks using v2 are working properly. We can expand this as necessary.


packages/static-build/src/utils/build-output-v2.ts
```diff
@@ -0,0 +1,125 @@
+import path from 'path';
+import { pathExists } from 'fs-extra';
+import { promises as fs } from 'fs';
```
TooTallNate:
```suggestion
import { stat, pathExists } from 'fs-extra';
```

Just use `stat()` from fs-extra.

packages/static-build/src/utils/build-output-v2.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+  const middlewareAbsolutePath = path.join(workPath, middlewareRelativePath);
+
+  const fileExists = await fs.stat(middlewareAbsolutePath);
+  if (!fileExists) {
+    return;
+  }
+
+  const route = {
+    src: '/api/(.*)',
```
TooTallNate:
```suggestion
    src: '/(.*)',
```

packages/static-build/test/build-fixtures/09-build-output-v3/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
styfle: Is this lock file necessary?

packages/static-build/test/build-fixtures/10-build-output-v2/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
styfle: Is this lock file necessary?

packages/static-build/test/build-fixtures/09-build-output-v3/yarn.lock
EndangeredMassa: They were generated when running tests. We can either keep them or .gitignore them. They are not necessary, no. Maybe gitignore is best, then.

EndangeredMassa:
Switched target branch to main

examples/solidstart/.gitignore
```diff
@@ -2,6 +2,7 @@ dist
 worker
 .solid
 .vercel
+.output
```
styfle, line 5: Why does the example need to ignore .output

Seems like we should use `.vercel/output` instead which is already covered by the ignore of `.vercel`

packages/cli/test/integration.js
```diff
@@ -1090,29 +1090,47 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
-  await vcDevAndFetchCloudVars();
-  await enableAutoExposeSystemEnvs();
-  await vcEnvPullFetchSystemVars();
-  fs.unlinkSync(path.join(target, '.env'));
-  await vcDevAndFetchSystemVars();
-  await vcEnvRemove();
-  await vcEnvRemoveWithArgs();
-  await vcEnvRemoveWithNameOnly();
-  await vcEnvLsIsEmpty();
+  function vcEnvRemoveByName(name) {
```
styfle: I don't see `vc env` command changing. Why is the test changing?

packages/static-build/test/build.test.ts
```diff
@@ -1,54 +1,88 @@
 import path from 'path';
+import { remove } from 'fs-extra';
 import { build } from '../src';
 
 describe('build()', () => {
-  it('should detect Builder Output v3', async () => {
```
styfle, line 5: What happened to `should detect Builder Output v3`?

TooTallNate: It's still there, moved down below.

examples/solidstart/.gitignore
TooTallNate: SolidStart actually outputs FS API v2. That's the main purpose of this PR.

packages/static-build/src/index.ts
```diff
@@ -647,23 +648,27 @@ export const build: BuildV2 = async ({
         };
       }
 
-      if (framework) {
```
TooTallNate, line 650: It seems like this code is being deleted?

packages/static-build/src/utils/build-output-v2.ts
```diff
@@ -0,0 +1,124 @@
...
+    debug(`Detected Routes Configuration in "${BUILD_OUTPUT_DIR}"`);
+  }
+
+  return outputs;
+}
+
+async function getMiddleware(
+  workPath: string
+): Promise<{ route: Route; file: FileFsRef } | undefined> {
+  const middlewareRelativePath = path.join(
```
TooTallNate: Instead of assuming this file, let's read the `.output/functions-manifest.json` file as the source of truth (middleware has `runtime: 'web'`).

packages/static-build/src/index.ts
EndangeredMassa: Huh, yeah it was. I restored it in the extracted function `getUpdatedDistPath`.

TooTallNate (approved):

PR #7696, opened 2022-04-20
[cli] check isTTY before prompting to `pull`

EndangeredMassa:
In a CI or other non-TTY setting, the `build` command should not prompt for executing `pull`. This causes the process to hang forever.


EndangeredMassa:
@TooTallNate should we pair soon on getting some tests around this command?

TooTallNate (approved):

PR #7697, opened 2022-04-20
[cli] check for prebuilt directory

EndangeredMassa:
When using `vc deploy --prebuilt`, we need to check for an actual ".vercel/output" before attempting to deploy. This PR adds a a check for that. Now...

In a directory without `.vercel/output`:

```
$ vc deploy --prebuilt
Error! Option `--prebuilt` was used, but no prebuilt deploy found in ".vercel/output"
```

In a direcotry with `.vercel/output` (where I've not linked it yet):

```
$ vc deploy --prebuilt
? Set up and deploy “~/source/vercel/examples/build-output-api/serverless-function”? [Y/n]
```

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -181,6 +181,17 @@ export default async (client: Client) => {
     );
   }
 
+  // build `--prebuilt`
+  if (argv['--prebuilt']) {
+    const prebuiltExists = await fs.pathExists(join(path, '.vercel/output'));
+    if (!prebuiltExists) {
+      error(
+        'Option `--prebuilt` was used, but no prebuilt deploy found in ".vercel/output"'
+      );
+      return 1;
+    }
+  }
+
   // retrieve `project` and `org` from .vercel
   const link = await getLinkedProject(client, path);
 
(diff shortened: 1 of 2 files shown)


TooTallNate (changes requested):
Let's tweak the message to inform the user that they need to run `vercel build`.

TooTallNate (approved):

TooTallNate (approved):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -181,6 +181,21 @@ export default async (client: Client) => {
...
   }
 
+  // build `--prebuilt`
+  if (argv['--prebuilt']) {
+    const prebuiltExists = await fs.pathExists(join(path, '.vercel/output'));
+    if (!prebuiltExists) {
+      error(
+        `The ${param(
+          '--prebuilt'
+        )} option was used, but no prebuilt deploy found in ".vercel/output". Run ${getCommandName(
```
styfle, line 191: "no prebuilt deploy found" should probably be "no prebuilt output found"

EndangeredMassa: Agreed! https://github.com/vercel/vercel/pull/7702

PR #7699, opened 2022-04-20
[client] Remove single file deployment root path `/` route

TooTallNate:
This is strange behavior, and also inconsistent compared to how a Git deployment is created. Let's remove it.


styfle (approved):
I believe `vc file.txt` has been deprecated for a couple years so it seems safe to remove this behavior too.

PR #7700, opened 2022-04-21
[cli] require non-TTY uses of `pull` to also pass `--yes`

EndangeredMassa:
Checks for `TTY` input to decide to handle the command or not. If the input is not TTY and the user did not pass the `--yes` option, the CLi will exit with the message:

> Command `vercel pull` requires confirmation. Use option "--yes" to confirm.


TooTallNate (approved):

packages/cli/src/commands/pull.ts
```diff
@@ -146,6 +148,16 @@ export default async function main(client: Client) {
   const yes = Boolean(argv['--yes']);
   const environment = parseEnvironment(argv['--environment'] || undefined);
 
+  const isTTY = process.stdin.isTTY;
+  if (!isTTY && !yes) {
+    client.output.print(
```
TooTallNate:
```suggestion
    client.output.error(
```

TooTallNate (approved):

packages/cli/test/commands/pull.test.ts
```diff
@@ -46,7 +46,7 @@ describe('pull', () => {
         id: 'vercel-pull-next',
         name: 'vercel-pull-next',
       });
-      client.setArgv('pull', cwd);
+      client.setArgv('pull', cwd, '--yes');
```
TooTallNate: So technically this is a breaking change - when the env vars are used (cause we won't prompt to link the project). I'm rethinking that we should move this isTTY check to right before the linking prompt happens so that `vc pull` + env vars still doesn't need `--yes`.

TooTallNate (changes requested):

packages/cli/test/commands/pull.test.ts
EndangeredMassa: @TooTallNate It is already right before the linking prompt, isn't it? https://github.com/vercel/vercel/pull/7700/files#diff-9559f83bc71400a51427651e1011d2c2acf33ec1fba9f62caa7c7cfa6e90eeaaR161

TooTallNate: This test being altered (with the env vars being set) needing to be changed indicates that the error message is happening too soon in the `pull` command. We shouldn't have to specify `--yes` if the env vars have been set since conceptually the project is already "linked" when the env vars are set.

EndangeredMassa:
@TooTallNate I moved the check to right before the confirm prompt would show up. I think this makes the most sense.

I noticed that `--yes` wasn't necessary in the other tests because they already had projects set up. I also added some tests for this behavior.

EndangeredMassa:
@TooTallNate This appears to work, but the build still fails because of the connection reset issue.

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -68,6 +70,15 @@ export default async function setupAndLink(
     remove(vercelDir);
   }
 
+  if (!isTTY && !autoConfirm) {
+    output.error(
+      `Command ${getCommandName(
+        'pull'
```
styfle: I don't think you can hard code `vc pull` here because this file is also used for `vc link` and I think `vc dev`

EndangeredMassa: Ah thanks! Will fix.

EndangeredMassa: I moved the responsibility for rendering this error message out to the command that calls `setupAndLink`. This makes it clearer how to change the error message per command (and recommend the correct command argument to `autoConfirm`, which is sometimes `--yes` and sometimes `--confirm`).

This is a good separation for this specific case, but I did not update all other cases to work this way. That something we could review and change as a whole. For now, I did export all relevant `reason` codes on error-like responses as a step towards this approach.

If there's a better way to solve this, let me know.

packages/cli/test/integration.js
```diff
@@ -794,15 +794,10 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
-    t.true(lines.has('MY_STDIN_VAR="{"expect":"quotes"}"'), 'MY_STDIN_VAR');
-    t.true(
-      lines.has('MY_DECRYPTABLE_SECRET_ENV="decryptable value"'),
-      'MY_DECRYPTABLE_SECRET_ENV'
-    );
-    t.false(lines.has('MY_PREVIEW'), 'MY_PREVIEW');
+    t.regex(contents, /MY_NEW_ENV_VAR="my plaintext value"/);
+    t.regex(contents, /MY_STDIN_VAR="{"expect":"quotes"}"/);
+    t.regex(contents, /MY_DECRYPTABLE_SECRET_ENV="decryptable value"/);
+    t.notRegex(contents, /MY_PREVIEW/);
```
EndangeredMassa, line 800: these provide better failure messages

packages/cli/test/integration.js
```diff
@@ -1087,29 +1082,48 @@ test('Deploy `api-env` fixture and test `vercel env` command', async t => {
...
+    await vcEnvPullFetchSystemVars();
+    fs.unlinkSync(path.join(target, '.env'));
+    await vcDevAndFetchSystemVars();
+    await vcEnvRemove();
+    await vcEnvRemoveWithArgs();
+    await vcEnvRemoveWithNameOnly();
+    await vcEnvLsIsEmpty();
+  } finally {
+    await vcEnvRemoveAll();
+  }
```
EndangeredMassa, line 1126: This test was failing, but failures polluted the env variables in a way that caused future test suite runs to fail. This change makes sure the env variables are always deleted after the test runs (failure or not).

packages/cli/test/commands/pull.test.ts
```diff
@@ -17,7 +17,7 @@ describe('pull', () => {
       id: 'vercel-pull-next',
       name: 'vercel-pull-next',
     });
-    client.setArgv('pull', '--yes', cwd);
```
TooTallNate, line 20: I think `--yes` is still necessary. This test is failing when I run it locally.

TooTallNate (changes requested):

packages/cli/test/commands/pull.test.ts
```diff
@@ -28,6 +28,34 @@ describe('pull', () => {
     expect(devFileHasDevEnv).toBeTruthy();
   });
 
+  it('should fail with message to pull without a link and without --env', async () => {
```
TooTallNate, line 31: This one doesn't seem to be passing for me locally (note that the prompt is displayed):

```
$ jest test/commands/pull.test.ts  
? Set up “/var/folders/qx/jr4274cx1_dbhytl3_rb6frc0000gn/T/tmp-47194jgRxSgyuVwVv/0/vercel-pull-next”? [Y/n] 
 FAIL  test/commands/pull.test.ts (5.929 s) 
  ● Console

    console.warn
      [Vercel API Mock] `GET /v8/projects/vercel-pull-next` was not handled.

      at test/test/mocks/client.ts:43:15
      at Layer.handle [as handle_request] (../../node_modules/express/lib/router/layer.js:95:5)
      at trim_prefix (../../node_modules/express/lib/router/index.js:317:13)
      at ../../node_modules/express/lib/router/index.js:284:7
      at Function.process_params (../../node_modules/express/lib/router/index.js:335:12)
      at next (../../node_modules/express/lib/router/index.js:275:10)
      at ../../node_modules/express/lib/router/index.js:635:15
      at next (../../node_modules/express/lib/router/index.js:260:14)

    console.warn
      [Vercel API Mock] `GET /v8/projects/vercel-pull-next` was not handled.

      at test/test/mocks/client.ts:43:15
      at Layer.handle [as handle_request] (../../node_modules/express/lib/router/layer.js:95:5)
      at trim_prefix (../../node_modules/express/lib/router/index.js:317:13)
      at ../../node_modules/express/lib/router/index.js:284:7
      at Function.process_params (../../node_modules/express/lib/router/index.js:335:12)
      at next (../../node_modules/express/lib/router/index.js:275:10)
      at ../../node_modules/express/lib/router/index.js:635:15
      at next (../../node_modules/express/lib/router/index.js:260:14)

  ● pull › should fail with message to pull without a link and without --env

    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      at test/test/commands/pull.test.ts:31:6
      at Object.<anonymous> (test/test/commands/pull.test.ts:10:1)
      at TestScheduler.scheduleTests (../../node_modules/@jest/core/build/TestScheduler.js:317:13)
      at runJest (../../node_modules/@jest/core/build/runJest.js:407:19)
      at _run10000 (../../node_modules/@jest/core/build/cli/index.js:338:7)
      at runCLI (../../node_modules/@jest/core/build/cli/index.js:190:3)
  
Test Suites: 1 failed, 1 total
Tests:       1 failed, 5 skipped, 6 total
Snapshots:   0 total
Time:        6.047 s
Ran all test suites matching /test\/commands\/pull.test.ts/i.
```

EndangeredMassa: Huh, this was passing. I can't tell if a merge to main caused something to shift, but I did push up a change that fixes the problem.

packages/cli/test/commands/pull.test.ts
```diff
@@ -28,6 +28,40 @@ describe('pull', () => {
...
+    try {
+      process.stdout.isTTY = undefined;
```
EndangeredMassa, line 33: This is definitely not ideal. We'll want to refactor the commands to not directly reference global values (like `process.stdout`), but the approach in this PR seems fine for now.

packages/cli/test/commands/pull.test.ts
TooTallNate: I guess it was because I'm running the tests from my terminal so `isTTY` was indeed true. Mocking it here seems like the correct approach.

packages/cli/test/commands/pull.test.ts
TooTallNate: `client.input` mock SGTM 😄

This works for now though 👍

TooTallNate (approved):

PR #7702, opened 2022-04-21
[cli] update language of prebuilt error

EndangeredMassa:
Updating the language of the error that shows when `vc deploy --prebuilt` is used without a prebuilt output.


styfle (approved):
Thanks!

TooTallNate (approved):

PR #7704, opened 2022-04-21
[static-build] Support Build Output API v3 `cache` property in `prepareCache()`

TooTallNate:
When a framework that outputs Build Output API v3 specifies the `cache` property in the `config.json` file then static-build will include those files in the `prepareCache()` function result.


packages/static-build/test/prepare-cache.test.ts
```diff
@@ -59,4 +59,18 @@ describe('prepareCache()', () => {
...
   });
+
+  test('should cache Build Output API v3 "cache" assets from `config.json` file', async () => {
+    const files = await prepareCache({
+      config: {},
+      workPath: path.resolve(__dirname, './cache-fixtures/build-output-api-v3'),
+      entrypoint: 'package.json',
+      files: {},
+    });
+    expect(Object.keys(files).sort()).toStrictEqual([
```
styfle, line 70: Should this also cache `node_modules`?

styfle (approved):

packages/static-build/test/prepare-cache.test.ts
TooTallNate: I don't think so. Frameworks can be explicit about that if they want to (that's how it works in v1 anyway).

PR #7709, opened 2022-04-22
[python] Add discontinue date for Python 3.6

styfle:
This PR does a few things:
- Changes the existing warning message for Python 3.6 to print a discontinue date
- Will automatically fail new Python 3.6 deployments created after that date
- Consolidates logic to make Python version selection work in a similar manner to Node.js version selection
- Changes tests from JS to TS


packages/python/src/version.ts
```diff
@@ -0,0 +1,94 @@
...
+      code: 'BUILD_UTILS_PYTHON_VERSION_DISCONTINUED',
+      link: 'http://vercel.link/python-version',
+      message: `Python version "${selection.version}" detected in Pipfile.lock is discontinued and must be upgraded. ${upstreamProvider}.`,
+    });
+  }
+
+  if (selection.discontinueDate) {
+    const d = selection.discontinueDate.toISOString().split('T')[0];
+    console.warn(
+      `Error: Python version "${selection.version}" detected in Pipfile.lock is deprecated. Deployments created on or after ${d} will fail to build. ${upstreamProvider}. http://vercel.link/python-version`
```
EndangeredMassa, line 85: Should the docs at https://vercel.com/docs/runtimes#official-runtimes/python/python-version also be updated to say when python 3.6 will no longer be supported?

packages/python/test/unit.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  global.Date.now = () => new Date('2021-07-01').getTime();
+
+  expect(
+    getSupportedPythonVersion({ pipLockPythonVersion: '3.6' })
+  ).toHaveProperty('runtime', 'python3.6');
+  expect(warningMessages).toStrictEqual([
+    'Error: Python version "3.6" detected in Pipfile.lock is deprecated. Deployments created on or after 2022-07-18 will fail to build. This change is the result of a decision made by an upstream infrastructure provider (AWS). http://vercel.link/python-version',
+  ]);
+
+  global.Date.now = realDateNow;
```
EndangeredMassa: Can we wrap this test body in a `try-finally` so that an exception wouldn't cause test pollution in future tests? Same goes for the other tests that manipulate `global.Date.now`.

```suggestion
  const realDateNow = Date.now.bind(global.Date);
  try {
    global.Date.now = () => new Date('2021-07-01').getTime();

    expect(
      getSupportedPythonVersion({ pipLockPythonVersion: '3.6' })
    ).toHaveProperty('runtime', 'python3.6');
    expect(warningMessages).toStrictEqual([
      'Error: Python version "3.6" detected in Pipfile.lock is deprecated. Deployments created on or after 2022-07-18 will fail to build. This change is the result of a decision made by an upstream infrastructure provider (AWS). http://vercel.link/python-version',
    ]);
  } finally {
    global.Date.now = realDateNow;
  }
```

packages/python/src/version.ts
```diff
@@ -0,0 +1,94 @@
...
+
+interface PythonVersion {
+  version: string;
+  pipPath: string;
+  pythonPath: string;
+  runtime: string;
+  discontinueDate?: Date;
+}
+
+const allOptions: PythonVersion[] = [
```
EndangeredMassa, line 12: Should this have a comment that explains the the order is important? It seems that the first element needs to be the most recent version.

EndangeredMassa (approved):

packages/python/src/version.ts
styfle: With the Node.js version selection, order matters because we're doing a semver match on the range. However, with Python version selection here, order does _not_  matter because its doing an exact match.

packages/python/src/version.ts
styfle: Yep, will do 👍

packages/python/test/unit.test.ts
styfle: I updated the code to use `beforeEach()` and `afterEach()` to avoid putting try/finally in every test 4a5bea1

EndangeredMassa (approved):

TooTallNate (approved):

PR #7710, opened 2022-04-22
[node][static-build][redwood] add root path pattern to prepareCache

Ethan-Arrowood:
### Related Issues

Adds root path caching support to other builders

#### Tests

#### Code Review


packages/node/src/index.ts
```diff
@@ -381,7 +381,10 @@ export const build: BuildV3 = async ({
 };
 
 export const prepareCache: PrepareCache = async ({ workPath }) => {
-  const cache = await glob('node_modules/**', workPath);
+  const cache = {
```
Ethan-Arrowood: @TooTallNate @styfle - i didn't see a test for this block. Any ideas how I can test it? This package has an integration test folder but those don't assert the cache as far as i can tell. WDYT?

packages/static-build/src/index.ts
```diff
@@ -786,6 +786,8 @@ export const prepareCache: PrepareCache = async ({
     await glob('{.shadow-cljs,node_modules}/**', workPath)
   );
 
+  Object.assign(cacheFiles, await glob('**/node_modules/**', workPath));
```
styfle: Is there a way to get the glob pattern above "Default cache files" to include this pattern as well? That way we don't need to pass over all the files twice?

Ethan-Arrowood: let me try adjusting the previous one - i don't see why it wouldn't include all the paths

packages/node/src/index.ts
```diff
@@ -381,7 +381,10 @@ export const build: BuildV3 = async ({
...
+    ...(await glob('node_modules/**', workPath)),
+    ...(await glob('**/node_modules/**', workPath)),
```
styfle: Same for this glob, can it be a single pattern to match both cases?

packages/node/src/index.ts
styfle: I think the same style of unit tests for static-build are best

packages/node/src/index.ts
TooTallNate: I think we need to consider `repoRootPath` here, when it's set.

packages/node/src/index.ts
TooTallNate: I suggest taking inspiration from [here](https://github.com/vercel/vercel/blob/main/packages/static-build/test/prepare-cache.test.ts) (that is, unit test the `prepareCache()` function).

TooTallNate (approved):

styfle (approved):

Ethan-Arrowood:
Any ideas why the E2E tests keep hanging indefinitely? @styfle @TooTallNate

styfle:
@Ethan-Arrowood Looks like its not hanging anymore but one did time out. You can also run E2E tests locally to confirm they work since GH Actions has been really flaky with network IO.

Ethan-Arrowood:
i will attempt to run locally - usually many things break when I run the whole suite locally

PR #7727, opened 2022-04-27
[tests] Bump dev dependencies for jest/eslint/prettier/turbo/etc

styfle:
Bump dev dependencies for jest/eslint/prettier/turbo/etc

### DIFF
--- a/packages/build-utils/package.json
+++ b/packages/build-utils/package.json
@@ -23,7 +23,7 @@
     "@types/end-of-stream": "^1.4.0",
     "@types/fs-extra": "9.0.13",
     "@types/glob": "^7.1.1",
-    "@types/jest": "27.0.1",
+    "@types/jest": "27.4.1",
     "@types/js-yaml": "3.12.1",
     "@types/ms": "0.7.31",
     "@types/multistream": "2.1.1",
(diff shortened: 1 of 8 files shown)


packages/build-utils/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/end-of-stream": "^1.4.0",
     "@types/fs-extra": "9.0.13",
     "@types/glob": "^7.1.1",
-    "@types/jest": "27.0.1",
+    "@types/jest": "27.4.1",
```
Ethan-Arrowood, line 26: jest types only at v27 but jest and ts-jest are at 28

Ethan-Arrowood (changes requested):

AndyBitz (approved):

packages/build-utils/package.json
styfle: Unfortunately, there are no `@types/jest` for 28 yet https://www.npmjs.com/package/@types/jest?activeTab=versions

Do you see any incompatible TS types with 28? https://jestjs.io/docs/upgrading-to-jest28

Ethan-Arrowood: I'll take a closer look first thing today to be sure; but if builds pass I assume we are all good too

Ethan-Arrowood (approved):

PR #7748, opened 2022-05-03
[cli] check prebuilt target env against deploy target env

EndangeredMassa:
This check prevents a user from creating a build output targeting one environment (like `preview`) and deploying targeting another environment (like `production`). It checks `.vercel/output/builds.json` for the `target` property to decide this. If that file is missing, the check is not run.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +235,21 @@ export default async (client: Client) => {
       );
       return 1;
     }
+
+    const prebuiltBuild = fs.readJsonSync(
+      join(path, '.vercel/output/builds.json'),
+      { throws: false }
+    );
+    if (prebuiltBuild && prebuiltBuild.target !== target) {
```
TooTallNate: Let's also handle if `builds.json` exists, but `target` was not set.

```suggestion
    if (prebuiltBuild?.target && prebuiltBuild.target !== target) {
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +235,21 @@ export default async (client: Client) => {
       );
       return 1;
     }
+
+    const prebuiltBuild = fs.readJsonSync(
```
TooTallNate: Not a huge deal, but for consistency let's stick with async operations.

```suggestion
    const prebuiltBuild = await fs.readJSON(
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +235,21 @@ export default async (client: Client) => {
...
+      join(path, '.vercel/output/builds.json'),
+      { throws: false }
+    );
+    if (prebuiltBuild && prebuiltBuild.target !== target) {
+      error(
+        `The ${param(
```
TooTallNate: We'll want to include a suggestion (i.e. "Please run `vc --prebuilt --target=___` instead") and have an err.sh link to a docs page / FAQ section with a more detailed explanation.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -62,6 +62,41 @@ import { Output } from '../../util/output';
 import { help } from './args';
 import { getDeploymentChecks } from '../../util/deploy/get-deployment-checks';
 
+function parseTarget(
```
TooTallNate: Move to `src/utils/deploy/parse-target.ts`? Can add unit tests that way, and also could be re-used in `vc build`.

packages/cli/src/commands/deploy/index.ts
styfle: I believe we use vercel.link instead of err.sh now

EndangeredMassa: I linked this to https://vercel.com/docs/cli#introduction/extended-usage where we can add some more details.

EndangeredMassa: How do I make a vercel.link ?

EndangeredMassa: I requested access to bit.ly

EndangeredMassa: Is there any guidance on how exactly to make these links? Do we add utm params and/or make a "campaign"? I searched Notion, but didn't find anything.

Screen Shot 2022-05-04 at 10 25 57 AM

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +201,23 @@ export default async (client: Client) => {
...
+    );
+    if (prebuiltBuild?.target && prebuiltBuild.target !== target) {
+      error(
+        `The ${param(
+          '--prebuilt'
+        )} option was used with the target environment "${target}", but the prebuilt output found in ".vercel/output" was built with target environment "${
+          prebuiltBuild.target
+        }". Please run ${getCommandName(
+          `--prebuilt --target=${target}`
+        )}. See https://vercel.com/docs/cli#introduction/extended-usage for more details.`
```
TooTallNate: For the link part you can use `output.prettyError()` which has a property to specify a link.

packages/cli/src/commands/deploy/index.ts
TooTallNate: There's not really AFAIK, but the nice part is that we can update the target later if we need to, and the link that we use here in CLI won't need to change.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +201,28 @@ export default async (client: Client) => {
       );
       return 1;
     }
+
+    const prebuiltBuild = await fs.readJSON(
+      join(path, '.vercel/output/builds.json'),
+      { throws: false }
+    );
+    const assumedTarget = target || 'preview';
```
EndangeredMassa, line 207: This defaults the target to "preview" for the purposes of comparison only. If we did this at the root of the CLI, other things will break.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +201,28 @@ export default async (client: Client) => {
...
+    if (prebuiltBuild?.target && prebuiltBuild.target !== assumedTarget) {
+      let specifyTarget = '';
+      if (prebuiltBuild.target === 'production') {
+        specifyTarget = ` --prod`;
```
EndangeredMassa, line 211: I used `--prod` here because `--target` is deprecated.

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -65,7 +67,8 @@ export default async function setupAndLink(
 
   if (forceDelete) {
     const vercelDir = getVercelDirectory(path);
-    remove(vercelDir);
+    remove(join(vercelDir, VERCEL_DIR_README));
+    remove(join(vercelDir, VERCEL_DIR_PROJECT));
```
EndangeredMassa, line 71: Is this OK? I wasn't sure what the full intent of `forceDelete` was, but it was destroying all fixture setup under `.vercel`.

TooTallNate: I think this is fine. We only set this to true in the `vc link` command, to re-force the prompt.

TooTallNate (approved):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +201,28 @@ export default async (client: Client) => {
       );
       return 1;
     }
+
+    const prebuiltBuild = await fs.readJSON(
+      join(path, '.vercel/output/builds.json'),
+      { throws: false }
```
TooTallNate:
```
$ vc --prebuilt
Vercel CLI 24.2.1-canary.1 — https://vercel.com/feedback
Error! ENOENT: no such file or directory, open '/Users/nrajlich/Code/vercel/vercel/examples/solidstart/.vercel/output/builds.json'
```

This option doesn't seem to work.

TooTallNate (changes requested):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -194,6 +201,28 @@ export default async (client: Client) => {
       );
       return 1;
     }
+
+    const prebuiltBuild = fs.readJSONSync(
```
EndangeredMassa: `fs.readJSONSync` behaves differently from `fs.readJSON` when `throws: false` is passed in. The sync version will not throw when the file is missing, but the async version will. (They both won't throw if parsing the json in that file fails.)

packages/cli/src/util/deploy/get-prebuilt-json.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs-extra';
+import { join } from 'path';
+
+export default async function getPrebuiltJson(directory: string) {
+  try {
+    return await fs.readJSON(join(directory, '.vercel/output/builds.json'));
```
EndangeredMassa, line 6: Turns out, the behavior of `readJSON` changes between the sync and async version of the function. The sync version will catch errors when finding the file or parsing it, but the async version only catches errors when parsing the file.
https://github.com/jprichardson/node-jsonfile/blob/master/index.js#L10-L61

It looks like they modified the behavior for the sync version without making the same change to the async version.
https://github.com/jprichardson/node-jsonfile/pull/42/files

TooTallNate (approved):

PR #7754, opened 2022-05-04
[cli] Track source of getting decrypted environment variables

JohnPhamous:
### Related Issues

We're adding audit logs/activities for accessing decrypted environment variable values. This PR adds the the command that requested for the environment variables to the API call so we can attribute how someone accessed the decrypted values.

API Change: https://github.com/vercel/api/pull/12207

#### Tests

#### Code Review


packages/cli/src/util/env/get-env-records.ts
```diff
@@ -7,6 +7,8 @@ export default async function getEnvRecords(
   output: Output,
   client: Client,
   projectId: string,
+  /** The CLI command that was used that needs the environment variables. */
+  source: string,
```
baruchadi: what do you think of using an enum or a union type for what source could be, i.e.

`EnvRecordsSourceType = 'vercel-cli:env:rm' | 'vercel-cli:pull' | ...`
- Will give us autocomplete 
- Will prevent easily avoidable errors like misspelling the source

baruchadi (changes requested):

packages/cli/src/util/env/get-env-records.ts
JohnPhamous: Ideally there's type sharing between `cli` and `api` to prevent typos. I like this idea to prevent typos in `cli`.

baruchadi (approved):

packages/cli/src/util/env/get-env-records.ts
```diff
@@ -3,10 +3,19 @@ import Client from '../client';
 import { ProjectEnvVariable, ProjectEnvTarget } from '../../types';
 import { URLSearchParams } from 'url';
 
+/** The CLI command that was used that needs the environment variables. */
+export type EnvRecordsSource =
+  | 'vercel-cli:env:ls'
+  | 'vercel-cli:env:add'
+  | 'vercel-cli:env:rm'
```
styfle, line 10: What about `vercel-cli:env:pull`?

That is a different command than `vercel-cli:pull`

packages/cli/src/commands/env/pull.ts
```diff
@@ -90,7 +90,13 @@ export default async function pull(
   output.spinner('Downloading');
 
   const [{ envs: projectEnvs }, { systemEnvValues }] = await Promise.all([
-    getDecryptedEnvRecords(output, client, project.id, environment),
+    getDecryptedEnvRecords(
+      output,
+      client,
+      project.id,
+      'vercel-cli:pull',
```
styfle:
```suggestion
      'vercel-cli:env:pull',
```

packages/cli/src/util/env/get-env-records.ts
JohnPhamous: Thanks for catching this! Adding that.

styfle (approved):
Looks good once the API ships

PR #7755, opened 2022-05-05
[build-utils] Add `ShouldServe` type

TooTallNate:
Version 3 Builders can define a `shouldServe()` function that is used in `vercel dev`, so add the proper type for that.


styfle (approved):
I was thinking this was only used for v2.
But if its still used for v3 then looks good 👍

EndangeredMassa (approved):

packages/build-utils/src/should-serve.ts
```diff
@@ -1,12 +1,12 @@
 import { parse } from 'path';
-import { ShouldServeOptions } from './types';
-import FileFsRef from './file-fs-ref';
+import type FileFsRef from './file-fs-ref';
+import type { ShouldServe } from './types';
 
-export default function shouldServe({
```
EndangeredMassa, line 5: CI didn't like the removal of `default`

TooTallNate:
@styfle I checked and I don't see any v2 Builders using `shouldServe()`. In `vc dev` we just fall back to proxying to the frontend dev server when none of the matched builders' `shouldServe()` returns true.

EndangeredMassa (approved):

PR #7737, opened 2022-04-29
[build-utils] Add supported list of workspace managers

agadzik:
In order to support monorepo detection, we need to build out the list of supported workspace managers so that our helper functions and API endpoints can utilize the same `detectFramework` function / logic to detect a workspace manager for a given git repository.

### Related Issues

- Closes https://github.com/vercel/vercel/issues/7731

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [x] Issue from task tracker has a link to this PR

PR #7760, opened 2022-05-05
Add list of supported monorepo managers

### LINKED ISSUE #7757: [Monorepo Detection] Add monorepo-managers.ts
This file will mimic the structure of `vercel/frameworks/frameworks.ts` in the sense that it defines the list of supported monorepo managers.  These are the ones were targeting for the initial launch,

- Turborepo

```ts
export const monorepoManagers = [
  {
    name: "Turborepo",
    slug: "turbo",
    detectors: {
      every: [
        {
          path: 'turbo.json'
        },
      ],
    },
  },
]
```

agadzik:
This file will mimic the structure of `vercel/frameworks/frameworks.ts` in the sense that it defines the list of supported monorepo managers.  These are the ones were targeting for the initial launch,

- Turborepo

```ts
export const monorepoManagers = [
  {
    name: "Turborepo",
    slug: "turbo",
    detectors: {
      every: [
        {
          path: 'turbo.json'
        },
      ],
    },
  },
]
```

### Related Issues

> Closes https://github.com/vercel/vercel/issues/7757

#### Tests

#### Code Review


packages/build-utils/src/workspaces/workspace-managers.ts
```diff
@@ -6,7 +6,7 @@ import type { Framework } from '@vercel/frameworks';
  * This list is designed to work with the @see {@link detectFramework} function.
  *
  * @example
- *   import { workspaceManagers as frameworkList } from '@vercel/build-utils/workspaces'
+ *   import { workspaceManagers as frameworkList } from '@vercel/build-utils'
```
chloetedder, line 9:
```suggestion
 *   import { workspaceManagers as workspaceManagerList } from '@vercel/build-utils'
```

agadzik: I think we're ok with my changes as we're showing that we can rename `workspaceManagers` to `frameworkList` and pass that into `detectFramework` 🙌

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,53 @@
...
+      outputDirectory: {
+        value: '',
+        placeholder: '',
+      },
+    },
+    getOutputDirName: () => Promise.resolve(''),
+  },
+];
+
+export default monorepoManagers;
```
chloetedder, line 36: Wondering if we should add these to the index file to export just from `@vercel/build-utils`?

agadzik: opps - good catch!  Added

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @example
+ *   import { monorepoManagers as frameworkList } from '@vercel/build-utils'
+ *   import { detectFramework } from '@vercel/build-utils'
+ *
+ *   const fs = new GitDetectorFilesystem(...)
+ *   detectFramwork({ fs, frameworkList }) // returns the 'slug' field if detected, otherwise null
+ *
+ * @todo Will be used by the detect-eligible-projects API endpoint for a given git url.
+ */
+export const monorepoManagers: Array<Framework> = [
```
styfle: Should this code live in `@vercel/frameworks` instead?

I'm wondering if we need a `@vercel/monorepos` pkg instead for all this new code?

Thoughts @TooTallNate ?

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,53 @@
...
+    name: 'Turborepo',
+    slug: 'turbo',
+    detectors: {
+      every: [
+        {
+          path: 'turbo.json',
+        },
+      ],
+    },
+    // unused props - needed for typescript
```
styfle: This seems like the wrong solution. Perhaps we need a new type

packages/build-utils/src/monorepos/monorepo-managers.ts
nathanhammond: Realistically there are three things at play: frameworks (lol no way to list them all), package managers (npm, pnpm, yarn, berry), package manager managers (ecosystem integration tools) (Turborepo, lerna, lerna-lite, [wireit](https://github.com/google/wireit), ...)

Note that I'd *maybe* shy away from the term "monorepo" as we don't consider Turborepo to be constrained by monorepo.

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @todo Will be used by the detect-eligible-projects API endpoint for a given git url.
+ */
+export const monorepoManagers: Array<Framework> = [
+  {
+    name: 'Turborepo',
+    slug: 'turbo',
+    detectors: {
+      every: [
+        {
+          path: 'turbo.json',
```
nathanhammond, line 25: This isn't guaranteed. Previously our config was at the `turbo` key inside of `$ROOT/package.json`.

packages/build-utils/src/monorepos/monorepo-managers.ts
agadzik: I suspect we'll find an use for them in the future, but for now they're unused.  These changes were based on @Ethan-Arrowood's comments here https://github.com/vercel/vercel/pull/7737#discussion_r865340977

packages/build-utils/src/monorepos/monorepo-managers.ts
agadzik: Happy to move code if needed.  Our thoughts were to just add them to `build-utils` at first, then refactor if needed

agadzik: could call the new package (if needed) `@vercel/repo-utils`

nathanhammond: Yeah, this is just a naming/organization question; I have no strong opinions other than avoiding the word "monorepo." And maybe even "repo" ... `@vercel/project-utils` maybe?

nathanhammond (approved):
From the perspective of Turborepo this is good to go.

packages/build-utils/src/monorepos/monorepo-managers.ts
styfle: The suggestion there mentions Omit

Let use something like this:

```
Array<Omit<Framework, 'logo' | 'settings' | 'getOutputDirName'>>
```

agadzik: image

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,37 @@
...
+ * This list is designed to work with the @see {@link detectFramework} function.
+ *
+ * @example
+ *   import { monorepoManagers as frameworkList } from '@vercel/build-utils'
+ *   import { detectFramework } from '@vercel/build-utils'
+ *
+ *   const fs = new GitDetectorFilesystem(...)
+ *   detectFramwork({ fs, frameworkList }) // returns the 'slug' field if detected, otherwise null
+ *
+ * @todo Will be used by the detect-eligible-projects API endpoint for a given git url.
```
EndangeredMassa: Will this `@todo` be removed after https://github.com/vercel/api/issues/12179 is done? If so, can we add a note on that issue to make sure this is removed later?

(I'm still learning how different teams are tracking tasks. Thanks!)

agadzik: oh shoot - ill just remove this! good catch

EndangeredMassa (approved):
Given the current state of the `@vercel/frameworks` package, I think this is a reasonable approach.

That said, the use of `Omit` to remove required keys from the `Framework` type seems brittle (because an addition to `Framework` may require a change here) and incomplete (because the optional keys still exist on the new resulting type, but there's not clear intent to use those.

Not in this PR, but in the near future: I think we should refactor frameworks to better accommodate this kind of use, colocate appropriate logic (based on concerns brought up in this PR discussion), rename terms and repos (based on concerns brought up in this PR discussion), and allow for different versions to have different values ([identified here](https://github.com/vercel/vercel/pull/7808#issuecomment-1130333998)). Not all of that would have to be done at once, of course. I can write up an RFC, unless this really isn't a big deal right now.

packages/build-utils/src/monorepos/monorepo-managers.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @example
+ *   import { monorepoManagers as frameworkList } from '@vercel/build-utils'
+ *   import { detectFramework } from '@vercel/build-utils'
+ *
+ *   const fs = new GitDetectorFilesystem(...)
+ *   detectFramwork({ fs, frameworkList }) // returns the 'slug' field if detected, otherwise null
+ *
+ */
+export const monorepoManagers: Array<
```
styfle, line 16: In hindsight, this should probably be

```ts
Array<Pick<Framework, 'name' | 'slug' | 'detectors'>>
```

PR #7766, opened 2022-05-09
[frameworks] Add defaults routes for vite

cesargdm:
Vite works really similar to CRA, but currently it lacks the ability to work as a SPA to redirect 404s to the entry file (index.html).

As you can see in  [this example created with vercel init vite](https://vite-phi-blush.vercel.app) when you refresh in any other route you will get a 404.

So this PR mimics the existing configuration for CRA, to make a React (or any framework) SPA created with Vite.

### Related Issues

None yet

#### Tests

#### Code Review


cesargdm:
@AndyBitz any change this can get reviewed?

packages/frameworks/src/frameworks.ts
```diff
@@ -1992,7 +1992,7 @@ export const frameworks = [
     logo: 'https://raw.githubusercontent.com/vercel/vercel/main/packages/frameworks/logos/vite.svg',
     tagline:
       'Vite is a new breed of frontend build tool that significantly improves the frontend development experience.',
-    description: 'A Vue.js app, created with Vite.',
+    description: 'A Vue.js or React app, created with Vite.',
```
styfle: The description here is for the `demo: https://vite-vue-template.vercel.app` which comes from the examples repo. Since the example is using Vue.js, the description should remain

https://vercel.com/new/templates

That being said, do we need two different detection mechanisms for vue and react. Or do they both share the same settings/routes?

packages/frameworks/src/frameworks.ts
```diff
@@ -2022,6 +2022,61 @@ export const frameworks = [
     },
     dependency: 'vite',
     getOutputDirName: async () => 'dist',
+    defaultRoutes: [
+      {
+        src: '/public/(.*)',
```
styfle: I don't think this is correct. This would only work if `vite build` emits a `./dist/public` directory with static assets hashed. Perhaps it should be `/assets/(.*)` instead?

packages/frameworks/src/frameworks.ts
```diff
@@ -2022,6 +2022,61 @@ export const frameworks = [
...
+        headers: { 'cache-control': 's-maxage=0' },
+        continue: true,
+      },
+      {
+        handle: 'filesystem',
+      },
+      { src: '/public/(.*)', status: 404, dest: '/404.html' },
+      {
+        src: '/(.*)',
+        headers: { 'cache-control': 's-maxage=0' },
```
styfle, line 1826: This header not necessary

cesargdm: If it's a PWA, the cache control is important to be kept at maxage=0, this headers comes from the react routing settings

packages/frameworks/src/frameworks.ts
cesargdm: They definitely share the same settings and routes.

packages/frameworks/src/frameworks.ts
cesargdm: Actually the output file is dist for any framework.


Screen Shot 2022-05-13 at 2 43 22 PM

packages/frameworks/src/frameworks.ts
styfle: The default Cache-Control header for static files is `cache-control: public, max-age=0, must-revalidate`.

See https://vercel.com/docs/concepts/edge-network/caching#static-files-caching

packages/frameworks/src/frameworks.ts
styfle: Right, and the routes are relative the the Output Directory (in this case `dist`)

Notice how your screenshot doesn't have a `public` directory.

PR #7772, opened 2022-05-09
[build-utils] Add Node.js 16.x

styfle:
Add support for Node.js 16

- Related to https://github.com/aws/aws-lambda-base-images/issues/14#issuecomment-1120864028

### DIFF
--- a/packages/build-utils/src/fs/node-version.ts
+++ b/packages/build-utils/src/fs/node-version.ts
@@ -4,6 +4,7 @@ import { NowBuildError } from '../errors';
 import debug from '../debug';
 
 const allOptions = [
+  { major: 16, range: '16.x', runtime: 'nodejs16.x' },
   { major: 14, range: '14.x', runtime: 'nodejs14.x' },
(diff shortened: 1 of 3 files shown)


AndyBitz (approved):

packages/build-utils/src/fs/node-version.ts
```diff
@@ -4,6 +4,7 @@ import { NowBuildError } from '../errors';
 import debug from '../debug';
 
 const allOptions = [
+  { major: 16, range: '16.x', runtime: 'nodejs16.x' },
```
gudmundur, line 7: **Question**: Does this mean that we'll default to NodeJS 16.x when this change is merged and deployed? If so, I wonder if want to gradually roll it out?

styfle: Yes if its a new project and there is no package.json engines.node defined.

Existing projects will not be changed.

pcjmfranken:
Released within 5 hours of AWS' announcement, nice work!

PR #7788, opened 2022-05-12
[tests] Use `turbo` for unit tests

TooTallNate:
Update to use `turbo` in CI for unit testing. This applies splitting of the tests into multiple runners so that they run in parallel. Currently in this PR we lose Mac + Windows because I couldn't figure out how to add them to the GH Actions matrix. We can add the E2E + `vc dev` tests next in a follow-up PR.


styfle (commented):
I feel like mac and windows is pretty important so we wouldn't want to lose those

TooTallNate:
@styfle See https://github.com/vercel/vercel/pull/7795.

PR #7791, opened 2022-05-12
[static-config] Support extracting `export const config` from swc's AST

nuta:
This PR adds support for extracting `config` from swc's AST. `static-config` supports parsing from the source program but in some cases we already have parsed an AST to do other static code analysis.

### Related Issues

None

#### Tests

#### Code Review


packages/static-config/package.json
```diff
@@ -17,6 +17,7 @@
     "dist"
   ],
   "dependencies": {
+    "@swc/core": "1.2.182",
```
TooTallNate: Is this only being used for "types"? I wonder if we can move to `devDependencies` if that's the case since this package includes a bunch of binaries (I wish there was a `typesDependencies` in package.json 😢).

packages/static-config/src/swc.ts
```diff
@@ -0,0 +1,188 @@
+import {
```
styfle:
```suggestion
import type {
```

Are these all types?

packages/static-config/src/swc.ts
```diff
@@ -0,0 +1,188 @@
...
+  return node.type === 'ArrayExpression';
+}
+
+function isObjectExpression(node: Node): node is ObjectExpression {
+  return node.type === 'ObjectExpression';
+}
+
+function isKeyValueProperty(node: Node): node is KeyValueProperty {
+  return node.type === 'KeyValueProperty';
+}
```
styfle: Seems like the types are not defined properly if these assertions are necessary? I would expect `if (node.type === 'KeyValueProperty')` to narrow the type automatically 🤔

packages/static-config/src/swc.ts
```diff
@@ -0,0 +1,188 @@
...
+
+export class UnsupportedValueError extends Error {}
+
+function extractValue(node: Node): any {
```
styfle: Why is the return type `any` if going through all the trouble of getting the types correct?

packages/static-config/package.json
styfle: This is huge 🤯 

https://packagephobia.com/result?p=@swc/core@1.2.194

Agreed, it should be moved to `devDependencies`

packages/static-config/src/swc.ts
nuta: It wish so but it doesn't. tsc cannot infer `KeyValueProperty` from something like `if (node.type === 'KeyValueProperty')` and `node` is still `Node` within the then block.

JFYI, some [suggestions](https://github.com/microsoft/TypeScript/issues/10421) are in proposed but seems not active.

packages/static-config/package.json
nuta: Yes only types are used in `src`. Moved to devDependencies. What a big package! 😮

packages/static-config/src/swc.ts
nuta: Because it returns arbitrary literals but I agree that we can do better. To avoid circular references of `Value`, I defined as follows in the latest commit:

```ts
export type Value = undefined | null | boolean | string | number | any[] | Record<string, any>;
```

packages/static-config/src/swc.ts
styfle: In this case, I don't think the issue is TS.

Its likely the `@swc/core` types are wrong. It just says string here 

Can we fix it upstream in swc?

image

packages/static-config/src/swc.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+function isKeyValueProperty(node: Node): node is KeyValueProperty {
+  return node.type === 'KeyValueProperty';
+}
+
+export type Value = undefined | null | boolean | string | number | any[] | Record<string, any>;
+export class UnsupportedValueError extends Error {}
+
+function extractValue(node: Node): Value {
```
styfle:
```suggestion
function extractValue(node: Expression): Value {
```

This should remove the need for all the additional functions. You can just compare `node.type === 'FOO'` to narrow inline

packages/static-config/src/swc.ts
```diff
@@ -0,0 +1,188 @@
...
+export function extractExportedConstValue(
+  module: Module,
+  exportedName: string
+): Value | null {
+  for (const moduleItem of module.body) {
+    if (!isExportDeclaration(moduleItem)) {
+      continue;
+    }
+
+    const declration = moduleItem.declaration;
```
styfle: This has a typo. Perhaps use the following:

```suggestion
    const { declaration } = moduleItem;
```

packages/static-config/src/swc.ts
nuta: Aha! Thanks for the help 😄

styfle (approved):
Thanks!

PR #7792, opened 2022-05-12
[redwood][static-build] move path logic up so both commands get pnpm7

Ethan-Arrowood:
### Related Issues

Adds pnpm7 support to custom build commands. Takes logic from install command moves it up so both command executions get the right path.

#### Tests

#### Code Review


packages/static-build/src/index.ts
```diff
@@ -408,6 +408,10 @@ export const build: BuildV2 = async ({
     );
     const spawnOpts = getSpawnOptions(meta, nodeVersion);
 
+    if (!spawnOpts.env) {
+      spawnOpts.env = {};
+    }
```
styfle, line 413: This might cause a bug because env undefined means `process.env` if I remember correctly

packages/static-build/src/index.ts
```diff
@@ -442,34 +463,11 @@ export const build: BuildV2 = async ({
       } else if (typeof installCommand === 'string') {
         if (installCommand.trim()) {
           console.log(`Running "install" command: \`${installCommand}\`...`);
-          const { cliType, lockfileVersion } = await scanParentDirs(
-            entrypointDir
-          );
```
styfle, line 447: We used to only scan when a custom install command was used. Are you sure we need to scan above too?

Ethan-Arrowood: yes cause this to add custom build command support. otherwise I have to clone this logic to the custom build command block

packages/static-build/src/index.ts
Ethan-Arrowood: this logic was just moved from the `create-react-app` block lines 421-423. It looks like a necessary step in order to modify `.env.PATH`

styfle: Right but I'm thinking it might have been a bug for CRA 🤔 

I bet we merge spawnOpts.env with process.env so its probably fine 👍

styfle (approved):

TooTallNate (approved):

PR #7794, opened 2022-05-12
[cli] distinguish error messages

EndangeredMassa:
There was some confusion around what the actual problem was when an auth error like this was encountered in the CLI:

> Vercel CLI 24.0.1-canary.9 — https://vercel.com/feedback
> Error! Could not retrieve Project Settings. To link your Project, remove the `.vercel` directory and deploy again.
> Learn More: https://vercel.link/cannot-load-project-settings

This could happen if the user auth token was missing, expired, or the user did not have access to the team.

This PR makes it clearer which is which and what you can do about it.


packages/cli/test/mocks/team.ts
```diff
@@ -16,6 +25,25 @@ export function useTeams(teamId?: string) {
...
+      }
+
+      if (options.failNoAccess) {
+        res.statusCode = 403;
+        res.send({
+          code: 'team_unauthorized',
+          message: 'You are not authorized',
+        });
+        return;
+      }
```
EndangeredMassa, line 45: I'm not a huge fan of this logic being here because it can fall out of sync with the actual API responses. I couldn't get an integration test that would have a missing or expired token working properly.

If someone has a suggestion or wants to pair on that, let me know!

packages/cli/src/util/projects/link.ts
```diff
@@ -150,15 +150,20 @@ export async function getLinkedProject(
       getProjectByIdOrName(client, link.projectId, link.orgId),
     ]);
   } catch (err) {
-    if (err?.status === 403) {
+    if (err instanceof APIError && err.status === 403) {
```
styfle: I'm not so sure that the error is an instance of `APIError`.

Where is it thrown?

EndangeredMassa: Good call. `APIError` is thrown in some cases, but not in the `getOrgById` -> `getUser` path. Updated!

EndangeredMassa: Although, those don't have a status of 403 anyway. The end result appears to be the same regardless. Those non-`APIError` errors will get re-thrown and their messages get logged out.

TooTallNate (approved):

PR #7795, opened 2022-05-12
[tests] Add cross platform chunked testing and leverage turbo more

ijjk:
### Related Issues

x-ref: https://github.com/vercel/vercel/pull/7788

#### Tests

#### Code Review


ijjk:
Seems the one integration test failure is also failing on main so this should be good now https://github.com/vercel/vercel/runs/6413856446?check_suite_focus=true

turbo.json
```diff
@@ -19,7 +19,23 @@
...
+    "test-integration-cli": {
+      "dependsOn": ["build"],
+      "outputs": []
+    },
+    "test-integration-once": {
+      "dependsOn": ["build"],
+      "outputs": []
+    },
+    "test": {
+      "dependsOn": ["build"],
```
styfle, line 38: What’s the purpose of this? Does it run every test regardless if unit/integration/e2e?

styfle (commented):
Looks like the vercel/node tests were not executed. Is that expected?

turbo.json
ijjk: This allows turbo to run the specific tests detected from running `test-unit` as `test-unit` lists the paths in the command already but we want to be able to pass specific paths when splitting them across jobs.

ijjk:
The `@verce/node` tests were run in this job https://github.com/vercel/vercel/runs/6416624520?check_suite_focus=true

ijjk:
@TooTallNate do we want to increase the timeouts for the failing unit tests or should these be passing within that range?

utils/chunk-tests.js
```diff
@@ -0,0 +1,152 @@
...
+  // ['test-integration-dev', ['macos-latest', 'ubuntu-latest']],
+  ['test-integration-once', ['ubuntu-latest']],
+]);
+
+async function getChunkedTests() {
+  const scripts = ['test-integration-once'];
+  const rootPath = path.resolve(__dirname, '..');
+
+  const dryRunText = (
+    await turbo([`run`, ...scripts, `--cache-dir=.turbo`, '--', '--listTests'])
```
styfle, line 13: Any reason why the custom cache dir is used instead of the default?

ijjk: The default is `./node_modules/.cache/turbo` ([related doc](https://turborepo.org/docs/reference/command-line-reference#--cache-dir)) which makes it trickier to restore/persist, we use the same config in Next.js' CI ([x-ref](https://github.com/vercel/next.js/blob/c72d32d3b5cce4fcf33cf621f0e43b899a493424/.github/workflows/build_test_deploy.yml#L1016))

styfle: Should we cache all of `node_modules` to make it even faster? 🤔

ijjk: Possibly although that may cause faster cache evictions as we would need a separate cache for each CI job which can add up fairly quick

styfle (approved):

utils/chunk-tests.js
```diff
@@ -0,0 +1,152 @@
+const child_process = require('child_process');
+const path = require('path');
+
+const NUMBER_OF_CHUNKS = 5;
+const MINIMUM_PER_CHUNK = 1;
+const runnersMap = new Map([
+  // ['test-integration-dev', ['macos-latest', 'ubuntu-latest']],
```
EndangeredMassa: Should this be removed or restored?

packages/static-build/test/integration-1.test.js
```diff
@@ -45,8 +46,20 @@ const testsThatFailToBuild = new Map([
   ],
 ]);
 
+const fixtureGroups = intoChunks(
+  NUMBER_OF_CHUNKS,
+  fs.readdirSync(fixturesPath)
+);
+const fixturesIndex =
+  parseInt(
+    (module.parent?.filename || __filename).match(/integration-(\d)\./)[1],
```
EndangeredMassa: Using `module` or `__filename` is confusing to follow. I suggest exporting a function that the `test/integration-*.test.js` files can require and call where they pass in the appropriate fixture group. It could be something like:

```js
const runTestGroup = require('./integration-tests-by-group');
runTestGroup(1); // 2, 3, ... in the other files
```

utils/chunk-tests.js
```diff
@@ -0,0 +1,152 @@
...
+  for (let i = 0; i < totalChunks; i++) {
+    chunks.push(arr.slice(i * chunkSize, (i + 1) * chunkSize));
+  }
+  return chunks.filter(x => x.length > 0);
+}
+
+async function main() {
+  try {
+    const chunks = await getChunkedTests(process.argv[2]);
+    // TODO: pack and build the runtimes for each package and cache it so we only deploy it once
```
EndangeredMassa, line 134: Is this work tracked somewhere?

utils/chunk-tests.js
```diff
@@ -0,0 +1,152 @@
...
+  const chunkSize = Math.max(
+    MINIMUM_PER_CHUNK,
+    Math.ceil(arr.length / totalChunks)
+  );
+  const chunks = [];
+  for (let i = 0; i < totalChunks; i++) {
+    chunks.push(arr.slice(i * chunkSize, (i + 1) * chunkSize));
+  }
+  return chunks.filter(x => x.length > 0);
+}
```
EndangeredMassa, line 129: Can we add tests for `intoChunks`? It looks like it would be easy to be off by one or not support some edge case.

EndangeredMassa (approved):
Looks great! These tests run a lot faster now and it looks like it picked them all up properly.

TooTallNate (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: 'yarn'
+      - run: yarn install --network-timeout 1000000
+
+      - name: Use Turborepo cache
+        uses: actions/cache@v2
+        with:
+          path: .turbo
+          key: turbo-test-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-${{matrix.packageName}}-${{matrix.scriptName}}-${{matrix.chunkNumber}}
```
jaredpalmer: I think the best move is to likely drop this and just specify TURBO_TOKEN and TURBO_TEAM and TURBO_REMOTE_ONLY as env. This way you don't need to worry about local caching at all

jaredpalmer (changes requested):

utils/chunk-tests.js
ijjk: This util is from [the runtimes repo](https://github.com/vercel/runtimes/blob/main/utils/chunk-tests.js) so not sure if we're tracking this somewhere related to that.

utils/chunk-tests.js
ijjk: Since the dev integration tests currently use ava which doesn't have `--listTests` I don't think we can split those up the same way so will remove for now

.github/workflows/test.yml
ijjk: Updated in https://github.com/vercel/vercel/pull/7795/commits/ea56981221eeece39779d01b297c138b65c8c0ec

packages/static-build/test/integration-1.test.js
ijjk: Updated to remove the module workaround in favor of a setup util in https://github.com/vercel/vercel/pull/7795/commits/a53d099852feb9019791ab523022315b64226ba4

utils/chunk-tests.js
ijjk: removed in https://github.com/vercel/vercel/pull/7795/commits/a53d099852feb9019791ab523022315b64226ba4

utils/chunk-tests.js
ijjk: Added in https://github.com/vercel/vercel/pull/7795/commits/a53d099852feb9019791ab523022315b64226ba4

.github/workflows/test-unit.yml
```diff
@@ -33,11 +33,20 @@ jobs:
       - run: git fetch origin ${{ github.ref }} --depth=100
       - run: git diff origin/main...HEAD --name-only
       - run: yarn install --network-timeout 1000000
+      - name: Use Turborepo cache
```
jaredpalmer: Can also remove here. Just use same env setup as the other one.

jaredpalmer (dismissed):

.github/workflows/test-unit.yml
ijjk: Updated the others separately in https://github.com/vercel/vercel/pull/7795/commits/dc23f983e7e646bac90812359c2686dfe8dece75 was waiting to see the initial cache work alright and it's looking good

PR #7808, opened 2022-05-16
[static-build] Support subset of Build Output API v2

EndangeredMassa:
Reverts: vercel/vercel#7803

Reimplements: https://github.com/vercel/vercel/pull/7690

I added this commit ([`9e15575` (#7808)](https://github.com/vercel/vercel/pull/7808/commits/9e15575a6a87c28eebe809a74d5b27a0ae4b938a)) to make the v2 detection more specific. This resolved the issue with v1 or non-Build Output API builds being improperly detected as v2.

Added a couple tests to make sure this works.


styfle (commented):
Lets add a test with `dependencies: { nuxt }` to prove it fixes the original issue

EndangeredMassa:
### The Problem

The problem with nuxt output was that some situations produced `.output`, which would confuse the Build Output API (BOA) version detection into thinking it was a BOAv2 output, when it wasn't even BOA at all or it was actually BOAv1. Here's how it happens:

Deployment Builds:

- `nuxt@2` build: outputs custom format to `dist`; our framework detection finds nuxt and leverages zero config
- `nuxt@3` build when `NOW_BUILDER='1'`: the `nuxi build` outputs BOAv1 to `.vercel_build_output`; our framework detection does NOT recognize `nuxt`, but does see `.vercel_build_output` and assumes it's BOAv1

Local Builds
- `nuxt@3` build: outputs custom format to `.output`

If you do a local build with `nuxt@3`, you'll get `.output`. If you then do a `vc deploy`, you'll push that `.output` along with your source code. The nuxt builder will produce `.vercel_build_output` because it sees `NOW_BUILDER='1'`, leaving both `.output` and `.vercel_build_output` in the working directory.

### The Solution

This PR changes the BOAv2 detection logic to look for at least one manifest file in `.output`, not just the presence of `.output` itself. This allows the `nuxt@3` build to properly produce BOAv1 and get recognized for that.

### Tests

I updated the tests with a real `nuxt` version 3 example to make it clear that this is now working.

Note that the package versions `nuxt@3.0.0-rc.3` and `nuxt3@3.0.0-rc.3-*` exist on different published npm packages. It looks like the `nuxt3` package is a test package used to house a published version for each merged PR. Then, when they are ready to declare one of those a new RC, they publish it under `nuxt@3.0.0-rc.*`. For this test, I wanted to use the latest published RC, which was `nuxt@3.0.0-rc.3`.

### Follow Up: Frameworks Detection

It looks like the `vercel/frameworks` detection can only handle version differences if they vary by slug or dependency name. You can see this in action with the multiple docusaurus entries. For any framework that will end up using the same slug and dependency name for a major version bump (that changes a detail we care about, like output directory), we'll have to refactor `vercel/frameworks` to handle the concept of different versions of frameworks more directly.

In this specific case, because nuxt@3 generates BOAv1, we didn't need to zero-config detect nuxt@3. I imagine there will be a time when we do need to detect something like this.

### Follow Up: nuxt3 Feedback

We probably want to follow up with the nuxt team about version 3:

- they should probably add `.output` to their `.vercelignore` file in any templates they use for vercel; I've already done this for [our example](https://github.com/vercel/vercel/pull/7817)
- they should update to BOAv3 if they want to take advantage of the fuller feature set or what to be on a better supported version

styfle (approved):

TooTallNate (approved):

EndangeredMassa:
@leerob @mcsdevv @AndyBitz this is ready for deploy. Do you have any review comments before we do?

PR #2812, opened 2019-08-21
Create monorepo

styfle:
This PR creates a monorepo by combining the existing `now-client`, `now-builders`, and `now-cli` repositories.

Related to #2782

PR #7809, opened 2022-05-16
[tests] Skip broken go test

styfle:
This test was added several years ago, before the monorepo in https://github.com/vercel/vercel/pull/2812 but it was never working correctly due to the way zero config behaves differently than the test with `builds` in vercel.json

We can fix it in a follow up PR

### DIFF
--- a/packages/go/test/test.js
+++ b/packages/go/test/test.js
@@ -21,10 +21,15 @@ beforeAll(async () => {
   console.log('builderUrl', builderUrl);
 });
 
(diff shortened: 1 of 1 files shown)


ijjk (approved):

EndangeredMassa (approved):
Is there a card somewhere to track fixing this test in the future?

styfle:
@EndangeredMassa https://linear.app/vercel/issue/BUI-111/

PR #7811, opened 2022-05-16
[cli] Implement `vercel.json` configuration overrides

Ethan-Arrowood:
### Related Issues

This PR adds vercel.json configuration overrides support for both new and existing project flows through the CLI. It includes integration and unit tests.

#### Tests

#### Code Review


packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -2,45 +2,109 @@ import inquirer from 'inquirer';
...
+      _framework => _framework.slug === localConfigurationOverrides.framework
+    ) ?? null) as Framework | null;
+
+    if (framework) {
+      output.print(
+        `Merging default project settings for framework ${framework.name}. Previously listed overrides are prioritized.\n`
+      );
+    } else {
+      output.print(
+        `Framework slug matching ${localConfigurationOverrides.framework} not found.`
```
styfle: When will a slug exist that doesn't match any framework?
Should this be an error instead?

Ethan-Arrowood: Good point - the api already validates against the list of known slugs so if the user provides something else the API itself will fail.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -453,6 +453,24 @@ export default async (client: Client) => {
...
 
+  // const localConfigurationOverrides = {}
+  // let overrideDetected = false;
+  // ['buildCommand', 'devCommand'].forEach(setting => {
+  //   const value = localConfig?.[setting]
+  //   if (value) {
+  //     overrideDetected = true;
+  //     localConfigurationOverrides[setting === 'ignoreCommand' ? 'commandForIgnoringBuildStep' : value]
+  //   }
+  // })
```
styfle: Is this comment still needed or can we delete it?

Ethan-Arrowood: deleting

packages/cli/src/commands/deploy/index.ts
```diff
@@ -475,7 +493,12 @@ export default async (client: Client) => {
 
     if (!localConfig.builds || localConfig.builds.length === 0) {
       // Only add projectSettings for zero config deployments
-      createArgs.projectSettings = { sourceFilesOutsideRootDirectory };
+      createArgs.projectSettings =
+        status === 'not_linked'
```
styfle, line 490: Do we need to handle `status === 'error'` here too?

Ethan-Arrowood: Unreachable I believe. There is a check higher up that is:

```js
  const link = await getLinkedProject(client, path);

  if (link.status === 'error') {
    return link.exitCode;
  }

  let { org, project, status } = link;
  ```

File: packages/cli/src/util/input/edit-project-settings.ts [hunk 4]
```diff
@@ -2,45 +2,110 @@ import inquirer from 'inquirer';
...
-  devCommand: string | null;
-}
+const settingKeys = [
+  'buildCommand',
+  'installCommand',
+  'devCommand',
+  'ignoreCommand',
+  'outputDirectory',
+  'framework',
+] as const;
```
styfle: Can we avoid defining these twice?

How about `const settingKeys = Object.keys(settingMap)` instead?

Ethan-Arrowood: Unfortunately, `Object.keys` doesn't maintaint the type automatically so I will have to write them twice regardless... but as:

```ts
const settingKeys = Object.keys(settingMap) as unknown as readonly ["buildCommand", "installCommand", "devCommand", "ignoreCommand", "outputDirectory", "framework"];
```

packages/client/src/types.ts
```diff
@@ -129,6 +129,12 @@ export interface VercelConfig {
   alias?: string | string[];
   regions?: string[];
   projectSettings?: ProjectSettings;
```
styfle, line 131: Hmm, I wonder if `projectSettings` should be removed since its not available on `vercel.json`

Ethan-Arrowood: if i remember correct it is on there, just undocumented and only used by some of our internal stuff?

packages/cli/src/util/input/edit-project-settings.ts
styfle: @Ethan-Arrowood This works:

```ts
const settingKeys = Object.keys(settingMap);

type ConfigKeys = keyof typeof settingMap;
```

styfle: And you don't need `settingsKeys` at all because you can iterate over an object like this:

```ts
for (const [key, value] of Object.entries(settingMap)) 
```

Ethan-Arrowood: Yes but a lot of the code relies on `settingKeys` being of type `readonly ["buildCommand", "installCommand", "devCommand", "ignoreCommand", "outputDirectory", "framework"]` (essentially a tuple of those strings)

Ethan-Arrowood: I can do this tho: 

```ts
type ConfigKeys = keyof typeof settingMap;
const settingKeys = Object.keys(settingMap) as unknown as readonly [ConfigKeys];
```

packages/client/src/types.ts
styfle: Is it on vercel.json or part of the API? Because this type represents vercel.json as far as I can tell

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -2,45 +2,110 @@ import inquirer from 'inquirer';
...
+type ProjectSettingKeys =
+  | Exclude<ConfigKeys, 'ignoreCommand'>
+  | 'commandForIgnoringBuildStep';
+type PartialProjectSettings = Pick<ProjectSettings, ProjectSettingKeys>;
+
+type LocalConfiguration = Pick<
+  VercelConfig,
+  | keyof Omit<PartialProjectSettings, 'commandForIgnoringBuildStep'>
+  | 'ignoreCommand'
+>;
```
styfle: This type is failing for me when I checkout the branch locally 
image

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -2,45 +2,110 @@ import inquirer from 'inquirer';
...
 
-const fields: { name: string; value: keyof PartialProjectSettings }[] = [
-  { name: 'Build Command', value: 'buildCommand' },
-  { name: 'Output Directory', value: 'outputDirectory' },
-  { name: 'Development Command', value: 'devCommand' },
-];
+/** A quick transform for 'ignoreCommand' to 'commandForIgnoringBuildStep' as they are the only key
+ * that differs between the top level vercel.json configuration settings and the api `projectSettings` object.
+ */
+const normalizeSettingName = (setting: typeof settingKeys[number]) =>
```
styfle:
```suggestion
const normalizeSettingName = (setting: ConfigKeys) =>
```

packages/cli/src/util/input/edit-project-settings.ts
Ethan-Arrowood: did you build `build-utils` and link it?

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -2,45 +2,110 @@ import inquirer from 'inquirer';
...
+const settingMap = {
+  buildCommand: 'Build Command',
+  devCommand: 'Development Command',
+  ignoreCommand: 'Ignore Command',
+  installCommand: 'Install Command',
+  outputDirectory: 'Output Directory',
+  framework: 'Framework',
+} as const;
+
+type ConfigKeys = typeof settingKeys[number];
```
styfle:
```suggestion
type ConfigKeys = keyof typeof settingMap;
```

packages/cli/src/util/input/edit-project-settings.ts
styfle: Haha nope 🤦  

Working now 👍

packages/cli/test/integration.js
```diff
@@ -3909,3 +3926,159 @@ test('[vc build] should build project with `@vercel/static-build`', async t => {
...
+  //       react: 'latest',
+  //       'react-dom': 'latest',
+  //     },
+  //   })
+  // );
+
+  // deployment = await deploy();
+  // page = await fetch(deployment.stdout);
+  // text = await page.text();
+  // t.is(text, '2\n');
```
EndangeredMassa: What's the failure when this code is run?

EndangeredMassa (dismissed):
It looks like we should figure this out before merging, right?

packages/cli/test/integration.js
Ethan-Arrowood: `Error! Command "yarn run build" exited with 1`

packages/client/src/types.ts
Ethan-Arrowood: im not sure. it looks like its been on this interface for a long while though

packages/cli/test/integration.js
Ethan-Arrowood: fixed!

EndangeredMassa:
@Ethan-Arrowood looks like there are still some test failures. Are those flakey or legit?

Ethan-Arrowood:
Oh really?! Ugh no those look legit. These tests are going to be the death of me 🙄

packages/cli/src/commands/deploy/index.ts
```diff
@@ -453,6 +453,15 @@ export default async (client: Client) => {
   let deployStamp = stamp();
   let deployment = null;
 
+  const localConfigurationOverrides = {
```
styfle: I added a type here and it demonstrated a bug.
```suggestion
  const localConfigurationOverrides: LocalConfiguration = {
```

Basically the `commandForIgnoringBuildStep` is used here and then the object is passed to `editProjectSettings()` which is expecting `ignoreCommand` instead

packages/client/src/types.ts
styfle: I noticed that `vc dev` uses projectSettings so I guess we should leave it as-is https://github.com/vercel/vercel/blob/960c66584c58ad1e88176796db829514f5db6b15/packages/cli/src/util/dev/server.ts#L567

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -216,7 +216,8 @@ export default async function setupAndLink(
         output,
         projectSettings,
         framework,
-        autoConfirm
+        autoConfirm,
+        null
```
styfle: Do we need to consider line 183 too?

https://github.com/vercel/vercel/blob/7d5d63965ecb15a888b8ed407fc72d89a1a91737/packages/cli/src/util/link/setup-and-link.ts#L183

styfle (commented):
Left a couple more comments

packages/cli/src/commands/deploy/index.ts
Ethan-Arrowood: Great catch; I think I'm actually able to simplify things significantly now too

Ethan-Arrowood: Okay latest commit simplified things quite a bit. I will ping when integration tests all pass

packages/cli/src/util/link/setup-and-link.ts
Ethan-Arrowood: This block is only executed when `isZeroConfig` is `true`. The expression for `isZeroConfig` starts with `!localConfig ||`, and so if I understand it correctly, we'd never reach line 183 if a local configuration is provided. In fact, we'd never even create a new deployment if a local configuration exists (the deployment block is wrapped in an if conditional using `isZeroConfig`

However,

styfle: Its an "OR" statement so it executes when theres no vercel.json OR no `builds` prop OR `builds` prop is zero length array.

Ethan-Arrowood: Ah i had the short circuiting logic backwards. Okay let me dig deep here

packages/cli/test/integration.js
```diff
@@ -645,6 +662,26 @@ test('should allow deploying a directory that was prebuilt, but has no builds.js
...
+    ...defaultArgs,
+  ]);
+
+  t.is(exitCode, 0, formatOutput({ stderr, stdout }));
+
+  const page = await fetch(stdout);
+  const text = await page.text();
+  t.is(text, '1\n');
+});
+
```
Ethan-Arrowood: This test is not passing. The fetch call returns a 404 - i expect to at least see `2` based on the structure of the fixture in the `prepare` script. Any ideas are appreciated 🤔

styfle: The test says `[vc link]` but its runnning `vc --prebuilt`. Is that expected?

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -176,13 +188,12 @@ export default async function setupAndLink(
...
+        projectSettings: {
+          ...localConfigurationOverrides,
+          sourceFilesOutsideRootDirectory,
+        },
       };
 
-      if (isZeroConfig) {
-        // Only add projectSettings for zero config deployments
-        createArgs.projectSettings = { sourceFilesOutsideRootDirectory };
-      }
```
styfle, line 184: I think we still need this `if` statement.

Project Settings are not applied when `builds` prop is present.

Ethan-Arrowood: This whole block is already wrapped in a if(isZeroConfig) though

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -2,45 +2,88 @@ import inquirer from 'inquirer';
...
+      if (framework) {
+        output.print(
+          `Merging default project settings for framework ${framework.name}. Previously listed overrides are prioritized.\n`
+        );
+      }
+    }
   }
 
-  // skip editing project settings if no framework is detected
+  // return early if the framework is null.
```
styfle: This comment is a bit ambiguous because "null" could be the framework object or the slug which have two different meanings

I think the previous comment was better so lets use that

styfle: Fixed in [662fcc1](https://github.com/vercel/vercel/pull/7811/commits/662fcc1a40ab4b848b54fa208da4a5a93cea8b39) along with a couple other things

packages/cli/test/helpers/prepare.js
```diff
@@ -440,6 +440,23 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
...
+        buildCommand: 'mkdir -p output && echo "1" > output/index.txt',
+      }),
+      'package.json': '{}',
+    },
+    'vercel-json-configuration-overrides-link': {
+      'vercel.json': JSON.stringify({
+        buildCommand: 'mkdir public && echo "1" > public/index.txt',
+      }),
+      '.vercel/output/config.json': JSON.stringify({ version: 3 }),
+      '.vercel/output/public/index.txt': '2',
```
styfle: Why does this test fixture have `.vercel/output`? Does that change the behavior of `vc link`? I thought that would only affect `vc build`.

Ethan-Arrowood: Yeah I think these can be removed now that the integration test only runs `vc link` instead of also making a deployment

Ethan-Arrowood: removed.

packages/cli/src/util/input/edit-project-settings.ts
Ethan-Arrowood: lgtm! ty!

packages/cli/test/integration.js
```diff
@@ -307,17 +314,27 @@ async function setupProject(process, projectName, overrides) {
...
     );
-    process.stdin.write(`${buildCommand ?? ''}\n`);
+    process.stdin.write(`${buildCommand || ''}\n`);
 
     await waitForPrompt(process, chunk =>
-      chunk.includes(`What's your Output Directory?`)
+      chunk.includes(`What's your Ignore Command?`)
     );
-    process.stdin.write(`${outputDirectory ?? ''}\n`);
+    process.stdin.write(`${ignoreCommand || ''}\n`);
```
styfle: I don't think this is correct. We shouldn't as for an ignore command because this isn't a Git deployment, its from the CLI.

Similarly, I don't think we should ask for Install Command since we don't do that on vercel.com/new (although maybe we should)

Ethan-Arrowood: I'm good with removing the ignoreCommand prompt. But I think having installCommand would be useful. Is there a specific reason when it is not included on `/new`? Or is it just cause it was never added?

styfle: As we discussed in the meeting, lets save that feature for another PR and keep this PR to the scope of overrides

Ethan-Arrowood: (this question was answered during a call - posting answer here for consistency)

- decided to remove install and ignore command in order to match how the `/new/import` ux works. 
- install command is not included in the ux flow because it would be "too many options". essentially, probably only a minimal % of users are even configuring this setting so its okay to omit. if it was included more users may accidentally configure it and break their deployments (this parallels to some customer issues we run into along the lines of misunderstanding that they dont _have_ to enter an override when prompted)
-

Ethan-Arrowood: (this question was answered during a call - posting answer here for consistency)

- decided to remove install and ignore command in order to match how the `/new/import` ux works. 
- install command is not included in the ux flow because it would be "too many options". essentially, probably only a minimal % of users are even configuring this setting so its okay to omit. if it was included more users may accidentally configure it and break their deployments (this parallels to some customer issues we run into along the lines of misunderstanding that they dont _have_ to enter an override when prompted)
-

styfle (approved):
Great work, thanks! 🥇

PR #7814, opened 2022-05-17
[remix] Add `@vercel/remix` Builder

TooTallNate:
Adds a new Builder for the Remix framework. This Builder has support for a few different configurations of Remix including the "just the basics" Remix server as well as the "vercel" server build target configuration. If the project does not have the `@remix-run/vercel` runtime adapter as a dependency in package.json then this Builder will implicitly add it when generating the SSR serverless function.

In a follow-up PR, we can make this Builder be used in the frameworks configuration so that it is executed when `remix` framework is selected for the project, and we can update the remix example template in this repo to no longer require the `ENABLE_FILE_SYSTEM_API` env var.


.github/workflows/test-unit.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [12]
+        node: [14]
```
TooTallNate, line 19: Note that Remix needs Node 14 to install dependencies. Figured it's a good time to update usage in the tests, considering that Node 12 is EOL.

packages/remix/test/build.test.ts
```diff
@@ -0,0 +1,53 @@
+import assert from 'assert';
+import { join } from 'path';
+import { NodejsLambda } from '@vercel/build-utils';
+import { build, prepareCache } from '../src';
+
+const fixture = (name: string) => join(__dirname, 'fixtures', name);
+
+describe('build()', () => {
+  jest.setTimeout(10 * 60 * 1000);
```
EndangeredMassa: Because `jest.setTimeout` applies to the whole file, can we move it out of the `describe` block to make it clearer it applies beyond this `describe` block?

EndangeredMassa (approved):

styfle (commented):
Do we need to add `@vercel/remix` as a dependency in `packages/cli/package.json`?

packages/remix/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "main": "./dist/index.js",
+  "homepage": "https://vercel.com/docs",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/vercel/vercel.git",
+    "directory": "packages/remix"
+  },
+  "scripts": {
+    "build": "node build.js",
+    "test-unit": "jest --env node --verbose --bail test/build.test.ts",
```
styfle: How about `test-integration-once` for the e2e tests?

TooTallNate:
> Do we need to add `@vercel/remix` as a dependency in `packages/cli/package.json`?

Yes, but in the `vc build` PR.

packages/remix/src/build.ts
```diff
@@ -0,0 +1,266 @@
...
+      (nodeVersion?.major || 0) < 16
+    ) {
+      // Ensure that npm 7 is at the beginning of the `$PATH`
+      spawnOpts.env.PATH = `/node16/bin-npm7:${spawnOpts.env.PATH}`;
+      console.log('Detected `package-lock.json` generated by npm 7...');
+    }
+  } else if (cliType === 'pnpm') {
+    if (typeof lockfileVersion === 'number' && lockfileVersion === 5.4) {
+      // Ensure that pnpm 7 is at the beginning of the `$PATH`
+      spawnOpts.env.PATH = `/pnpm7/node_modules/.bin:${spawnOpts.env.PATH}`;
```
styfle, line 68: We really need to consolidate this 😓

packages/remix/src/build.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  if (needsHandler) {
+    // Copy the `default-server.js` file into the "build" directory
+    const sourceHandlerPath = join(__dirname, '../default-server.js');
+    await fs.copyFile(sourceHandlerPath, handlerPath);
+  }
+
+  // Trace the handler with `@vercel/nft`
+  const trace = await nodeFileTrace([handlerPath], {
+    base: rootDir,
```
styfle, line 215: I think the base won't work if this is a monorepo because it should be repoRootPath

styfle (approved):

packages/remix/src/build.ts
mikedijkstra: Was this missed?

I've been unable to use the Remix preset since Friday because it's not finding the yarn.lock file in the root of my monorepo and I think this might be the cause.

styfle: Please try this https://github.com/vercel/vercel/pull/8077#pullrequestreview-1033551949

PR #7819, opened 2022-05-18
[tests] update CI to use node@14 and drop support for node@12

EndangeredMassa:
Node 12 is EOL and we're starting to reference packages that don't install on node lower than 14, such as nuxt@3 and remix. Let's update to 14 for now.


styfle (commented):
We should also update tsconfig.json target and package.json engines, then we can ship semver major to drop support for 12

EndangeredMassa:
@TooTallNate we'll need to change the "Required" checks. We should remove the node@12 versions, add the node@14 versions, and probably also add the new matrix of checks from the turborepo change, right?

packages/cli/package.json
```diff
@@ -40,7 +40,7 @@
     ]
   },
   "engines": {
-    "node": ">= 12"
+    "node": "14.x"
```
styfle:
```suggestion
    "node": ">= 14"
```

We don't want to prevent users from using Node.js 16

packages/client/package.json
```diff
@@ -20,7 +20,7 @@
     "test-unit": "yarn test tests/unit.*test.*"
   },
   "engines": {
-    "node": ">= 12"
+    "node": "14.x"
```
TooTallNate:
```suggestion
    "node": ">= 14"
```

EndangeredMassa:
I updated the tsconfig files based on the [Node Target Mapping](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping) guidance for typescript. I changed every `target` and `lib` key that I found to `ES2020`, even if they were `esnext` before. I think this makes the most sense, but if `esnext` is a better choice for these, please let me know why and I'll change them.

I left the examples alone.

TooTallNate (approved):

TooTallNate:
`next` and `remix` packages may need to be updated as well

EndangeredMassa:
@TooTallNate good call. Updated!

packages/node/src/typescript.ts
```diff
@@ -450,17 +450,14 @@ export function fixConfig(
...
 
-  // Target esnext output by default (instead of ES3).
   // This will prevent TS from polyfill/downlevel emit.
   if (config.compilerOptions.target === undefined) {
     // See https://github.com/tsconfig/bases/tree/main/bases
     let target: string;
     if (nodeVersionMajor >= 16) {
       target = 'ES2021';
-    } else if (nodeVersionMajor >= 14) {
-      target = 'ES2020';
```
styfle: We need to leave this until August 9th

https://vercel.com/changelog/node-js-12-is-being-deprecated

PR #7821, opened 2022-05-19
[cli] Remove "monorepo-dynamic-paths" Dev test fixture

TooTallNate:
This fixture test is failing (https://github.com/vercel/vercel/runs/6498856153) since it's invoking `@vercel/next` with a version that returns `Prerender` instances, however the test itself is invalid since we no longer prefer monorepos to be set up with multiple builds in the same deployment anymore (instead multiple projects should be created with the different subpaths).


ijjk (approved):

TooTallNate:
See #7822 instead.

PR #7822, opened 2022-05-19
[cli] Migrate Dev tests from `ava` to `jest`

ijjk:
### Related Issues

As discussed with @TooTallNate it would be more ideal to migrate from ava to jest for the dev CLI tests and this also allows us to split them apart and parallelize them like the `test-integration-once` tests. 

This also removes an outdated Next.js test we found after merging `@vercel/next` back in (x-ref: https://github.com/vercel/vercel/pull/7821)

#### Tests

#### Code Review


.github/workflows/cancel.yml
```diff
@@ -13,6 +13,5 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.9.1
         with:
-          workflow_id: 849295, 849296, 849297, 849298
+          workflow_id: 849295, 849296, 849297, 849298, test.yml
```
styfle: Lets swap these all to use names

ijjk: Sounds good, updated in https://github.com/vercel/vercel/pull/7822/commits/d9ac99a4fa29c1a3b61c9f9ccbc8cae820909953

packages/cli/test/dev/fixtures/26-nextjs-secrets/vercel.json
```diff
@@ -1,10 +1,10 @@
 {
   "build": {
     "env": {
-      "SECRET": "@buildtime"
+      "SECRET": "buildtime"
     }
   },
   "env": {
-    "ANOTHER_SECRET": "@runtime"
+    "ANOTHER_SECRET": "runtime"
```
styfle: I'm pretty sure the `@` is necessary, otherwise its not testing secrets

ijjk: Previously this test wasn't deployed but is now required but fails since the secrets don't exist. This used to use the `@vercel/next` builder which wasn't fully testing secrets so maybe we want to rename this to `26-nextjs-env`, wdyt @TooTallNate?

utils/chunk-tests.js
```diff
@@ -3,10 +3,13 @@ const path = require('path');
...
 const MINIMUM_PER_CHUNK = 1;
-const runnersMap = new Map([['test-integration-once', ['ubuntu-latest']]]);
+const runnersMap = new Map([
+  ['test-integration-once', ['ubuntu-latest']],
+  ['test-integration-dev', ['ubuntu-latest', 'macos-latest']],
+]);
 
 async function getChunkedTests() {
-  const scripts = ['test-integration-once'];
+  const scripts = ['test-integration-once', 'test-integration-dev'];
```
styfle: Should this be `runnersMap.keys()` instead of defining a new array?

packages/cli/test/dev/fixtures/26-nextjs-secrets/vercel.json
TooTallNate: I think we should remove this test completely. The `dev-server.ts` file in vercel/next doesn't even exist anymore.

utils/chunk-tests.js
ijjk: Updated in https://github.com/vercel/vercel/pull/7822/commits/f0ccb8115583ca0796335296406442a232babf5b

packages/cli/test/dev/fixtures/26-nextjs-secrets/vercel.json
ijjk: Updated in https://github.com/vercel/vercel/pull/7822/commits/97dfc46080b76096d95ee8cef3917a5b1af0777a

TooTallNate (approved):

styfle (approved):
Thanks!

PR #7824, opened 2022-05-19
[cli] Fix "now-dev-next" and "now-dev-static-build-routing" unit tests

TooTallNate:
These two tests have been problematic on MacOS for a long time, but now that we've re-introduced `@vercel/next` into this repo, they've started to fail on Linux/Windows as well.

So the idea is to use the `devCommand` property so that `@vercel/next` Builder doesn't get invoked at all.

Since the `devCommand` property was not yet being unit tested, some logic in `vc build` needed to be adjusted in order to properly shut down the dev command (via `tree-kill` module) when the process was being stopped in order to cleanly shut down.

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -89,6 +89,7 @@ import {
 } from './types';
 import { ProjectEnvVariable, ProjectSettings } from '../../types';
 import exposeSystemEnvs from './expose-system-envs';
+import { treeKill } from '../tree-kill';
 
 const frontendRuntimeSet = new Set(
   frameworkList.map(f => f.useRuntime?.use || '@vercel/static-build')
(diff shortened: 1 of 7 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -2164,17 +2148,14 @@ export default class DevServer {
       process.stdout.write(data.replace(proxyPort, devPort));
     });
 
-    p.on('exit', (code: number) => {
-      if (code > 0) {
-        process.exit(code);
```
AndyBitz, line 2169: Why do we not need to exit the main process anymore?

AndyBitz (approved):

styfle (commented):
Looks like CI is failing with the following:

> Error! Unknown file type: Prerender

https://github.com/vercel/vercel/runs/6504726075?check_suite_focus=true#step:12:6724

TooTallNate:
@styfle This PR is fixing the unit tests specifically. The Dev tests are being addressed in #7822.

styfle (approved):

ijjk (approved):

Issue #1, opened 2016-09-09
Logging in should revoke existing token if any


rauchg:
If we're logging out to log in under a different account, we should revoke the existing token.

leo:
[Here](https://github.com/zeit/now-app/blob/master/src/main/actions/logout.js)'s an example of how we're revoking the token in Now.app.

codyzu:
Looks like Now.app revokes the token by calling:
**DELETE:** `https://zeit.co/api/www/user/tokens/{tokenId}`

What is the api call for now? Will this work?
**DELETE:** `https://api.zeit.co/now/registration/{token}`

@leo can you confirm on the server side what the api call should be?

codyzu:
Looking through old issues... @leo any update on this?

leo:
Yep, that's how it should be done. Not sure if he's talking about deleting the token from `~/.now.json` or about revoking it through the API. Can you test that, @codyzu? If the former is already the case, then the latter needs to follow.

codyzu:
@leo ok I'll try to take a look tomorrow.

PR #7826, opened 2022-05-19
[build-utils] Add getProjectPaths that will evaluate if there is a monorepo

chloetedder:
Screen Shot 2022-05-19 at 3 53 55 PM

If there is no workspace detected, we still want to check if there are potential projects in the monorepo. An example of this case would be:

Screen Shot 2022-05-19 at 10 29 12 AM

The directories `backend/app-three`, `frontend/app-one`, `frontend/app-two` all have definitions within their `package.json` files that define a framework assosiated with the project. This is what distinguishes them from the `package.json` files within `backend` and `frontend`

`getProjectPaths` loops through a depth of 3 (this can be changed but in most cases of this it won't be more than 3). For each directory it recursively calls itself until it finds a `package.json` file. It then checks if the directory has a framework defined and if so it adds the path to the final return value of all projects.

If any `skipPaths` are sent into the function, no projects will be looked for within those path(s). This would include the case for a hybrid monorepo where a workspace has already been detected

### Related Issues

> Fixes #1
> Related to #2

#### Tests

#### Code Review


packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,63 @@
...
+  path?: string;
+  depth?: number;
+}
+
+export type ProjectPath = string;
+
+export const getProjectPaths = async ({
+  fs,
+  path,
+  depth = 0,
```
agadzik: As a public function, I think `depth` feels out of place.  As a consumer of this function I dont want to worry about depth, or allowing consumers to pass in depth arbitrarily

chloetedder: I agree, removed it as a param

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
+import { detectFramework } from './detect-framework';
+import { DetectorFilesystem } from './detectors/filesystem';
+import frameworks from '@vercel/frameworks';
+
+const MAX_DEPTH_TRAVERSE = 3;
```
styfle, line 5: How/why was 3 chosen?

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+  skipPaths?: string[];
+}
+
+export type ProjectPath = string;
+
+export const getProjectPaths = async ({
+  fs,
+  path,
+  skipPaths,
+}: GetProjectPathsOptions): Promise<ProjectPath[] | any> => {
```
styfle: Can we use a more specific type than `any`?

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+  const allPaths: Array<ProjectPath> = [];
+  for (
+    let currentDepth = 0;
+    currentDepth < MAX_DEPTH_TRAVERSE;
+    currentDepth++
+  ) {
+    if (path && skipPaths?.includes(path)) {
+      return allPaths;
+    } else {
```
styfle: You can remove `else` here since the early return on line 27 handles it

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+      const directoryContents = await fs.readdir(path ?? './');
+
+      // This should only ever return one file bc there should never be more than one package.json in a directory
+      const childFiles = directoryContents.filter(
```
styfle:
```suggestion
      const hasPackageJson = directoryContents.some(
```

You can use a boolean instead and it will exit early instead of traversing all files.

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+          childDirs.flatMap(async current => {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            const paths = await getProjectPaths({
+              fs,
+              path: current.path,
+            });
+            return paths;
+          })
+        );
+        dirPaths.flatMap(path => path).forEach(path => allPaths.push(path));
```
styfle: Isn't this a no-op? `dirPaths.flatMap(path => path)`

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+    let currentDepth = 0;
+    currentDepth < MAX_DEPTH_TRAVERSE;
+    currentDepth++
+  ) {
+    if (path && skipPaths?.includes(path)) {
+      return allPaths;
+    } else {
+      const directoryContents = await fs.readdir(path ?? './');
+
+      // This should only ever return one file bc there should never be more than one package.json in a directory
```
EndangeredMassa: Should we add a check for `childFiles.length > 1` and throw an explicit error, just in case?

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      // Scenario 3
+      if (!childDirs.length && !childFiles.length) {
+        return allPaths;
+      }
+
+      if (childDirs.length) {
+        const dirPaths = await Promise.all(
+          childDirs.flatMap(async current => {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
EndangeredMassa: `paths` seems to be used. Does this `eslint-disable` need to be here for this file to lint successfully?

packages/build-utils/src/get-project-paths.ts
chloetedder: 3 was chosen because Jared mentioned how the most common non workspace monorepo would be structured like: 
repo-name
    folder-one
        app-one
        app-two
    folder-two
        app-three

packages/build-utils/src/get-project-paths.ts
chloetedder: Let me try removing it but it is actually for the recursive call is what the eslint was for

packages/build-utils/src/get-project-paths.ts
chloetedder: Changing this check to just see if there is a package.json

packages/build-utils/src/get-project-paths.ts
chloetedder: Removed

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+describe.each<[ProjectPath[], string, ProjectPath[]]>([
+  [
+    ['backend/app-three', 'frontend/app-one', 'frontend/app-two'],
+    '30-monorepo-no-workspaces',
+    [],
+  ],
+  [[], '31-no-monorepo', []],
+  [[], '32-monorepo-highly-nested', []],
+  [['backend/app-three'], '33-hybrid-monorepo', ['frontend']],
```
nathanhammond: Can you duplicate this test to also run without the `skipPaths`?

packages/build-utils/src/get-project-paths.ts
nathanhammond: To some degree the decision that we make here actually informs the decisions that people make about project layout. I don't suspect people are particularly dogmatic about these things.

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,63 @@
...
+  const allPaths: Array<ProjectPath> = [];
+  for (
+    let currentDepth = 0;
+    currentDepth < MAX_DEPTH_TRAVERSE;
+    currentDepth++
+  ) {
+    if (path && skipPaths?.includes(path)) {
+      return allPaths;
+    }
+    const directoryContents = await fs.readdir(path ?? './');
```
nathanhammond: At the leaf level this `readdir` call is unnecessary; we only need to know if a folder contains `package.json` and we can check that location directly. Effectively we can check N+1 levels deep while only traversing N. This could matter a lot for performance in some repositories.

nathanhammond (changes requested):
This requires special-casing but I think it's worth it.

packages/build-utils/src/get-project-paths.ts
chloetedder: Changed it to just `flat` because potentially could be an array of arrays

packages/build-utils/src/get-project-paths.ts
chloetedder: I am understanding half of this. But how can you check a location directly for `package.json` of child directories without getting the contents of the directory you are in?

chloetedder: If i already have all the directories then I can do:
But I would still need to go through for the children directories unless those are already give too

How do I get all the paths?
`directoryPaths.forEach(path => {
  const hasPackageJson = fs.isFile(path);
  if (hasPackageJson) {
    allPaths.push(path)
  }
})`

nathanhammond: I'm sorry, let me try again. I believe this is how the current implementation works:
```
traverse                         <= first call to `readdir`
├── 1                            <= second call to `readdir 
│   └── 2                        <= third call to `readdir`
│       └── package.json         <= found in third call to `readdir`
└── a                            <= seen in first call to `readdir`, fourth call to `readdir`
    ├── b                        <= fifth call to `readdir`
    │   └── c                    <= `readdir` not called for depth reasons
    │       └── package.json     <= not found
    ├── beta                     <= sixth call to `readdir`
    ├── boxer                    <= seventh call to `readdir`
    │   └── package.json         <= found in seventh call to `readdir`
    └── brawl                    <= eighth call to `readdir`
        └── package.json         <= found in eighth call to `readdir`
```

I think we can get away with this:
```
traverse                         <= first call to `readdir`
├── 1                            <= second call to `readdir 
│   └── 2                        <= found folder in second call to `readdir`
│       └── package.json         <= fs.exists("2/package.json")
└── a                            <= seen in first call to `readdir`, third call to `readdir`
    ├── b                        <= found folder in third call to `readdir`
    │   └── c                    <= !fs.exists("b/package.json")
    │       └── package.json     <= not found
    ├── beta                     <= !fs.exists("beta/package.json")
    ├── boxer                    <= found folder in third call to `readdir`
    │   └── package.json         <= fs.exists("boxer/package.json")
    └── brawl                    <= found folder in third call to `readdir`
        └── package.json         <= fs.exists("boxer/package.json")
```

By directly checking `fs.exists` at the leaf nodes we can cut down on the readdir calls which are much more expensive (and the deeper we go the greater the odds we stumble into absolutely massive directories).

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,63 @@
...
+  fs,
+  path,
+  skipPaths,
+}: GetProjectPathsOptions): Promise<ProjectPath[]> => {
+  const allPaths: Array<ProjectPath> = [];
+  for (
+    let currentDepth = 0;
+    currentDepth < MAX_DEPTH_TRAVERSE;
+    currentDepth++
+  ) {
```
nathanhammond: This appears to be unused?

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,63 @@
...
+    const childDirs = directoryContents.filter(stat => stat.type === 'dir');
+
+    if (!childDirs.length && !hasPackageJson) {
+      return allPaths;
+    }
+
+    if (childDirs.length) {
+      const dirPaths = await Promise.all(
+        childDirs.flatMap(async current => {
+          const paths = await getProjectPaths({
```
nathanhammond: You're halfway between iteration and recursion here, right now landing in the recursion to full depth.

nathanhammond (changes requested):
I am guessing that this started with an iterative approach and ended up with a recursive approach. Right now I believe it's starting a recursive walk of all directories three times.

packages/build-utils/src/get-project-paths.ts
chloetedder: I updated the function where it matches the structure not excessively calling `readdir`

packages/build-utils/src/get-project-paths.ts
chloetedder: I changed this to depth being sent in because it was looping through all directories

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,30 @@
+import path from 'path';
+import { getProjectPaths, ProjectPath } from '../src/get-project-paths';
+import { FixtureFilesystem } from './utils/fixture-filesystem';
+
+describe.each<[ProjectPath[], string, ProjectPath[]]>([
```
nathanhammond: Can you add a test that stubs `fs.readDir` and `fs.hasPath` to get a count of the number of calls? It's straightforward to look at a file tree and figure out the number of times a function should be called, but it's a bit harder for me to do so reasoning my way through this code.

This will also make sure that we don't actually regress in a end-reult compatible way in some future update.

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,78 @@
...
+  const allPaths: Array<ProjectPath> = [];
+  if (path && skipPaths?.includes(path)) {
+    return allPaths;
+  }
+  const directoryContents = await fs.readdir(path ?? './');
+
+  if (depth < MAX_DEPTH_TRAVERSE) {
+    const hasPackageJson = directoryContents.some(
+      dir => dir.type === 'file' && dir.name === 'package.json'
+    );
```
nathanhammond: I instinctively believe that this is duplicating work done in the previous recurse, on lines 52-59, but it is hard to be certain.

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        } for ${fixturePath}`
+      : `should not detect any path for ${fixturePath}`;
+
+  it(testName, async () => {
+    const fixture = path.join(__dirname, 'fixtures', fixturePath);
+    const fs = new FixtureFilesystem(fixture);
+
+    const actualPaths = await getProjectPaths({ fs, skipPaths });
+
+    expect(actualPaths).toEqual(expect.arrayContaining(paths));
```
nathanhammond: `arrayContaining` here is not a good proxy for what we're trying to do—it won't catch duplicates, which I hypothesize are present. I'd much prefer a `===` string comparison of `.sort().join("\u0000")` (since the null byte is invalid in path names on every system).

nathanhammond: (If this test passes after this change I *think* we're okay on my other comment about possible duplicate work.)

nathanhammond (changes requested):
One more thing about tracking how much work we do and locking it down in tests!

packages/build-utils/src/get-project-paths.ts
chloetedder: Simplified this

nathanhammond (approved):
The tests constrain the behavior we want to pay attention to; this is good to go! Thanks for all of the work @chloetedder!

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  [['backend/app-three'], 2, 3, '33-hybrid-monorepo', ['frontend']],
+  [
+    ['backend/app-three', 'frontend/app-one', 'frontend/app-two'],
+    3,
+    6,
+    '34-monorepo-no-workspaces',
+  ],
+  [[], 1, 1, '35-no-monorepo'],
+  [['frontend/app-two'], 2, 4, '36-monorepo-some-nested'],
+])(
```
styfle, line 43: Can we swap the order so the test fixture name is first?

Similar to this file:

https://github.com/vercel/vercel/blob/main/packages/build-utils/test/unit.detect-workspace-managers.test.ts#L6

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,42 @@
+import path from 'path';
+import { normalizePath } from '../src';
+import { getProjectPaths, ProjectPath } from '../src/get-project-paths';
+import { FixtureFilesystem } from './utils/fixture-filesystem';
+
+describe.each<[ProjectPath[], number, number, string, ProjectPath[]?]>([
```
styfle: Perhaps this should use the object notation instead of array so that each parameter can be named

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,75 @@
...
+        stat => stat.type === 'dir' && !skipPaths?.includes(stat.path)
+      ),
+      ...topDir,
+    ];
+
+    if (!childDirs.length) {
+      return allPaths;
+    }
+
+    if (childDirs.length) {
```
styfle: You can remove this `if` statement since it will always be true (see line 41)

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,75 @@
...
+  const allPaths: Array<ProjectPath> = [];
+  if (path && skipPaths?.includes(path)) {
+    return allPaths;
+  }
+
+  if (depth < MAX_DEPTH_TRAVERSE) {
+    const directoryContents = await fs.readdir(path ?? './');
+
+    const topDir = !path
+      ? [{ path: path ?? './', type: 'dir', name: path ?? './' }]
```
styfle: I see `path ?? './'` 3 times. Can this be stored as a variable? Perhaps `topPath = path ?? './'`?

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,75 @@
...
+          if (hasPackageJsonDir && current.path) {
+            const framework = await detectFramework({
+              fs: fs.chdir(current.path),
+              frameworkList: frameworks,
+            });
+            if (framework) {
+              return current.path;
+            }
+          }
+          if (current.path !== './') {
```
styfle: Do we need to do any normalization before this comparison? For example `.` and `./` are equivalent.

packages/build-utils/src/get-project-paths.ts
```diff
@@ -0,0 +1,75 @@
...
+  path,
+  skipPaths,
+  depth = 1,
+}: GetProjectPathsOptions): Promise<ProjectPath[]> => {
+  const allPaths: Array<ProjectPath> = [];
+  if (path && skipPaths?.includes(path)) {
+    return allPaths;
+  }
+
+  if (depth < MAX_DEPTH_TRAVERSE) {
```
styfle: Lets invert this to follow the same pattern from getWorkspaces which starts at 3 and works towards a base case of 0.

https://github.com/vercel/vercel/blob/86886e6b6005c84bc39afd20fa2603da03fe74a3/packages/build-utils/src/workspaces/get-workspaces.ts#L22-L27

packages/build-utils/src/get-project-paths.ts
chloetedder: I added in `isRootPath` check to cover if `.` or `./` comes in

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    resultPaths: [],
+    readdirCalls: 1,
+    hasPathCalls: 1,
+  },
+  {
+    fixturePath: '36-monorepo-some-nested',
+    resultPaths: ['frontend/app-two'],
+    readdirCalls: 2,
+    hasPathCalls: 4,
+  },
```
styfle, line 42: Let's add a test fixture that gets one level deep. All of these tests are 0 levels or 2 levels as far as I can see

packages/build-utils/test/unit.get-project-paths.test.ts
```diff
@@ -0,0 +1,75 @@
...
+describe.each<{
+  fixturePath: string;
+  resultPaths: ProjectPath[];
+  skipPaths?: ProjectPath[];
+  readdirCalls: number;
+  hasPathCalls: number;
+}>([
+  {
+    fixturePath: '32-monorepo-highly-nested',
+    resultPaths: [],
```
styfle, line 14: Lets update this fixture so it returns something. That way we confirm it can find 3 levels deep but not 4

styfle (approved):
Great, thanks!

styfle (commented):
How about changing `non-deployed-fixtures` to `unit-fixtures`?

PR #7832, opened 2022-05-19
[frameworks] Remove `defaultHeaders` / `defaultRedirects` / `defaultRewrites` / `getFsOutputDir()`

TooTallNate:
These properties are leftover from the previous iteration of `vc build` / File System API v2, and are no longer in use.


styfle (approved):
Great, thanks!

EndangeredMassa (approved):

PR #7839, opened 2022-05-20
[next] Use `substring()` instead of `substr()`

TooTallNate:
Fixes linting warnings, i.e.:

> packages/next/src/server-build.ts#L858
>
> substr() is deprecated, use slice() or substring() instead


ijjk (approved):

styfle (approved):
I prefer `.slice()` but `.substring()` will work in most cases too 😃

EndangeredMassa (approved):

PR #7842, opened 2022-05-20
[cli] Refactor vc build `installBuilders()` logic to not use a loop

TooTallNate:
This will prevent the possibility of an infinite loop occurring when importing builders that throw a `MODULE_NOT_FOUND` error. Instead, we'll just try to import the builders, then run yarn if necessary, and then try to import the builders a second time when needed. If that fails then we'll throw the original error.


EndangeredMassa (approved):

packages/cli/src/util/build/import-builders.ts
```diff
@@ -30,168 +34,192 @@ export async function importBuilders(
...
+
+    importResult = await resolveBuilders(
+      buildersDir,
+      builderSpecs,
+      output,
+      installResult.resolvedSpecs
+    );
+
+    if ('buildersToAdd' in importResult) {
+      throw new Error('Something went wrong!');
```
styfle, line 56: Can we improve this error message?

PR #7843, opened 2022-05-20
[cli] Hard-code `VERCEL: 1` env var for "vc build"

TooTallNate:
This env var should normally come from the `.env` file that is created from `vc pull`, but let's just be extra sure and hard-code it directly in to `vc build` command as well.

SvelteKit `adapter-vercel` depends on this env var being set for its detection logic.


styfle (approved):
I wonder if we need this to vary based on the this checkbox https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables

PR #7844, opened 2022-05-20
[cli] Symlink `@now/build-utils` when installing builders in "vc build"

TooTallNate:
To support legacy Builders like `@frontity/now` which still depends on the old `@now/build-utils` package name, create a symlink to `@vercel/build-utils` to make import the package work transparently.

Depends on #7842 being merged first.


packages/cli/test/unit/util/dev/server.test.ts
```diff
@@ -1,438 +0,0 @@
...
-async function runNpmInstall(fixturePath: string) {
-  if (await fs.pathExists(path.join(fixturePath, 'package.json'))) {
-    return execa('yarn', ['install', '--network-timeout', '1000000'], {
-      cwd: fixturePath,
-      shell: true,
-    });
-  }
-}
-
-interface TestFixtureOptions extends Omit<DevServerOptions, 'output'> {
```
styfle: Why was this deleted?

packages/cli/src/util/build/import-builders.ts
```diff
@@ -203,6 +203,11 @@ async function installBuilders(
     cwd: buildersDir,
   });
 
+  // Symlink `@now/build-utils` -> `@vercel/build-utils` to support legacy Builders
+  const nowScopePath = join(buildersDir, 'node_modules/@now');
```
styfle, line 207:
```suggestion
  const nowScopePath = join(buildersDir, 'node_modules', '@now');
```

TooTallNate: There's no difference

packages/cli/test/unit/util/dev/server.test.ts
TooTallNate: Carry-over from #7845. Not included in the diff here anymore since that PR got merged.

styfle (approved):

PR #7845, opened 2022-05-20
[cli] Convert `DevServer` "Unit" suite tests to "Dev" suite tests

TooTallNate:
The `DevServer` class as it's currently implement is not really suitable for unit testing, since it calls `process.exit()` when shutting down the server. This causes Jest to exit prematurely, and prevents tests after these one from running. This was in fact hiding some failing tests. So these tests were ported over to be "Dev" suite tests which spawn a child process of `vc dev` rather than running them in-process, so that the `process.exit()` doesn't exit Jest.


styfle (approved):
Nice find, thanks!

EndangeredMassa (approved):

ijjk (approved):

PR #7847, opened 2022-05-21
[next] Fix middleware order and add e2e test

ijjk:
This ensures we don't change order of middleware during sorting `continue` routes as `redirects` from `next.config.js` should come before middleware. This adds an actual e2e fixture to ensure the middleware behaves as expected once deployed along with a local integration test to ensure the routes order output are correct.  

### Related Issues

x-ref: https://github.com/vercel/edge-functions/issues/93
x-ref: [slack thread](https://vercel.slack.com/archives/C0289CGVAR2/p1652807280232989)

#### Tests

#### Code Review


packages/next/src/utils.ts
```diff
@@ -2265,6 +2265,7 @@ export async function getMiddlewareBundle({
       source.edgeFunctions[edgeFile] = worker.edgeFunction;
       const route = {
         continue: true,
+        override: true,
```
javivelasco, line 2268: What's the purpose of this property exactly??

javivelasco (approved):
Awesome! Thanks JJ!

packages/next/src/utils.ts
ijjk: It overrides the sorting done during merging routes after the build, it was initially added to fix sorting for `beforeFiles` rewrites x-ref: https://github.com/vercel/vercel/blob/5a7851a7f7d4fdf3e99119f059fa34b897736780/packages/routing-utils/src/merge.ts#L40-L44

styfle: I think this will need to take into account the Next.js version so that it doesn't break existing customers.

Something like `override: gte(version, '12.1.6')`

ijjk: It shouldn't break existing customers, I need to investigate this further but I think we may want to move this before the `beforeFiles` rewrites

PR #7853, opened 2022-05-22
[tests] Revert "update CI to use node@14 and drop support for node@12 (#7819)"

TooTallNate:
This reverts commit 89b5aad36775eb897339ad79dc6d9502b7f5e28b.

This PR seemed to have been prematurely merged since there was still failing tests.


ijjk (approved):

ijjk:
We might need to mark individual checks as required instead of the final E2E one to ensure Kodiak doesn't merge with failing ones 🤔

styfle:
Do all checks run every time? Seems best to mark them all as required if we can.

ijjk:
Yeah they should all run every time and then turbo decides to use a cached version or not 

Updated required checks for `main` branch

PR #7854, opened 2022-05-22
[next] Remove unused middleware definition

Schniz:
This is an unused value anyway, so we should just remove it from the codebase.


ijjk (approved):

TooTallNate (approved):

TooTallNate:
We should also get rid of the `BuildResult` type in this package and use the from from `@vercel/build-utils` instead.

PR #7857, opened 2022-05-22
[next][build-utils] Update to re-use build result type from build-utils

ijjk:
### Related Issues

We were previously using a custom type in Next.js for the build result although there is an existing type for this in the build-utils package which we can centralize around. 

x-ref: https://github.com/vercel/vercel/pull/7854#issuecomment-1133934764

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/types.ts
+++ b/packages/build-utils/src/types.ts
@@ -352,11 +352,41 @@ export interface BuilderV3 {
 
 type ImageFormat = 'image/avif' | 'image/webp';
 
+export type RemotePattern = {
+  /**
+   * Must be `http` or `https`.
+   */
+  protocol?: 'http' | 'https';
+
+  /**
+   * Can be literal or wildcard.
+   * Single `*` matches a single subdomain.
+   * Double `**` matches any number of subdomains.
+   */
+  hostname: string;
+
+  /**
+   * Can be literal port such as `8080` or empty string
+   * meaning no port.
+   */
+  port?: string;
+
+  /**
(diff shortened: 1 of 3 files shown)


AndyBitz (approved):

packages/build-utils/src/types.ts
```diff
@@ -352,11 +352,41 @@ export interface BuilderV3 {
...
+};
+
 export interface Images {
   domains: string[];
+  remotePatterns?: RemotePattern[];
   sizes: number[];
   minimumCacheTTL?: number;
   formats?: ImageFormat[];
+  dangerouslyAllowSVG?: boolean;
+  contentSecurityPolicy?: string;
```
TooTallNate, line 389: These should be added to the type in `@vercel/build-utils` instead, and this interface should be able to be removed as well.

ijjk: Are the types listed twice in build-utils?

TooTallNate: I'm not sure what you mean "twice", but we have `Images` defined in build-utils that we should use instead: https://github.com/vercel/vercel/blob/main/packages/build-utils/src/types.ts#L355-L360

ijjk: Isn't that the same type/file?

TooTallNate: Yup. Nvm. I'm blind apparently 🤦

TooTallNate (approved):

PR #7859, opened 2022-05-23
[build-utils] Add getWorkspacePackagePaths function

### LINKED ISSUE #7749: [Monorepo Detection] Add getWorkspacePackagePaths helper function
```ts
/** 
 *  Gets the list of workspace package paths based on the configuration 
 * 
 * @example
 *
 * my-repo
 * |-- packages
 * |    |-- api-1
 * |    |-- api-2
 * |-- package.json
 * |-- pnpm-workspaces.yaml // packages: ["packages/*"]
 *
 * const fs = new ... // based on 'my-repo'
 * const workspace = { ..., type: "pnpm" }
 * getWorkspacePackagePaths({ fs, workspace }) => [
 *   "my-repo/packages/api-1",
 *   "my-repo/packages/api-2"
 * ]
 */
function getWorkspacePackagePaths(fs: DetectorFilesystem, workspace: Workspace): string[]
```

agadzik:
Gets the package paths from the workspace definition for a given git repository

```ts
/** 
 *  Gets the list of workspace package paths based on the configuration 
 * 
 * @example
 *
 * my-repo
 * |-- packages
 * |    |-- api-1
 * |    |-- api-2
 * |-- package.json
 * |-- pnpm-workspaces.yaml // packages: ["packages/*"]
 *
 * const fs = new ... // based on 'my-repo'
 * const workspace = { ..., type: "pnpm" }
 * getWorkspacePackagePaths({ fs, workspace }) => [
 *   "my-repo/packages/api-1",
 *   "my-repo/packages/api-2"
 * ]
 */
function getWorkspacePackagePaths(fs: DetectorFilesystem, workspace: Workspace): string[]
```

### Related Issues

Closes https://github.com/vercel/vercel/issues/7749

#### Tests

#### Code Review


packages/build-utils/src/fs/get-glob-fs.ts
```diff
@@ -0,0 +1,68 @@
...
+import { DetectorFilesystem } from '../detectors/filesystem';
+
+type GlobFs = typeof fs;
+
+export function getGlobFs(_fs: DetectorFilesystem): GlobFs {
+  const readdir = (
+    path: fs.PathLike,
+    callback: (err: NodeJS.ErrnoException | null, files: string[]) => void
+  ): void => {
+    console.log('readdir');
```
EndangeredMassa: Remove the `console.log` calls?

packages/build-utils/src/fs/glob.ts
```diff
@@ -45,6 +45,8 @@ export default async function glob(
 
   const files = await vanillaGlob(pattern, options);
 
+  console.log('glob', files, pattern, options);
```
styfle: Please remove

packages/build-utils/src/fs/glob.ts
```diff
@@ -67,5 +69,7 @@ export default async function glob(
     }
   }
 
+  console.log('glob results', results);
```
styfle: Please remove

packages/build-utils/src/fs/get-glob-fs.ts
```diff
@@ -0,0 +1,73 @@
...
+        });
+      })
+      .catch(err => callback(err, { isSymbolicLink: () => false } as fs.Stats));
+  };
+
+  return {
+    readdir: readdir as typeof fs.readdir,
+    lstat: stat as typeof fs.lstat,
+    stat: stat as typeof fs.stat,
+  } as typeof fs;
```
styfle: `as typeof fs` seems wrong since it doesn't have all methods defined. The caller might try `fs.writeFile()` and it would be undefined at runtime but TS would say its defined at build time

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
+import _path from 'path';
+import yaml from 'js-yaml';
+import glob from 'glob';
+import { DetectorFilesystem } from '../detectors/filesystem';
+import { Workspace } from './get-workspaces';
+import { getGlobFs } from '../fs/get-glob-fs';
+
+const path = _path.posix;
```
styfle: Lets not call this `path` because it could be misleading. How about keeping it as `posix` or `posixPath`?

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
...
+export interface GetWorkspacePackagePathsOptions extends GetPackagePathOptions {
+  fs: DetectorFilesystem;
+  workspace: Workspace;
+}
+
+export async function getWorkspacePackagePaths({
+  fs,
+  workspace,
+}: GetWorkspacePackagePathsOptions): Promise<string[]> {
+  const { type, rootPath: workspaceRootPath } = workspace;
```
styfle: Any reason to use `workspaceRootPath` but not `workspaceType`? Seems like they should both be renamed or neither

packages/build-utils/src/fs/glob.ts
```diff
@@ -8,7 +8,7 @@ import FileFsRef from '../file-fs-ref';
 
 export type GlobOptions = vanillaGlob_.IOptions;
 
-interface FsFiles {
+export interface FsFiles {
```
styfle: Why is this exported? I'm not seeing it used anywhere

packages/build-utils/test/unit.get-workspaces-package-paths.test.ts
```diff
@@ -0,0 +1,38 @@
+import path from 'path';
+import { getWorkspaces } from '../src/workspaces/get-workspaces';
+import { getWorkspacePackagePaths } from '../src/workspaces/get-workspace-package-paths';
+import { FixtureFilesystem } from './utils/fixture-filesystem';
+
+describe.each<[string[], string]>([
+  [['/a', '/b'], '21-npm-workspaces'],
```
styfle: Lets swap the order here so the fixture name is first

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
...
+  ).flat();
+}
+
+async function getPackageJsonWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  const packageJsonAsBuffer = await fs.readFile('package.json');
+  const { workspaces } = JSON.parse(
+    packageJsonAsBuffer.toString()
+  ) as PackageJsonWithWorkspace;
```
styfle, line 91:
```suggestion
  const pkgContent = await fs.readFile('package.json', 'utf8');
  const { workspaces } = JSON.parse(pkgContent) as PackageJsonWithWorkspace;
```

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
...
+      throw new Error(`Unknown workspace implementation: ${type}`);
+  }
+
+  return results.map(packagePath => {
+    return path.join(workspaceRootPath, path.dirname(packagePath));
+  });
+}
+
+type PackageJsonWithWorkspace = {
+  workspaces:
```
styfle:
```suggestion
  workspaces?:
```

This should be optional in the case when its missing

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
...
+    | {
+        packages: string[];
+        noHoist?: string[];
+      }
+    | string[];
+};
+
+type PnpmWorkspaces = {
+  packages: string[];
```
styfle:
```suggestion
  packages?: string[];
```

This should be optional for the case when its missing

packages/build-utils/src/fs/get-glob-fs.ts
agadzik: `glob` only ever uses these three functions.  If we were to remove the casting, how would we define `writeFile` for example?

Something like this? 

```ts
writeFile: (...) => throw new Error("not implemented")
```

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,112 @@
...
+  packages: string[],
+  fs: DetectorFilesystem
+): Promise<string[]> {
+  return (
+    await Promise.all(
+      packages.map(
+        packageGlob =>
+          new Promise<string[]>((resolve, reject) => {
+            glob(
+              path.join(packageGlob, 'package.json').replace(/\\/g, '/'),
```
styfle: You might be able to use the existing `normalizePath()` function from build-utils here

packages/build-utils/src/fs/get-glob-fs.ts
styfle: Yeah that might be good so at least the line number will tell us what when wrong if glob ever changes its implementation. Not sure how many methods need to be added here 🤔

packages/build-utils/src/fs/get-glob-fs.ts
```diff
@@ -0,0 +1,73 @@
...
+import { DetectorFilesystem } from '../detectors/filesystem';
+
+type GlobFs = typeof fs;
+
+function normalizePath(path: string) {
+  // on windows, this will return a path like
+  // D:/c/package.json
+  // since we abstract the filesystem, we need to remove windows specific info from the path
+  // and let the FS decide how to process the path
+  return path.replace(/^[a-zA-Z]:/, '');
```
styfle, line 12: Do we need to account for backslash? `\`

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
agadzik: The `readFile` function of `DetectorFilesystem` only supports a `name` as a argument, and returns back a Promised `Buffer`


```ts
protected abstract _readFile(name: string): Promise<Buffer>;
```

packages/build-utils/src/fs/get-glob-fs.ts
agadzik: No because we call `normalizePath` from build-utils when invoking `glob`. Just looking to remove the drive letter from the path in the case of win32

`D:/c/package.json` => `/c/package.json`

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
styfle: Seems fine as long as `glob` or other places calling `readFile()` dont expect to be able to use the options param

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
chloetedder: changed to just use `type, rootPath`

packages/build-utils/src/fs/get-glob-fs.ts
chloetedder: I tried adding the above `writeFile: (...) => throw new Error("not implemented")` but was getting some wild errors. Going to try a couple other ways

chloetedder:
```
type GlobFs = {
  readdir: typeof fs.readdir,
  lstat: typeof fs.lstat,
  stat: typeof fs.stat,
};
```

What if we change the return type for `getGlobFs` to the above because those are the only 3 functions we care about

chloetedder: Actually that wouldn't work for where we utilize `getGlobFs`
```return {
    ...fs,
    readdir: readdir as typeof fs.readdir,
    lstat: stat as typeof fs.lstat,
    stat: stat as typeof fs.stat,
  };
```

Do we not want to return any of the other properties from `fs`? There are like ~82 besides `writeFile` we would have to manually add which seems also not right to do

agadzik: > Actually that wouldn't work for where we utilize `getGlobFs`
> 
> ```
>     ...fs,
>     readdir: readdir as typeof fs.readdir,
>     lstat: stat as typeof fs.lstat,
>     stat: stat as typeof fs.stat,
>   };
> ```
> 
> Do we not want to return any of the other properties from `fs`? There are like ~82 besides `writeFile` we would have to manually add which seems also not right to do

This would definitely prevent errors, but might make `glob` return unexpected results.

styfle: Lets use a Proxy here to catch unexpected method calls

agadzik: @styfle recommended we use `Proxy` instead 🙌 

```ts
return new Proxy(fs, {
    get(_target, prop) {
      switch (prop) {
        case 'readdir':
          return readdir;
        case 'lstat':
        case 'stat':
          return stat;
        default:
          throw new Error('Not Implemented');
      }
    },
  });
```

packages/build-utils/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  return results.map(packagePath => {
+    return posixPath.join(rootPath, posixPath.dirname(packagePath));
+  });
+}
+
+type PackageJsonWithWorkspace = {
+  workspaces?:
+    | {
+        packages: string[];
```
styfle:
```suggestion
        packages?: string[];
```

packages/build-utils/src/fs/get-glob-fs.ts
```diff
@@ -0,0 +1,82 @@
...
+          isBlockDevice: () => false,
+          isCharacterDevice: () => false,
+          isDirectory: () => !isPathAFile,
+          isFIFO: () => false,
+          isFile: () => isPathAFile,
+          isSocket: () => false,
+          isSymbolicLink: () => false,
+        });
+      })
+      .catch(err => callback(err, { isSymbolicLink: () => false } as fs.Stats));
```
styfle: I thought that error cases should return `null` for the second param in the callback

packages/build-utils/test/unit.get-workspaces-package-paths.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const fs = new FixtureFilesystem(fixture);
+
+    const workspaces = await getWorkspaces({ fs });
+    const actualPackagePaths = (
+      await Promise.all(
+        workspaces.map(workspace => getWorkspacePackagePaths({ fs, workspace }))
+      )
+    ).flat();
+
+    expect(actualPackagePaths).toEqual(expect.arrayContaining(packagePaths));
```
styfle:
```suggestion
    expect(actualPackagePaths).toEqual(packagePaths);
```

Lets ensure an exact match

styfle (approved):
Thanks!

PR #7861, opened 2022-05-23
[tests] update CI to use node@14 and drop support for node@12

EndangeredMassa:
Node 12 is EOL and we're starting to reference packages that don't install on node lower than 14, such as nuxt@3 and remix. Update to 14 for now.

---

- Reimplements: https://github.com/vercel/vercel/pull/7819
- Reverts: vercel/vercel#7853


EndangeredMassa:
This ended up being implemented elsewhere.

PR #7864, opened 2022-05-23
[tests] improve error message when a deploy fails

EndangeredMassa:
When a deploy fails in a test, you typically only see the `State of https://${deploymentUrl} is ERROR` message, not what caused the deploy to fail.

This PR adds the source error message directly to the one that's logged out, removing a step when debugging.


styfle (approved):
Great, thanks! 🎉

PR #7865, opened 2022-05-23
[cli][client][tests] update to node14-compatible target (drop support for node12)

EndangeredMassa:
Node 12 is EOL and we're starting to reference packages that don't install on node lower than 14, such as nuxt@3 and remix. Let's update to 14 for now.

---

Reimplements: https://github.com/vercel/vercel/pull/7819


styfle (approved):
Thanks!

TooTallNate:
Unfortunately Jest is doing something different with the dynamic requires that makes the Unit tests not fail here, but with the built CLI I am getting this with `vc build`:

```
> [debug] [2022-05-25T19:00:07.084Z] Failed to import "@vercel/static-build": Error: Cannot find module '@vercel/static-build/package.json'
> [debug] [2022-05-25T19:00:07.085Z] Installing Builders: @vercel/static-build
yarn add v1.22.18
warning ../../../package.json: No license field
info No lockfile found.
[1/4] 🔍  Resolving packages...
...
Error! Importing "@vercel/static-build": Cannot find module '/Users/nrajlich/r/.vercel/builders/node_modules/@vercel/static-build/dist/index'
```

TooTallNate:
We probably need to do the `eval('require')` stuff in `import-builders.ts` as well.

packages/next/test/fixtures/00-i18n-support/index.test.js
```diff
@@ -29,9 +31,16 @@ async function checkForChange(url, initialValue, hardError) {
 const ctx = {};
 
 describe(`${__dirname.split(path.sep).pop()}`, () => {
-  it('should deploy and pass probe checks', async () => {
+  beforeAll(async () => {
```
styfle: This might be worth moving to a separate PR: seems like a bug when the tests were written?

packages/next/src/templated-launcher.ts
```diff
@@ -9,9 +9,9 @@ if (!process.env.NODE_ENV) {
   process.env.NODE_ENV = region === 'dev1' ? 'development' : 'production';
 }
 
+// NOTE: `eval('require')` is necessary to avoid bad transpilation to `__webpack_require__`
```
TooTallNate: Are we sure this one is necessary? I don't believe this file is compiled by webpack.

TooTallNate: @ijjk could maybe confirm.

packages/next/src/utils.ts
```diff
@@ -244,8 +244,8 @@ export async function getRoutesManifest(
     });
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const routesManifest: RoutesManifest = require(pathRoutesManifest);
+  // NOTE: `eval('require')` is necessary to avoid bad transpilation to `__webpack_require__`
+  const routesManifest: RoutesManifest = eval('require')(pathRoutesManifest);
```
TooTallNate: I'd suggest `fs.readJSON()` instead.

packages/next/src/templated-launcher.ts
ijjk: Yeah this one shouldn't be touched by `ncc` just `tsc` x-ref: https://github.com/vercel/vercel/blob/0b56caba45c4ab0a74e7ae0a451cda033db2db04/packages/next/build.js#L9

TooTallNate:
I pushed a failing CLI test for `vc build` (from #7876) to this branch to make sure that the test is passing before we merge this.

packages/next/test/fixtures/00-i18n-support/index.test.js
TooTallNate: It's not a bug, it's just that the first test creates the deployment and the other tests perform some assertions on the behavior of the deployment.

This `index.test.js` file is afaik copy/pasted into all the other `next/test/fixtures` directories so if we're making a change here then it should probably be applied to all of them.

EndangeredMassa:
Holding off on merging until we get a new minor release out. Then this can be in a new major release.

packages/next/test/fixtures/00-i18n-support/index.test.js
EndangeredMassa: I would consider this a bug because it relies on test order to execute properly.

I reverted the change here and will make a change in all of these files in a separate PR.

AndyBitz (approved):

packages/next/src/templated-launcher.ts
```diff
@@ -10,7 +10,6 @@ if (!process.env.NODE_ENV) {
 }
 
 // @ts-ignore
-// eslint-disable-next-line @typescript-eslint/no-var-requires
 const page = require(__LAUNCHER_PAGE_PATH__);
```
styfle: Should this one use `eval()` too? Why does lint no longer complain?

packages/next/src/utils.ts
```diff
@@ -536,8 +534,9 @@ export async function getImagesManifest(
     return undefined;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const imagesManifest: NextImagesManifest = require(pathImagesManifest);
+  // NOTE: `eval('require')` is necessary to avoid bad transpilation to `__webpack_require__`
+  const imagesManifest: NextImagesManifest =
+    eval('require')(pathImagesManifest);
```
styfle: Lets use `fs.readJSON()` like we do for routes manifest above

packages/next/src/utils.ts
```diff
@@ -2299,8 +2298,8 @@ async function getMiddlewareManifest(
     return;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  return require(middlewareManifestPath);
+  // NOTE: `eval('require')` is necessary to avoid bad transpilation to `__webpack_require__`
+  return eval('require')(middlewareManifestPath);
```
styfle: Lets use `fs.readJSON()` like we do for routes manifest above

packages/node-bridge/tsconfig.json
```diff
@@ -2,11 +2,13 @@
...
     "checkJs": true,
-    "lib": ["esnext"],
+    "lib": ["ES2020"],
     "noEmit": true,
     "noImplicitReturns": true,
     "strict": true,
-    "declaration": true
+    "target": "ES2020",
+    "declaration": true,
+    "module": "ES2020"
```
styfle: I'm not sure this is correct. The `module` should likely be `commonjs` since we even changed TS files to JS in this package to preserve CJS

packages/next/src/templated-launcher.ts
EndangeredMassa: This one doesn't need to be `eval`, according to: https://github.com/vercel/vercel/pull/7865#discussion_r882232084

The linting rule is off for some reason. I enabled it and restored this opt-out line. [`a3c5a3d` (#7865)](https://github.com/vercel/vercel/pull/7865/commits/a3c5a3d37a2474d97fddcd500bb513721de32b26)

EndangeredMassa: Ahh, that's why the rule was off: https://github.com/vercel/vercel/runs/6698194648?check_suite_focus=true

I'll mess with that in a separate PR. Turning this rule back off.

packages/node-bridge/tsconfig.json
EndangeredMassa: Makes sense. I updated it to `commonjs`. [`d905a75` (#7865)](https://github.com/vercel/vercel/pull/7865/commits/d905a759cffded27b6cb4ce8a2ec74a3176284b5)

packages/next/src/templated-launcher.ts
EndangeredMassa: There are 247 violations of this rule. We can discuss if we want to enforce it or not. Maybe for now we remove these comments disabling the rule on a per-line level.

TooTallNate (approved):

PR #7866, opened 2022-05-23
[cli] Lazy-load "metrics" object to fix jest reporting open handle

TooTallNate:
Jest was saying that there's an open handle in the unit tests due to the `crypto.pbkdf2Sync()` call in this metrics file, so lazy-load that object in the case when metrics should not be collected.


packages/cli/src/index.ts
```diff
@@ -690,7 +690,8 @@ const main = async () => {
     if (shouldCollectMetrics) {
       const category = 'Command Invocation';
 
-      metric
+      if (!metric) metric = metrics();
+      metrics()
```
```suggestion
      metric
```

This should reuse the reference that was just created, right?

styfle (commented):
What about something like this?

```ts
// define this once
let metric: ReturnType<typeof metrics> | undefined;
function metricsSingleton() {
  if (!metric) metric = metrics();
  return metric
}

// then use it
metricsSingleton().event(eventCategory, `${exitCode}`, pkg.version).send();
```

styfle (approved):

PR #7869, opened 2022-05-24
[cli] Add initial `vc build` unit tests

TooTallNate:
Adds some initial unit tests for the `vc build` command which includes some fixes for Windows.


packages/cli/src/commands/build.ts
```diff
@@ -176,10 +176,17 @@ export default async function main(client: Client): Promise<number> {
...
+
   // Get a list of source files
-  const files = (await getFiles(workPath, client)).map(f =>
-    relative(workPath, f)
-  );
+  const files = (await getFiles(workPath, client)).map(f => {
+    let rel = relative(workPath, f);
+    if (isWin) {
+      // On Windows, normalize forward slashes to backslashes
+      rel = rel.replace(/\\/g, '/');
```
styfle: We have a shared function for this in `@vercel/build-utils`. Can we reuse the `normalizePath` func here?

TooTallNate: e9bc16831f1c7837d57ad7647f3d5a2193c26a0b

styfle (approved):
Thanks!

PR #7871, opened 2022-05-25
[cli] Add experimental corepack support

styfle:
This PR adds support for experimental [corepack](https://nodejs.org/api/corepack.html) to `vc build`.

Since this is still experimental, we only enable if the env var `ENABLE_EXPERIMENTAL_COREPACK=1` is set.

#### Tests


Ethan-Arrowood (commented):
lgtm! just needs a test

packages/cli/src/commands/build.ts
```diff
@@ -312,6 +313,43 @@ export default async function main(client: Client): Promise<number> {
...
+    } else {
+      console.log(
+        `Detected ENABLE_EXPERIMENTAL_COREPACK=1 and "${pkg.packageManager}" in package.json`
+      );
+      const cacheDir = join(
+        rootPackageJsonPath,
+        '.vercel',
+        'cache',
+        'corepack'
+      );
```
styfle: @TooTallNate Do you have a preference for cache path here?

We'll need to add it to `prepareCache()` and I think we talked about adding `.vercel/cache/**` to always cache.

Thoughts?

TooTallNate: Seems good to me 👍

packages/cli/src/commands/build.ts
```diff
@@ -312,6 +313,43 @@ export default async function main(client: Client): Promise<number> {
...
+      await fs.mkdirp(cacheDir);
+      process.env.COREPACK_HOME = join(cacheDir);
```
TooTallNate: Is `join()` here doing anything?

packages/cli/src/commands/build.ts
```diff
@@ -312,6 +313,43 @@ export default async function main(client: Client): Promise<number> {
...
+      process.env.DEBUG = 'corepack';
```
TooTallNate: Should we set this on the `spawnAsync()` opts instead of globally?

styfle: Globally is much safer which is why I put it in `vc build` instead of build-utils.

Basically, we want everything after this point to use corepack.

However we might need to unset after `vc build` runs.

During my testing it didn't seem like I needed to unset it (perhaps there is a child process for build??)

packages/cli/src/commands/build.ts
styfle: Ok cool, I'll add `.vercel/cache/**` to the default build cache and then we can reuse this for new features going forward 👍

packages/cli/src/commands/build.ts
styfle: Here's an example where pnpm7 is detected in package.json but after `vc build` runs, it reverts to the original/global pnpm6

image

And then here is the same with the cache:

image

styfle: Now that we have #7876 ,  I'll add a similar integration test so we can verify this is working as expected 🔜

styfle: I added a `corepack disable` after the build completes, but I'm worried this is actually worse due to https://github.com/nodejs/corepack/issues/112

I'm starting to think this doesn't belong in `vc build` and should happen in the build container prior to `vc build`

styfle: I realize now that we can't use `corepack disable` without deleting the cache.

So I updated the code to use `corepack enable <pkgmgr> --install-dir <path>` so that we can cache the shim and add it to the PATH manually.

TooTallNate: My point was more about if someone is setting `DEBUG` to something else because they use it in their own build script, then we'd be overwriting it here. Maybe let's just augment the variable if it's already set - you can add multiple namespaces to the `DEBUG` env var separated by colon `:`.

styfle: @TooTallNate Do you know if the `:` delimiter should be `;` on windows? I can't find any docs on that for debug 🤔

TooTallNate: Whoops, sorry, looks like it's comma-delimited `,`. Same on Windows.

styfle: @TooTallNate That explains why I couldn't find `:` or `:` anywhere 🤣 

Ok updated in [b15fb49](https://github.com/vercel/vercel/pull/7871/commits/b15fb49b9ed94cebe7223265291e6213abd4cedf)

packages/cli/src/commands/build.ts
```diff
@@ -366,6 +408,17 @@ export default async function main(client: Client): Promise<number> {
     );
   }
 
+  // Cleanup any globals remaining from corepack
```
TooTallNate: All of this cleanup shouldn't be necessary. We don't spawn and more child processes after this point in `vc build` and env vars don't "persist" after the lifetime of the Node process.

EndangeredMassa: I don't mind keeping things tidy

packages/cli/test/integration.js
```diff
@@ -390,6 +398,93 @@ test('login', async t => {
...
+    ),
+    '7.1.0\n'
+  );
+  // Ensure corepack will be cached
+  t.true(fs.existsSync(path.join(directory, '.vercel/cache/corepack/home')));
+  t.true(fs.existsSync(path.join(directory, '.vercel/cache/corepack/shim')));
+  process.env.ENABLE_EXPERIMENTAL_COREPACK = '1';
+});
+
+test('[vc build] should build project with corepack and select yarn@2.4.3', async t => {
```
EndangeredMassa, line 463: Should this also set the env var? `process.env.ENABLE_EXPERIMENTAL_COREPACK = '1';`

packages/cli/test/integration.js
```diff
@@ -390,6 +398,93 @@ test('login', async t => {
...
+  });
+  const output = await execute(['build'], { cwd: directory });
+  t.is(output.exitCode, 0, formatOutput(output));
+  t.regex(output.stderr, /Build Completed/gm);
+  const after = await _execa('npm', ['--version'], {
+    cwd: directory,
+    reject: false,
+  });
+  // Ensure global npm didn't change
+  t.is(before.stdout, after.stdout);
```
EndangeredMassa, line 416: It looks like `before` and `after` are called without anything changing between. What exactly is being tested here?

packages/cli/src/commands/build.ts
```diff
@@ -312,6 +313,48 @@ export default async function main(client: Client): Promise<number> {
...
+      await fs.mkdirp(corepackHomeDir);
+      await fs.mkdirp(corepackShimDir);
+      process.env.COREPACK_HOME = corepackHomeDir;
+      process.env.PATH = `${corepackShimDir}${delimiter}${process.env.PATH}`;
+      process.env.DEBUG = process.env.DEBUG
+        ? `corepack,${process.env.DEBUG}`
+        : 'corepack';
+      const pkgManagerName = pkg.packageManager.split('@')[0];
+      // We must explicitly enable npm since its not enabled by default
+      // See https://github.com/nodejs/corepack/pull/24
```
EndangeredMassa: Just so I understand, this means that we have to call `corepack enable <package-manager>` so that `npm` will still work. Otherwise, we could have called `corepack enable`, right?

This comment made it sound like we only need to do anything here if `npm` was the package manager, but that's not the case. Can we make the comment clearer for future readers?

packages/cli/src/commands/build.ts
```diff
@@ -312,6 +313,48 @@ export default async function main(client: Client): Promise<number> {
...
+      await spawnAsync(
+        'corepack',
+        ['enable', pkgManagerName, '--install-directory', corepackShimDir],
+        {
+          prettyCommand: `corepack enable ${pkgManagerName}`,
+        }
+      );
+    }
+  }
```
EndangeredMassa: This logic is all specific to corepack. Can we extract it to a named function for easier reading?

The cleanup below (if we keep it) makes this a bit more complex. We could have the extracted function return a cleanup function that can be invoked later. Or it could return the only shared value `corepackShimDir` that gets passed into a separate extracted cleanup function.

I'd find this a lot easier to read if this was extracted.

packages/cli/test/integration.js
```diff
@@ -390,6 +398,93 @@ test('login', async t => {
...
+  // Ensure global yarn didn't change
+  t.is(before.stdout, after.stdout);
+  // Ensure version is correct
+  t.is(
+    await fs.readFile(
+      path.join(directory, '.vercel/output/static/index.txt'),
+      'utf8'
+    ),
+    '2.4.3\n'
+  );
```
EndangeredMassa, line 486: In general, I don't think we should have assertions that have no messages. Can we add some assertion messages (or use other assertions that have messages built in like `t.regex`)?

packages/cli/test/integration.js
styfle: @EndangeredMassa The comment mentions what its checking `// Ensure global npm didn't change`

The `before` runs before `vc build` and the `after` runs after `vc build`

packages/cli/test/integration.js
styfle: This will print "expected: 2.4.3, actual: 1.18.0" automatically on failure so I dont think any additional message is necessary

packages/cli/test/integration.js
EndangeredMassa: Yup, I couldn't see the `build` execution in there for some reason.

packages/cli/test/integration.js
EndangeredMassa: Yeah, this one is a bad example of my general point. The `t.true` calls will not  print anything useful, though.

styfle: Good point! I updated the tests in [72d4b0a](https://github.com/vercel/vercel/pull/7871/commits/72d4b0ad43e680e1466331933dc37e9414c1bad9)

packages/cli/src/commands/build.ts
styfle: I updated the comment in [a36bdec](https://github.com/vercel/vercel/pull/7871/commits/a36bdecaa3a2511ef78cfbfd86d66b6644fae3ac)

packages/cli/src/commands/build.ts
styfle: I refactored to a separate module in [c0fe411](https://github.com/vercel/vercel/pull/7871/commits/c0fe4111c4f1c1ba0d0d0a14283da68663f8f024)

EndangeredMassa (approved):

EndangeredMassa:
Looks good!

packages/cli/src/util/build/corepack.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return null;
+}
+
+export function cleanupCorepack(corepackShimDir: string) {
+  if (process.env.COREPACK_HOME) {
+    delete process.env.COREPACK_HOME;
+  }
+  if (process.env.PATH) {
+    process.env.PATH.replace(`${corepackShimDir}${delimiter}`, '');
```
TooTallNate:
```suggestion
    process.env.PATH = process.env.PATH.replace(`${corepackShimDir}${delimiter}`, '');
```

packages/cli/src/util/build/corepack.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  if (process.env.DEBUG) {
+    process.env.DEBUG.replace('corepack,', '');
```
TooTallNate: This won't get cleaned up when `DEBUG` wasn't previously set (no `,` in the env var in that case). Maybe like this?

```suggestion
    const DEBUG = process.env.DEBUG;
    if (DEBUG === 'corepack') {
      delete process.env.DEBUG;
    } else {
      process.env.DEBUG = DEBUG.replace('corepack,', '');
```

TooTallNate (commented):
Suggested a couple minor fixes.

PR #7878, opened 2022-05-26
[node] Remove `mkdirp-promise` dependency

TooTallNate:
There's no need to use this module since Node.js has `recursive: true` since Node v10.


styfle (approved):
Thanks!

PR #7894, opened 2022-05-31
[examples] Add Vercel Analytics for CRA/SvelteKit templates.

leerob:
Successfully tested with:

- [CRA](https://front-git-analytics-for-all-frameworks.vercel.sh/now-examples/create-react-app/analytics?device=desktop&timePeriod=1d&percentile=75&timeInterval=3h&sort=datapoints&search=)
- [SvelteKit](https://front-git-analytics-for-all-frameworks.vercel.sh/now-examples/sveltekit/analytics?device=desktop&timePeriod=1d&percentile=75&timeInterval=3h&sort=datapoints&search=)

Dependencies have also been updated 👍


examples/sveltekit/svelte.config.js
```diff
@@ -1,7 +1,13 @@
 import adapter from '@sveltejs/adapter-auto';
+import preprocess from 'svelte-preprocess';
 
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
+  preprocess: preprocess({
+    replace: [
+      ['import.meta.env.VERCEL_ANALYTICS_ID', JSON.stringify(process.env.VERCEL_ANALYTICS_ID)]
+    ]
+  }),
```
Rich-Harris, line 10: Is the env var hard-coded? If there's a way to make it `VITE_VERCEL_ANALYTICS_ID` instead, Vite can take care of the replacement automatically. (In future we plan to have [better handling of env vars](https://github.com/sveltejs/kit/issues/4296) but that's the current state of play.)

Either way, it would be better to use [`@rollup/plugin-replace`](https://www.npmjs.com/package/@rollup/plugin-replace) here (it'll work as a Vite plugin) so that it applies to all modules, not just Svelte components

examples/sveltekit/src/routes/__layout.svelte
```diff
@@ -1,6 +1,21 @@
...
+
+  if (browser && analyticsId) {
+    page.subscribe(({ url, params }) =>
+      webVitals({
+        path: url.pathname,
+        params,
+        analyticsId
+      })
+    );
+  }
```
Rich-Harris, line 18: It's _possible_ for this layout to get unmounted and remounted, even though it's the root layout, if a page somewhere uses a named layout that doesn't inherit from the root. In that case this subscription would keep getting re-added. We plan to add a universal entry point so that stuff like this doesn't need to be put in the root layout, but in the meantime this would work:

```suggestion
  $: if (browser && analyticsId) {
    webVitals({
      path: $page.url.pathname,
      params: $page.params,
      analyticsId
    })
  }
```

examples/sveltekit/svelte.config.js
Rich-Harris: Actually I think you could just use `define`, no need for `plugin-replace`: https://vitejs.dev/config/#define

Timer (approved):
LGTM for CRA. Leaving Svelte to @Rich-Harris !

examples/sveltekit/svelte.config.js
leerob: Trying to do this: 

```js
import adapter from '@sveltejs/adapter-auto';

/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    adapter: adapter(),

    // Override http methods in the Todo forms
    methodOverride: {
      allowed: ['PATCH', 'DELETE']
    },
    vite: {
      define: {
        'import.meta.env.VERCEL_ANALYTICS_ID': process.env.VERCEL_ANALYTICS_ID
      }
    }
  }
};

export default config;
```

And it seems to be swapping the value correctly, but I get an error

```js


ReferenceError: REDACTED is not defined
--
21:19:44.164 | at file:///vercel/path0/.svelte-kit/output/server/entries/pages/__layout.svelte.js:59:21
21:19:44.165 | at Object.$render (file:///vercel/path0/.svelte-kit/output/server/chunks/index-5d6c164c.js:110:18)
21:19:44.165 | at file:///vercel/path0/.svelte-kit/output/server/index.js:62:97
21:19:44.165 | at $render (file:///vercel/path0/.svelte-kit/output/server/chunks/index-5d6c164c.js:110:18)
21:19:44.165 | at Object.render (file:///vercel/path0/.svelte-kit/output/server/chunks/index-5d6c164c.js:118:20)
21:19:44.165 | at render_response (file:///vercel/path0/.svelte-kit/output/server/index.js:844:29)
21:19:44.165 | at async respond$1 (file:///vercel/path0/.svelte-kit/output/server/index.js:1886:25)
21:19:44.166 | at async resolve (file:///vercel/path0/.svelte-kit/output/server/index.js:2148:105)
21:19:44.166 | at async Object.handle (file:///vercel/path0/.svelte-kit/output/server/chunks/hooks-3ad7760f.js:5:20)
```

Any thoughts? Where REDACTED is the literal analytics ID string value of the env var.

Rich-Harris: try this:

```diff
define: {
-  'import.meta.env.VERCEL_ANALYTICS_ID': process.env.VERCEL_ANALYTICS_ID
+  'import.meta.env.VERCEL_ANALYTICS_ID': JSON.stringify(process.env.VERCEL_ANALYTICS_ID)
}
```

(you could also remove the `import.meta.env.` prefix from both the `define` map and your code, if you're so inclined)

leerob: That worked! https://github.com/vercel/vercel/pull/8002

PR #7898, opened 2022-05-31
Revert "[next] Allow edge api endpoints in Next.js"

ijjk:
Reverts vercel/vercel#7855

We can re-land after we ensure the cases we have found failing are resolved.

PR #7901, opened 2022-06-01
[next] Add regression test for nested middleware

ijjk:
### Related Issues

x-ref: https://github.com/vercel/vercel/pull/7898

#### Tests

#### Code Review

### DIFF
--- a/packages/next/test/fixtures/00-middleware-nested/vercel.json
+++ b/packages/next/test/fixtures/00-middleware-nested/vercel.json
@@ -1,0 +1,16 @@
+{
+  "version": 2,
+  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "probes": [
+    {
+      "path": "/redirect-me",
+      "status": 307,
+      "responseHeaders": {
+        "Location": "/from-next-config/"
+      },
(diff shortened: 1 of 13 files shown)


packages/next/test/fixtures/00-middleware-nested/vercel.json
```diff
@@ -0,0 +1,16 @@
...
+      "path": "/redirect-me",
+      "status": 307,
+      "responseHeaders": {
+        "Location": "/from-next-config/"
+      },
+      "fetchOptions": {
+        "redirect": "manual"
+      }
+    }
+  ]
```
Schniz, line 15: I think I'm missing what is being tested here 😨 
I wonder if we should have a simpler middleware that adds a header or something and then make a probe that tries to use it? Seems like this only probe has nothing to do with the nested middleware

ijjk: This is checking the priority of redirects in `next.config.js` and a redirect from middleware similar to the existing middleware test added in https://github.com/vercel/vercel/pull/7847

Schniz (approved):

PR #7904, opened 2022-06-01
[tests] Remove ava `compileEnhancements` property from package.json

TooTallNate:
VS Code was complaining about this property, so I guess it's not doing anything:

> Property `compileEnhancements` is not allowed.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -30,7 +30,6 @@
     "scripts/preinstall.js"
   ],
   "ava": {
-    "compileEnhancements": false,
     "extensions": [
       "ts"
     ],


EndangeredMassa (approved):
Yeah, looks like a babel config option that we probably don't need.

packages/cli/package.json
```diff
@@ -30,7 +30,6 @@
     "scripts/preinstall.js"
   ],
   "ava": {
-    "compileEnhancements": false,
```
styfle, line 33: Looks like it was supposed to be nested under `babel` https://github.com/avajs/babel#disable-enhancements

styfle (approved):

PR #7905, opened 2022-06-01
[next] Support Edge API Endpoints

Schniz:
### Related Issues

- #7898
- #7855
- I have merged #7901 by @ijjk into this PR, to ensure this does not break

#### Tests

#### Code Review


packages/next/src/index.ts
```diff
@@ -2575,3 +2584,23 @@ export const prepareCache: PrepareCache = async ({
...
+  pagesDir: string;
+  entryPath: string;
+  outputDirectory: string;
+}) {
+  const [pages, middlewareManifest] = await Promise.all([
+    glob('**/!(_middleware).js', params.pagesDir),
+    getMiddlewareManifest(params.entryPath, params.outputDirectory),
+  ]);
+
+  // Edge Functions do not consider as Serverless Functions
```
cramforce, line 2599: Nit: Typo "are not considered"

Schniz:
```suggestion
  // Edge Functions are not considered as Serverless Functions
```

packages/next/src/index.ts
```diff
@@ -2575,3 +2584,23 @@ export const prepareCache: PrepareCache = async ({
...
+  const [pages, middlewareManifest] = await Promise.all([
+    glob('**/!(_middleware).js', params.pagesDir),
+    getMiddlewareManifest(params.entryPath, params.outputDirectory),
+  ]);
+
+  // Edge Functions are not considered as Serverless Functions
+  for (const edgeFunctionFile of Object.keys(
+    middlewareManifest?.functions ?? {}
+  )) {
+    delete pages[edgeFunctionFile.slice(1) + '.js'];
```
ijjk, line 2612: Are we planning to add an error when `runtime: 'edge'` is used with a `getStaticProps` page? Currently it isn't compatible with the `Prerender` type as it expects the `lambda` output so will break for that case.

Schniz: That's a good question, but right now with https://github.com/vercel/next.js/pull/37481 it's only going to be api endpoints and not pages, so `Prerender` is not expected to be emitted for this output.

cc @shuding that mentioned some work related to pages 👀

ijjk: Might be good to add an invariant here to ensure we only delete API pages for now to be safe then

shuding: Although we don't have FS in the edge runtime, we still allow `getStaticProps` together with `runtime: 'edge'`. It's just that it currently behaves like SSR.

shuding: Soon we will have to remove this constraint as we're refactoring the edge SSR implementation to use functions here :)

shuding (approved):

Schniz:
@ijjk can you re-review?

ijjk (approved):

PR #7909, opened 2022-06-01
[build-utils] Make `repoRootPath` non-optional

TooTallNate:
This value is always set in the Vercel build infra, so mark the type as non-optional.


styfle (approved):
Looks good. We just need to make sure that `vc build` is always run _before_ the `cd RootDirectory` so that cwd is correct.

PR #7910, opened 2022-06-01
[cli] Add git metadata to `vc deploy`

MatthewStanciu:
Currently, git metadata is not included on the Vercel dashboard when a project is deployed via the CLI. This PR populates the git metadata.

### Related PRs

This PR is one of a few required to make this feature work:

- https://github.com/vercel/api/pull/12612
- https://github.com/vercel/api/pull/12648
- https://github.com/vercel/front/pull/14147
- https://github.com/vercel/front/pull/14242
- https://github.com/vercel/front/pull/14268

#### Tests

#### Code Review


packages/cli/src/commands/deploy/index.ts
```diff
@@ -411,9 +412,11 @@ export default async (client: Client) => {
   // build `meta`
   const meta = Object.assign(
     {},
+    await createGitMeta(path),
     parseMeta(localConfig.meta),
     parseMeta(argv['--meta'])
   );
+  console.log(meta);
```
TooTallNate: Remove 
```suggestion
```

packages/cli/test/fixtures/unit/create-git-meta/no-origin/config
```diff
@@ -0,0 +1,7 @@
+[core]
+        repositoryformatversion = 0
+        filemode = true
+        bare = false
+        logallrefupdates = true
+        ignorecase = true
+        precomposeunicode = true
```
EndangeredMassa: The tests will still work, but for consistency with real git directories, can we move this into `packages/cli/test/fixtures/unit/create-git-meta/no-origin/git/config`?

packages/cli/test/fixtures/unit/create-git-meta/no-email/config
```diff
@@ -0,0 +1,11 @@
...
+    repositoryformatversion = 0
+    fileMode = false
+    bare = false
+    logallrefupdates = true
+[user]
+    name = TechBug2012
+    email = <>
+[remote "origin"]
+    url = https://github.com/MatthewStanciu/git-test
+    fetch = +refs/heads/*:refs/remotes/origin/*
```
EndangeredMassa: The tests will still work, but for consistency with real git directories, can we move this into `packages/cli/test/fixtures/unit/create-git-meta/no-email/git/config`?

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  });
+
+  /*
+   * 1. Git config is empty or doesn't exist
+   * 2. Git config doesn't contain email
+   * 3. Git config doesn't contain origin url
+   * 4. Last commit doesn't exist; getLastCommit() throws an error
+   * 5. Last commit fetches correctly and git config contains email and origin url
+   */
```
EndangeredMassa: Remove comments?

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expect(data).toBeUndefined();
+  });
+});
+
+describe('parseRepoUrl', () => {
+  it('should parse github https url', () => {
+    const parsedUrl = parseRepoUrl('https://github.com/vercel/vercel.git');
+    expect(parsedUrl?.provider).toEqual('github');
+    expect(parsedUrl?.org).toEqual('vercel');
+    expect(parsedUrl?.repo).toEqual('vercel');
```
EndangeredMassa: As written, if `parseRepoUrl` returned `null`, the error you would see is something like "undefined not equal to github". I like to make sure that the failure messages are as specific as possible to ease debugging.

You can assert that `parsedUrl` is defined first, like `expect(parsedUrl, 'parseRepoUrl()').toBeDefined()`. This will make the failure message mention "parseRepoUrl()" and you'll know that it returned undefined, instead of something to do with the "github" value.

Note: Jest natively doesn't allow custom messages as a second argument to `expect`. This repo has [mattphillips/jest-expect-message](https://github.com/mattphillips/jest-expect-message) installed to allow it.

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expect(parsedUrl?.repo).toEqual('maven-project-example');
+  });
+});
+
+describe('createGitMeta', () => {
+  it('gets git metata from test-github', async () => {
+    const directory = fixture('test-github');
+    try {
+      await fs.rename(join(directory, 'git'), join(directory, '.git'));
+      const data = (await createGitMeta(directory)) as GitHubMeta;
```
EndangeredMassa: Every time we use `as`, we're telling Typescript to trust that we know the true type of something. This will often be true, but we don't have the automated protection of it being true.

In this case, we could check for the type we expect and fail the test if it's something else. This will provide a clearer error message if the test fails. Typescript itself doesn't let you do actual type checking at runtime, but you can do it indirectly.

One way I like to do this in cases like these is to add a property I can use to distinguish each type, like so:

```
export interface GitHubMeta {
  type: 'github';
  // ...
}

export interface GitLabMeta {
  type: 'gitlab';
  // ...
}

export interface BitbucketMeta {
  type: 'bitbucket';
  // ...
}
```

Then we can say:

```
await fs.rename(join(directory, 'git'), join(directory, '.git'));
const data = await createGitMeta(directory);
if (data.type !== 'github') {
  throw new Error(`Expected createGitMeta to return type "github', but received ${data.type}.`);
}
```

Typescript will assume the value of `data` after that block to be `GithubMeta`.

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,149 @@
...
+  GitLabMeta,
+  GitMeta,
+  RepoData,
+} from '../../types';
+
+function getLastCommit(directory: string): Promise<git.Commit> {
+  return new Promise((resolve, reject) => {
+    git.getLastCommit(
+      (err, commit) => {
+        if (err) reject(err);
```
EndangeredMassa: If there's an error here, `reject` will be called, but so will `resolve`. Can we add a `return` to the error case?

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,149 @@
...
+      { dst: directory }
+    );
+  });
+}
+
+export async function getRepoData(configPath: string) {
+  let gitConfig;
+  try {
+    gitConfig = ini.parse(await fs.readFile(configPath, 'utf-8'));
+  } catch (error) {}
```
EndangeredMassa: Can we catch the error and debug log it? 

```
debug(`Error while parsing repo data: ${error.message}`);
```

EndangeredMassa (approved):

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
TooTallNate:
```suggestion
      const data = await createGitMeta(directory);
      assert('githubCommitMessage' in data);
```

EndangeredMassa (approved):

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,155 @@
...
+  output: Output
+): Promise<GitMeta> {
+  const repoData = await getRepoData(join(directory, '.git/config'), output);
+  // If we can't get the repo URL, then don't return any metadata
+  if (!repoData) {
+    return {};
+  }
+  const commit = await getLastCommit(directory);
+
+  if (repoData.provider === 'github') {
```
styfle: I'm not a huge fan of adding this logic to the cli because it means adding new providers requires cli updates. Although it happens rarely enough that it might be okay.

packages/cli/src/util/output/create-output.ts
```diff
@@ -17,7 +17,7 @@ export interface LogOptions extends PrintOptions {
 }
 
 export class Output {
-  private debugEnabled: boolean;
+  debugEnabled: boolean;
```
styfle, line 19: Can we use `isDebugEnabled()` function instead?

.gitattributes
```diff
@@ -8,3 +8,7 @@ packages/*/test/* linguist-vendored
 # Go build fails with Windows line endings.
 *.go text eol=lf
 go.mod text eol=lf
+
+# Mark certain files as "binary" -- not diffable
+**/test/fixtures/**/git/**/* binary
+**/test/fixtures/**/git/**/* linguist-generated
```
styfle, line 14: I'm not sure if we want `linguist-generated` here. We use `linguist-vendored` above to disable stats but `linguist-generated` is going to hide diffs. I think we would want to see diffs when reviewing PRs.

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    expect(
+      client.outputBuffer.includes('Error while parsing repo data'),
+      'Debug message was not found'
+    ).toBeTruthy();
+  });
+});
+
+describe('parseRepoUrl', () => {
+  it('should parse github https url', () => {
+    const parsedUrl = parseRepoUrl('https://github.com/vercel/vercel.git');
```
styfle: Lets add another test for `parseRepoUrl('https://github.com/vercel/vercel')` since I know that github supports this without the `.git` suffix

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,155 @@
...
+  const originUrl = gitConfig['remote "origin"']?.url;
+  if (originUrl) {
+    return parseRepoUrl(originUrl);
+  }
+}
+
+export function parseRepoUrl(originUrl: string): RepoData | null {
+  const isSSH = originUrl.startsWith('git@');
+  // Matches all characters between (// or @) and (.com or .org)
+  const provider = originUrl.match(/(?<=(\/\/|@)).*(?=(.com|.org))/);
```
styfle:
```suggestion
  const provider = originUrl.match(/(?<=(\/\/|@)).*(?=(\.com|\.org))/);
```

Pretty sure these dots need to be escaped.

Lets add a unit test for `https://examplecom/foo` which would match right now but it shouldn't match

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,155 @@
...
+  let org;
+  let repo;
+
+  if (isSSH) {
+    org = originUrl.split(':')[1].split('/')[0];
+    repo = originUrl.split('/')[1].replace('.git', '');
+  } else {
+    // Assume https:// or git://
+    org = originUrl.split('/')[3];
+    repo = originUrl.split('/')[4].replace('.git', '');
```
styfle: Are there scenarios when [3] or [4] are null/undefined? These can be handled by checking the length or better yet, check `typeof string`

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,155 @@
...
+      (err, commit) => {
+        if (err) return reject(err);
+        resolve(commit);
+      },
+      { dst: directory }
+    );
+  });
+}
+
+export async function getRepoData(configPath: string, output: Output) {
```
styfle: It might be good to add return types of these functions so you catch any missing/incorrect returns early. For example, should this return both `null` and `undefined`?

.gitattributes
EndangeredMassa: I added this because the git directory copies in the fixture added a lot of noise in the PR and it's less likely to get detailed review. I'd be fine removing these, though.

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: My thinking is that the purpose of this feature is to populate data to be displayed on the deployments page of the dashboard, and that page specifically supports GitHub, Bitbucket, and GitLab with slightly different metadata for each one. If we wanted to add support for another provider, the dashboard would have to be updated to support the new provider and the CLI would have to be updated anyway.

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: I think this undefined behavior is okay because it's handled by `createGitMeta()`, which doesn't return anything (and therefore the git metadata isn't populated on the dashboard as expected) if the result of `parseRepoUrl()` is null or undefined. I agree though that it's probably better for readability to explicitly handle the error and return null. 👍

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: Thank you for catching this!

.gitattributes
styfle: Maybe I missed it. Where is the "git directory copies in the fixture" happening?

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
styfle: Oh also lets test `vercel/next.js` just to be sure org and repo and assign properly 🙂

.gitattributes
MatthewStanciu: I'm happy to remove this, but @EndangeredMassa and I added this on a call together, so it may be worth first discussing & getting on the same page about the best way to represent the `git` test files.

packages/cli/src/util/output/create-output.ts
MatthewStanciu: I'm not clear on what this would look like—can you help me understand?

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: Thanks for catching this!

packages/cli/src/util/deploy/create-git-meta.ts
styfle: The alternative is to let the deployment API parse the string so the CLI can just pass-through. This means that adding a new provider happens in 1 place: the API

.gitattributes
MatthewStanciu: Each test within the `createGitMeta()` tests starts by renaming `git` to `.git` and ends by renaming `.git` to `git`

packages/cli/src/util/output/create-output.ts
styfle: On second thought, this is only for tests right?

You can update the MockClient to accept options param and pass that through to the Output here

https://github.com/vercel/vercel/blob/b8114b8b39dd6067e68dee49a556dcc3320a31d8/packages/cli/test/mocks/client.ts#L26

packages/cli/src/util/deploy/create-git-meta.ts
styfle: Add a test like this:

```
const parsedUrl = parseRepoUrl('https://github.com/borked');
```

Then you'll see 

```
TypeError: Cannot read properties of undefined (reading 'replace')
```

styfle:
What about the case when there are uncommitted changes in the current directory so the metadata commit sha will be wrong? Do we need to add additional metadata like `gitDirty: 1`?

MatthewStanciu:
Hmm that's true. What's your opinion on the best way to display this on the deployments page?

.gitattributes
styfle: I see. Maybe that pattern should be added to `.gitignore` instead so the generated files are never committed?

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: Ohhh yes you're right

packages/cli/src/util/output/create-output.ts
TooTallNate: There is a mock `Client` object which gets its `output` property reset for each test. However, there's no way to specify whether the test should have debug enabled or not, so mutating the property on the mock `client.output` was the way we said to do it.

An alternative might be to have a proper `MockOutput` subclass, but actually `Output` isn't even a real class at the moment, so I think doing this is fine in this case and we could refactor at some point in the future (I'd _like_ for `Output` to be a proper class for example).

.gitattributes
TooTallNate: We _want_ the git files committed, since the tests operate on them. But Git doesn't like adding `.git` directories to the git repo (it tries to add them as git submodules instead), but in this case we want to commit those files, hence renaming them to something other than `.git`.

TooTallNate:
@MatthewStanciu For the dirty tree scenario, we'd probably want to ask Design for what they'd prefer.

MatthewStanciu:
FYI, just added the logic to determine if there are uncommitted files on your machine & populate `gitDirty` accordingly

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,197 @@
...
+import { Output } from '../output';
+
+export function isDirty(directory: string): Promise<boolean> {
+  return new Promise((resolve, reject) => {
+    process.exec(
+      'git status -s',
+      { cwd: directory },
+      function (err, stdout, stderr) {
+        if (err) return reject(err);
+        if (stderr) return reject(stderr);
```
TooTallNate: We don't want to reject with a string. Wrap this with `new Error()` instead.

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (stdout.length > 0) {
+          resolve(true);
+        }
+        resolve(false);
```
TooTallNate:
```suggestion
        resolve(stdout.trim().length > 0);
```

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,197 @@
...
+  directory: string,
+  output: Output
+): Promise<GitMeta> {
+  const repoData = await getRepoData(join(directory, '.git/config'), output);
+  const dirty = await isDirty(directory);
+  // If we can't get the repo URL, then don't return any metadata
+  if (!repoData) {
+    return {};
+  }
+  const commit = await getLastCommit(directory);
```
TooTallNate:
```suggestion
  // If we can't get the repo URL, then don't return any metadata
  if (!repoData) {
    return {};
  }
  const [commit, dirty] = await Promise.all([
    getLastCommit(directory),
    isDirty(directory)
  ]);
```

TooTallNate (changes requested):

packages/cli/src/util/deploy/create-git-meta.ts
TooTallNate:
```suggestion
        if (stderr) return reject(new Error(`Failed to determine if git repo has been modified: ${stderr.trim()}`));
```

packages/cli/src/util/deploy/create-git-meta.ts
MatthewStanciu: Oh duh

packages/cli/src/types.ts
```diff
@@ -296,6 +296,48 @@ export interface Token {
...
+}
+
+export interface GitLabMeta {
+  gitDirty?: string;
+  gitlabCommitAuthorName?: string;
+  gitlabCommitMessage?: string;
+  gitlabCommitRef?: string;
+  gitlabCommitSha?: string;
+  gitlabDeployment?: string;
+  gitlabProjectPath?: string;
```
styfle: Missing `gitlabProjectName` and `gitlabProjectNamespace`

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,201 @@
+import fs from 'fs-extra';
+import { join } from 'path';
+import ini from 'ini';
+import git from 'git-last-commit';
+import process from 'child_process';
```
styfle: Lets use { exec }

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -0,0 +1,201 @@
...
+    data.gitDirty = '1';
+  }
+
+  data.githubOrg = repoData.org;
+  data.githubCommitOrg = repoData.org;
+  data.githubRepo = repoData.repo;
+  data.githubCommitRepo = repoData.repo;
+
+  data.githubCommitAuthorName = commit.author.name;
+  data.githubCommitMessage = commit.subject;
```
styfle: We need to truncate the commit message to 1024 chars

```
.substring(0, 1024)
```

.gitattributes
styfle: Ok I understand now. This looks good since the diff is still there in the PR 

image

Its just collapsed by default 
image


Makes sense! I think the comment needs to be updated because "not diffable" doesn't seem true.

MatthewStanciu: Updated the comment!

styfle (approved):
Great work, thanks!

TooTallNate (approved):

mataspetrikas:
hey @MatthewStanciu, I have noticed that the metadata gets picked up when running `vercel deploy` locally, but not when I run the same command in the GitHub action started with `actions/checkout@v3`. 

Can the git metadata be read by the cli in the GitHub action or is it impossible?

Apologies for a probably misplaced PR comment 🙏

styfle:
@mataspetrikas If your project is OSS, I can take a look. 

Otherwise, please [submit a support ticket](https://vercel.com/help#issues) with the relevant deployment url in question, thanks!

PR #7911, opened 2022-06-02
[cli] update `vc env pull --help`

EndangeredMassa:
Updated `vercel pull --help` to make it clearer what it does and when to use `vercel env pull`.

### Related Issues

> Related to https://github.com/vercel/front/pull/14123

---

`vercel pull --help`:

Screen Shot 2022-06-02 at 4 08 45 PM

---

`vercel env --help`:

Screen Shot 2022-06-02 at 4 18 13 PM

### DIFF
--- a/packages/cli/src/commands/env/index.ts
+++ b/packages/cli/src/commands/env/index.ts
@@ -46,7 +46,7 @@ const help = () => {
     '–'
   )} Pull all Development Environment Variables down from the cloud
 
-      ${chalk.cyan(`$ ${getPkgName()} env pull`)}
+      ${chalk.cyan(`$ ${getPkgName()} env pull <file>`)}
       ${chalk.cyan(`$ ${getPkgName()} env pull .env.development.local`)}
 
   ${chalk.gray('–')} Add a new variable to multiple Environments
--- a/packages/cli/src/commands/env/index.ts
+++ b/packages/cli/src/commands/env/index.ts
@@ -46,8 +46,8 @@ const help = () => {
     '–'
   )} Pull all Development Environment Variables down from the cloud
 
-      ${chalk.cyan(`$ ${getPkgName()} env env pull`)}
-      ${chalk.cyan(`$ ${getPkgName()} env env pull .env.development.local`)}
+      ${chalk.cyan(`$ ${getPkgName()} env pull`)}
+      ${chalk.cyan(`$ ${getPkgName()} env pull .env.development.local`)}
 
   ${chalk.gray('–')} Add a new variable to multiple Environments
 
--- a/packages/cli/src/commands/env/index.ts
+++ b/packages/cli/src/commands/env/index.ts
@@ -42,6 +42,13 @@ const help = () => {
 
   ${chalk.dim('Examples:')}
 
+  ${chalk.gray(
+    '–'
+  )} Pull all Development Environment Variables down from the cloud
+
+      ${chalk.cyan(`$ ${getPkgName()} env env pull`)}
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/env/index.ts
```diff
@@ -42,6 +42,13 @@ const help = () => {
 
   ${chalk.dim('Examples:')}
 
+  ${chalk.gray(
+    '–'
+  )} Pull all Development Environment Variables down from the cloud
+
+      ${chalk.cyan(`$ ${getPkgName()} env env pull`)}
+      ${chalk.cyan(`$ ${getPkgName()} env env pull .env.development.local`)}
```
styfle: This looks like a bug: `env env`?

styfle: I'm also no sure we want `vc env pull` to be used the same way as `vc pull`. Is there a reason to document this here?

packages/cli/src/commands/pull.ts
```diff
@@ -42,18 +42,29 @@ const help = () => {
...
     )}
+
+  ${chalk.gray(
+    '–'
+  )} Pull the latest Environment Variables to a specific file. (Deprecated)
+    ${chalk.gray(
+      'If you want to download environment variables to a specific file, use `vercel env pull` instead.'
+    )}
+
+    ${chalk.cyan(`$ ${getPkgName()} pull --env .env.local`)}
```
styfle: I tried `vc pull --env .env.local` on Vercel CLI 25 and it doesn't create the `.env.local`

styfle (changes requested):
We should make sure the commands work that are being documented

packages/cli/src/commands/pull.ts
EndangeredMassa: You're right--It doesn't work. I saw the reference `--env` in the command logic and assumed it was still actually used. I'll check that more explicitly next time.

I removed this bit of documentation and the reference to `--env`.

packages/cli/src/commands/env/index.ts
EndangeredMassa: I noticed it was not mentioned here, but it was in the docs: https://vercel.com/docs/cli#commands/env

With the changes made to remove `vc pull --env some-file`, are we good here, or do you still think there's some unnecessary documentation?

packages/cli/src/commands/pull.ts
```diff
@@ -42,25 +42,29 @@ const help = () => {
...
+  ${chalk.gray(
+    'If you want to download environment variables to a specific file, use `vercel env pull` instead.'
+  )}
 `);
 };
 
 function processArgs(client: Client) {
   return getArgs(client.argv.slice(2), {
     '--yes': Boolean,
-    '--env': String, // deprecated
```
EndangeredMassa, line 63: This isn't used anymore.

EndangeredMassa:
Updated command ouput:

Screen Shot 2022-06-03 at 1 40 48 PM


Screen Shot 2022-06-03 at 1 36 18 PM

styfle (approved):

packages/cli/src/commands/env/index.ts
```diff
@@ -42,6 +42,13 @@ const help = () => {
 
   ${chalk.dim('Examples:')}
 
+  ${chalk.gray(
+    '–'
+  )} Pull all Development Environment Variables down from the cloud
+
+      ${chalk.cyan(`$ ${getPkgName()} env pull`)}
```
styfle:
```suggestion
      ${chalk.cyan(`$ ${getPkgName()} env pull <file>`)}
```

This follows the pattern of the other examples below

PR #7912, opened 2022-06-03
[next] Ensure middleware route comes before beforeFiles rewrites

ijjk:
This ensures `beforeFiles` rewrites come after middleware as there can be a conflict with redirects in middleware if they aren't expecting rewrites to modify the original path. 

### Related Issues

x-ref: https://github.com/vercel/vercel/pull/7847
x-ref: [slack thread](https://github.com/vercel/vercel/pull/7847)

#### Tests

#### Code Review

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -54,6 +54,7 @@ import prettyBytes from 'pretty-bytes';
 
 // related PR: https://github.com/vercel/next.js/pull/30046
 const CORRECT_NOT_FOUND_ROUTES_VERSION = 'v12.0.1';
+const CORRECT_MIDDLEWARE_ORDER_VERSION = 'v12.1.7-canary.29';
 
 export async function serverBuild({
   dynamicPages,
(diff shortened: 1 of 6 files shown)


packages/next/test/fixtures/00-middleware/middleware.js
```diff
@@ -47,6 +47,11 @@ export function middleware(request) {
     return;
   }
 
+  if (url.pathname === '/somewhere') {
```
styfle, line 50: How about a test to request the `/somewhere` path?

packages/next/src/server-build.ts
```diff
@@ -1025,6 +1025,8 @@ export async function serverBuild({
 
       ...redirects,
 
+      ...middleware.staticRoutes,
```
EndangeredMassa: Should we add a comment here about why this needs to come before `beforeFilesRewrites`?

ijjk: Updated in https://github.com/vercel/vercel/pull/7912/commits/658bbb8187c5db20ed39549272661ab8df6bfa56

packages/next/test/fixtures/00-middleware/middleware.js
ijjk: Added in https://github.com/vercel/vercel/pull/7912/commits/658bbb8187c5db20ed39549272661ab8df6bfa56

styfle (approved):
Thanks!

EndangeredMassa (approved):

PR #7922, opened 2022-06-06
[static-config] Use "runtime" instead of "use"

TooTallNate:
The static configuration property name has been decided to be `runtime`.

This is a precursor for #7877.


EndangeredMassa (approved):

styfle (approved):
Sounds good to me 👍 

I also added other stakeholders as reviewers since I'm not sure who decided on the naming here

PR #7924, opened 2022-06-06
[cli] Output `EdgeFunctions` from version 3 Builders in `vc build`

TooTallNate:
Serializes `EdgeFunction` outputs in `vc build` when provided by a version 3 Builder.


packages/cli/test/unit/commands/build.test.ts
```diff
@@ -171,4 +171,71 @@ describe('build', () => {
...
+      ).toEqual(false);
+
+      // "functions/api" directory has output Functions
+      const functions = await fs.readdir(join(output, 'functions/api'));
+      expect(functions.sort()).toEqual(['edge.func']);
+
+      const vcConfig = await fs.readJSON(
+        join(output, 'functions/api/edge.func/.vc-config.json')
+      );
+      console.log(vcConfig);
```
styfle: The log here can probably be removed

styfle (approved):

packages/cli/test/unit/commands/build.test.ts
TooTallNate: Whoops, good catch.

```suggestion
```

PR #7927, opened 2022-06-06
[cli][static-build] Fix case when `vc dev` is throwing the wrong error w/ BOAv3

styfle:
Fixes an issue brought up in https://github.com/vercel/og-image/issues/207

The problem is that we allow a corner case with `vc dev` to run the build script, but this can break if the user ran `vc build` first due to the BOA V3 detection.

The workaround is to delete the BOA V3 directory when there is no Development Command provided since that will run the Build Command during development.

We also add a nicer error in the case when someone actually intended to emit BOA V3.


packages/static-build/test/build.test.ts
```diff
@@ -145,5 +150,32 @@ describe('build()', () => {
         `Detected Build Output v3 from the "build" script, but this Deployment is not using \`vercel build\`.\nPlease set the \`ENABLE_VC_BUILD=1\` environment variable.`
       );
     });
+
+    it('should throw an Error without `vercel dev` is used with `@vercel/static-build`', async () => {
```
AndyBitz:
```suggestion
    it('should throw an Error when `vercel dev` is used with `@vercel/static-build`', async () => {
```

AndyBitz (approved):

TooTallNate (approved):

packages/cli/src/commands/dev/dev.ts
```diff
@@ -104,6 +106,13 @@ export default async function dev(
     devCommand = process.env.VERCEL_DEV_COMMAND;
   }
 
+  if (!devCommand) {
+    // If there is no Development Command, we must delete the
+    // v3 Build Output because it will incorrectly be detected by
+    // `@vercel/static-build` when it executes BuildOutputV3.getBuildOutputDirectory()
+    await fs.remove(OUTPUT_DIR);
```
EndangeredMassa: Can we log out that this directory is being removed? It may save a user from some confusion.

EndangeredMassa (approved):

packages/cli/src/commands/dev/dev.ts
styfle: Add in 9587966 👍

PR #7928, opened 2022-06-07
[node] Add zero-config support for root-level `middleware.js`/`middleware.ts` file

TooTallNate:
Adds support for `config.middleware` property in `@vercel/node` to output an Edge Middleware (by including a catch-all route in the output).

Also updates the zero-config detection logic to schedule a build of root-level `middleware.js`/`middleware.ts` files with `@vercel/node` with this middleware option enabled.


packages/build-utils/src/detect-builders.ts
```diff
@@ -355,7 +355,16 @@ function maybeGetApiBuilder(
   apiMatches: Builder[],
   options: Options
 ) {
-  if (!fileName.startsWith('api/')) {
+  const middleware =
+    fileName === 'middleware.js' || fileName === 'middleware.ts';
+
+  // Root-level Middleware file is handled by `@vercel/next`, so don't
+  // schedule a separate Builder when "nextjs" framwork is selected
```
Ethan-Arrowood:
```suggestion
  // schedule a separate Builder when "nextjs" framework is selected
```

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -78,18 +78,64 @@ describe('Test `detectBuilders`', () => {
...
-    expect(builders![1].src).toBe('!{api/**,package.json}');
+    expect(builders![1].src).toBe('!{api/**,package.json,middleware.[jt]s}');
     expect(builders!.length).toBe(2);
     expect(errors).toBe(null);
   });
 
+  it('no package.json + no build + root-level "middleware.js"', async () => {
+    const files = ['middleware.js', 'index.html', 'web/middleware.js'];
+    const { builders, errors } = await detectBuilders(files);
+    console.log(builders);
```
Ethan-Arrowood: do you mean to have console.log statements in the tests?

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -78,18 +78,64 @@ describe('Test `detectBuilders`', () => {
...
     expect(builders![0].src).toBe('api/users.js');
     expect(builders![1].use).toBe('@vercel/static');
-    expect(builders![1].src).toBe('!{api/**,package.json}');
+    expect(builders![1].src).toBe('!{api/**,package.json,middleware.[jt]s}');
     expect(builders!.length).toBe(2);
     expect(errors).toBe(null);
   });
 
+  it('no package.json + no build + root-level "middleware.js"', async () => {
+    const files = ['middleware.js', 'index.html', 'web/middleware.js'];
```
Ethan-Arrowood: do these tests need the `web/middleware.js` file?

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: Nope.

```suggestion
```

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: To make sure it's only the root-level that gets a Builder match.

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -78,18 +78,64 @@ describe('Test `detectBuilders`', () => {
...
+    expect(builders![1].use).toBe('@vercel/static');
+    expect(builders![1].src).toBe('!{api/**,package.json,middleware.[jt]s}');
+    expect(builders!.length).toBe(2);
+    expect(errors).toBe(null);
+  });
+
+  it('no package.json + no build + root-level "middleware.ts"', async () => {
+    const files = ['middleware.ts', 'index.html', 'web/middleware.js'];
+    const { builders, errors } = await detectBuilders(files);
+    console.log(builders);
```
TooTallNate:
```suggestion
```

packages/build-utils/src/detect-builders.ts
```diff
@@ -278,7 +278,7 @@ export async function detectBuilders(
       // and package.json can be served as static files
       frontendBuilder = {
         use: '@vercel/static',
-        src: '!{api/**,package.json}',
+        src: '!{api/**,package.json,middleware.[jt]s}',
```
Ethan-Arrowood, line 281: What syntax is this? Glob pattern? I could be wrong, but i thought the right pattern would be `[j|t]s`

TooTallNate: These are matched with `minimatch`, so it's the syntax that that module uses. Essentially it's standard RegExp though:

```
> /middleware\.[jt]s/.test('middleware.js')
true
> /middleware\.[jt]s/.test('middleware.ts')
true
> /middleware\.[jt]s/.test('middleware.xs')
false
```

Ethan-Arrowood (approved):
LGTM

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -78,18 +78,62 @@ describe('Test `detectBuilders`', () => {
...
   });
 
+  it('no package.json + no build + root-level "middleware.js"', async () => {
+    const files = ['middleware.js', 'index.html', 'web/middleware.js'];
+    const { builders, errors } = await detectBuilders(files);
+    expect(builders![0].use).toBe('@vercel/node');
+    expect(builders![0].src).toBe('middleware.js');
+    expect(builders![0].config?.middleware).toEqual(true);
+    expect(builders![1].use).toBe('@vercel/static');
+    expect(builders![1].src).toBe('!{api/**,package.json,middleware.[jt]s}');
```
styfle: Can we add some tests to verify this glob pattern works as expected?

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -78,18 +78,62 @@ describe('Test `detectBuilders`', () => {
...
+    expect(builders![0].use).toBe('@vercel/node');
+    expect(builders![0].src).toBe('middleware.ts');
+    expect(builders![0].config?.middleware).toEqual(true);
+    expect(builders![1].use).toBe('@vercel/static');
+    expect(builders![1].src).toBe('!{api/**,package.json,middleware.[jt]s}');
+    expect(builders!.length).toBe(2);
+    expect(errors).toBe(null);
+  });
+
+  it('should not add middleware builder when "nextjs" framework is selected', async () => {
```
styfle: Lets move all these new tests into the section that uses `const featHandleMiss = true;`

In the future we need delete the other code path because it is no longer used (featHandleMiss is always true)

packages/node/src/index.ts
```diff
@@ -373,36 +373,60 @@ export const build: BuildV3 = async ({
...
+    ? handler.substring(0, handler.length - 3)
+    : handler;
+
+  // Will output an `EdgeFunction` for when `config.middleware = true`
+  // (i.e. for root-level "middleware" file) or if source code contains:
+  // `export const config = { runtime: 'experimental-edge' }`
+  let isEdgeFunction = false;
+
+  // Add a catch-all `route` for Middleware
+  if (config.middleware === true) {
```
styfle, line 390: Where does this config come from? Is this like a feature flag?

packages/node/test/fixtures/62-edge-middleware/middleware.ts
```diff
@@ -0,0 +1,12 @@
+export default (req: Request) => {
+  const url = new URL(req.url);
+  if (url.pathname === '/' || url.pathname.startsWith('/api/')) {
+    // For `index.html` and `/api/edge.js`, pass through
+    const response = new Response();
+    response.headers.set('x-middleware-next', '1');
```
styfle: It would be good to add a header here to confirm that middleware was actually executed (since default behavior is pass-through)

packages/node/src/index.ts
TooTallNate: Added by the zero-config build match.

packages/node/test/fixtures/62-edge-middleware/middleware.ts
TooTallNate: 653bb8f8da4f4fa50c1bc5ec8baf8aeafd43dd95

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: df6d714c1d4d36951d1efea2af88898eba02ec85

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: Added a test using `vc build` in 8e51c9f6f7af409d45189c53f7ffb88678d7eef1

styfle: Thanks! I was thinking some unit tests like `middleware.ts` and `middleware.js` both match but `middleware.jsx` doesnt match

packages/node/src/index.ts
styfle: Oh yeah I missed that 👍

styfle (approved):

styfle (approved):

PR #7934, opened 2022-06-08
[node] Move `@vercel/build-utils` to "dependencies"

### LINKED ISSUE #7931: "Cannot find module '@vercel/build-utils'" when using @vercel/node and TypeScript without skipLibCheck
When using `@vercel/node` and TypeScript, some of its exports are imported from `@vercel/build-utils` which is a dev dependency, and not a production dependency, therefore not installed.

I am required to enable `skipLibCheck` to skip this, but it seems like it should be fixed in the package.

```
$ tsc --noEmit
node_modules/@vercel/node/dist/index.d.ts:1:68 - error TS2307: Cannot find module '@vercel/build-utils' or its corresponding type declarations.

1 import { shouldServe, BuildV3, PrepareCache, StartDevServer } from '@vercel/build-utils';
                                                                     ~~~~~~~~~~~~~~~~~~~~~
```

MichaelDeBoey:
Fixes #7931


styfle (commented):
The `@vercel/build-utils` are a little strange because they are injected into the build by the platform, because builders shouldn't maintain the version (its a platform responsibility), so `devDependencies` is expected.

cc @TooTallNate who has more history here. Perhaps `peerDependencies` is a better alternative

MichaelDeBoey:
@styfle `@vercel/build-utils` is imported in the actual code & `shouldServe` is also exported, so I would expect them in `dependencies`

https://github.com/vercel/vercel/blob/0c7b54edad6adf48505abf2cbec01691b85963bb/packages/node/src/index.ts#L22-L37
https://github.com/vercel/vercel/blob/0c7b54edad6adf48505abf2cbec01691b85963bb/packages/node/src/index.ts#L54

styfle (commented):
The purpose is just for TS types, correct?

So can we solve this by publishing a separate package with only types?

MichaelDeBoey:
@styfle As I said above, the `shouldServe` function is re-exported as well + other functionality from `@vercel/build-utils` are used internally, so this isn't a pure TS related problem.

TooTallNate (approved):

styfle (approved):

MichaelDeBoey:
@styfle Any ETA when this one will be published?

styfle:
This was published yesterday in `@vercel/node@2.2.0`

PR #7943, opened 2022-06-10
[cli] Normalize "src" property in `vc build`

TooTallNate:
This matches the production behavior.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -1,7 +1,7 @@
 import fs from 'fs-extra';
 import chalk from 'chalk';
 import dotenv from 'dotenv';
-import { join, relative } from 'path';
+import { join, normalize, relative } from 'path';
 import {
   detectBuilders,
   normalizePath,
@@ -15,6 +15,7 @@ import {
   BuildResultV2,
   BuildResultV2Typical,
   BuildResultV3,
+  NowBuildError,
 } from '@vercel/build-utils';
 import minimatch from 'minimatch';
 import {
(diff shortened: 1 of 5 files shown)


styfle (approved):
Thanks!

packages/cli/src/commands/build.ts
```diff
@@ -477,15 +478,33 @@ export default async function main(client: Client): Promise<number> {
...
+  if (!build.use) {
+    throw new NowBuildError({
+      code: `invalid_build_specification`,
+      message: 'Field `use` is missing in build specification',
+      link: 'https://vercel.com/docs/configuration#project/builds',
+      action: 'View Documentation',
+    });
+  }
+
+  let src = normalize(build.src || '**');
```
pvdz, line 488: I don't know what "the production behavior" is but this will change a few things that may trigger subtle bugs (or fix them, I can't tell)

- resolve `.` and `..`
- replace slashes with their platform native slash
  - if this is `vc build` could this affect windows users? Does our system expect all slashes to be normalized to linux slashes?
- an empty string results in `.` (could hide bugs?)

As per https://nodejs.org/api/path.html#pathnormalizepath

packages/cli/src/commands/build.ts
```diff
@@ -475,17 +476,33 @@ export default async function main(client: Client): Promise<number> {
 }
 
 function expandBuild(files: string[], build: Builder): Builder[] {
-  if (!build.src) return [];
+  if (!build.use) {
```
pvdz, line 479: This change seems unrelated to the PR :shrug:

pvdz (approved):
I assume that this indeed matches production and doesn't introduce expected changes :)

PR #7945, opened 2022-06-10
[tests] Remove `as any` casting in unit tests

TooTallNate:
Maybe this casting was necessary at some point, but it doesn't seem to be anymore.


styfle (approved):
Probably before we had jest configured correctly 😄

PR #7954, opened 2022-06-13
[node] Fix ESM output from TS source

### LINKED ISSUE #7908: @vercel/node: Support ESM with TS
Add support for ESM in Serverless Functions. AWS Lambda [supports ESM](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/) by having a `package.json` file with `{"type":"module"}`.

This is already possible with Build Output v3 ([SvelteKit](https://github.com/sveltejs/kit/blob/master/packages/adapter-vercel/index.js#L470) and [Astro](https://github.com/withastro/astro/blob/main/packages/integrations/vercel/src/serverless/adapter.ts#L49-L51) use this method), and it could be great to have something similar for standalone `api/` functions.

---

Related issues
- #7540

styfle:
- Depends on https://github.com/vercel/nft/pull/297
- Fixes #7908 
- Fixes #6194 

This PR bumps `@vercel/nft` (for production) and `ts-node` (for development) to fix ESM behavior.

Because Node.js cannot dynamically set [loader hooks](https://nodejs.org/docs/latest-v18.x/api/esm.html#loaders) once a process is started, I had to change the way `ts-node` is registered by using environment variables such as `NODE_OPTIONS`. Most of the logic for TS and ESM was pulled out of the child process and into the parent.


cheapsteak:
I think ts-node's support for ESM is no longer experimental, would you mind taking another look? 🙏

user72356:
asd

user72356:
We need some progress on this!

packages/node/src/index.ts
```diff
@@ -528,20 +531,99 @@ export const startDevServer: StartDevServer = async opts => {
     filename: 'package.json',
   });
   const pkg = pathToPkg ? require_(pathToPkg) : {};
+  const tsNodePath = require_.resolve('ts-node');
+  const esmLoader = join(tsNodePath, '..', '..', 'esm.mjs');
```
Ethan-Arrowood, line 535: I wonder if it would be better to read the `ts-node` package.json `"bin"` or `"exports` object to retrieve this path. In case it was ever to change in the future we'd have to manually figure it out again.

---

Since I still had it up locally, I quickly looked into what it might be and I think it would be `exports["./esm"]`, which is seemingly no better than the existing hardcoded path. This comment can probably be ignored but I'm leaving the thought here in case others have more ideas.

packages/node/src/index.ts
```diff
@@ -528,20 +531,99 @@ export const startDevServer: StartDevServer = async opts => {
...
+  const cjsLoader = join(tsNodePath, '..', '..', 'register');
```
Ethan-Arrowood: Interestingly enough this path may need to be more explicit. Based on the `exports` I'm seeing, its resolving `./register/index.js` automatically. We should be explicit like in the `esmLoader` variable. 
```suggestion
  const cjsLoader = join(tsNodePath, '..', '..', 'register', 'index.js');
```

packages/node/src/index.ts
```diff
@@ -528,20 +531,99 @@ export const startDevServer: StartDevServer = async opts => {
...
+
+    // In prod, `.ts` inputs use TypeScript and
+    // `.js` inputs use Babel to convert ESM to CJS.
+    // In dev, both `.ts` and `.js` inputs use ts-node
+    // without Babel so we must enable `allowJs`.
+    tsConfig.compilerOptions.allowJs = true;
+
+    // In prod, we emit outputs to the filesystem.
+    // In dev, we don't emit because we use ts-node.
+    tsConfig.compilerOptions.noEmit = true;
```
Ethan-Arrowood, line 597: Was worried that `compilerOptions` might not be set and thus this could throw an error, but then figured out thats happening in `fixConfig`. Maybe we can move these two lines into that function so it all happens in one place?

Ethan-Arrowood (approved):
lgtm! I didn't check locally, but are there other tests verifying all the various ways an esm/cjs project can be specified?

Scenarios I can think of off the top of my head are:
- `.cjs`
- `.mjs`
- `.js` and `package.json => module: true`
- and typescript versions of each of the above

packages/node/src/index.ts
styfle: @Ethan-Arrowood `fixConfig()` is used for both prod and dev but these 2 options here are only used for dev.

The comments explain why they are needed for dev.

packages/node/src/index.ts
styfle: I think its fine as-is. It wont change unless we bump `ts-node`. If the case when a newer version of `ts-node` changed the file format, it will cause the tests to fail.

krystof-k:
@styfle is it possible this somehow broke serverless functions? See this [discussion](https://github.com/orgs/vercel/discussions/1225) please.

PR #7957, opened 2022-06-13
[tests] Allow for multiple probes

TooTallNate:
The `probes` in the vercel.json files in tests fixtures weren't allowing both `mustContain` _and_ `responseHeaders` probes due to an `else` block. This PR allows to combine those checks into the same probe test.


ijjk (approved):

styfle (approved):
Looks good but we might want to run all tests again just to make sure it works (probably need to add this file as an input to turbo)

PR #7958, opened 2022-06-13
[build-utils] Return `middlewareRoutes` from `detectBuilders()`

TooTallNate:
Instead of `@vercel/node` being responsible for creating the middleware route, we can have it be part of the zero-config detection logic.

This will have the advantage of `vercel dev` being able to get that route since it runs the `detectBuilders()` function, and also seems like a cleaner separation of concerns since `@vercel/node` will not need be responsible for generating any routes, so the `routes` property is removed from the `BuildResultV3` type once again.


packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -2231,35 +2231,63 @@ describe('Test `detectBuilders` with `featHandleMiss=true`', () => {
 
   it('no package.json + no build + root-level "middleware.js"', async () => {
     const files = ['middleware.js', 'index.html', 'web/middleware.js'];
-    const { builders, errors } = await detectBuilders(files, null, {
-      featHandleMiss,
-    });
-    expect(builders![0].use).toBe('@vercel/node');
-    expect(builders![0].src).toBe('middleware.js');
+    const { builders, middlewareRoutes, errors } = await detectBuilders(
```
styfle: Lets also assert any other routes are what we expect here

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -2231,35 +2231,63 @@ describe('Test `detectBuilders` with `featHandleMiss=true`', () => {
...
-    const files = ['middleware.ts', 'index.html', 'web/middleware.js'];
-    const { builders, defaultRoutes, errors } = await detectBuilders(
+  it('no package.json + no build + root-level "middleware.ts" + API function', async () => {
+    const files = [
+      'api/index.go',
+      'middleware.ts',
+      'index.html',
+      'web/middleware.js',
+    ];
+    const { builders, middlewareRoutes, errors } = await detectBuilders(
```
styfle: Same here, do we expect other api routes for the go function?

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: 0bec2f2dc3efd1346dec9c453af402e718f658b4

packages/build-utils/test/unit.builds-and-routes-detector.test.ts
TooTallNate: 0bec2f2dc3efd1346dec9c453af402e718f658b4

PR #7959, opened 2022-06-13
[routing-utils] Convert tests to TypeScript

TooTallNate:
Converts the routing-utils tests to TypeScript so that they can be run without rebuilding the source code.


ijjk (approved):

styfle (commented):
Looks like CI is failing

PR #7964, opened 2022-06-14
[node] Remove duplicate entry of `@vercel/ncc`

TooTallNate:
Remove from "devDependencies" because it's also listed in "dependencies".


styfle (commented):
Do we also need to consider it external now?

TooTallNate:
Superseded by #7969.

PR #7967, opened 2022-06-14
[tests] Create tarballs of all packages in Vercel deployment

TooTallNate:
For the Vercel deployment, run `yarn pack` in each of the packages in the monorepo and place them in the "public/tarballs" directory so that we can have npm-installable URLs of each package for every commit. The `package.json` of each package is also updated to reference the tarball when a package depends on other packages within the monorepo.

Try it out like:

```
$ npm i -g https://vercel-biww73ffq.vercel.sh/tarballs/vercel.tgz

# Notice how the package.json has the monorepo dependencies
# updated to point to the related tarballs from the same deployment:

$ cat /usr/local/lib/node_modules/vercel/package.json | grep biww
    "@vercel/build-utils": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/build-utils.tgz",
    "@vercel/go": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/go.tgz",
    "@vercel/next": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/next.tgz",
    "@vercel/node": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/node.tgz",
    "@vercel/python": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/python.tgz",
    "@vercel/redwood": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/redwood.tgz",
    "@vercel/remix": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/remix.tgz",
    "@vercel/ruby": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/ruby.tgz",
    "@vercel/static-build": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/static-build.tgz",
    "@vercel/client": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/client.tgz",
    "@vercel/frameworks": "https://vercel-biww73ffq.vercel.sh/tarballs/@vercel/frameworks.tgz",

# Also notice that the "version" in `package.json` gets updated to include the commit
# SHA so that we can easily identify which commit a given tarball was generated from:

$ vercel --version
25.1.1-canary.1-727b290
```

### DIFF
--- a/packages/cli/src/util/pkg.ts
+++ b/packages/cli/src/util/pkg.ts
@@ -1,6 +1,11 @@
-import _pkg from '../../package.json';
+import fs from 'fs';
+import { join } from 'path';
 import { PackageJson } from '@vercel/build-utils';
 
-const pkg: PackageJson & typeof _pkg = _pkg;
+// Path here has only one `..` because the ncc'd file lives in "dist"
+const pkgPath = join(__dirname, '../package.json');
 
+const pkg: PackageJson & typeof import('../../package.json') = JSON.parse(
+  fs.readFileSync(pkgPath, 'utf8')
+);
 export default pkg;
--- a/packages/cli/src/util/pkg.ts
+++ b/packages/cli/src/util/pkg.ts
@@ -2,9 +2,12 @@ import fs from 'fs';
 import { join } from 'path';
 import { PackageJson } from '@vercel/build-utils';
 
-// Path here has only one `..` because the ncc'd file lives in "dist"
-const pkgPath = join(__dirname, '../package.json');
+let rootDir = __dirname;
+while (!fs.existsSync(join(rootDir, 'package.json'))) {
+  rootDir = join(rootDir, '..');
+}
 
+const pkgPath = join(rootDir, 'package.json');
 const pkg: PackageJson & typeof import('../../package.json') = JSON.parse(
   fs.readFileSync(pkgPath, 'utf8')
 );
--- a/utils/pack.ts
+++ b/utils/pack.ts
@@ -1,0 +1,46 @@
+import execa from 'execa';
+import path from 'path';
+import fs from 'fs-extra';
+import { TurboDryRun } from './types';
+
+const rootDir = path.join(__dirname, '..');
+
+async function main() {
+  const { stdout: turboStdout } = await execa(
+    'turbo',
+    ['run', 'build', '--dry=json'],
+    {
+      cwd: rootDir,
+    }
+  );
+  const turboJson: TurboDryRun = JSON.parse(turboStdout);
+  for (const task of turboJson.tasks) {
+    const dir = path.join(rootDir, task.directory);
+    const packageJsonPath = path.join(dir, 'package.json');
+    const originalPackageObj = await fs.readJson(packageJsonPath);
+    const packageObj = await fs.readJson(packageJsonPath);
+    if (task.dependencies.length > 0) {
+      for (const dependency of task.dependencies) {
+        const name = dependency.split('#')[0];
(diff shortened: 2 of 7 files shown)


utils/pack.ts
```diff
@@ -0,0 +1,52 @@
...
+    {
+      cwd: rootDir,
+    }
+  );
+  const turboJson: TurboDryRun = JSON.parse(turboStdout);
+  for (const task of turboJson.tasks) {
+    const dir = path.join(rootDir, task.directory);
+    const packageJsonPath = path.join(dir, 'package.json');
+    const originalPackageObj = await fs.readJson(packageJsonPath);
+    const packageObj = await fs.readJson(packageJsonPath);
```
EndangeredMassa, line 24: Do we want to read this file twice or clone the object? I doubt it's a perf issue to read twice. So, either is fine with me. It just seemed a bit weird on an initial read.

utils/pack.ts
```diff
@@ -0,0 +1,52 @@
...
+        const name = dependency.split('#')[0];
+        const tarballUrl = `https://${process.env.VERCEL_URL}/tarballs/${name}.tgz`;
+        if (packageObj.dependencies && name in packageObj.dependencies) {
+          packageObj.dependencies[name] = tarballUrl;
+        }
+        if (packageObj.devDependencies && name in packageObj.devDependencies) {
+          packageObj.devDependencies[name] = tarballUrl;
+        }
+      }
+    }
```
EndangeredMassa, line 38: Can we extract the logic that produces `packageObj` and write some tests for it?

EndangeredMassa (approved):

packages/cli/src/util/pkg.ts
```diff
@@ -1,6 +1,14 @@
-import _pkg from '../../package.json';
+import fs from 'fs';
+import { join } from 'path';
 import { PackageJson } from '@vercel/build-utils';
 
-const pkg: PackageJson & typeof _pkg = _pkg;
+let rootDir = __dirname;
+while (!fs.existsSync(join(rootDir, 'package.json'))) {
```
styfle, line 6: Any chance this will throw if the user doesn't have root permission?

utils/pack.ts
styfle: Hmm, that does seems strange to read it twice. Although if we need a deep clone, then there isn't a nice way to do that without a 3rd party package (or Node.js 17+ structuredClone)

utils/types.ts
```diff
@@ -0,0 +1,17 @@
+export interface TurboDryRun {
+  packages: Array<string>;
+  tasks: Array<Task>;
+}
+
+export interface Task {
```
styfle, line 6:
```suggestion
export interface TurboTask {
```

utils/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
styfle, line 1: I'm curious why these new files are in `utils` instead of `api/_lib/script`?

Since these new files are only used by the Vercel deployment, I would prefer to keep everything in there

packages/cli/src/util/pkg.ts
TooTallNate: Node.js literally reads the `package.json` file of a module when importing, so I suspect any issues here to be unlikely. I guess we'll see though.

utils/pack.ts
MatthewStanciu: Happy to clone it instead

utils/tsconfig.json
MatthewStanciu: That's a good point, thank you.

PR #7925, opened 2022-06-06
[cli] add support for Edge Functions in `vercel dev`

EndangeredMassa:
Add support for Edge Functions when using `vercel dev`. This uses the [`edge-runtime`](https://edge-runtime.vercel.app/) to emulate the proper environment, [`ncc`](https://github.com/vercel/ncc) to parse the entrypoint (supporting TypeScript and importing node modules), and some custom marshaling to pass the request/response values through.

PR #7969, opened 2022-06-14
[node] Add missing dependencies for `dev-server.ts`

TooTallNate:
Some new imports were introduced recently but are missing from the `"dependencies"` object, so they were missing when the package was installed from the npm registry.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@types/node": "*",
-    "@vercel/ncc": "0.34.0",
+    "@vercel/ncc": "0.24.0",
     "@vercel/node-bridge": "3.0.0",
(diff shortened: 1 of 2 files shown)


styfle (approved):

packages/node/package.json
```diff
@@ -48,14 +52,10 @@
     "@types/node-fetch": "^2.6.1",
     "@types/test-listen": "1.1.0",
     "@vercel/build-utils": "4.1.1-canary.1",
-    "@vercel/ncc": "0.24.0",
```
styfle, line 51: I noticed the version here doesn't match

styfle: I think we should stick with `"@vercel/ncc": "0.24.0"` for now since that was the version before #7925

PR #7973, opened 2022-06-15
[cli] Support root-level Middleware file in `vc dev`

TooTallNate:
Adds initial support for a root-level `middleware.js` / `middleware.ts` file in the `vercel dev` CLI command. This leverages the existing Edge Function invoking logic in `@vercel/node`'s `startDevServer()` function and applies the necessary response / rewrites / mutations to the HTTP request based on the result of the middleware invocation.


packages/cli/src/util/dev/server.ts
```diff
@@ -1430,37 +1406,144 @@ export default class DevServer {
...
+            } else {
+              res.end();
+            }
+            return;
+          }
+
+          if (rewritePath) {
+            // TODO: add validation?
+            debug(`Detected rewrite path from middleware: "${rewritePath}"`);
+            prevUrl = rewritePath;
```
Schniz, line 1538: Does the name `prevUrl` make sense here? It looks to me more `activeUrl` (if this is the value that holds the rewritten URL)

packages/cli/src/util/dev/server.ts
```diff
@@ -1430,37 +1406,144 @@ export default class DevServer {
...
+        startMiddlewareResult =
+          await middleware.builderWithPkg.builder.startDevServer?.({
+            files,
+            entrypoint: middleware.entrypoint,
+            workPath,
+            repoRootPath: this.cwd,
+            config: middleware.config || {},
+            meta: {
+              isDev: true,
+              //requestPath,
```
EndangeredMassa: remove?

packages/cli/src/util/dev/server.ts
```diff
@@ -1430,37 +1406,144 @@ export default class DevServer {
...
+    // Run the middleware file, if present, and apply any
+    // mutations to the incoming request based on the
+    // result of the middleware invocation.
+    const middleware = [...this.buildMatches.values()].find(
+      m => m.config?.middleware === true
+    );
+    if (middleware) {
+      let startMiddlewareResult: StartDevServerResult | undefined;
+      // TODO: can we add some caching to prevent (re-)starting
+      // the middleware server for every HTTP request?
```
EndangeredMassa, line 1418: Should we make a card for this?

packages/cli/src/util/dev/server.ts
```diff
@@ -1430,37 +1406,144 @@ export default class DevServer {
...
+              }
+              res.end(middlewareBody);
+            } else {
+              res.end();
+            }
+            return;
+          }
+
+          if (rewritePath) {
+            // TODO: add validation?
```
EndangeredMassa, line 1536: Should we make a card for this?

packages/cli/src/util/dev/server.ts
```diff
@@ -1430,37 +1406,144 @@ export default class DevServer {
...
-      if (middlewareResult.error) {
-        this.sendError(
+        await this.sendError(
           req,
           res,
           requestId,
-          'EDGE_FUNCTION_INVOCATION_FAILED',
-          500
+          'NO_RESPONSE_FROM_FUNCTION',
+          502
```
EndangeredMassa: Is middleware supposed to respond with a 502? We're making general Edge Functions respond with a 500.

EndangeredMassa (approved):

packages/cli/src/util/dev/server.ts
TooTallNate: Yes

packages/cli/src/util/dev/server.ts
TooTallNate: Yes

EndangeredMassa: @TooTallNate How's this?
https://linear.app/vercel/issue/VCCLI-153/validate-rewritepath-before-using-in-middleware-during-vc-dev

packages/cli/src/util/dev/server.ts
TooTallNate: Ya perhaps we could rename the var. Would prefer to save for a different PR though.

packages/cli/src/util/dev/server.ts
TooTallNate: 0b2ebcb7aafbda5c3f559f236c953ddd0534911b

PR #7975, opened 2022-06-15
Implement suggestions from #7967

MatthewStanciu:
#7967 was tagged for merge before some comments were resolved. This PR is an add-on to #7967 which continues its code review session.

#### Tests

#### Code Review


MatthewStanciu:
FYI, @EndangeredMassa and I tried to add tests but it wasn't as straightforward as we were hoping—ran into issues related to bridging the gap between ESModules and CommonJS modules. We should figure this out, so I created an issue in Linear, but in the meantime we know that this PR works fine, so I would like to defer that to another PR.

api/package.json
```diff
@@ -16,7 +16,7 @@
     "unzip-stream": "0.3.0"
   },
   "devDependencies": {
-    "@types/node": "13.1.4",
+    "@types/node": "17.0.44",
```
styfle, line 19: Technically this is using 16.x at runtime

package.json
```diff
@@ -50,7 +50,7 @@
     "test-integration-dev": "turbo run test-integration-dev",
     "lint": "eslint . --ext .ts,.js",
     "prepare": "husky install",
-    "pack": "cd utils && node -r ts-eager/register ./pack.ts"
+    "pack": "cd ./api/_lib/script && node -r ts-eager/register ./pack.ts"
```
styfle, line 53: Can we remove this script from package.json if we make the "pack" step part of the `_lib/script/build.ts`?

MatthewStanciu: Yes, but these steps were intentionally separated, cc @TooTallNate

styfle: Why where they separated? I feel like we already have a lot of scripts here and we should reduce if we can since its not obvious to new users when to run each

api/_lib/script/pack.ts
```diff
@@ -46,6 +37,20 @@ async function main() {
...
+    const name = dependency.split('#')[0];
+    const tarballUrl = `https://${process.env.VERCEL_URL}/tarballs/${name}.tgz`;
+    if (packageObj.dependencies && name in packageObj.dependencies) {
+      packageObj.dependencies[name] = tarballUrl;
+    }
+    if (packageObj.devDependencies && name in packageObj.devDependencies) {
+      packageObj.devDependencies[name] = tarballUrl;
+    }
+  }
+  return packageObj;
```
styfle, line 51: No need to return since its mutating the object (as the func name says)

api/package.json
MatthewStanciu: Oh should I revert to 13.1.4 or change the version to something in 16.x?

api/_lib/script/pack.ts
MatthewStanciu: Whoops, good catch

MatthewStanciu:
I'm going to close this and hope nobody notices or cares

Issue #800, opened 2017-09-04
For `now ls`, print `http://` prefix to allow cmd+clicking from terminal


vjpr:
To keep the interface cleaner, perhaps you'd like to put this behind a flag, or some kind of global config `~/.now/config.js`...

gkaemmer:
+1. I'm working on this, will open a PR in a few.

PR #7979, opened 2022-06-15
Make links in `vc ls` clickable

MatthewStanciu:
Currently, URLs in the `vc ls` command are not clickable, which makes them a lot less useful to me. It seems like this was intentionally prevented at some point; however, I (admittedly with no context about the history of this command) think this behavior should be enabled by default.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/list.ts
+++ b/packages/cli/src/commands/list.ts
@@ -95,7 +95,7 @@ export default async function main(client: Client) {
   }
 
   const meta = parseMeta(argv['--meta']);
-  const { currentTeam, includeScheme } = config;
+  const { currentTeam } = config;
 
   let contextName = null;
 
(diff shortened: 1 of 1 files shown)


TooTallNate:
FYI here's the historical context: https://github.com/vercel/vercel/issues/800

MatthewStanciu:
Oh man so this would be a breaking change then. Hmm

MatthewStanciu:
I'm going to mark this as a draft for now

TooTallNate:
Should probably also remove from the type definition here: https://github.com/vercel/vercel/blob/e0a62dabaa485fde7284d48e9c941fa653c680af/packages/cli/src/types.ts#L25

styfle:
If we're going to make a breaking change, we might as well remove `vc ls` and put the url in `vc project ls` instead.

I'm pretty sure `vc ls` precedes the concept of "projects" so perhaps its not relevant anymore.

MatthewStanciu:
Closing in favor of #7993, which includes this change

PR #7984, opened 2022-06-16
fix(node): dependencies should be stable

MichaelDeBoey:


TooTallNate:
The dependencies need to remain in lock-step with what the versions of the packages are in this monorepo, and get updated by lerna when a release is done, so thank you but this change isn't necessary.

PR #7985, opened 2022-06-16
chore(node): remove `typescript` dependency

MichaelDeBoey:
`typescript` is only used as a type or as bin when doing a type check.

Since type checks aren't necessary when using normal Node, we can move `typescript` to an optional peerDependency.

---

BREAKING CHANGE: `typescript` isn't a dependency anymore

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -38,8 +38,7 @@
     "edge-runtime": "1.0.1",
     "exit-hook": "2.2.1",
     "node-fetch": "2.6.1",
-    "ts-node": "8.9.1",
-    "typescript": "4.3.4"
+    "ts-node": "8.9.1"
   },
   "devDependencies": {
     "@babel/core": "7.5.0",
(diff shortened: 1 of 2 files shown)


styfle (commented):
It is used at runtime here via dynamic require https://github.com/vercel/vercel/blob/bfef989adae7340a853b02f00ba0a60d42d5bb2f/packages/node/src/typescript.ts#L142

MichaelDeBoey:
@styfle that's correct, but that file is also only used when using TypeScript, right?
If that's correct, this means that people who don't need that functionality (because they're not using TypeScript at all in their project), will still have `typescript` in their `node_modules` for no reason.

styfle:
No, its also used for JS users, not just TS users, during `vc dev`

MichaelDeBoey:
I still think `typescript` should be in `peerDependencies` instead of normal `dependencies` though, as now you're causing a possible double installation of the package by having a strict version in `dependencies` list.

If you look at other big projects that rely on `typescript`, they all have `typescript` in their `peerDependencies` (or have it listed as `'*'`).

styfle:
I believe `peerDependencies` isn't installed automatically (depending on package manager and version) so this would be a breaking change. What we can probably do is use `@vercel/ncc` to vendor typescript instead of using a regular dependency.

EndangeredMassa:
We've been discussing some from of this, but there's no clear decision yet. I'm going to close this for now.

PR #7988, opened 2022-06-17
[cli] Include "argv" in `builds.json` produced by `vc build`

TooTallNate:
It will be useful for debugging purposes to have access to the arguments that were passed to `vc build`.


styfle (approved):
What kind of args are you looking for?

I thought the args would be something like `--scope` or `--cwd` and I don't think we use either of those.

PR #7991, opened 2022-06-17
[cli] Remove beta warning for `vercel build`

leerob:
This removes the `(beta)` output from `vercel build` as we move towards stability.

image


styfle (commented):
Lets remove once we have rolled out to 100%

styfle (approved):
We have reached 100% hobby so I think its safe to remove the beta tag

PR #7995, opened 2022-06-17
[cli] Implement `vc build --output` parameter

TooTallNate:
Use `--output` parameter to output the Build Output API build artifacts to a different location than the default `.vercel/output` directory.


styfle (commented):
How about a test?

styfle (approved):
Thanks!

PR #7997, opened 2022-06-20
[next] allow to declare edge functions outside of /api/ in Next.js

Schniz:
Enable declaring edge functions outside of `/api/`, and therefore
allow Next.js pages to be deployed as an Edge Function.

#### Tests

#### Code Review


shuding (approved):

packages/next/src/index.ts
```diff
@@ -2613,15 +2612,6 @@ async function getServerlessPages(params: {
   for (const edgeFunctionFile of Object.keys(
     middlewareManifest?.functions ?? {}
   )) {
-    // `getStaticProps` are expecting `Prerender` output which is a Serverless function
```
ijjk, line 2616: Do we have an invariant elsewhere when attempting to mark a `getStaticProps` page as an edge function? 

We might want to update this error https://github.com/vercel/vercel/blob/50af9f5b75e518379bbba2b6dcba0330710b0425/packages/next/src/utils.ts#L1883 to check if there is an edge function and mention this is not yet supported.

Schniz: @shuding @huozhi what happens on `getStaticProps` with the preview build of the builder that you tested? does it work? do we need to add this invariant?

huozhi: I built one page with gSP and your runtime build it looks good so far https://empty-73bfip04p-huozhi.vercel.app/

the `world` is from gSP

ijjk: @huozhi it doesn't seem that the above deployment is using `Prerender` outputs which is the part that's incompatible with edge functions currently. Are we not properly detecting `getStaticProps` with the edge runtime in Next.js so they aren't added in the `prerender-manifest`?

huozhi: It can't work like static optimization and output as html, but work similar like `gSSP`. The detection for `runtime` with `getStaticProp` seems correct for enabling edge runtime if `runtime` is specified as `"edge"` and `getStaticProps` is present.

I wonder shall we keep it there without asserting it to just it works as gSSP? but it's bit confusing indeed 🤔

ijjk: I think we should throw an invariant in Next.js if `getStaticProps` is used with runtime edge as it won't actually have the static generation/caching guarantees that it should have. We can mention they can use `getServerSideProps` instead in the mean-time.

huozhi: In that way we may want to error it in next.js as well later, stopping user to use gSP from the beginning. cc @shuding

shuding: Sounds good, gSP doesn't seem like a high priority thing.

ijjk (approved):

PR #8004, opened 2022-06-21
[cli] Only show "Removing `.vercel/output`" log when directory exists in `vc dev`

TooTallNate:
Conditionally show the "Removing…" message, instead of always rendering it.


styfle (approved):
Thanks!

EndangeredMassa (approved):

PR #8007, opened 2022-06-23
[cli][node] make error handling of edge functions consistent with serverless functions in `vc dev`

EndangeredMassa:
When edge functions error, they were showing a basic text response instead of the error template. They were also sending a 502 status code instead of a 500.

This PR makes the error handling of Edge Functions consistent with Serverless Functions when executed through `vc dev`. If we want to update the error templates themselves, we can do that in a separate PR.

**Note:** Production currently treats Edge Function errors differently from Serverless Function errors, but that's a known issue that will be resolved.

---

*I deleted the original outputs (terminal and browser screenshots) because they are out of date and added a lot of content to this page. See my latest comment for updated examples.*


packages/cli/src/util/dev/server.ts
```diff
@@ -1106,6 +1106,7 @@ export default class DevServer {
         view = errorTemplate({
           http_status_code: statusCode,
           http_status_description,
+          error_code,
```
EndangeredMassa, line 1111: This wasn't being passed through, but the template already supported it.

packages/cli/src/util/dev/server.ts
```diff
@@ -2185,13 +2186,7 @@ function proxyPass(
...
       );
       if (!res.headersSent) {
-        devServer.sendError(
-          req,
-          res,
-          requestId,
-          'NO_RESPONSE_FROM_FUNCTION',
-          502
-        );
+        devServer.sendError(req, res, requestId, 'FUNCTION_INVOCATION_FAILED');
```
EndangeredMassa, line 2285: `FUNCTION_INVOCATION_FAILED` is a better message here and it matches production. This should be a 500 (the default for `sendError`), not 502.

packages/node/src/dev-server.ts
```diff
@@ -146,28 +146,35 @@ async function createEdgeEventHandler(
...
-      let response = edgeHandler(event.request, event);
-      return event.respondWith(response);
+        let edgeHandler = module.exports.default;
+        let response = await edgeHandler(event.request, event);
+
+        return event.respondWith(response);
+      } catch (error) {
+        event.respondWith(new Response(error.stack, {
+          status: 500
+        }));
```
EndangeredMassa: We handle errors within the code that runs inside the `edge-runtime` so that we don't trigger `edge-runtime`'s error handling. It was doing some extra logging we couldn't control. This gives us direct control of error handling in the response.

Schniz (commented):
Question: what if the user wants to respond with a 500 status code?

EndangeredMassa:
To see what different kinds of errors look like, based on @Schniz's suggestion, I created a project that has one kind of error per edge function.

I'm going to iterate on these.

---

*I deleted the original outputs (terminal and browser screenshots) because they are out of date and added a lot of content to this page. See my latest comment for updated examples.*

cramforce:
Looking at your screenshots, the edge function doesn't have a meaningful stack trace and filenames/line numbers while the serverless function does.

packages/cli/src/util/dev/server.ts
```diff
@@ -1740,7 +1741,10 @@ export default class DevServer {
             isDev: true,
             requestPath,
             devCacheDir,
-            env: { ...envConfigs.runEnv },
+            env: {
+              ...envConfigs.runEnv,
+              VERCEL_BUILDER_DEBUG: this.output.debugEnabled ? '1' : undefined,
```
EndangeredMassa, line 1839: This allows `debug` function calls (from build-utils) to log or not inside the spawned process

packages/node/src/dev-server.ts
```diff
@@ -136,77 +147,146 @@ async function serializeRequest(message: IncomingMessage) {
...
+): Promise<(request: IncomingMessage) => Promise<VercelProxyResponse>> {
+  const userCode = await compileUserCode(entrypoint);
+  const server = await createEdgeRuntime(userCode);
 
   return async function (request: IncomingMessage) {
+    if (!server) {
+      // this error state is already logged, but we have to wait until here to exit the process
+      // this matches the serverless function bridge launcher's behavior when
+      // an error is thrown in the function
+      process.exit(1);
```
EndangeredMassa, line 250: There's some dancing around being done at a few points in this file to make sure that errors are not thrown during setup and `process.exit` is called during `onDevRequest` instead. This maintains the interface of how the spawned dev server interacts with `startDevServer`.

We can probably refactor this to make it more obvious how these processes interact, but I wanted to see if I could get this to work with the same interface first. The good news is that this works as expected.

packages/node/src/index.ts
```diff
@@ -514,7 +514,7 @@ export const startDevServer: StartDevServer = async opts => {
     // Got "exit" event from child process
     const [exitCode, signal] = result;
     const reason = signal ? `"${signal}" signal` : `exit code ${exitCode}`;
-    throw new Error(`\`node ${entrypoint}\` failed with ${reason}`);
+    throw new Error(`Function \`${entrypoint}\` failed with ${reason}`);
```
EndangeredMassa, line 523: Because the function in question here could be a Serverless Function or an Edge Function, I changed the wording of this error. If something else makes more sense, let me know!

packages/node/src/dev-server.ts
```diff
@@ -136,77 +147,146 @@ async function serializeRequest(message: IncomingMessage) {
...
+          let response = await edgeHandler(event.request, event);
+
+          return event.respondWith(response);
+        } catch (error) {
+          // we can't easily show a meaningful stack trace
+          // so, stick to just the error message for now
+          event.respondWith(new Response(error.message, {
+            status: 500,
+            headers: {
+              'x-handled-from': 'edge-wrapper'
```
EndangeredMassa: This allows us to know that the 500 response comes from our code handling an error in user code instead of the user code intentionally responding with a 500 of their own. There's a test for this if you want more context on what the difference is.

EndangeredMassa:
Updated output examples!

All edge functions currently show the following before any other output. We're going to quiet this in a separate PR.

```
ncc: Version 0.24.0
ncc: Compiling file index.js
ncc: Using typescript@4.7.3 (local user-provided)
```

Sometimes this warning also shows up before other output. We'll quiet these as well.

```
(node:80390) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
(Use `node --trace-deprecation ...` to show where the warning was created)
```

---

### Serverless - Runtime Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ✅ proper error message
- ✅ useful stack trace

```
Unhandled rejection: Error: I broke in a serverless function!
    at serverless (/Users/smassa/source/demo/edge/ember/api/serverless-error.js:2:9)
    at Object.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:813:19)
    at Generator.next (<anonymous>)
    at /Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:4:12)
    at Server.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:795:59)
    at Server.emit (events.js:400:28)
    at parserOnIncoming (_http_server.js:900:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:127:17)
Error! Failed to complete request to /api/serverless-error: Error: socket hang up
```

---

### Edge - Runtime Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ✅ proper error message
- ❌ useful stack trace: I trimmed off the stack trace to at least remove the noise. We'll work on getting useful stack traces for these in a future PR.

```
Unhandled rejection: intentional runtime error
Error! Failed to complete request to /api/edge-error-runtime: Error: socket hang up
```

---

### Edge - Config Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ✅ proper error message
- ✅ useful stack trace: There is no stack trace, but I don't think we need one here. It would be nice to link to a docs or KB page here.

```
Invalid function runtime "bad-runtime-config-value" for "api/edge-error-config.ts". Valid runtimes are: ["experimental-edge"]
Error! Failed to complete request to /api/edge-error-config: Error: socket hang up
```

---

### Edge - Startup Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ✅ proper error message
- ❌ useful stack trace: I trimmed off the stack trace to at least remove the noise. We'll work on getting useful stack traces for these in a future PR.

```
Failed to instantiate edge runtime: intentional startup error
Error! Failed to complete request to /api/edge-error-startup: Error: socket hang up
```

---

### Edge - Syntax Error (500 / FUNCTION_INVOCATION_FAILED)

- ❌ clean output: output is messy with `ncc` messages beyond the compilation message
- ❌ proper error message: the error does say "Module parse failed: Unexpected token", but it's not specific enough
- ❌ useful stack trace: No stack trace available. We'll look into adding one in a future PR.

```
Failed to instantiate edge runtime: Module parse failed: Unexpected token (5:42)
File was processed with these loaders:
 * ../../../vercel/vercel/node_modules/@vercel/ncc/dist/ncc/loaders/empty-loader.js
 * ../../../vercel/vercel/node_modules/@vercel/ncc/dist/ncc/loaders/relocate-loader.js
 * ../../../vercel/vercel/node_modules/@vercel/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
| }
|
> export default async function edge(request: Request, event: Event) {
|   return new Response('some response body');
|
Error! Failed to complete request to /api/edge-error-syntax: Error: socket hang up
```

---

### Edge - Unknown Import Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ❌ proper error message: the error here is confusing. It's happening because `eval('require')` is being used in the `ncc`-generated code, but only when the module is not found. When the module is found, this error doesn't happen. Will investigate in a future PR.
- ❌ useful stack trace: No stack trace available. We'll look into adding one in a future PR.

```
Failed to instantiate edge runtime: Code generation from strings disallowed for this context
Error! Failed to complete request to /api/edge-error-unknown-module: Error: socket hang up
```

---

### Edge - No Default Export Error (500 / FUNCTION_INVOCATION_FAILED)

- ✅ clean output
- ✅ proper error message: the message is good, but it would be better if we added a link to a docs page or KB article.
- ✅ useful stack trace: No stack trace seems necessary here.

```
Unhandled rejection: No default export was found. Add a default export to handle requests.
Error! Failed to complete request to /api/edge-error-no-handler: Error: socket hang up
```

packages/node/src/dev-server.ts
```diff
@@ -136,77 +147,150 @@ async function serializeRequest(message: IncomingMessage) {
...
+            headers: requestDetails.headers,
+            method: requestDetails.method,
+            body: body
+          });
+
+          event.request = request;
+
+          let edgeHandler = module.exports.default;
+          if (!edgeHandler) {
+            throw new Error('No default export was found. Add a default export to handle requests.');
```
EndangeredMassa, line 183: @leerob is there a good link we can use here to point users to a place where they can read up on general edge functions? I know things are shifting around.

packages/node/src/dev-server.ts
```diff
@@ -136,77 +147,150 @@ async function serializeRequest(message: IncomingMessage) {
...
+  entryPointPath: string
+): string | undefined {
   const project = new Project();
-  const staticConfig = getConfig(project, entrypoint);
+  const staticConfig = getConfig(project, entryPointPath);
   const runtime = staticConfig?.runtime;
   if (runtime && !validRuntimes.includes(runtime)) {
-    throw new Error(`Invalid function runtime for "${entrypoint}": ${runtime}`);
+    throw new Error(
+      `Invalid function runtime "${runtime}" for "${entrypoint}". Valid runtimes are: ${JSON.stringify(
```
EndangeredMassa, line 291: @leerob is there a good link we can use here as well?

TooTallNate (changes requested):
We should use `testFixtureStdio` instead of `testFixture` so that we test the behavior against real deployments as well.

packages/node/src/dev-server.ts
leerob: https://vercel.com/docs/concepts/functions is a good, safe link

packages/node/src/dev-server.ts
leerob: Let's use https://vercel.com/docs/concepts/functions for now and swing back once docs URLs are finalized - sound good?

leerob: Maybe just "Learn more: https://vercel.com/docs/concepts/functions"

TooTallNate (approved):

packages/node/src/dev-server.ts
EndangeredMassa: Added in: https://github.com/vercel/vercel/pull/8048

packages/node/src/dev-server.ts
EndangeredMassa: Added in: https://github.com/vercel/vercel/pull/8048

PR #8014, opened 2022-06-25
[cli] Add `vc project connect` command

MatthewStanciu:
This PR adds a new subcommand, `vc project connect`, which connects a Git provider repository to the current project. Previously, this could only be done via the Dashboard.

This is the first part of a larger project—the goal is to include this functionality within `vc link`, so that you never have to leave the CLI if you want to set up a new Vercel project that's connected to Git.

#### Tests

#### Code Review


packages/cli/src/commands/projects.ts
```diff
@@ -103,16 +117,144 @@ export default async (client: Client) => {
...
+  if (subcommand === 'connect') {
+    const yes = Boolean(argv['--yes']);
+    if (args.length !== 0) {
+      console.error(
+        error(
+          `Invalid number of arguments. Usage: ${chalk.cyan(
+            `${getCommandName('projects connect')}`
+          )}`
+        )
+      );
```
TooTallNate:
```suggestion
      output.error(
        `Invalid number of arguments. Usage: ${chalk.cyan(
          `${getCommandName('project connect')}`
        )}`
      );
```

packages/cli/src/commands/projects.ts
```diff
@@ -103,16 +117,144 @@ export default async (client: Client) => {
...
+    let paths = [process.cwd()];
+
+    const validate = await validatePaths(client, paths);
+    if (!validate.valid) {
+      return validate.exitCode;
+    }
+    const { path } = validate;
+
+    const linkedProject = await ensureLink(
+      'projects connect',
```
TooTallNate:
```suggestion
      'project connect',
```

packages/cli/src/commands/projects.ts
```diff
@@ -103,16 +117,144 @@ export default async (client: Client) => {
...
+        `No local git repo found. Run ${chalk.cyan(
+          '`git clone <url>`'
+        )} to clone a remote Git repository first.`
+      );
+      return 1;
+    }
+    const remoteUrl = pluckRemoteUrl(gitConfig);
+    if (!remoteUrl) {
+      output.error(
+        `No remote origin url found in your Git config. Make sure you've connected your local Git repo to a Git provider first.`
```
TooTallNate:
```suggestion
        `No remote origin URL found in your Git config. Make sure you've connected your local Git repo to a Git provider first.`
```

packages/cli/src/commands/projects.ts
```diff
@@ -103,16 +117,144 @@ export default async (client: Client) => {
...
+      );
+      return 1;
+    }
+    const parsedUrl = parseRepoUrl(remoteUrl);
+    if (!parsedUrl) {
+      output.error(
+        `Can't parse Git repo data from the following remote url in your Git config: ${remoteUrl}`
```
TooTallNate:
```suggestion
        `Failed to parse Git repo data from the following remote URL in your Git config: ${remoteUrl}`
```

TooTallNate: Might also want to add formatting on the `remoteUrl` i.e. from `util/output/link.ts`.

packages/cli/src/commands/projects.ts
```diff
@@ -103,16 +117,144 @@ export default async (client: Client) => {
...
+        project.id,
+        provider,
+        repoPath
+      );
+      if (typeof connect === 'number') {
+        return connect;
+      }
+    }
+
+    output.log(`Connected ${chalk.cyan(repoPath)}!`);
```
TooTallNate: Maybe something like "Connected to GitHub repository org/name!"

packages/cli/src/commands/projects.ts
```diff
@@ -271,7 +413,9 @@ async function run({
     return;
   }
 
-  console.error(error('Please specify a valid subcommand: ls | add | rm'));
+  console.error(
+    error('Please specify a valid subcommand: ls | connect | add | rm')
+  );
```
TooTallNate:
```suggestion
  output.error(
    Please specify a valid subcommand: ls | connect | add | rm'
  );
```

packages/cli/src/util/projects/connect-git-provider.ts
```diff
@@ -0,0 +1,101 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        type,
+        repo,
+      }),
+    })
+    .catch(err => {
+      if (err.message.includes('install the GitHub integration')) {
```
TooTallNate: Does the error not contain a `code` or more reliable property that we can check to infer these states?

packages/cli/test/mocks/project.ts
```diff
@@ -131,6 +131,9 @@ export function useProject(project: Partial<Project> = defaultProject) {
   client.scenario.get(`/v8/projects/${project.id}`, (_req, res) => {
     res.json(project);
   });
+  client.scenario.get(`/v8/projects/:nameOrId`, (_req, res) => {
```
TooTallNate: Why is this one necessary? This will make `project` be returned even with a non-matching ID. Seems like maybe the test isn't set up correctly if this is needed.

packages/cli/test/unit/commands/projects.test.ts
```diff
@@ -0,0 +1,266 @@
...
+        process.chdir(cwd);
+        await fs.rename(join(cwd, 'git'), join(cwd, '.git'));
+        useUser();
+        useTeams('team_dummy');
+        useProject({
+          ...defaultProject,
+          id: 'unlink',
+          name: 'unlink',
+        });
+        client.setArgv('projects', 'connect', '--cwd', cwd);
```
TooTallNate: You're already doing `process.chdir()` so these `--cwd` flags shouldn't be necessary (on the tests below as well):

```suggestion
        client.setArgv('projects', 'connect');
```

TooTallNate (commented):
Left a few suggestions but overall looking very nice! Good looking tests 😉

packages/cli/src/util/projects/connect-git-provider.ts
MatthewStanciu: Both of these return 400. The message is the only reliable thing I can think of to differentiate them. Maybe there's something else that I'm missing, though.

TooTallNate: I suggest we look into augmenting the response from the API in that case then. Sniffing error messages is fragile and likely to break in the future if the message is ever changed.

packages/cli/src/util/projects/connect-git-provider.ts
```diff
@@ -0,0 +1,101 @@
...
+        client.output.error(
+          `Failed to link ${chalk.cyan(
+            repo
+          )}. You need to add a Login Connection to your Git provider first. Visit ${chalk.cyan(
+            `https://vercel.com/docs/concepts/personal-accounts/login-connections`
+          )} to learn how to do this.`
+        );
+      } else {
+        client.output.error(
+          `Failed to connect a Git provider repo.\nFetch url: ${fetchUrl}\n${err}`
```
TooTallNate: I don't think we need to log `fetchUrl` here - I don't think it would be useful to the user.

packages/cli/src/util/projects/connect-git-provider.ts
MatthewStanciu: For posterity: Nate and I merged some changes to vercel/api to address this. https://github.com/vercel/api/pull/12991

packages/cli/src/commands/projects.ts
```diff
@@ -85,7 +101,7 @@ const main = async (client: Client) => {
   }
 
   try {
-    await run({ client, contextName });
+    return await run({ client, scope });
```
styfle: I think this try/catch can be removed since its already handled on line 114

packages/cli/src/util/link-project.ts
```diff
@@ -0,0 +1,44 @@
+import { Org, Project } from '../types';
```
EndangeredMassa: This file is called `ensure-link.ts` in your other PRs, right?

EndangeredMassa (approved):

packages/cli/src/util/link-project.ts
MatthewStanciu: Yes, thank you for catching!

packages/cli/src/commands/projects.ts
```diff
@@ -54,6 +69,7 @@ const main = async (client: Client) => {
     argv = getArgs(client.argv.slice(2), {
       '--next': Number,
       '-N': '--next',
+      '--yes': Boolean,
```
styfle, line 72: This should be `--confirm` right? Like `vc dev --confirm` and `vc link --confirm`?

PR #8015, opened 2022-06-25
[next] Ensure trailing slash is handled while resolving _next/data rewrites

ijjk:
This ensures we handle a trailing slash being included in a rewrite while resolving `_next/data` rewrites as this can occur when `NextUrl`'s normalizing isn't able to be applied. 

This also fixes our `mustNotContain` probe as it previously wasn't asserting correctly and also didn't allow being combined with a `mustContain` check in the same probe. 

### Related Issues

x-ref: https://github.com/vercel/examples/pull/331
x-ref: [slack thread](https://vercel.slack.com/archives/C01MYAM1G0K/p1656130164891489)

#### Tests

#### Code Review

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -945,28 +945,13 @@ export async function serverBuild({
             continue: true,
             ...(isOverride ? { override: true } : {}),
           },
-          // handle trailing slash
-          {
-            src: path.join('^/', entryDirectory, '((?!_next/).*)/$'),
-            has: [
-              {
-                type: 'header',
-                key: 'x-nextjs-data',
-              },
-            ],
-            dest: `${path.join(
-              '/',
-              entryDirectory,
-              '/_next/data/',
-              buildId,
-              '/$1.json'
-            )}`,
-            continue: true,
-            ...(isOverride ? { override: true } : {}),
-          },
           // handle non-trailing slash
           {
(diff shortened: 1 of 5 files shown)


packages/next/src/server-build.ts
```diff
@@ -945,6 +945,26 @@ export async function serverBuild({
...
+              '/_next/data/',
+              buildId,
+              '/$1.json'
+            )}`,
+            continue: true,
+            ...(isOverride ? { override: true } : {}),
+          },
+          // handle non-trailing slash
           {
             src: path.join('^/', entryDirectory, '((?!_next/).*)$'),
```
cramforce: Could we instead change this one to `'((?!_next/).*)/?$'`

I'd love to avoid adding another has route

cramforce (changes requested):

packages/next/src/server-build.ts
ijjk: That doesn't exclude the trailing slash from the capture group correctly x-ref: https://regexr.com/6ogac

I initially tried just adding that as well.

cramforce: How about https://regexr.com/6ogao

ijjk: Nice, yeah looks like that works. Updated in https://github.com/vercel/vercel/pull/8015/commits/e5638fc1feeb72e829e74cae98dddf23b63fd632

cramforce (approved):

PR #8020, opened 2022-06-27
Add middleware test

javivelasco:
TBD


EndangeredMassa:
@javivelasco is this PR still relevant? It looks like there are tests in this area of functionality.

PR #8023, opened 2022-06-27
[cli] Support "http:" scheme in `vc bisect`

MatthewStanciu:
`vc bisect` currently prepends `https://` to a passed-in url if it doesn't begin with https—which means that if someone passes in a url that begins with `http://`, it'll turn the url into `https://http://url.com`. This PR fixes this.

#### Tests

#### Code Review


packages/cli/src/commands/bisect/index.ts
```diff
@@ -370,6 +366,16 @@ export default async function main(client: Client): Promise<number> {
...
 }
 
+function hasScheme(url: string): Boolean {
+  return url.startsWith('http://') || url.startsWith('https://');
+}
+
+function validateUrl(url: string): string {
+  if (!hasScheme(url)) {
+    return `https://${url}`;
+  } else return url;
```
styfle:
```suggestion
  }
  return url;
```

No `else` needed here because of the early return

styfle: Actually, you can do it with a single return

```ts
return hasScheme(url) ? url : `https://${url}`
```

packages/cli/src/commands/bisect/index.ts
```diff
@@ -370,6 +366,16 @@ export default async function main(client: Client): Promise<number> {
   return 0;
 }
 
+function hasScheme(url: string): Boolean {
+  return url.startsWith('http://') || url.startsWith('https://');
+}
+
+function validateUrl(url: string): string {
```
styfle: How about `normalizeURL` since usually validate implies throwing an error if its invalid

EndangeredMassa (approved):

TooTallNate (approved):

packages/cli/src/commands/bisect/index.ts
```diff
@@ -370,6 +366,14 @@ export default async function main(client: Client): Promise<number> {
   return 0;
 }
 
+function hasScheme(url: string): Boolean {
+  return url.startsWith('http://') || url.startsWith('https://');
+}
+
+function normalizeURL(url: string): string {
+  return hasScheme(url) ? url : `https://${url}`;
+}
```
TooTallNate: For the purposes of testing this PR, I would recommend moving this into a file in `src/utils/commands/bisect` dir and adding some unit tests with various input values.

MatthewStanciu: 👍 Just added tests

TooTallNate: Awesome. Thanks!

packages/cli/test/unit/util/bisect/normalize-url.test.ts
```diff
@@ -0,0 +1,16 @@
...
+describe('normalize-url', () => {
+  it('should add https to url without scheme', () => {
+    const normalizedUrl = normalizeURL('vercel.com');
+    expect(normalizedUrl).toEqual('https://vercel.com');
+  });
+  it('should not add anything to a url that starts with https', () => {
+    const normalizedUrl = normalizeURL('https://vercel.com');
+    expect(normalizedUrl).toEqual('https://vercel.com');
+  });
+  it('should not add anything to a url that starts iwth http', () => {
```
TooTallNate:
```suggestion
  it('should not add anything to a url that starts with http', () => {
```

styfle (approved):
Thanks!

PR #8029, opened 2022-06-28
[remix] Don't depend on `@remix-run/vercel`

TooTallNate:
Instead, just add it to the project's `package.json` file before installing the dependencies.

Fixes warning about missing peer dependencies when installing CLI.

CleanShot 2022-05-22 at 09 40 09@2x

Now, a warning is shown in the Deployment build logs, saying that the dep was added, but that the user should commit the change:

Screen Shot 2022-06-27 at 8 15 19 PM


packages/remix/src/build.ts
```diff
@@ -70,6 +78,52 @@ export const build: BuildV2 = async ({
     }
   }
 
+  // Ensure `@remix-run/vercel` is in the project's `package.json`
+  const packageJsonPath = await walkParentDirs({
+    base: repoRootPath,
+    start: workPath,
+    filename: 'package.json',
+  });
+  console.log({ packageJsonPath });
```
styfle: Remove this log?

packages/remix/src/build.ts
```diff
@@ -70,6 +78,52 @@ export const build: BuildV2 = async ({
...
+    let modified = false;
+    if (REMIX_RUNTIME_ADAPTER_NAME in devDependencies) {
+      dependencies[REMIX_RUNTIME_ADAPTER_NAME] =
+        devDependencies[REMIX_RUNTIME_ADAPTER_NAME];
+      delete devDependencies[REMIX_RUNTIME_ADAPTER_NAME];
+      console.log(
+        `Warning: Moving "${REMIX_RUNTIME_ADAPTER_NAME}" from \`devDependencies\` to \`dependencies\`. You should commit this change.`
+      );
+      modified = true;
+    } else if (!('@remix-run/vercel' in dependencies)) {
```
styfle: Should this use the constant too?

packages/remix/src/build.ts
```diff
@@ -70,6 +78,52 @@ export const build: BuildV2 = async ({
...
+      const packageJsonString = JSON.stringify(
+        {
+          ...packageJson,
+          dependencies,
+          devDependencies,
+        },
+        null,
+        2
+      );
+      await fs.writeFile(packageJsonPath, `${packageJsonString}\n`);
```
styfle, line 120: Will this break lock files?

TooTallNate: Shouldn't. `npm` updates the `package-lock.json` file after this update, at least.

packages/remix/package.json
```diff
@@ -21,7 +21,6 @@
     "default-server.js"
   ],
   "dependencies": {
-    "@remix-run/vercel": "1.4.3",
     "@vercel/nft": "0.19.1"
```
styfle, line 24: Should we move all deps to devDeps and run ncc like the other builders?

styfle (approved):

PR #8031, opened 2022-06-28
[cli][dev] fix `vc dev` hanging when `content-length` response header is wrong

EndangeredMassa:
When responding from an Edge Function during `vc dev`, if the `content-length` response header has a value higher than the actual response body, `vc dev` will never deliver the response because it's still waiting for more content.

This can happen (for some reason) with some `fetch` responses. To work around this, the `content-length` header is re-calculated in the `edge-runtime` handler. This makes the responses actually close and respond with the proper content.


packages/node/src/dev-server.ts
```diff
@@ -197,8 +197,14 @@ async function compileUserCode(entrypoint: string) {
           }
 
           let response = await edgeHandler(event.request, event);
+          let responseBlob = await response.blob();
 
-          return event.respondWith(response);
+          response.headers.set('content-length', responseBlob.size.toString());
```
Schniz: This is going to break response streaming.
Next version of `edge-runtime` will remove the `content-length` header, as we're always streaming the data in local env anyway in `runtime.createServer` and in Next.js: https://github.com/vercel/edge-runtime/commit/1fa8e94de79e2fa11bd7d338ec4ff42bcd685c23

can we support response streaming in `vc dev` too? In any case, the lack of `content-header` will mean it's going to be re-calculated implicitly, right?

(there's no way to know whether the given body was a stream or not, at least without patching the `Response` constructor... which we might do in the future, but right now we don't)

EndangeredMassa: I updated this to do the same header deletes. I'm still not seeing it stream the response properly and I'm not sure what's happening here.

In the current version of this PR, if you visit http://localhost:3000/api/edge-return-fetch-binary you'll see a broken media player. If you visit http://localhost:3000/static/bunny.mp4 you'll see the video start playing.

PR #8032, opened 2022-06-28
[cli][node] switch to esbuild for compiling edge functions

EndangeredMassa:
Edge Functions were being compiled by `ncc` on every request. This was:

- slow
- noisy (in the logs output)
- inflexible (couldn't easily support TS without a `tsconfig.json`, couldn't easily add sourcemaps that work in an edge environment)

Switching to `esbuild` at mitigates these problems. The requests are still not as fast as they could be, but they are faster than `ncc`.


packages/node/package.json
```diff
@@ -36,6 +36,7 @@
     "@vercel/node-bridge": "3.0.0",
```
styfle, line 35: Can we move `@vercel/ncc` back to devDependencies?

packages/node/src/dev-server.ts
```diff
@@ -149,10 +150,22 @@ async function serializeRequest(message: IncomingMessage) {
...
-    const buildResult = await ncc(entrypoint, {
-      target: 'es2022',
+    const result = await esbuild.build({
+      platform: 'node',
+      target: 'node14',
+      sourcemap: 'inline',
+      bundle: true,
+      entryPoints: [entrypoint],
+      write: false, // operate in memory
+      format: 'cjs',
```
styfle, line 160: I think we also need to detect ESM since we support ESM for production deployments with `@vercel/node`

EndangeredMassa: `format` here is output format. It does ingest ESM.

packages/node/package.json
EndangeredMassa: Good call. Moved: [`8e4c7fe` (#8032)](https://github.com/vercel/vercel/pull/8032/commits/8e4c7fecf4ab3ccdc104d0239899f5994f7f7c10)

TooTallNate (approved):

packages/node/src/dev-server.ts
styfle: What about if the user has `package.json` with `type: module`? They would expect the output to be ESM but esbuild will convert to CJS

EndangeredMassa: The output from this is only when executing an edge function during `vc dev` inside the `edge-runtime`. The user never sees this output or has any way to use it directly.

I tested it in a new project where the `package.json` has `"type": "module"` to be sure and it does still execute edge functions properly.

EndangeredMassa: I tested this logic on an edge function that uses mixed modules as suggested by @styfle:

```
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);

// a CommonJS module
const add = require('./sibling-module');

export const config = {
  runtime: 'experimental-edge'
}

export default async function edge(request, event) {
  return new Response('hello there: ' + add(3, 5));
}
```

I made sure the sibling module was commonjs and was used in the response for the edge function. I can’t get any form of this to work on main, my PR, or in a deployment.

---

On `ncc` (on main), it fails in `vc dev` with:

```
Failed to instantiate edge runtime: Module build failed (from ../../vercel/vercel/node_modules/@vercel/ncc/dist/ncc/loaders/shebang-loader.js):
UnhandledSchemeError: Reading from "node:module" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

Deploying it had an error: `Could not resolve "node:module"`

If I remove the `node:` prefix, I get a different error in `vc dev`: `require is not defined`.

Deploying that gave me the error: `Could not resolve "module" (use "platform: 'node'" when building for node)`

---

On `esbuild`, it fails with:

```
require is not defined
 import_node_module = require("module");
                         ^

ReferenceError: require is not defined
```

styfle (approved):

PR #8033, opened 2022-06-28
[cli][dev][node] Support `matchers` config for Middleware in `vc dev`

TooTallNate:
Adds support for `config.matchers` exported property in Middleware during `vc dev`.

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -460,7 +460,26 @@ export const prepareCache: PrepareCache = ({ repoRootPath, workPath }) => {
 
 export const startDevServer: StartDevServer = async opts => {
   const { entrypoint, workPath, config, meta = {} } = opts;
-  const entryDir = join(workPath, dirname(entrypoint));
+  const entrypointPath = join(workPath, entrypoint);
+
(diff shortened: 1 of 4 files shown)


packages/node/src/index.ts
```diff
@@ -460,7 +460,26 @@ export const prepareCache: PrepareCache = ({ repoRootPath, workPath }) => {
 
 export const startDevServer: StartDevServer = async opts => {
   const { entrypoint, workPath, config, meta = {} } = opts;
-  const entryDir = join(workPath, dirname(entrypoint));
+  const entrypointPath = join(workPath, entrypoint);
+
+  if (config.middleware === true && typeof meta.requestUrl === 'string') {
```
EndangeredMassa, line 465: Curious: why do we need the `meta.requestUrl` here?

EndangeredMassa (approved):

packages/node/src/index.ts
TooTallNate: To match against with the `matchers` 😄

TooTallNate: If you mean why are we checking if it's a string here, it's to appease TypeScript since `meta` props are type `unknown`.

PR #8038, opened 2022-06-28
[cli] upgrade `update-notifier` to fix vulnerability

alex-grover:
#### Description

Upgrade `update-notifier` for security vulnerability fixed in https://github.com/yeoman/update-notifier/pull/222. An older version is included via `ava` but I didn't touch it since that's a dev dep.

The unit tests pass but seem to modify the fixture file so I included that as well.

#### Tests

#### Code Review

- (N/A) Issue from task tracker has a link to this PR


styfle (commented):
This version switched to ESM so we'll need to update TS as well so we can use `await import()` from CJS

https://github.com/yeoman/update-notifier/blob/3b6b9b18428aa3848960f02df53cb571a7620a51/package.json#L13

I'm starting to wonder if it makes more sense to drop `update-notifier` since its [quite large](https://packagephobia.com/result?p=update-notifier@6.0.2) from something that should be a single fetch.

alex-grover:
@styfle got it, thanks for the pointers! happy to take a stab at upgrading TS, let me know how you'd like to proceed.

alex-grover:
closing in favor of #8090

PR #8039, opened 2022-06-28
[cli] Add `Client#stdin` / `Client#stdout`

TooTallNate:
This will allow for mockability of the input streams (i.e. prompts) for CLI commands in unit tests.

There will be a follow-up to this PR that adjusts the way output streams are mocked as well.

**Example:**

```typescript
import confirm from '../../src/util/input/confirm';
import { client } from '../mocks/client';

describe('MockClient', () => {
  it('should mock `confirm()`', async () => {
    const confirmedPromise = confirm(client, 'Do the thing?', false);

    client.stdin.write('yes\n');

    const confirmed = await confirmedPromise;
    expect(confirmed).toEqual(true);
  });
});
```

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -140,6 +140,7 @@ export default async function main(client: Client): Promise<number> {
       }
 
       confirmed = await confirm(
+        client,
         `No Project Settings found locally. Run ${cli.getCommandName(
           'pull'
         )} for retrieving them?`,
--- a/packages/cli/src/commands/env/pull.ts
+++ b/packages/cli/src/commands/env/pull.ts
@@ -74,6 +74,7 @@ export default async function pull(
     exists &&
     !skipConfirmation &&
     !(await confirm(
+      client,
       `Found existing file ${param(filename)}. Do you want to overwrite?`,
       false
     ))
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -65,7 +65,7 @@ export default async function init(
     return extractExample(client, name, dir, force, 'v1');
   }
 
-  const found = await guess(exampleList, name);
+  const found = await guess(client, exampleList, name);
 
   if (typeof found === 'string') {
     return extractExample(client, found, dir, force);
@@ -194,7 +194,7 @@ function prepareFolder(cwd: string, folder: string, force?: boolean) {
 /**
  * Guess which example user try to init
  */
-async function guess(exampleList: string[], name: string) {
+async function guess(client: Client, exampleList: string[], name: string) {
   const GuessError = new Error(
     `No example found for ${chalk.bold(name)}, run ${getCommandName(
       `init`
@@ -208,7 +208,7 @@ async function guess(exampleList: string[], name: string) {
   const found = didYouMean(name, exampleList, 0.7);
 
   if (typeof found === 'string') {
-    if (await promptBool(`Did you mean ${chalk.bold(found)}?`)) {
+    if (await promptBool(`Did you mean ${chalk.bold(found)}?`, client)) {
       return found;
     }
   } else {
--- a/packages/cli/src/util/input/input-project.ts
+++ b/packages/cli/src/util/input/input-project.ts
@@ -47,11 +47,16 @@ export default async function inputProject(
 
   if (!detectedProject) {
     // did not auto-detect a project to link
-    shouldLinkProject = await confirm(`Link to existing project?`, false);
+    shouldLinkProject = await confirm(
+      client,
+      `Link to existing project?`,
+      false
+    );
   } else {
     // auto-detected a project to link
     if (
       await confirm(
+        client,
         `Found project ${chalk.cyan(
           `“${org.slug}/${detectedProject.name}”`
         )}. Link to it?`,
@@ -63,6 +68,7 @@ export default async function inputProject(
 
     // user doesn't want to link the auto-detected project
     shouldLinkProject = await confirm(
+      client,
       `Link to different existing project?`,
       true
     );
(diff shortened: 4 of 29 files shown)


packages/cli/src/commands/env/pull.ts
```diff
@@ -74,6 +74,7 @@ export default async function pull(
     exists &&
     !skipConfirmation &&
     !(await confirm(
+      client,
```
EndangeredMassa, line 77: For the prompting functions, can we make the signature consistent? It looks like most of them have `client` as the first argument, but `promptBool` has it as the second argument. I think it makes the most sense to have them all take `client` as the first argument.

packages/cli/src/util/input/input-project.ts
```diff
@@ -73,7 +79,11 @@ export default async function inputProject(
     let project: Project | ProjectNotFound | null = null;
 
     while (!project || project instanceof ProjectNotFound) {
-      const answers = await inquirer.prompt({
+      const prompt = inquirer.createPromptModule({
+        input: client.stdin,
+        output: client.stdout,
+      });
```
EndangeredMassa, line 85: Does the creation of `prompt` need to go inside the loop? If not, can we pull it out?

Same for: https://github.com/vercel/vercel/pull/8039/files#diff-93886c8513f6d89a4ad25d3916a215c98dd6432754e988ff561ef31b056c5baaR18-R21

packages/cli/test/mocks/client.ts
```diff
@@ -53,6 +57,12 @@ export class MockClient extends Client {
   }
 
   reset() {
+    this.stdin = new PassThrough();
+    this.stdin.isTTY = true;
+
+    this.stdout = new PassThrough();
+    this.stdout.isTTY = true;
```
EndangeredMassa, line 64: Should these have `isTTY` set in the constructor as well?

packages/cli/test/unit/mock.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    client.stdin.write('yes\n');
+
+    client.stdout.setEncoding('utf8');
+    client.stdout.on('data', d => console.log({ d }));
+
+    let confirmed = await confirmedPromise;
+    expect(confirmed).toEqual(true);
+
+    // false
```
EndangeredMassa, line 17: I think this would work better as two separate tests. That way they can run and pass/fail independently.

packages/cli/src/commands/init/init.ts
```diff
@@ -194,7 +194,7 @@ function prepareFolder(cwd: string, folder: string, force?: boolean) {
 /**
  * Guess which example user try to init
  */
-async function guess(exampleList: string[], name: string) {
```
EndangeredMassa, line 197: Should this function check for `client.stdout.isTTY` instead of `process.stdout.isTTY`? (The line where it happens is folded in Github UI.)

packages/cli/src/commands/build.ts
```diff
@@ -140,6 +140,7 @@ export default async function main(client: Client): Promise<number> {
       }
```
EndangeredMassa, line 140: Should `const isTTY = process.stdin.isTTY;` above the fold here check `client.stdin.isTTY` instead?

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -80,6 +80,7 @@ export default async function setupAndLink(
   const shouldStartSetup =
```
EndangeredMassa, line 80: Line 59: Should this `const isTTY = process.stdout.isTTY;` check `client.stdout.isTTY` instead?

EndangeredMassa (approved):
Because the tests currently pass, I think we can merge this, but there are more changes I think we should make as part of this effort.

EndangeredMassa:
Should `Output` accept an override for `isTTY` that we can pull from `client.stdout.isTTY`?

```
packages/cli/src/util/output/create-output.ts:
  28: this.isTTY = process.stdout.isTTY || false;
```

EndangeredMassa:
Do we need to set `process.stdout.isTTY` at all in this test?

```
packages/cli/test/unit/commands/pull.test.ts:
  33: process.stdout.isTTY = undefined;
  47: process.stdout.isTTY = true;
```

EndangeredMassa:
I think we should also check these locations for replacement of `process` with `client`:

```
packages/cli/src/commands/projects.ts:
  295: process.stdin
  297: process.stdin.pause();

packages/cli/src/commands/remove.ts:
  328: process.stdin
  330: process.stdin.pause();

packages/cli/src/commands/certs/rm.ts:
  119: process.stdin
  121: process.stdin.pause();

packages/cli/src/commands/dns/rm.ts:
  90:  process.stdin
  92:  process.stdin.pause();

packages/cli/src/commands/projects.ts:
  286: process.stdout.write(
  291: process.stdout.write(

packages/cli/src/commands/whoami.ts:
  62:  output.print(`${contextName}\n`, { w: process.stdout });

packages/cli/src/commands/billing/add.js:
   99: process.stdout.write(
  103: process.stdout.write(
  111: process.stdout.write(
  115: process.stdout.write(
  144: process.stdout.write(ansiEscapes.eraseLines(linesToClear));
  157: process.stdout.write(ansiEscapes.eraseLines(linesToClear));

packages/cli/src/commands/certs/issue.ts:
  211: process.stdout.write(`${rows.join('\n')}\n\n`);

packages/cli/src/commands/teams/add.ts:
   82: process.stdout.write(eraseLines(2));
  110: process.stdout.write(eraseLines(2));

packages/cli/src/util/emoji.ts:
  18: if (emoji && process.stdout.isTTY) {

packages/cli/src/util/events.ts:
  171: process.stdout.write(eraseLines(o + 1));

packages/cli/src/util/deploy/process-deployment.ts:
  163: process.stdout.write(`https://${event.payload.url}`);

packages/cli/src/util/domains/purchase-domain-if-available.ts:
  14: const isTTY = process.stdout.isTTY;

packages/cli/src/util/input/list.ts:
  119: process.stdout.write(eraseLines(2));
```

EndangeredMassa:
There are still plenty of places that use `console.log` or `console.error` directly, but that's something we should probably address in a different effort.

TooTallNate:
@EndangeredMassa All valid points. In the interest of improving this situation iteratively, I'll merge this one and start addressing those other suggestions in different PRs.

PR #8041, opened 2022-06-28
[cli][dev] fix: creating "api/some-func.js" after "vc dev" now works

EndangeredMassa:
If there is no `api` directory, then you run `vc dev`, then you create a new function `api/some-func.js`, then this file would not be served as a new function.

This was being caused by incomplete "new file" handling logic. This PR ensures that the proper detection is done in each new file (`getVercelConfig`) that populates key properties (`apiDir`, `apiExtensions`, and extensionless `files`) for determining when a file is used to serve a request.


packages/cli/test/dev/integration-2.test.ts
```diff
@@ -334,3 +344,44 @@ test(
     await testPath(200, '/', /A simple deployment with the Vercel API!/m);
   })
 );
+
+test(
+  '[vercel dev] add a `api/fn.ts` when `api` does not exist at startup`',
+  testFixtureStdio('no-api', async (_testPath: any, port: any) => {
```
EndangeredMassa, line 350: This test [fails with](https://github.com/vercel/vercel/runs/7105237751?check_suite_focus=true):

> Error: Command failed with ENOENT: /home/runner/work/vercel/vercel/packages/cli/scripts/start.js -t qDsVTyaR9ignOzx5EcwwtVot link --confirm

Any ideas why?

packages/cli/test/dev/fixtures/no-api/vercel.json
```diff
@@ -0,0 +1,3 @@
+{
+  "version": 2
+}
```
EndangeredMassa, line 3: This is basically an empty fixture

TooTallNate (approved):

PR #8045, opened 2022-06-29
[python] support Sanic >=21 and python >= 3.10

stkevintan:
### Related Issues

1. exception from python 3.10:
```
TypeError: As of 3.10, the *loop* parameter was removed from Queue() since it is no longer necessary`
```

Remove the deprecated argument `loop` from `Queue`, which can also be omitted in python version < 3.10

2. exception from Sanic > 21.3:
```
File "C:\Users\Kevin\AppData\Local\Temp\zeit-fun-03f18b2d2c7d7\sanic\signals.py", line 93, in get
    group, param_basket = self.find_route(
TypeError: 'NoneType' object is not callable
```
As of Sanic > 21.3, it cannot serve requests immediately after initializing, instead, we need implement the [ASGI lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html) and wait for the startup event completed.  

here I complemented the protocol copied from (same source of the previous HTTP procotol): <https://github.com/jordaneremieff/mangum/blob/main/mangum/protocols/lifespan.py>

### Related link:
https://github.com/encode/uvicorn/pull/498

#### Tests

#### Code Review


TooTallNate:
The changes seem ok to me, but I'm also not a Python expert. Adding an integration test would be preferred though.

stkevintan:
Hi, @TooTallNate just added two integrations on the lifespan protocols, they ( and the rest existing tests) are all passed in my local.

~~However, I am not sure how to test python 3.10, since currently vercel doesn't support python 3.10 runtimes although everything runs perfectly in my local python3.10 env.~~
It seems python 3.10 is supported actually, it also passed the local test.

stkevintan:
All green when test in local:
image

stkevintan:
Hi, @TooTallNate is there anything I should do to push forward?

styfle (approved):
Great, thanks!

simonw:
I'm getting 500 errors which I think are caused by this PR:
- #8231

styfle:
@stkevintan @simonw Let's revert in #8241

PR #8048, opened 2022-06-29
[node] add links to edge error messages

EndangeredMassa:
Add links to some Edge errors.

---

Follow up to: https://github.com/vercel/vercel/pull/8007


packages/node/src/dev-server.ts
```diff
@@ -305,7 +305,7 @@ function parseRuntime(
     throw new Error(
       `Invalid function runtime "${runtime}" for "${entrypoint}". Valid runtimes are: ${JSON.stringify(
         validRuntimes
-      )}`
+      )}. Learn more: https://vercel.com/docs/concepts/functions`
```
styfle: This link doesn't mention the runtimes

EndangeredMassa: Good point. It's what @leerob suggested. https://github.com/vercel/vercel/pull/8007#discussion_r907834570

@leerob is this OK? Will that page mention the runtime explicitly or should we link elsewhere?

TooTallNate: Would be better to use `vercel.link` URL instead, no? So that we can change where it redirects to without having to change CLI code in the future?

megbird: If the problem is not defining a default export, we have content for that in https://vercel.com/docs/concepts/functions/edge-middleware#creating-edge-middleware. Would be useful to provide the example there rather than pointing to the top level doc.

EndangeredMassa:
@styfle @TooTallNate I updated the links to be freshly minted (didn't see any existing ones) vercel.link URLs.

styfle (approved):
Thanks!

PR #8047, opened 2022-06-29
Revert "[next] Update max size warning to handle initial layer better"

Schniz:
Reverts vercel/vercel#8013

PR #8049, opened 2022-06-29
[next] Ensure uncompressed limit is correct

ijjk:
This fixes https://github.com/vercel/vercel/pull/8013 as an argument got swapped and the limit was used as the initial uncompressed size causing bundling to not be applied correctly when close to the uncompressed size limit. Since the size wasn't actually close to hitting the limit our size warning didn't show here. 

A test case for this specific setup with uncompressed files has been added to prevent regressing. 

Fix also verified on this [Vercel deployment](https://vercel.com/vercel/front/8cCjqMrPU3GmjfMLNk3KrXiukErz)

### Related Issues

x-ref: https://github.com/vercel/vercel/pull/8013
x-ref: https://github.com/vercel/vercel/pull/8047

#### Tests

#### Code Review


packages/next/test/integration/index.test.js
```diff
@@ -105,6 +105,23 @@ it('should build using server build', async () => {
       log.includes('WARNING: Unable to find source file for page')
     )
   ).toBeFalsy();
+
+  const lambdas = new Set();
+  let totalLambdas = 0;
+
+  for (const key of Object.keys(output)) {
```
styfle: Nit pick: since `key` is only used to get the value, you could use `Object.values(output)` instead (same for the other usages below

ijjk: Updated in https://github.com/vercel/vercel/pull/8049/commits/729d1319685db3037e40a81ae2eba2f28e93e362

styfle (approved):
Thanks!

PR #8050, opened 2022-06-29
[cli] Fix `vc build` lambda serialization when there's a broken symlink

TooTallNate:
There's some cleanup directory walking logic that was choking when a Lambda outputs a file with a broken symlink. We shouldn't need to traverse into those directories in the case of a symlink anyways, so use `lstat()` instead of `stat()` to prevent that filesystem call from throwing an error.

### DIFF
--- a/packages/cli/src/util/build/write-build-result.ts
+++ b/packages/cli/src/util/build/write-build-result.ts
@@ -385,8 +385,14 @@ export async function* findDirs(
   }
   for (const path of paths) {
     const abs = join(dir, path);
-    const s = await fs.lstat(abs);
-    if (s.isDirectory()) {
+    let stat: fs.Stats;
+    try {
+      stat = await fs.lstat(abs);
+    } catch (err: any) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -385,7 +385,7 @@ export async function* findDirs(
   }
   for (const path of paths) {
     const abs = join(dir, path);
-    const s = await fs.stat(abs);
+    const s = await fs.lstat(abs);
```
ijjk: Do we still want the try/catch for the ENOENT case?

TooTallNate: It shouldn't really happen now, but ya sure, I'll add it just because this piece of logic should never make the build fail.

TooTallNate: da0e8eade299cdde3126020e279fc53e064cccf3

ijjk (approved):

PR #8052, opened 2022-06-29
[cli] Add full stdio mockability for unit tests

TooTallNate:
This PR is a follow-up to #8039, which provides an intuitive syntax for writing unit tests for interactive CLI commands.

The heart of this is the new `await expect(stream).toOutput(test)` custom Jest matcher, which is intended for use with the mock Client `stdout` and `stderr` stream properties. The `test` is a string that will wait for the stream to output via "data" events until a match is found, or it will timeout (after 3 seconds by default). The timeout error has nice Jest-style formatting so that you can easily identify what was output:

Screen Shot 2022-06-29 at 10 33 06 PM

Below is an example of a unit test that was added for an interactive `vc login` session:

```typescript
it('should allow login via email', async () => {
  const user = useUser();

  const exitCodePromise = login(client);

  // Wait for login interactive prompt
  await expect(client.stderr).toOutput(`> Log in to Vercel`);

  // Move down to "Email" option
  client.stdin.write('\x1B[B'); // Down arrow
  client.stdin.write('\x1B[B'); // Down arrow
  client.stdin.write('\x1B[B'); // Down arrow
  client.stdin.write('\r'); // Return key

  // Wait for email input prompt
  await expect(client.stderr).toOutput('> Enter your email address:');

  // Write user email address into prompt
  client.stdin.write(`${user.email}\n`);

  // Wait for login success message
    `Success! Email authentication complete for ${user.email}`
  );

  // Assert that the `login()` command returned 0 exit code
});
```

**Note:**  as a consequence of this PR, prompts are now written to stderr instead of stdout, so this change _may_ be considered a breaking change, in which case we should tag it for major release.

### DIFF
--- a/packages/cli/test/mocks/matchers/index.ts
+++ b/packages/cli/test/mocks/matchers/index.ts
@@ -1,0 +1,83 @@
+/**
+ * This file registers the custom Jest "matchers" that are useful for
+ * writing CLI unit tests, and sets them up to be recognized by TypeScript.
+ * 
+ * References:
+ *  - https://haspar.us/notes/adding-jest-custom-matchers-in-typescript
+ *  - https://gist.github.com/hasparus/4ebaa17ec5d3d44607f522bcb1cda9fb
+ */
+
+/// <reference types="@types/node" />
+/// <reference types="@types/jest" />
+import { PassThrough } from 'stream';
+import type { MatcherState } from 'expect';
+
+import * as _matchers from './matchers';
+
+const matchers = {
+  ..._matchers,
+  toHaveWordsCount(this: MatcherState, sentence: string, wordsCount: number) {
+    // implementation redacted
+  },
+};
+
+
+type Tail<T extends unknown[]> = T extends [infer _Head, ...infer Tail]
+  ? Tail
+  : never;
+
+type AnyFunction = (...args: never[]) => unknown;
+
+type GetMatchersType<
+  TMatchers,
+  TResult,
+  > = {
+    [P in keyof TMatchers]: TMatchers[P] extends AnyFunction
+      ? AnyFunction extends TMatchers[P] 
+        ? (...args: Tail<Parameters<TMatchers[P]>>) => TResult
+        : TMatchers[P]
+      : TMatchers[P]
+  };
+
+
+type FirstParam<T extends AnyFunction> = Parameters<T>[0]
+
+type OnlyMethodsWhereFirstArgIsOfType<
+  TObject,
+  TWantedFirstArg
+> = {
+    [P in keyof TObject]: TObject[P] extends AnyFunction
+      ? TWantedFirstArg extends FirstParam<TObject[P]>
+        ? TObject[P]
+        : [`Error: this function is present only when received is:`, FirstParam<TObject[P]>]
+      : TObject[P]
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Matchers<R, T = {}>
+      extends GetMatchersType<
+        OnlyMethodsWhereFirstArgIsOfType<typeof matchers, T>,
+        R
+      > {}
+  }
+}
+
+const jestExpect = (global as any).expect;
+
+if (jestExpect !== undefined) {
+  jestExpect.extend(matchers);
+} else {
+  console.error("Couldn't find Jest's global expect.");
+}
+
+// ✅
+expect(new PassThrough()).toWaitFor('test')
+expect(process.stdout).toWaitFor('test')
+expect('foo bar').toWaitFor(2);
+
+// 🔥 error as expected
+expect('foo bar').toHaveWordsCount(2);
+expect(20).toHaveWordsCount(2);
--- a/packages/cli/test/mocks/matchers/index.ts
+++ b/packages/cli/test/mocks/matchers/index.ts
@@ -7,32 +7,25 @@
  *  - https://gist.github.com/hasparus/4ebaa17ec5d3d44607f522bcb1cda9fb
  */
 
-/// <reference types="@types/node" />
 /// <reference types="@types/jest" />
-import { PassThrough } from 'stream';
-import type { MatcherState } from 'expect';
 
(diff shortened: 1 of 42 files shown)


MatthewStanciu:
I'm a little concerned about prompts being written to stderr instead of stdout. I imagine that will be confusing for people who contribute to this codebase in the future. I don't understand these changes well enough to be able to confidently suggest a different implementation, but I think it may be worth trying to figure out how to write to stdout.

packages/cli/test/mocks/matchers/to-output.ts
```diff
@@ -0,0 +1,68 @@
...
+  getLabelPrinter,
+  matcherHint,
+  printExpected,
+  printReceived,
+} from 'jest-matcher-utils';
+import type { Readable } from 'stream';
+import type { MatcherState } from 'expect';
+import type { MatcherHintOptions } from 'jest-matcher-utils';
+
+export async function toOutput(
```
styfle, line 11: Can we call this `toMatchOutput()`?

packages/cli/test/mocks/matchers/to-output.ts
```diff
@@ -0,0 +1,68 @@
...
+        printExpected(test) +
+        '\n' +
+        printLabel(labelReceived) +
+        (isNot ? '    ' : '') +
+        printReceived(output)
+      );
+    };
+
+    function onData(data: string) {
+      //console.log({ data });
```
styfle: Remove console log?

packages/cli/test/unit/commands/deploy.test.ts
```diff
@@ -10,38 +10,38 @@ import { useUser } from '../../mocks/user';
 describe('deploy', () => {
   it('should reject deploying a single file', async () => {
     client.setArgv('deploy', __filename);
-    const exitCode = await deploy(client);
-    expect(exitCode).toEqual(1);
```
styfle, line 14: ~~Should we continue to check for the exit code?~~

styfle: Oh nevermind I see it below

packages/cli/test/unit/commands/deploy.test.ts
```diff
@@ -10,38 +10,38 @@ import { useUser } from '../../mocks/user';
...
   it('should reject deploying multiple files', async () => {
     client.setArgv('deploy', __filename, join(__dirname, 'inspect.test.ts'));
-    const exitCode = await deploy(client);
-    expect(exitCode).toEqual(1);
-    expect(client.outputBuffer).toEqual(
+    const exitCodePromise = deploy(client);
+    await expect(client.stderr).toOutput(
       `Error! Can't deploy more than one path.\n`
     );
+    await expect(exitCodePromise).resolves.toEqual(1);
```
styfle, line 26: Is the await needed here?

styfle (approved):

packages/cli/test/unit/commands/deploy.test.ts
TooTallNate: Yes

packages/cli/test/mocks/matchers/to-output.ts
TooTallNate:
```suggestion
```

packages/cli/test/mocks/matchers/to-output.ts
TooTallNate: We certainly _can_. I'm not sure if I agree because "output" is already being used as a verb with the current naming, plus it's shorter.

MatthewStanciu (approved):

packages/cli/test/mocks/matchers/to-output.ts
styfle: I was thinking of Output as a noun, like `toStdOut()`, not a verb. The verb makes more sense 👍

packages/cli/test/mocks/matchers/index.ts
```diff
@@ -0,0 +1,65 @@
...
+type AnyFunction = (...args: any[]) => any;
+type PromiseFunction = (...args: any[]) => Promise<any>;
+
+type GetMatcherType<TP, TResult> = TP extends PromiseFunction
+  ? (...args: Tail<Parameters<TP>>) => Promise<TResult>
+  : TP extends AnyFunction
+  ? (...args: Tail<Parameters<TP>>) => TResult
+  : TP;
+
+//type T = GetMatcherType<typeof matchers['toOutput'], void>;
```
EndangeredMassa, line 28: Remove?

packages/cli/test/unit/commands/login.test.ts
```diff
@@ -5,22 +5,45 @@ import { useUser } from '../../mocks/user';
...
+      const user = useUser();
+      client.setArgv('login');
+      const exitCodePromise = login(client);
+      await expect(client.stderr).toOutput(`> Log in to Vercel`);
+
+      // Move down to "Email" option
+      client.stdin.write('\x1B[B'); // Down arrow
+      client.stdin.write('\x1B[B'); // Down arrow
+      client.stdin.write('\x1B[B'); // Down arrow
+      client.stdin.write('\r'); // Return key
```
EndangeredMassa, line 36: Maybe we can make this kind of interaction easier in a future PR. I'm certainly not going to remember the proper code for "down arrow" for example.

packages/cli/test/unit/util/confirm.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    client.stdin.write('\n');
+    confirmed = await confirmedPromise;
+    expect(confirmed).toEqual(true);
+
+    // false (default)
+    confirmedPromise = confirm(client, 'Default false?', false);
+    await expect(client.stderr).toOutput('Default false? [y/N]');
+    client.stdin.write('\n');
+    confirmed = await confirmedPromise;
+    expect(confirmed).toEqual(false);
```
EndangeredMassa, line 32: I think these should be separate tests that can pass/fail on their own.

packages/cli/test/mocks/matchers/index.ts
EndangeredMassa: https://github.com/vercel/vercel/pull/8109

PR #7993, opened 2022-06-17
[cli] 🌈 Give `vc ls` a makeover

MatthewStanciu:
`vc ls` hasn't been touched in years, and it shows. But I think it can be a highly useful command! This PR makes sweeping changes to the `vc ls` command:

- `vc ls` is now scoped to your current directory's Vercel project by default, instead of your user/team
- A new `--all` flag allows you to see all the projects in your user/team
- Deployment URLs are always preceded by `https://` in order to be clickable; the `includeScheme` config flag no longer does anything
- A new `--prod` flag filters only for production deployments
- A new `--inspect` flag displays dashboard inspect urls instead of deployment urls
- The table omits the `username` field when the scope is set to your username
- `age` is now the first field in the table
- The table omits the `project` field when showing deployments for a project
- The project view now specifies which project you're looking at
- The table now displays the build duration in the project view
- `BUILDING`, `QUEUED`, and `CANCELED` are now supported project states, AKA they no longer show `Unknown`
- The state string now displays in title case preceded by a colored circle, to match the dashboard
- Table headings are now colored cyan and bolded
- Project and scope names are now colored magenta
- The command still works almost entirely as it used to—`vc ls [project-name]` will still list deployments for a given project name. The only difference is that the default scope is no longer necessarily your user

...and some other small touches!

#### Try it!

Pull this branch. `cd` into `packages/cli`. Run `ts-eager src/index.ts ls --cwd <full path to a local project that you've linked to Vercel>`.

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8053, opened 2022-06-29
[cli] MAJOR: Scope `vc ls` to linked project

MatthewStanciu:
Currently, `vc ls` is scoped to your team, and you have to type out a project name if you want to see deployments for a project. This PR instead scopes it to the linked project.

Under these changes, `vc ls` still works similarly to how it currently works: users can still specify a project name to get the deployments for a project. The difference is that:

1. The selected team is the one that the linked project belongs to, instead of the one that the user has selected.
2. `vc ls` with no arguments displays the latest deployments for the linked project.

This is the first part of a larger effort to change the behavior of `vc ls`, plucked from https://github.com/vercel/vercel/pull/7993. More PRs to follow.

#### Tests

#### Code Review


packages/cli/src/commands/list.ts
```diff
@@ -42,14 +47,26 @@ const help = () => {
...
-  ${chalk.gray('–')} List all deployments for the app ${chalk.dim('`my-app`')}
+  ${chalk.gray(
+    '–'
+  )} List all projects in the team of the currently linked project
+
+    ${chalk.cyan(`$ ${getPkgName()} ls --all`)}
 
+  ${chalk.gray('–')} List all deployments for the app ${chalk.dim(
+    '`my-app`'
+  )} in the team of the currently linked project
```
EndangeredMassa, line 55: Should we call this a "project" instead of an "app"? How does this sound: "List all deployments for the project `my-app`" ?

packages/cli/test/integration.js
```diff
@@ -1813,31 +1813,6 @@ test('remove the wildcard alias', async t => {
 });
 */
 
-test('ensure username in list is right', async t => {
```
EndangeredMassa, line 1816: Why do we not need these test anymore?

EndangeredMassa (approved):

packages/cli/test/integration.js
MatthewStanciu: I decided to remove this test because it's no longer necessary now that there are unit tests that cover this. It will also start failing once I remove the username field from the table, so I figured it was better not to deal with it.

packages/cli/src/commands/list.ts
MatthewStanciu: I think that's a great idea. `project` is used a lot more within Vercel than `app`

packages/cli/test/unit/commands/list.test.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+function readOutputStream(client: MockClient): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const chunks: Buffer[] = [];
+    client.stderr.resume();
+    client.stderr.on('data', chunk => {
+      chunks.push(chunk);
+      if (chunks.length === 3) {
+        resolve(chunks.toString().replace(/,/g, ''));
```
EndangeredMassa, line 134: The timeout below will trigger after this resolve happens. I think we should be careful with promise resolution. I don't think it will matter in this specific case, but generally it's a good idea to prevent resolving and rejecting the same promise.

You could keep a reference to the timeout handle when it's created:

```
const timeoutHandle = setTimeout(...);
```

Then clear it here when resolving the promise:

```
clearTimeout(timeoutHandle);
```

packages/cli/test/unit/commands/list.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  const project = output.match(/(?<=Deployments for )(.*)(?= under)/);
+  const org = output.match(/(?<=under )(.*)(?= \[)/);
+
+  return {
+    project: project?.[0],
+    org: org?.[0],
+  };
+}
+
+function formatTable(output: string): string[] {
```
EndangeredMassa: This function name makes it sound like it's formatting the table, but it actually seems to be parsing it. I think something like `parseTable` makes more sense here.

EndangeredMassa (approved):
Good stuff!

packages/cli/test/fixtures/unit/commands/list/with-team/.vercel/project.json
```diff
@@ -0,0 +1,4 @@
+{
+  "projectId": "prj_Am19DF8JBL9g89tn4RdDVD59axFi",
+  "orgId": "team_MtLD9hKuWAvoDd3KmiHs9zUg"
+}
```
styfle, line 4: These look like real project/org ids 🤔 

I would prefer the test is mocked using fake IDs (or use an integration test to run vc list after creating deployments)

packages/cli/src/commands/list.ts
```diff
@@ -42,14 +47,26 @@ const help = () => {
...
 
-  ${chalk.gray('–')} List all deployments
+  ${chalk.gray('–')} List all deployments for the currently linked project
 
     ${chalk.cyan(`$ ${getPkgName()} ls`)}
 
-  ${chalk.gray('–')} List all deployments for the app ${chalk.dim('`my-app`')}
+  ${chalk.gray(
+    '–'
+  )} List all projects in the team of the currently linked project
```
styfle: This sounds like its the same as `vc project ls`. I think we should keep it as a separate command instead of overloading the `vc ls` command

EndangeredMassa: I'm good with either, but if we remove `--all`, it may still be useful to include help text here saying to use `vc project ls` for this case.

packages/cli/src/util/link-project.ts
```diff
@@ -0,0 +1,44 @@
...
+import setupAndLink from './link/setup-and-link';
+import param from './output/param';
+import { getCommandName } from './pkg-name';
+import { getLinkedProject } from './projects/link';
+
+type LinkResult = {
+  org: Org;
+  project: Project;
+};
+export async function ensureLink(
```
styfle: Lets make the exported function name match the file name (or rename the file)

packages/cli/src/commands/list.ts
```diff
@@ -31,6 +34,7 @@ const help = () => {
     'DIR'
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
+    -y, --yes                      Skip the confirmation prompt
```
styfle: This should be `--confirm` because its not asking a yes or no question. See `vc link --confirm` and `vc dev --confirm` for existing usage.

packages/cli/src/commands/list.ts
```diff
@@ -86,18 +93,64 @@ export default async function main(client: Client) {
...
+  let app: string | undefined = appArg || project?.name;
+  let host: string | undefined = undefined;
+
+  if (app && !isValidName(app)) {
+    error(`The provided argument "${app}" is not a valid project name`);
+    return 1;
+  }
 
-  let contextName = null;
+  // If there's no linked project and user doesn't pass `app` or `all` args,
```
EndangeredMassa: The `all` arg doesn't exist anymore. Update?

EndangeredMassa (approved):

packages/cli/src/util/ensure-link.ts
```diff
@@ -0,0 +1,44 @@
...
+      return 0;
+    }
+  }
+
+  if (link.status === 'error') {
+    if (link.reason === 'HEADLESS') {
+      client.output.error(
+        `Command ${getCommandName(
+          commandName
+        )} requires confirmation. Use option ${param('--yes')} to confirm.`
```
styfle, line 37: This should be `--confirm`, right?

PR #8058, opened 2022-06-30
[build-utils] Handle `npm bin` exit code 7

lucleray:
In some rare cases, `npm bin` exits with code 7, but still outputs the right bin path.

To reproduce, try:
```
npm init -y
echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
vc
# enter "echo build" for the build command, leave the other configuration as default
```

The build will fail with `Error: Command exited with 7` because `npm bin` fails with code 7, for some reason.

In this PR, we do 2 things:
(1) Ignore exit codes from `npm bin`. It still outputs the right path when it exits with code 7 so we just read the output and check if it's a valid path.
(2) Throw a more specific error message when `npm bin` fails to give us the bin path. The current error was hard to debug because it looked like it was coming from the install commmand. We can do better by emitting a custom error.

Alternative considered for (2): Do not throw errors. If `npm bin` fails, emit a warning and let the build continue.

Related Issues:
- https://github.com/vercel/customer-issues/issues/585 (internal)

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -172,9 +172,32 @@ export async function execCommand(command: string, options: SpawnOptions = {}) {
   return true;
 }
 
-export async function getNodeBinPath({ cwd }: { cwd: string }) {
-  const { stdout } = await execAsync('npm', ['bin'], { cwd });
-  return stdout.trim();
+export async function getNodeBinPath({
+  cwd,
+}: {
+  cwd: string;
+}): Promise<string | undefined> {
+  const { code, stdout, stderr } = await execAsync('npm', ['bin'], {
+    cwd,
+    returnInsteadOfThrowing: true,
+  });
+
+  // in some rare cases, we saw `npm bin` exit with code 7, but still
+  // output the right bin path, so we consider both code 0 and 7 as success
+  if (code !== 0 && code !== 7) {
+    let message = `Running \`npm bin\` failed with code ${code}`;
+    if (stderr) {
+      message += `: ${stderr}`;
+    }
+
+    throw new NowBuildError({
+      code: `BUILD_UTILS_GET_NODE_BIN_PATH`,
+      message,
+    });
+  }
+
+  const nodeBinPath = stdout.trim();
+  return nodeBinPath;
 }
 
 async function chmodPlusX(fsPath: string) {
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -181,25 +181,22 @@ export async function getNodeBinPath({
   const { code, stdout, stderr } = await execAsync('npm', ['bin'], {
     cwd,
     prettyCommand: 'npm bin',
+
+    // in some rare cases, we saw `npm bin` exit with a non-0 code, but still
+    // output the right bin path, so we ignore the exit code
     ignoreNon0Exit: true,
   });
 
-  // in some rare cases, we saw `npm bin` exit with code 7, but still
-  // output the right bin path, so we consider both code 0 and 7 as success
-  if (code !== 0 && code !== 7) {
-    let message = `Running \`npm bin\` failed with code ${code}`;
-    if (stderr) {
-      message += `: ${stderr}`;
-    }
+  const nodeBinPath = stdout.trim();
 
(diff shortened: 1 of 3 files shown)


craigandrews (approved):

styfle (commented):
Please add tests 🙂

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -123,24 +129,24 @@ export function execAsync(
...
-            ? `Command "${opts.prettyCommand}"`
-            : 'Command';
-
-          return reject(
-            new NowBuildError({
-              code: `BUILD_UTILS_EXEC_${code || signal}`,
-              message: `${cmd} exited with ${code || signal}`,
-            })
-          );
+        if (code === 0 || opts.returnInsteadOfThrowing) {
```
styfle: Should `returnInsteadOfThrowing` also be used for the `.on('error')` case?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -166,9 +172,32 @@ export async function execCommand(command: string, options: SpawnOptions = {}) {
...
-  const { stdout } = await execAsync('npm', ['bin'], { cwd });
-  return stdout.trim();
+export async function getNodeBinPath({
+  cwd,
+}: {
+  cwd: string;
+}): Promise<string | undefined> {
+  const { code, stdout, stderr } = await execAsync('npm', ['bin'], {
+    cwd,
+    returnInsteadOfThrowing: true,
```
styfle: Lets also add `prettyCommand: 'npm bin'` here so the error is clear

packages/build-utils/src/fs/run-user-scripts.ts
lucleray: Hmm, I might rename `returnInsteadOfThrowing`. I just want to handle the case where the process exits with some other code than 0, not other types of errors.

packages/build-utils/src/fs/run-user-scripts.ts
lucleray: `prettyCommand` is only read when `returnInsteadOfThrowing` is not true, but I can still add it.

packages/build-utils/src/fs/run-user-scripts.ts
styfle: Is there any precendent for that in `execa` or others? Trying to think of a better name here.

I know there is a [`reject: false`](https://www.npmjs.com/package/execa#reject) that works the way I'm describing: 

> resolves the promise with the error instead of rejecting it.

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
lucleray: I renamed it to `ignoreNon0Exit`, what do you think about that name?

`reject: false` seems a bit vague 🤔

lucleray:
@styfle I added tests and tackled your suggestion 👍

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -166,9 +173,33 @@ export async function execCommand(command: string, options: SpawnOptions = {}) {
...
+}): Promise<string | undefined> {
+  const { code, stdout, stderr } = await execAsync('npm', ['bin'], {
+    cwd,
+    prettyCommand: 'npm bin',
+    ignoreNon0Exit: true,
+  });
+
+  // in some rare cases, we saw `npm bin` exit with code 7, but still
+  // output the right bin path, so we consider both code 0 and 7 as success
+  if (code !== 0 && code !== 7) {
```
styfle: What about if we ignored the exit code and instead simplified this: 

```js
const nodeBinPath = stdout.trim()

if (path.isAbsolute(nodeBinPath)) {
  return nodeBinPath
}

throw new NowBuildError()
```

lucleray: I like it too 👍

lucleray: Updated in https://github.com/vercel/vercel/pull/8058/commits/8d448a82080910aef52f1a2704b802ac7bf09f38 👍 Thank you for the suggestion!

styfle (approved):
Great, thanks!

PR #8064, opened 2022-06-30
[node] Fix public TypeScript types

### LINKED ISSUE #7951: Cannot find module '@vercel/build-utils' or its corresponding type declarations.
### System Info

- yarn: `1.22.17`
- pnpm: `7.1.6`
- npm: `8.1.2`

All three had the same issue.

### Dependencies

- `"@types/node": "^17.0.42"`
- `"@vercel/node": "^2.1.0"`
- `"typescript": "^4.7.3"`

### Error

```
node_modules/.pnpm/@vercel+node@2.1.0/node_modules/@vercel/node/dist/index.d.ts:1:29 - error TS2307: Cannot find module '@vercel/build-utils' or its corresponding type declarations.

1 import { shouldServe } from '@vercel/build-utils';

node_modules/.pnpm/@vercel+node@2.1.0/node_modules/@vercel/node/dist/index.d.ts:2:60 - error TS2307: Cannot find module '@vercel/build-utils' or its corresponding type declarations.

2 import type { BuildV3, PrepareCache, StartDevServer } from '@vercel/build-utils';
```

## Reproduction

```js
import type { VercelRequest, VercelResponse } from '@vercel/node';
import type { Readable } from 'node:stream';

export const config = {
  api: {
    bodyParser: false,
  },
};

async function buffer(readable: Readable) {
  const chunks = [];
  for await (const chunk of readable) {
    chunks.push(typeof chunk === 'string' ? Buffer.from(chunk) : chunk);
  }
  return Buffer.concat(chunks);
}

export default async function (req: VercelRequest, res: VercelResponse) {
  if (req.method === 'POST') {
    const buf = await buffer(req);
    const rawBody = buf.toString('utf8');

    res.json({ rawBody });
  } else {
    res.setHeader('Allow', 'POST');
    res.status(405).end('Method Not Allowed');
  }
}
```

styfle:
There was a regression some time between [1.12.2-canary.6](https://unpkg.com/browse/@vercel/node@1.12.2-canary.6/dist/index.d.ts) and [1.12.2-canary.7](https://unpkg.com/browse/@vercel/node@1.12.2-canary.7/dist/index.d.ts) where the `index.d.ts` types changed.

This PR reverts back to the old behavior so users can properly use

```ts
import { VercelRequest, VercelResponse } from '@vercel/node';
```

The existing test was also changed to use the result of the build rather than copy at the beginning

https://github.com/vercel/vercel/blob/cfae7ec3c291d1802fb6eec6642a0ebf792fb71d/packages/node/test/fixtures/15-helpers/ts/index.ts#L1

- Fixes https://github.com/vercel/vercel/issues/7951


TooTallNate (approved):

packages/node/src/example-import.ts
```diff
@@ -1,14 +0,0 @@
-// We intentionally import these types here
-// which will fail at compile time if exports
-// are not found in the index file
```
styfle, line 3: This file was not actually testing the build output so I deleted it.

Instead, we'll rely on the existing `test/fixtures/15-helpers` and make sure the build output is copied as the final step in the build before tests begin.

PR #8071, opened 2022-07-01
[hydrogen] Add `@vercel/hydrogen` Builder

TooTallNate:
Adds a new `@vercel/hydrogen` Builder package so that Vercel can support Shopify Hydrogen projects with zero config. It outputs an Edge Function for the server-side render code and includes a catch-all route to invoke that function after a `handle: "filesystem"` to serve static files that were generated by the build command.

**Examples:**

 * [`hello-world-ts` template](https://hydrogen-hello-world-otm2vmw6w-tootallnate.vercel.app/)
 * [`demo-store-ts` template](https://hydrogen-demo-store-1gko2fst3-tootallnate.vercel.app/)


packages/hydrogen/edge-entry.js
```diff
@@ -0,0 +1,12 @@
+import handleRequest from '__RELATIVE__/src/App.server';
+import indexTemplate from '__RELATIVE__/dist/client/index.html?raw';
+
+// ReadableStream is bugged in Vercel Edge, overwrite with polyfill
+import { ReadableStream } from 'web-streams-polyfill/ponyfill';
```
styfle, line 5: Do we need to add `web-streams-polyfill` as a dependency to `@vercel/hydrogen`?

packages/hydrogen/src/build.ts
```diff
@@ -0,0 +1,147 @@
...
+      });
+    } else {
+      console.log(`Skipping "install" command...`);
+    }
+  } else {
+    await runNpmInstall(entrypointDir, [], spawnOpts, meta, nodeVersion);
+  }
+
+  // Copy the edge entrypoint file into `.vercel/cache`
+  const edgeEntryDir = join(workPath, '.vercel/cache/hydrogen');
```
styfle, line 66: What is the purpose of copying `edge-entry.js` file to the cache if its an input file? I would expect that only output files are cached.

packages/hydrogen/src/build.ts
```diff
@@ -0,0 +1,147 @@
...
+  const edgeFunction = new EdgeFunction({
+    name: 'hydrogen',
+    deploymentTarget: 'v8-worker',
+    entrypoint: 'index.js',
+    files: edgeFunctionFiles,
+  });
+
+  // The `index.html` file is a template, but we want to serve the
+  // SSR version instead, so omit this static file from the output
+  delete staticFiles['index.html'];
```
styfle, line 125: Is it possible there is more than one `index.html`? For example `about/index.html` or `blog/index.html`?

packages/hydrogen/edge-entry.js
TooTallNate: It's a [dependency of Hydrogen itself](https://github.com/Shopify/hydrogen/blob/v1.x-2022-07/packages/hydrogen/package.json#L145).

```
$ npm ls web-streams-polyfill
hydrogen-app@0.0.0 /Users/nrajlich/hydrogen-app
├─┬ @shopify/cli-hydrogen@3.0.25
│ ├─┬ @shopify/cli-kit@3.0.25
│ │ └─┬ node-fetch@3.2.6
│ │   └─┬ fetch-blob@3.1.5
│ │     └── web-streams-polyfill@3.2.1 deduped
│ └─┬ @shopify/hydrogen@0.26.1
│   └── web-streams-polyfill@3.2.1 deduped
└─┬ @shopify/hydrogen@1.0.2
  └── web-streams-polyfill@3.2.1
```

packages/hydrogen/src/build.ts
TooTallNate: Seems like a fine place to put the file. I consider it a "build artifact" of `@vercel/hydrogen` builder. I'm open to suggestions though if you have a preferred location to copy the file to.

packages/hydrogen/src/build.ts
TooTallNate: I don't think so. I can only find mention of a root-level `index.html` template in the docs, [for example here](https://shopify.dev/custom-storefronts/hydrogen/framework#configuring-default-entry-points).

packages/hydrogen/src/build.ts
styfle: What I mean is that I don't think it needs to be copied to the disk at all. Its similar to node-bridge where you can use a FileBlob to copy the contents into the build output without creating a file on disk.

TooTallNate: Not really. The file is referenced via env var when invoking `shopify hydrogen build`, so their bundler needs access to the file. It also need to be "within" the project source code so that it has access to `node_modules`.

styfle (approved):

PR #8072, opened 2022-07-02
[cli] Add `vc open` command

MatthewStanciu:
This PR adds a new command, `vc open`, which opens a few select pages on the dashboard.

Screen Shot 2022-07-07 at 3 41 07 PM

You can also pass in a combination of the arguments `dash`, `latest`, and `prod` to skip the prompt and open the given page directly.

Examples:

- `vc open dash` opens the dashboard
- `vc open latest` opens the latest preview deployment
- `vc open dash latest` opens the latest preview deployment in the dashboard
- `vc open latest --prod`  opens the production deployment
- `vc open dash latest --prod`  opens the production deployment in the dashboard
- `vc open dash my-deployment-ji2fjij2.vercel.app` opens a specific deployment in the dashboard

#### Tests

#### Code Review


almonk:
❤️ this feature! I love the `vc open` menu command but I have some suggestions on how we could maybe make the other syntax more memorable below;

```bash
vc open dash # Open the dashboard
vc open latest # Open the latest preview deploymwnt
vc open dash latest # Open the latest preview deployment in the dashboard
vc open prod # Open the product deployment
vc open dash prod # Open the production deployment in the dashboard

# For parity with other commands...
vc open dash my-deployment-ji2fjij2.vercel.app # Open a specific deployment in the dashboard
vc open my-deployment-ji2fjij2.vercel.app # Open a specific deployment
```

I think this is a simple syntax and concept to remember, since now you only need to remember 3 keywords: `latest`, `prod`, `dash` and the combination of them allows you acess all routes.

FWIW I would also make sure we alias `production -> prod` and `dashboard -> dash` to avoid being annoying.

packages/cli/src/commands/deploy/args.ts
```diff
@@ -21,6 +21,7 @@ export const help = () => `
       link                 [path]      Link local directory to a Vercel Project
       login                [email]     Logs into your account or creates a new one
       logout                           Logs out of your account
+      open                 [options]   Opens a project URL in your browser
```
almonk: Capitilisation: project -> Project

almonk (approved):
Some super minor copy stuff – but this is great

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  ${chalk.gray('–')} Open the dashboard for the latest production deployment
+
+    ${chalk.cyan(`$ ${getPkgName()} open dash prod`)}
+
+  ${chalk.gray('–')} Open a specific deployment URL
+  
+    ${chalk.cyan(`$ ${getPkgName()} open [url]`)}
+
+  ${chalk.gray('–')} Open the dashboard for a specific deployment
```
EndangeredMassa:
```suggestion
  ${chalk.gray('–')} Open the dashboard for a specific deployment URL
```

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    ${chalk.cyan(`$ ${getPkgName()} open dash latest`)}
+
+  ${chalk.gray('–')} Open the dashboard for the latest production deployment
+
+    ${chalk.cyan(`$ ${getPkgName()} open dash prod`)}
+
+  ${chalk.gray('–')} Open a specific deployment URL
+  
+    ${chalk.cyan(`$ ${getPkgName()} open [url]`)}
```
EndangeredMassa: What's the benefit to `vc open [url]` over `open [url]` (on a mac terminal, anyway)?

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  const dashboardUrl = getDashboardUrl(org, project);
+  const inspectorUrl =
+    (await getInspectorUrl(client, project, org, team)) || 'not_found';
+  const prodInspectorUrl =
+    (await getInspectorUrl(client, project, org, team, true)) || 'not_found';
+  const latestDeployment =
+    (await getLatestDeploymentUrl(client, project, team)) || 'not_found';
+  const latestProdDeployment =
+    (await getLatestDeploymentUrl(client, project, team, true)) || 'not_found';
```
EndangeredMassa: Can we defer these requests until they are needed? If any one of these is slow, the entire command is slow.

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+  if (choice === '') {
+    // User aborted
+    return 0;
+  }
+
+  if (!test) execa('open', [choice]);
+  output.log(`🪄 Opened ${link(choice)}`);
+  return 0;
+}
```
EndangeredMassa, line 162: This function is pretty long and getting hard to scan. Can we extract some parts into named functions?

packages/cli/test/unit/commands/open.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  let teamSlug: string = '';
+
+  it('should open the dashboard', async () => {
+    const cwd = fixture('project');
+    try {
+      process.chdir(cwd);
+
+      useUser();
+      const team = useTeams('team_dummy');
+      teamSlug = team[0].slug;
```
EndangeredMassa: It looks like other tests depend on this one running. I suggest avoiding test dependencies because it makes them brittle and impossible to run in isolation.

EndangeredMassa (approved):

packages/cli/src/commands/open.ts
MatthewStanciu: Nothing really, it's just like `open` except you can only open Vercel deployments. Happy to remove if it's confusing/weird UX.

packages/cli/test/unit/commands/open.test.ts
MatthewStanciu: This was a good catch, thank you.

packages/cli/src/commands/open.ts
EndangeredMassa: I found it confusing, at least. I assumed it must do more than just open the provided url, but wasn't sure what that could be.

EndangeredMassa (approved):

packages/cli/src/commands/open.ts
styfle: Agreed, it seems confusing to me because I would expect it to do something different specific to Vercel

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    ${chalk.cyan(`$ ${getPkgName()} open dash`)}
+
+  ${chalk.gray('–')} Open the latest preview deployment URL
+
+    ${chalk.cyan(`$ ${getPkgName()} open latest`)}
+
+  ${chalk.gray('–')} Open the latest production deployment URL
+
+    ${chalk.cyan(`$ ${getPkgName()} open prod`)}
```
styfle: This should probably be `vc open latest` and `vc open latest --prod` to follow the same pattern as `vc deploy` and `vc deploy --prod`.

MatthewStanciu: Great idea, thanks!

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,325 @@
...
+
+const help = () => {
+  console.log(`
+  ${chalk.bold(`${logo} ${getPkgName()} open`)} [options]
+
+  ${chalk.dim('Options:')}
+
+    -h, --help                     Output usage information
+    --confirm                      Skip confirmation prompts
+    --prod                        Filter for production deployments
```
EndangeredMassa: Is the alignment off here?

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,325 @@
...
+      )} to create a deployment.`
+    );
+    return 1;
+  }
+  if (choice === '') {
+    // User aborted
+    return 0;
+  }
+
+  if (!test) execa('open', [choice]);
```
EndangeredMassa: This will only work on natively macOS, I believe. Should we use a package that will make sure this is cross-platform?

I've not used a package like this in the past. I found this one that looks alright: https://www.npmjs.com/package/open

EndangeredMassa (changes requested):
I think we need to make sure this works cross-platform before merging

packages/cli/src/commands/open.ts
MatthewStanciu: Wow, thank you for catching this.

EndangeredMassa (approved):

packages/cli/src/commands/open.ts
MatthewStanciu: Just removed 👍

packages/cli/package.json
```diff
@@ -52,6 +52,7 @@
     "@vercel/remix": "1.0.6",
     "@vercel/ruby": "1.3.13",
     "@vercel/static-build": "1.0.5",
+    "open": "8.4.0",
```
styfle, line 55: This should be a devDependency so it gets bundled with ncc

dferber90:
Today I found myself wanting this too (opened https://github.com/vercel/vercel/discussions/8174 before realizing this PR exists) 🥳 

I imagined `vc open` to open `vercel.com/:team/:project` right away, without any prompt. I can see myself being inconvenienced by the extra confirmation step or having to type the long form `vc open dash`.

PR #8085, opened 2022-07-05
[cli] MAJOR: remove `vc deploy` clipboard copy feature

MatthewStanciu:
https://vercel.slack.com/archives/C03F2CMNGKG/p1656971502881949

Right now, `vc deploy` automatically copies the deploy url to your clipboard after the deployment has finished. You can opt out via the `--no-clipboard` flag, but the feature is enabled by default.

This is strange behavior—there's no indication that the CLI will hijack your clipboard, and you don't know it's been hijacked until after it happens.

This PR removes the clipboard copying feature as well as the `--no-clipboard` flag.

#### Tests

#### Code Review


styfle (approved):
Looks good 👍  

I would take it a step further and remove the feature completely. Instead, rely on the terminal printing the necessary information and the user clicking to open the url.

MatthewStanciu:
I'm on board with that!

packages/cli/src/commands/deploy/args.ts
```diff
@@ -68,7 +68,7 @@ export const help = () => `
     -m, --meta                     Add metadata for the deployment (e.g.: ${chalk.dim(
       '`-m KEY=value`'
     )}). Can appear many times.
-    -C, --no-clipboard             Do not attempt to copy URL to clipboard
+    -C, --clipboard                Copy URL to clipboard after deployment
```
styfle: This should be removed

Also the docs need to be removed here https://vercel.com/docs/cli#introduction/unique-options/no-clipboard

styfle (changes requested):
The help message needs to be updated.

Also the docs need to be removed here https://vercel.com/docs/cli#introduction/unique-options/no-clipboard

styfle (approved):

EndangeredMassa (approved):

test/benchmark.js
```diff
@@ -3,9 +3,7 @@ const { tmpdir } = require('os');
 const { mkdirSync, writeFileSync } = require('fs');
```
TooTallNate: I'm not even sure what this file is. Perhaps we should delete it if it's not being used by CI.

TooTallNate (approved):

PR #8086, opened 2022-07-05
[cli] improve `isBundledBuilder` logic

EndangeredMassa:
The logic that determines if a builder needs to be installed has a check to see if the builder should already be bundled: `isBundledBuilder`. This was looking for specific conditions that made it (1) a bit hard to follow and (2) very sensitive to `canary` tags (and having "canary" in the version specifier).

This causes general development problems because local changes weren't always used by local CLI builds. Depedendant packages (like `@vercel/node`) would be installed from the latest `canary` release instead.

This caused problems in CI and released CLI versions where the latest `canary` of dependant packages might be rather old, causing that old code to be used instead of the latest non-canary releases.

The issue was mitigated for now by publishing canary releases for all packages.

---

Paired with @styfle @MatthewStanciu.

@TooTallNate: Is this change too broad? Are there cases where we wouldn't want to do this?


packages/cli/test/dev/fixtures/runtime-not-installed/vercel.json
```diff
@@ -1,3 +1,3 @@
 {
-  "builds": [{ "src": "*.js", "use": "@vercel/does-not-exist" }]
```
styfle: This is the only one I'm worried about. I'm thinking perhaps the `isBundledBuilder()` should still check if the name exists in the CLI package.json dependencies.

packages/cli/test/unit/util/dev/builder-cache.test.ts
```diff
@@ -213,89 +148,45 @@ describe('getBuildUtils', () => {
...
+    expect(result).toEqual(false);
+  });
 
-    // URL
-    {
-      const parsed = npa('https://example.com');
-      const result = isBundledBuilder(parsed, cliPkg);
-      expect(result).toEqual(false);
-    }
+  it('shouldNOT  detect git url releases', () => {
```
TooTallNate:
```suggestion
  it('should NOT detect git url releases', () => {
```

TooTallNate (approved):
I think this is a good first step towards ultimately sharing the logic with the `importBuilders()` function in `vc build`.

packages/cli/test/dev/fixtures/runtime-not-installed/vercel.json
TooTallNate: Ya I think that makes sense to me as well.

packages/cli/src/util/dev/builder-cache.ts
```diff
@@ -126,31 +120,6 @@ export function filterPackage(
...
-    const semverInstalled = semver.parse(parsedInstalled.rawSpec);
-    if (!semverInstalled) {
-      return true;
-    }
-    if (semverInstalled.prerelease.length > 0) {
-      return semverInstalled.prerelease[0] !== distTag;
-    }
-    if (distTag === 'latest') {
-      return false;
-    }
```
EndangeredMassa, line 151: After removing the two last cases here, the rest could only return true, which means they did nothing. Removed this whole section as a result.

styfle (approved):
Perfect, thanks! 🎉

PR #8090, opened 2022-07-06
[cli] Replace `update-notifier` dependency with built in

styfle:
This PR replaces the `update-notifier` dependency with a custom implementation. 

There are a few reasons: the dependency is quite large, it requires ESM in order to update, and can sometimes suggest an update to an older version. For example:

image

- Related to #8038


packages/cli/scripts/build.ts
```diff
@@ -47,9 +47,15 @@ async function main() {
     });
   }
 
+  const pkg = await readJSON(join(dirRoot, 'package.json'));
+  const dependencies = Object.keys(pkg?.dependencies ?? {});
```
TooTallNate, line 47: Nice

packages/cli/src/util/update-notifier.ts
```diff
@@ -0,0 +1,107 @@
...
+      config = readJSONSync(cacheFile);
+    } catch (e) {
+      console.error(`failed to read ${cacheFile}`);
+      return null;
+    }
+
+    if (
+      config.version !== pkg.version &&
+      !config.notified &&
+      process.stdout.isTTY
```
TooTallNate: If we want to write unit tests for this, we probably want `client.stdout.isTTY`.

packages/cli/src/util/update-notifier.ts
```diff
@@ -0,0 +1,107 @@
...
+    const agent = new HttpsAgent({
+      keepAlive: true,
+      maxSockets: 15,
+    });
+    const headers = {
+      accept:
+        'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
+    };
+
+    const url = `https://registry.npmjs.org/${pkg.name}`;
```
TooTallNate: Perhaps make `url` configurable so we could test against a mock server?

packages/cli/src/index.ts
```diff
@@ -149,22 +142,28 @@ const main = async () => {
...
+            )} to install ${getTitleName()} CLI ${latest}`
+        )
+      );
 
-    console.log(
-      info(
-        `Changelog: https://github.com/vercel/vercel/releases/tag/vercel@${latest}`
-      )
-    );
+      console.log(
```
TooTallNate, line 163: We should be using `output.log()` here instead of `console.log()` otherwise we'll break the pipe-ability of commands (i.e. deployment URL goes to stdout for `vc deploy`).

I realize that this was the previous behavior as well, but since we're here we might as well fix it.

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,103 @@
+import semver from 'semver';
+import XDGAppPaths from 'xdg-app-paths';
+import { dirname, resolve as resolvePath } from 'path';
+// import { fileURLToPath } from 'url';
```
TooTallNate: Remove?

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * This file is spawned in the background and checks npm for the latest version
+ * of the CLI, then writes the version to the cache file.
+ */
+
+// import fetch from 'node-fetch';
+// import { Agent as HttpsAgent } from 'https';
+// import { outputJSON } from 'fs-extra';
```
TooTallNate: Remove?

```suggestion
```

packages/cli/test/unit/util/update-notifier.test.ts
```diff
@@ -0,0 +1,117 @@
+import fs from 'fs-extra';
+import sleep from '../../../src/util/sleep';
+import tmp from 'tmp-promise';
+import updateNotifier from '../../../src/util/update-notifier';
+import { join } from 'path';
+// import { client } from '../../mocks/client';
```
TooTallNate:
```suggestion
```

EndangeredMassa:
@styfle is this still WIP? Are you looking for more review or should we hold off for now?

styfle:
@EndangeredMassa I think @cb1kenobi has been working on this branch, not sure if its ready for review though

cb1kenobi:
@EndangeredMassa, @styfle this is ready for review. Please take a peek and let me know if there's something I should revisit.

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    cacheDir,
+    'update-notifier',
+    `${pkg.name}-${distTag}.json`
+  );
+
+  const loadCache = async (): Promise<UpdateNotifierConfig | undefined> => {
+    try {
+      return await readJSON(cacheFile);
+    } catch (e: any) {
+      // cache does not exist or malformed
```
EndangeredMassa: **suggestion (non-blocking):** debug log this error

Can we debug log this error? I really prefer to have someway to access this if something goes wrong.

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      const script = resolvePath(
+        dirname(__filename),
+        '..',
+        '..',
+        '..',
+        'dist',
+        'update-worker.js'
+      );
+      // const script = resolvePath(dirname(fileURLToPath(import.meta.url)), '..', '..', '..', 'dist', 'update-worker.js');
```
EndangeredMassa:
```suggestion
```

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      const worker = spawn(process.execPath, [script], {
+        stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
+        windowsHide: true,
+      });
+
+      worker.on('close', () => resolve());
+      worker.on('error', () => resolve());
```
EndangeredMassa: **suggestion (non-blocking):** debug log this error

Can we debug log this error? I really prefer to have someway to access this if something goes wrong.

styfle:
@EndangeredMassa @cb1kenobi Looks like something broke during merge conflicts

cb1kenobi:
@styfle Fixed!

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,103 @@
...
+        '..',
+        'dist',
+        'update-worker.js'
+      );
+      // const script = resolvePath(dirname(fileURLToPath(import.meta.url)), '..', '..', '..', 'dist', 'update-worker.js');
+
+      const worker = spawn(process.execPath, [script], {
+        stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
+        windowsHide: true,
+      });
```
EndangeredMassa: **question (soft-blocking):** Why are we spawning a process for this?

EndangeredMassa (approved):
Approved, but I think we should makes some tweaks in this or a follow-up PR.

packages/cli/src/util/update-notifier/index.ts
styfle: We only want to log if malformed since the first run will throw due to the file not existing

packages/cli/src/util/update-notifier/index.ts
styfle: Its supposed to run in the background so it doesn't wait to check for an update before executing the user's command.

packages/cli/src/index.ts
```diff
@@ -149,22 +142,28 @@ const main = async () => {
...
-      info(
-        `${chalk.black.bgCyan('UPDATE AVAILABLE')} ` +
-          `Run ${cmd(
-            await getUpdateCommand()
-          )} to install ${getTitleName()} CLI ${latest}`
-      )
-    );
+  if (isTTY && !process.env.NO_UPDATE_NOTIFIER) {
+    // Checks for available update and returns an instance
+    const latest = await updateNotifier({
```
styfle: @cb1kenobi The fact that we are awaiting here makes me think this is not running in the background. It should read the file and continue but only write the file in the background.

cb1kenobi: The `await` has nothing to do with whether it's running in the background or not. `updateNotifier()` performs async logic prior to spawning the worker process and we need to block in case an error occurs (which is unlikely).

Another reason we `await` was added was because I added a `wait` option which will block until the update check completes and is useful for debugging.

Blocking while it quickly tries to read the cache file is a relatively cheap operation compared to always spawning the process to perform the check.

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: The commented out line was simply to help @TooTallNate with the ESM conversion. Once we move to ESM, we no longer have `__filename` and I thought it would be more helpful to have the line ready to go than have to waste cycles porting the current `script` value.

packages/cli/src/util/update-notifier/index.ts
EndangeredMassa: That sounds reasonable, but currently the parent process is still waiting for the child process to exit. I can't figure out why, either. I tried removing event listeners and fiddling with spawn options, but it's still always waiting.

I'd be happy to pair on this. I want to figure out why the parent isn't letting the child be free.

It should also be noted that a bug in this process can cause orphaned node process to pile up on a machine. I managed to get several myself when debugging this.

packages/cli/src/util/update-notifier/index.ts
EndangeredMassa: That sounds reasonable, but currently the parent process is still waiting for the child process to exit. I can't figure out why, either. I tried removing event listeners and fiddling with spawn options, but it's still always waiting.

I'd be happy to pair on this. I want to figure out why the parent isn't letting the child be free.

It should also be noted that a bug in this process can cause orphaned node process to pile up on a machine. I managed to get several myself when debugging this.

packages/cli/src/util/update-notifier.ts
cb1kenobi: The `isTTY` check has been moved to the CLI's `index.ts` as to decouple the update notifier from the console.

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: Removed. You'll need to add it back when you port it to ESM.

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: Removed. You'll need to add it back when you port it to ESM.

packages/cli/test/unit/util/update-notifier.test.ts
cb1kenobi: Removed.

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: Done!

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: I removed this line not because it was unused, but because I found an issue in resolving the path. It will need to be eventually be updated to use `fileURLToPath(import.meta.url)` when the CLI moves to ESM.

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: Done!

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: I figured it out and fixed the problem. Essentially the worker process would not disconnect the IPC channel until the check completed.

I believe the code originally worked, but when writing the tests, logic was changed and the dist version was not re-checked.

Since we currently do not have any tests that test the update notifier using the dist version, perhaps a few tests should be added?

EndangeredMassa: The parent is still waiting for the child, in my testing. It's really fast, but it does still happen.

I messed with this some more and I think I figured it out. I think it was a combination of:

- IPC was not disconnected until the check completed
- the `message` event handler returned a promise that waited for the check to complete

If I remove both issues, the parent no longer waits for the child at all. Here's an example:

```
process.on('message', msg => {
  process.disconnect();

  setImmediate(async () => {
    // pretend this is taking a while
    await new Promise((resolve) => {
      setTimeout(resolve, 10000);
    });

    const pending = fetchLatest(msg);
    if (msg.wait) {
      await pending;
    }
  });
});
```

Running this, I can see the initial command exit and an extra node process running for ~10 more seconds. After that process ends, I see the cache file populated.

EndangeredMassa: We discussed this on a call: the code as written does allow the parent to exit reasonably soon without waiting on the child. We decided to add a few more protections to ensure a clean system.

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  let cache = await loadCache();
+
+  if (!cache || cache.expireAt < Date.now()) {
+    await new Promise<void>(async resolve => {
+      // spawn the worker, wait for the worker to report it's ready, then
+      // signal the worker to fetch the latest version
+
+      // we need to find the update worker script since the location is
+      // different based on production vs tests because
```
EndangeredMassa:
```suggestion
      // different based on production vs tests
```

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -57,33 +60,47 @@ export default async function updateNotifier({
...
       const worker = spawn(process.execPath, [script], {
         stdio: ['inherit', 'inherit', 'inherit', 'ipc'],
         windowsHide: true,
       });
 
+      // we allow the child 2 seconds to let us know it's ready before we give up
+      const timer = setTimeout(() => {
+        worker.kill();
+        resolve();
+      }, 2000);
```
EndangeredMassa: **suggestion (non-blocking):** debug log

Can we debug log in this case?

packages/cli/src/util/update-notifier/index.ts
```diff
@@ -57,33 +60,47 @@ export default async function updateNotifier({
...
 
+      // spawn the worker with an IPC channel
       output?.debug(`Spawning ${script}`);
       const worker = spawn(process.execPath, [script], {
         stdio: ['inherit', 'inherit', 'inherit', 'ipc'],
         windowsHide: true,
       });
 
+      // we allow the child 2 seconds to let us know it's ready before we give up
+      const timer = setTimeout(() => {
```
EndangeredMassa: **suggestion (non-blocking):** rename `timer` -> `workerReadyTimer`

Can we use a more specific name for the timeout handle? It will make it clearer what the `clearTimeout` line is doing, for example:

```
clearTimeout(workerReadyTimer);
```

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -20,7 +20,14 @@ process.on('unhandledRejection', err => {
   process.exit(1);
 });
 
-process.on('message', async msg => {
+// this timer will prevent this worker process from running longer than 10s
+const timer = setTimeout(() => {
```
EndangeredMassa: **suggestion (non-blocking):** rename `timer` -> `workerCompleteTimer`

Can we use a more specific name for the timeout handle? It will make it clearer what the `clearTimeout` line is doing, for example:

```
clearTimeout(workerCompleteTimer);
```

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -20,7 +20,14 @@ process.on('unhandledRejection', err => {
...
+  console.error('Update worker timed out, exiting');
```
EndangeredMassa: **suggestion (soft-blocking):** move to debug log

In this or a follow-up PR: Can wove this to a debug log? Because the worker inherits stdio from the parent, if the `vc` command takes >= 10 seconds and the work does time out, the user will see this message. That would be confusing and not actionable for them.

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -57,11 +64,15 @@ async function fetchLatest({ cacheFile, distTag, name, updateCheckInterval }) {
...
 }
 
-/**
- * Signal the parent process we're ready to fetch.
- */
+// signal the parent process we're ready
 if (process.connected && process.send) {
   process.send({ type: 'ready' });
+} else {
+  console.error('No IPC bridge detected, exiting');
```
EndangeredMassa: **suggestion (soft-blocking):** move to debug log

In this or a follow-up PR: This will be confusing for a user and not actionable. Can we move it to a debug log?

EndangeredMassa (approved):

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: LOL

cb1kenobi: Fixed

packages/cli/src/util/update-notifier/index.ts
cb1kenobi: Fixed

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: Users will never see this because `process.connected` and `process.send` will always be defined in the context of the Vercel CLI.

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: I thought about that, but I don't see why it's necessary. The worker will have disconnected and the parent will have exited long before that timeout fires.

Besides, we'd have to create a new `output` instance and then we'd need to figure out if we should pass `--debug` through from the parent or parse it again in the worker.

Maybe we just switch `stdio` to `['ignore', 'ignore', 'ignore', 'ipc']` or just remove the log message?

packages/cli/scripts/build.ts
```diff
@@ -43,15 +43,15 @@ async function main() {
...
   // Do the initial `ncc` build
-  console.log();
-  const args = [
-    'ncc',
-    'build',
-    '--external',
-    'update-notifier',
-    'src/index.ts',
-  ];
+  console.log('Dependencies:', dependencies);
```
Ethan-Arrowood, line 49: Non-blocking - what if dependencies is empty? Should we still log?

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * NOTE: Since this file runs asynchronously in the background, it's possible
+ * for multiple instances of this file to be running at the same time leading
+ * to a race condition where the most recent instance will overwrite the
+ * previous cache file resetting the `notified` flag and cause the update
+ * notification to appear for multiple consequetive commands. Not the end of
+ * the world, but something to be aware of.
+ */
+
+const fetch = require('node-fetch');
```
Ethan-Arrowood: Can we try `undici` here? As a company we are trying to start moving away from node-fetch and using the fetch included with Node.js (which is undici)

Ethan-Arrowood (approved):
Well done!

packages/cli/src/util/update-notifier/update-worker.js
EndangeredMassa: I agree that's it's more likely that the user will never see this, but a bad connection could still show this log.

We could pass a debug flag in (probably already parsed into a boolean) and only log if that's true. I wouldn't create a new `output` for this.

But if we don't want to debug log this, I'd rather remove the log entirely.

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: I'm fine with switching to undici eventually, but that's out-of-scope for this PR.

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: Bad connection? The IPC channel? I think once it's disconnected, it's disconnected. Maybe it's possible the stdout/stderr file descriptors are still valid even after the child process has disconnected.

Yeah, I'll just rip out the log instead of passing flags and such. Probably one of the best solutions is to log all debug to a file in the update cache directory, but I don't think that's necessary.

EndangeredMassa: I mean a bad internet connection. It could cause the CLI command (like `vc deploy`) to take a while as well as the call to look up the latest CLI version. If both of those take 10+ seconds, this `console.log` would fire.

EndangeredMassa (approved):

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: Ah, now I'm tracking.

FWIW, just as I suspected, even if the IPC channel is closed and the parent exits, the child still has a valid stdout/stderr file descriptors. Setting `stdio` to `"ignore"` will not pass the stdio file descriptors to the child and thus any console logs from the child will not be displayed.

packages/cli/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "preinstall": "node ./scripts/preinstall.js",
-    "test": "jest --env node --verbose --runInBand --bail --forceExit",
+    "test": "jest --env node --verbose --runInBand --bail",
```
cb1kenobi: Removed `--forceExit` because the update notifier is async and the process will out live the test and causes jest to complain.

cb1kenobi:
I've updated the API to be completely async and as such, we have no way of knowing if the update worker process has errored. I did add a lock/pid file to prevent multiple worker processes from running at the same time.

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -73,3 +80,15 @@ if (process.connected && process.send) {
   console.error('No IPC bridge detected, exiting');
   process.exit(1);
 }
+
+function isRunning(lockFile) {
+  try {
+    const pid = parseInt(fs.readFileSync(lockFile, 'utf-8'));
+    process.kill(pid, 0);
+    return true;
+  } catch (e) {
```
EndangeredMassa: **suggestion (soft-blocking):** debug log error

In this or a follow-up PR, can we debug log this error to help with debugging future issues, should they arise?

packages/cli/test/unit/util/update-notifier.test.ts
```diff
@@ -100,6 +90,35 @@ describe('update notifier', () => {
...
+    latest = updateNotifier({
+      cacheDir,
+      pkg,
+      updateCheckInterval: 1,
+    });
+    expect(latest).toEqual(undefined);
+
+    await waitForFiles(join(cacheDir, 'update-notifier'));
+
+    // 3. immediately call again, but should hopefully still be undefined
```
EndangeredMassa: **suggestion (non-blocking):** update comment

This comment looks like a copy-paste error. Should it not say "immediately" or mention the value should be "undefined"?

packages/cli/src/index.ts
```diff
@@ -149,22 +142,29 @@ const main = async () => {
...
-    console.log(
-      info(
-        `${chalk.black.bgCyan('UPDATE AVAILABLE')} ` +
-          `Run ${cmd(
-            await getUpdateCommand()
-          )} to install ${getTitleName()} CLI ${latest}`
-      )
-    );
+  if (isTTY && !process.env.NO_UPDATE_NOTIFIER) {
+    // Checks for available update and returns an instance
```
EndangeredMassa: **suggestion (non-blocking):** update or remove comment

This returns a version, not an instance of something, right?

packages/cli/src/index.ts
```diff
@@ -149,22 +142,29 @@ const main = async () => {
...
+    const latest = updateNotifier({
+      distTag: isCanary ? 'canary' : 'latest',
+      output,
+      pkg,
+    });
```
EndangeredMassa, line 152: **suggestion (non-blocking):** rename `updateNotifier` to something like `getLatestVersion`

Seeing `updateNotifier` in use, I don't think this is a good name for the the script. It's not notifying at all. It's determining if there's a new version, but we still have to notify the user in the consumer code.

The original package could `.notify()`, which makes the original name make sense. In our replaced version, we don't do that.

I think this should be named something like `getLatestVersion`.

packages/cli/test/unit/util/update-notifier.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+async function waitForCacheFile() {
+  for (let i = 0; i < 20; i++) {
+    await sleep(100);
+    try {
+      await fs.readFile(cacheFile, 'utf-8');
+      return;
+    } catch (e) {
+      // cache file may not have been created yet
+    }
```
EndangeredMassa: **suggestion (non-blocking):** don't rely on exceptions

```suggestion
    if (await fs.pathExists(cacheFile)) {
      return;
    }
```

Can we use `pathExists` instead? This would allow any other error that could come up to fail the test.

packages/cli/src/util/update-notifier/update-worker.js
```diff
@@ -0,0 +1,95 @@
...
+  process.send({ type: 'ready' });
+} else {
+  console.error('No IPC bridge detected, exiting');
+  process.exit(1);
+}
+
+async function isRunning(lockFile) {
+  try {
+    const pid = parseInt(await fs.readFile(lockFile, 'utf-8'));
+    process.kill(pid, 0);
```
EndangeredMassa: **suggest (non-blocking):** add comment about signal `0`

I was assuming this actually killed the process until I checked on what signal `0` means. This instead checks for the existence of a process.

Can we add a comment here to make that clear? I didn't expect `process.kill` to not actually try to kill a process.

The [Docs: Signal Events](https://nodejs.org/api/process.html#signal-events) say:

> 0 can be sent to test for the existence of a process, it has no effect if the process exists, but will throw an error if the process does not exist.

Maybe we can say:

```suggestion
    // checks for existence of a process; throws if not found
    process.kill(pid, 0);
```

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: In this case, we don't really care if an error is thrown here. If the file doesn't exist, then no lock. If the file exists, but is empty/invalid, then treat as no lock. If the file exists, but the pid in it isn't running, then we're good to go. If the file exists and the pid is running, then we know we can bail.

cb1kenobi: Also, `process.kill(pid, 0)` is designed to throw if the process is not running.

EndangeredMassa: That's true, but if we need to debug an issue around here, hiding the error that gets thrown will make it harder to figure out what's going on.

I'm always skeptical of completely ignoring errors. I've encountered far too many ignored errors that would have helped me in debugging if the message was available somewhere.

packages/cli/test/unit/util/update-notifier.test.ts
cb1kenobi: Done!

packages/cli/src/index.ts
cb1kenobi: Done!

packages/cli/src/index.ts
cb1kenobi: Fixed!

packages/cli/test/unit/util/update-notifier.test.ts
cb1kenobi: Fixed!

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: I agree that debug logging would be nice in the worker process, however that requires writing the log to a file. I'd be happy to add that if you think it's worth it.

As far as this particular function, I wouldn't be so worried about errors being thrown here. By design, it's meant to catch and return true/false.

packages/cli/src/util/update-notifier/update-worker.js
cb1kenobi: Done!

packages/cli/src/util/update-notifier/update-worker.js
EndangeredMassa: A log file would be nice, but I agree that it's not necessary for this. We could pass the debug flag through the IPC message and debug log it to stdio, though. I still think that would be useful.

---

The intent of the code is one thing, but things can go wrong. If I'm trying to debug an issue with the spawned worker, I would probably comment out the part where it detaches, run a CLI command with debug on, then see the log output of the worker because of the `inherit` io.

Currently, any exception thrown on these lines would be hidden. We could have a bug where writing the lockfile puts in `"234"` instead of `234`, breaking the `parseInt`, for example. We'd have to dig deep to even find the error, but if it was debug logged, we would have seen it a lot sooner in the process.

EndangeredMassa (approved):
I'd still like to see the additional debug logging I commented on, but this PR is useful and mergable now. ✅

cb1kenobi:
Added logging to the get latest worker process. It will write a log file in the cache directory that looks like this:

```
[2022-09-16T22:30:38.661Z] [debug] Notifying parent we're ready
[2022-09-16T22:30:38.664Z] [debug] Received message from parent: {
  cacheFile: '/Users/chris/Library/Caches/com.vercel.cli/package-updates/vercel-latest.json',
  distTag: 'latest',
  name: 'vercel'
}
[2022-09-16T22:30:38.664Z] [debug] Disconnecting from parent
[2022-09-16T22:30:38.665Z] [debug] Checking lock file: /Users/chris/Library/Caches/com.vercel.cli/package-updates/vercel-latest.lock
[2022-09-16T22:30:38.665Z] [debug] Resetting lock file: Error: ENOENT: no such file or directory, open '/Users/chris/Library/Caches/com.vercel.cli/package-updates/vercel-latest.lock'
[2022-09-16T22:30:38.665Z] [debug] Initializing lock file with pid 53421
[2022-09-16T22:30:38.666Z] [debug] Fetching https://registry.npmjs.org/vercel
[2022-09-16T22:30:38.913Z] [debug] Found dist tag "latest" with version "28.2.5"
[2022-09-16T22:30:38.914Z] [debug] Writing cache file: /Users/chris/Library/Caches/com.vercel.cli/package-updates/vercel-latest.json
[2022-09-16T22:30:38.914Z] [debug] Releasing lock file: /Users/chris/Library/Caches/com.vercel.cli/package-updates/vercel-latest.lock
[2022-09-16T22:30:38.915Z] [debug] Worker finished successfully!
```

Also added support for `--debug`:

image

Notice that some of the debug output is printed after the `vercel` command exits. Simply hit `<enter>` to get the prompt to re-render.

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -14,36 +14,101 @@ const fetch = require('node-fetch');
...
+        )} ${str}`
+      );
+    } else if (type === 'error') {
+      console.error(`${red(`Error:`)} ${str}`);
+    }
   }
+
+  setLogFile(file) {
+    if (this.logFile === null) {
+      process.on('exit', () => {
```
EndangeredMassa, line 60: **suggestion (non-blocking):** use `process.once` instead

It looks like the check for `this.logFile === null` is used here to ensure that the exit handler is only registered once. It took me a minute to realize that, though. To clarify, could we do one of the following?

- add a comment explaining the intent of the condition
- store the condition value in a variable named with intent, like `isFirstCall`
- remove the condition entirely and use `process.once('exit', ...`

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -14,36 +14,101 @@ const fetch = require('node-fetch');
...
+          fs.outputFileSync(this.logFile, this.debugLog.join('\n'));
+        }
+      });
+    }
+    this.logFile = file;
+  }
+}
+
+const output = new WorkerOutput({
+  debug: process.argv.includes('--debug') || !process.connected,
```
EndangeredMassa, line 73: **suggestion (non-blocking):** check `process.connected` before debug logging

`process.connected` will always be true here, right? The value changes later after the IPC message is received. Should this value be checked right before debug logging?

`debugOutputEnabled` could become a getter that checks the underlying value as well as `process.connected`. I wouldn't be a fan of the running instance constantly checking global values, but it would work. Wrapping the access to `process.connected` in its own getter would at least allow a test to override the global access behavior.

Example:

```
class WorkerOutput {
  get processConnected() {
    return process.connected;
  }

  get debugOutputEnabled() {
    return this._debugOutputEnabled || !this.processConnected;
  }

  // ...
}
```

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -14,36 +14,101 @@ const fetch = require('node-fetch');
...
   // check for a lock file and either bail if running or write our pid and continue
   const lockFile = path.join(
     cacheFileParsed.dir,
     `${cacheFileParsed.name}.lock`
   );
+  output.debug(`Checking lock file: ${lockFile}`);
   if (await isRunning(lockFile)) {
-    return;
+    output.debug('Worker already running, exiting');
+    process.exit(2);
```
EndangeredMassa: **question (non-blocking):** Does this need to exit instead of return?

**suggestion (non-blocking):** change exit code to `1`

In a team meeting, we aligned on using these exit codes.

- 0: success
- 1: internal system error
- 2: help output
- 4: user error
- 5: external system error (ex: API Error)

While not official, currently followed, or enforced, if we want to start moving towards this convention, we can start here by changing the exit code of `2` to `1`.

EndangeredMassa (approved):
I really like the new debug logging. I have a few suggestions.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: `process.once()` won't work here because it's not a matter of the event being fired more than once, but rather we only want to wire up the exit handler once.

No need to define a variable when a comment will work.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: There are 2 ways the worker script can be invoked:

1. Manually run `node /path/to/get-latest-worker.js`. For this we force debug logging so that we see any unhandled rejection errors, of which should never happen. Manually running the script should just print `"No IPC bridge detected, exiting"`. If we don't care about unhandled rejections, which should never occur, we could simply remove the `process.connected` all together.

2. The worker is spawned by the CLI, `process.connected` will be `true` and `debug` will be `false` by default. If `--debug` is set, then the debug log is printed to `stderr`.

We don't need to overcomplicate the debug logging with checking `process.connected` because the debug logger doesn't care if the worker process is connected or not. The `stderr` file descriptor is still valid even after the processes disconnect.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: Yes, we need an exit. I found that without an explicit exit here, the 10 second time was firing despite clearing the timer. I'm not sure what exactly was keeping the process from exiting, but this "already running" check was added after I realized I needed the exit calls.

Good call on the exit code. Fixed!

packages/cli/src/util/get-latest-version/get-latest-worker.js
EndangeredMassa: I was reading too much intent behind the check. This sounds good to me.

EndangeredMassa (approved):

packages/cli/src/index.ts
EndangeredMassa: Because of the `isTTY` check, this won't break piped output, but by convention and for future testing, we should still probably change these to use `output` in a follow-up PR.

TooTallNate (approved):
I'll admit that I'm confused about how the worker detaching logic is working, because `child.unref()` in the Node.js docs always is accompanied by using `detached: true` in the spawn() function. But after testing locally it does seem to work as expected.

I do think the IPC stuff is unnecessary, and we could pass that information over argv or env vars instead. But I'm approving as-is and we can follow-up in the future if necessary.

cb1kenobi:
> I'll admit that I'm confused about how the worker detaching logic is working, because `child.unref()` in the Node.js docs always is accompanied by using `detached: true` in the spawn() function. But after testing locally it does seem to work as expected.

Since we're using IPC, we don't want to detach right away. The key is having the child process do a `process.disconnect()` so that the child can out live the parent process. The parent still needs to `worker.unref()`.

> I do think the IPC stuff is unnecessary, and we could pass that information over argv or env vars instead. But I'm approving as-is and we can follow-up in the future if necessary.

While creating an IPC channel and doing a handshake does seem complex, I've been burned in the past with spawning a child process and passing data in via the command line args or env vars. IPC is a good way to pass data to another process and you don't have to worry about data truncation, size limitations, malformed data, and it makes it super easy if we ever need to add more data as we're dealing with a simple JSON object where Node.js will take care of the serialization/deserialization. I have built this exact IPC channel architecture several times and it's robust. Neither here nor there, if you prefer a non-IPC solution, I'd be happy to refactor this code again.

TooTallNate:
I don't think a refactor is necessary :shipit: 

Looks like the Dev tests are timing out now though, I'm guessing related to the removal of `--forceExit` on Jest.

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -0,0 +1,142 @@
...
+ * version, otherwise `undefined`.
+ */
+export default function getLatestVersion({
+  cacheDir = XDGAppPaths('com.vercel.cli').cache(),
+  distTag = 'latest',
+  output,
+  pkg,
+  getLatestInterval,
+}: GetLatestVersionOptions): string | undefined {
+  const cacheFile = resolvePath(
```
styfle, line 54: I think we want to use `path.join()`, not `path.resolve()`

https://stackoverflow.com/a/39836259/266535

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  try {
+    cache = readJSONSync(cacheFile);
+  } catch (err: any) {
+    // cache does not exist or malformed
+    if (err.code !== 'ENOENT') {
+      output?.debug(`Error reading latest package cache file: ${err}`);
+    }
+  }
+
+  if (!cache || cache.expireAt < Date.now()) {
```
styfle, line 61: Should this be `>` instead?

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @returns {String|undefined} If a newer version is found, then the lastest
+ * version, otherwise `undefined`.
+ */
+export default function getLatestVersion({
+  cacheDir = XDGAppPaths('com.vercel.cli').cache(),
+  distTag = 'latest',
+  output,
+  pkg,
+  getLatestInterval,
```
styfle: What is the purpose of `getLatestInterval`?

Its not being passed in by the `src/index.ts` caller.

packages/cli/src/util/get-latest-version/index.ts
cb1kenobi: In this case, it doesn't really matter. Considering `cacheDir` is sourced elsewhere, the worker script should resolve it as it could be a relative path.

The Stack Overflow post is highlighting an issue where arguments aside from the first argument begin with a `/` which could only happen if `pkg.name` starts with a `/`, but then the package name wouldn't be a valid package name.

packages/cli/src/index.ts
```diff
@@ -149,22 +142,30 @@ const main = async () => {
...
-          )} to install ${getTitleName()} CLI ${latest}`
-      )
-    );
+  if (isTTY && !process.env.NO_UPDATE_NOTIFIER) {
+    // Check if an update is available. If so, `latest` will contain a string
+    // of the latest version, otherwise `undefined`.
+    const latest = getLatestVersion({
+      distTag: isCanary ? 'canary' : 'latest',
+      output,
+      pkg,
```
styfle, line 151: Lets pass in `updateCheckInterval` like we were before

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,188 @@
...
+    if (version) {
+      output.debug(`Found dist tag "${distTag}" with version "${version}"`);
+    } else {
+      output.error(`Dist tag "${distTag}" not found`);
+      output.debug('Available dist tags:', Object.keys(tags));
+    }
+
+    output.debug(`Writing cache file: ${cacheFile}`);
+    await fs.outputJSON(cacheFile, {
+      expireAt: Date.now() + (getLatestInterval || defaultGetLatestInterval),
```
styfle: Lets call this `updateCheckInterval` instead

packages/cli/src/util/get-latest-version/index.ts
cb1kenobi: Just an option. If it's not beneficial, it can be removed in favor of a constant.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: 100%. `getLatestInterval` is a horrible name. Naming is hard.

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,189 @@
...
+  // was directly executed
+  debug: process.argv.includes('--debug') || !process.connected,
+});
+
+process.on('unhandledRejection', err => {
+  output.error('Exiting worker due to unhandled rejection:', err);
+  process.exit(1);
+});
+
+const defaultUpdateCheckInterval = 1000 * 60 * 60 * 24 * 7; // 1 week
```
styfle: @cb1kenobi This default is no longer needed

packages/cli/test/unit/util/get-latest-version.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      cacheDir,
+      pkg,
+    });
+    expect(latest).toEqual(undefined);
+
+    await waitForCacheFile();
+
+    let cache = await fs.readJSON(cacheFile);
+    expect(typeof cache).toEqual('object');
+    expect(typeof cache.expireAt).toEqual('number');
```
styfle, line 37:
```suggestion
    expect(cache.expireAt).toBeGreaterThan(Date.now());
```

packages/cli/test/unit/util/get-latest-version.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  it('should not check twice', async () => {
+    // 1. first call, no cache file
+    let latest = getLatestVersion({
+      cacheDir,
+      updateCheckInterval: 1,
+      pkg,
+    });
+    expect(latest).toEqual(undefined);
+
+    // 2. immediately call again, but should hopefully still be undefined
```
styfle, line 98: "hopefully" can probably be "always" if you increase `updateCheckInterval` from 1ms to 1000ms

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,189 @@
...
+
+// this timer will prevent this worker process from running longer than 10s
+const timer = setTimeout(() => {
+  output.error('Worker timed out after 10 seconds');
+  process.exit(1);
```
styfle, line 84: I think this needs to clean up the lock file. Otherwise the user will never be able to update again.

I'm wondering if we even need a lock file at all 🤔

packages/cli/test/unit/util/get-latest-version.test.ts
cb1kenobi: Fixed, though I left the `typeof` check in there.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: I have moved the default out of the worker and into the `getLatestVersion()` options defaults.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: This `setTimeout()` exists in the event the parent never sends the message, thus no lock file will be created to be cleaned up. The lock file name is derived from the cache file name, which is the package name plus the dist tag. Those two bits of info are not available at the time the timer is created.

We could rename `timer` to `waitForParentTimer`, then clear it once it gets the message to start, then create an `AbortController` to make sure the `fetch()` returns in a reasonable time.

Lock files are only 4KB on disk and I have yet to come across a scenario where the lock file wasn't cleaned up. I'm sure it will happen at some point and that's why I added the stale pid check in `isRunning()`.

As for needing the lock file, I did some testing and it's possible to run `vercel` commands fast enough that one worker is clobbering the cache file with stale data (e.g. the `notified` flag being toggled). Since the access to the cache file should be atomic, a lock file will synchronize multiple processes ensuring a predictable state. Obviously we're not building a database, so if you feel the lock file, or the whole IPC design for that matter, is overkill, I'd be happy to refactor the code.

packages/cli/src/index.ts
```diff
@@ -149,22 +142,31 @@ const main = async () => {
...
+      updateCheckInterval: 1000 * 60 * 60 * 24 * 7, // 1 week
```
```suggestion
```

I guess we dont need to pass it in since the default param exists now

cb1kenobi: Fixed!

packages/cli/scripts/build.ts
EndangeredMassa: Can we update this in a follow-up PR?

We probably don't want to log if `dependencies` is empty. Or, at least we'd want to log a clearer message saying so.

packages/cli/src/index.ts
EndangeredMassa: Can we update this in a follow-up PR?

This was about using `output` instead of `console.log`.

EndangeredMassa (approved):
I think we should do a release before this to reduce the surface area of a new version. Then we can release this.

packages/cli/scripts/build.ts
cb1kenobi: On one hand, this is a build script and if there are no dependencies, then I'd think we'd want to see that.

On the other hand, `dependencies` will likely never be empty.

PR #8091, opened 2022-07-07
[cli] Refactor `vc project`

MatthewStanciu:
Since the `vc project` command is about to be expanded with 2 new subcommands, I think it makes sense to do a little bit of refactoring. The current `vc project` command is all in one file, with the logic for subcommands nested within if statements. I think the structure of `vc project` should look more like `vc env`, which is consistent with how commands with subcommands look throughout the rest of the codebase.

This PR moves the logic for the `project` subcommands into their own files.

#### Tests

#### Code Review


TooTallNate (approved):

packages/cli/test/helpers/parse-table.ts
```diff
@@ -0,0 +1,19 @@
+export function getDataFromIntro(output: string): {
```
EndangeredMassa, line 1: This file name and location makes it look like a general helper, but it contains logic specific to `packages/cli/test/unit/commands/project.test.ts`. Can we either rename this file to something like `parse-project-table.ts` and/or move it to sit next to the test where it's used?

If we move this file next to the test, it shouldn't be picked up by the test runner as a test because it won't fit the `*.test.ts` pattern. So, that shouldn't be a problem.

packages/cli/src/index.ts
```diff
@@ -653,7 +653,7 @@ const main = async () => {
         func = require('./commands/logout').default;
         break;
       case 'projects':
-        func = require('./commands/projects').default;
```
EndangeredMassa, line 656: Is the official command supposed to be `vc projects` or `vc project`? I think it's reasonable that both commands would work, but we should keep consistent on what the official one is. It looks like "projects" so far.

If that's so, can this directory match the name? I'd move it to `./commands/projects`.

packages/cli/test/helpers/read-output-stream.ts
```diff
@@ -0,0 +1,23 @@
+import { MockClient } from '../mocks/client';
+
+export function readOutputStream(
```
EndangeredMassa, line 3: The need for this was created by removing `outputBuffer` in: https://github.com/vercel/vercel/pull/8052

I think we should add `outputBuffer` (or something like it) back in. I created a PR to do just that, which would remove this file. https://github.com/vercel/vercel/pull/8108

packages/cli/src/index.ts
MatthewStanciu: It seems the jury is still out, but I've been told to use `project` in the past, so I've been moving to that. Agree it should be consistent though, I'll change what's left of `projects` over to `project`.

packages/cli/test/helpers/parse-table.ts
MatthewStanciu: I moved these helpers into their own file because it's used in `vc project ls` as well as `vc ls`. For this reason I'm not sure if it's a good idea to move them next to `project.test.ls`. I agree though that it could benefit from more specific naming.

EndangeredMassa: Ahh, right. They are used in multiple places.

Agreed. Just a name update would be good.

PR #8093, opened 2022-07-07
[@vercel/node] favor workPath over repoRootPath

Schniz:
Middleware in monorepos are getting the wrong name, instead of being a simple `middleware`
the name provided is `[full_path_to]/middleware`

This PR changes the `baseDir` to favor `workPath` over `repoRootPath`

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -339,7 +339,7 @@ export const build: BuildV3 = async ({
   config = {},
   meta = {},
 }) => {
-  const baseDir = repoRootPath || workPath;
+  const baseDir = workPath || repoRootPath;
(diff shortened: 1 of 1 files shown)


styfle (changes requested):
This will certainly break existing deployments.

The `workPath` is always defined. Its only `repoRootPath` that might be undefined.

Furthermore, `nodeFileTrace` must use the repo root as the base or else it wont be able to trace files higher up in the repo.

I suggest changing the middleware code to handle monorepos rather than change all `@vercel/node`.

TooTallNate:
Let's go with https://github.com/vercel/vercel/pull/8097 instead.

PR #8094, opened 2022-07-07
[cli] Fix `vc deploy` erroring when `.git` is corrupt

MatthewStanciu:
### Related Issues

- Fixes https://github.com/vercel/customer-issues/issues/597

`vc deploy` currently fails with a confusing error if the user has a corrupt `.git` directory. This is caused by `create-git-meta` improperly handling the scenario where it detects a `.git` directory but cannot find git information. This PR handles this error.

#### Tests

#### Code Review


styfle (commented):
How about adding a test with a corrupt `.git` directory (or whatever throws)?

MatthewStanciu:
It's difficult to test for the exact scenario since the test itself is running in a Git repository, but I think I've added a test that still checks for what matters.

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -125,4 +126,28 @@ describe('createGitMeta', () => {
...
     }
   });
+  it('fails when .git is corrupt)', async () => {
+    const directory = fixture('git-corrupt');
+    try {
+      await fs.rename(join(directory, 'git'), join(directory, '.git'));
+
+      const data = await createGitMeta(directory, client.output);
+
+      // `getLastCommit()` runs `git log -1`, which searches parent directories as well.
```
TooTallNate: How about if we copy `directory` to a temp directory i.e. `path.join(os.tmpdir(), 'git-corrupt')`, so that the fixture is "outside" of `vercel/vercel`.

MatthewStanciu: Splendid idea

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -125,4 +126,28 @@ describe('createGitMeta', () => {
       await fs.rename(join(directory, '.git'), join(directory, 'git'));
     }
   });
+  it('fails when .git is corrupt)', async () => {
```
TooTallNate:
```suggestion
  it('fails when `.git` is corrupt', async () => {
```

packages/cli/src/util/create-git-meta.ts
```diff
@@ -6,26 +6,26 @@ import { exec } from 'child_process';
...
   });
 }
 
-function getLastCommit(directory: string): Promise<git.Commit> {
-  return new Promise((resolve, reject) => {
+function getLastCommit(directory: string): Promise<git.Commit | void> {
+  return new Promise(resolve => {
     git.getLastCommit(
       (err, commit) => {
-        if (err) return reject(err);
```
TooTallNate: I think rejecting here is still more "proper". We can ignore the error in a `.catch()` on line 77 instead.

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -6,22 +6,22 @@ import { exec } from 'child_process';
...
         );
+        return resolve(false);
+      }
       resolve(stdout.trim().length > 0);
     });
   });
 }
 
-function getLastCommit(directory: string): Promise<git.Commit> {
+function getLastCommit(directory: string): Promise<git.Commit | void> {
```
TooTallNate:
```suggestion
function getLastCommit(directory: string): Promise<git.Commit> {
```

Is this return type still valid?

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -64,10 +64,14 @@ export async function createGitMeta(
     return;
   }
   const [commit, dirty] = await Promise.all([
-    getLastCommit(directory),
-    isDirty(directory),
+    getLastCommit(directory).catch(() => {}),
```
TooTallNate: I suggest returning `null` or `undefined` here. The `() => {}` syntax confuses my brain and I thought it was an empty object at first.

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -125,4 +126,17 @@ describe('createGitMeta', () => {
...
+    try {
+      await fs.rename(join(directory, 'git'), join(directory, '.git'));
+      await fs.copy(directory, join(os.tmpdir(), 'git-corrupt'));
+      const cpDirectory = join(os.tmpdir(), 'git-corrupt');
+
+      const data = await createGitMeta(cpDirectory, client.output);
+      expect(data).toBeUndefined();
+    } finally {
+      await fs.rename(join(directory, '.git'), join(directory, 'git'));
+    }
```
TooTallNate: Let's do the rename part in the tmp dir, and then clean up afterwards:

```suggestion
    const directory = fixture('git-corrupt');
    const tmpDir = join(os.tmpdir(), 'git-corrupt');
    try {
      // Copy the fixture into a temp dir so that we don't pick
      // up Git information from the `vercel/vercel` repo itself
      await fs.copy(directory, tmpDir);
      await fs.rename(join(tmpDir, 'git'), join(tmpDir, '.git'));

      const data = await createGitMeta(tmpDir, client.output);
      expect(data).toBeUndefined();
    } finally {
      await fs.remove(tmpDir);
    }
```

MatthewStanciu: This is much better, thanks

TooTallNate (approved):

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -64,10 +64,16 @@ export async function createGitMeta(
     return;
   }
   const [commit, dirty] = await Promise.all([
-    getLastCommit(directory),
-    isDirty(directory),
+    getLastCommit(directory).catch(() => {
+      return;
```
EndangeredMassa, line 68: Can we debug log this error?

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -6,16 +6,16 @@ import { exec } from 'child_process';
...
+export function isDirty(directory: string, output: Output): Promise<boolean> {
+  return new Promise(resolve => {
     exec('git status -s', { cwd: directory }, function (err, stdout, stderr) {
-      if (err) return reject(err);
-      if (stderr)
-        return reject(
-          new Error(
-            `Failed to determine if git repo has been modified: ${stderr.trim()}`
-          )
+      if (err) return resolve(false);
```
EndangeredMassa, line 12: Can we debug log this error?

PR #8097, opened 2022-07-07
[node] Provide correct `middlewarePath` when using "Root Directory" setting

TooTallNate:
The `middlewarePath` property was not correctly stripping the root directory setting from the value, causing Middleware to not work correctly with Root Directory.


TooTallNate:
Merged because tests passing and needed to get this included as part of the v27 CLI release.

packages/cli/test/unit/commands/build.test.ts
```diff
@@ -423,6 +423,67 @@ describe('build', () => {
...
+          { handle: 'filesystem' },
+          { src: '^/api(/.*)?$', status: 404 },
+          { handle: 'error' },
+          { status: 404, src: '^(?!/api).*$', dest: '/404.html' },
+        ],
+      });
+
+      // "static" directory contains `index.html`, but *not* `middleware.js`
+      const staticFiles = await fs.readdir(join(output, 'static'));
+      expect(staticFiles.sort()).toEqual(['index.html']);
```
EndangeredMassa, line 476: Should we add an assertion that `middleware.js` is not in `staticFiles`?

PR #8102, opened 2022-07-07
[cli] `vc ls` visual improvements

MatthewStanciu:
Featuring:

- State string now includes all states, including `QUEUED`, `INITIALIZING`, `BUILDING`, `DEPLOYING`, and `ANALYZING`
- `includeScheme` is no longer considered; deployment URLs now always include `https://` and are clickable
- Removed double backticks in "To list more deployments for a project" message

Before:

Screen Shot 2022-07-13 at 6 29 15 PM

After:

Screen Shot 2022-07-13 at 6 29 20 PM

#### Tests

#### Code Review


packages/cli/src/commands/list.ts
```diff
@@ -326,16 +322,23 @@ function getProjectName(d: Deployment) {
 
 // renders the state string
 export function stateString(s: string) {
-  switch (s) {
+  const CIRCLE = '● ';
+  // make `s` title case
+  s = `${s.substring(0, 1)}${s.toLowerCase().substring(1)}`;
```
EndangeredMassa: Because we need both values, I don't think we should change the value of `s`, just to temporarily revert it with `.toUpperCase`. When reading this logic, I feel like I'm missing some detail of why this is necessary.

How about leaving `s` as is and making a new variable to hold the titlecase version?

packages/cli/src/commands/list.ts
```diff
@@ -326,16 +322,23 @@ function getProjectName(d: Deployment) {
...
     case 'READY':
-      return s;
-
+      return chalk.green(CIRCLE) + s;
+    case 'QUEUED':
+      return chalk.white(CIRCLE) + s;
+    case 'CANCELED':
+      return chalk.gray(s);
     default:
       return chalk.gray('UNKNOWN');
```
EndangeredMassa, line 336: Do we want `UNKNOWN` to still be uppercase?

packages/cli/test/mocks/deployment.ts
```diff
@@ -26,13 +46,15 @@ export function useDeployment({
...
     createdIn: 'sfo1',
+    buildingAt: Date.now(),
     ownerId: creator.id,
     creator: {
       uid: creator.id,
       email: creator.email,
       username: creator.name,
     },
     readyState: 'READY',
+    state: states[Math.floor(Math.random() * states.length)],
```
EndangeredMassa: I prefer to not use randomness in tests. It can often be fine, but will sometime cause a test to pass or fail unpredictably.

EndangeredMassa (approved):

packages/cli/src/commands/list.ts
MatthewStanciu: I like it. To me it indicates that it's not a normal state, which is what I think the "unknown" state should indicate. But happy to make it title case if that's confusing.

packages/cli/test/mocks/deployment.ts
```diff
@@ -12,12 +12,32 @@ export function useDeployment({
...
+    'ERROR',
+    'CANCELED',
+    'QUEUED',
+  ];
+
   const deployment: Deployment = {
     id: `dpl_${chance().guid()}`,
     url: url.hostname,
-    name: '',
+    name: `dpl_${chance().guid()}`,
```
styfle:
```suggestion
    name: chance().name(),
```

packages/cli/test/mocks/deployment.ts
```diff
@@ -12,12 +12,32 @@ export function useDeployment({
 }: {
   creator: Pick<User, 'id' | 'email' | 'name'>;
 }) {
+  type state =
```
styfle: This should be capital `State` since its a type. Also you can probably use the existing type from `@vercel/client`

```ts
type State = Deployment['readyState'];
```

packages/cli/src/commands/list.ts
EndangeredMassa: I agree that either approach is fine. I wasn't sure if this was intentional or not.

packages/cli/src/commands/list.ts
styfle: Agreed. We also have a module designed for title case

```ts
import title from 'title';
```

MatthewStanciu: TIL!

packages/cli/test/mocks/deployment.ts
styfle: Agreed. Also, why would state and readyState be different?

packages/cli/test/unit/commands/list.test.ts
```diff
@@ -35,11 +35,13 @@ describe('list', () => {
 
       await list(client);
 
-      const output = await readOutputStream(client);
+      const output = await readOutputStream(client, 4);
+      console.log(output);
```
styfle: Is this console.log() supposed to be here?

MatthewStanciu: Nope, thanks for catching that

EndangeredMassa (approved):

Ethan-Arrowood (approved):

MatthewStanciu:
Closing in favor of #8151

PR #8108, opened 2022-07-08
[tests] restore outputBuffer

EndangeredMassa:
In the PR https://github.com/vercel/vercel/pull/8052, the `client.outputBuffer` was removed in favor of a new `toOutput` matchers: https://github.com/vercel/vercel/pull/8052/files#diff-0838ea92c3dfb1e78ce8b2950b7f01ad0be324bd615ab1287b5ef5bc14a141b7L85-L87

This works really well for streaming output until you get a match. If you want to gather up the full output and parse it, however, you have to reimplement something like `outputBuffer` anyway.

I think we should bring back `outputBuffer` (or some other approach that as the same effect. This PR does a simple restore of that functionality.


EndangeredMassa:
Discussed this with @TooTallNate. We probably want a more complete solution.

PR #8112, opened 2022-07-08
[cli] Debug log error messages in `create-git-meta`

MatthewStanciu:
This is a follow-up to #8094 which debug logs errors.

#### Tests

#### Code Review


packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -65,6 +66,9 @@ export async function createGitMeta(
   }
   const [commit, dirty] = await Promise.all([
     getLastCommit(directory).catch(() => {
+      output.debug(
+        'Failed to get last commit. The directory is likely not a Git repo, there are no latest commits, or it is corrupted.'
```
EndangeredMassa: Can we include the error message at the end?

packages/cli/src/util/deploy/create-git-meta.ts
```diff
@@ -9,10 +9,11 @@ import { Output } from '../output';
 export function isDirty(directory: string, output: Output): Promise<boolean> {
   return new Promise(resolve => {
     exec('git status -s', { cwd: directory }, function (err, stdout, stderr) {
-      if (err) return resolve(false);
-      if (stderr) {
+      if (err || stderr) {
         output.debug(
-          `Failed to determine if git repo has been modified: ${stderr.trim()}`
+          `Failed to determine if git repo has been modified:\n${err}\n${
+            stderr && stderr.trim()
```
EndangeredMassa: if `err` is `null`, there will be two newlines here. Can we optionally add these values to the resulting string?

EndangeredMassa (approved):

TooTallNate:
Let's add some unit tests?

EndangeredMassa (approved):

PR #8113, opened 2022-07-08
[cli] Explicitly use `vc project` vs. `vc projects`

MatthewStanciu:
This is a follow-up to #8091 which:

- Makes `vc project` the default command, with `vc projects` aliased to `vc project` (previously it was not clear in the code which one was the "real" command)
- Makes some helper names for `ls` more specific

#### Tests

#### Code Review


packages/cli/src/commands/index.ts
```diff
@@ -25,8 +25,8 @@ export default new Map([
   ['logout', 'logout'],
   ['logs', 'logs'],
   ['ls', 'list'],
-  ['project', 'projects'],
-  ['projects', 'projects'],
+  ['project', 'project'],
+  ['projects', 'project'],
```
styfle, line 29: Does this affect the end user? I think the answer is "no" but just wanted to check.

MatthewStanciu: No, the end user can still use `vc project` and `vc projects` interchangeably. The only thing that has changed here is that `vc project` is the "real" command.

packages/cli/test/helpers/parse-table.ts
```diff
@@ -11,7 +11,7 @@ export function getDataFromIntro(output: string): {
   };
 }
 
-export function parseTable(output: string): string[] {
```
EndangeredMassa, line 14: This function parses a row more than a table. What do you think of `parseSpacedTableRow`?

packages/cli/src/commands/index.ts
styfle: I have been thinking that these plural commands should all be changed to singular. Then the singular should print help instead of assuming list.

So something like:

- `vc project` is the same as `vc project --help`
- `vc projects` is the same as `vc project ls`
- `vc project ls` list the projects
- `vc project connect` connect a project

styfle (approved):
I have been thinking that we should change `vc project` to print

packages/cli/test/helpers/parse-table.ts
```diff
@@ -1,4 +1,4 @@
-export function getDataFromIntro(output: string): {
```
EndangeredMassa, line 1: I think this should be even more specific to what it does. It looks for `Deployments for`, for example. What about `pluckIdentifiersFromDeploymentList`? (I can't think of a great name for this.)

MatthewStanciu: It's a little long but I think it's great

PR #8127, opened 2022-07-11
[next][build-utils] allow to declare assets in EdgeFunction constructor

Schniz:
There's a [Next.js PR](https://github.com/vercel/next.js/pull/38492) to include binary assets in Edge Functions. This PR enables to declare assets in the `EdgeFunction` constructor.

This is a configuration level setting, compared to WebAssembly modules which the `middleware-manifest.json` converts into `import` statements. This is because the asset generation is done in Webpack and it is not dynamic, it's a static string. We can use AST to replace it but we're risking in:

* worse performance
* accidentally replacing a string that isn't an asset


packages/build-utils/src/edge-function.ts
```diff
@@ -33,12 +33,18 @@ export class EdgeFunction {
...
+  blobs?: { name: string; path: string }[];
+
   constructor(params: Omit<EdgeFunction, 'type'>) {
     this.type = 'EdgeFunction';
     this.name = params.name;
     this.deploymentTarget = params.deploymentTarget;
     this.entrypoint = params.entrypoint;
     this.files = params.files;
     this.envVarsInUse = params.envVarsInUse;
+    this.blobs = params.blobs;
```
styfle: Is this going to be confusing calling this `blob` because `files` can have `FileBlob` types? Seems like we should call this something else like `assets`?

packages/next/src/utils.ts
```diff
@@ -2234,6 +2235,23 @@ export async function getMiddlewareBundle({
...
+              const blobFiles = (edgeFunction.blobs ?? []).reduce(
+                (acc: Files, { filePath, name }) => {
+                  const fullFilePath = path.join(
+                    entryPath,
+                    outputDirectory,
+                    filePath
+                  );
+                  acc[`blobs/${name}`] = new FileFsRef({
+                    mode: 0o644,
+                    contentType: 'application/octet-stream',
```
styfle, line 2247: Is this always octet stream? What about if the asset is `./file.txt`?

styfle (approved):

ijjk (approved):

PR #8128, opened 2022-07-11
[fs-detectors] process detectFramework in parallel

pbt:
The Vercel API's `detect-framework` API contacts GitHub's API using its own file adapter. 

However, currently the `detectFramework` function in the Vercel CLI (on which the API depends) calls the file system adapter in series using a `for` loop. This results in large amounts of lag when in a networked situation such as calling GitHub's API as each API call takes a couple hundred milliseconds.

I propose changing `detectFramework` to process the directories it scans in parallel.

Linked to: https://github.com/vercel/turbo-internal/issues/193

#### Tests

#### Code Review

### DIFF
--- a/packages/fs-detectors/src/detect-framework.ts
+++ b/packages/fs-detectors/src/detect-framework.ts
@@ -91,7 +91,7 @@ export async function detectFramework({
         throw new Error(`Framework not detected: "${frameworkMatch}"`);
       })
     );
-  } catch (errorOrErrors: any) {
+  } catch (errorOrErrors: unknown) {
     // note the promise.any polyfill works differently from the native Promise.any!
     // it returns an array of errors instead of an `AggregateError`
     if (Array.isArray(errorOrErrors)) {
--- a/packages/fs-detectors/src/detect-framework.ts
+++ b/packages/fs-detectors/src/detect-framework.ts
@@ -77,11 +77,6 @@ async function matches(fs: DetectorFilesystem, framework: BaseFramework) {
   return result.every(res => res === true);
 }
 
-// "error" message used to reject the promise if no framework is detected
-// under the current path
-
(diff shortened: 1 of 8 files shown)


agadzik (approved):
This looks good to me - I'll defer to the owners to see if theres anything odd about this PR!

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -76,15 +77,37 @@ async function matches(fs: DetectorFilesystem, framework: BaseFramework) {
...
+  try {
+    framework = await Promise.any(
+      frameworkList.map(async framework => {
+        if (await matches(fs, framework)) {
+          return framework.slug;
+        }
+        throw new Error(FRAMEWORK_ERR);
+      })
+    );
+  } catch (errorOrErrors: any) {
```
styfle:
```suggestion
  } catch (errorOrErrors: unknown) {
```

styfle (dismissed):
This will likely cause bugs because the order of the match matters and the order of the promise resolution is no longer guaranteed with Promise.any(). For example, the user might have both `next` and `storybook` in their package.json `dependencies`. Since `next` is the first in the array, it should match first, regardless of other matches (there can only be one framework selected afterall).

agadzik:
> This will likely cause bugs because the order of the match matters. For example, the user might have `next` and they might have `storybook` in their package.json `dependencies`, however since `next` is the first in the array, it should match.

🤔 Maybe we switch to `Promise.all` if thats the case and then return back the first successful result...The con of this approach is that we make more network calls (one for each framework) that will 404, but it speeds up the overall method.

pbt:
Will benchmark on `vercel/api` and report back on the results in a few hours.

styfle (approved):
Thanks, looks great! 🥇

PR #8131, opened 2022-07-11
[cli] Remove `DEBUG=corepack` env var

styfle:
This debug log was originally added in #7871 because corepack has no output by default. In particular, it was nice to see the first deployment was not stalled when the package manager is being installed.

That being said, this gets noisy really fast because cache detections also print a log line.
For example, here's a deployment that prints 3 times:

```
Detected ENABLE_EXPERIMENTAL_COREPACK=1 and "npm@8.10.0" in package.json
Running "install" command: `npm install --prefix=.. --no-audit --engine-strict=false`...
2022-07-11T18:27:00.696Z corepack Reusing npm@8.10.0
356 packages are looking for funding
Running "npm run vercel-build"
2022-07-11T18:27:06.664Z corepack Reusing npm@8.10.0
> front@0.0.0 vercel-build
> npm run buildonly && npm run build:rss
2022-07-11T18:27:07.088Z corepack Reusing npm@8.10.0
> front@0.0.0 buildonly
> next build
```

I think its best to let users add this env var themselves if they want to debug what corepack is doing, so this PR removes that environment variable.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -336,8 +336,7 @@ export default async function main(client: Client): Promise<number> {
   const buildResults: Map<Builder, BuildResult> = new Map();
   const overrides: PathOverride[] = [];
   const repoRootPath = cwd;
-  const rootPackageJsonPath = repoRootPath || workPath;
-  const corepackShimDir = await initCorepack({ cwd, rootPackageJsonPath });
+  const corepackShimDir = await initCorepack({ repoRootPath });
 
   for (const build of builds) {
     if (typeof build.src !== 'string') continue;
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -336,8 +336,7 @@ export default async function main(client: Client): Promise<number> {
   const buildResults: Map<Builder, BuildResult> = new Map();
   const overrides: PathOverride[] = [];
   const repoRootPath = cwd;
-  const rootPackageJsonPath = repoRootPath || workPath;
```
EndangeredMassa, line 339: Is `workPath` never what we want here, then?

EndangeredMassa (approved):

MatthewStanciu (approved):

packages/cli/src/commands/build.ts
styfle: Yeah because you can see on line 338 that `repoRootPath` is never undefined. So `repoRootPath` and `rootPackageJsonPath` were always equal. So I removed the useless line 🙂

EndangeredMassa: Oh, hah, indeed.

PR #8135, opened 2022-07-11
[tests] Add retry for 3rd party builder test

styfle:
These tests fail occasionally on GH Actions due to flakey network IO, so this PR adds retry logic.

https://github.com/vercel/vercel/runs/7291444541?check_suite_focus=true#step:13:1774


MatthewStanciu (approved):
Thank you for your service

EndangeredMassa (approved):

styfle:
One problem I found is that `async-retry` doesn't seem to cancel so this doesn't actually solve the hanging fetch.

We probably need to use `vm` or `child_process` to isolate this test, kill the process once it exceeds the timeout, then retry.

PR #8142, opened 2022-07-12
[remix] Fix `isMonorepo` check

TooTallNate:
`config.projectSettings` is not available when running `@vercel/remix` directly (i.e. when not using `vc build`), so calculate the correct root directory value based on the `workPath` relative to the `repoRootPath` value.

### DIFF
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -192,6 +192,8 @@ export const build: BuildV2 = async ({
     // If we are, prepend the app root directory from config onto the build path.
     // e.g. `/apps/my-remix-app/api/index.js`
     const isMonorepo = repoRootPath && repoRootPath !== workPath;
+    console.log('config', config);
+    console.log('config.projectSettings', config.projectSettings);
     if (isMonorepo && config.projectSettings?.rootDirectory) {
       serverBuildPath = join(
         config.projectSettings.rootDirectory,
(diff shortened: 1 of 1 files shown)


packages/remix/src/build.ts
```diff
@@ -192,11 +192,9 @@ export const build: BuildV2 = async ({
...
     // e.g. `/apps/my-remix-app/api/index.js`
     const isMonorepo = repoRootPath && repoRootPath !== workPath;
-    if (isMonorepo && config.projectSettings?.rootDirectory) {
-      serverBuildPath = join(
-        config.projectSettings.rootDirectory,
-        serverBuildPath
-      );
+    if (isMonorepo) {
+      const rootDirectory = relative(repoRootPath, workPath);
+      serverBuildPath = join(rootDirectory, serverBuildPath);
```
styfle, line 197: I think it should be sufficient to say that `rootDirectory` is the same as `workPath` when `isMonorepo` is true.

So this could be one line:

```js
serverBuildPath = join(workPath, serverBuildPath);
```

styfle (approved):

packages/remix/src/build.ts
TooTallNate: Keeping it as-is because we want to pass a relative path, not absolute path.

PR #8100, opened 2022-07-07
[cli] Add ability to connect Git provider repo

MatthewStanciu:
This PR adds a new subcommand `vc git`, which allows you to create a Git provider repository from your local `.git` config to your Vercel project.

Usage:

- `vc git connect` – connects a Git provider repository
- `vc git disconnect` – disconnects a Git provider repository

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8145, opened 2022-07-13
[cli] Support multiple remote URLs in Git config

MatthewStanciu:
Two features that handle a user's local Git config have been shipped:

- #8100 
- #7910 

Both of these features currently pull only from the user's remote origin URL. This covers 90% of cases, but there are cases in which the user has more than one remote URL, and may want to use one other than the origin URL, especially in `vc git connect`. This PR:

- Adds support for multiple remote URLs in a user's Git config
- Updates `vc git connect` to prompt the user to select a URL if they have multiple remote URLs
- Updates `createGitMeta` to send the connected Git repository url by default, origin url otherwise

#### Tests

#### Code Review


MatthewStanciu:
Keeping as a draft for now because more work needs to be done in `vc deploy` before this is ready to merge.

packages/cli/src/commands/git/connect.ts
```diff
@@ -64,20 +65,37 @@ export default async function connect(
...
     );
     return 1;
   }
-  output.log(`Identified Git remote "origin": ${link(remoteUrl)}`);
+
+  let remoteUrl: string;
+
+  if (Object.keys(remoteUrls).length > 1) {
+    output.log(`Found multiple remote URLs.`);
+    remoteUrl = await listRemoteUrls(client, remoteUrls);
```
EndangeredMassa: I think `listRemoteUrls` would make more sense as something like `selectRemoteUrl`. That better describes what's happening from the call site's perspective.

packages/cli/test/unit/commands/git.test.ts
```diff
@@ -25,7 +25,7 @@ describe('git', () => {
           id: 'unlinked',
           name: 'unlinked',
         });
-        client.setArgv('projects', 'connect');
```
EndangeredMassa, line 28: How was this test working before?

EndangeredMassa (approved):

packages/cli/test/unit/commands/git.test.ts
MatthewStanciu: The test calls `git()` directly, which only looks at the subcommand.

packages/cli/src/commands/git/connect.ts
```diff
@@ -166,3 +184,22 @@ async function confirmRepoConnect(
...
+  remoteUrls: { [key: string]: string }
+): Promise<string> {
+  let choices: ListChoice[] = [];
+  Object.keys(remoteUrls).map(urlKey => {
+    choices.push({
+      name: `${remoteUrls[urlKey]} ${chalk.gray(`(${urlKey})`)}`,
+      value: remoteUrls[urlKey],
+      short: urlKey,
+    });
+  });
```
styfle:
```suggestion
  for (const [urlKey, urlValue] of Object.entries(remoteUrls)) {
    choices.push({
      name: `${urlValue} ${chalk.gray(`(${urlKey})`)}`,
      value: urlValue,
      short: urlKey,
    });
  }
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -166,3 +184,22 @@ async function confirmRepoConnect(
...
+
+  return await list(client, {
+    message: 'Which one do you want to connect?',
```
styfle:
```suggestion
    message: 'Which remote do you want to connect?',
```

packages/cli/src/commands/git/connect.ts
MatthewStanciu: Thanks, this is much cleaner

TooTallNate (approved):

PR #8147, opened 2022-07-13
[cli] Consume `--no-clipboard`

MatthewStanciu:
#8085 removed the clipboard copy feature in `vc deploy`, along with the `--no-clipboard` flag. Right now, the CLI exits and returns an error when someone includes the `--no-clipboard` flag. This PR instead consumes the flag and warns the user that the flag is deprecated.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -187,7 +187,7 @@ export default async (client: Client) => {
   if (argv['--no-clipboard']) {
     output.print(
       `${prependEmoji(
-        `The ${param('--no-clipboard')} option is deprecated.`,
+        `The ${param('--no-clipboard')} option was ignored because it is the default behavior. Please remove it.`,
         emoji('warning')
       )}\n`
     );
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/deploy/index.ts
```diff
@@ -182,6 +183,15 @@ export default async (client: Client) => {
       )}\n`
     );
   }
+  // deprecate --no-clipboard
```
styfle: This comment is not needed

packages/cli/src/commands/deploy/index.ts
```diff
@@ -182,6 +183,15 @@ export default async (client: Client) => {
...
+  if (argv['--no-clipboard']) {
+    output.print(
+      `${prependEmoji(
+        `The ${param('--no-clipboard')} option is deprecated.`,
```
```suggestion
        `The ${param('--no-clipboard')} option was ignored because it is the default behavior. Please remove it.`,
```

styfle (approved):
Thanks!

PR #8149, opened 2022-07-13
[cli] Normalize output file paths in `vercel build`

TooTallNate:
When a Builder returns an output asset that contains irregular slashes (multiple slashes or leading/trailing slashes), then have them be removed from the file path before creating the output asset.

This fixes an edge case where `@vercel/next` could end up outputting a Serverless Function with a trailing slash (i.e. `en-US/`). Before this PR, that would be serialized to the filesystem at `en-US/.func`, but after this fix it's saved in the correct `en-US.func` directory.


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -67,6 +75,13 @@ export interface PathOverride {
   path?: string;
 }
 
+/**
+ * Remove duplicate slashes as well as leading/trailing slashes.
+ */
+function normalizePath(path: string): string {
```
styfle: Let’s give this a better name like stripSlashes since we already have several normalizePath functions

TooTallNate: 3e4e9536c08035523618986c6c14a2cd376a9cad

styfle (approved):

EndangeredMassa (approved):

PR #8151, opened 2022-07-13
[cli] MAJOR: `vc ls` visual changes

MatthewStanciu:
- Username field does not show if all deployment usernames in the page match the context name
- The `age` field shows first in the table
- A new `duration` field shows the time it took to deploy
- A new `--prod` flag filters for production deployments
- Table headers are colored cyan

### Before

Screen Shot 2022-07-13 at 6 29 15 PM

Screen Shot 2022-07-13 at 6 31 15 PM

### After

Screen Shot 2022-07-26 at 2 23 27 PM

Screen Shot 2022-07-26 at 2 23 21 PM

#### Tests

#### Code Review


TooTallNate:
Maybe could be in a separate PR, but I think it would be nice to include another column "Inspect" which would render a clickable link to the deployment inspection page. Perhaps we'd only include this column when support for clickable links is detected by the terminal emulator.

packages/cli/src/commands/list.ts
```diff
@@ -271,18 +292,32 @@ export default async function main(client: Client) {
...
+              'Status',
+              'Duration',
+              'Username',
+            ]
+          : [
+              'Age',
+              inspect ? 'Inspect URL' : 'Deployment URL',
+              'Status',
+              'Duration',
+            ]
```
Ethan-Arrowood: You could simplify this logic by putting these into a variable above like so:

```js
const headers = ['Age', inspect ? 'Inspect URL' : 'Deployment URL', 'Status', 'Duration'];
if (showUsername) headers.push('Username');
```

packages/cli/test/unit/commands/list.test.ts
```diff
@@ -75,6 +79,84 @@ describe('list', () => {
...
+      const header: string[] = parseSpacedTableRow(output.split('\n')[3]);
+      const data: string[] = parseSpacedTableRow(output.split('\n')[4]);
+      data.shift();
+
+      expect(org).toEqual(user.username);
+      expect(header).toEqual(['Age', 'Deployment URL', 'Status', 'Duration']);
+      expect(data).toEqual([
+        'https://' + deployment.url,
+        stateString(deployment.state || ''),
+        getDeploymentDuration(deployment as unknown as Deployment),
```
Ethan-Arrowood: why is the type casting necessary here? `useDeployment` should return a `Deployment`

Ethan-Arrowood (approved):
really well done. tests look good and so does the implementation. Like how much more color there is now too!

packages/cli/src/commands/list.ts
MatthewStanciu: Thanks!

packages/cli/test/unit/commands/list.test.ts
MatthewStanciu: This is necessary because there are two `Deployment` types: one in `@vercel/client` and one in `cli/src/types`. The one in `cli` is used by `list` and the one in `client` is used by the mock deployments.

This should probably be fixed, but I would imagine that's for a different PR. But happy to do it in this one.

MatthewStanciu: Turns out I was stupid and it was actually very easy to fix. Fixed in latest commit

MatthewStanciu:
Merging into #8102

MatthewStanciu:
@TooTallNate re inspect: this PR includes a new flag, `--inspect`, which shows the inspect URL instead of the deployment URL. What do you think about that as a solution?

I want the table to show inspect URLs, but I don't want people who use terminals that don't support terminal links to have a majorly different experience using this command—and I worry that not having a way to access the inspect URL when people with other terminals do would make that happen.

packages/cli/src/commands/list.ts
```diff
@@ -317,27 +357,67 @@ export default async function main(client: Client) {
...
-  return d.name;
+function getDeployUrl(
+  deployment: Deployment,
+  projectName: string | undefined,
+  inspect?: boolean
+): string {
+  const url = inspect ? deployment.inspectorUrl : 'https://' + deployment.url;
+  if (terminalLink.isSupported) {
+    const depId = getDeploymentIdFromUrl(url, projectName, inspect);
+    return terminalLink(depId, url);
```
EndangeredMassa: **question (blocking):** Did we decide no terminal linking at all or just not in the other case? My understanding was that we'd forgo terminal linking as a replacement for full URLs, at least.

packages/cli/src/commands/list.ts
```diff
@@ -317,27 +357,67 @@ export default async function main(client: Client) {
...
+  inspect?: boolean
+): string {
+  if (!projectName) {
+    return url;
   }
+  return inspect
+    ? url.split('/').slice(-1)[0]
+    : url
+        .match(
+          /(?<=https:\/\/)(.*)(?=(.*)(\.vercel\.app|\.now\.sh|\.vercel\.sh))/g
```
EndangeredMassa: **question (non-blocking):** should we report misses here as an error?

If we don't match one of these cases, it seems like something is seriously wrong, but we probably shouldn't bother the user about it. Can we report an error somewhere, like Sentry, when this happens?

packages/cli/src/commands/list.ts
```diff
@@ -317,27 +357,67 @@ export default async function main(client: Client) {
...
+        )?.[0]
+        ?.split(`${projectName}-`)?.[1]
+        ?.split('-')?.[0] || url;
```
EndangeredMassa: **suggestions (soft-blocking):** refactor for clarity, add tests

I think I understand the intent, but this is hard to follow. Can we break it into named variables/functions?

Can we add tests for this?

packages/cli/test/helpers/read-output-stream.ts
```diff
@@ -8,7 +8,9 @@ export function readOutputStream(
     let output: string = '';
     let lines = 0;
     const timeout = setTimeout(() => {
-      reject();
+      reject(
+        `Was waiting for ${length} lines, but only received ${chunks.length}`
```
EndangeredMassa: **suggestion (blocking):** 

```suggestion
        new Error(`Was waiting for ${length} lines, but only received ${chunks.length}`)
```

Can we reject with an Error instead of a string? It makes catching errors easier.

packages/cli/test/mocks/deployment.ts
```diff
@@ -7,17 +7,23 @@ import { Build, User } from '../../src/types';
...
+  state?: State;
 }) {
   const createdAt = Date.now();
   const url = new URL(chance().url());
+
   const deployment: Deployment = {
     id: `dpl_${chance().guid()}`,
     url: url.hostname,
-    name: '',
+    inspectorUrl: `https://vercel.com/${chance()}`,
```
EndangeredMassa: **suggestion (soft-blocking):** Use a static arbitrary value instead

I don't think this does what you expect. The result of `chance()` is an object, making this: `https://vercel.com/[object Object]`.

I don't think we should use random values in tests anyway. It can cause tests to be flakey and make failures harder to reproduce, depending on how you handle randomness seeds.

In this or a follow-up PR: Can this be a static arbitrary value instead?

packages/cli/test/mocks/deployment.ts
```diff
@@ -7,17 +7,23 @@ import { Build, User } from '../../src/types';
...
+    name: chance.name,
```
EndangeredMassa: **suggestion (soft-blocking):** use a static arbitrary value instead

This always evaluates to `Chance`. I think you would need `chance().name()` to get a random name.

As above, I think we should avoid randomness in tests.

In this or a follow-up PR: Can this be a static arbitrary value instead?

packages/cli/test/mocks/deployment.ts
```diff
@@ -26,13 +32,17 @@ export function useDeployment({
...
       uid: creator.id,
       email: creator.email,
-      username: creator.name,
+      name: creator.name,
+      username: creator.username,
     },
-    readyState: 'READY',
+    readyState: state,
+    state: state,
+    ready: Date.now() + Math.floor(Math.random() * 300000),
```
EndangeredMassa: **suggestion (soft-blocking):** as above, let's avoid randomness in tests

EndangeredMassa (changes requested):

packages/cli/src/commands/list.ts
MatthewStanciu: Agh, yeah I think since that was the takeaway from our team discussion, I'll remove `terminal-link` for now. BUT for the record I think `vc ls` is a great use case for `terminal-link`. I mean, does this not look way better?

181114770-ca947618-e0af-4b20-a387-b06552018b98

We should consider re-implementing this for `vc ls` using Nate's new thing in another PR sometime.

packages/cli/src/commands/list.ts
MatthewStanciu: I think that would be great, but I'd be in favor of doing this in another PR—I'm not sure how to do this, and it doesn't appear to be urgent and I want to optimize for getting the feature out the door.

What do you think about that?

packages/cli/src/commands/list.ts
EndangeredMassa: It does, but I think the user expectation issue is a valid one.

packages/cli/src/commands/list.ts
EndangeredMassa: Certainly! I marked my comment as `non-blocking` to mean exactly that. This doesn't have to be addressed in the current PR or the current task, in my opinion.

packages/cli/src/commands/list.ts
MatthewStanciu: This function was used for displaying `terminal-link`ed URLs, so I ended up removing it altogether.

EndangeredMassa (approved):

TooTallNate (approved):

PR #8155, opened 2022-07-14
[routing-utils] MAJOR refactor getTransformedRoutes and types

styfle:
This is a semver major change to the public API for `@vercel/routing-utils` which includes the following breaking changes.

1. `getTransformedRoutes({ nowConfig })` props changed to `getTransformedRoutes(nowConfig)`
2. `type Source` renamed `type RouteWithSrc`
3. `type Handler` renamed `type RouteWithHandle`
4. `interface VercelConfig` removed
5. `type NowConfig` removed
6. `type NowRewrite` removed
7. `type NowRedirect` removed
8. `type NowHeader` removed
9. `type NowHeaderKeyValue` removed

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -206,7 +206,7 @@ export default async function main(client: Client): Promise<number> {
     normalizePath(relative(workPath, f))
   );
 
-  const routesResult = getTransformedRoutes({ nowConfig: vercelConfig || {} });
+  const routesResult = getTransformedRoutes(vercelConfig || {});
   if (routesResult.error) {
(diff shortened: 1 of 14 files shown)


styfle:
The redwood failure is happening because `@vercel/redwood` doesn't use ncc which means the version of `@vercel/routing-utils` is not bundled and thus fails E2E tests. You can see the others pass because they use ncc. This will work once published to npm and we might want to consider using ncc for `@vercel/redwood` or changing our E2E tests to utilize the tarballs from #7967

EndangeredMassa (approved):

ijjk (approved):

PR #8157, opened 2022-07-14
[cli] `vc project ls` visual updates

MatthewStanciu:
Related to #8102 & #8151. This PR updates the look of `vc project ls`:

- Header names are cyan
- A new "Latest Production URL" section

### Before

Screen Shot 2022-08-11 at 10 46 03 AM

### After

Screen Shot 2022-08-11 at 10 45 54 AM

#### Tests

#### Code Review


TooTallNate:
Would be great IMO if we could use a clickable link to the project page under the project name section, instead of having an entire column for that. I'd prefer to print something like the production domain URL in that new column instead (also clickable 😄).

It's also not super clear to me what "Status" section is referring to, I'm not entirely convinced we need that column (the project card on the Vercel dashboard doesn't have the status dot, for comparison).

packages/cli/src/commands/project/list.ts
```diff
@@ -48,33 +55,38 @@ export default async function list(
...
+            [
+              project.name,
+              chalk.bold(getInspectUrl(contextName, project.name)),
+              stateString(project.latestDeployments[0]?.readyState),
+              chalk.gray(ms(Date.now() - project.updatedAt)),
+            ],
+          ])
+          // flatten since the previous step returns a nested
+          // array of the deployment and (optionally) its instances
+          .flat(),
```
Ethan-Arrowood: could you just remove the nested array now? It looks like its unnecessary

packages/cli/src/util/list/state-string.ts
```diff
@@ -0,0 +1,24 @@
+import chalk from 'chalk';
+import title from 'title';
+
+export function stateString(s: string) {
+  const CIRCLE = '● ';
+  const sTitle = s && title(s);
```
Ethan-Arrowood: `s` can't be undefined or null here (by type declaration) so remove the `s &&` part. 

Also, fwiw this kinda thing is a code smell IMO. `&&` should really only be used for conditional expressions. I'd rather see an explicit `s ? title(s) : 's is null/undefined case'` expression instead

Ethan-Arrowood (commented):
really like these visual improvement prs. take a look at my comments then I believe it is good to go

packages/cli/src/commands/project/list.ts
MatthewStanciu: The nested array is necessary—it's creating an array containing the pieces of data to appear under the header for each project, generating:

```
['project name', 'manage', 'updated'],
['name 1', 'vercel.com/user/name-1', '35s'],
['name 2', 'vercel.com/user/name-2', '1m']
// ...etc
```

packages/cli/src/util/list/state-string.ts
MatthewStanciu: 👍 Decided to remove the state string altogether after seeing Nate's comment

MatthewStanciu:
@TooTallNate @Ethan-Arrowood curious to hear your thoughts on this new behavior with `terminal-link` (screenshots in the original message have been updated).

`terminal-link`'s default fallback does not look very good—here's what the "with terminal-link" case would look like on terminals that don't support it:

Screen Shot 2022-07-21 at 4 32 51 PM

My concern is that I'm not sure if it's bad to have the command behave differently based on whether the user's terminal supports hyperlinks. But also I think the command needs to look good on terminals that don't support hyperlinks.

Ethan-Arrowood:
Can you create a custom fallback? 

https://github.com/sindresorhus/terminal-link/blob/8311a11c5ab95c7c7bf4983c565276759c86fc31/test.js#L72-L81

Make it pretty for non-supported users?

MatthewStanciu:
@Ethan-Arrowood I could create a custom fallback, but I think the best custom fallback in this case is adding another section containing the URL—since the hyperlink for supported users is in the `Project Name` section, I don't think there's another fallback that works otherwise.

My concern though is that maybe it's bad for a command to have different behavior based on which terminal people are using. But I'm not sure if that's a real problem or not.

EndangeredMassa:
It seems like there is decent support, but it's not ubiquitous, according to [this support tracker](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) (no longer being updated) I found. We should have a fallback if we want to add terminal links.

Having the output change based on support for this isn't great for cross-terminal scripting of the command, sure. I don't imagine that's a common concern, though. I'd be fine with different outputs here.

packages/cli/test/unit/commands/project.test.ts
```diff
@@ -9,14 +9,17 @@ import {
   pluckIdentifiersFromDeploymentList,
   parseSpacedTableRow,
 } from '../../helpers/parse-table';
+import { useDeployment } from '../../mocks/deployment';
 
 describe('project', () => {
   describe('list', () => {
-    it('should list deployments under a user', async () => {
+    it('should list projects', async () => {
```
EndangeredMassa, line 15: **question (blocking):** Is this test working properly?

This passes in CI, but not for me locally. I see:

```
 FAIL  test/unit/commands/project.test.ts
  ● project › list › should list projects

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
    -   "cli",
    +   "cli",
        "--",
      ]

      37 |         'Updated',
      38 |       ]);
    > 39 |       expect(data).toEqual([project.project.name, `--`]);
         |                    ^
      40 |     });
      41 |     it('should use inspector url', async () => {
      42 |       const user = useUser();

      at Object.<anonymous> (test/unit/commands/project.test.ts:39:20)
```

@MatthewStanciu does this pass for you locally?

packages/cli/src/commands/project/list.ts
```diff
@@ -48,39 +51,69 @@ export default async function list(
...
+                  )
+                : chalk.bold(project.name),
+              inspect
+                ? getInspectUrl(contextName, project.name)
+                : getLatestProdUrl(project),
+              chalk.gray(ms(Date.now() - project.updatedAt)),
+            ],
+          ])
+          .flat(),
+      ],
```
EndangeredMassa, line 72: **suggestion (non-blocking):** break up complex expressions

This is getting to be a pretty complex set of expressions, which is hard to follow. I'd be happy to pair on refactoring this to make it easier to follow.

EndangeredMassa (changes requested):
I think the tests changed here may be broken, but maybe something else is going on. Let's check before merging.

packages/cli/test/unit/commands/project.test.ts
MatthewStanciu: Hmm, it does pass for me locally. I can't tell what's happening to make it fail on your end.

packages/cli/src/commands/project/list.ts
```diff
@@ -48,39 +51,69 @@ export default async function list(
...
-          chalk.bold(secret.name),
-          chalk.gray(`${ms(cur - secret.updatedAt)} ago`),
-        ])
-      ),
+  if (projectList.length > 0) {
+    const tablePrint = table(
+      [
+        [
+          'Project Name',
+          inspect ? 'Inspect Project' : 'Latest Production URL',
```
EndangeredMassa: **suggestion (non-blocking):** 

```suggestion
          inspect ? 'Inspect Latest Production Deployment' : 'Latest Production URL',
```

I think our jargon aligns better with `Inspect Deployment` instead of `Inspect Project`. What do you think of this? I couldn't think of something shorter that still made it clear.

MatthewStanciu: But `--inspect` takes you to the homepage for the Project on Vercel's dashboard, so I don't think it's correct to say "Inspect Latest Production Deployment". Would you be in favor of something along the lines of "Manage Project"? That way it's short and accurate, but is also separate from the expectations of "Inspect".

EndangeredMassa: Ohh, I missed that part. Yeah, maybe `Inspect` is a loaded term. Or maybe I'm reading too much into it.

I do like `Manage Project` or `View Project` or just `Project`, but then the flag should probably be called something else, too.

packages/cli/test/unit/commands/project.test.ts
EndangeredMassa: It turns out this was a difference between terminals that supported terminal link. In CI, terminal link is not supported. In iterm2, it is!

packages/cli/src/commands/project/list.ts
MatthewStanciu: Now that the `terminal-link` logic is removed, do you still feel this is the case?

EndangeredMassa: Personally, I'd still break it up, but it is significantly leaner now. I'm good with this.

packages/cli/test/unit/commands/project.test.ts
```diff
@@ -28,29 +31,12 @@ describe('project', () => {
...
-
-      expect(org).toEqual(team[0].slug);
-      expect(header).toEqual(['name', 'updated']);
-      expect(data).toEqual([project.project.name]);
+      expect(header).toEqual([
+        'Project Name',
+        'Latest Production URL',
+        'Updated',
+      ]);
+      expect(data).toEqual([project.project.name, `--`]);
```
EndangeredMassa: **suggestion (soft-blocking):** add test for deployment URL showing up in output

In this or a follow-up PR, can we test that a deployment URL makes it in as well?

packages/cli/src/commands/project/index.ts
```diff
@@ -57,6 +57,8 @@ export default async function main(client: Client) {
     argv = getArgs(client.argv.slice(2), {
       '--next': Number,
       '-N': '--next',
+      '--inspect': Boolean,
+      '-i': '--inspect',
```
EndangeredMassa: **suggestion (soft-blocking):** remove the leftover args

```suggestion
```

In this or a follow-up PR, can we remove these leftovers?

EndangeredMassa (approved):

packages/cli/test/unit/commands/project.test.ts
```diff
@@ -36,7 +34,33 @@ describe('project', () => {
...
+      expect(data).toEqual([project.project.name, 'https://foobar.com']);
+    });
+    it('should list projects when there is no production deployment', async () => {
+      const user = useUser();
+      useTeams('team_dummy');
+      defaultProject.alias = [];
+      const project = useProject({
+        ...defaultProject,
+      });
+      console.log(project);
```
EndangeredMassa: **suggestion (soft-blocking):** remove line

```suggestion
```

In this or a follow-up PR: Can we remove this left over `console.log`?

EndangeredMassa (approved):

packages/cli/test/unit/commands/project.test.ts
MatthewStanciu: Thanks for catching!

packages/cli/src/commands/project/list.ts
```diff
@@ -48,39 +49,48 @@ export default async function list(
...
+      [
+        ['Project Name', 'Latest Production URL', 'Updated'].map(header =>
+          chalk.bold(chalk.cyan(header))
+        ),
+        ...projectList
+          .map(project => [
+            [
+              chalk.bold(project.name),
+              getLatestProdUrl(project),
+              chalk.gray(ms(Date.now() - project.updatedAt)),
```
TooTallNate, line 68: Was it intentional to drop the "ago" part?

TooTallNate (approved):

PR #8158, opened 2022-07-14
[next] Update allowQuery for prerendered paths

ijjk:
This updates our `allowQuery` generating to ignore all query values for build-time prerender paths as these will match before dynamic routes since they are filesystem routes and the query values will not be overridden properly like they are for fallback prerender paths. This also adds testing for both prerender path types with on-demand ISR to ensure the cache is updated as expected regardless of the query.  

Deployment with patch can be seen here https://nextjs-issue-odr-simple-hrjt2dagm-ijjk-testing.vercel.app/

### Related Issues

x-ref: https://github.com/vercel/next.js/issues/38306
x-ref: https://github.com/vercel/next.js/issues/38653

#### Tests

#### Code Review

### DIFF
--- a/packages/next/test/fixtures/00-server-build/index.test.js
+++ b/packages/next/test/fixtures/00-server-build/index.test.js
@@ -1,8 +1,144 @@
+/* eslint-env jest */
 const path = require('path');
-const { deployAndTest } = require('../../utils');
+const cheerio = require('cheerio');
+const { deployAndTest, check, waitFor } = require('../../utils');
+const fetch = require('../../../../../test/lib/deployment/fetch-retry');
+
+async function checkForChange(url, initialValue, getNewValue) {
+  return check(async () => {
+    const res = await fetch(url);
+
+    if (res.status !== 200) {
+      throw new Error(`Invalid status code ${res.status}`);
+    }
+    const newValue = await getNewValue(res);
+
+    return initialValue !== newValue
+      ? 'success'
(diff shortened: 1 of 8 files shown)


packages/next/test/fixtures/00-server-build/index.test.js
```diff
@@ -1,8 +1,144 @@
...
+    const revalidateRes = await fetch(
+      `${ctx.deploymentUrl}/api/revalidate?urlPath=${pathsToCheck[0].urlPath}`
+    );
+    expect(revalidateRes.status).toBe(200);
+    expect((await revalidateRes.json()).revalidated).toBe(true);
+
+    await checkPaths(pathsToCheck);
+
+    expect(preRevalidateRandom).not.toBe(initialRandom);
+    expect(preRevalidateRandomData).not.toBe(initialRandomData);
```
styfle: Since both these initial vars are undefined at this point, these two assertions are equivalent to the following, right?

```js
expect(preRevalidateRandom).toBeDefined()
expect(preRevalidateRandomData).toBeDefined()
```

ijjk: Yeah should be the same, updated in https://github.com/vercel/vercel/pull/8158/commits/a15615a51f554eeb93d401639030e61b6d8a57e9 for clarity

styfle (approved):
Great, thanks!

PR #8160, opened 2022-07-14
[cli] Minor `vc inspect` visual updates

MatthewStanciu:
Featuring:

- Aliases are now shown
- `readyState` –> `status` in "General" section
- `createdAt` –> `created` in "General" section
- Update `stateString` to support more states & be more visually appealing

### Before

Screen Shot 2022-07-14 at 3 10 40 PM

### After

Screen Shot 2022-07-14 at 3 10 29 PM

#### Tests

#### Code Review


EndangeredMassa (approved):

packages/cli/src/commands/inspect.ts
```diff
@@ -165,19 +184,24 @@ export default async function main(client: Client) {
...
-
+      return chalk.red(CIRCLE) + sTitle;
     case 'READY':
-      return s;
-
+      return chalk.green(CIRCLE) + sTitle;
+    case 'QUEUED':
+      return chalk.white(CIRCLE) + sTitle;
+    case 'CANCELED':
+      return chalk.gray(sTitle);
```
styfle:
```suggestion
      return chalk.gray(CIRCLE) + sTitle;
```

I think this should say canceled, right?

packages/cli/src/commands/inspect.ts
```diff
@@ -165,19 +184,24 @@ export default async function main(client: Client) {
...
     default:
-      return chalk.gray(s || 'UNKNOWN');
+      return chalk.gray('  --');
```
styfle: Lets keep this `UNKNOWN`

packages/cli/src/commands/inspect.ts
```diff
@@ -165,19 +184,24 @@ export default async function main(client: Client) {
...
     case 'ERROR':
-      return chalk.red(s);
-
+      return chalk.red(CIRCLE) + sTitle;
     case 'READY':
-      return s;
-
+      return chalk.green(CIRCLE) + sTitle;
+    case 'QUEUED':
+      return chalk.white(CIRCLE) + sTitle;
```
styfle: I believe we use gray for queued status on the dashboard so best to match that

TooTallNate:
How about using `https://` for the deployment URL and aliases?

TooTallNate:
What do you think about dropping the "Ready" in the Builds section? Seems redundant to me.

PR #8162, opened 2022-07-15
[cli] Add `--environment` flag to `vc env pull`

MatthewStanciu:
Right now, `vc env pull` only pulls development environment variables. This PR adds a new flag, `--environment,` which allows users to specify which environment to pull from.

#### Tests

#### Code Review


packages/cli/src/commands/env/index.ts
```diff
@@ -111,6 +112,7 @@ export default async function main(client: Client) {
     argv = getArgs(client.argv.slice(2), {
       '--yes': Boolean,
       '-y': '--yes',
+      '--environment': String,
```
EndangeredMassa, line 117:
```suggestion
      '--environment': String,
      '--env': '--environment',
```

I think adding a `--env` alias for this is pretty reasonable. We could add that alias to `vc pull` as well.

packages/cli/src/commands/env/index.ts
```diff
@@ -126,6 +128,24 @@ export default async function main(client: Client) {
...
+      target = ProjectEnvTarget.Production;
+    } else {
+      output.error(
+        `Invalid environment \`${chalk.cyan(
+          environmentArg
+        )}\`. Options: development | preview | production`
+      );
+      return 1;
+    }
+  }
```
EndangeredMassa, line 142: Converting strings to enums is messy business. There's a helper that makes this a bit easier and will work with any future changes to `ProjectEnvTarget`. I think you could do something like this:

```
// TODO: you'll have to export the `envTargets` function in this file
import { isValidEnvTarget, envTargets } from '../../util/env/env-target.ts';

// ...

function parseEnvironmentTarget(environment): ProjectEnvTarget | undefined {
  if (isValidEnvTarget(environment)) {
    return environment;
  }
  return undefined;
}

// ...

const target = environmentArg
  ? parseEnvironmentTarget(environmentArg)
  : ProjectEnvTarget.Development;

if (!target) {
  output.error(
    `Invalid environment \`${chalk.cyan(
      environmentArg
    )}\`. Valid options: ${envTargets()}`
  );
  return 1;
}
```

EndangeredMassa (approved):

EndangeredMassa (approved):

packages/cli/src/commands/env/index.ts
```diff
@@ -126,6 +140,22 @@ export default async function main(client: Client) {
   const subArgs = argv._.slice(1);
   const { subcommand, args } = getSubcommand(subArgs, COMMAND_CONFIG);
   const { output, config } = client;
+
+  const environmentArg = argv['--environment']?.toLowerCase();
+
+  const target = environmentArg
+    ? parseEnvironmentTarget(environmentArg)
+    : ProjectEnvTarget.Development;
```
styfle: Let's throw when an invalid target is provided instead of falling back to development

packages/cli/src/commands/env/index.ts
```diff
@@ -126,6 +140,22 @@ export default async function main(client: Client) {
...
+
+  if (!target) {
```
styfle: Can this ever be undefined? The development value is the fallback above

packages/cli/src/commands/env/index.ts
```diff
@@ -104,13 +108,23 @@ const COMMAND_CONFIG = {
   pull: ['pull'],
 };
 
+function parseEnvironmentTarget(
+  environment = 'development'
```
styfle: This default value seem unnecessary since this function isnt called when there is no value

styfle (changes requested):
Let's remove the `--env` shorthand. We don't typically use two dashes for short options (usually its just one dash followed by one letter). 

Since it doesn't seem necessary for this PR, lets remove so we can discuss in a future PR.

packages/cli/src/commands/env/index.ts
MatthewStanciu: It only falls back to `.Development` if the user doesn't pass in environment argument. If they do, it runs `parseEnvironmentTarget`, which returns undefined if the argument is invalid.

packages/cli/src/commands/env/index.ts
MatthewStanciu: This is undefined when the user passes in an invalid argument—it's only `.Development` when they don't. Will add a comment to clarify!

packages/cli/src/commands/env/index.ts
styfle: I think this should be `if (!isValidEnvTarget(target))` so you don't check validity twice. You can simplify this to something like this:

```js
  const target = argv['--environment'] || 'development';
  if (!isValidEnvTarget(target) {
    output.error(...);
    return 1;
  }
  // continue with project link etc
```

styfle (approved):
Perfect, thanks! 🎉

PR #8167, opened 2022-07-15
[build-utils][node][python][ruby] Update error message for EOL runtimes

styfle:
This PR updates the error message when the runtime version detected is EOL

### DIFF
--- a/packages/build-utils/src/fs/node-version.ts
+++ b/packages/build-utils/src/fs/node-version.ts
@@ -33,9 +33,6 @@ function getHint(isAuto = false) {
     : `Please set "engines": { "node": "${range}" } in your \`package.json\` file to use Node.js ${major}.`;
 }
 
-const upstreamProvider =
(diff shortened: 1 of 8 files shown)


AndyBitz (approved):

swarnava:
Do you want to include ?:

> This change is because Node.js 12 has reached end-of-life and has stopped receiving security updates.

ethomson (approved):

swarnava (approved):

styfle:
@swarnava I think the message is already too long so I would prefer to keep it shorter to avoid burying the lede. Probably better to add more details to the linked docs.

swarnava:
@styfle 

> @swarnava I think the message is already too long so I would prefer to keep it shorter to avoid burying the lede. Probably better to add more details to the linked docs.

good point. But doesn't look like we are showing https://vercel.link/node-version in current build error.

styfle:
@swarnava It should be there at the top of the page where we show an error message and a button to click for more info

PR #8170, opened 2022-07-15
[cli] Log delta in `vc env pull`

MatthewStanciu:
This PR adds a new message to the end of `vc env pull` which shows delta of what was added, changed, and removed.

Example: you have a local `.env` file in a directory linked to a Vercel project. You change `TEST_1`, remove `TEST_2` via `vc env rm TEST_2`, and add `TEST_3` via `vc env add`. Then you run `vc env pull`. Here's what you will see:

```
> Updating existing .env file
> Downloading `development` Environment Variables for Project name
✅  Updated .env file

+ TEST_3
~ TEST_1
- TEST_2
```

Screen Shot 2022-07-20 at 11 09 56 AM

#### Tests

#### Code Review


packages/cli/src/commands/env/pull.ts
```diff
@@ -125,6 +136,12 @@ export default async function pull(
     )}\n`
   );
 
+  if (deltaString) {
+    output.print('\n' + deltaString);
+  } else if (exists) {
+    output.log('No changes found.');
```
EndangeredMassa, line 145: If `createEnvObject` throws an error, it will end up returning `undefined`. When it gets down here, there could have been actual changes, but we'll log here that there were none. In this case, I think it's better to either say nothing or be clear that we couldn't detect the changes, but the latest values were updated.

packages/cli/src/commands/env/pull.ts
```diff
@@ -135,3 +152,79 @@ function escapeValue(value: string | undefined) {
...
+  deltaString += chalk.red(addDeltaSection('-', removed));
+
+  return deltaString;
+}
+
+function addDeltaSection(prefix: string, arr: string[]): string {
+  let deltaSection = '';
+  let ellipsis = '';
+  for (const [i, item] of arr.entries()) {
+    if (i === 2) ellipsis = '…';
```
EndangeredMassa: Should this check that there are more items as well? If there are exactly 3 items, will this render all 3 items plus an ellipses?

Can we add a test for boundary cases like this?

packages/cli/src/commands/env/pull.ts
```diff
@@ -135,3 +152,79 @@ function escapeValue(value: string | undefined) {
         .replace(new RegExp('\r', 'g'), '\\r') // combine newlines (windows) into one line
     : '';
 }
+
+async function createEnvObject(
```
EndangeredMassa: These functions related to env diffing could be better organized into a separate file, in my opinion.

packages/cli/src/commands/env/pull.ts
```diff
@@ -135,3 +152,79 @@ function escapeValue(value: string | undefined) {
...
+
+function buildDeltaString(
+  oldEnv: Dictionary<string | undefined>,
+  newEnv: Dictionary<string | undefined>
+): string {
+  let { added, changed, removed } = findChanges(oldEnv, newEnv);
+
+  let deltaString = '';
+  deltaString += chalk.green(addDeltaSection('+', added));
+  deltaString += chalk.yellow(addDeltaSection('~', changed));
```
EndangeredMassa: I think it makes sense that `~` would mean "changed", but not immediately. I think a user may not know what this means, especially if their first interactions with this output only shows one variable changed.

Should we log a `(changed)` suffix (maybe dimmed) at the end of these lines? Or can we show a brief legend without taking up too much space?

MatthewStanciu: I'll explore a legend, though not quite sure yet how to make it look good & not feel awkward. Big fan of the `(changed)` suffix.

packages/cli/src/commands/env/pull.ts
MatthewStanciu: Will add a test for the ellipsis case. 👍 

The intention is to render 3 items plus an ellipsis. The thinking here is that if someone made some major changes to their env variables, I wouldn't want the diff to take up too much of their screen.

packages/cli/src/commands/env/pull.ts
MatthewStanciu: Added a `Changed:` prefix, but not super confident in this

EndangeredMassa: I think the `Changed:` prefix is fine, but not necessary. I think the issue with `~` still stands, though.

packages/cli/src/commands/env/pull.ts
EndangeredMassa: Did you intentionally remove the ellipses code? I think both approaches (with or without truncation) is fine, but wanted to make sure this was the intent. If so, the fixture `env-pull-delta-ellipsis` is still in the PR.

If not, the test you added was removed. You'll want to restore that. I also suggest adding a test for exactly 3 items not showing an ellipses.

MatthewStanciu: Ah, thanks for catching. I intentionally removed the ellipsis in favor of displaying changed items in a list. I think it didn't really make sense to show each variable on its own line, and I think showing them in a list makes more sense.

Will remove the fixture.

EndangeredMassa: Oh, yeah, I like listing them a lot

packages/cli/src/commands/env/pull.ts
MatthewStanciu: Thinking about it more, I'm not convinced this is an issue. I think there is enough context as to what the tilde means: there is now a "changed" prefix which gives a clue, especially if there's only one edited variable, and the yellow color indicates (at least to me) that the variable wasn't added or removed, but somewhere in between—AKA edited.

I think a legend, or more explicit context, will take up a lot of space, and is the sort of thing that may be useful to see once but will be annoying to see every time after that.

I would be happy to ship this as is. Though I think it's worth bringing this to the rest of the team or to `#design` / `#engineering` for additional feedback.

packages/cli/src/commands/env/pull.ts
```diff
@@ -107,6 +111,22 @@ export default async function pull(
...
+
+      if (oldEnv && records) {
+        deltaString = buildDeltaString(oldEnv, records);
+      }
+    } catch (err) {
+      deltaString = `Failed to find changes, but the \`${chalk.cyan(
+        filename
+      )}\` file was updated.`;
+      output.debug(`Failed to find changes: ${err}`);
+    }
```
Ethan-Arrowood: can you add a test for this part

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,77 @@
...
+function findChanges(
+  oldEnv: Dictionary<string | undefined>,
+  newEnv: Dictionary<string | undefined>
+): {
+  added: string[];
+  changed: string[];
+  removed: string[];
+} {
+  let added = [];
+  let changed = [];
```
Ethan-Arrowood: never re-assigned so these can be `const`
```suggestion
  const added = [];
  const changed = [];
```

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,77 @@
...
+  let removed = [];
```
Ethan-Arrowood:
```suggestion
```
remove this and just declare as constant on line when its assigned

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  for (const key of Object.keys(newEnv)) {
+    if (!oldEnv[key] && oldEnv[key] !== '') {
+      added.push(key);
+    } else if (oldEnv[key] !== newEnv[key]) {
+      changed.push(key);
+    }
+    delete oldEnv[key];
+  }
+  removed = Object.keys(oldEnv);
```
Ethan-Arrowood:
```suggestion
  const removed = Object.keys(oldEnv);
```

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,77 @@
...
+    changed,
+    removed,
+  };
+}
+
+export function buildDeltaString(
+  oldEnv: Dictionary<string | undefined>,
+  newEnv: Dictionary<string | undefined>
+): string {
+  let { added, changed, removed } = findChanges(oldEnv, newEnv);
```
Ethan-Arrowood:
```suggestion
  const { added, changed, removed } = findChanges(oldEnv, newEnv);
```
this should be caught by a lint rule but preference for using `const` when something is not re-assigned

packages/cli/src/util/parse-env.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const parseEnv = (env?: string[] | Dictionary<string>) => {
+  if (!env) {
+    return {};
+  }
+
+  if (typeof env === 'string') {
+    // a single `--env` arg comes in as a String
+    env = [env];
+  }
```
Ethan-Arrowood, line 12: surprised typescript isn't catching this... make sure to update the `env` function parameter type to include `string`

packages/cli/src/util/parse-env.ts
```diff
@@ -0,0 +1,34 @@
+// Converts `env` Arrays, Strings and Objects into env Objects.
+import { Dictionary } from '@vercel/client';
+
+export const parseEnv = (env?: string[] | Dictionary<string>) => {
```
Ethan-Arrowood:
```suggestion
export const parseEnv = (env?: string | string[] | Dictionary<string>) => {
```

packages/cli/src/commands/env/pull.ts
```diff
@@ -125,6 +138,19 @@ export default async function pull(
...
   );
 
+  output.print('\n');
+  if (deltaString) {
+    output.print(deltaString);
+  } else if (!oldEnv) {
+    output.log(
+      `Failed to find changes, but the \`${chalk.cyan(
+        filename
+      )}\` file was updated.`
```
EndangeredMassa: Should we say anything about failing in this case? I think we should only say `the .env file was updated`.

EndangeredMassa (approved):
I'm good with this version, but you may want to wait for confirmation from others who had suggestions on this before merging.

packages/cli/src/commands/env/pull.ts
MatthewStanciu: I think you're right. The goal is to make the user aware that something went wrong if they were expecting a diff, but this is an awkward interaction. Will remove 👍

packages/cli/src/util/parse-env.ts
```diff
@@ -0,0 +1,34 @@
+// Converts `env` Arrays, Strings and Objects into env Objects.
```
styfle: This comment should be on the function below, not the import

packages/cli/src/util/parse-env.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Dictionary } from '@vercel/client';
```
styfle:
```suggestion
import type { Dictionary } from '@vercel/client';
```

packages/cli/test/unit/commands/env.test.ts
```diff
@@ -184,23 +184,5 @@ describe('env', () => {
 
       await expect(pullPromise).resolves.toEqual(0);
     });
-
-    it('should show a custom message when it fails to read a file', async () => {
```
EndangeredMassa: I think we should still have a test for this case. The output should say that the `.env` file was updated, but it should not show a diff.

EndangeredMassa (approved):

packages/cli/test/unit/commands/env.test.ts
MatthewStanciu: I re-added the test, but I think it's hard to test for this. In every case where the env file was updated, it outputs "Updated <env name> file". In most cases, it then also outputs "No changes found", but in the case where it can't parse the file, it doesn't output that. So we want to test that something _doesn't_ output, which I'm not sure is possible right now.

EndangeredMassa: Ah, right, because our output mocking changed. This will at least make sure there's no error in this case.

EndangeredMassa (approved):

cb1kenobi (approved):

TooTallNate (approved):

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,78 @@
...
+): {
+  added: string[];
+  changed: string[];
+  removed: string[];
+} {
+  const added = [];
+  const changed = [];
+
+  for (const key of Object.keys(newEnv)) {
+    if (!oldEnv[key] && oldEnv[key] !== '') {
```
styfle: What is the value its checking for here? Is this basically the same

```suggestion
    if (oldEnv[key] === undefined) {
```

Or are you expecting null value in addition to undefined?

I'm also curious how to tell the difference between undefined and empty string. I don't think we allow an undefined value today.

MatthewStanciu: I'm just checking for undefined. This is better, thank you

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  return deltaString ? chalk.gray('Changes:\n') + deltaString : deltaString;
+}
+
+function addDeltaSection(prefix: string, arr: string[]): string {
+  let deltaSection = '';
+  for (const item of arr.sort()) {
+    deltaSection += `${prefix} ${item}\n`;
+  }
+  return deltaSection;
```
styfle:
```suggestion
  return arr.sort().map(item =>`${prefix} ${item}`).join('\n') + '\n'
```

Nit pick: this can be simplified

MatthewStanciu: Thank you 🙏

styfle (approved):
Looks good, thanks!

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,74 @@
...
+  let deltaString = '';
+  deltaString += chalk.green(addDeltaSection('+', added));
+  deltaString += chalk.yellow(addDeltaSection('~', changed));
+  deltaString += chalk.red(addDeltaSection('-', removed));
+
+  return deltaString ? chalk.gray('Changes:\n') + deltaString : deltaString;
+}
+
+function addDeltaSection(prefix: string, arr: string[]): string {
+  return arr.sort().map(item =>`${prefix} ${item}`).join('\n')
```
styfle:
```suggestion
  return arr.sort().map(item =>`${prefix} ${item}`).join('\n') + '\n';
```

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,74 @@
...
+  let deltaString = '';
+  deltaString += chalk.green(addDeltaSection('+', added));
+  deltaString += chalk.yellow(addDeltaSection('~', changed));
+  deltaString += chalk.red(addDeltaSection('-', removed));
+
+  return deltaString ? chalk.gray('Changes:\n') + deltaString : deltaString;
+}
+
+function addDeltaSection(prefix: string, arr: string[]): string {
+  return arr.sort().map(item =>`${prefix} ${item}`).join('\n') + '\n'
```
styfle:
```suggestion
  return arr.sort().map(item =>`${prefix} ${item}`).join('\n') + '\n';
```

packages/cli/src/util/env/diff-env-files.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Dictionary } from '@vercel/client';
+import { readFile } from 'fs-extra';
+import { parseEnv } from '../parse-env';
+import chalk from 'chalk';
+
+export async function createEnvObject(
+  envPath: string,
+  output: Output
+): Promise<Dictionary<string | undefined> | undefined> {
+  // Partially taken from https://github.com/tswaters/env-file-parser/blob/master/lib/parse.js
```
styfle:
```suggestion
  // Originally authored by Tyler Waters under MIT License: https://github.com/tswaters/env-file-parser/blob/f17c009b39da599380e069ee72728d1cafdb56b8/lib/parse.js
  // https://github.com/tswaters/env-file-parser/blob/f17c009b39da599380e069ee72728d1cafdb56b8/LICENSE
```

PR #8172, opened 2022-07-15
[next] Ensure manifests are specific to the included pages

ijjk:
### Related Issues

This updates to filter the `routes-manifest` and `pages-manifest` to only include entries for the pages that are being included in the specific serverless function. This fixes the case where multiple dynamic routes could match a path but one is configured with `fallback: false` so shouldn't match when executing for a different dynamic route. 

A regression test for this specific scenario has been added in the `00-mixed-dynamic-routes` fixture. 

#### Tests

#### Code Review


packages/next/src/server-build.ts
```diff
@@ -676,18 +681,77 @@ export async function serverBuild({
...
+        // in this specific serverless function so that we don't
+        // accidentally match a dynamic route while resolving that
+        // is not actually in this specific serverless function
+        for (const manifest of [
+          'routes-manifest.json',
+          'server/pages-manifest.json',
+        ]) {
+          const fsPath = path.join(entryPath, outputDirectory, manifest);
+
+          const relativePath = path.join(path.relative(baseDir, fsPath));
```
styfle: Is it expected that path.join() has one param instead of 2?

packages/next/src/server-build.ts
```diff
@@ -676,18 +681,77 @@ export async function serverBuild({
...
+          delete group.pseudoLayer[relativePath];
+
+          const manifestData = await fs.readJSON(fsPath);
+          const normalizedPages = new Set(
+            group.pages.map(page => {
+              page = `/${page.replace(/\.js$/, '')}`;
```
styfle, line 712: Do we need to consider .mjs here?

packages/next/src/server-build.ts
```diff
@@ -676,18 +681,77 @@ export async function serverBuild({
...
+            case 'server/pages-manifest.json': {
+              for (const key of Object.keys(manifestData)) {
+                if (isDynamicRoute(key) && !normalizedPages.has(key)) {
+                  delete manifestData[key];
+                }
+              }
+              break;
+            }
+            default: {
+              break;
```
styfle: Are there other cases for manifest? If not, let’s add a TS exhaustive check by using “as const“ on the array declaration and then throwing here in the default case

packages/next/src/server-build.ts
ijjk: We shouldn't as we're expecting `.js` files here in other usage and our page outputs should currently always be `.js`.

packages/next/src/server-build.ts
ijjk: Looks like the `join` isn't needed here and must have been left over updated in https://github.com/vercel/vercel/pull/8172/commits/25db496dcfad31095802c3da3a5b4c27ddc017b3

packages/next/src/server-build.ts
ijjk: These should be the only manifests yeah updated to use const/invariant in https://github.com/vercel/vercel/pull/8172/commits/25db496dcfad31095802c3da3a5b4c27ddc017b3

styfle (approved):

TooTallNate (approved):

PR #8183, opened 2022-07-18
[cli] Fix dev query string merging

### LINKED ISSUE #7283: [CLI] When "vercel dev" proxies a request to a framework, its queries are not original ones and will result in errors for Vite.
When Vite plugins processing a Vue Single File Component(SFC) file, it will transform its style tag into a js file with a url like 
`/foo/bar/AComponent.vue?vue&type=style&index=0&lang.css`. Then it was used as a js module in the page.

However when Vercel CLI dev server proxying it to frontend dev server, the cli doesn't use original url but generates a new one.
Generated one is like: `/foo/bar/AComponent.vue?vue=&type=style&index=0&lang.css=`, which is not a valid css url that can be recognized by Vite. Its content is not processed by css plugins of Vite, but it's still used as a js module.

Because of that, when running "vercel dev", if you write some css rules in the style tag of a Vue SFC file, the page will become blank and throw errors in browser console. It's very inconvenient for Vue developers if we can't use style tag in SFC.

Cause is at: 
https://github.com/vercel/vercel/blob/bae26cc0178ff49c3050ceaf5db00969e510cc6b/packages/cli/src/util/dev/server.ts#L1703-L1707

Here it deletes original querystring (`origUrl.search`), and generates a new one together with pathname (`url.format(origUrl)`), which is the one with additional "=". (`url.format()` internally uses `querystring.stringify()` that will add "=" for empty fields in the object).

In my local repo, I deleted L1704 & L1706 then it worked as expected. `url.format()` now uses original search string, so additional "="s are gone. But I don't know if it will impact other frameworks.

styfle:
Since `vercel dev` performs query string merging with `vercel.json` routes and the requested url, there is sometimes a slight difference between the request that the browser sends and the request that the upstream dev server receives.

Most servers treat the query string key with "empty" value the same as undefined value. Meaning, these two URLs should be equivalent because each has empty values:

- `http://example.com/src/App.vue?vue&type=style&index=0&lang.css`
- `http://example.com/src/App.vue?vue=&type=style&index=0&lang.css=`

However, `vite dev` handles these two URLs differently because of [string comparisons](https://github.com/vitejs/vite/blob/2289d04af5398791c3a01c9e597dc976e593c852/packages/plugin-vue/src/handleHotUpdate.ts#L98-L99) instead of URL parsing, which causes requests from `vercel dev` to fail. Thus, this PR changes from Node.js native URL parsing to a custom implementation in order to handle this corner case and preserve the difference between `?a=` and `?a`.

Ideally this would be [fixed in vite](https://github.com/vitejs/vite/pull/9589), however we might run into other dev servers that also fail in the same way in the future.

- Fixes https://github.com/vercel/vercel/issues/7283
- Closes https://github.com/vitejs/vite/pull/9589
- Related to https://github.com/nodejs/node/issues/9500


micahlt:
What's the status of this?  It'd be great to have this merged.

styfle:
The tests are failing and there are merge conflicts. 

It might be faster to [submit a PR to Vite](https://github.com/vitejs/vite/pull/9589) to fix this since it’s the only framework that has an issue.

packages/cli/src/util/dev/parse-query-string.ts
```diff
@@ -0,0 +1,41 @@
...
+  for (let param of params) {
+    const [key, value] = param.split('=');
+
+    let existing = map.get(key);
+    if (!existing) {
+      existing = [];
+      map.set(key, existing);
+    }
+
+    existing.push(value);
```
cb1kenobi, line 29: Do you need to `decodeURIComponent()` the value?

TooTallNate: Ya I believe we need to decode both the key and value.

styfle: Just to be clear, are suggesting decoding during parse() and encoding during format()?

cb1kenobi: I only tested the parse, but you should probably also `encodeURIComponent()` during format.

styfle: I updated in [a5aee22](https://github.com/vercel/vercel/pull/8183/commits/a5aee2278269a19d2974e3888330bb720ff3ee01) so this behavior matches `url.parse()` and `url.format()` with decode and encode respectively. Nice catch! 👍

packages/cli/src/util/dev/parse-query-string.ts
```diff
@@ -0,0 +1,47 @@
+export function parseQueryString(querystring?: string) {
+  let map = new Map<string, string[]>();
```
Ethan-Arrowood: nit: prefer const

```suggestion
  const map = new Map<string, string[]>();
```

packages/cli/test/unit/util/dev/parse-query-string.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const parsed = parseQueryString(querystring);
+    const map = new Map<string, string[]>([
+      ['a', []],
+      ['b', ['1']],
+      ['c', ['2', '3']],
+      ['d', ['']],
+      ['space bar', ['4']],
+      ['html', ['<h1>']],
+    ]);
+    expect(parsed).toEqual(map);
```
Ethan-Arrowood: should these be strict/deep equals?

Ethan-Arrowood (approved):
lgtm, just one code nit and then a question about the assertions used in the tests

packages/cli/test/unit/util/dev/parse-query-string.test.ts
styfle: This is jest which is deep equal already https://jestjs.io/docs/expect#toequalvalue

If it wasn't deep equal it would fail because its comparing different references 🙂

packages/cli/src/util/dev/parse-query-string.ts
styfle: Fixed in [2d36b10](https://github.com/vercel/vercel/pull/8183/commits/2d36b101432ad91b96128c33547568c3fc1c8d0a). I guess we should update to https://github.com/vercel/style-guide 🙂

packages/cli/test/unit/util/dev/parse-query-string.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const format = formatQueryString(parsed);
+    expect(format).toEqual(undefined);
+  });
+  it('should work with question mark', async () => {
+    const parsed = parseQueryString('?');
+    expect(parsed).toEqual(new Map());
+    const format = formatQueryString(parsed);
+    expect(format).toEqual(undefined);
+  });
+  it('should work without question mark', async () => {
```
TooTallNate, line 34: What is this case for?

packages/cli/test/unit/util/dev/parse-query-string.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  parseQueryString,
+  formatQueryString,
+} from '../../../../src/util/dev/parse-query-string';
+
+describe('parseQueryString', () => {
+  it('should parse to Map and format back to original String', async () => {
+    const querystring = '?a&b=1&c=2&c=3&d=&space%20bar=4&html=%3Ch1%3E';
+    const parsed = parseQueryString(querystring);
+    const map = new Map<string, string[]>([
+      ['a', []],
```
TooTallNate: I thought this case would have `[undefined]` as the value here? What if we're parsing `?a&a&a&a&a=1`?

TooTallNate: I also wonder if Map is even really the correct data structure here. How do we retain ordering when formatting again? i.e. `parse('?a=1&b=2&a=3'`)? I think the current code would format _both_ `a` values before the `b` value.

packages/cli/test/unit/util/dev/parse-query-string.test.ts
styfle: These are just unit tests so I wanted to have full coverage. I don't think this is actually possible when passing the result of `url.parse(foo).search` into this function.

packages/cli/test/unit/util/dev/parse-query-string.test.ts
cb1kenobi: That's a good question. I have used `Map` many times, but never with the intent of preserving key order, which I know `Object` does. Silly question, why not just use `URLSearchParams`?

styfle: > The Map object holds key-value pairs and remembers the original insertion order of the keys. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

styfle: Although I think you might be right about the duplicate keys that are undefined. Let me add a test for that 👀

TooTallNate: The point about Map insertion order doesn't really address the issue though. My example had `a,b,a` and right now the code will format back to `a,a,b`.

cb1kenobi: FYI, `URLSearchParams` allows for duplicate keys:

```
> const u = new URLSearchParams('?a=1&b=2&a=3');
undefined
> u
URLSearchParams { 'a' => '1', 'b' => '2', 'a' => '3' }
> u.entries()
URLSearchParams Iterator { [ 'a', '1' ], [ 'b', '2' ], [ 'a', '3' ] }
```

It also supports with or without the `?` and empty values:

```
> new URLSearchParams('a=&b=&a')
URLSearchParams { 'a' => '', 'b' => '', 'a' => '' }
```

styfle: @cb1kenobi Unfortunetly this suffers from the same bug:

```
> new URLSearchParams('a=&b=&a')
URLSearchParams { 'a' => '', 'b' => '', 'a' => '' }
```

Because we need the first `a` to be empty string and the second `a` to be undefined. Otherwise we can't tell which one had an equals sign (see PR description).

I switched the code back to a plain Object so that its compatible with the existing `url.parse()` and `url.format()` API. 

I think the order of `a,b,a` to `a,a,b` will need to be fixed in a future PR since that is out of scope for the `vite dev` fix and that is the issue customers are blocked on.

EndangeredMassa (approved):

TooTallNate (approved):

packages/cli/src/util/dev/parse-query-string.ts
```diff
@@ -0,0 +1,59 @@
+/**
```
cb1kenobi, line 1: Since this implementation does not preserve the order of the params, and assuming we care, perhaps there should be a TODO stating the limitation.

packages/cli/src/util/dev/router.ts
```diff
@@ -56,7 +57,8 @@ export async function devRouter(
   phase?: HandleValue | null
 ): Promise<RouteResult> {
   let result: RouteResult | undefined;
-  let { query, pathname: reqPathname = '/' } = url.parse(reqUrl, true);
+  let { pathname: reqPathname = '/', search: reqSearch } = url.parse(reqUrl);
```
cb1kenobi, line 60: With `url.parse()` having been deprecated, perhaps we should switch to the new, futureproof method:

```javascript
let { pathname: reqPathname = '/', search: reqSearch } = new URL(reqUrl);
```

Note that the URL must have a protocol and domain, otherwise the call would look like `new URL(reqUrl, 'http://localhost')`.

micahlt:
🎉 Thank you so much for your hard work on this @styfle!

packages/cli/src/util/dev/router.ts
styfle: Its not really deprecated https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost

image

cb1kenobi: Ha, revoked!

PR #8193, opened 2022-07-19
[cli] Write top-level error to `builds.json` file in "vc build"

TooTallNate:
If an error happens outside of a Builder (i.e. `detectBuilders()` function fails), then write the serialized error into the `builds.json` file at the top-level of the file (there is no `builds[]` when an error happens at the top-level).


packages/cli/test/unit/util/error.test.ts
```diff
@@ -187,3 +191,43 @@ describe('responseError', () => {
...
   });
 });
+
+describe('toEnumerableError()', () => {
+  it('should JSON stringify Error', () => {
+    const err = new Error('An error');
+    const enumerable = toEnumerableError(err);
+    expect(JSON.stringify(err)).toEqual('{}');
+
+    // Delete `stack` since it makes stringify undeterministinc
```
styfle:
```suggestion
    // Delete `stack` since it makes stringify nondeterministic
```

styfle (approved):

PR #8194, opened 2022-07-19
[cli] Prevent `vc deploy --prebuilt` when `vc build` failed

TooTallNate:
If `vc build` failed then don't allow a deployment to be created with `vc deploy --prebuilt`.


packages/cli/src/util/deploy/get-prebuilt-json.ts
```diff
@@ -1,7 +1,10 @@
...
 import { join } from 'path';
+import { BuildsManifest } from '../../commands/build';
 
-export default async function getPrebuiltJson(directory: string) {
+export default async function getPrebuiltJson(
+  directory: string
+): Promise<BuildsManifest | null> {
   try {
     return await fs.readJSON(join(directory, '.vercel/output/builds.json'));
   } catch (error) {
```
EndangeredMassa, line 10: Can we debug log this error?

EndangeredMassa (approved):

styfle (approved):
Thanks!

PR #8199, opened 2022-07-20
[cli] Remove legacy config file migration logic

TooTallNate:
Removes the legacy config file migration logic from back in the days when Zeit CLI supported multiple "providers". This was from a _very_ long time ago and we should expect that anyone who would have migrated at this point, has.


packages/cli/src/util/config/get-default.ts
```diff
@@ -1,75 +1,15 @@
...
-      migrated = true;
-    } catch (err) {}
-  }
-
-  return { config, migrated };
+export const defaultGlobalConfig: GlobalConfig = {
+  '// Note':
+    'This is your Vercel config file. For more information see the global configuration documentation: https://vercel.com/docs/configuration#global',
+  '// Docs':
+    'https://vercel.com/docs/project-configuration#global-configuration/config-json',
```
TooTallNate, line 7: Fixed these docs links too since they pointed to an incorrect page.

packages/cli/src/util/config/get-default.ts
```diff
@@ -1,75 +1,15 @@
...
-      }
-
-      migrated = true;
-    } catch (err) {}
-  }
-
-  return { config, migrated };
+export const defaultGlobalConfig: GlobalConfig = {
+  '// Note':
+    'This is your Vercel config file. For more information see the global configuration documentation: https://vercel.com/docs/configuration#global',
```
styfle: This link is redirecting, should it be updated too?

TooTallNate: Whoops, removed.

styfle (approved):

styfle (approved):

TooTallNate:
Though this is technically a breaking change, we decided to merge as a "patch" change since the logic is so legacy that anyone who needed to migrate should have at this point.

PR #7747, opened 2022-05-02
[frameworks] Add Astro

tony-sull:
This PR adds [Astro](https://astro.build) to the list of supported frameworks.

The example used in this PR is our latest [`basics`](https://github.com/withastro/astro/tree/main/examples/basics) starter that is also available with `create-astro`.

The public demo can be found [here](https://astro-template.vercel.app/)

### Related Issues

Related to [#6477](https://github.com/vercel/vercel/pull/6477)

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8200, opened 2022-07-21
[frameworks] Fix astro default routes since it is not SPA

lucleray:
The default routes for Astro are redirecting all non-existing files to `index.html`, which means that the `404.html` page is not used, and instead `index.html` is shown for all not found pages.

Astro outputs files for each page (ie. `about.html`, `blog.html`, ...) so the `{handle: 'filesystem'}` route should be enough to route all existing pages correctly.

The missing part to ship this fix is to answer the following question: can we safely assume that Astro will always output a `404.html` file?

Internal ref: https://github.com/vercel/customer-issues/issues/638


packages/frameworks/src/frameworks.ts
```diff
@@ -277,8 +277,9 @@ export const frameworks = [
         handle: 'filesystem',
       },
       {
-        src: '/(.*)',
-        dest: '/index.html',
```
styfle, line 279: This will break existing astro deployments that use SPA pattern.

See https://vercel.com/guides/custom-404-page

okikio: Yeah, we need to explicitly specify that only 404 statuses can return a 404 error page, but other redirects work as expected

okikio (dismissed):
We need a way to have the redirects work but still have the 404 pages work, ideally leaving it to Astro to decide. Yesterday I did some research into it, and the problem is much larger in scope on the Astro side, I'll see if I can get a possible fix out, but if not this might be the only solution we have for now.

packages/frameworks/src/frameworks.ts
lucleray: @styfle Ah. I didn't know that Astro outputs SPA pages. I thought it would always a file for each page (I was reading the [docs here](https://docs.astro.build/en/core-concepts/astro-pages/) but I'm not really familiar with Astro).

In other words, I thought Astro is a "static site generator", and should work [like specified here](https://vercel.com/guides/custom-404-page#static-site-generator-(ssg)).

lucleray: > Yeah, we need to explicitly specify that only 404 statuses can return a 404 error page, but other redirects work as expected

Aren't all Astro outputs static? How would these static assets return 404 statuses?

okikio: Yeah but that's only if you're using Astro in static mode, Astro also support SSR mode, I think that what @styfle is referencing. 
> I'm not familiar with how Vercel handles SSR, so I might be misunderstanding somethings

styfle: Can we detect SPA vs SSG mode, and change the routes dynamically based on astro config?

lucleray: @styfle How do you use Astro in SPA mode?

okikio: Call it SSR mode not SPA mode, it sounds confusing when it's called SPA mode, as Astro doesn't have an official SPA mode. To answer your question @styfle yes you should be able to, I know that the Astro team has discussed adding some sort of variable value to determine what mode is being used, I'll contact the Astro core team, to determine if this issue can be quickly fixed.

styfle: @lucleray I'm not sure, I haven't used Astro.

This route was added by @tony-sull in the original PR to add Astro support https://github.com/vercel/vercel/pull/7747

Presumably that route was necessary for the Astro example/demo that was added there too.

okikio: Ok, I've got confirmation it seems we're supposed to use ~`import.meta.env.SSR`~ to determine if Astro is running in SSR or SSG mode

lucleray: > Yeah but that's only if you're using Astro in static mode, Astro also support SSR mode, I think that what @styfle is referencing.

If Astro is used in SSR mode, I think that it's Astro itself (the Astro "server") that should render the 404.html page and return a 404 status when the page is not available. 

But I don't think that we currently support Astro in SSR mode (at least not with this configuration - it might work with the Astro adapter and `vc build`).

okikio: Squash that, `import.meta.env.SSR` is currently buggy, I'll keep doing some investigation

okikio: The more I think about it, @lucleray I think your solution here should work for Astro as there is no SPA mode for Astro. However, it may break other frameworks. The SPA frameworks specified by @styfle here https://vercel.com/guides/custom-404-page, might break

okikio: ~Good news everybody Astro has merged a fix from their end https://github.com/withastro/astro/pull/4018#event-7048931698~

EndangeredMassa:
@okikio @lucleray what's the path forward on this work?

lucleray:
I'm going to close it for now. If it still makes sense to ship it, we can reopen.

packages/frameworks/src/frameworks.ts
okikio: Astro doesn't conventionally do SPA's by default

> @delucis : Astro doesn’t really do SPA builds. I mean, they’re technically possible if you just have src/pages/index.astro and mount a client-side routed app there, but it’s MPA as a basic premise both in static and SSR modes.

Is there a way to enable this pr while not harming other SPA frameworks?

delucis:
Hi everyone, Astro core maintainer here 👋

I was about to make a PR more or less identical to this and thought it might be best to discuss here first — maybe we can re-open.

We currently document a weird hack in [the Astro Vercel deploy guide](https://docs.astro.build/en/guides/deploy/vercel/#static-site), where setting `"cleanUrls": true` in `vercel.json` makes custom 404 pages display for static builds, but I’ve never understood why that would work. Would be cool if that wasn’t needed. 

There seems to be a bit of confusion between “SPA” mode and “SSR” in the discussion above, Astro doesn’t have an SPA-mode where you would route everything to `/index.html` and let client-side routing take over, but does support SSR via [its Vercel adapter](https://docs.astro.build/en/guides/integrations-guide/vercel/).

In SSR mode, Astro’s adapter outputs routing configuration so I believe this `defaultRoutes` is ignored. In static mode, Astro builds an HTML file per page, so the filesystem routing should work, we just need to handle the special 404 case.

> The missing part to ship this fix is to answer the following question: can we safely assume that Astro will always output a 404.html file?

The answer here is no: if a user does not create a `src/pages/404.astro` file, there won’t be a `dist/404.html` to route to. I assume in this case Vercel displays a generic 404 page?

styfle (approved):
I tried the example from the original PR https://github.com/vercel/vercel/pull/7747 as well as the latest example and both are SSG style, so I don't know why routes were added to begin with.

I discussed with @TooTallNate and we found that removing these routes (and fixing the asset header) would make astro work perfectly, thanks!

PR #8204, opened 2022-07-21
[node] Add WebAssembly importing support for the @vercel/node builder

Schniz:
This PR implements WebAssembly importing support like Next.js has in the Edge Runtime for `vc dev`.
It works on a similar idea: find all `*.wasm?module` imports and convert them to a global import.

For production builds, I patched the resolution algorithm for nodeFileTrace to remove the `?module` suffix
so that it will find the WASM files and mark them as dependencies.

### Related Issues

- https://github.com/vercel/next.js/pull/34437

#### Tests

#### Code Review


packages/node/src/edge-wasm-plugin.ts
```diff
@@ -0,0 +1,89 @@
+import type { Plugin } from 'esbuild';
```
EndangeredMassa, line 4: Is this file copy-pasted from the other place where we support this? Should we share this logic directly somewhere?

packages/node/test/fixtures/61-edge-function/api/wasm.js
```diff
@@ -0,0 +1,13 @@
+import wasm from '../increment.wasm?module';
```
EndangeredMassa, line 1: Should we also support `.wat` files? Are those supposed to be executable or just a human-legible representation?

EndangeredMassa (approved):
I tested this with some small wasm examples and everything works great. Is there anything else we should test?

It would be nice if we included the wasm source and a script to produce it in case we need to change it (or produce a different fixture) in the future.

packages/node/src/edge-wasm-plugin.ts
Schniz: Nope but it is very similar to what we do in Next.js and when deploying edge functions

packages/node/test/fixtures/61-edge-function/api/wasm.js
Schniz: These files are not yet supported in Next.js either so we can wait until there’s a requirement for such files I think

Schniz:
The wasm file was copied from the Next.js repo 🫣

packages/node/src/edge-wasm-plugin.ts
styfle: Can you add a link to the source code in a comment?

packages/node/test/fixtures/61-edge-function/api/wasm.js
styfle: I wouldn't expect `.wat` to work since no runtime supports them currently that I'm aware of

packages/node/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "strict": true,
     "esModuleInterop": true,
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "DOM"],
```
styfle, line 5: What is DOM used for here?

packages/node/src/dev-server.ts
```diff
@@ -228,21 +229,30 @@ async function compileUserCode(
...
+    const wasmBindings = await params.wasmAssets.getContext();
     const edgeRuntime = new EdgeRuntime({
-      initialCode: userCode,
+      initialCode: params.userCode,
       extend: (context: Primitives) => {
-        Object.assign(context, {
-          __dirname: '',
-          module: {
-            exports: {},
+        Object.assign(
```
styfle, line 245: Looks like there's a merge conflict here

EndangeredMassa: My env var PR caused this conflict. I just resolved it here.

packages/node/tsconfig.json
EndangeredMassa: This is necessary for the `WebAssembly` type. If we remove it, we get the following typescript errors:

```
src/edge-wasm-plugin.ts(22,46): error TS2503: Cannot find namespace 'WebAssembly'.
src/edge-wasm-plugin.ts(22,46): error TS4055: Return type of public method from exported class has or is using private name 'WebAssembly'.
src/edge-wasm-plugin.ts(24,35): error TS2503: Cannot find namespace 'WebAssembly'.
src/edge-wasm-plugin.ts(29,33): error TS2304: Cannot find name 'WebAssembly'.
```

packages/node/src/edge-wasm-plugin.ts
EndangeredMassa: I tried looking for the original, but couldn't find it. I asked around for someone to point it out.

EndangeredMassa: I added a comment at the top of this file linking to the original.

packages/node/src/dev-server.ts
```diff
@@ -228,24 +229,33 @@ async function compileUserCode(
...
-          __dirname: '',
-          module: {
-            exports: {},
-          },
-          process: {
-            env: process.env,
+        Object.assign(
+          context,
+          {
+            __dirname: '',
```
cb1kenobi, line 248: Do you also need `__filename` for consistency?

EndangeredMassa: I was only adding what was necessary to support the wrapping logic used by esbuild to bundle the code into a single file. If we add too much here, it may hide the actual errors from the user code because `__filename` will exist, but it won't have an expected value, causing some other error.

I should have added a comment about what exactly this was doing.

It's good that you brought this up! These values were necessary for the `ncc` bundle, but after switching to `esbuild`, I didn't re-evaluate that they were all still required. It turns out they are not.

I updated this with less values and comments about why they are here.

EndangeredMassa: This change isn't strictly related to this PR, though. I opened a new PR with these changes: https://github.com/vercel/vercel/pull/8260

styfle (approved):

PR #8209, opened 2022-07-21
[utils] allow a github PAT token to get past rate limit

EndangeredMassa:
If you run `yarn changelog` when your IP is already rate limited by github, you'll get an error. This allows you to set a [Github PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) to get past the generic rate limit.

Given a PAT with value `MY_PAT_TOKEN`, you can run:

```
$ GITHUB_TOKEN=MY_PAT_TOKEN yarn changelog
```

If you do get an error, it now actually shows up in the output:

```
$  yarn changelog
yarn run v1.22.18
$ node utils/changelog.js
Error: Failed to fetch releases from github: API rate limit exceeded for 98.139.180.149. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

### DIFF
--- a/utils/changelog.js
+++ b/utils/changelog.js
@@ -11,7 +11,7 @@ process.chdir(join(__dirname, '..'));
 async function getLatestStableTag() {
   const headers = {};
 
-  const token = process.env.GITHUB_PAT;
+  const token = process.env.GITHUB_TOKEN;
   if (token) {
     headers['authorization'] = `token ${token}`;
   }
--- a/utils/changelog.js
+++ b/utils/changelog.js
@@ -22,8 +22,13 @@ async function getLatestStableTag() {
       headers,
     }
   );
-  const { tag_name } = await res.json();
-  return tag_name;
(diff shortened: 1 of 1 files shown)


utils/changelog.js
```diff
@@ -9,11 +9,26 @@ const groupLog = require('./changelog/group');
 process.chdir(join(__dirname, '..'));
 
 async function getLatestStableTag() {
+  const headers = {};
+
+  const token = process.env.GITHUB_PAT;
```
TooTallNate: My instinctual naming would be `GITHUB_TOKEN`, but not a huge deal.

TooTallNate (approved):

utils/changelog.js
EndangeredMassa: That is better. You can use different types of tokens and this will still work. I updated it.

styfle (approved):

PR #8211, opened 2022-07-21
[cli] recreate symlinked files instead of copying

cb1kenobi:
Instead of copying symlinked files during a build, recreate the symlink.

### Related Issues

> https://linear.app/vercel/issue/VCCLI-122/handle-symlinks-when-writing-build-results

#### Tests

#### Code Review


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -235,7 +235,20 @@ async function writeStaticFile(
   const dest = join(outputDir, 'static', fsPath);
   await fs.mkdirp(dirname(dest));
 
-  // TODO: handle (or skip) symlinks?
+  // if the source is a symlink, try to create it instead of copying the file
+  // however, if the symlink creation fails (e.g. Windows), fallback to copying
+  if (
+    file.type === 'FileFsRef' &&
```
TooTallNate: This is correct, but only handles the `FileFsRef` case. To be complete, we should handle the other `File` type implementations.

Thinking about it a bit more now, I realize that we already have the necessary logic [in this function](https://github.com/vercel/vercel/blob/a82f1172179b4906b35b005e7a45b27a760fd82f/packages/build-utils/src/fs/download.ts#L45-L57). So perhaps let's just export that function from build-utils and we can use it here.

TooTallNate (changes requested):

packages/build-utils/src/fs/download.ts
```diff
@@ -42,14 +42,23 @@ async function prepareSymlinkTarget(
...
   if (isSymbolicLink(mode)) {
     const target = await prepareSymlinkTarget(file, fsPath);
 
-    await symlink(target, fsPath);
-    return FileFsRef.fromFsPath({ mode, fsPath });
+    try {
+      await symlink(target, fsPath);
+      return FileFsRef.fromFsPath({ mode, fsPath });
+    } catch (e) {
+      // fall through
```
TooTallNate: This fall through case is a bit tricky because for non-`FileFsRef` types (`FileBlob`, specifically), the "contents" of the file are actually the symlink target. So do do this properly we'd need to pass in the full `files` object to this function and replace the `file` reference here to be the resolved one (`file = files[target]` or similar).

TooTallNate (changes requested):

packages/build-utils/src/fs/download.ts
cb1kenobi: OK, so what you're saying is if `file` is a `FileBlob` or `FileRef`, the contents IS a symlink and if it fails, we really don't want to fall through.

I'm just worried because a month ago I ran into a problem with creating symlinks on Windows where it will fail with `EPERM` if not running as administrator.

If we truly want symlink creation failures to throw, then perhaps in the `catch` we say if on Windows and code equals `EPERM`, then rewrite the error to say `"Unable to create symlink due to Windows group policy setting; either allow symlinks or run as administrator"`.

Not sure I'm following your reference to `files`. I'm calling `downloadFile()`, not `download()`.

cb1kenobi: I removed the try/catch. We can improve the error should it become an issue.

cb1kenobi:
Closed in favor of #8270.

PR #8217, opened 2022-07-22
[remix] add image optimization configuration

theMosaad:
### Feature
Makes it possible to use [Vercel's image optimization](https://vercel.com/docs/build-output-api/v3#build-output-configuration/supported-properties/images).

### Demo
This example is built locally with this patch:
https://github.com/theMosaad/remix-vercel-image


leerob:
This is amazing!

theMosaad:
Happy to add tests if needed. If that's the case, please, let me know the kind of tests you expect.

jorisre:
Heyy this is awesome ! 
How can I help to release that feature faster?

packages/remix/src/build.ts
```diff
@@ -196,6 +197,157 @@ export const build: BuildV2 = async ({
       const rootDirectory = relative(repoRootPath, workPath);
       serverBuildPath = join(rootDirectory, serverBuildPath);
     }
+
+    if (remixConfig.images) {
```
styfle: Is this a feature of `remix.config` or are you adding a new property that would only be used for Vercel?

If its the latter, then it should probably be `remixConfig.vercel.images` instead to avoid conflicts with Remix official configuration properties.

theMosaad: Makes total sense. Done!

packages/remix/src/build.ts
```diff
@@ -196,6 +198,157 @@ export const build: BuildV2 = async ({
...
+    if (images) {
+      if (typeof images !== 'object') {
+        throw new Error(
+          `remix.config.js "images" should be an object received ${typeof images}.`
+        );
+      }
+
+      if (!Array.isArray(images.domains)) {
+        throw new Error(
+          `remix.config.js "images.domains" should be an Array received ${typeof images.domains}.`
```
styfle: Should all of these error messages be `vercel.images.*` now?

theMosaad: Nice catch! they are now.

styfle (approved):
Thanks!

TooTallNate:
I wonder if it would make more sense to add this property to the `vercel.json` file instead, that way, any project could configure the Image Optimization feature, regardless of which framework is being used.

EndangeredMassa:
Hey @theMosaad! Sorry for the trouble, but our bot auto-merged this PR before it was ready. I just reverted it here https://github.com/vercel/vercel/pull/8375.

If you could open a new PR about this, moving the property to `vercel.json`, that'd be awesome. Thanks!

theMosaad:
Hey @EndangeredMassa, no trouble!

Opened a new PR https://github.com/vercel/vercel/pull/8379 but it seems that Vercel's deployment api rejects the request with the new `images` property. Is there a way to get around that or should I wait until you support it internally?

PR #8222, opened 2022-07-22
[cli] Add "dev" script, remove "build-dev" script

TooTallNate:
I don't think anyone is using this command anymore in the world of M1.

Anyways, we should be using `ts-node` to execute the CLI from source code instead of skipping steps at build-time to iterate quickly.


EndangeredMassa (approved):

packages/cli/scripts/build.ts
```diff
@@ -27,33 +27,31 @@ function envToString(key: string) {
...
+  console.log();
+  await execa(process.execPath, [join(__dirname, 'compile-templates.js')], {
+    stdio: 'inherit',
+  });
 
   // Do the initial `ncc` build
   console.log();
-  const args = ['ncc', 'build', '--external', 'update-notifier'];
-  if (isDev) {
-    args.push('--source-map');
```
styfle, line 54: I think we need to keep this because we want to emit source maps when building for development so we can attach a debugger

styfle (changes requested):
> we should be using ts-node to execute the CLI from source code instead of skipping steps at build-time to iterate quickly

Can you add a script for that instead of removing the existing one?

packages/cli/scripts/build.ts
TooTallNate: Did you do `yarn build-dev` in these situations?

TooTallNate:
@styfle I added a `dev` script. Can you take another look?

packages/cli/scripts/build.ts
styfle: Yes

packages/cli/scripts/build.ts
```diff
@@ -27,33 +27,31 @@ function envToString(key: string) {
...
 
-  if (!isDev) {
-    // Read the secrets from GitHub Actions and generate a file.
-    // During local development, these secrets will be empty.
-    await createConstants();
+  // `vercel dev` uses chokidar to watch the filesystem, but opts-out of the
+  // `fsevents` feature using `useFsEvents: false`, so delete the module here so
+  // that it is not compiled by ncc, which makes the npm package size larger
+  // than necessary.
+  await remove(join(dirRoot, '../../node_modules/fsevents'));
```
styfle, line 38: Is it okay to remove this every time?

styfle (approved):

packages/cli/scripts/build.ts
TooTallNate: Doesn't hurt. Was already doing that before anyways.

PR #8224, opened 2022-07-22
[cli] Use `ts-node` with SWC instead of `ts-eager`

TooTallNate:
`ts-eager` isn't really maintained anymore and doesn't support ESM packages.

`ts-node` with SWC is more or less just as fast, so let's use that instead since it's well maintained any we can dogfood SWC more.


styfle (approved):
Great, thanks!

MatthewStanciu:
💯

PR #8230, opened 2022-07-22
[build-utils] Fix package.json and lockfile detection

styfle:
This PR fixes a couple issues where `vercel build` was not correctly detecting the package.json files

```
Error: @vercel/node:test: ERROR: frontend/index.ts(1,12): error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
```

It also fixes an issue where all deployments were incorrectly detecting the lock file because the lock file doesn't always live in the same directory as the package.json file. So we need to do 2 passes: one to find the nearest package.json and one to find the nearest lock file.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -392,6 +372,34 @@ export async function walkParentDirs({
   return null;
 }
 
+async function walkParentDirs2({
```
styfle: This is similar to `walkParentDirs` above but `walkParentDirs2` supports an array of filenames and returns an array of filenames. I think we could eventually merge the two but since `walkParentDirs` is exported publicly, I didn't want to change it.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -305,67 +305,47 @@ export async function scanParentDirs(
 ): Promise<ScanParentDirsResult> {
   assert(path.isAbsolute(destPath));
 
-  let cliType: CliType = 'yarn';
-  let packageJson: PackageJson | undefined;
-  let packageJsonPath: string | undefined;
-  let currentDestPath = destPath;
+  const pkgJsonPath = await walkParentDirs({
+    base: '/',
```
TooTallNate, line 309: Maybe let's add a TODO comment here to allow `base` to be specified as a parameter? I'm also not sure if this will work on Windows.

styfle: I guess we'll find out when the unit tests run 😄 

Ideally I would like to pass in the `repoRootPath` here but `scanParentDirs` doesn't accept that parameter today, so we at least need a fallback

styfle: I tested windows and it is indeed a problem so I added a special case here [1a31a66](https://github.com/vercel/vercel/pull/8230/commits/1a31a665e4cb17d0f70fd3acf923ea2d29edc0f6)

packages/build-utils/test/unit.test.ts
```diff
@@ -454,41 +454,69 @@ it('should return lockfileVersion 2 with npm7', async () => {
...
+  const result = await scanParentDirs(fixture);
+  expect(result.cliType).toEqual('yarn');
+  expect(result.lockfileVersion).toEqual(undefined);
+  expect(result.packageJsonPath).toEqual(path.join(fixture, 'package.json'));
+});
+
+it('should detect package.json in nested frontend', async () => {
+  const fixture = path.join(
+    __dirname,
+    '../../node/test/fixtures/18.1-nested-packagejson/backend'
```
EndangeredMassa: Was this fixture intended to be `/frontend`?

styfle: Yep, nice catch!

EndangeredMassa (approved):

packages/build-utils/test/unit.test.ts
styfle:
```suggestion
    '../../node/test/fixtures/18.1-nested-packagejson/frontend'
```

PR #8232, opened 2022-07-23
[cli] Update to TypeScript v4.7.4

TooTallNate:


packages/cli/scripts/build.ts
```diff
@@ -77,6 +77,7 @@ async function main() {
 
   // Band-aid to bundle stuff that `ncc` neglects to bundle
   await cpy(join(dirRoot, 'src/util/projects/VERCEL_DIR_README.txt'), distRoot);
+  await cpy(join(dirRoot, 'src/util/dev/builder-worker.js'), distRoot);
```
TooTallNate, line 80: Strangely, `ncc` is no longer including this file into the dist dir as a static asset. This line is missing from the build logs:

```
vercel:build:    2kB  dist/builder-worker.js
```

packages/cli/src/util/dev/server.ts
```diff
@@ -1533,18 +1544,10 @@ export default class DevServer {
...
         // `startDevServer()` threw an error. Most likely this means the dev
         // server process exited before sending the port information message
         // (missing dependency at runtime, for example).
-        if (err.code === 'ENOENT') {
-          err.message = `Command not found: ${chalk.cyan(
-            err.path,
-            ...err.spawnargs
-          )}\nPlease ensure that ${cmd(err.path)} is properly installed`;
-          err.link = 'https://vercel.link/command-not-found';
-        }
```
styfle, line 1546: Should this be removed or should we keep it?

styfle: Maybe we should pretty print the error

packages/cli/src/util/env/known-error.ts
```diff
@@ -7,7 +9,8 @@ const knownErrorsCodes = new Set([
   'ENV_SHOULD_BE_A_SECRET',
 ]);
 
-export function isKnownError(error: { code?: string }) {
-  const code = error && typeof error.code === 'string' ? error.code : '';
+export function isKnownError(error: unknown) {
+  const code = isErrnoException(error) ? error.code : null;
+  if (!code) return false;
```
styfle, line 14:
```suggestion
  const code = isErrnoException(error) ? error.code : '';
```

packages/cli/src/util/login/email.ts
```diff
@@ -51,9 +53,9 @@ export default async function doEmailLogin(
         'Email',
         ssoUserId
       );
-    } catch (err) {
-      if (err.serverMessage !== 'Confirmation incomplete') {
-        output.error(err.message);
+    } catch (err: unknown) {
+      if (!isAPIError(err) || err.serverMessage !== 'Confirmation incomplete') {
```
styfle, line 57: This seems wrong since the API error would not print an error and instead print success

packages/cli/src/util/projects/remove-domain-from-project.ts
```diff
@@ -21,7 +21,7 @@ export async function removeDomainFromProject(
     );
 
     return response;
-  } catch (err) {
+  } catch (err: any) {
```
styfle: Should this use `unknown` and check for `isAPIError(err)`?

packages/cli/src/util/login/email.ts
TooTallNate: It's using `||` here, so it _will_ print the API error message so long as `serverMessage !== 'Confirmation incomplete'`, so I think the logic is correct here.

packages/cli/src/util/env/known-error.ts
TooTallNate: `error.code` can be undefined in the case of `NodeJS.ErrnoException`, so returning early here makes the `code.toUpperCase()` happy below.

styfle (approved):

packages/cli/src/util/dev/server.ts
TooTallNate: Reverted it actually, since printing the error causes the test to fail since we search for "Error!" in the logs. We can address this in another PR.

cb1kenobi (approved):

PR #8239, opened 2022-07-25
[fs-detectors] Exclude the middleware builder if it's a Next.js app

nuta:
While `maybeGetApiBuilder()` excludes the middleware builder, however, we need another check for the case where `projectSettings.framework == null`.

#### Tests

#### Code Review


EndangeredMassa (approved):

packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -2281,6 +2281,30 @@ describe('Test `detectBuilders` with `featHandleMiss=true`', () => {
...
   });
+
+  it('should not add middleware builder when building "nextjs"', async () => {
+    const files = ['package.json', 'pages/index.ts', 'middleware.ts'];
+    const pkg = {
+      scripts: { build: 'next build' },
+      dependencies: { next: '12.2.0' },
+    };
+    const projectSettings = {
+      framework: null,
```
styfle: Lets add `createdAt` and set an old date since I think that is the only time we should be detecting Next.js from the dependencies (I believe this was to grandfather in old projects when we introduced project settings)

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -305,6 +305,22 @@ export async function detectBuilders(
     };
   }
 
+  // Exclude the middleware builder for Next.js apps since @vercel/next
+  // will build middlewares.
+  //
+  // While maybeGetApiBuilder() excludes the middleware builder, however,
+  // we need to check if it's a Next.js app here again for the case where
+  // `projectSettings.framework == null`.
+  if (frontendBuilder?.use === '@vercel/next' && apiBuilders.length > 0) {
```
styfle: Lets use the same logic that is used to detect nextjs here since the project createdAt date matters 

https://github.com/vercel/vercel/blob/181a492d912b4c11970f824afc43638519a5ef7f/packages/fs-detectors/src/detect-builders.ts#L497

styfle: In fact, you might be able to use `detectFrontBuilder()` instead

packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
nuta: I wrote this fix not for old projects, but because framework detection was not working in the one-click deploy button. The problem has already been fixed so this PR is meaningful only for:

- Apps incorrectly configured by users (settings "others" for Next.js).
- When framework detection become broken again for some reason in the future.

Both will unlikely happen but if happens they would cause relatively hard-to-debug behavior: build succeeds, but only middleware might behave differently.

styfle: If its not for old projects, then lets close it because thats the only valid scenario for framework "Other" doing any framework detection.

styfle (changes requested):
This should take into account the project creation date

https://github.com/vercel/vercel/blob/181a492d912b4c11970f824afc43638519a5ef7f/packages/fs-detectors/src/detect-builders.ts#L497

You might even use `detectFrontBuilder()`  instead

packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
nuta: At first I wrote this PR for fixing the one-click deploy button the problem but indeed it's still helpful for old projects. Will update this for old projects 👍

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -305,6 +305,27 @@ export async function detectBuilders(
...
+  // Exclude the middleware builder for Next.js apps since @vercel/next
+  // will build middlewares.
+  //
+  // While maybeGetApiBuilder() excludes the middleware builder, however,
+  // we need to check if it's a Next.js app here again for the case where
+  // `projectSettings.framework == null`.
+  if (
+    framework === null &&
+    detectFrontBuilder(pkg, files, usedFunctions, fallbackEntrypoint, options)
+      .use === '@vercel/next' &&
```
styfle:
```suggestion
    frontendBuilder?.use === '@vercel/next' &&
```

This can be simplified to use the result from detectFrontBuilder above

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -305,6 +305,27 @@ export async function detectBuilders(
...
+    apiBuilders.length > 0
+  ) {
+    apiBuilders = apiBuilders.filter(builder => {
+      const isMiddlewareBuilder =
+        builder.use === '@vercel/node' &&
+        builder.config?.middleware &&
+        (builder.src === 'middleware.js' || builder.src === 'middleware.ts');
```
styfle: Is there a scenario where `builder.config?.middleware` is true but the `src !== 'middleware.js/ts`?

Seems like checking for `builder.config?.middleware` should be sufficient.

packages/fs-detectors/src/detect-builders.ts
nuta: Fixed. I used `detectFrontBuilder` to reuse the `createdAt < Date.parse('2020-03-01')` condition. I copied that into this if statement instead.

nuta:
I'm seeing failures in CI but it appears the same happen in some other PRs. Are they just flaky?

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -305,6 +305,26 @@ export async function detectBuilders(
...
+  // will build middlewares.
+  //
+  // While maybeGetApiBuilder() excludes the middleware builder, however,
+  // we need to check if it's a Next.js app here again for the case where
+  // `projectSettings.framework == null`.
+  if (
+    framework === null &&
+    frontendBuilder?.use === '@vercel/next' &&
+    projectSettings.createdAt &&
+    projectSettings.createdAt < Date.parse('2020-03-01') &&
```
styfle, line 318: I don’t think you need to check the date here since it’s already checked when frontendBuilder is assigned

packages/fs-detectors/src/detect-builders.ts
styfle: I don’t think you need to check the date here since it’s already checked when frontendBuilder is assigned

styfle (approved):
Thanks!

PR #8242, opened 2022-07-25
[cli][node] update edge-runtime to allow `instanceof` to work with primitives

EndangeredMassa:
Code executed in `edge-runtime` would not match `instanceof` comparisons to certain primitives because of how exactly those values were injected into the VM. This was resolved by an overhaul of how exactly that happens in: https://github.com/vercel/edge-runtime/pull/27

This PR updates the use of `edge-runtime` to the latest version, which allows `instanceof` to match correctly in all cases.

The test for successful edge functions was updated to use `snake-case` because that logic was failing before inside `edge-runtime` because of the `instanceof` issue with: `Unhandled rejection: re.reduce is not a function`.

With this update, users can now use `instanceof` directly, but also import packages that happen to rely on `instanceof` working.


TooTallNate (approved):

EndangeredMassa:
Looks like the version update changed the interface. Will update.

packages/node/package.json
```diff
@@ -30,11 +30,12 @@
     }
   },
   "dependencies": {
+    "@edge-runtime/vm": "1.1.0-beta.20",
     "@types/node": "*",
     "@vercel/build-utils": "5.0.4",
     "@vercel/node-bridge": "3.0.0",
     "@vercel/static-config": "2.0.1",
-    "edge-runtime": "1.0.1",
+    "edge-runtime": "v1.1.0-beta.20",
```
cb1kenobi: Is the `v` allowed?

Kikobeats: it isn't, also the newest version is `1.1.0-beta.23`

EndangeredMassa: It's allowed, but we don't normally use it. I'll remove it and update the version.

packages/cli/test/dev/integration-1.test.ts
```diff
@@ -55,10 +55,17 @@ test('[vercel dev] should support edge functions', async () => {
...
-  })
+  testFixtureStdio(
+    'edge-function',
+    async (testPath: any) => {
+      await testPath(500, '/api/edge-500-response');
+      await testPath(200, '/api/edge-success');
+    },
+    {
+      // TODO: why is this failing in prod?
+      skipDeploy: true,
```
EndangeredMassa: I believe this (importing `snake-case`) used to work in production, but doesn't seem to anymore. I have a question out to the team about it.

cb1kenobi (approved):

EndangeredMassa:
The CI failures are known and being worked on elsewhere. They do not apply to this PR.

PR #8243, opened 2022-07-25
[cli] Apply `functions` configuration to Serverless Functions in "vc build"

TooTallNate:


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -162,19 +183,31 @@ async function writeBuildResultV2(
...
+  vercelConfig: VercelConfig | null
 ) {
   const { output } = buildResult;
   const src = build.src;
   if (typeof src !== 'string') {
     throw new Error(`Expected "build.src" to be a string`);
   }
+
+  let customFunctionConfiguration: FunctionConfiguration | undefined;
+  if (vercelConfig?.functions) {
```
styfle: Can we re-use the `getLambdaOptionsFromFunction()` function from `@vercel/build-utils`?

TooTallNate: 6d13ed54c551c20307f4610a5359e55bef59e8e0

styfle (approved):

PR #8247, opened 2022-07-25
[tests] Update node tests to run concurrently

styfle:
This applies the same chunking algorithm from `@vercel/static-build` tests to `@vercel/node` tests that we can run up to 5 integration tests concurrently.


packages/node/test/integration-setup.js
```diff
@@ -0,0 +1,72 @@
...
+  let fixtures = fs.readdirSync(fixturesPath);
+
+  if (typeof groupIndex !== 'undefined') {
+    fixtures = intoChunks(NUMBER_OF_CHUNKS, fixtures)[groupIndex - 1];
+
+    console.log('testing group', groupIndex, fixtures);
+  }
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const fixture of fixtures) {
```
EndangeredMassa, line 45: It would be neat if we reversed the order so that tests in each chunk counted up with the chunks themselves.

EndangeredMassa (approved):

packages/node/test/integration-setup.js
styfle: @ijjk Do you know why the tests are backwards? I'm pretty sure its happening before this line since the tests fixtures used to be in order so I suspect the chunking is backwards.

ijjk: The list tests currently relies on the order returned from `jest` here https://github.com/vercel/vercel/blob/afd303b94a3201c9bb6e11fe7d2c6f5825b4cb47/utils/chunk-tests.js#L17 

We could sort the tests alphabetically in jest if desired x-ref: https://jestjs.io/docs/configuration#testsequencer-string

PR #8251, opened 2022-07-26
[cli] WIP Refactor to ESM

TooTallNate:


packages/cli/scripts/build.ts
```diff
@@ -1,8 +1,12 @@
 import cpy from 'cpy';
 import execa from 'execa';
+import fs from 'fs-extra';
 import { join } from 'path';
-import { remove, writeFile } from 'fs-extra';
+import { fileURLToPath } from 'url';
```
cb1kenobi, line 5: Do we want to adopt `node:` for all Node.js internal modules? That seems to be the futureproof pattern.

`node:` was introduced in Node 12 and since we support Node >=14, we should be able to safely switch. We would need to update `@types/node`, which you've bumped it already to v14.

packages/cli/src/commands/init/init.ts
```diff
@@ -138,10 +137,14 @@ async function extractExample(
...
         const extractor = tar.extract(folder);
-        res.body.on('error', reject);
         extractor.on('error', reject);
         extractor.on('finish', resolve);
-        res.body.pipe(extractor);
+        if (res.body) {
+          res.body.on('error', reject);
+          res.body.pipe(extractor);
+        } else {
+          reject(new Error(`res.body not defiled`));
```
cb1kenobi, line 146: `s/defiled/defined/`?

packages/cli/src/util/config/files.ts
```diff
@@ -1,16 +1,16 @@
 import { join, basename } from 'path';
 import loadJSON from 'load-json-file';
-import writeJSON from 'write-json-file';
+import { writeJsonFileSync } from 'write-json-file';
```
cb1kenobi, line 3: Since this is the only file in `cli` that uses `write-json-file`, perhaps we can remove `write-json-file` in favor of `fs-extra`'s `fs.outputJson()`.

packages/cli/src/util/dev/server.ts
```diff
@@ -44,36 +44,39 @@ import {
...
-import { MissingDotenvVarsError } from '../errors-ts';
-import cliPkg from '../pkg';
-import { getVercelDirectory } from '../projects/link';
-import { staticFiles as getFiles } from '../get-files';
-import { validateConfig } from './validate';
-import { devRouter, getRoutesTypes } from './router';
-import getMimeType from './mime-type';
-import { executeBuild, getBuildMatches, shutdownBuilder } from './builder';
-import { generateErrorMessage, generateHttpStatusDescription } from './errors';
+import frameworks from '@vercel/frameworks';
```
cb1kenobi, line 47: I've seen elsewhere you do `import { frameworks } from '@vercel/frameworks';`, so for the sake of consistency consider using the same pattern. Plus, you can do `frameworks.map()` below instead of `frameworks.default.map()`.

packages/cli/tsconfig.json
```diff
@@ -3,19 +3,21 @@
     "strict": true,
     "noEmitOnError": true,
     "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "node",
-    "module": "commonjs",
+    "moduleResolution": "Node16",
```
cb1kenobi, line 6: If this works, then cool, though I believe you can leave this as `"node"`.

packages/cli/tsconfig.json
```diff
@@ -3,19 +3,21 @@
...
+    "module": "Node16",
```
cb1kenobi, line 7: FWIW, in my projects, I set this to `"esnext"`.

PR #8255, opened 2022-07-26
[static-build] Add .vercel to static-build ignore list regardless of config

Ethan-Arrowood:
### Related Issues

Adds `.vercel` path to the ignore list regardless if `zeroConfig` is enabled. This fixes a bug where the `.vercel` folder was being copied into the resulting `.vercel/output/static` directory after running `vc build` with `distDir: "."` configured for static-build.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/test/helpers/prepare.js
+++ b/packages/cli/test/helpers/prepare.js
@@ -503,7 +503,7 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
         },
       }),
     },
-    'zero-config-dist-dir': {
+    'static-build-dist-dir': {
       '.vercel/project.json': JSON.stringify({
         orgId: '.',
         projectId: '.',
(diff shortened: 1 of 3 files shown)


packages/static-build/src/index.ts
```diff
@@ -707,13 +707,12 @@ export const build: BuildV2 = async ({
           routes.push(...frameworkRoutes);
         }
 
-        let ignore: string[] = [];
+        let ignore: string[] = ['.vercel/**'];
         if (config.zeroConfig && config.outputDirectory === '.') {
           ignore = [
             '.env',
             '.env.*',
             '.git/**',
```
styfle: You also need to add it back here or else it will be overridden

packages/cli/test/helpers/prepare.js
```diff
@@ -503,6 +503,31 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
...
+          framework: null,
+        },
+      }),
+      'vercel.json': JSON.stringify({
+        version: 2,
+        builds: [
+          {
+            src: 'package.json',
+            use: '@vercel/static-build',
+            config: { distDir: '.' },
```
styfle, line 520: Let's also add a test with `zeroConfig: true` to make sure both work

styfle (approved):

packages/static-build/src/index.ts
```diff
@@ -707,7 +707,7 @@ export const build: BuildV2 = async ({
           routes.push(...frameworkRoutes);
         }
 
-        let ignore: string[] = [];
+        let ignore: string[] = ['.vercel/**'];
         if (config.zeroConfig && config.outputDirectory === '.') {
```
styfle: There should be a second case with `else if (!config.zeroConfig && config.distDir === '.')`

styfle: Or we could use a single 

`if (config.outputDirectory === '.' || config.distDir === '.')`

EndangeredMassa (approved):

styfle (approved):

PR #8227, opened 2022-07-22
Test with vc build

styfle:
Test

PR #8257, opened 2022-07-26
[tests] Update tests to run with vercel cli tarball

styfle:
This PR update the tests suite to wait for Vercel CLI tarball and then use that tarball to run E2E tests.

This is valuable because it will package all the packages in this monorepo to make the tests follow more closely what will happen in production once merged.

Since the current "Find Changes" step takes about 2 minutes, we run that first (it happens concurrently with the tarball preparation). Then once we complete "Find Changes" we wait for the tarball but it will likely be ready by that point since it also takes about 2 minutes. After both of those steps, the E2E tests continue as usual but with the `VERCEL_CLI_VERSION` set to the tarball.

- Related to #7967 
- Closes #8245 
- Closes #8227


.github/workflows/test.yml
```diff
@@ -32,6 +33,12 @@ jobs:
           echo "Files to test:"
           echo "$TESTS_ARRAY"
           echo "::set-output name=tests::$TESTS_ARRAY"
+      - uses: patrickedqvist/wait-for-vercel-preview@ae34b392ef30297f2b672f9afb3c329bde9bd487
```
ijjk, line 36: Does the `on: deployment_status` workflow trigger not work here? That seems to be how front's playwright setup is triggered.

styfle: Good point, that would work. However, if we changed to `on: deployment_status` we would have to block for 2 minutes while the deployment is going and wait for the status to be available to GitHub before we could start running turbo to check for the changes. Since both of these take about 2 minutes, it would increase the time before tests beginning running to 4 minutes. However, the way its written in my PR, it will remain 2 minutes.

ijjk: Makes sense, sounds good to kick this one off eagerly instead of waiting for the deployment to start.

packages/node/test/integration-setup.js
```diff
@@ -1,30 +1,15 @@
...
-  beforeEach(async () => {
-    if (builderUrlLastUpdated < Date.now() - ms('30min')) {
-      const builderPath = path.resolve(__dirname, '..');
-      builderUrl = await packAndDeploy(builderPath);
-      builderUrlLastUpdated = Date.now();
-      console.log('builderUrl', builderUrl);
-    }
-  });
-
+  // TODO: wait for vercel cli to create a tarball
```
Ethan-Arrowood: Is this comment still necessary?

Ethan-Arrowood (approved):
lgtm + did you mean to commit the yarn.lock in the puppeteer test?

packages/node/test/integration-setup.js
styfle: Nice catch, removed in b7575a6dc

EndangeredMassa (commented):
Looking good, but the `35-puppeteer` test is still failing

cb1kenobi (approved):

ijjk (approved):

PR #8262, opened 2022-07-27
[node] Consider "browser" and "module" fields in `package.json` when tracing Edge Functions

TooTallNate:
For Edge Functions, we are now considering the `"browser"` and `"module"` fields of packages to find the proper entrypoint. This PR updates `@vercel/node` to consider those same fields when using the node-file-trace tracing, so that the proper files get included in the output Edge Function.


packages/node/src/index.ts
```diff
@@ -200,23 +201,48 @@ async function compile(
...
-            entry = new FileFsRef({ fsPath, mode });
-          } else {
+
+          if (!entry) {
             entry = new FileBlob({ data: source, mode });
           }
           fsCache.set(relPath, entry);
           sourceCache.set(relPath, source);
-          return source.toString();
+          return source;
```
styfle: This can be a buffer according to line 205 so I think you need to keep the tostring()

styfle (commented):
Looks good, just need a test

EndangeredMassa:
Added some tests and they seemed to all be passing

packages/node/src/index.ts
```diff
@@ -201,18 +202,43 @@ async function compile(
...
+            for (const prop of ['browser', 'module']) {
+              const val = pkgJson[prop];
+              if (typeof val === 'string') {
+                debug(`Using "${prop}" field in ${fsPath}`);
+                pkgJson.main = val;
+
+                // Create the `entry` with the original so that the output is unmodified
+                if (!entry) {
+                  entry = new FileBlob({ data: source, mode });
+                }
```
EndangeredMassa, line 225: Do we need to create the `entry` here? It seems like it would be created below on line 241 regardless of what happens here.

EndangeredMassa (approved):

packages/node/src/index.ts
styfle: Agreed! In fact, I think you can move the `if` at line 213 to an `else if` following line 239

TooTallNate: `source` gets mutated on the next line though, so we need to add the entry here so that it contains the original `package.json` contents.

Ethan-Arrowood: agreed

styfle (approved):

PR #8267, opened 2022-07-28
[dev] log middleware errors in vc dev

EndangeredMassa:
Middleware server setup wasn't logging errors the same way that dev server setup was. This meant that middleware instantiation errors (like invalid config) would cause requests to 500, but no errors to be logged to the console.

This PR updates the invalid config error, makes sure errors in this area are logged out, and adds a test for this behavior.

**It may be appropriate to fail the deploy (and crash `vc dev`) in this case instead, though. What do others think?**

---

During `vc dev` with middleware that has an invalid `config.matcher` value...

Before: You see a 500 response in the browser and no output in the terminal.

After: You see a 500 response in the browser and this output in the terminal:

```
Error! Middleware's `config.matcher` values must start with "/". Received: not-a-valid-matcher
```

---

Related Issue: https://github.com/vercel/edge-functions/issues/220


packages/cli/test/dev/integration-4.test.ts
```diff
@@ -503,13 +503,6 @@ test(
   })
 );
 
-test(
-  '[vercel dev] Middleware with error at init',
-  testFixtureStdio('middleware-error-at-init', async (testPath: any) => {
-    await testPath(500, '/', /EDGE_FUNCTION_INVOCATION_FAILED/);
-  })
-);
```
EndangeredMassa, line 511: I'm not sure how this test was passing before. Will look into it more.

packages/cli/test/dev/utils.js
```diff
@@ -430,7 +433,7 @@ function testFixtureStdio(
           );
         }
 
-        if (stderr.includes('Command failed') || stderr.includes('Error!')) {
+        if (stderr.includes('Command failed')) {
```
EndangeredMassa, line 436: This `[vercel dev] Middleware with error at init` test relies on the `vc dev` not having `Error!` in the `stderr` output, which it now does.

`vc dev` and real deployments do still have the same behavior, but `testFixtureStdio` is being a bit aggressive in what counts as a failure. I removed this check for `Error!` and it now passes.

I wanted to make sure this check wasn't being relied on for other tests. So, I ran a full set of tests on just removing this check, which passed: https://github.com/vercel/vercel/pull/8293

If we really should keep this, we'll have to change the `[vercel dev] Middleware with error at init` test.

packages/node/src/dev-server.ts
```diff
@@ -224,7 +224,7 @@ async function compileUserCode(
   } catch (error) {
     // We can't easily show a meaningful stack trace from ncc -> edge-runtime.
     // So, stick with just the message for now.
-    console.error(`Failed to instantiate edge runtime.`);
+    console.error(`Failed to compile user code for edge runtime.`);
```
EndangeredMassa, line 227: I changed this slightly to make it obvious when this is the error vs. the other location of exactly the same log line where we actually instantiate the edge runtime.

styfle (approved):

PR #8273, opened 2022-07-29
[cli] Fix `vc build` with legacy `@now/static`

styfle:
There was a special case for `@vercel/static` but vc build was missing a special case for the legacy `@now/static`, which should work the same way.


EndangeredMassa (commented):
👍

PR #8274, opened 2022-07-29
[tests] Run dev e2e against same cli version

styfle:
The dev integration tests compare `vc dev` with a real deployment to make sure the results are the same. This PR ensures the deployment uses the same version of Vercel CLI as the local `vc dev` instance.


packages/cli/test/dev/utils.js
```diff
@@ -347,6 +347,9 @@ function testFixtureStdio(
               ? ['--scope', process.env.VERCEL_TEAM_ID]
               : []),
             'deploy',
+            ...(process.env.VERCEL_CLI_VERSION
+              ? ['--build-env', process.env.VERCEL_CLI_VERSION]
```
EndangeredMassa:
```suggestion
              ? ['--build-env', `VERCEL_CLI_VERSION=${process.env.VERCEL_CLI_VERSION}`]
```

Should this be set like this?

packages/cli/test/dev/integration-4.test.ts
```diff
@@ -380,18 +380,22 @@ test(
...
+
+//     await testPath(200, `/api`, 'This is the index page');
+//     await testPath(200, `/api/index`, 'This is the index page');
+//     await testPath(200, `/api/index.go`, 'This is the index page');
+//     await testPath(200, `/api/another`, 'This is another page');
+//     await testPath(200, '/api/another.go', 'This is another page');
+//     await testPath(200, `/api/foo`, 'Req Path: /api/foo');
+//     await testPath(200, `/api/bar`, 'Req Path: /api/bar');
+//   })
+// );
```
EndangeredMassa, line 398: Temporarily disabling this test. It exposed the same issue we saw in https://github.com/vercel/vercel/pull/7809

PR #8276, opened 2022-07-29
[all] Revert ncc back to 0.24.0

styfle:
- Reverts https://github.com/vercel/api/pull/13466


baruchadi (approved):

TooTallNate (approved):

EndangeredMassa:
`[vercel dev] does not display directory listing after 404` failed because the vc-build builder is being used, but we can merge with that failure for now as we're fixing issues

PR #8278, opened 2022-07-29
[next] Fix _next/data resolving priority for dynamic routes

ijjk:
### Related Issues

This ensures we include all dynamic routes as data routes when we are resolving data routes with middleware. Without this we can incorrectly resolve a more generic dynamic route that has a data route before matching a more specific dynamic route that doesn't have a data route (e.g. an automatic static optimized page). 

Fixes: [slack thread](https://vercel.slack.com/archives/CGU8HUTUH/p1659120663980089)

#### Tests

#### Code Review


packages/next/test/fixtures/00-middleware/vercel.json
```diff
@@ -1,7 +1,42 @@
 {
   "version": 2,
-  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@vercel/next@https://files-7mzyyh21l-ijjk-testing.vercel.app"
```
ijjk:
```suggestion
      "use": "@vercel/next"
```

packages/next/src/server-build.ts
```diff
@@ -1363,7 +1380,30 @@ export async function serverBuild({
...
       // /_next/data routes for getServerProps/getStaticProps pages
-      ...dataRoutes,
+      ...(isNextDataServerResolving
+        ? // when resolving data routes for middleware we need to include
+          // all dynamic routes including non-SSG/SSP so that the priority
+          // is correct
+          dynamicRoutes
+            .map(route => {
+              route = Object.assign({}, route);
+              route.src = path.join(
```
styfle: Since these are routes, I wonder if it should use `path.posix.join()` instead so this always uses `/` instead of `\` on windows

packages/next/test/fixtures/00-middleware/vercel.json
```diff
@@ -1,7 +1,42 @@
 {
   "version": 2,
-  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@vercel/next"
+    }
+  ],
```
styfle, line 8: You can remove `builds` now 😄

ijjk: This is still needed for the integration tests that run the builder manually I think since it uses the `runBuildLambda` util.

packages/next/src/server-build.ts
ijjk: Updated in https://github.com/vercel/vercel/pull/8278/commits/9e084beb1f9bfe6456a4ad83aa758f0c8c91f87b seems there are other cases we are using this so can update those in a follow-up PR.

packages/next/src/server-build.ts
```diff
@@ -1363,7 +1385,30 @@ export async function serverBuild({
...
+              route = Object.assign({}, route);
+              route.src = path.posix.join(
+                '^/',
+                entryDirectory,
+                '_next/data/',
+                escapedBuildId,
+                route.src.replace(/(^\^|\$$)/g, '') + '.json$'
+              );
+
+              const { pathname } = new URL(route.dest || '/', 'http://n');
```
styfle, line 1403: Does this need to consider both absolute and relative urls for `dest`? For example, what if `dest="https://example.com/foo"`?

packages/next/src/server-build.ts
```diff
@@ -1395,6 +1440,7 @@ export async function serverBuild({
                 'x-nextjs-matched-path': '/$1',
               },
               continue: true,
+              override: true,
```
styfle, line 1443: Should the override be gated behind `NON_NESTED_MIDDLEWARE_VERSION`?

ijjk: This is already gated behind `isNextDataServerResolving` and should be applied for the same range as that

packages/next/src/server-build.ts
ijjk: These should only be relative paths as they need to match the related build output for a dynamic route

styfle (approved):

PR #8287, opened 2022-08-01
[go] remove `meta.isDev` references and restore partial tests

EndangeredMassa:
Remove `meta.isDev` checks inside the `build` function because it's never set there. Instead, `startDevServer` would be used.

Also restored the Go tests in a partial form. Will fix the Go builder issues and make sure those features are tested completely in follow-up PRs.


packages/cli/test/dev/fixtures/go/api/[segment].go
```diff
@@ -1,10 +0,0 @@
-package handler
-
-import (
-	"fmt"
-	"net/http"
-)
-
-func Handler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "Req Path: %s", r.URL.Path)
-}
```
Comment (styfle) [inline packages/cli/test/dev/fixtures/go/api/[segment].go:10]: Should this file be restored, or else the test won't be testing the correct behavior?

TooTallNate (approved):

packages/go/index.ts
```diff
@@ -100,7 +94,7 @@ export async function build({
   workPath,
   meta = {},
 }: BuildOptions) {
-  if (process.env.GIT_CREDENTIALS && !meta.isDev) {
+  if (process.env.GIT_CREDENTIALS) {
```
styfle, line 97: Are you sure this should happen in `vc dev`?

packages/cli/test/dev/fixtures/go/api/[segment].go
Comment (EndangeredMassa) [reply packages/cli/test/dev/fixtures/go/api/[segment].go:10]: The correct behavior doesn't currently work. It will be restored in a follow-up PR.

Comment (styfle) [reply packages/cli/test/dev/fixtures/go/api/[segment].go:10]: Lets comment it out instead? Seems pretty critical to this test since the fixture is meant to test path segments like this

packages/go/index.ts
EndangeredMassa: It doesn't look like `vc dev` calls `build`. In `startDevServer`, there is no mention of `initPrivateGit`. I believe this is what already happens because `isDev` is never true when `build` is called, but if I'm mistaken we can dig deeper into this.

packages/cli/test/dev/fixtures/go/api/[segment].go
Comment (EndangeredMassa) [reply packages/cli/test/dev/fixtures/go/api/[segment].go:10]: I can't comment the contents because the file name is what causes the problem. I could rename the file, but I don't think it's that much different. I'm going to restore this file as soon as possible.

packages/go/index.ts
styfle: Hmm, might be left over from before startDevServer was implemented 👍

styfle (approved):

PR #8288, opened 2022-08-01
[build-utils] Fix symlink on download

styfle:
Some builders, such as `@vercel/next`, return both the symlinked directory and the resolved file.

When `vc build` iterates over the files to recreate them in `.vercel/output`, it fails with `EEXIST: file already exists` when creating the symlink because it first creates the file `node_modules/<symlink>/package.json` and then attempts to create the symlink `node_modules/<symlink>`.

This happened to work before `vc build` because yazl would accept the symlinked directory instead of the package.json file, so this PR is created to match that behavior.

### DIFF
--- a/packages/build-utils/src/fs/download.ts
+++ b/packages/build-utils/src/fs/download.ts
@@ -47,22 +47,22 @@ async function downloadFile(file: File, fsPath: string): Promise<FileFsRef> {
 
   if (isSymbolicLink(mode)) {
     const target = await prepareSymlinkTarget(file, fsPath);
-    console.log('Found symlink', { target, fsPath });
 
     try {
       await symlink(target, fsPath);
     } catch (error: any) {
       if (error?.code === 'EEXIST') {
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/fs/download.ts
```diff
@@ -92,12 +93,28 @@ export default async function download(
...
         return;
       }
 
+      // Some builders resolve symlinks and return both
+      // a file, node_modules/<symlink>/package.json, and
+      // node_modules/<symlink>, a symlink.
+      // Removing the file matches how the yazl lambda zip
+      // behaves so we can use download() with `vercel build`.
+      const parts = name.split('/');
+      for (let i = 0; i < parts.length; i++) {
```
TooTallNate:
```suggestion
      for (let i = 1; i < parts.length; i++) {
```

Minor optimization: I think we want to start at index 1 otherwise we'll always check an empty string entry.

TooTallNate (approved):

PR #8290, opened 2022-08-01
[cli] MAJOR: Connect a Git provider repository in `vc link`

MatthewStanciu:
#8100 added a new `vc git` command, which allows users to connect a Git provider repository, enabling them to set up a full Git workflow for their Vercel projects without having to leave the CLI.

This PR takes this functionality a step further by including it as part of the `vc link` flow. This way, users can set up a Vercel project and add a Git provider repository all in one step.

This PR is blocked by a PR to `front` which adds an option to the Git Settings page for a Project to re-enable the prompt if the user opted out (in review).

#### Tests

#### Code Review


packages/cli/src/util/link/handle-git-connection.ts
```diff
@@ -0,0 +1,271 @@
...
+} from '../git/connect-git-provider';
+import { Output } from '../output';
+import list from '../input/list';
+import Client from '../client';
+import { Org, Project, ProjectSettings } from '../../types';
+import { getCommandName } from '../pkg-name';
+import updateProject from '../projects/update-project';
+import chalk from 'chalk';
+
+function getProjectSettings(project: Project): ProjectSettings {
```
TooTallNate: Is this function necessary? Couldn't you just do `settings || project` below instead?

packages/cli/src/util/link/handle-git-connection.ts
```diff
@@ -0,0 +1,271 @@
...
+  );
+}
+
+export async function promptGitConnectSingleUrl(
+  client: Client,
+  replace = false
+) {
+  return await list(client, {
+    message: replace
+      ? 'Do you want to replace it?'
```
TooTallNate: What is "it" in this context? Can we print something more specific?

MatthewStanciu: This message appears right after another message that gives more context. What do you think of this?

Screen Shot 2022-08-04 at 8 51 27 AM

May want to move the code around a little bit to make it more clear that this is what's happening.

MatthewStanciu: Moved the logic that prints those messages next to each other so that it's clearer what's going on

packages/cli/src/util/link/handle-git-connection.ts
MatthewStanciu: I removed the function, but brought it back briefly. `settings || project` doesn't work because the API expects a list of project settings, so additional `Project` properties like `accountId` make the request fail. I initially created this function because I needed a way to strip the extra properties from `project` to turn it into the type `ProjectSettings`.

I'm sure there's a better way to do this, though.

packages/cli/src/util/git/connect-git-provider.ts
```diff
@@ -43,22 +49,22 @@ export async function connectGitProvider(
...
-        client.output.error(
-          err.message.replace(repo, chalk.cyan(repo)) +
-            `\nVisit ${link(err.link)} for more information.`
-        );
-      }
+    const apiError = isAPIError(err);
+    if (
+      apiError &&
+      (err.meta?.action === 'Install GitHub App' ||
+        err.code === 'repo_not_found')
```
EndangeredMassa: **question (soft-blocking):**  Should this check `err.action` as well?

It was done here, but not sure if it was necessary. https://github.com/vercel/vercel/pull/8351/files#diff-d7690ab6ccbb22582fe76743d64bbd2909e179a7930c35cd52f76d3128a81554R50

packages/cli/test/mocks/project.ts
```diff
@@ -125,13 +125,78 @@ export const defaultProject = {
   ],
 };
 
+export function useUnknownProject() {
+  let project: any; // TODO: properly type
```
EndangeredMassa: **suggestion (soft-blocking):** add the type

In this or a follow-up PR, can we pull in the right type here?

packages/cli/test/mocks/project.ts
```diff
@@ -125,13 +125,78 @@ export const defaultProject = {
   ],
 };
 
+export function useUnknownProject() {
```
EndangeredMassa, line 139: **suggestion (non-blocking):** add comment explaining that ordering matters

That way, hovering over the function in an IDE will show this message.

```suggestion
/**
 * Responds to any GET for a project with a 404.
 * `useUnknownProject` should always come after `useProject`, if any,
 * to allow `useProject` responses to still happen.
 */ 
export function useUnknownProject() {
```

EndangeredMassa: We could save a flag that `useUnknownProject` was called, then check for it in `useProject` and throw an error. I can't think of a reason I'd want to call `useProject` after and have that mean something.

If we did this, we'd have to remember to clear that flag between each test run in a global `afterEach`, or somewhere.

packages/cli/src/util/projects/update-project.ts
```diff
@@ -17,7 +17,7 @@ export default async function updateProject(
     `/v2/projects/${encodeURIComponent(prjNameOrId)}`,
     {
       method: 'PATCH',
-      body: JSON.stringify(settings),
+      body: settings as JSONObject, // TODO: figure out how to not make this cast necessary
```
EndangeredMassa: **suggestion (non-blocking):** move casting to the top of the function body

This conflicts because `ProjectSettings` doesn't (and shouldn't) have the index signature (`[key: Type]: Type;`). Technically, values of type `ProjectSettings` will be compatible with values of type `JSONObject`, but I don't know how to tell TypeScript that.

If we leave the casting, I'd pull it to the top of the function with a comment. I think it makes the logic easier to follow if things like this are called out separately. I suggest something like this:

```
// `ProjectSettings` is technically compatible with JSONObject
const body = settings as JSONObject;
```

packages/cli/src/util/link/handle-git-connection.ts
```diff
@@ -0,0 +1,127 @@
...
+    buildCommand: project.buildCommand,
+    outputDirectory: project.outputDirectory,
+    rootDirectory: project.rootDirectory,
+    directoryListing: project.directoryListing,
+    nodeVersion: project.nodeVersion,
+    skipGitConnectDuringLink: project.skipGitConnectDuringLink,
+  };
+}
+
+export async function handleGitConnection(
```
EndangeredMassa: **suggestion (non-blocking):** rename `handleGitConnection` to something like `addGitConnection`

I think that `handle*` can make sense when talking about handling requests, but here it seems pretty generic. What do you think about a more specific name?

packages/cli/src/util/link/handle-git-connection.ts
```diff
@@ -0,0 +1,127 @@
...
+) {
+  const remoteUrl = Object.values(remoteUrls)[0];
+  const parsedUrl = parseRepoUrl(remoteUrl);
+  if (!parsedUrl) {
+    output.debug(`Could not parse repo url ${parsedUrl}.`);
+    return 1;
+  }
+  const { org: parsedOrg, repo, provider } = parsedUrl;
+  const alreadyLinked =
+    project.link?.org === parsedOrg &&
```
EndangeredMassa: **preference (non-blocking):** use `project.link && ...`, as below

Only using `?.` on one check here seems correct, but visually threw me for a second. I think it's fine either way, but using the other pattern (checking for `project.link && ...` a few lines below makes this seem even more like it does something else, on first read.

Can we use the same pattern in both places? I prefer the second, but either is good.

packages/cli/src/util/link/git-connect-prompts.ts
```diff
@@ -0,0 +1,83 @@
...
+import { formatProvider } from '../git/connect-git-provider';
+import list from '../input/list';
+import { Output } from '../output';
+
+export async function promptGitConnectSingleUrl(
+  client: Client,
+  output: Output,
+  project: Project,
+  remoteUrl: string,
+  replace = false
```
EndangeredMassa: **suggestion (soft-blocking):** rename `replace` to `hasDiffConnectedProvider`

In this or a follow-up PR, can we change the name `replace` to something that better represents the value it holds OR better represents how it should affect the flow?

Better represents value: `hasDiffConnectedProvider`

Better represents affected flow: `shouldPromptToReplaceConnectedProvider`

packages/cli/src/util/link/git-connect-prompts.ts
```diff
@@ -0,0 +1,83 @@
...
+      `Found Git remote url ${chalk.cyan(
+        remoteUrl
+      )}, which is different from the connected ${formatProvider(
+        currentProvider
+      )} repository ${chalk.cyan(currentRepoPath)}.`
+    );
+  } else {
+    output.log(`Found local Git remote URL: ${chalk.cyan(remoteUrl)}`);
+  }
+  return await list(client, {
```
cb1kenobi, line 32: Could be just me, but after accepting the project settings, the command does work and then a few seconds later it prints out 8 lines for the Git remote URL prompt.

For me, it seems like a bit too much to process: the bold text, the `>`, the `●`, the `○`, the colors. What do you think of inserting a line break before prompting?

image

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -128,6 +130,20 @@ export default async function setupAndLink(
   } else {
     const project = projectOrNewProjectName;
 
+    const remoteUrls = await getRemoteUrls(join(path, '.git/config'), output);
+    if (remoteUrls && !project.skipGitConnectDuringLink) {
+      const connectGit = await handleGitConnection(
+        client,
+        org,
+        output,
```
cb1kenobi: Since `output` and `client.output` are the same, and the code is already passing `client` around, maybe remove `output` and pull `output` out of `client` when needed?

packages/cli/src/util/link/handle-git-connection.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export async function handleGitConnection(
+  client: Client,
+  org: Org,
+  output: Output,
+  project: Project,
+  remoteUrls: Dictionary<string>,
+  settings?: ProjectSettings
+): Promise<number | void> {
```
cb1kenobi: Out of curiosity, why does this return a number? I know `connectGitProvider()` is where the number originates on error, but it's not clear what `/v4/projects/{id}/link` returns or why an `Error` isn't thrown instead with a message and code.

packages/cli/src/util/projects/update-project.ts
```diff
@@ -1,5 +1,5 @@
 import Client from '../client';
-import { ProjectSettings } from '../../types';
+import { JSONObject, ProjectSettings } from '../../types';
```
cb1kenobi: Do we ever use `import type ...` if we're only importing types?

TooTallNate: Ideally yes, but we don't have a lint rule set up for that yet (we should use `@vercel/style-guide` 😬).

EndangeredMassa (approved):
I like having more functions for following the logic. Most of them require a lot of parameters, which is a flag that there may be a better structure for them. We could dig deeper into that to see if there is a better structure, if you were interested. For this PR, I think it's good.

I flipped the labels around so this wouldn't automerge with a single approval.

packages/cli/src/util/link/git-connect-prompts.ts
EndangeredMassa: I had the same experience. I think a line break somewhere here would be useful. If it goes before the line "Found local Git remote", it keeps that line in the context of the prompt. It may be clearer to have the origin name in these lines.

It could look like:

```
Auto-detected Project Settings (Hexo):
- Build Command: hexo generate
- Development Command: hexo server --port $PORT
- Install Command: `yarn install`, `pnpm install`, or `npm install`
- Output Directory: public
? Want to modify these settings? [y/N] n

> Found local Git remote "origin": https://github.com/EndangeredMassa/boa-v3-hexo.git
> Do you want to connect "origin" to your Vercel project? yes
> Connecting...
> Connected GitHub repository EndangeredMassa/boa-v3-hexo!
✅  Linked to endangeredmassa/test-link-2 (created .vercel)
```

packages/cli/src/util/link/handle-git-connection.ts
EndangeredMassa: There is a convention of returning numbers up a call stack to use as an exit code, but I agree that it's probably not one we want to perpetuate. There have been discussions before about wanting to refactor this convention away. I think we can refine this in a different PR.

cb1kenobi: Related/unrelated, is there a reason we haven't upgraded to v9 of the api and secondly are there api docs for v4?

packages/cli/src/util/link/handle-git-connection.ts
cb1kenobi: FWIW, more and more recently, I've been doing `?.` to short circuit as this code is written. The only times I am more explicit is when the logic is a bit more complex (or's) or when TypeScript complains.

MatthewStanciu:
sorry about that ^ blindly ran find-and-replace to change the name of my variable, and didn't check to see if that variable name was also present in other places in the code 🤦

packages/cli/src/util/link/git-connect-prompts.ts
MatthewStanciu: I love the suggestion. I think I just addressed this.

packages/cli/src/util/link/setup-and-link.ts
MatthewStanciu: Thanks, great suggestion.

packages/cli/src/util/projects/update-project.ts
MatthewStanciu: What Nate said, but I just changed it to import type.

packages/cli/src/util/git/connect-git-provider.ts
MatthewStanciu: I still can't figure out why this is the case, but it appears `err.action` is the way now. `err.meta` is gone. Just updated to use `err.action`.

packages/cli/src/util/link/handle-git-connection.ts
MatthewStanciu: I think `addGitConnection()` is perfect! Thanks.

MatthewStanciu: What do you think about the functions like `handleOptions()`, which parses the responses to the connection prompt and handles them accordingly?

packages/cli/src/util/link/handle-git-connection.ts
MatthewStanciu: Thanks for catching the inconsistency. I don't have a strong preference yet, so I went with the second.

packages/cli/src/util/link/handle-git-connection.ts
MatthewStanciu: I think the only reason I used `v4` was because I was looking at code in `front` that used `v4`. I just updated it to use `v9` and tested and it appears to work!

packages/cli/test/unit/commands/git.test.ts
```diff
@@ -39,6 +39,11 @@ describe('git', () => {
         await expect(client.stderr).toOutput('Found project');
         client.stdin.write('y\n');
 
+        await expect(client.stderr).toOutput(
+          'Do you want to connect "origin" to your Vercel project?'
+        );
+        client.stdin.write('n\n');
```
cb1kenobi, line 45: Missing a `\`?

cb1kenobi (changes requested):

packages/cli/test/unit/commands/git.test.ts
cb1kenobi: Nevermind! :)

cb1kenobi (approved):

packages/cli/src/util/link/handle-options.ts
```diff
@@ -0,0 +1,96 @@
...
+  option: string,
+  client: Client,
+  org: Org,
+  project: Project,
+  settings: ProjectSettings,
+  repoInfo?: RepoInfo
+) {
+  const { output } = client;
+  if (option === 'no') {
+    return skip(output);
```
EndangeredMassa: **preference (non-blocking):** separate `return` to make it clearer that no value is expected from `skip`

```suggestion
    skip(output);
    return;
```

Skip only returns `undefined`. This code works, but it implies more is going on. I suggest separating them. The same goes for `return optOut;`.

I think this all stems from the convention of functions used by CLI commands to return a number or whatever other type they would successfully return otherwise. We should discuss and possibly refactor these out, but not in this PR or overall task.

EndangeredMassa (approved):

PR #8291, opened 2022-08-01
[go] refactor away use of `downloaded` result

EndangeredMassa:
Review Notes: Turn off diff whitespace.

### Refactor

This PR refactors away the use of the result of `download` for a couple of reasons:
- Keeping `files` and `downloadedFiles` in sync with the file system (like when we rename a file that starts with a bracket) is easy to forget to do, causing bugs
- Nate says that `files` is something we've wanted to move away from using anyway
- It simplifies the code in a few places
- It was getting in the way of other fixes that need to be made

We do still call `download`, but it should be a no-op most of the time.

As a consequence of these changes, this PR also addresses:
- the builder no longer leaves build artifacts around, in many cases
- the builder can compile files that start with brackets again; routes don't seem to allow this to file to respond to a dynamic segments yet, though

### Next Steps

Upcoming PRs will resolve builder issues:

- bracket endpoints responding to dynamic segments
- exported function name conflict handling
- compilation targets should only apply to the source code build, not the analyze go utility

### Operating In-place

We also now have a cleanup step that clears out created files, created directories, and undoes file renames. This fixes an issue where multiple builds on the same directory would fail. It also cleans the user's project code when they are using `vc build`.

Ideally, we'd probably copy all of the code to a separate location, then freely do filesystem operations there. It's not clear to me if this is preferred for large projects because it would have to happen once per endpoint: 100 Go files would cause 10,000 (100 * 100) file copies (or symlinks). 

It has to copy once per endpoint because we potentially need all of the files around in case any of them are imported. If we had nft-style tracing for Go, we could copy only what we needed.

This gets more complex in the next step where the exported function names will be renamed during compilation to fix the name conflict issue.


packages/cli/test/dev/fixtures/go/api/another.go
```diff
@@ -1,10 +1,10 @@
-package handler
```
EndangeredMassa, line 1: reverting an earlier change in https://github.com/vercel/vercel/pull/8287/files to make sure this fixture still tests having multiple packages

packages/go/index.ts
```diff
@@ -94,332 +99,458 @@ export async function build({
...
   meta = {},
 }: BuildOptions) {
-  if (process.env.GIT_CREDENTIALS) {
-    debug('Initialize Git credentials...');
-    await initPrivateGit(process.env.GIT_CREDENTIALS);
-  }
+  const goPath = await getWriteableDirectory();
+  const srcPath = join(goPath, 'src', 'lambda');
+  const downloadPath = meta.skipDownload ? workPath : srcPath;
+  await download(files, downloadPath, meta);
```
EndangeredMassa, line 105: Download should mostly be a no-op because `skipDownload` is always true, according to @TooTallNate. I checked that this was the case locally and in CI.

packages/go/index.ts
```diff
@@ -94,332 +99,458 @@ export async function build({
...
 
-  if (process.env.GO111MODULE) {
-    console.log(`\nManually assigning 'GO111MODULE' is not recommended.
+  // keep track of file system actions we need to undo
+  // the keys "from" and "to" refer to what needs to be done
+  // in order to undo the action, not what the original action was
+  const undoFileActions: UndoFileAction[] = [];
+  const undoDirectoryCreation: string[] = [];
```
EndangeredMassa, line 111: This keeps a mostly clean user project space. It's not a complete cleanup where `go.mod` file changes are necessary, for example, but it's a lot better than it was.

packages/go/index.ts
```diff
@@ -94,332 +99,454 @@ export async function build({
...
+    throw error;
+  } finally {
     try {
-      await go.get();
-    } catch (err) {
-      console.log('Failed to `go get`');
-      throw err;
+      await cleanupFileSystem(undoFileActions, undoDirectoryCreation);
+    } catch (error) {
+      console.log(`Build succeeded, but cleanup failed: ${error.message}`);
```
EndangeredMassa: Do we necessarily want to log anything here? I think it will make it clearer that we messed with the project directory and failed to undo that, but maybe we don't need to show the actual error message.

packages/cli/test/dev/fixtures/go/api/another.go
```diff
@@ -1,10 +1,10 @@
...
+package another
```
EndangeredMassa, line 1: reverting an earlier change in https://github.com/vercel/vercel/pull/8287/files to make sure this fixture still tests having multiple packages

EndangeredMassa, line 1: reverting an earlier change in https://github.com/vercel/vercel/pull/8287/files to make sure this fixture still tests having multiple packages

packages/go/index.ts
styfle: In a future PR, we can probably remove download() from all the builders

packages/go/index.ts
```diff
@@ -94,332 +99,454 @@ export async function build({
...
+          isGoModInRootDir = true;
+          goModPath = join(fileDirname, '..');
+          const pathParts = file.split(sep);
+          pathParts.pop(); // Remove go.mod
+          pathParts.pop(); // Remove api
+          pathParts.push('go.mod');
+          const newFsPath = pathParts.join(sep);
+
+          debug(`Moving api/go.mod to root: ${file} to ${newFsPath}`);
+          await move(file, newFsPath);
```
styfle, line 212: Will this work now that `newFsPath` was changed from an absolute path to a relative path? 

(Pro tip: most of the builders use fspath, to mean absolute path)

packages/go/index.ts
```diff
@@ -94,332 +99,454 @@ export async function build({
...
+    // find `go.mod` in downloadedFiles
+    const entrypointDirname = dirname(entrypointAbsolute);
+    let isGoModExist = false;
+    let goModPath = '';
+    let isGoModInRootDir = false;
+
+    const modFileRefs = await glob('**/*.mod', workPath);
+    const modFiles = Object.keys(modFileRefs);
+
+    for (const file of modFiles) {
```
styfle, line 183: Is `file` a relative or absolute path? It should probably use the name `fsPath` if its absolute

packages/go/index.ts
styfle: This is only relevant for running `vercel build` locally, right? We don't have this problem when running `vercel dev`

EndangeredMassa: Correct. `vc dev` does not have this problem.

packages/go/index.ts
EndangeredMassa: No, that looks wrong. I thought this condition was covered by the tests, but because of the `zeroConfig` check, I don't believe it is. I'll make sure it's covered and resolve this issue.

packages/go/index.ts
EndangeredMassa: It's relative. I didn't realize the convention was to use `fsPath` for absolute values. I'll do that going forward!

packages/go/index.ts
EndangeredMassa: Oh, it happened to work because the relative path was correct, but to make sure this is true, I updated it to use an absolute path. [`7308609` (#8291)](https://github.com/vercel/vercel/pull/8291/commits/7308609dd9bbb19bea1eb224225adca669457d62)

Turns out fixture `17-go-mod-api` was already testing this case. So, no new test is necessary. The fact that this test still passes shows the change here still works.

packages/go/index.ts
```diff
@@ -94,332 +99,456 @@ export async function build({
...
+  Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
+        `
+      );
+      console.log(err.message);
+      throw err;
+    }
+
+    const parsedAnalyzed = JSON.parse(analyzed) as Analyzed;
+
+    // find `go.mod` in downloadedFiles
```
cb1kenobi: More like find `go.mod` in modFiles?

EndangeredMassa: I missed that in the rename. Thanks! [`0432d38` (#8291)](https://github.com/vercel/vercel/pull/8291/commits/0432d38772a69f7830ecad0c3c30459a213e6c16)

packages/go/index.ts
```diff
@@ -94,332 +99,456 @@ export async function build({
...
+          // We didn't find `/go.mod` but we found `/api/go.mod` so move it to the root
+          isGoModExist = true;
+          isGoModInRootDir = true;
+          goModPath = join(fileDirname, '..');
+          const pathParts = file.split(sep);
+          pathParts.pop(); // Remove go.mod
+          pathParts.pop(); // Remove api
+          pathParts.push('go.mod');
+
+          const newRoot = pathParts.join(sep);
```
cb1kenobi, line 208: I understand your changes don't really change the method the path was rewritten, but I've run into cross platform issues with manually splitting on `path.sep`. I generally play it safe and just do:

```javascript
const newRoot = join(dirname(dirname(file)), 'go.mod');
```
or
```javascript
const newRoot = join(file, '../../go.mod');
```
Note: `path.join()` retains the relativeness of `file` whereas `path.resolve()` will return an absolute path and in this case, we want a relative path.

packages/go/index.ts
cb1kenobi: Note that regardless if the build succeeds or fails, if the cleanup fails, it will always say the build succeeded. Could lead to confusion.

packages/go/index.ts
EndangeredMassa: Interesting! Do you recall the specific issues with splitting on `path.sep`?

I'm for this change, but I'd rather do it separately. I'll keep a note of it. Thanks!

packages/go/index.ts
EndangeredMassa: @cb1kenobi good call. Updated! [`ec46234` (#8291)](https://github.com/vercel/vercel/pull/8291/commits/ec462347b1d8d00fd0639cb3d42e4585a0814a68)

packages/go/index.ts
cb1kenobi: Silly question, but would it be better to just scan and copy the files into a tmp dir so that we don't need to touch the user's project files?

packages/go/index.ts
cb1kenobi: I believe this was more of an issue with super old versions of Node.js, but today it is possible to be on a posix machine and doing a `join('foo\\bar', 'baz')` will result in `foo\\bar/baz` and `join(process.cwd(), '\\foo\\bar')` will result in `/path/to/cwd\\foo\\bar`.

Under the hood on Windows, Node converts all `\` to `/`, so on Windows things are surprisingly robust. On posix platforms, `\` is a valid file name character, so Node leaves it alone.

In this case, I don't know if the relative paths returned by `glob()` use platform specific path separators.

I guess the moral of the story is using `path.sep` is no more dangerous than `join()` (on Mac/Linux) and probably more importantly don't mix posix and Windows paths, which the codes doesn't appear to be doing anyways.

packages/go/index.ts
EndangeredMassa: There are pros and cons, but ultimately, I think we'd still want to do something like that. Maybe it's a virtualized file system, a full project copy, or a bunch of symlinks. I touched on this in the "Operating In-place" section of the description. I'm very open to alternative approaches to these shenanigans.

cb1kenobi: I love the idea of a virtual FS! Symlinks work pretty well on posix platforms. Copying an entire project is ok if you're already copying and modifying many files anyways, otherwise maybe there's a JS implementation of rsync. In any case, my money is on the virtual filesystem.

cb1kenobi (approved):
I've combed through the code and I can't really find anything complain about. Tests run locally and this is a great improvement!

TooTallNate (approved):

PR #8299, opened 2022-08-03
[go] fix function name conflict

EndangeredMassa:
When compiling multiple Go endpoints that share a function name (and the files are not in the `go.mod` root), the builder will fail to arrange the files such that the go compiler can compile them without error. Go compile fails because the exported handler function already exists.

This PR renames all handler function names as they are compiled, then renames them back. This resolves the issue, but adds to the dirty user project code state. So, it also adds an undo system for function renames.

---

### Alternatives

Instead of renaming every entrypoint during its build, we could only rename if it was necessary. However, determining if it is necessary requires some Go project parsing that would be easy to get wrong. It would probably also be slower given the number of filesystem operations that would have to happen.

Instead of renaming the entrypoint's function name at all, we could rearrange files such that these endpoints are not in the project dir when compiled. However, this would be hard to assess and potentially not a complete solution: two Go endpoint files could define the same function name, but share a different exported function during their execution.


packages/go/index.ts
```diff
@@ -501,17 +520,54 @@ export async function build({
...
       console.log(`Build succeeded, but cleanup failed: ${error.message}`);
       debug('Cleanup Error: ' + error);
     }
   }
 }
 
+async function renameHandlerFunction(fsPath: string, from: string, to: string) {
+  let fileContents = (await readFile(fsPath)).toString('utf-8');
+
+  const fromRegex = new RegExp(`\\b${from}\\b`, 'g');
```
EndangeredMassa, line 538: Relying on word boundaries here accounts for some things, like replacing all instances of a function name, like:

```go
var newIndex = Index

func Index(...)
```

But this will also catch the same value in comments and strings. I've asked internally about a better way to approach this, likely with actual AST parsing.

@TooTallNate and I agree that this is shippable as is, but we'd still work on the more complete solution immediately.

packages/go/index.ts
```diff
@@ -501,17 +520,54 @@ export async function build({
...
+      );
     } catch (error) {
       console.log(`Build succeeded, but cleanup failed: ${error.message}`);
       debug('Cleanup Error: ' + error);
     }
   }
 }
 
+async function renameHandlerFunction(fsPath: string, from: string, to: string) {
+  let fileContents = (await readFile(fsPath)).toString('utf-8');
```
styfle:
```suggestion
  let fileContents = await readFile(fsPath, 'utf8');
```

packages/go/index.ts
```diff
@@ -501,17 +520,54 @@ export async function build({
...
+  debug(`Found exported function "${originalFunctionName}" in "${entrypoint}"`);
+
+  const pathSlug = entrypoint
+    .replace(/\\/g, '_')
+    .replace(/\//g, '_')
+    .replace(/\[/g, '_')
+    .replace(/\]/g, '_')
+    .replace(/ /g, '_')
+    .replace(/-/g, '_')
+    .replace(/\./g, '_');
```
styfle: I think you can do this all in a single regex using `/(this|that|other)/g` and we might want to add a unit test for this.

styfle:
```suggestion
  const pathSlug = entrypoint.replace(/( \\|\/|\]|\[|-|\.)/g, '_');
```

cb1kenobi: Good catch, though you forgot a `|` after the space and before the `\\`.

This could be simplified further by doing:

```javascript
const pathSlug = entrypoint.replace(/[\\/[\] -.]+/g, '_');
```

Also, you probably will want to add a `+` near the end so it replaces multiple consecutive characters with a single `_`:

```javascript
'a\\b/c[d]e f-g.h/-./\\i'.replace(/[\\/[\] -.]+/g, '_') // 'a_b_c_d_e_f_g_h_i'
'a\\b/c[d]e f-g.h/-./\\i'.replace(/[\\/[\] -.]/g, '_') // 'a_b_c_d_e_f_g_h_____i'
```

EndangeredMassa: I think we should optimize for legibility on this one. However, I don't think one is necessarily easier to read than the other. With syntax highlighting at least, using `|` a bunch is helpful.

I went with `/(\s|\\|\/|\]|\[|-|\.)/g`. I added a bunch of tests for it.

I don't think it's critical to replace multiple instances. No one is supposed to even see this value.

https://github.com/vercel/vercel/pull/8299/commits/2b4d729f88a044994d026ec0b82bde56a3b6792f

EndangeredMassa:
I added some unit-level tests to this, but the test suites weren't set up in this repo for anything but integration tests. I can follow up with updates to the test suites here to match other builders. For now, maybe we can leave the few unit tests where they are?

I could also split the integration tests in half. Right now they take ~14 minutes to run.

packages/go/test/fixtures/21-func-conflicts/vercel.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 2,
+  "builds": [{ "src": "**/*.go", "use": "@vercel/go" }],
+  "probes": [
+    { "path": "/", "mustContain": "from index.go" },
+    { "path": "/dupe-handler.go", "mustContain": "from dupe-handler.go" },
+    { "path": "/other-package.go", "mustContain": "from other-package.go" },
+    { "path": "/sub/one.go", "mustContain": "from /sub/one.go" },
```
cb1kenobi, line 8: Out of curiosity, what is this `probes` thing? The docs describe something else and this looks like a Jest thing, but it's obviously in the `vercel.json` file.

Secondly, is it just doing a string match on an absolute path? In other words, it's not actually treating `/sub/one.go` as absolute.

cb1kenobi (approved):

packages/go/test/fixtures/21-func-conflicts/vercel.json
EndangeredMassa: Probes are a way to write some simple request/response tests against a deployed fixture.

The `path` is relative. It's appended on to the deployment url.

The logic for this is in: https://github.com/vercel/vercel/blob/9c66ebc536c4acf6ca395aa24fe43100582a817f/test/lib/deployment/test-deployment.js#L34

Note that we recently made a change to allow `probes.json` to contain these instead of putting them in `now.json` or `vercel.json`, which makes it easier to deploy a fixture manually.

EndangeredMassa:
I don't have the patch label on this PR yet because I don't want it to automerge. I'll add when I'm ready to merge.

packages/go/test/fixtures/21-func-conflicts/vercel.json
styfle: @EndangeredMassa @cb1kenobi FYI, you can remove `vercel.json` for most tests now and instead use `probes.json`

See https://github.com/vercel/vercel/pull/8279

EndangeredMassa: Indeed! I'll switch that over in my follow-up test suite updates PR.

styfle (approved):

PR #8305, opened 2022-08-03
[cli] Print error from Builder in `vc build`

TooTallNate:
Ensures that errors that are serialized into `builds.json` are also printed to the terminal when running the `vc build` command.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -469,6 +469,8 @@ async function doBuild(
         )
       );
     } catch (err: any) {
+      output.prettyError(err);
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -469,6 +469,8 @@ async function doBuild(
         )
       );
     } catch (err: any) {
+      output.prettyError(err);
```
styfle, line 472: @TooTallNate Do we need this to conditionally print so we don't accidentally print twice in the build container?

styfle: Looks like we can solve this in the build container 👍

styfle (approved):

PR #8307, opened 2022-08-03
[cli] Strip scheme from `vc inspect` argument

MatthewStanciu:
Right now, `vc inspect` fails to find a deployment if you include `http://` before it. But it works with no scheme and with `https://`.

Since it appears no scheme is what the API looks for anyway, and to avoid confusion, this PR strips any included scheme from the `deploymentIdOrHost` argument.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/commands/inspect.ts
+++ b/packages/cli/src/commands/inspect.ts
@@ -66,7 +66,7 @@ export default async function main(client: Client) {
   const { print, log, error } = client.output;
 
   // extract the first parameter
-  const [, deploymentIdOrHost] = argv._;
+  let [, deploymentIdOrHost] = argv._;
 
   if (argv._.length !== 2) {
     error(`${getCommandName('inspect <url>')} expects exactly one argument`);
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/inspect.ts
```diff
@@ -212,3 +213,11 @@ function stateString(s: Deployment['readyState']) {
       return chalk.gray('UNKNOWN');
   }
 }
+
+function stripScheme(deploymentIdOrHost: string): string {
+  const match = deploymentIdOrHost.match(/http:\/\/|https:\/\//g)?.[0];
```
styfle: This should probably use `^` to match that start because a url might have a query string too.

In fact, this should instead use `new URL(deploymentIdOrHost).hostname` in a try/catch instead to handle all these cases

styfle (commented):
How about a test?

styfle (approved):
Perfect, thanks! 🥇

PR #8316, opened 2022-08-04
[cli] Fix flaky error in `builds.json` tests

TooTallNate:
Follow-up to #8305.

The `expect().toOutput()` call was frequently timing out before the command writes the error to the terminal, causing the test to fail. So wait for the command to return the exit code before running assertion on the printed output.


TooTallNate:
Test failures are not related to this PR, but unit tests are passing so merging.

PR #8318, opened 2022-08-04
[tests] Update package.json scripts

styfle:
This PR consolidates all the `test` scripts to be the same and removes the `prepublishOnly` script since we always run `build` before publishing to npm.


cb1kenobi (approved):
LGTM

PR #8320, opened 2022-08-04
[cli] Fix vc secret rm

styfle:
Fixes a regression from the the refactor in https://github.com/vercel/vercel/pull/8039 that was causing the following error:

```
Error! client.prompt is not a function
```


cb1kenobi (approved):
Verified and approved

MatthewStanciu (approved):

TooTallNate (approved):

PR #8326, opened 2022-08-04
[cli] Always set `BROWSER=none` env var for frontend dev server in `vc dev`

TooTallNate:
Previously `vc dev` would only set this env var for "create-react-app" Framework preset, but other frameworks do the "auto-open browser window" behavior as well, and respect this env var (Docusaurus, specifically).

So always set the `BROWSER=none` env var regardless of which framework preset is selected. Also simplifies the code since this was the only place `frameworkSlug` property was being used.


styfle (approved):
Thanks!

PR #8330, opened 2022-08-05
[cli] Standardize on `--yes` instead of `--confirm`

cb1kenobi:
### Related Issues

> https://linear.app/vercel/issue/VCCLI-226/rename-confirm-to-yes-and-update-help
> https://linear.app/vercel/issue/VCCLI-227/add-y-to-all-commands-that-have-yes-flag

#### Tests

#### Code Review


packages/cli/src/commands/dev/index.ts
```diff
@@ -74,11 +74,14 @@ export default async function main(client: Client) {
...
       '--listen': String,
       '-l': '--listen',
-      '--confirm': Boolean,
+      '--yes': Boolean,
+      '-y': '--yes',
 
       // Deprecated
       '--port': Number,
       '-p': '--port',
+      '-c': '--yes',
```
TooTallNate: Let's add a deprecation warning for when `--confirm` is used?

packages/cli/src/commands/domains/index.ts
```diff
@@ -45,6 +45,7 @@ const help = () => {
   )}        Login token
     -S, --scope                    Set a custom scope
     -N, --next                     Show next page of results
+    -y, --yes                      Skip questions when setting up new project using default scope and settings
```
styfle: Is there a “confirm” option for domains?

I’m not sure why “yes” was added here. Where is it used?

cb1kenobi: Good catch! The description is inaccurate in this case. It's used here: https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/domains/rm.ts#L81.

packages/cli/src/commands/dev/index.ts
cb1kenobi: I asked that exact question on the Linear card (https://linear.app/vercel/issue/VCCLI-226#comment-65e30f77) and Sean was concerned about writing a warning to `stdout` for the `deploy` command. I'd be certainly happy to add a warning to all or some of the commands. Let me know what you think.

TooTallNate: I saw that comment and I thought that I followed-up but I guess I didn't hit send. We write to `stderr` (via `client.output`) for logging messages so that shouldn't be a concern 😃

cb1kenobi: Do you mean `client.warn()` like this: https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/dev/index.ts#L86?

cb1kenobi: Fixed! :)

TooTallNate (approved):
LGTM. Can you also update the usage in `test/unit/commands/git.test.ts` (well, all the tests really).

verbose117 (approved):
👍🏻

PR #8349, opened 2022-08-08
[node] adds /index to / middleware matchers

feugy:
### 📖 What's in there?

Relates to: https://github.com/vercel/edge-functions/issues/235
When used to enforce accesses, Middleware matchers can be confusing, as reported in the ticket above.
While it is the current expected behavior, it is quite misleading, because `/index` resolves to `/` on Vercel platform (which often are 404, but still). 

We agreed on [slack](https://vercel.slack.com/archives/C03S8ED1DKM/p1659625216874639?thread_ts=1659619656.516699&cid=C03S8ED1DKM) to automatically adds `/index` when user specifies `/`

Open question: when users specify a route like `/about`, should we also add a match for `/about/index`? What about [path matching](https://nextjs.org/docs/api-reference/next.config.js/redirects#path-matching)?  

### 🧪 How to test?

It's fully covered with unit tests:

```shell
yarn workspace @vercel/node test edge-matcher.test
```
I have an integration test ready, but we'll need to deploy a new build-container before adding it.

#### Tests

#### Code Review


ijjk (approved):

styfle (approved):

packages/node/src/utils.ts
```diff
@@ -24,8 +25,11 @@ function getRegExpFromMatcher(matcher: unknown): string {
     );
   }
 
-  const re = pathToRegexp(matcher);
-  return re.source;
+  const regExps = [pathToRegexp(matcher).source];
+  if (matcher === '/') {
+    regExps.push(pathToRegexp('/index').source);
```
EndangeredMassa, line 30: **suggestion:** check for `/index` already existing before adding in

If the matchers are `['/', '/index']`, the result is that the regex for `/index` is in the resulting regex twice. I don't think this will cause any problems, but we should probably keep it unique.

feugy: A good idea Sean.
I'll open a followup PR

PR #8350, opened 2022-08-08
[go] add lambda wrapper support to the go runtime

craigandrews:
Check for `AWS_LAMBDA_EXEC_WRAPPER` when starting Go lambdas and, if present, restart the process as a child of the wrapper. This enabled the use of large environment.

Simplify the build by having only a single `main.go` file. In the case of a legacy build, completely remove the `"__VC_HANDLER_PACKAGE_NAME"` import.

https://nextjs-15smhwlwk-craigandrews.vercel.app/api/golang

Example of a Go lambda with large environment support. Built by linking the `@vercel/go` package into `.vercel/builders/node_modules` and using BOA to prebuild and deploy it.

#### Tests

#### Code Review


craigandrews:
I have created a test for large environment, but it will fail until after this change is merged due to the CLI and build container pulling the latest version from npm instead of using this version.

craigandrews:
If there is advice on how to write tests for the go builder that build against the un-merged version I would be interested in updating the tests to do that.

packages/go/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vercel/go",
-  "version": "2.0.15",
+  "version": "2.0.16",
```
EndangeredMassa: We publish as a separate step. I'll follow up with more details in the slack thread.

```suggestion
  "version": "2.0.15",
```

EndangeredMassa (approved):

PR #8351, opened 2022-08-08
[cli] Accept URL argument in `vc git connect`

MatthewStanciu:
Right now, `vc git connect` connects a Git provider repository by looking for a remote URL in the user's local `.git` config. But sometimes a user may want to explicitly state which Git provider repository they want to connect. This PR adds support for an optional URL argument.

#### Tests

#### Code Review


packages/cli/test/unit/commands/git.test.ts
```diff
@@ -426,5 +426,188 @@ describe('git', () => {
...
+          updatedAt: 1656109539791,
+        });
+
+        await expect(gitPromise).resolves.toEqual(0);
+      } finally {
+        await fs.rename(join(cwd, '.git'), join(cwd, 'git'));
+        process.chdir(originalCwd);
+      }
+    });
+    it('should continue as normal when input matches singlegit remote', async () => {
```
styfle:
```suggestion
    it('should continue as normal when input matches single git remote', async () => {
```

packages/cli/src/commands/git/index.ts
```diff
@@ -16,7 +16,7 @@ const help = () => {
 
   ${chalk.dim('Commands:')}
 
-    connect                          Connect your Git config "origin" remote as a Git provider to your project
+    connect [url]                    Connect a local Git remote or given Git provider repository URL as a Git provider to your project
```
styfle: "local git remote" sounds contradictory

```suggestion
    connect [url]                    Connect your Vercel Project your Git repository or provide the remote URL to your Git repository
```

packages/cli/src/commands/git/index.ts
```diff
@@ -28,9 +28,17 @@ const help = () => {
 
   ${chalk.dim('Examples:')}
 
-  ${chalk.gray('–')} Connect a Git provider repository
+  ${chalk.gray(
+    '–'
+  )} Connect a Git provider repository from your local Git config
```
styfle:
```suggestion
  )} Connect your Vercel Project your Git repository defined in your local .git config
```

packages/cli/src/commands/git/index.ts
```diff
@@ -28,9 +28,17 @@ const help = () => {
...
 
     ${chalk.cyan(`$ ${getPkgName()} git connect`)}
+  
+  ${chalk.gray('–')} Connect a Git provider repository
```
styfle:
```suggestion
  ${chalk.gray('–')} Connect your Vercel Project to a Git repository using the remote URL
```

packages/cli/test/unit/commands/git.test.ts
```diff
@@ -426,5 +426,188 @@ describe('git', () => {
...
+        await fs.rename(join(cwd, 'git'), join(cwd, '.git'));
+        useUser();
+        useTeams('team_dummy');
+        useProject({
+          ...defaultProject,
+          id: 'no-remote-url',
+          name: 'no-remote-url',
+        });
+
+        client.setArgv('git', 'connect', 'https://github.com/user2/repo2'); // TODO: accept urls with no scheme
```
EndangeredMassa: **suggestion (non-blocking):** remove `TODO`?

If we want to do this, can we make a card for it?

packages/cli/src/util/projects/connect-git-provider.ts
```diff
@@ -45,6 +47,7 @@ export async function connectGitProvider(
   } catch (err: unknown) {
     if (isAPIError(err)) {
       if (
+        err.action === 'Install GitHub App' ||
         err.meta?.action === 'Install GitHub App' ||
```
EndangeredMassa: **suggestion (non-blocking):** remove `err.meta?.action` check, if possible

This action can be at the top level or under `meta`? Do we still need to check for `meta`? It doesn't look like it would hurt to keep it in, but I'm wondering more about why the old check wasn't good enough.

packages/cli/src/commands/git/index.ts
```diff
@@ -16,7 +16,7 @@ const help = () => {
 
   ${chalk.dim('Commands:')}
 
-    connect                          Connect your Git config "origin" remote as a Git provider to your project
+    connect [url]                    Connect your Vercel Project your Git repository or provide the remote URL to your Git repository
```
EndangeredMassa: **suggestion (blocking):** typo

```suggestion
    connect [url]                    Connect your Vercel Project to your Git repository or provide the remote URL to your Git repository
```

packages/cli/src/commands/git/index.ts
```diff
@@ -28,9 +28,17 @@ const help = () => {
 
   ${chalk.dim('Examples:')}
 
-  ${chalk.gray('–')} Connect a Git provider repository
+  ${chalk.gray(
+    '–'
+  )} Connect your Vercel Project your Git repository defined in your local .git config
```
EndangeredMassa: **suggestion (blocking):** typo

```suggestion
  )} Connect your Vercel Project to your Git repository defined in your local .git config
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+      client,
+      `Do you still want to connect ${chalk.cyan(arg)}?`,
+      false
+    );
+    if (!shouldConnect) {
+      output.log(`Aborted. Repo not connected.`);
+    }
+  }
+  return shouldConnect;
+}
```
EndangeredMassa, line 324: **suggestion (non-blocking):** use clearer function and variable identifiers

I think the names of functions and variables around this area can be confusing. This function uses `arg` as a variable name, which I believe is the command-line provided repo url, which it needs to compare to `parsedUrl`, which I believe is the git-config provided repo url.

For example: If these were called `currentRepoUrl` (for the git config one) and `newRepoUrl`, I'd be able to tell them apart better.

I think we could make other names around this area better in the same way. I'd be happy to pair on it. It's certainly not blocking.

EndangeredMassa (changes requested):
**blocking:** I think we should fix the typos before merging, but the other comments are not blocking.

I like the use of more functions here. It made it easier for me to follow the logic. Thanks!

packages/cli/src/commands/git/connect.ts
```diff
@@ -25,8 +27,9 @@ export default async function connect(
 ) {
   const { output } = client;
   const confirm = Boolean(argv['--confirm']);
+  const repoArg = argv._[1];
 
-  if (args.length !== 0) {
+  if (args.length !== 0 && args.length !== 1) {
```
styfle:
```suggestion
  if (args.length > 1) {
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+  confirm: boolean,
+  org: Org,
+  project: Project,
+  repoArg: string
+) {
+  output.log(`Connecting Git remote: ${link(repoArg)}`);
+  const parsedRepoArg = parseRepoUrl(repoArg);
+  if (!parsedRepoArg) {
+    output.error(
+      `Cannot parse URL ${repoArg}. Make sure you didn't make any typos.`
```
styfle:
```suggestion
      `Failed to parse URL "${repoArg}". Please ensure the URL is valid.`
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+  org: Org,
+  project: Project,
+  confirm: boolean,
+  gitConfig: Dictionary<any>,
+  repoArg: string
+) {
+  const parsedUrlArg = parseRepoUrl(repoArg);
+  if (!parsedUrlArg) {
+    output.error(
+      `Cannot parse URL ${repoArg}. Make sure you didn't make any typos.`
```
styfle:
```suggestion
      `Failed to parse URL "${repoArg}". Please ensure the URL is valid.`
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+  const multiple = Object.keys(remoteUrls).length > 1;
+  if (multiple) {
+    output.log(`Found multiple Git repositories in your local Git config:`);
+    printRemoteUrls(output, remoteUrls);
+  } else {
+    const url = Object.values(remoteUrls)[0];
+    const parsedUrl = parseRepoUrl(url);
+    if (!parsedUrl) {
+      output.error(
+        `Cannot parse URL ${url}. Make sure you didn't make any typos.`
```
styfle:
```suggestion
        `Failed to parse URL "${url}". Please ensure the URL is valid.`
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+      );
+      return 1;
+    }
+    const parsedUrlArg = parseRepoUrl(arg);
+    if (!parsedUrlArg) {
+      output.error(
+        `Cannot parse URL ${arg}. Make sure you didn't make any typos.`
```
styfle:
```suggestion
        `Failed to parse URL "${arg}". Please ensure the URL is valid.`
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+      );
+      return 1;
+    }
+    if (JSON.stringify(parsedUrl) === JSON.stringify(parsedUrlArg)) {
+      return true;
```
styfle, line 301: Is there a reason this is returning `true` instead of `0`? 

The errors above return `1` so I would assume success should return `0`

MatthewStanciu: Good catch, thanks

MatthewStanciu: Oh actually, it was returning true because `promptConnectArg()` returns a boolean, so `true` means the code should continue, not exit successfully.

I updated it to only return a boolean and exit with code 1 if it returns false.

packages/cli/src/util/projects/connect-git-provider.ts
MatthewStanciu: I removed the meta check. Tested and it appears to work.

Though one confusing thing is that `err.meta` used to be the way to do it. At some point it stopped working and the metadata moved out of `meta`. 🤔

packages/cli/src/util/projects/connect-git-provider.ts
```diff
@@ -118,3 +120,14 @@ export function parseRepoUrl(originUrl: string): {
...
+
+export function printRemoteUrls(
+  output: Output,
+  remoteUrls: Dictionary<string>
+) {
+  let str = '';
+  for (const [name, url] of Object.entries(remoteUrls)) {
+    str += `  • ${name}: ${chalk.cyan(url)}\n`;
+  }
+  output.print(str);
```
cb1kenobi: Is this method of building a potentially large buffer and calling print once standard? I would assume the overhead of growing the string buffer would be greater than just calling `print()` and writing the output for each entry.

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+    );
+    if (!shouldConnect) {
+      output.log(`Aborted. Repo not connected.`);
+    }
+  }
+  return shouldConnect;
+}
+
+async function checkExistsAndConnect(
+  client: Client,
```
cb1kenobi: 10 function arguments is a lot. Consider passing in a single argument which is an object that is destructured. It's too easy to confuse the arguments, especially if another argument is added or an argument is removed.

```javascript
interface GitRepoCheckParams {
  client: Client;
  confirm: boolean;
  gitProviderLink?: ProjectLinkData;
  org: Org;
  output: Output;
  parsedOrg: string;
  project: Project;
  provider: string;
  repo: string;
  repoPath: string;
}

async function checkExistsAndConnect({
  client,
  confirm,
  gitProviderLink
  org,
  output,
  parsedOrg,
  project,
  provider,
  repoPath
}: GitRepoCheckParams) {
  // snip
```

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+async function promptConnectArg(
+  client: Client,
+  output: Output,
+  yes: boolean,
+  arg: string,
+  remoteUrls: Dictionary<string>
+) {
+  const multiple = Object.keys(remoteUrls).length > 1;
+  if (multiple) {
+    output.log(`Found multiple Git repositories in your local Git config:`);
```
cb1kenobi: Do we encourage template literals if there are no expressions? I've seen this in a few other places in the PR.

packages/cli/src/commands/git/connect.ts
```diff
@@ -101,10 +121,194 @@ export default async function connect(
...
+}
+
+async function promptConnectArg(
+  client: Client,
+  output: Output,
+  yes: boolean,
+  arg: string,
+  remoteUrls: Dictionary<string>
+) {
+  const multiple = Object.keys(remoteUrls).length > 1;
```
cb1kenobi: Nitpick, but not sure this needed to be in its own variable.

packages/cli/src/commands/git/connect.ts
```diff
@@ -57,6 +60,23 @@ export default async function connect(
...
+        client,
+        output,
+        org,
+        project,
+        confirm,
+        gitConfig,
+        repoArg
+      );
+    } else {
+      return await connectArg(client, output, confirm, org, project, repoArg);
```
cb1kenobi: Unless others have opinions, you don't need the `else` and instead keep the code to the left.

packages/cli/test/unit/commands/git.test.ts
MatthewStanciu: Done in latest commit!

packages/cli/src/commands/git/connect.ts
MatthewStanciu: I don't think there's a standard, but I agree that it's probably better to not use template literals here. It's also not consistent throughout this PR.

packages/cli/src/commands/git/connect.ts
MatthewStanciu: Thanks, did a bit of refactoring.

packages/cli/src/commands/git/connect.ts
MatthewStanciu: Thanks for helping me get started! I think it's a bit better now.,

packages/cli/src/util/git/connect-git-provider.ts
```diff
@@ -83,15 +92,14 @@ export function formatProvider(type: string): string {
...
-  org: string;
-  repo: string;
-} | null {
+export function parseRepoUrl(originUrl: string): ParsedRepoUrl | null {
   const isSSH = originUrl.startsWith('git@');
   // Matches all characters between (// or @) and (.com or .org)
   // eslint-disable-next-line prefer-named-capture-group
-  const provider = /(?<=(\/\/|@)).*(?=(\.com|\.org))/.exec(originUrl);
+  // const provider = /(?<=(\/\/|@)).*(?=(\.com|\.org))/.exec(originUrl);
+  let provider =
```
cb1kenobi: I'm not not tracking why the `const` line was commented out. Furthermore, prefer `const`?

Also, could you please add a few example origin URLs in the comment above so we can get an idea of what sort of URLs are being matched?

FWIW, I think you could write a single regex that could test the URL and extract the org and repo so you don't need to do all that splitting and replacing.

Side note, if this function was called frequently, it would be better to cache the compiled regex outside the function and prefer `String.match()` over `RegExp.exec()` since `exec()` modifies the `RegExp` instance and `match()` does not. Your code in this instance is just fine, but more of a FYI.

MatthewStanciu: Sorry, this one was me being sloppy. I initially changed it to `let` because of some other logic I was trying that involved reassigning it, but then I forgot to change it back and remove the comment above.

Re regex: there's definitely a pattern that will cover everything I want, but I couldn't figure it out :( so I decided it was better to move on for now. If you're good with regex, I would love to pair with you sometime today to figure it out.

cb1kenobi (approved):

EndangeredMassa (approved):

PR #8356, opened 2022-08-08
[client] Send single `source.tgz` archive when creating deployment

MatthewStanciu:
Right now, deployments via Vercel CLI are created by uploading every source file individually. This PR creates a tarball of the user's source files and sends the packed tarball instead. This speeds up deployment times, especially for large projects, and also bypasses the 15,000 file upload limit.

Currently, it requires a `--archive=tgz` flag. Perhaps we can enable this by default for prebuilt, but that may be for a separate PR.

#### Tests

#### Code Review


packages/client/src/upload.ts
```diff
@@ -86,17 +85,6 @@ export async function* upload(
 
         await semaphore.acquire();
 
-        const fPath = file.names[0];
```
MatthewStanciu: The file object already contains the contents as a buffer, so there's no need to read the files again.

Ethan-Arrowood (approved):
lgtm. nice work with the data streaming for efficiency 👍

packages/client/src/types.ts
```diff
@@ -11,6 +11,10 @@ export interface Dictionary<T> {
   [key: string]: T;
 }
 
+export enum ArchiveFormat {
+  Tgz = 'tgz',
+}
```
styfle: I'm not sure we want to use an enum because this is one of the quirky TS features that works differently depending on using TS proper or some other compiler. And it also behaves differently if adding a `const enum`.

How about separating the valid values (runtime) from the type (build time) like this:

https://github.com/vercel/vercel/blob/181a492d912b4c11970f824afc43638519a5ef7f/packages/routing-utils/src/index.ts#L26-L35

packages/cli/test/mocks/client.ts
```diff
@@ -94,7 +94,9 @@ export class MockClient extends Client {
     this.output = new Output(this.stderr);
 
     this.argv = [];
-    this.authConfig = {};
+    this.authConfig = {
+      token: 'token_dummy',
+    };
```
styfle, line 99: Was this affecting something? I'm not seeing where `token_dummy` is used

MatthewStanciu: This was added because deployments created via unit tests were failing with a `missing_token` error, because `now.create()` expects a token to be present in the client's auth config, and checks for this before making any API calls.

packages/client/src/create-deployment.ts
```diff
@@ -82,7 +86,34 @@ export default function buildCreateDeployment() {
...
 
-    const files = await hashes(fileList);
+    // Populate Files -> FileFsRef mapping
+    const workPath = typeof path === 'string' ? path : path[0];
+
+    let files;
+
+    if (clientOptions.archive === ArchiveFormat.Tgz) {
+      debug('Packing tarball');
+      fileList = fileList.map(file => file.replace(workPath, ''));
```
styfle: Can you inline this on line 99 or do we expect to mutate `fileList` to use it later on?

styfle: Also, replace might cause problems if it replaces more than one pattern.

Lets use relative path since I think that is the intention:

```suggestion
      const entries = fileList.map(file => path.relative(workPath, file));
```

packages/client/src/create-deployment.ts
```diff
@@ -82,7 +86,34 @@ export default function buildCreateDeployment() {
...
+      const tarStream = tar
+        .pack(workPath, {
+          entries: fileList,
+        })
+        .pipe(createGzip());
+      const tarBuffer: Buffer = await streamToBuffer(tarStream);
```
styfle:
```suggestion
      const tarBuffer = await streamToBuffer(tarStream);
```

The type here is redundant

packages/client/src/create-deployment.ts
```diff
@@ -82,7 +86,34 @@ export default function buildCreateDeployment() {
...
+      debug('Packed tarball');
+      files = new Map([
+        [
+          hash(tarBuffer),
```
styfle, line 104: Do we need to worry about buffering all files in memory at once? I wonder how big this will get for large projects? We can probably optimize it later to use a streaming hash perhaps? Although not sure if our files API a supports streaming uploads 😄

packages/client/src/upload.ts
styfle: Was this change necessary for this PR? Might be best to do it in a separate PR because I'm not sure the implications here.

packages/client/src/create-deployment.ts
MatthewStanciu: I don't see why the logic can't be inline on line 99

packages/client/src/create-deployment.ts
TooTallNate: Interestingly, the `hashes()` function already does buffer all files into memory, which is why the logic in `client/src/upload.ts` was able to be removed. At least in this case the hashes are thrown away for a compressed version instead 😆

EndangeredMassa (approved):

packages/client/src/upload.ts
TooTallNate: https://github.com/vercel/vercel/pull/8366

packages/client/src/types.ts
```diff
@@ -25,6 +28,7 @@ export interface VercelClientOptions {
   defaultName?: string;
   isDirectory?: boolean;
   skipAutoDetectionConfirmation?: boolean;
+  archive?: ArchiveFormat | undefined;
```
styfle, line 31: I think this can be `archive?: ArchiveFormat` since undefined is implied by the `?`

PR #8359, opened 2022-08-09
[next] Ensure $$ in ISR page is handled correctly

ijjk:
### Related Issues

This ensures we handle `$` in page names correctly when creating ISR paths. 

x-ref: [slack thread](https://vercel.slack.com/archives/CGU8HUTUH/p1656065059531169)

#### Tests

#### Code Review


packages/next/src/utils.ts
```diff
@@ -1832,7 +1832,7 @@ export const onPrerenderRoute =
     if (nonDynamicSsg || isFallback || isOmitted) {
       outputPathData = outputPathData.replace(
         new RegExp(`${escapeStringRegexp(origRouteFileNoExt)}.json$`),
-        `${routeFileNoExt}.json`
+        `${routeFileNoExt.replace(/\$/g, '$$$$')}.json`
```
lucleray, line 1838: Maybe good to add a comment explaining:
- what we're doing (the `$$$$` can be confusing 😄 i'm only aware it results in `$$` because i read [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter) recently)
- why we're doing it (i'm actually not sure about this, why do we need that line here?)

ijjk: Good point, added comment in https://github.com/vercel/vercel/pull/8359/commits/cfad73551fbc59716a6db5ea6df8af7ae1ab630c

lucleray (approved):

PR #8369, opened 2022-08-10
[cli] MAJOR: Remove update command

cb1kenobi:
The `vc update` command is essentially a noop command that doesn't actually update the CLI. This command is not listed in the `--help` output, nor the CLI documentation. It should be safe to remove in the next major release.

#### Tests

#### Code Review


packages/cli/src/commands/update.ts
```diff
@@ -1,64 +0,0 @@
...
-  }
-
-  if (argv['--help']) {
-    help();
-    return 2;
-  }
-
-  output.log(
-    `Please run ${cmd(
-      await getUpdateCommand()
```
styfle, line 60: This might be kind of useful because the update command here determines if you installed with npm or yarn, which is not that easy for the user to discover, right?

cb1kenobi: Considering most users don't know this command even exists, I don't believe the command is called often.

The case of which package manager command to use is covered by the update notifier, furthermore once https://linear.app/vercel/issue/VCCLI-156/cli-update-notification-should-support-pnpm is implemented.

EndangeredMassa (approved):

styfle (approved):
Let's remove it. We can always add it back if we decide to implement some way to update.

PR #8370, opened 2022-08-10
[cli] Add `Output#link()` function to format terminal hyperlinks

TooTallNate:
This adds a `link()` helper function to the `Output` class that is inspired by the `terminal-link` npm package.

The main difference with this version is that it's more tightly integrated with the `Output` class for the purposes of being able to toggle hyperlinks support on/off for [unit tests](https://github.com/vercel/vercel/blob/4a54b19f46953af2a808bad3e461798e5a40e350/packages/cli/test/unit/util/output/create-output.test.ts) by setting the `output.supportsHyperlink` boolean.

> **Note:** Since hyperlinks are still a relatively new feature, and users might not yet understand how to interact with them, we should only use this function for progressive enhancement scenarios at this time, and _not_ as part of a critical UX.


packages/cli/src/util/output/create-output.ts
```diff
@@ -164,4 +184,28 @@ export class Output {
...
+
+  /**
+   * Returns an ANSI formatted hyperlink when support has been enabled.
+   */
+  link = (
+    text: string,
+    url: string,
+    { fallback }: LinkOptions = {}
+  ): string => {
+    // Credit: https://github.com/sindresorhus/terminal-link
```
MatthewStanciu: I think we should start following a convention that's something like [how @styfle and I attributed part of the `vc env pull` code](https://github.com/vercel/vercel/pull/8170/files#diff-32c1018d233f2c92e3dd39a35fa9abe29eeb1e9dc26d49e603b22023e860fb83). Something like

```
Originally authored by Sindre Sorhus under [license]: [link to license]
[link to piece of code]
```

MatthewStanciu (approved):

packages/cli/src/util/output/create-output.ts
EndangeredMassa: This doesn't seem like a credit-worth amount or significance of code to me, but I'm happy to include it if anyone would like to. If we keep it, an expanded comment (as Matthew mentioned) would probably be best.

EndangeredMassa (approved):
Are we going to use this? If the plan is "not for now", I'd rather close this until we need it.

TooTallNate:
I think we should be.

javivelasco (approved):

TooTallNate:
Btw [`xterm.js` has a PR to support hyperlinks](https://github.com/xtermjs/xterm.js/pull/4005), so they will be supported in VSCode's terminal soon.

MatthewStanciu:
I expect the answer to "Are we going to use this?" to be "Not for now", but even if that's the case I think this should be merged and we should start using terminal links where they can be used.

Related: a good hackathon project or part of a hackathon project could be making the CLI sparkle. Examples could include adding little details like Turbo's "FULL TURBO" thing, and adding terminal links.

packages/cli/src/util/output/create-output.ts
```diff
@@ -49,15 +61,19 @@ export class Output {
     str: string,
     slug: string | null = null,
     link: string | null = null,
-    action: string | null = 'Learn More'
+    action: string = 'Learn More'
```
styfle:
```suggestion
    action = 'Learn More'
```

packages/cli/src/util/output/create-output.ts
```diff
@@ -49,15 +61,19 @@ export class Output {
...
     const details = slug ? `https://err.sh/vercel/${slug}` : link;
 
     this.print(
       chalk.yellow(
         chalk.bold('WARN! ') +
           str +
-          (details ? `\n${action}: ${renderLink(details)}` : '')
+          (details
+            ? `\n${this.link(action, details, {
+                fallback: () => `${action}: ${renderLink(details)}`,
```
styfle: How will users know to click the link? I think we need to be careful to only use this when the link matches the text.

TooTallNate: I reverted the change to `warn()` and `error()`. We can discuss that at a later time.

cb1kenobi (approved):
LGTM

packages/cli/src/util/output/create-output.ts
```diff
@@ -196,8 +196,6 @@ export class Output {
     url: string,
     { fallback }: LinkOptions = {}
   ): string => {
-    // Credit: https://github.com/sindresorhus/terminal-link
```
styfle: I think we need to leave this here if its copy pasta. Should say MIT license

styfle (approved):
Approving based on this important caveat:

> Note: Since hyperlinks are still a relatively new feature, and users might not yet understand how to interact with them, we should only use this function for progressive enhancement scenarios at this time, and not as part of a critical UX

We need to be careful how to use this and I think a good start would be @TooTallNate's idea for `--help` to make options/flags clickable and link to the docs.

PR #8371, opened 2022-08-11
[cli] Support more granular file upload progress bar

TooTallNate:
### Before

Previously, the file upload progress bar would only get updated upon completion of uploading an individual file. This would cause the the progress bar to appear "chunky" when uploading large files. This problem is exasperated with #8356, since there is only a single (potentially large) file that gets uploaded, which would cause the progress bar to not render at all, making the deployment process seem "stuck".

https://user-images.githubusercontent.com/71256/184241009-ba6d1e63-a0fe-462f-804b-e58f56ec310c.mov

### After

Now, the progress bar gets updated incrementally as individual files get uploaded which makes the progress bar updates smoother and play nicely with large files. This is done by "chunking" the individual file into slices and creating a Readable stream from those chunks.

https://user-images.githubusercontent.com/71256/184241050-51832996-d42b-46fd-9a71-d8a1f87a026c.mov


cb1kenobi (approved):
CR'd and LGTM

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -127,6 +123,24 @@ export default async function processDeployment({
...
+
+        uploads.forEach((e: any) =>
+          e.on('progress', () => {
+            if (!bar) return;
+
+            const totalBytesUploaded = uploads.reduce((acc: number, e: any) => {
+              return acc + e.bytesUploaded;
+            }, 0);
+            bar.curr = totalBytesUploaded;
+            bar.tick(0);
```
EndangeredMassa: **suggestion (non-blocking):** make the progress bar updating clearer

Being unfamiliar with the `progress` package, I had to dig a bit to figure out what exactly `bar.tick(0)` was doing here. Turns out, it's [rendering without advancing the counter, checking if the progress is complete](https://github.com/visionmedia/node-progress/blob/master/lib/node-progress.js#L75-L107).

---

We could add a comment about what `bar` is doing here:

```suggestion
            // set the current progress bar value
            bar.curr = totalBytesUploaded;
            // trigger rendering
            bar.tick(0);
```

---

Or, we could leverage `bar.render` directly, which wouldn't need a comment, as much:
 
```suggestion
            bar.curr = totalBytesUploaded;
            bar.render();
```

but it looks like that doesn't call the "are we done" logic. So we'd also have to add a `bar.tick(0)` before checking `bar.complete` to make it work. Or we'd have to do our own checking that `totalBytesUploaded >= missingSize`. Bah.

---

Or we could leverage `bar.update`:

```suggestion
            bar.update(totalBytesUploaded / missingSize);
```

---

Or we could leverage `bar.tick`, but using the diff:

```suggestion
            bar.tick(missingSize - totalBytesUploaded);
```

---

I got carried away with the options. Ultimately, I think this could be a bit clearer, but it's not necessary to change.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -127,6 +123,24 @@ export default async function processDeployment({
           total: missingSize,
           clear: true,
         });
+
+        bar.tick(0);
+
+        uploads.forEach((e: any) =>
+          e.on('progress', () => {
+            if (!bar) return;
```
EndangeredMassa, line 131: **question (non-blocking):** When would `bar` be falsey? Is this here just to satisfy TS?

cb1kenobi (approved):

packages/client/src/upload.ts
```diff
@@ -87,6 +96,27 @@ export async function* upload(
 
         const { data } = file;
 
+        uploadProgress.bytesUploaded = 0;
```
EndangeredMassa, line 99: **suggestion (non-blocking):** remove?

```suggestion
```

Is this necessary? The default value is already `0`, right?

packages/client/src/upload.ts
```diff
@@ -87,6 +96,27 @@ export async function* upload(
...
+
+        // Split out into chunks
+        const body = new Readable();
+        const read = body.read;
```
EndangeredMassa: **suggestion (non-blocking):** rename `read` to `originalRead`

Can we call this `originalRead` (or something like it) to make it clearer what it is?

EndangeredMassa (approved):
Much better user experience!

packages/cli/src/util/deploy/process-deployment.ts
TooTallNate: The reason that I'm updating `curr` directly is in case there's ever a retry on a file upload, in which case we should reset the bytes uploaded count for that file - effectively "subtracting" from the progress. I think adding some comments to the existing code is fine enough.

This `progress` module is kinda "meh" though and I don't think it's really necessary. I'd like to just stick to the Ora spinner and calculate the progress manually. Wanted to do that in a follow-up PR though since this one is already busy enough.

packages/cli/src/util/deploy/process-deployment.ts
TooTallNate: Never, and ya.

packages/client/src/upload.ts
TooTallNate: This is in case the file upload fails and is being retried. So it's a reset of the bytes uploaded counter.

packages/client/src/upload.ts
TooTallNate: Updated

packages/client/src/upload.ts
```diff
@@ -87,6 +96,27 @@ export async function* upload(
...
+          if (chunk) {
+            uploadProgress.bytesUploaded += chunk.length;
+            uploadProgress.emit('progress');
+          }
+          return chunk;
+        };
+
+        const chunkSize = 16384; /* 16kb - default Node.js `highWaterMark` */
+        for (let i = 0; i < data.length; i += chunkSize) {
+          const chunk = data.slice(i, i + chunkSize);
```
styfle, line 115: This can probably be optimized because each call to `.slice()` is going to iterate over the data again (pretty sure .slice() is O(N) time complexity so this algo is O(N^2) potentially). Its probably negligible for small `data` but as it gets larger it might be noticeable.

TooTallNate: I'm not sure what you mean that `.slice()` is going to iterate over the data. It basically just creates a new Buffer "view" ([source](https://github.com/nodejs/node/blob/main/lib/buffer.js#L1122-L1132)). Should be very fast. Are you thinking of `Array#slice()` behavior?

MatthewStanciu (approved):
🔥

packages/client/src/upload.ts
styfle: Yes, I am indeed thinking of `Array#slice()` 😅 

Thanks for clarifying 👍

PR #8375, opened 2022-08-11
Revert "[remix] add image optimization configuration"

EndangeredMassa:
Reverts vercel/vercel#8217

The previous PR had unresolved comments, but the kodiak bot automerged it. Let's revert for now, then we can continue to iterate on code in a new PR.


styfle (approved):
Agreed, adding this to `vercel.json` would be desirable so the feature could be framework agnostic

PR #8377, opened 2022-08-11
[cli] MAJOR: Remove `vc billing` command

MatthewStanciu:
The `vc billing` command has been broken for a while, but nobody has reported it, likely because nobody uses it anyway. `vc billing` is also one of the only JavaScript files left that powers a Vercel CLI command, and it hasn't been touched in years. Because of this, I think it makes sense to get rid of it altogether.

#### Tests

#### Code Review


packages/cli/src/commands/billing/add.js
```diff
@@ -1,170 +0,0 @@
-import ansiEscapes from 'ansi-escapes';
-import chalk from 'chalk';
-import ccValidator from 'credit-card';
```
TooTallNate, line 3: Can we remove this dependency?

packages/cli/src/commands/billing/add.js
```diff
@@ -1,170 +0,0 @@
...
-import textInput from '../../util/input/text';
-import cardBrands from '../../util/billing/card-brands';
```
TooTallNate, line 5: I think we can remove `util/billing` too

packages/cli/src/commands/billing/add.js
```diff
@@ -1,170 +0,0 @@
-import ansiEscapes from 'ansi-escapes';
-import chalk from 'chalk';
-import ccValidator from 'credit-card';
-import textInput from '../../util/input/text';
```
TooTallNate, line 4: Let's double-check, but I think this might be the only place where this input text module was used as well.

TooTallNate (changes requested):
Looks like there's some more stuff we can delete.

packages/cli/src/commands/billing/add.js
MatthewStanciu: It's used in a few other commands as well: `domains`, `dns`, and `teams`.

TooTallNate (approved):

PR #8382, opened 2022-08-12
[cli] Fix `vc env pull` delta changes bug

MatthewStanciu:
#8170 shipped with a small bug: double quotes aren't stripped when pulling existing environment variables, but double quotes are stripped when reading `.env`, so if an environment variable uploaded to Vercel contained double quotes, it would always show as "Changed". This PR fixes the bug.

#### Tests

#### Code Review


packages/cli/src/commands/env/pull.ts
```diff
@@ -117,7 +117,8 @@ export default async function pull(
   if (exists) {
     oldEnv = await createEnvObject(fullPath, output);
     if (oldEnv) {
-      deltaString = buildDeltaString(oldEnv, records);
+      const newEnv = JSON.parse(JSON.stringify(records).replace(/\\"/g, ''));
```
cb1kenobi, line 124: The intent is to remove all `"` from within strings? Since that's specific to building the delta string, perhaps this should go in `buildDeltaString()`?

MatthewStanciu: Possibly, but I think it would be more consistent to keep it outside. The quotes etc. from `oldEnv` are removed in `createEnvObject()`, which is also called before. I think it would be strange to move it to `buildDeltaString()`. But I'm happy to do it if we feel it's better.

cb1kenobi: OK, makes sense. If `createEnvObject()` strips the quotes, then would it be consistent if `exposeSystemEnvs()` did the same?

packages/cli/test/fixtures/unit/vercel-env-pull-delta-quotes/.env
```diff
@@ -0,0 +1,2 @@
+NEW_VAR=testvalue
+SPECIAL_FLAG=1
```
styfle: Should this have a newline? I cant remember if we generate one or not

packages/cli/test/fixtures/unit/vercel-env-pull-delta-quotes/.env
```diff
@@ -0,0 +1,2 @@
+NEW_VAR=testvalue
```
styfle, line 1: What about a test when the env var contains quotes?

packages/cli/src/commands/env/pull.ts
MatthewStanciu: I'm concerned that will have unintended side effects that will be difficult to determine. The point of the manipulation here is just to get each set of environment variables in a state where they can both be compared, but `exposeSystemEnvs()` is used in other places in the code that may expect to receive environment variables with no manipulations.

I don't know too much about how this part of Vercel works to feel confident in this, though.

packages/cli/test/fixtures/unit/vercel-env-pull-delta-quotes/.env
MatthewStanciu: Added newline 👍

packages/cli/test/fixtures/unit/vercel-env-pull-delta-quotes/.env
MatthewStanciu: Added test 👍

cb1kenobi (approved):
LGTM

styfle (approved):

PR #8381, opened 2022-08-12
[ruby] Fix ruby version error message

styfle:
This error message was incorrectly telling user to use the `2.7.x` value even though it will fail `bundle install` with the following error:

```
Your Ruby patchlevel is 0, but your Gemfile specified -1
Command failed with exit code 18 (EXDEV)
```

The exact version, `2.7.0`, is necessary.

PR #8386, opened 2022-08-12
Test ruby gemfile ruby "~> 2.7.x"

styfle:
wip


styfle:
Confirmed every test failed so we'll need to update the message in #8381

PR #8388, opened 2022-08-12
[ruby] Handle exit code 18 gracefully

styfle:
This reverts #8381 and instead supports the "2.7.x" syntax.

- Related to https://github.com/vercel/customer-issues/issues/684

### DIFF
--- a/packages/ruby/index.ts
+++ b/packages/ruby/index.ts
@@ -47,30 +47,27 @@ async function bundleInstall(
 ) {
   debug(`running "bundle install --deployment"...`);
   const bundleAppConfig = await getWriteableDirectory();
+  const { exitCode, stderr, stdout } = await execa(
+    bundlePath,
+    ['install', '--deployment', '--gemfile', gemfilePath, '--path', bundleDir],
+    {
+      stdio: 'pipe',
+      reject: false,
+      env: {
+        ...process.env,
+        BUNDLE_SILENCE_ROOT_WARNING: '1',
+        BUNDLE_APP_CONFIG: bundleAppConfig,
+        BUNDLE_JOBS: '4',
+      },
+    }
+  );
 
-  try {
-    await execa(
-      bundlePath,
-      [
-        'install',
-        '--deployment',
-        '--gemfile',
-        gemfilePath,
-        '--path',
-        bundleDir,
-      ],
-      {
-        stdio: 'pipe',
-        env: {
-          BUNDLE_SILENCE_ROOT_WARNING: '1',
-          BUNDLE_APP_CONFIG: bundleAppConfig,
-          BUNDLE_JOBS: '4',
-        },
-      }
-    );
-  } catch (err) {
-    debug(`failed to run "bundle install --deployment"...`);
-    throw err;
+  if (exitCode === 0 || exitCode === 18) {
+    // Gemfile might contain "2.7.x" so install might exit with code 18 and message:
+    // "Your Ruby patchlevel is 0, but your Gemfile specified -1"
+    // See https://github.com/rubygems/bundler/blob/3f0638c6c8d340c2f2405ecb84eb3b39c433e36e/lib/bundler/errors.rb#L49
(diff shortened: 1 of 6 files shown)


packages/ruby/index.ts
```diff
@@ -142,14 +139,7 @@ export async function build({
       } else {
         // try installing. this won't work if native extesions are required.
         // if that's the case, gems should be vendored locally before deploying.
-        try {
-          await bundleInstall(bundlerPath, bundleDir, gemfilePath);
-        } catch (err) {
-          debug(
-            'unable to build gems from Gemfile. vendor the gems locally with "bundle install --deployment" and retry.'
-          );
-          throw err;
```
styfle, line 151: This was a double try/catch, so I removed this one in favor of the inner one in `bundleInstall`

packages/ruby/index.ts
```diff
@@ -47,30 +47,27 @@ async function bundleInstall(
...
-    );
-  } catch (err) {
-    debug(`failed to run "bundle install --deployment"...`);
-    throw err;
+  if (exitCode === 0 || exitCode === 18) {
+    // Gemfile might contain "2.7.x" so install might exit with code 18 and message:
+    // "Your Ruby patchlevel is 0, but your Gemfile specified -1"
+    // See https://github.com/rubygems/bundler/blob/3f0638c6c8d340c2f2405ecb84eb3b39c433e36e/lib/bundler/errors.rb#L49
+  } else {
+    throw new Error(stdout + stderr);
```
cb1kenobi: Do we care about the `exitCode` when there is a non-zero, non-18 exit code? Append a ` (code ${exitCode})` to the error message or set an `err.exitCode = exitCode`?

packages/ruby/index.ts
```diff
@@ -142,14 +139,7 @@ export async function build({
...
         // if that's the case, gems should be vendored locally before deploying.
-        try {
-          await bundleInstall(bundlerPath, bundleDir, gemfilePath);
-        } catch (err) {
-          debug(
-            'unable to build gems from Gemfile. vendor the gems locally with "bundle install --deployment" and retry.'
-          );
-          throw err;
-        }
+        await bundleInstall(bundlerPath, bundleDir, gemfilePath);
```
cb1kenobi, line 147: That debug message is no longer relevant if an error occurs? Are there enough clues in the stdout + stderr to indicate how to proceed when an error occurs?

packages/ruby/index.ts
styfle: Good point! I updated this message to look more like the previous message:

```
"bundle install" failed with exit code ${exitCode}
```

packages/ruby/index.ts
styfle: It should have more information than before because we were previously hiding the logs of bundle install

Also see the latest change in [df4f8ee](https://github.com/vercel/vercel/pull/8388/commits/df4f8eec75b7c93f66d405119a31f22ab4ded854)

packages/ruby/index.ts
```diff
@@ -47,30 +47,29 @@ async function bundleInstall(
...
   debug(`running "bundle install --deployment"...`);
   const bundleAppConfig = await getWriteableDirectory();
+  const { exitCode, stderr, stdout } = await execa(
+    bundlePath,
+    ['install', '--deployment', '--gemfile', gemfilePath, '--path', bundleDir],
+    {
+      stdio: 'pipe',
+      reject: false,
+      env: {
+        ...process.env,
```
cb1kenobi, line 57: Very nice

cb1kenobi (approved):

cb1kenobi (approved):

packages/ruby/index.ts
styfle: Subtle change but I think it was a bug to not include the env vars. Didn't make a difference in this case.

styfle:
This didn't work as expected. See #8392

PR #8392, opened 2022-08-12
[ruby] Fix `bundle install`

styfle:
This PR is a follow up to #8388. The tests were not running as expected because only the "vendored" test used `2.7.x` which was never failing `bundle install` to begin with. I had to add `2.7.x` to the non-vendored test to get `bundle install` to fail. The fix is to rename `2.7.x` to `2.7.0` before running `bundle install`.


cb1kenobi (approved):
Guess we don’t need to worry about exit code 18. :)

PR #8397, opened 2022-08-14
[cli] Replace "progress" module with custom implementation

TooTallNate:
The `progress` module is overkill for what we are using it for and it's inconsistent with our styling for the rest of the deployment process.

So drop the npm module in favor of an in-house progress bar implementation that can be used inside of the Ora `output.spinner()` that we already use for the other steps of the deployment process.

**Before:**

Screen Shot 2022-08-25 at 5 07 28 PM

**After:**

Screen Shot 2022-08-25 at 5 09 35 PM


packages/cli/src/util/output/progress.ts
```diff
@@ -0,0 +1,22 @@
...
+export function progress(
+  current: number,
+  total: number,
+  opts: ProgressOptions = {}
+): string {
+  const width = opts?.width ?? 20;
+  const complete = opts?.complete ?? '=';
+  const incomplete = opts?.incomplete ?? '_';
+  const unit = total / width;
+  let pos = unit === 0 ? width : Math.floor(current / unit);
```
EndangeredMassa: **suggestion (non-blocking):** validate `total > 0`

If `unit` is `0`, that means `total` must have been `0`, which seems like an invalid case to call this function. Am I missing something?

If that's the case, I think we should validate that total is `> 0` at the top of this function and maybe debug log an error message and return an empty progress bar. The call site can decide what to do with an empty result, like showing a generic "Downloading..." message.

While we're validating inputs, asserting that `current >= 0` is also a good idea.

packages/cli/test/unit/output/progress.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    { current: 0, total: 5, opts: { width: 5 }, expected: '_____' },
+    { current: 1, total: 5, opts: { width: 5 }, expected: '=____' },
+    { current: 2, total: 5, opts: { width: 5 }, expected: '==___' },
+    { current: 3, total: 5, opts: { width: 5 }, expected: '===__' },
+    { current: 4, total: 5, opts: { width: 5 }, expected: '====_' },
+    { current: 5, total: 5, opts: { width: 5 }, expected: '=====' },
+    { current: 0, total: 12, expected: '____________________' },
+    { current: 1, total: 12, expected: '=___________________' },
+    { current: 2, total: 12, expected: '===_________________' },
+    { current: 600, total: 1200, expected: '==========__________' },
```
EndangeredMassa: **suggestion (non-blocking):** add test cases

Can we add a case where `current > total`?

Should we add a case for the `complete`, and `incomplete` options?

EndangeredMassa (approved):

packages/cli/src/util/output/progress.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Returns a raw progress bar string.
+ */
+export function progress(
+  current: number,
+  total: number,
+  opts: ProgressOptions = {}
+): string {
+  const width = opts?.width ?? 20;
+  const complete = opts?.complete ?? '=';
+  const incomplete = opts?.incomplete ?? '_';
```
styfle:
```suggestion
  const {
    width = 20,
    complete = '=', 
    incomplete = '_',
} = opts;
```

packages/cli/test/unit/output/progress.test.ts
styfle: Agreed, we should definitely have a test when `current > total` since I think we had that bug in the past.

EndangeredMassa (approved):

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -114,32 +113,34 @@ export default async function processDeployment({
...
-            // set the current progress bar value
-            bar.curr = totalBytesUploaded;
-            // trigger rendering
-            bar.tick(0);
 
-            if (bar.complete) {
+            if (uploadedBytes === missingSize) {
               output.spinner(deployingSpinnerVal, 0);
+            } else {
+              const bar = progress(uploadedBytes, missingSize, progressOpts);
```
styfle: Since `bar` can be null now, should we add a debug log in that case and assume 100%?

TooTallNate: Ok added.

packages/cli/src/util/output/progress.ts
EndangeredMassa: If we do this:

```suggestion
  let pos = Math.floor(current / unit);
```

All the tests still pass. What's the intent behind the `unit === 0` check? If this function is called with a `total` of `0`, the result is empty string.

Should we make this change and add validations?

```
if (total <= 0 || current < 0) {
  return null;
} 
```

EndangeredMassa (approved):
I think we should modify the logic a bit, but this is shippable in its current form.

styfle (approved):

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -114,32 +113,34 @@ export default async function processDeployment({
...
+        const progressOpts = {
           width: 20,
           complete: '=',
-          incomplete: '',
-          total: missingSize,
-          clear: true,
-        });
-
-        bar.tick(0);
+          incomplete: '-',
```
cb1kenobi: Perhaps instead of passing in the complete/incomplete characters, use the defaults defined in `progress()`. That will ensure we have better consistency when we go to add a progress bar elsewhere.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -114,32 +113,34 @@ export default async function processDeployment({
...
-            if (bar.complete) {
+            const bar = progress(uploadedBytes, missingSize, progressOpts);
+            if (!bar || uploadedBytes === missingSize) {
               output.spinner(deployingSpinnerVal, 0);
+            } else {
+              const uploadedHuman = bytes.format(uploadedBytes, {
+                decimalPlaces: 1,
+                fixedDecimals: true,
+              });
+              output.spinner(
```
cb1kenobi, line 132: I like this!

packages/cli/src/util/deploy/process-deployment.ts
cb1kenobi: In hindsight, maybe just remove complete/incomplete? Don’t even give devs the option to be inconsistent.

Ethan-Arrowood (approved):
lgtm

PR #8399, opened 2022-08-15
[cli] fix vc bisect off by one

EndangeredMassa:
There's a bug in `vc bisect` where the known "good" deployment ends up in the list of deployments to check. This can result in the command asking the user if the known "good" deployment is good or bad, which it should already know.

This PR fixes the issue and adds a test that proves it works.

### DIFF
--- a/packages/cli/src/commands/bisect/index.ts
+++ b/packages/cli/src/commands/bisect/index.ts
@@ -149,7 +149,9 @@ export default async function main(client: Client): Promise<number> {
 
   if (badDeployment) {
     if (badDeployment instanceof Error) {
-      badDeployment.message += ` when requesting "${normalizeURL(bad)}"`;
+      badDeployment.message += ` when requesting bad deployment "${normalizeURL(
+        bad
+      )}"`;
       output.prettyError(badDeployment);
       return 1;
     }
@@ -164,7 +166,9 @@ export default async function main(client: Client): Promise<number> {
 
   if (goodDeployment) {
     if (goodDeployment instanceof Error) {
-      goodDeployment.message += ` "${good}"`;
+      goodDeployment.message += ` when requesting good deployment "${normalizeURL(
+        good
+      )}"`;
       output.prettyError(goodDeployment);
       return 1;
     }
--- a/packages/cli/src/commands/bisect/index.ts
+++ b/packages/cli/src/commands/bisect/index.ts
@@ -2,7 +2,6 @@ import open from 'open';
 import boxen from 'boxen';
 import execa from 'execa';
 import plural from 'pluralize';
-import inquirer from 'inquirer';
 import { resolve } from 'path';
 import chalk, { Chalk } from 'chalk';
 import { URLSearchParams, parse } from 'url';
@@ -150,7 +149,7 @@ export default async function main(client: Client): Promise<number> {
 
   if (badDeployment) {
     if (badDeployment instanceof Error) {
-      badDeployment.message += ` "${bad}"`;
+      badDeployment.message += ` when requesting "${normalizeURL(bad)}"`;
       output.prettyError(badDeployment);
       return 1;
     }
@@ -226,7 +225,8 @@ export default async function main(client: Client): Promise<number> {
     // If we have the "good" deployment in this chunk, then we're done
     for (let i = 0; i < newDeployments.length; i++) {
       if (newDeployments[i].url === good) {
-        newDeployments = newDeployments.slice(0, i + 1);
+        // grab all deployments up until the good one
+        newDeployments = newDeployments.slice(0, i);
         next = undefined;
         break;
       }
@@ -316,7 +316,7 @@ export default async function main(client: Client): Promise<number> {
       if (openEnabled) {
         await open(testUrl);
       }
-      const answer = await inquirer.prompt({
+      const answer = await client.prompt({
         type: 'expand',
         name: 'action',
         message: 'Select an action:',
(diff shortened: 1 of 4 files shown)


packages/cli/test/unit/commands/bisect.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    );
+    client.stdin.write(`${deployment1.url}\n`);
+
+    await expect(client.stderr).toOutput(
+      'Specify the URL subpath where the bug occurs:'
+    );
+    client.stdin.write('/docs\n');
+
+    await expect(client.stderr).toOutput(
+      `The first bad deployment is: https://${deployment1.url}`
```
EndangeredMassa, line 42: Is something wrong with my test setup? This passes, but `deployment1` should be a good deployment, not a bad one.

EndangeredMassa: Ah, it's because of the order of the deployments returned in the request for all of them

packages/cli/test/unit/commands/bisect.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    await expect(client.stderr).toOutput('Log in with SAML?');
+    client.stdin.write('n\n');
+
+    await expect(client.stderr).toOutput(
+      'Response Error (403) when requesting "https://some-9s8df0.vercel.sh"'
+    );
+
+    expect(bisectPromise).resolves.toBe(1);
+
+    console.log('TEST OVER');
```
EndangeredMassa: This test passes, but the test suite fails. It looks like there's an unhandled rejected promise or something, but I cannot figure out why.

```
$  yarn jest test/unit/commands/bisect.test.ts
yarn run v1.22.18
$ /Users/smassa/source/vercel/vercel/node_modules/.bin/jest test/unit/commands/bisect.test.ts
 FAIL  test/unit/commands/bisect.test.ts
  ● Console

    console.log
      TEST OVER

      at Object.<anonymous> (test/unit/commands/bisect.test.ts:90:13)



  ● Test suite failed to run

    Response Error (403) when requesting "https://some-9s8df0.vercel.sh"

      23 |
      24 |   const msg = bodyError?.message || fallbackMessage || 'Response Error';
    > 25 |   return new APIError(msg, res, bodyError);
         |          ^
      26 | }
      27 |

      at responseError (src/util/response-error.ts:25:10)
      at src/util/client.ts:146:23
      at main (src/commands/bisect/index.ts:148:25)

Test Suites: 1 failed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        0.716 s, estimated 3 s
Ran all test suites matching /test\/unit\/commands\/bisect.test.ts/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

You can see that the test completes. The error it prints is handled and has assertions run against it just fine. The test itself passes, but the test suite fails.

Does anyone have any ideas here?

packages/cli/src/util/client.ts
```diff
@@ -186,7 +191,7 @@ export default class Client extends EventEmitter implements Stdio {
           `Failed to re-authenticate for ${bold(error.scope)} scope`
         );
       }
-      process.exit(1);
+      throw error;
```
EndangeredMassa, line 195: We need to remove `process.exit` to make this testable.

packages/cli/src/commands/bisect/index.ts
```diff
@@ -228,7 +223,8 @@ export default async function main(client: Client): Promise<number> {
     // If we have the "good" deployment in this chunk, then we're done
     for (let i = 0; i < newDeployments.length; i++) {
       if (newDeployments[i].url === good) {
-        newDeployments = newDeployments.slice(0, i + 1);
+        // grab all deployments up until the good one
+        newDeployments = newDeployments.slice(0, i);
```
EndangeredMassa, line 233: The actual bug fix

packages/cli/src/commands/bisect/index.ts
```diff
@@ -148,14 +143,13 @@ export default async function main(client: Client): Promise<number> {
   }
 
   output.spinner('Retrieving deployments…');
-  const [badDeployment, goodDeployment] = await Promise.all([
-    badDeploymentPromise,
-    goodDeploymentPromise,
-  ]);
+
+  // `getDeployment` cannot be parallelized because it might prompt for login
```
EndangeredMassa: This is the main bug fix. When we parallelize the lookups for the good and bad deployment, if the current user's auth doesn't have the proper scope, the user will be prompted to log in with SAML. This is async, though, which stuffs that prompt into your face when you are trying to interact with other prompts, like selecting a subpath.

packages/cli/test/unit/commands/bisect.test.ts
EndangeredMassa: I tried updating jest, just in case there would be some guidance on what was wrong. Same result.

TooTallNate (approved):

packages/cli/src/util/client.ts
```diff
@@ -146,10 +146,15 @@ export default class Client extends EventEmitter implements Stdio {
...
-          // to re-authenticate
-          await this.reauthenticate(error);
+          try {
+            // A SAML error means the token is expired, or is not
+            // designated for the requested team, so the user needs
+            // to re-authenticate
+            await this.reauthenticate(error);
+          } catch (reauthError) {
+            // there's no sense in retrying
+            return bail(reauthError as Error);
```
TooTallNate: To avoid the `as` cast, import this function from `is-error.ts`.

```suggestion
            return bail(normalizeError(reauthError));
```

PR #8400, opened 2022-08-15
[cli][tests] remove dupe `inspectorUrl`

EndangeredMassa:
This duplicate value was added in https://github.com/vercel/vercel/pull/8151/files#diff-912ecd0b3f7b44cac96f04be048647f24d0efbb34593c0145455b9edea7c104fR27 but the diff window didn't show that the value already existed.

I added the lint rule for this. There are no other offenders to fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
           "selector": "MemberExpression > Identifier[name='substr']"
         }
       ],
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
LGTM

PR #8403, opened 2022-08-15
[cli] serialize getDeployment calls in `vercel bisect`

EndangeredMassa:
When we parallelize the lookups for the good and bad deployment, if the current user's auth doesn't have the proper scope, the user will be prompted to log in with SAML. This is async, though, which stuffs that prompt into your face when you are trying to interact with other prompts, like selecting a subpath.

I tried writing tests for this, but am having some trouble. See https://github.com/vercel/vercel/pull/8399

### DIFF
--- a/packages/cli/src/commands/bisect/index.ts
+++ b/packages/cli/src/commands/bisect/index.ts
@@ -116,8 +116,6 @@ export default async function main(client: Client): Promise<number> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


EndangeredMassa:
I manually tested that this still works

cb1kenobi (approved):
LGTM

PR #8411, opened 2022-08-16
[cli] Change mentions of "Aborted" to "Canceled"

MatthewStanciu:
@leerob flagged that the language "Aborted" throughout the CLI should be changed to be more inclusive. This PR updates this throughout.

I didn't go as far as to update type names because I wanted to avoid potentially unintended consequences, but happy to also do that if we want.

#### Tests

#### Code Review


styfle (changes requested):
Let's use "Canceled" with one L since that's how we spell it elsewhere.

For example https://vercel.com/blog/canceling-ongoing-deployments

packages/cli/src/commands/alias/rm.ts
```diff
@@ -59,7 +59,7 @@ export default async function rm(
 
   const removeStamp = stamp();
   if (!opts['--yes'] && !(await confirmAliasRemove(client, alias))) {
-    output.log('Aborted');
+    output.log('Canceled.');
```
cb1kenobi: I think we should be consistent whether the period is present. I would argue that if it's only a single word, then omit the period.

packages/cli/src/commands/build.ts
```diff
@@ -171,7 +171,7 @@ export default async function main(client: Client): Promise<number> {
       );
     }
     if (!confirmed) {
-      client.output.print(`Aborted. No Project Settings retrieved.\n`);
+      client.output.print(`Cancelled. No Project Settings retrieved.\n`);
```
```suggestion
      client.output.print(`Canceled. No Project Settings retrieved.\n`);
```

styfle (approved):
Thanks!

packages/cli/src/commands/alias/rm.ts
MatthewStanciu: Thanks, just did this in the latest commit

cb1kenobi (approved):
LGTM

leerob (approved):
Thank you so much!

javivelasco (approved):

EndangeredMassa (approved):

PR #8412, opened 2022-08-16
[cli] Fix `vc link` not respecting `--yes`

MatthewStanciu:
As of 28.0.0, `vc link` stopped respecting `--yes` and would wait for prompt if it found a Git repository to connect. This PR passes `autoConfirm` through to the Git prompt, so that `vc link` continues to respect `--yes`.

Here's what `vc link` looks like after this PR if a user has a local Git repository:

Screen Shot 2022-08-16 at 10 05 56 AM

Not sure if we want to remove the log messages to be consistent with the rest of `vc link`?

#### Tests

#### Code Review


cb1kenobi (approved):
LGTM

packages/cli/src/util/link/add-git-connection.ts
```diff
@@ -81,13 +85,17 @@ async function addSingleGitRemote(
...
+  if (!shouldConnect) {
+    shouldConnect = await promptGitConnectSingleUrl(
+      client,
+      project,
+      remoteName,
+      remoteUrl,
+      replace
+    );
+  }
   return handleOptions(shouldConnect, client, org, project, settings, repoInfo);
```
EndangeredMassa: **suggestion (non-blocking):** rearrange logic to make it clearer

```suggestion
  let shouldConnectOption: string | undefined;
  if (autoConfirm) {
    shouldConnectOption = 'yes';
  } else {
    shouldConnectOption = await promptGitConnectSingleUrl(
      client,
      project,
      remoteName,
      remoteUrl,
      replace
    );
  }
  return handleOptions(shouldConnectOption, client, org, project, settings, repoInfo);
```

I had to read this several times to understand what was going on. The `option` parameter of `promptGitConnectSingleUrl` made it a bit more confusing for me.

I suggest rearranging this logic to clarify.

packages/cli/src/util/link/add-git-connection.ts
```diff
@@ -96,12 +104,13 @@ async function addMultipleGitRemotes(
...
-  settings: ProjectSettings
+  settings: ProjectSettings,
+  autoConfirm: Boolean
 ) {
-  client.output.log('Found multiple Git remote URLs in Git config.');
-  const remoteUrlOrOptions = await promptGitConnectMultipleUrls(
-    client,
-    remoteUrls
-  );
+  let remoteUrlOrOptions = autoConfirm ? Object.values(remoteUrls)[0] : '';
```
EndangeredMassa: **suggestion (non-blocking):** use more elaborate defaults

I suggest defaulting like so:

- if only 1 remote
  - use it
- else (many remotes)
  - if `origin` exists
    - use `origin` 
  - else
    - use first remote

Changing this later should probably be considered a breaking change. We may want to address this now.

packages/cli/src/util/link/add-git-connection.ts
```diff
@@ -96,12 +104,13 @@ async function addMultipleGitRemotes(
...
+  if (!remoteUrlOrOptions) {
+    client.output.log('Found multiple Git remote URLs in Git config.');
+    remoteUrlOrOptions = await promptGitConnectMultipleUrls(client, remoteUrls);
+  }
   return handleOptions(remoteUrlOrOptions, client, org, project, settings);
```
EndangeredMassa: **suggestion (non-blocking):** rearrange logic for clarity

The use of `remoteUrlOrOptions` for multiple values is confusing.

```suggestion
  let remoteUrl: string | undefined;
  if (autoConfirm) {
    remoteUrl = Object.values(remoteUrls)[0];
  } else {
    client.output.log('Found multiple Git remote URLs in Git config.');
    remoteUrl = await promptGitConnectMultipleUrls(client, remoteUrls);
  }
  return handleOptions(remoteUrl, client, org, project, settings);
```

EndangeredMassa (approved):

packages/cli/src/util/link/add-git-connection.ts
MatthewStanciu: Thanks 👍 I think I addressed this in my latest commit, but let me know if you had something else in mind.

packages/cli/test/unit/commands/link.test.ts
```diff
@@ -306,5 +306,28 @@ describe('link', () => {
...
+        client.setArgv('--yes');
+        const linkPromise = link(client);
+        expect(client.stderr).not.toOutput('Do you want to connect "origin"');
+        await expect(client.stderr).toOutput('Linked to');
+        await expect(linkPromise).resolves.toEqual(0);
+      } finally {
+        await fs.rename(join(cwd, '.git'), join(cwd, 'git'));
+        process.chdir(originalCwd);
+      }
+    });
```
EndangeredMassa, line 331: **suggestion (soft-blocking):** add test for multiple remotes

In this or a follow-up PR: Can we add a test for multiple remotes that prefer `origin` when `origin` is not the first item in the list?

EndangeredMassa (approved):

packages/cli/test/unit/commands/link.test.ts
MatthewStanciu: Added test!

EndangeredMassa (approved):

PR #8413, opened 2022-08-16
[hydrogen] Polyfill Oxygen global in Hydrogen

frandiox:
### Related Issues

It's currently not possible to access environment variables in Hydrogen projects deployed to Vercel. The reason is that Hydrogen currently relies on a global [`Oxygen.env` that needs to be polyfilled](https://shopify.dev/custom-storefronts/hydrogen/framework/environment-variables#private-variables-in-production) in the platform. This PR adds the global object using `process.env`, which is [mentioned in the docs](https://vercel.com/docs/concepts/functions/edge-functions/edge-functions-api#environment-variables) as the way to access environment variables in Edge Functions.

@TooTallNate Can you double check if this is correct? I'm not quite sure `process.env` is the right thing here even if it's mentioned in the docs 😅 

#### Tests

#### Code Review

### DIFF
--- a/packages/hydrogen/edge-entry.js
+++ b/packages/hydrogen/edge-entry.js
@@ -5,8 +5,10 @@ import indexTemplate from '__RELATIVE__/dist/client/index.html?raw';
 import { ReadableStream } from 'web-streams-polyfill/ponyfill';
 Object.assign(globalThis, { ReadableStream });
 
-export default (request, event) =>
-  handleRequest(request, {
+export default (request, event) => {
+  globalThis.Oxygen = { env: process.env };
+  return handleRequest(request, {
     indexTemplate,
     context: event,
   });
+}


packages/hydrogen/edge-entry.js
```diff
@@ -5,8 +5,10 @@ import indexTemplate from '__RELATIVE__/dist/client/index.html?raw';
 import { ReadableStream } from 'web-streams-polyfill/ponyfill';
 Object.assign(globalThis, { ReadableStream });
```
EndangeredMassa, line 6: We may want to do `globalThis` changes in one clear place up top, like:

```
Object.assign(globalThis, {
  // ReadableStream is bugged in Vercel Edge, overwrite with polyfill
  ReadableStream,

  // Hydrogen exposes env vars through `Oxygen.env`
  Oxygen: {
    env: process.env
  }
});
```

EndangeredMassa (commented):
Can you also add a test that shows the environment variables being used? You can make a new fixture, [set env vars](https://vercel.com/docs/project-configuration#legacy/env), make sure they are used in the fixture app somewhere, then leverage a vercel.json probe to test for that value.

If not, no worries! We can add a test for this.

packages/hydrogen/edge-entry.js
cb1kenobi: I second this idea. Global is global and it's best outside of a function.

mperreux: 👋 hey I made this change here https://github.com/vercel/vercel/pull/8496 if you want to take a look

styfle:
Closing in favor of #8496

PR #8417, opened 2022-08-17
[node] fix WebAssembly bindings evaluating to `undefined`

Schniz:
This was broken since https://github.com/vercel/vercel/pull/8242, which accidentally pushed all WebAssembly modules to be nested under `global.wasmBindings` instead of being directly under the global scope.

### Related Issues

- This was broken since https://github.com/vercel/vercel/pull/8242

#### Tests

#### Code Review


Schniz:
cc @EndangeredMassa @TooTallNate how would you suggest me to test this to make sure it does not break in the future? I am not sure we have the foundations to test `vc dev` right now, but I might be wrong!

cb1kenobi (approved):
Tested and works!

EndangeredMassa (approved):
There is a test that basic WASM support works. That test passed in the offending PR: https://github.com/vercel/vercel/runs/7585545987?check_suite_focus=true

Is there deeper functionality we should add to that test? I'm confused about why that test wasn't failing.

cb1kenobi:
Are you thinking of adding a `dev` test or fixing the test which incorrectly passed?

EndangeredMassa:
Either one. As long as this is covered appropriately.

Schniz:
@EndangeredMassa  can you point me to where I can find these tests so I would add another example?

EndangeredMassa:
@Schniz certainly! https://github.com/vercel/vercel/blob/main/packages/node/test/fixtures/61-edge-function/vercel.json

EndangeredMassa (approved):
I verified that the new tests only works with the changes.

cb1kenobi (approved):

PR #8422, opened 2022-08-18
[go] ignore bad go handlers and rename functions with more precision

EndangeredMassa:
The `analyze.go` program uses the Go AST package to parse the user's Go file to look for the exported handler function. To be found, the function has to meet certain criteria:

- have two parameters
- one parameter must be of type `*http.ResponseWriter`
- one parameter must be of type `*http.Request`

The problem is that Receiver Functions can match these parameters, but will fail to be executed by our Go wrapper logic. The error you get in this case is [very cryptic to the user](https://github.com/vercel/vercel/runs/7911138926?check_suite_focus=true):

```
../main__vc__go__.go:30:28: undefined: handler.BadReceiverHandler_api_bad_receiver_go
```

This PR adds another condition that:

- function is not a receiver function

The error shown to the user in this case is now:

```
Could not find an exported function in "${entrypointWithExt}"
Learn more: https://vercel.com/docs/runtimes#official-runtimes/go
```

---

Example Good Handler:

```
func GoodHandler(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "version:%v:from GoodHandler", runtime.Version())
}

```

Example Bad Handler (Receiver Function):

```
func (d *SampleDecoder) BadReceiverHandler(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "version:%v:from BadHandler", runtime.Version())
}
```


packages/go/test/fixtures/24-bad-handler/vercel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "version": 2,
+  "probes": [
+    {
+      "path": "/api/bad-receiver",
+      "mustContain": "version:go1.13.15:from GoodHandler"
+    }
```
EndangeredMassa: I ran this test without the fix to make sure it fails. It does: https://github.com/vercel/vercel/runs/7911138926?check_suite_focus=true

packages/go/util/analyze.go
```diff
@@ -223,7 +225,7 @@ func main() {
 		if fn.Name.IsExported() == true {
 			for _, param := range fn.Type.Params.List {
 				paramStr := fmt.Sprintf("%s", param.Type)
-				if strings.Contains(string(paramStr), "http ResponseWriter") && len(fn.Type.Params.List) == 2 {
+				if strings.Contains(string(paramStr), "http ResponseWriter") && len(fn.Type.Params.List) == 2 && (fn.Recv == nil || len(fn.Recv.List) == 0) {
```
EndangeredMassa, line 228: Should this fallback logic also check for `strings.Contains(string(params), "*http.Request")`, or is that absent because of the parsing issue this fallback exists to solve?

packages/go/go.mod
```diff
@@ -0,0 +1,3 @@
+module main
+
+go 1.18
```
EndangeredMassa, line 3: This allows the direct invocation of `utils/analyze.go` for integrated debugging.

Ethan-Arrowood (approved):
i'm not super confident in my Go skills, but this lgtm. the tests seem to demonstrate what was broken and how it was fixed so im happy to approve 👍

cb1kenobi:
I can reproduce the issue, but I can't seem to get the fix to work. How do I test this?

EndangeredMassa:
@cb1kenobi to test this locally you have to specify the tarballs to use. This is most easily done by grabbing the CLI tarball generated from the deployment in this PR https://vercel-3k4aefbfm.vercel.sh/tarballs. Then running it locally from `./packages/go`:

```
VERCEL_CLI_VERSION=https://vercel-3k4aefbfm.vercel.sh/tarballs/vercel.tgz yarn test test/fixtures.test.js
```

cb1kenobi:
@EndangeredMassa I tried that, though I'm trying to reproduce with a real project with a bad go handler that uses the same `BadReceiverHandler()` that you had and I still see the cryptic message:

~~~
[12:57:15.392] failed to `go build`
[12:57:15.396] Error! Command failed: go build -ldflags -s -w -o /tmp/616b540f/handler /vercel/path1/main__vc__go__.go
[12:57:15.396] # command-line-arguments
[12:57:15.396] ../main__vc__go__.go:30:36: undefined: handler.BadHandler_api_fun_go
~~~

cb1kenobi (approved):

EndangeredMassa:
For reference: @cb1kenobi and I resolved the testing issue. When using a tarball as part of a deploy, you have to set the build env: `vc deploy -b VERCEL_CLI_VERSION=...`

packages/go/go.mod
styfle: Does go.mod support comments? Might be good to add that here

packages/go/test/fixtures/24-bad-handler/vercel.json
styfle: Lets rename this `vercel.json` file to `probes.json` and remove `version` prop

packages/go/index.ts
```diff
@@ -534,8 +531,8 @@ export async function build({
 async function renameHandlerFunction(fsPath: string, from: string, to: string) {
   let fileContents = await readFile(fsPath, 'utf8');
 
-  const fromRegex = new RegExp(`\\b${from}\\b`, 'g');
-  fileContents = fileContents.replace(fromRegex, to);
+  const fromRegex = new RegExp(String.raw`func (${from})\(`, 'g');
```
EndangeredMassa: The simpler version before would rename parts it shouldn't which is obvious in retrospect. The updated tests confirms that will not happen going forward.

The right balance for this replacement is important. Is this current form too restrictive? @cb1kenobi @Ethan-Arrowood @styfle @TooTallNate

EndangeredMassa:
I updated this PR with a more complete fix for Go handler detection and renaming.

packages/go/go.mod
EndangeredMassa: Yup! [`9cb054c` (#8422)](https://github.com/vercel/vercel/pull/8422/commits/9cb054c07547479fc8394479c6a14f6b2d16e8c2)

packages/go/index.ts
cb1kenobi: From what I know about Go, I can't say if it's too restrictive from a language standpoint, but I believe your change is certainly safer than a naive word boundary based replacement.

The only thing I'd say is the parens around the `${from}` are unnecessary and can lead to confusion, so they could be omitted:

```javascript
const fromRegex = new RegExp(String.raw`func ${from}\(`, 'g');
```

packages/go/test/fixtures/24-bad-handler/vercel.json
EndangeredMassa: Sure thing: [`550fd0f` (#8422)](https://github.com/vercel/vercel/pull/8422/commits/550fd0f8582f26910834a4654caef704c6a46ae6)

packages/go/index.ts
styfle: I think the correct thing to do here is switch from regex to parsing the AST. We should already have it from `analyze.go`, right?

cb1kenobi: That would be ideal, but is that reasonable? Wouldn't `analyze.go` need to be reworked to rewrite the function name as it walks the AST and then serialize the tree back to disk? I'm sure it's cheap to do, but I'd guess the level of effort is more than a simple string replacement in Node.

styfle: Right, I dont know the effort to write the AST back to a string

EndangeredMassa: I agree that the best solution is to use the parsed AST, change it, and write out the result. I just don't know nearly enough Go to handle that. We asked around about doing this before with resident Go devs, but they were pretty busy with other work. I think we can ask again now.

I just updated this to be more lax. The new form is ` ${from}\b`, which accounts for all known syntax variations.

packages/go/index.ts
```diff
@@ -531,8 +531,14 @@ export async function build({
 async function renameHandlerFunction(fsPath: string, from: string, to: string) {
   let fileContents = await readFile(fsPath, 'utf8');
 
-  const fromRegex = new RegExp(String.raw`func (${from})\(`, 'g');
-  fileContents = fileContents.replace(fromRegex, `func ${to}(`);
+  // This regex has to walk a fine line where it replaces the mostl-likey occurences
```
styfle:
```suggestion
  // This regex has to walk a fine line where it replaces the most-likely occurrences
```

packages/go/util/analyze.go
EndangeredMassa: @jarneson do you have any idea why this condition was not included here? I can't tell if it's intentional or accidental.

styfle (approved):

PR #8432, opened 2022-08-19
Update blitz framework properties to match legacy blitz detection

v3dillon:
A simple update to the framework array, updating the blitz properties to match the legacy framework. This PR detects a `blitz.config.(ts|js)` file which is specific to the legacy framework. With the blitz 2.0 update the framework detection should automatically be next.js.

#### Tests

- [X] The code changed/added as part of this PR has been covered with tests
- [X] All tests pass locally with `yarn test-unit`

#### Code Review

- [X] This PR has a concise title and thorough description useful to a reviewer


packages/frameworks/src/frameworks.ts
```diff
@@ -18,22 +18,20 @@ const { readdir, readFile, unlink } = promises;
 
 export const frameworks = [
   {
-    name: 'Blitz.js',
-    slug: 'blitzjs',
+    name: 'Blitz.js (Legacy)',
+    slug: 'legacy-blitzjs',
```
flybayer: changing the slug may cause problems for users who already have the blitz preset?

styfle: Correct, slugs should never change

packages/frameworks/src/frameworks.ts
```diff
@@ -24,16 +24,14 @@ export const frameworks = [
...
     website: 'https://blitzjs.com',
     envPrefix: 'NEXT_PUBLIC_',
     useRuntime: { src: 'package.json', use: '@vercel/next' },
     detectors: {
       every: [
         {
-          path: 'package.json',
-          matchContent:
-            '"(dev)?(d|D)ependencies":\\s*{[^}]*"blitz":\\s*".+?"[^}]*}',
+          path: 'blitz.config.(js|ts)',
```
styfle, line 34: Can you create a Blitz.js app without a config file?

packages/frameworks/src/frameworks.ts
```diff
@@ -24,16 +24,14 @@ export const frameworks = [
     logo: 'https://raw.githubusercontent.com/vercel/vercel/main/packages/frameworks/logos/blitz.svg',
     tagline: 'Blitz.js: The Fullstack React Framework',
     description:
-      'A brand new Blitz.js app - the result of running `npx blitz new`.',
+      'A brand new Blitz.js app - the result of running `npx blitz@0.45.4 new`.',
```
styfle, line 27: Lets revert the description since the version is irrelevant

packages/frameworks/src/frameworks.ts
flybayer: all legacy blitz apps have this config. With 2.0, this file no longer exists

leerob:
👋 There's a few outstanding comments here - are you interested in continuing with this PR?

v3dillon:
> 👋 There's a few outstanding comments here - are you interested in continuing with this PR?

Yup! Going to work on it today. Thanks for the reminder

packages/frameworks/src/frameworks.ts
v3dillon: Hey @styfle, We have to set the version otherwise it will installed the latest version of blitz which is `2.0.0-beta.17`.

packages/frameworks/src/frameworks.ts
v3dillon: @styfle Since this is for legacy blitz, we can keep the detection for `blitz.config`?

flybayer:
@styfle @leerob this is ready to go as far as we know

packages/frameworks/src/frameworks.ts
```diff
@@ -18,22 +18,20 @@ const { readdir, readFile, unlink } = promises;
 
 export const frameworks = [
   {
-    name: 'Blitz.js',
+    name: 'Blitz.js (Legacy)',
```
leerob, line 21: Should there be an entry for non-legacy then?

flybayer: @leerob no, it's just nextjs. Nothing special for blitz 2

leerob: Got it 🙏

PR #8419, opened 2022-08-18
Sometimes the project not have git infomations

fewensa:
### Related Issues

```
 + vercel --token *** --scope itering link --yes
Vercel CLI 28.1.0
Loading scopes…
Searching for existing projects…
> Connecting...
Error! Failed to link helix-bridge/helix-ui. Make sure there aren't any typos and that you have access to the repository if it's private.
```

https://github.com/helix-bridge/helix-ui/actions/runs/2879537072/attempts/2

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8433, opened 2022-08-20
[cli] prevent git connection during vc pull

EndangeredMassa:
When running `vc pull`, if the project is not yet linked, the logic behind `vc link` is called. When that happens, if there is no Git Provider on the project, the Git Provider Connection logic can run.

This is confusing for the user running `vc pull`. I think we should skip Git Provider Connection logic when the original CLI command was not `vc link`. This PR implements that.

### DIFF
--- a/packages/cli/src/util/create-git-meta.ts
+++ b/packages/cli/src/util/create-git-meta.ts
@@ -93,6 +93,11 @@ export function isDirty(directory: string, output: Output): Promise<boolean> {
 
 export async function parseGitConfig(configPath: string, output: Output) {
   try {
+    const configExists = await fs.pathExists(configPath);
+    if (!configExists) {
+      return;
+    }
+
     return ini.parse(await fs.readFile(configPath, 'utf-8'));
   } catch (err: unknown) {
     output.debug(`Error while parsing repo data: ${errorToString(err)}`);
--- a/packages/cli/src/util/link/add-git-connection.ts
+++ b/packages/cli/src/util/link/add-git-connection.ts
@@ -70,13 +70,18 @@ async function addSingleGitRemote(
     return 1;
   }
   const { org: parsedOrg, repo, provider } = repoInfo;
-  const alreadyLinked =
+
+  // The Project calls a Git Provider Connection a "link",
+  // which is not the same as linking a local repo to a Vercel Project
+  const alreadyConnected =
     project.link &&
     project.link.org === parsedOrg &&
     project.link.repo === repo &&
     project.link.type === provider;
-  if (alreadyLinked) {
-    client.output.debug('Project already linked. Skipping...');
+  if (alreadyConnected) {
+    client.output.debug(
+      `Project already connected to Git Provider: ${remoteUrl}. Skipping...`
+    );
     return;
   }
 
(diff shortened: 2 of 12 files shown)


packages/cli/src/util/create-git-meta.ts
```diff
@@ -93,6 +93,11 @@ export function isDirty(directory: string, output: Output): Promise<boolean> {
 
 export async function parseGitConfig(configPath: string, output: Output) {
   try {
+    const configExists = await fs.pathExists(configPath);
+    if (!configExists) {
+      return;
+    }
```
EndangeredMassa, line 99: This prevents the unhelpful error message (from line 103) from showing up when there's no `git.config` to even try to parse.

packages/cli/src/util/link/add-git-connection.ts
```diff
@@ -70,13 +70,18 @@ async function addSingleGitRemote(
...
     project.link &&
     project.link.org === parsedOrg &&
     project.link.repo === repo &&
     project.link.type === provider;
-  if (alreadyLinked) {
-    client.output.debug('Project already linked. Skipping...');
+  if (alreadyConnected) {
+    client.output.debug(
+      `Project already connected to Git Provider: ${remoteUrl}. Skipping...`
+    );
```
EndangeredMassa, line 84: The variable name and message here was confusing Project linking with Git Provider Connection.

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -258,18 +263,20 @@ export default async function setupAndLink(
...
-        project,
-        remoteUrls,
-        autoConfirm,
-        settings
-      );
-      if (typeof connectGit === 'number') {
-        return { status: 'error', exitCode: connectGit };
+    if (!skipGitProviderConnection) {
+      const remoteUrls = await getRemoteUrls(join(path, '.git/config'), output);
+      if (remoteUrls && !project.skipGitConnectDuringLink) {
```
EndangeredMassa, line 268: This logic was not respecting the `skipGitConnectDuringLink` project setting.

packages/cli/test/unit/commands/pull.test.ts
```diff
@@ -65,6 +67,68 @@ describe('pull', () => {
     await expect(exitCodePromise).resolves.toEqual(1);
   });
 
+  it('should link an unlinked project with a git config without connecing a git provider', async () => {
```
EndangeredMassa, line 70: This test got messy. Part of it is that we want to test that something doesn't happen, but the other part is that our API mocking makes it hard to do anything uncommon.

We may not really need this test. If we want to keep it, I could clean it up a bit more.

cb1kenobi:
After doing some thinking on this, I'm starting to think we should never connect the git provider when linking a project regardless of the command. I don't like some commands connecting and some commands not connecting. I prefer consistency. In the case of `--yes`, I would say that connecting the git provider is optional and thus should be skipped. So I think my vote would be to accept #8419 instead of this PR as is.

EndangeredMassa:
@cb1kenobi I agree that we probably want to never connect a git provider during linking. I added that topic to our weekly team meeting, but we can try to resolve this sooner in Slack.

EndangeredMassa:
Closing in favor of: https://github.com/vercel/vercel/pull/8450

PR #8450, opened 2022-08-23
[cli] revert git connect inside `vc link`

EndangeredMassa:
In https://github.com/vercel/vercel/pull/8290, functionality to connect projects to git providers was added as a step in `vercel link`. This caused problems with various commands that leverage linking functionality and caused confusion.

This PR reverts that behavior. Users can still `vc git connect` explicitly, but `vc link` (or any command that relies on linking) will no longer connect git providers at all.

### Related Issues

- partially reverts https://github.com/vercel/vercel/pull/8290
- supersedes https://github.com/vercel/vercel/pull/8419
- supersedes https://github.com/vercel/vercel/pull/8433


packages/cli/test/unit/commands/git.test.ts
```diff
@@ -39,11 +39,6 @@ describe('git', () => {
         await expect(client.stderr).toOutput('Found project');
         client.stdin.write('y\n');
 
-        await expect(client.stderr).toOutput(
-          'Do you want to connect "origin" to your Vercel project?'
-        );
-        client.stdin.write('n\n');
```
styfle, line 45: Why was this test changed? This is testing `vc git connect` right?

cb1kenobi (approved):

packages/cli/test/unit/commands/git.test.ts
EndangeredMassa: This does test `vc git connect`, which calls `ensureLink` (from the [command's `index.js`](https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/git/index.ts#L91) file), which was calling the git connection logic, which caused this prompt. The reverted logic keeps this within `vc git connect`'s domain, which does not have this prompt.

styfle: I was expecting `vc git connect` to ask about the remote if there was more than one in the `.git` directory

styfle (approved):

packages/cli/test/unit/commands/git.test.ts
EndangeredMassa: It still does that. In a different test, you can see that behavior being tested: https://github.com/vercel/vercel/blob/main/packages/cli/test/unit/commands/git.test.ts#L324-L327

unitof:
Could someone at Vercel update https://vercel.com/changelog/vercel-cli-v28 to mention this is no longer automatic, and maybe mention the handy new `vc git connect` instead?

I kept returning to that blog post every time doesn't prompt to connect Git when I `vc` a new project, assuming something must be funky with my local Git setup. Until just now when I finally tracked down this reversion—happy it's not me!

EndangeredMassa:
@unitof we updated [the changelog](https://vercel.com/changelog/vercel-cli-v28) to say that this was reverted. Sorry for the trouble!

PR #8451, opened 2022-08-23
[cli] add override support for vc build and vc dev

Ethan-Arrowood:
### Related Issues

Adds configuration override support to `vc build` and `vc dev` commands.

#### Tests

#### Code Review


packages/cli/src/commands/build.ts
```diff
@@ -313,7 +319,7 @@ async function doBuild(
     // Detect the Vercel Builders that will need to be invoked
     const detectedBuilders = await detectBuilders(files, pkg, {
       ...vercelConfig,
-      projectSettings: project.settings,
+      projectSettings: projectSettings,
```
TooTallNate:
```suggestion
      projectSettings,
```

packages/cli/src/commands/dev/dev.ts
```diff
@@ -70,19 +65,6 @@ export default async function dev(
 
     projectSettings = project;
 
-    if (project.devCommand) {
```
styfle, line 73: We don't need the project dev command anymore?

styfle: I see now, this was moved to `get devCommand()` in the dev server 👍

packages/cli/test/fixtures/unit/commands/build/project-settings-override/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "scripts": {
+    "build": "this will be ignored because of the Project-level settings"
```
styfle:
```suggestion
    "build": "mkdir -p output && echo 1 > output/index.txt"
```

How about a different value to prove it was overridden?

packages/cli/src/util/projects/project-settings.ts
```diff
@@ -51,3 +53,26 @@ export async function writeProjectSettings(
 export async function readProjectSettings(cwd: string) {
   return await getLinkFromDir<ProjectLinkAndSettings>(cwd);
 }
+
+export function pickOverrides(
+  vercelConfig: VercelConfig
+): PartialProjectSettings {
+  const overrides: PartialProjectSettings = {};
+  for (const prop of [
```
styfle, line 61: I've seen similar code like this in the API and now in the CLI. In a future PR, we should consider consolidating some of these into helpers so we can add new overrides in the future.


https://github.com/vercel/vercel/blob/f3ed2790072ca64a5e9d2d7dcca6b6af28cbba74/packages/cli/src/commands/deploy/index.ts#L512

https://github.com/vercel/vercel/blob/9963965e9abeed8fb6e6ad7d6c2ac7803d714942/packages/cli/src/util/link/setup-and-link.ts#L192

styfle (approved):

packages/cli/src/util/dev/server.ts
```diff
@@ -2203,10 +2233,21 @@ export default class DevServer {
...
 
+    if (devCommand === this.currentDevCommand) {
+      // `devCommand` has not changed, so don't restart frontend dev process
+      return;
+    }
+
+    this.currentDevCommand = devCommand;
+
     if (!devCommand) {
       return;
```
EndangeredMassa, line 2246: **suggestion (non-blocking):** move condition below killing of dev process

If we get here, the `devCommand` was changed to nothing. Should we move this condition down so that we can kill the `devProcess` first or does it not matter at that point?

EndangeredMassa (approved):

packages/cli/src/util/dev/server.ts
Ethan-Arrowood: I do not believe it matters at this point

PR #8447, opened 2022-08-23
[tests] Automatically update Next.js example periodically

styfle:
Automatically update Next.js example every 4 hours using `create-next-app`

PR #8452, opened 2022-08-23
[tests] Fix update-next script

styfle:
Follow up to #8447 since exec returns a buffer, not a string.

### DIFF
--- a/utils/update-next.js
+++ b/utils/update-next.js
@@ -2,16 +2,17 @@ const { execFileSync } = require('child_process');
 
 function exec(cmd, args, opts) {
   // eslint-disable-next-line no-console
-  console.log(`${cmd} ${args.join(' ')}`);
-  return execFileSync(cmd, args, opts);
+  console.log({ input: `${cmd} ${args.join(' ')}` });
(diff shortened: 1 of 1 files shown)


utils/update-next.js
```diff
@@ -9,7 +9,9 @@ function exec(cmd, args, opts) {
 module.exports = async ({ github, context }) => {
   const oldVersion = require('../examples/nextjs/package.json').dependencies
     .next;
-  const newVersion = exec('npm', ['view', 'next', 'dist-tags.latest']);
+  const newVersion = exec('npm', ['view', 'next', 'dist-tags.latest'])
+    .toString()
+    .trim();
   const branch = `next-${newVersion.replaceAll('.', '-')}`;
```
ijjk: Do we want to log the output from `exec` for `create-next-app` in case it fails at all so we can debug it easier?

styfle: Good point! Added in [8881207](https://github.com/vercel/vercel/pull/8452/commits/8881207c0e29606097c2cc9efb088e455ee3b6bf)

ijjk (approved):

PR #8456, opened 2022-08-24
[dev] rewrites from middleware should change path too

EndangeredMassa:
Middleware rewrites are supposed to change the path, but are currently not.

**Tests will fail until this is fixed in production as well. We will not merge until then.**

---

Blocked by:

- a similar change needs to be made in production middleware handling


cramforce (approved):

EndangeredMassa:
Turns out that this was correct behavior to start with.

PR #8457, opened 2022-08-24
[dev] allow middleware rewrites to change origin in `vc dev`

EndangeredMassa:
Middleware rewrites in `vc dev` were not respecting changes to the origin. This PR implements that to match prodution.

---

Paired with: @TooTallNate


packages/cli/src/util/dev/server.ts
```diff
@@ -1536,9 +1537,34 @@ export default class DevServer {
...
+                debug(`ProxyPass: ${rewritePath}`);
+                this.setResponseHeaders(res, requestId);
+                proxyPass(req, res, rewritePath, this, requestId);
+                return;
+              }
+            } else {
+              // Retain orginal pathname, but override query parameters from the rewrite
+              const rewriteUrlParsed = url.parse(beforeRewriteUrl);
+              rewriteUrlParsed.search = url.parse(rewritePath).search;
+              req.url = url.format(rewriteUrlParsed);
```
EndangeredMassa, line 1565: We're keeping the broken path writes in place because it matches production. This will be fixed in production, then in `vc dev`, here: https://github.com/vercel/vercel/pull/8456

packages/cli/test/dev/utils.js
```diff
@@ -98,7 +98,6 @@ function shouldSkip(name, versions) {
 
 function validateResponseHeaders(res, podId) {
   if (res.status < 500) {
-    expect(res.headers.get('server')).toEqual('Vercel');
```
EndangeredMassa: This assertion is no longer always true. We may want keep the original `server` header, though.

packages/cli/src/util/dev/server.ts
```diff
@@ -1566,9 +1567,31 @@ export default class DevServer {
 
             // Retain orginal pathname, but override query parameters from the rewrite
             const beforeRewriteUrl = req.url || '/';
-            const rewriteUrlParsed = url.parse(beforeRewriteUrl);
-            rewriteUrlParsed.search = url.parse(rewritePath).search;
-            req.url = url.format(rewriteUrlParsed);
+
+            if (isURL(rewritePath)) {
+              const rewriteUrlParsed = new URL(rewritePath);
+              const rewriteOrigin = `${rewriteUrlParsed.protocol}//${rewriteUrlParsed.host}`;
```
```suggestion
              const rewriteOrigin = rewriteUrlParsed.origin;
```

Can we use the origin prop?

packages/cli/src/util/dev/server.ts
```diff
@@ -1566,9 +1567,31 @@ export default class DevServer {
...
+
+              // `this.address` already has localhost normalized from ip4 and ip6 values
+              const devServerHost = new URL(this.address).host;
+              const isSameOrigin = devServerHost === rewriteUrlParsed.host;
```
styfle: `isSameOrigin` is not a great name because its comparing `host`, not `origin` (origin includes protocol and port)

packages/cli/src/util/dev/server.ts
```diff
@@ -1566,9 +1567,31 @@ export default class DevServer {
...
+
+              if (isSameOrigin) {
+                req.url = rewritePath.replace(rewriteOrigin, '');
```
styfle: Might be better to do a `rewritePath.slice(rewriteOrigin.length)` since you know the shape of the string already.

packages/cli/src/util/dev/server.ts
EndangeredMassa: I think `host` is the confusing part because it does include port, but `hostname` does not. This check is comparing hostname and port.

`origin` will include protocol, which may still be better

packages/cli/src/util/dev/server.ts
EndangeredMassa: Updated in [`0e9013a` (#8457)](https://github.com/vercel/vercel/pull/8457/commits/0e9013a125a62c331e088fb11c81b09d19e93b55)

packages/cli/src/util/dev/server.ts
EndangeredMassa: Updated in [`0e9013a` (#8457)](https://github.com/vercel/vercel/pull/8457/commits/0e9013a125a62c331e088fb11c81b09d19e93b55)

packages/cli/src/util/dev/server.ts
```diff
@@ -1566,9 +1567,31 @@ export default class DevServer {
...
-            req.url = url.format(rewriteUrlParsed);
+
+            if (isURL(rewritePath)) {
+              const rewriteUrlParsed = new URL(rewritePath);
+
+              // `this.address` already has localhost normalized from ip4 and ip6 values
+              const devServerOrigin = new URL(this.address).origin;
+              const isSameOrigin = devServerOrigin === rewriteUrlParsed.origin;
+
+              if (isSameOrigin) {
```
```suggestion
              const devServerParsed = new URL(this.address);
              if (rewriteUrlParsed.origin === devServerParsed.origin) {
```

Nit pick: remove a variable so its clear what is being compared and they aren't used later

packages/cli/src/util/dev/server.ts
```diff
@@ -1566,9 +1567,31 @@ export default class DevServer {
 
             // Retain orginal pathname, but override query parameters from the rewrite
```
styfle: This comment is the same on line 1589. Is that expected?

styfle:
We don't need to do it in this PR, but I think a future enhancement is to save the parsed `this.address` on the class so we only need to do it once. cc @TooTallNate @EndangeredMassa

packages/cli/src/util/dev/server.ts
EndangeredMassa: I find both roughly equal in clarity. I'd rather come through and update these references to `this.address` to be a new property that's already parsed or update `this.address` itself to already be parsed, as you mentioned below. That would make this clearer.

packages/cli/test/dev/utils.js
TooTallNate: I believe it's true in production. I think we should look into maintaining that behavior for dev.

packages/cli/src/util/dev/server.ts
EndangeredMassa: Nope! Removed in [`9a511e1` (#8457)](https://github.com/vercel/vercel/pull/8457/commits/9a511e1191a6bdfe9fd312bc1655765557590ec0)

packages/cli/test/dev/utils.js
styfle: Why is it not true in dev? Is it not added for middleware responses?

TooTallNate: It is true for a lot of cases but I believe the case that needs to be looked into is when vc dev does a `proxyPass()`, since then the destination headers will take precedence over the ones already set.

I'm guessing that this validation logic was not being used in tests that do a proxy pass until this PR.

packages/cli/src/util/dev/server.ts
cb1kenobi: I do like @styfle's recommendation. It's a bit easier to read as it has less "origins" that my mind has to process.

packages/cli/test/dev/utils.js
styfle: Perhaps we should add this line back and remove the `validateResponseHeaders()` from the proxy pass test.

packages/cli/src/util/dev/server.ts
EndangeredMassa: Updated [`4bd0b08` (#8457)](https://github.com/vercel/vercel/pull/8457/commits/4bd0b08ace0308e37aa20b5573f0367a7cd6b9a2)

packages/cli/test/dev/utils.js
EndangeredMassa: When the proxy proxies to another origin, it was setting the response header from that target.

I figured out how to override this, which matches what happens in production. Updated [`ae53c87` (#8457)](https://github.com/vercel/vercel/pull/8457/commits/ae53c87890768b8b760c2dd690da5b38aacbd3cd)

cb1kenobi (approved):
LGTM

styfle (approved):
Thanks!

PR #8464, opened 2022-08-25
[tests] Add required labels

styfle:
This action will fail if the required labels are not applied to a PR.

We allow multiple `area` labels and only a single `semver` label.

- Example failure: https://github.com/vercel/vercel/runs/8024949804?check_suite_focus=true#step:2:19
- Example success: https://github.com/vercel/vercel/runs/8026212031?check_suite_focus=true#step:2:19

### DIFF
--- a/.github/workflows/pull-request-labels.yml
+++ b/.github/workflows/pull-request-labels.yml
@@ -1,0 +1,16 @@
+name: Pull Request Labels
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


EndangeredMassa (approved):
Looks good to me!

ijjk (approved):

cb1kenobi (approved):

PR #8467, opened 2022-08-25
[node] Added 'cause' to error message

cb1kenobi:
When running `vc dev` for a project using `experimental-edge` runtime and a fetch-related error occurs, the message is a little vague:

```
Unhandled rejection: fetch failed
```

In this case, the fetch promise (defined in @edge-runtime/primitives) is being rejected with a `TypeError: fetch failed` that has a `cause` property containing the actual `response.error` message. The `response.error` is the actual error and is never reported to the user.

This PR simply appends the `cause` message, if exists, to the fetch failure message.

```
Unhandled rejection: fetch failed: Error: getaddrinfo ENOTFOUND undefined
```

#### Tests

#### Code Review


styfle (approved):
Cool, looks like `error.cause` was added in Node.js 16 👍

cb1kenobi:
@styfle ref https://github.com/nodejs/undici/blob/21fdda54ca290456083874ab0727e3c343ee8e45/lib/fetch/index.js#L199

TooTallNate (approved):

packages/node/src/dev-server.ts
```diff
@@ -212,7 +212,8 @@ async function compileUserCode(
         } catch (error) {
           // we can't easily show a meaningful stack trace
           // so, stick to just the error message for now
-          event.respondWith(new Response(error.message, {
+          const msg = error.cause ? (error.message + ': ' + (error.cause.message || error.cause)) : error.message;
```
EndangeredMassa:
```suggestion
          const msg = error.cause
            ? (error.message + ': ' + (error.cause.message || error.cause))
            : error.message;
```

I think we should make this a bit clearer with line breaks or breaking apart the ternary.

EndangeredMassa (approved):

Ethan-Arrowood (approved):

PR #8485, opened 2022-08-30
[cli] Use Builder importing logic from `vc build` in `vc dev`

TooTallNate:
This makes `vc dev` utilize the same Builder installing and importing logic that `vc build` uses. So now the Builders will be installed locally on a per-project basis, rather than being installed to a shared global installation directory.


cb1kenobi:
The builder cache _was_ the shared global installation directory, correct?

TooTallNate:
@cb1kenobi Correct

cb1kenobi (approved):
CR'd and manually tested. Verified with the `vercel-deno` builder that the builder is in fact downloaded into the `./.vercel/builders` directory.

It's worth noting that I did run into an issue testing using the Deno builder. It works for `vc deploy`, but `vc dev` errors out with `Error: Failed to complete request to /api/hello: Error: socket hang up`. Since this error occurs without the changes in this PR, the issue is likely in the Deno builder and outside the scope of this PR.

packages/cli/src/util/dev/builder.ts
```diff
@@ -198,16 +198,12 @@ export async function executeBuild(
   }
 
   // Sort out build result to builder v2 shape
-  if (!builder.version || builder.version === 1) {
+  if (!builder.version || (builder as any).version === 1) {
```
EndangeredMassa, line 201: **question (non-blocking):** is `builder.version === 1` even possible? Should we augment the types?

packages/cli/src/util/build/static-builder.ts
```diff
@@ -39,3 +40,18 @@ export const build: BuildV2 = async ({ entrypoint, files, config }) => {
...
+
+export const shouldServe: ShouldServe = _opts => {
+  const opts = { ..._opts };
+  const {
+    config: { zeroConfig, outputDirectory },
+  } = opts;
+
+  // Add the output directory prefix
+  if (zeroConfig && outputDirectory) {
+    opts.entrypoint = `${outputDirectory}/${opts.entrypoint}`;
```
styfle, line 52: Should this use `path.join()` so it works with windows?

cb1kenobi: I'd say using `path.join()` is a best practice, but it's not necessary with modern Node.js versions:

```
> const p = process.cwd().replaceAll('\\', '/')
undefined

> p
'C:/Users/chris/projects/vercel/packages'

> fs.readdirSync(p)
[
  'build-utils',   'cli',
  'client',        'edge',
  'frameworks',    'fs-detectors',
  'go',            'hydrogen',
  'next',          'node',
  'node-bridge',   'python',
  'redwood',       'remix',
  'routing-utils', 'ruby',
  'static-build',  'static-config'
]
```

packages/cli/src/util/dev/builder.ts
```diff
@@ -198,16 +198,12 @@ export async function executeBuild(
...
     // `BuilderOutputs` map was returned (Now Builder v1 behavior)
     result = {
       output: buildResultOrOutputs as BuilderOutputs,
       routes: [],
       watch: [],
-      distPath:
```
styfle, line 207: Are you sure this isn't used anymore?

packages/cli/src/util/dev/builder.ts
```diff
@@ -439,6 +437,8 @@ export async function getBuildMatches(
     for (const file of files) {
       src = relative(cwd, file);
 
+      const entrypoint = mapToEntrypoint.get(src) || src;
```
styfle, line 440: Is this just a formatting change or is there something else depending on it now?

styfle (approved):

packages/cli/src/util/dev/builder.ts
TooTallNate: It's not strictly possible, but Builders are inherently "untrusted code" we should be defensive about it.

packages/cli/src/util/dev/builder-cache.ts
```diff
@@ -1,412 +0,0 @@
...
-export function filterPackage(
-  builderSpec: string,
-  buildersPkg: PackageJson,
-  cliPkg: Partial<PackageJson>
-) {
-  if (builderSpec in localBuilders) return false;
-  const parsed = npa(builderSpec);
-  const parsedVersion = parseVersionSafe(parsed.rawSpec);
-
-  // Skip install of Runtimes that are part of Vercel CLI's `dependencies`
```
styfle, line 108: I assume the answer is yes, but I just want to make sure: are we still skipping these for `vc build`?

packages/cli/src/util/dev/builder.ts
TooTallNate: It was a change to make `entrypoint` be calculated _before_ the update to `src` to strip the output directory. It was necessary to do it like this to make `@vercel/static` work without changes.

packages/cli/src/util/dev/builder-cache.ts
TooTallNate: Yes, the `import-builders.test.ts` file tests for this explicitly.

packages/cli/test/unit/util/dev/builder-cache.test.ts
```diff
@@ -1,235 +0,0 @@
...
-    expect(result).toEqual(true);
-  });
-
-  it('should detect latest releases', () => {
-    const parsed = npa('@vercel/node');
-    const result = isBundledBuilder(parsed, cliPkg);
-    expect(result).toEqual(true);
-  });
-
-  it('should detect "latest" tagged releases', () => {
```
styfle, line 212: Do we have similar tests somewhere else?

If not, we should convert these test cases to run against the new code.

packages/cli/src/util/dev/builder-cache.ts
styfle: I'm looking at the tests and it doesn't seem to check if the "install" was skipped. Perhaps we need to check `client.stderr` for each one to ensure it wasn't installed.

And we're missing test cases for `@latest`, etc (see my other comment)

packages/cli/test/unit/util/dev/builder-cache.test.ts
TooTallNate: Here's a follow-up that adds a few more cases that were being covered in these previous tests https://github.com/vercel/vercel/pull/8554.

PR #8496, opened 2022-08-31
[hydrogen] Polyfill Oxygen global in Hydrogen

mperreux:
### Related Issues

Description stolen from https://github.com/vercel/vercel/pull/8413
It's currently not possible to access environment variables in Hydrogen projects deployed to Vercel. The reason is that Hydrogen currently relies on a global [Oxygen.env that needs to be polyfilled](https://shopify.dev/custom-storefronts/hydrogen/framework/environment-variables#private-variables-in-production) in the platform. This PR adds the global object using process.env, which is [mentioned in the docs](https://vercel.com/docs/concepts/functions/edge-functions/edge-functions-api#environment-variables) as the way to access environment variables in Edge Functions.

#### Tests

#### Code Review

### DIFF
--- a/packages/hydrogen/edge-entry.js
+++ b/packages/hydrogen/edge-entry.js
@@ -3,7 +3,15 @@ import indexTemplate from '__RELATIVE__/dist/client/index.html?raw';
 
 // ReadableStream is bugged in Vercel Edge, overwrite with polyfill
 import { ReadableStream } from 'web-streams-polyfill/ponyfill';
-Object.assign(globalThis, { ReadableStream });
+Object.assign(globalThis, {
+  // ReadableStream is bugged in Vercel Edge, overwrite with polyfill
+  ReadableStream,
+
+  // Hydrogen exposes env vars through `Oxygen.env`
+  Oxygen: {
+    env: process.env
+  }
+});
 
 export default (request, event) =>
   handleRequest(request, {


packages/hydrogen/edge-entry.js
```diff
@@ -3,7 +3,15 @@ import indexTemplate from '__RELATIVE__/dist/client/index.html?raw';
...
 // ReadableStream is bugged in Vercel Edge, overwrite with polyfill
 import { ReadableStream } from 'web-streams-polyfill/ponyfill';
-Object.assign(globalThis, { ReadableStream });
+Object.assign(globalThis, {
+  // ReadableStream is bugged in Vercel Edge, overwrite with polyfill
+  ReadableStream,
+
+  // Hydrogen exposes env vars through `Oxygen.env`
+  Oxygen: {
+    env: process.env
```
TooTallNate, line 12: Unfortunately I actually think that this won't work the way you're expecting. This is because for Edge functions, `process.env` usage is statically analyzed to determine which variables to include in the edge function bundle.

To make it work with `Oxygen.env` usage then we would have to do similar static analysis in `@vercel/hydrogen` and populate the `envVarsInUse` array in the `EdgeFunction` constructor. Perhaps Hydrogen could provide that information if it is already doing static analysis during the build process?

TooTallNate:
Also, any reason in particular that you created a new PR instead of the other one?

ryanleichty:
Any idea when this might get merged/resolved? Would love to use Vercel for hosting a current project 🙌🏻

packages/hydrogen/edge-entry.js
frandiox: We're not doing any static analysis in the build process in Hydrogen, afaik 🤔 
Perhaps you can replace `Oxygen.env.XYZ` matches in `@vercel/hydrogen`?

Also, [we are moving to using `Hydrogen.env` instead of `Oxygen.env`](https://github.com/Shopify/hydrogen/pull/2150) to avoid confusion when deploying to other platforms, so something to keep in mind.

TooTallNate:
@ryanleichty The current workaround would be to just use `process.env` instead.

EndangeredMassa:
Closing this for now. We can revisit this fresh if someone wants to kick it off.

PR #8497, opened 2022-09-01
[fs-detectors] optimize readdir for framework detection

agadzik:
Adding the ability to explicitly set the `hasPath` and `isFile` caches via `readdir`

**Why is this needed?**

For framework detection, this allows us to fetch the contents of a directory (the root directory) of a git repo, and make assumptions about the existence of files,

```ts
import frameworks from "@vercel/frameworks";

// we can make the set of all file paths needed for framework detection like this
const setOfPaths = new Set(
  frameworks.frameworks
    .flatMap((f) => [
      ...(f.detectors?.every ?? []),
      ...(f.detectors?.some ?? []),
    ])
    .map((d) => d.path)
);

// then we can read the contents of the root directory of a git repo via the `DetectorFilesystem`
fs.readdir('./', { potentialFiles: [...setOfPaths] });

// The filesystem has been fully primed - any network calls from here will only happen if needed.
// This works because the logic of `detectFramework` calls `hasPath` and then `isFile` before 
// fetching file contents.  Since the filesystem knows this information, it doesn't have to make 
// any additional network calls 🙌 
const framework = detectFramework(fs);
```

### Related Issues

Related to [HIT-57](https://linear.app/vercel/issue/HIT-57/monorepo-detection-api-prevent-rate-limits)

#### Tests

#### Code Review


packages/fs-detectors/test/unit.framework-detector.test.ts
```diff
@@ -164,11 +164,23 @@ describe('DetectorFilesystem', () => {
     const files = {};
     const fs = new VirtualFilesystem(files);
 
-    fs.writeFile('file.txt', 'Hello World');
+    fs.writeFile('file.txt', true, 'Hello World');
```
styfle: Lets use an object instead of multiple params.

Also, can we follow the Node.js `fs.writeFile` API here (not sure if there is an equivalent)

agadzik: Sounds good!

I initially followed the nodejs `writeFile` API here, but will add an `options` object as the third parameter 🙌 

image

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,13 @@ export abstract class DetectorFilesystem {
   /**
    * Writes a file to the filesystem cache.
    * @param name the name of the file to write
-   * @param content The content of the file
+   * @param exists true if the files exists, false otherwise
+   * @param content contents of the file
    */
-  public writeFile(name: string, content?: string): void {
-    if (content)
+  public writeFile(name: string, exists?: boolean, content?: string): void {
```
styfle: In what scenario would you call `writeFile('./somefile')` without any content?

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,13 @@ export abstract class DetectorFilesystem {
   /**
    * Writes a file to the filesystem cache.
    * @param name the name of the file to write
-   * @param content The content of the file
+   * @param exists true if the files exists, false otherwise
```
styfle: The comment here needs updating because I didn't understand the purpose until reading the PR description. Also it doesn't mention what happens in the undefined case.

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,13 @@ export abstract class DetectorFilesystem {
...
+   * @param content contents of the file
    */
-  public writeFile(name: string, content?: string): void {
-    if (content)
+  public writeFile(name: string, exists?: boolean, content?: string): void {
+    if (content !== undefined && exists)
       this.readFileCache.set(name, Promise.resolve(Buffer.from(content)));
-    this.fileCache.set(name, Promise.resolve(true));
-    this.pathCache.set(name, Promise.resolve(true));
+    if (exists !== undefined) this.fileCache.set(name, Promise.resolve(exists));
```
chloetedder: could these go in one if block?

```
if (exists !== undefined) {
  this.fileCache.set(name, Promise.resolve(exists));
  this.pathCache.set(name, Promise.resolve(exists));
}
```

packages/fs-detectors/src/detectors/filesystem.ts
agadzik: In my PR description I tried to be explicit on usage.  In this case, we can prime the `hasPath` and `isFile` caches like this (think framework detection)

```ts
// we can read the contents of the root directory of a git repo via the `DetectorFilesystem`
const files = fs.readdir('./');
for (const file of files) {
  if (file.type === 'file') {
    // we know this file exists, mark it as so on the filesystem
    // this primes the `hasPath` and `isFile` caches
    fs.writeFile(file.name, '');
  }
}
```

agadzik: Maybe it's better to mark `content` as a required parameter here 🤔 

**Edit:** Yeah, I like this approach better for the API

packages/fs-detectors/src/detectors/filesystem.ts
agadzik: Went in another direction with the writeFile signature - these were removed 🥳

packages/fs-detectors/src/detectors/filesystem.ts
agadzik: Refactored to make `content` a required parameter again, this comment should still hold 🤞

packages/fs-detectors/src/detectors/filesystem.ts
agadzik: @styfle Made `content` a required parameter again - thank you for the suggestions!  Feel free to mark this convo as resolved if you are satisfied.

cb1kenobi: I understand this function is not async, however since it shares the same name as Node's `fs.writeFile()`, it might confuse people thinking this is Node's writeFile or that this function is async. What if someday we want to add something async, but this function name is already used for synchronous logic. I'd consider renaming it to something else like `addFile()` or appending "Sync" to the name.

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,22 @@ export abstract class DetectorFilesystem {
   /**
    * Writes a file to the filesystem cache.
    * @param name the name of the file to write
-   * @param content The content of the file
+   * @param content contents of the file
+   * @param options.exists true if the files exists, false otherwise
```
styfle: Is there a default value? What about undefined?

packages/fs-detectors/src/detectors/filesystem.ts
styfle: @cb1kenobi Its the same as `import fs from 'fs/promises'`. I think the name is fine.

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,22 @@ export abstract class DetectorFilesystem {
...
+  ): void {
+    if (options?.exists !== false) {
+      this.readFileCache.set(
+        name,
+        Promise.resolve(Buffer.from(String(content)))
+      );
+    }
+
+    this.fileCache.set(name, Promise.resolve(options?.exists ?? true));
+    this.pathCache.set(name, Promise.resolve(options?.exists ?? true));
```
styfle:
```suggestion
    const exists = typeof options?.exists === 'boolean' ? options.exists : true
    if (exists) {
      this.readFileCache.set(
        name,
        Promise.resolve(Buffer.from(String(content)))
      );
    }

    this.fileCache.set(name, Promise.resolve(exists));
    this.pathCache.set(name, Promise.resolve(exists));
```

packages/fs-detectors/test/unit.framework-detector.test.ts
```diff
@@ -169,6 +169,18 @@ describe('DetectorFilesystem', () => {
     expect(await fs.readFile('file.txt')).toEqual(Buffer.from('Hello World'));
     expect(await fs.hasPath('file.txt')).toBe(true);
     expect(await fs.isFile('file.txt')).toBe(true);
+
+    fs.writeFile('file2.txt', '', { exists: false });
```
styfle: How about adding a test with `{ exists: true }`

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,23 @@ export abstract class DetectorFilesystem {
...
+  public writeFile(
+    name: string,
+    content: string,
+    options?: { exists?: boolean }
+  ): void {
+    const exists = typeof options?.exists === 'boolean' ? options.exists : true
+    if (exists) {
+      this.readFileCache.set(
+        name,
+        Promise.resolve(Buffer.from(String(content)))
```
styfle:
```suggestion
        Promise.resolve(Buffer.from(content))
```

Since `content` must be a string, no need to convert

styfle (commented):
Something about this `writeFile()` API feels wrong.

Why not modify `readdir()` on line 80 to set the existence flag instead?

agadzik:
> Something about this `writeFile()` API feels wrong.
> 
> 
> 
> Why not modify `readdir()` on line 80 to set the existence flag instead?

That's a good catch and agreed that it's a better way of solving half of this optimization. 

We still need a way to to take the set of know filepaths needed for framework detection, filter by the known filepaths from `readdir` and mark the fs caches appropriately. 

In my pseudo code above, it's the second call to `writeFile` we still need to solve for. 

Maybe the name of the function is inappropriate for what it's actually doing? `writeFile` seemed good at the time 😂. 

Another option,

* we could pass in an `options` object into `readdir` and have a prop for this use case. Something like `options.knownFiles`.

Thoughts?

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -96,12 +96,23 @@ export abstract class DetectorFilesystem {
...
+   * @param options.exists optional. undefined or true (default) if the files exists, false otherwise
    */
-  public writeFile(name: string, content?: string): void {
-    if (content)
-      this.readFileCache.set(name, Promise.resolve(Buffer.from(content)));
-    this.fileCache.set(name, Promise.resolve(true));
-    this.pathCache.set(name, Promise.resolve(true));
+  public writeFile(
+    name: string,
+    content: string,
```
chloetedder: Now `content` is not an optional parameter. I am wondering if that breaks any existing uses of `writeFile` 🤔

styfle:
As we discussed in [slack](https://vercel.slack.com/archives/C03C26D5QTD/p1662492172905849?thread_ts=1662472088.848189&cid=C03C26D5QTD):

> remove writeFile all together, and just make changes to readdir

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -76,12 +76,39 @@ export abstract class DetectorFilesystem {
...
+            directoryContent.find(file => file.name === path) === undefined
+        );
+        for (const filePath of filesThatDoNotExist) {
+          // we know this file does not exist, mark it as so on the filesystem
+          this.fileCache.set(filePath, Promise.resolve(false));
+          this.pathCache.set(filePath, Promise.resolve(false));
+        }
+      }
     }
     return p;
```
styfle, line 119: I think you can return `directoryContent` here now that its already awaited above

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -92,16 +119,4 @@ export abstract class DetectorFilesystem {
   public chdir = (name: string): DetectorFilesystem => {
     return this._chdir(name);
   };
-
-  /**
-   * Writes a file to the filesystem cache.
-   * @param name the name of the file to write
```
chloetedder, line 98: Do you want to delete the whole writeFile?

packages/fs-detectors/src/detectors/filesystem.ts
styfle: Oh nevermind - its only awaited in the if statement

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -76,12 +76,39 @@ export abstract class DetectorFilesystem {
...
+        if (file.type === 'file') {
+          // we know this file exists, mark it as so on the filesystem
+          this.fileCache.set(file.name, Promise.resolve(true));
+          this.pathCache.set(file.name, Promise.resolve(true));
+        }
+      }
+
+      if (options?.potentialFiles) {
+        // calculate the set of paths that truly do not exist
+        const filesThatDoNotExist = options?.potentialFiles.filter(
```
styfle:
```suggestion
        const filesThatDoNotExist = options.potentialFiles.filter(
```

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -76,12 +76,39 @@ export abstract class DetectorFilesystem {
...
+          path =>
+            directoryContent.find(file => file.name === path) === undefined
```
styfle: Right now, this algorithm is O(N×M). How big are we expecting `potentialFiles` and `directoryContent` to be? You can get it down to O(N) if you use a `Set<string>`

packages/fs-detectors/src/detectors/filesystem.ts
chloetedder: > remove writeFile all together, and just make changes to readdir

I see

packages/fs-detectors/test/unit.framework-detector.test.ts
```diff
@@ -135,11 +135,18 @@ describe('DetectorFilesystem', () => {
     };
 
     const fs = new VirtualFilesystem(files);
+    const hasPathSpy = jest.spyOn(fs, '_hasPath');
 
-    expect(await fs.readdir('/')).toEqual([
+    expect(await fs.readdir('/', { potentialFiles: ['config.rb'] })).toEqual([
```
styfle, line 140: What about adding `potentialFiles` with `packages/app1/package.json`? I think we need more test coverage here

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -1,3 +1,7 @@
+import path from 'path';
+
+const posixPath = path.posix;
```
styfle:
```suggestion
import { posix: posixPath } from 'path';
```

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -76,12 +80,43 @@ export abstract class DetectorFilesystem {
 
   /**
    * Returns a list of Stat objects from the current working directory.
+   * @param name The name of the directory to read.
+   * @param options.potentialFiles optional. Array of potential file paths.  If provided, these paths will be used to mark the filesystem caches as existing or not existing.
```
styfle:
```suggestion
   * @param options.potentialFiles optional. Array of potential file names (relative to the path).  If provided, these will be used to mark the filesystem caches as existing or not existing.
```

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -76,12 +80,43 @@ export abstract class DetectorFilesystem {
 
   /**
    * Returns a list of Stat objects from the current working directory.
+   * @param name The name of the directory to read.
```
styfle: Should the be `path` instead of `name`?

Or perhaps `dirPath`.

packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -111,7 +109,7 @@ export abstract class DetectorFilesystem {
         );
         for (const filePath of filesThatDoNotExist) {
           const fullFilePath =
-            name === '/' ? filePath : posixPath.join(name, filePath);
+            path === '/' ? filePath : posixPath.join(path, filePath);
```
styfle: Lets use `dirPath` because line 108 shadows `path` and its super confusing

styfle (approved):
Thanks!

Ethan-Arrowood (approved):

PR #8499, opened 2022-09-01
[next] Revert nft to 0.21.0

styfle:
In Next.js 12, we introduce `outputFileTracing` as the default behavior. This moved `@vercel/nft` from `@vercel/next` into `next` itself so users can upgrade or downgrade Next.js if there is a bug in `@vercel/nft`.

Unfortunately, some users are setting `outputFileTracing: false` which deopts tracing back from `next` to `@vercel/next`. So we should pin `@vercel/nft` here and never upgrade in case of any bugs.

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -45,7 +45,7 @@
     "@types/text-table": "0.2.1",
     "@types/webpack-sources": "3.2.0",
     "@vercel/build-utils": "5.4.0",
-    "@vercel/nft": "0.22.0",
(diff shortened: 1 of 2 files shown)


ijjk (approved):

cb1kenobi (approved):
Sounds good.

Do you think someday we could remove nft from @vercel/next?

styfle:
@cb1kenobi No because of the exact reason described above. Its needed for backwards compat

PR #8500, opened 2022-09-01
[node] update middleware-no-response behavior to match production changes

EndangeredMassa:
Production was updated to allow middleware without responses to pass through. This PR makes `vc dev` behave the same way. Tests were updated.


cb1kenobi (approved):

styfle (approved):

styfle (commented):
Looks like this test is failing now:

```
[vercel dev] throws an error when an edge function has no response
```

PR #8503, opened 2022-09-02
[build-utils] Support empty directory entries in `createZip()`

TooTallNate:
Adds support for creating empty directories inside of a Lambda zip file.

Some refactoring was also done so I suggest reviewing with [whitespace ignored](https://github.com/vercel/vercel/pull/8503/files?w=1).


TooTallNate:
My use-case for this is no longer necessary, so I'm gonna close.

PR #8505, opened 2022-09-02
[frameworks] Update Next.js Logos

chibicode:
Update them to match the new logos. See: https://vercel.com/design/brands#next-js

Screen Shot 2022-09-02 at 10 43 52 AM

PR #8508, opened 2022-09-02
[frameworks] Remove dark mode nextjs logo

styfle:
This is a follow up to PR #8505 because the logos are now identical for dark/light mode and therefore we only need one logo.


styfle:
Closing since they are not identical 😄

PR #8529, opened 2022-09-07
[tests] set retry count to 0 when running tests locally

EndangeredMassa:
When running tests locally that fail to make a fetch request, the retries add a lot of noise to debugging. This PR sets those retry counts to `0` locally, but keeps them at their current value for CI.

### DIFF
--- a/packages/cli/test/dev/utils.js
+++ b/packages/cli/test/dev/utils.js
@@ -13,6 +13,7 @@ const {
 
 jest.setTimeout(6 * 60 * 1000);
 
+const isCI = !!process.env.CI;
 const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
 const isCanary = () => getDistTag(cliVersion) === 'canary';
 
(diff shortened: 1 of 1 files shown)


packages/cli/test/dev/utils.js
```diff
@@ -52,7 +53,7 @@ function fetchWithRetry(url, opts = {}) {
       return res;
     },
     {
-      retries: opts.retries || 3,
+      retries: opts.retries ?? 3,
```
EndangeredMassa, line 56: Allows `0` to return from this expression

packages/cli/test/dev/utils.js
```diff
@@ -150,9 +151,9 @@ async function testPath(
   fetchOpts = {}
 ) {
   const opts = {
+    retries: isCI ? 5 : 0,
```
EndangeredMassa, line 154: Moving this allows calls to `testPath` to specifically override the retry count

Ethan-Arrowood (approved):

styfle (approved):

cb1kenobi (approved):
LGTM

PR #8530, opened 2022-09-07
[next] Use posix path separator to join routes and overrides paths

cb1kenobi:
Use posix path separators for routes (static pages, lambdas, etc) instead of the native platform path separator (e.g. backslash on Windows).

#### Tests

#### Code Review


ijjk (approved):

styfle (approved):
Looks good. How about a unit test that runs on windows to verify the output is correct now?

TooTallNate (approved):
LGTM but ya, would be good to add a `vc build` test to ensure `routes` in `config.json` uses forward slashes.

packages/next/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "node build.js",
     "build-dev": "node build.js --dev",
     "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "yarn test test/build.test.ts  test/unit/",
```
styfle, line 11: Was this intentional?

styfle (changes requested):
I think you either need to move `test/build.test.ts` to `test/unit/build.test.ts` or change the script back

packages/next/package.json
cb1kenobi: Yes, 100%.

cb1kenobi:
> I think you either need to move `test/build.test.ts` to `test/unit/build.test.ts` or change the script back

I just checked again and unless I'm completely missing something, `test/build.test.ts` does not exist.

styfle (approved):
Hmm, you're right. Looks good 👍

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
```diff
@@ -0,0 +1,184 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@next/env@12.3.1-canary.0":
```
ijjk: Should this lockfile be committed? We don't commit it for other fixtures to continue testing new canaries.

packages/next/test/unit/fixtures/01-normalize-routes/vercel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "version": 2,
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@vercel/next"
+    }
+  ]
+}
```
styfle: Lets delete this file since its not needed, right?

packages/next/test/unit/fixtures/01-normalize-routes/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies": {
+    "next": "canary",
+    "react": "experimental",
+    "react-dom": "experimental"
+  },
+  "scripts": {
+    "vercel-build": "next build"
```
styfle:
```suggestion
    "build": "next build"
```

styfle (changes requested):
A couple clean up things

packages/next/test/unit/utils.test.js
```diff
@@ -8,8 +8,8 @@ const {
 const { FileRef } = require('@vercel/build-utils');
 
 describe('getNextConfig', () => {
-  const workPath = path.join(__dirname, 'fixtures');
-  const entryPath = path.join(__dirname, 'fixtures', 'entry');
+  const workPath = path.join(__dirname, 'fixtures', '00-config');
```
styfle, line 11: Was this intended?

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
cb1kenobi: Removing, 1 sec

packages/next/test/unit/utils.test.js
cb1kenobi: Yes. The fixtures directory seemed to contain a single fixture for testing the config stuff and since I needed to add a fixture for this PR, I moved the config stuff into a subdir and had to update the next config test. Now we can use the next/test/unit/fixtures dir for other unit test fixtures.

packages/next/test/unit/fixtures/01-normalize-routes/package.json
cb1kenobi: Happy to change it, but does it matter? I probably copied it from here: https://github.com/vercel/vercel/blob/623e43f865e927406200849aa3d7b6a9e9c2dcfc/packages/next/test/fixtures/11-export-clean-urls/package.json.

styfle: `vercel-build` is only needed if you already have a `build` command defined and want to override it.

In fact, you can even override `buildCommand` in vercel.json now so the `vercel-build` script is even less useful.

packages/next/test/unit/fixtures/01-normalize-routes/vercel.json
cb1kenobi: Correct, removing

styfle (approved):

packages/next/test/unit/utils.test.js
styfle: I see now, I was confused because this was `utils.test.js` but it was really testing one of the fixtures.

We can probably do some more clean up here but that can wait for a future PR 🙂

PR #8531, opened 2022-09-07
[tests] Skip tests that won't run on Windows due to 'mkdir -p' and symlinks

cb1kenobi:
Found a few tests that won't run on Windows. The project settings override test runs `mkdir -p` which errors. The other two build tests fail because they try to create symlinks.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/test/unit/commands/build.test.ts
+++ b/packages/cli/test/unit/commands/build.test.ts
@@ -895,6 +895,12 @@ describe('build', () => {
   });
 
   it('should apply project settings overrides from "vercel.json"', async () => {
+    if (process.platform === 'win32') {
+      // this test runs a build command with `mkdir -p` which is unsupported on Windows
+      console.log('Skipping test on Windows');
+      return;
+    }
+
     const cwd = fixture('project-settings-override');
     const output = join(cwd, '.vercel/output');
     try {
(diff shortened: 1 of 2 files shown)


EndangeredMassa:
These tests run in [CI on Windows](https://github.com/vercel/vercel/runs/8235943942?check_suite_focus=true). What's the problem, specifically?

cb1kenobi:
For the project settings override test (in cli), the unit test fails locally on a Windows machine because `yarn run build` exits with code `1` instead of `0`. It's trying to run `mkdir -p output && echo 3 > output/index.txt`. It seems to be fine when you run the tests from a clean fresh git clone (e.g. CI). However, subsequent unit test runs error out, possibly due to the test fixture directory not being clean.

```
A subdirectory or file -p already exists.
Error occurred while processing: -p.
A subdirectory or file output already exists.
Error occurred while processing: output.
```

If I check the filesystem, then there's a directory called `-p`. Windows aliases `mkdir` to `md`, but apparently it doesn't support the `-p` flag and proceeds to create a directory called `-p`. In any case, I still think the test is not Windows friendly.

The `importBuilders()` tests fail because of symlinks:

```
EPERM: operation not permitted, symlink '..\@vercel\build-utils' -> 'C:\Users\chris\AppData\Local\Temp\720c7e57\.vercel\builders\node_modules\@now\build-utils'
```

styfle (approved):

EndangeredMassa:
Ah, so these happen to pass in CI, but they aren't really clean tests and they fail locally. Got it.

EndangeredMassa (approved):

cb1kenobi:
@EndangeredMassa and the whole `mkdir -p` thing. :)

PR #7881, opened 2022-05-26
[build-utils][go][next][redwood][static-build] Fix path delimiter on windows

styfle:
This PR fixes an issue where the path delimiter was incorrect for windows and caused yarn (and other global CLIs) to no longer be found.

https://github.com/vercel/vercel/runs/6602572000?check_suite_focus=true#step:13:1357

PR #8532, opened 2022-09-07
[build-utils][cli][go][node][ruby][static-build] Explicitly set PATH when copying env vars

cb1kenobi:
On Windows 10 and 11 machines, environment variables are not case sensitive. The system PATH is actually defined as `process.env.Path`, however Node.js kindly handles the case sensitivity and will automatically return the system path when specifying `process.env.PATH`.

When we clone the environment variables via `{ ...process.env }`, we lose the automatic resolving of `Path` to `PATH`. To fix this, we need to explicitly copy the `PATH`.

#### Tests

#### Code Review


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -217,7 +217,10 @@ export function getSpawnOptions(
   nodeVersion: NodeVersion
 ): SpawnOptions {
   const opts = {
-    env: { ...process.env },
+    env: {
+      PATH: process.env.PATH,
+      ...process.env,
+    },
```
EndangeredMassa, line 223: **suggestion (non-blocking):** extract to `cloneEnv`

If we extract this into a utility function (like `cloneEnv`), we can add comment on the function explaining why this is necessary.

```
function cloneEnv(currentEnv) {
  return {
    // set uppercase `PATH` for Windows, which would normally happen
    // for us because Node.js proxies the value, but when we enumerate
    // the values here, we lose that proxied value
    PATH: currentEnv.PATH,

    ...currentEnv
  }
}
```

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -449,7 +452,8 @@ export async function runNpmInstall(
     debug(`Installing to ${destPath}`);
 
     const opts: SpawnOptionsExtended = { cwd: destPath, ...spawnOpts };
-    const env = opts.env ? { ...opts.env } : { ...process.env };
+    const envObj = opts.env || process.env;
+    const env: typeof process.env = { PATH: envObj.PATH, ...envObj };
```
EndangeredMassa: **question (soft-blocking):** Do we also need to change these places?

```
packages/cli/src/commands/bisect/index.ts:
  288:           ...process.env,

packages/cli/src/util/dev/builder.ts:
  68:     ...process.env,

packages/cli/test/dev/utils.js:
  391:         ? { ...process.env, __VERCEL_SKIP_DEV_CMD: 1 }

packages/go/go-helpers.ts:
  126:     ...process.env,

packages/go/index.ts:
  695:     ...process.env,

packages/ruby/index.ts:
  67:         ...process.env,

utils/run.js:
  94:           ...process.env,
```

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: **suggestion (non-blocking):** extract to `cloneEnv`

If we extract this into a utility function (like `cloneEnv`), we can add comment on the function explaining why this is necessary.

```
function cloneEnv(currentEnv) {
  return {
    // set uppercase `PATH` for Windows, which would normally happen
    // for us because Node.js proxies the value, but when we enumerate
    // the values below, we lose that proxied value
    PATH: currentEnv.PATH,

    ...currentEnv
  }
}
```

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
cb1kenobi: I thought of that. Wasn't sure if `build-utils` was the best place or what. I will fix it.

packages/build-utils/src/fs/run-user-scripts.ts
styfle: I don't think `envObj.PATH` will work. It needs to be `process.env.PATH` because we had to fix this here:

- https://github.com/vercel/vercel/pull/7881

cb1kenobi: `packages/cli/src/commands/bisect/index.ts:`
Bisect is using `...process.env` when spawning `--run <cmd>` and assumes the cmd is resolvable.

`packages/cli/src/util/dev/builder.ts:`
On the next line after the `...process.env`, an overwritten `PATH` is being explicitly set.

`packages/cli/test/dev/utils.js:`
I didn't fix `...process.env` in tests. I assumed if the tests pass as is on my Windows machine, then no need to touch them.

`packages/go/go-helpers.ts:`
On the next line after the `...process.env`, an overwritten `PATH` is being explicitly set.

`packages/go/index.ts:`
I will fix this. I missed this one as the go builder doesn't have unit tests.

`packages/ruby/index.ts:`
I will fix this. I missed this one as the go builder doesn't have unit tests.

`utils/run.js:`
I'm not sure what calls this script. I'm not sure exactly how to test it. Do we call this script from any Windows machines? I'm leaning towards if it ain't broke, then don't fix it.

cb1kenobi: Seems to work in both the unit tests and the Node.js REPL:

```
C:\Users\chris\projects\vercel>node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> const envObj = process.env;
undefined
> envObj.PATH
C:\Program Files (x86)\Razer Chroma SDK\bin;C:\Program Files\Razer Chroma SDK\bin;C:\Program Files (x86)\Razer\ChromaBroadcast\bin;C:\Program Files\Razer\ChromaBroadcast\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps;C:\Users\chris\AppData\Roaming\nvm;C:\Program Files\nodejs;C:\Program Files\Go\bin;C:\Program Files\Git\cmd;C:\Program Files\Docker\Docker\resources\bin;C:\ProgramData\DockerDesktop\version-bin;C:\Users\chris\.cargo\bin;C:\Users\chris\AppData\Local\Microsoft\WindowsApps;C:\Users\chris\AppData\Local\Programs\Microsoft VS Code\bin;C:\Users\chris\AppData\Roaming\nvm;C:\Program Files\nodejs;C:\Users\chris\go\bin
> process.env.PATH === envObj.PATH
true
> envObj.Path === envObj.PATH
true
```

styfle: Oh yeah I see now the issue in the previous PR was that opts was already a clone haha

packages/build-utils/src/fs/run-user-scripts.ts
styfle: Definitely needs a helper function. I think build-utils is the best place for it.

packages/build-utils/src/clone-env.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @returns {Object}
+ */
+export function cloneEnv(currentEnv = process.env): typeof process.env {
+  return {
+    // the system path is called `Path` on Windows and Node.js will
+    // automatically return the system path when accessing `PATH`,
+    // however we lose this proxied value when we destructure and
+    // thus we must explicitly copy it
+    PATH: currentEnv.PATH,
+    ...currentEnv,
```
styfle: Should PATH be moved to after instead of before?

What happens when you want to override the PATH after cloning such as here:

https://github.com/vercel/vercel/blob/177835b58864e6750e10c590721f487c0b0f73a5/packages/build-utils/src/fs/run-user-scripts.ts#L241

cb1kenobi: I don't believe there is a difference whether `PATH` is explicitly set before or after the destructure of the current environment variable object. I think it just comes down to a coding style preference and/or readability.

Not sure I follow you with the second question. `opts.env` is a new object that is a shallow clone of `process.env` with `Path` copied to `PATH`. It is possible for `opts.env` on Windows to have both a `Path` and `PATH` with identical values. If later you want to override `opts.env.PATH`, it will only mutate the cloned object.

EndangeredMassa (approved):

styfle (commented):
Looks good, the only thing you might consider is changing the function signature to match `Object.assign()`

That way we could do something like:

```js
{
  env: cloneEnv(process.env, spawnOpts.env, meta.env)
}
```

packages/build-utils/src/clone-env.ts
```diff
@@ -1,16 +1,29 @@
...
  *
  * @param {Object} [currentEnv] The environment object to clone
  * @returns {Object}
  */
-export function cloneEnv(currentEnv = process.env): typeof process.env {
-  return {
+export function cloneEnv(...envs: (ProcessEnv | undefined)[]): ProcessEnv {
+  if (!envs.length) {
+    envs.push(process.env);
+  }
```
styfle: I think we should remove this default cause and use it explicitly like `Object.assign()`. That way its clear which object is getting mutated.

packages/build-utils/src/clone-env.ts
```diff
@@ -1,16 +1,29 @@
...
+    }
+
     // the system path is called `Path` on Windows and Node.js will
     // automatically return the system path when accessing `PATH`,
     // however we lose this proxied value when we destructure and
     // thus we must explicitly copy it
-    PATH: currentEnv.PATH,
-    ...currentEnv,
-  };
+    if (obj.PATH === undefined) {
```
styfle: Not sure this is correct. 

I would expect `if (env.PATH !== undefined))`

packages/build-utils/src/clone-env.ts
cb1kenobi: Agreed

packages/build-utils/src/clone-env.ts
cb1kenobi: Gah, coding too fast.

styfle: Lets make sure to add a test to cover this case

packages/build-utils/src/types.ts
```diff
@@ -444,3 +444,5 @@ export type ShouldServe = (
 export type StartDevServer = (
   options: StartDevServerOptions
 ) => Promise<StartDevServerResult>;
+
+export type ProcessEnv = typeof process.env;
```
styfle: I don't think we should export this.

We need to be really careful about what we export from `@vercel/build-utils` because we can never change it in the future since this package is injected (community builders/runtimes dont use an explicit dependency on it)

packages/build-utils/src/clone-env.ts
```diff
@@ -0,0 +1,28 @@
...
+const { hasOwnProperty } = Object.prototype;
+
+/**
+ * Identical to `Object.assign()`, but ensures that the `PATH` environment
+ * variable is defined even when the variable is spelled `Path`.
+ *
+ * @param {Object} [...envs] Objects and/or `process.env` to merge.
+ * @returns {Object}
+ */
+export function cloneEnv(...envs: (ProcessEnv | undefined)[]): ProcessEnv {
```
styfle:
```suggestion
export function cloneEnv<T>(...envs: (T | undefined)[]): T {
```

Can we make this generic?

packages/build-utils/src/types.ts
styfle: I just realized we already have `Env` exported so lets use that instead

https://github.com/vercel/vercel/blob/b8110d97d10cb59251355fa01e8095c2761285d1/packages/build-utils/src/types.ts#L8-L10

packages/build-utils/src/clone-env.ts
cb1kenobi: I updated the code such that it will copy `PATH` if the property exists, regardless if it's `undefined`. It might be handy to unset the `PATH` with a simple `PATH: undefined`.

packages/build-utils/src/types.ts
cb1kenobi: Fixed and it actually made the code cleaner elsewhere.

packages/build-utils/src/clone-env.ts
cb1kenobi: I don't think generics will benefit here now that I'm using `Env` which is already pretty generic.

packages/cli/test/fixtures/unit/commands/git/connect/unlinked/.gitignore
```diff
@@ -1,2 +1,4 @@
 !.vercel
 .vercel
+
+.vercel
```
styfle: This looks unnecessary

packages/build-utils/src/clone-env.ts
```diff
@@ -0,0 +1,28 @@
+import type { Env } from './types';
+
+const { hasOwnProperty } = Object.prototype;
+
+/**
+ * Identical to `Object.assign()`, but ensures that the `PATH` environment
```
styfle: I realize now that `cloneEnv()` is not identical to `Object.assign()` because the former always returns a clone and the latter also mutates the first arg.

We should keep it that way but the comment should be updated since its not true.

cb1kenobi: Fixed!

styfle (approved):
Thanks!

PR #8535, opened 2022-09-08
[dev] fix middleware rewrites to relative paths that do not exist

EndangeredMassa:
When using a `rewrite` in middleware to a relative path that does not exist, the logic gets confused and falls back to the original path.

This was caused by two changes that, in combination, caused this behavior.

1. `prevUrl` was created to keep track of route results that point to other routes, but falls back to `req.url` (`prevUrl`'s initial value): [#4033](https://github.com/vercel/vercel/commit/40e4b6926756528364a4ed9d4c5a7f871f89c945#diff-142c93a61d03a1718eb765cd25e5f18d07a95358bb27ce5d5d4da314ee2fa286R1283-R1284)
2. `prevUrl` was reassigned when handling middleware: [#7973](https://github.com/vercel/vercel/commit/ee1211416f50ac04e2644b24e89403414179a3e9#diff-00ef6e7b63ed4cae298afc2a8c84f49247855a2506270f748e4d3e41da97ad99R1538)

Because `prevUrl` was reset back to `req.url` after the first `phase` (null) was run, the updated `prevUrl` value from middleware was lost. This only matters if the second `phase` ("filesystem") was going to be hit.

Further confusing matters, this was partially fixed by https://github.com/vercel/vercel/pull/8457 because it either returned a proxy pass (which doesn't have this problem) or assigned `req.url` to include the `rewritePath`, which meant that later when `prevUrl` would default to `req.url`, it still had come from `rewritePath`. So, this is fixed for the absolute URL case, but not the relative path case.

Given all that, I think the fix we need is to keep `prevUrl` at its current value when it's not being updated based on the `routeResult`.

---

So, I made that change here. I added a test that exercises this specific behavior.

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1645,7 +1645,7 @@ export default class DevServer {
       prevUrl =
         routeResult.continue && routeResult.dest
           ? getReqUrl(routeResult)
-          : req.url;
+          : prevUrl;
       prevHeaders =
         routeResult.continue && routeResult.headers ? routeResult.headers : {};
 
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1644,12 +1644,16 @@ export default class DevServer {
         missRoutes,
         phase
       );
-      prevUrl =
-        routeResult.continue && routeResult.dest
-          ? getReqUrl(routeResult)
-          : prevUrl;
-      prevHeaders =
-        routeResult.continue && routeResult.headers ? routeResult.headers : {};
+
+      if (routeResult.continue) {
+        if (routeResult.dest) {
+          prevUrl = getReqUrl(routeResult);
+        }
+
+        if (routeResult.headers) {
+          prevHeaders = routeResult.headers;
+        }
+      }
 
       if (routeResult.isDestUrl) {
         // Mix the `routes` result dest query params into the req path
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1579,7 +1579,9 @@ export default class DevServer {
               const devServerParsed = new URL(this.address);
               if (devServerParsed.origin === rewriteUrlParsed.origin) {
                 // remove origin, leaving the path
-                req.url = rewritePath.slice(rewriteUrlParsed.origin.length);
+                req.url =
+                  rewritePath.slice(rewriteUrlParsed.origin.length) || '/';
+                prevUrl = req.url;
               } else {
                 // Proxy to absolute URL with different origin
                 debug(`ProxyPass: ${rewritePath}`);
--- a/packages/cli/test/dev/fixtures/middleware-rewrite-404/api/edge.ts
+++ b/packages/cli/test/dev/fixtures/middleware-rewrite-404/api/edge.ts
@@ -1,0 +1,7 @@
+export const config = {
+  runtime: 'experimental-edge',
+};
+
+export default async function edge(request: Request, event: Event) {
+  return new Response('heyo');
+}
--- a/packages/cli/test/dev/fixtures/middleware-rewrite-404/index.html
+++ b/packages/cli/test/dev/fixtures/middleware-rewrite-404/index.html
@@ -1,0 +1,1 @@
+<h1>Hello from Index</h1>
--- a/packages/cli/test/dev/fixtures/middleware-rewrite-404/middleware.ts
+++ b/packages/cli/test/dev/fixtures/middleware-rewrite-404/middleware.ts
@@ -1,0 +1,21 @@
+export const config = {
+  runtime: 'experimental-edge',
+};
+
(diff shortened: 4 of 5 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -1579,7 +1579,9 @@ export default class DevServer {
               const devServerParsed = new URL(this.address);
               if (devServerParsed.origin === rewriteUrlParsed.origin) {
                 // remove origin, leaving the path
-                req.url = rewritePath.slice(rewriteUrlParsed.origin.length);
+                req.url =
+                  rewritePath.slice(rewriteUrlParsed.origin.length) || '/';
```
EndangeredMassa, line 1583: This is empty string in some cases. Default it to `/`.

packages/cli/src/util/dev/server.ts
```diff
@@ -1579,7 +1579,9 @@ export default class DevServer {
...
+                prevUrl = req.url;
```
EndangeredMassa, line 1584: I think this doesn't matter because the `prevUrl` would stay a full url while `req.url` would be a relative one, but for complete correctness (which may prevent future bugs), I think we should still update `prevUrl` here.

packages/cli/src/util/dev/server.ts
```diff
@@ -1645,7 +1647,7 @@ export default class DevServer {
       prevUrl =
         routeResult.continue && routeResult.dest
           ? getReqUrl(routeResult)
-          : req.url;
+          : prevUrl;
```
EndangeredMassa: This is the specific bug fix.

packages/cli/src/util/dev/server.ts
cb1kenobi: I'm fine with the current implementation, though I am curious if it would be safe if the URL could be pieced together by doing a `req.url = rewriteUrlParsed.pathname + rewriteUrlParsed.search;`? No need to slice or worry about empty strings.

packages/cli/src/util/dev/server.ts
cb1kenobi: Small knit pick, but consider a simple `if` instead of a ternary when the l-value and one of the r-values are the same.

```javascript
if (routeResult.continue && routeResult.dest) {
  prevUrl = getReqUrl(routeResult);
}
```

EndangeredMassa: I do like this more, yeah. I realize now that `prevHeaders` would have the same problem. I updated both values. [`de4c550` (#8535)](https://github.com/vercel/vercel/pull/8535/commits/de4c550d5e4f9d8d824a7ebe6fe29f8c2df2d4fe)

packages/cli/src/util/dev/server.ts
styfle: I think the slice is actually safer because concatenation would also need to account for the port. Its really easy to get url parsing wrong.

packages/cli/src/util/dev/server.ts
```diff
@@ -1642,12 +1644,16 @@ export default class DevServer {
         missRoutes,
         phase
       );
-      prevUrl =
```
styfle, line 1645: `prevUrl` is no longer assigned `req.url` in each phase. Is that expected?

packages/cli/src/util/dev/server.ts
cb1kenobi: `port` is apart of the `origin`. `pathname` is everything after the port up to the query string.

styfle: Ah right. Well then the missing piece is `hash` then

cb1kenobi: I forgot about the `hash`, but given this is a server I believe the server never receives the `hash`.

styfle: Yeah good point. I think either way is fine then 👍

packages/cli/src/util/dev/server.ts
EndangeredMassa: Yes, I believe the bug was caused by the fact that it was being re-set to the original value, before the rewrite was applied.

Here's an example of what would happen:
- given app with
	- "about.html" file 
	- "other.html" file
	- "middleware.js" that will rewrite to "/other.html"
- request to "/about.html", sets `req.url` to "/about.html"
- `prevUrl` is set to "/about.html"
- middleware executed, sets `prevUrl` to "/other.html"
- phase 1 (`null`) runs, doesn't find a route at "/other.html", sets `prevUrl` to `req.url` (THE BUG), which is "/about.html" (NOT the rewrite value of "other.html")
- phase 2 (`filesystem`) runs, sees "about.html", returns it
- note that these are the only 2 `phases` for this logic to loop over

So, we made a request to "/about.html" which the middleware rewrote to "other.html", but we were still served "about.html".

---

The intent of this logic is not super clear to me, though. I may not have the right of it.

styfle (approved):

cb1kenobi (approved):

PR #8536, opened 2022-09-08
[fs-detectors] Add the writeFile function to DetectorFilesystem

agadzik:
Adds a `writeFile` function to `DetectorFilesystem` that will be used to update the various file cache maps.

**Why is this needed?**

When detecting npm7+ monorepos, we identified a performance improvement where the service can inspect the `package-lock.json` file for workspaces, and reuse the package information for each workspace in framework-detection.

The pseudo code in `vercel/api` will look something like this

For a given lockfile
```json
{
  ...,
  "packages": {
    "": {
      "name": "npm-workspaces",
      "version": "1.0.0",
      "license": "ISC",
      "workspaces": {
        "packages": [
          "apps/*"
        ]
      }
    },
    "apps/admin": {
      "version": "0.1.0",
      "dependencies": {
        "next": "12.2.5",
        "react": "18.2.0",
        "react-dom": "18.2.0"
      },
      "devDependencies": {
        "eslint": "8.23.0",
        "eslint-config-next": "12.2.5"
      }
    },
    ...,
}
```

```ts
// for each projectPath we detect in package-lock.json
// switch the cwd of the fs to the project directory
const projectFs = fs.chdir(projectPath);
// gets the package info from the lockfile
const projectPackageInfo = lockFile.packages[projectPath];
// insert this content into fs cache
await projectFs.writeFile('package.json', projectPackageInfo)
// call detectFramework, which should now have a cached "package.json" file 
const projectFramework = await detectFramework(projectFs);
```

### Related Issues

Related to https://linear.app/vercel/issue/HIT-57/monorepo-detection-api-prevent-rate-limits

#### Tests

#### Code Review

### DIFF
--- a/packages/fs-detectors/src/detectors/filesystem.ts
+++ b/packages/fs-detectors/src/detectors/filesystem.ts
@@ -131,9 +131,9 @@ export abstract class DetectorFilesystem {
    * @param name the name of the file to write
    * @param content The content of the file
    */
-  public writeFile(name: string, content: string): void {
+  public writeFile = async (name: string, content: string): Promise<void> => {
     this.readFileCache.set(name, Promise.resolve(Buffer.from(content)));
     this.fileCache.set(name, Promise.resolve(true));
     this.pathCache.set(name, Promise.resolve(true));
-  }
+  };
 }
--- a/packages/fs-detectors/src/detectors/filesystem.ts
+++ b/packages/fs-detectors/src/detectors/filesystem.ts
@@ -125,4 +125,15 @@ export abstract class DetectorFilesystem {
   public chdir = (name: string): DetectorFilesystem => {
     return this._chdir(name);
   };
+
+  /**
+   * Writes a file to the filesystem cache.
+   * @param name the name of the file to write
+   * @param content The content of the file
+   */
+  public writeFile(name: string, content: string): void {
+    this.readFileCache.set(name, Promise.resolve(Buffer.from(content)));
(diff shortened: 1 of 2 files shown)


packages/fs-detectors/src/detectors/filesystem.ts
```diff
@@ -125,4 +125,15 @@ export abstract class DetectorFilesystem {
...
+
+  /**
+   * Writes a file to the filesystem cache.
+   * @param name the name of the file to write
+   * @param content The content of the file
+   */
+  public writeFile(name: string, content: string): void {
+    this.readFileCache.set(name, Promise.resolve(Buffer.from(content)));
+    this.fileCache.set(name, Promise.resolve(true));
+    this.pathCache.set(name, Promise.resolve(true));
```
styfle, line 137: Do we need to invoke `this._writefile()` too so the class inheriting this abstract class can change the behavior? (I'm comparing this new `writeFile()` to the existing `readFile()` above)

agadzik: I'd say we should revisit this when there's a use-case for it.  Was trying to avoid a major version bump

🤔 I could see the function signature changing to `public async writeFile(...): Promise<void>` if we wanted to support `this._writeFile()`. 

I'll go ahead and update the function signature for now to allow us in the future to add support for `this._writeFile()`

styfle (approved):

cb1kenobi (approved):

PR #8538, opened 2022-09-08
[dev] consolidate dev server address parsing

EndangeredMassa:
Consolidates parsing and access to `this.address` so that we don't have to `new URL(this.address)` all over the place.

---

Follow-up from: https://github.com/vercel/vercel/pull/8457#issuecomment-1238653904

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1036,7 +1036,7 @@ export default class DevServer {
 
     await devCommandPromise;
 
-    // log address without trailing slash to maintian backwards compatibility
+    // log address without trailing slash to maintain backwards compatibility
     const addressFormatted = this.addressUrl.toString().replace(/\/$/, '');
     this.output.ready(`Available at ${link(addressFormatted)}`);
   }
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -1036,7 +1036,9 @@ export default class DevServer {
 
     await devCommandPromise;
 
-    this.output.ready(`Available at ${link(this.addressUrl.toString())}`);
+    // log address without trailing slash to maintian backwards compatibility
+    const addressFormatted = this.addressUrl.toString().replace(/\/$/, '');
+    this.output.ready(`Available at ${link(addressFormatted)}`);
   }
 
   /**
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -135,9 +135,18 @@ export default class DevServer {
   public proxy: httpProxy;
   public envConfigs: EnvConfigs;
   public files: BuilderInputs;
-  public address: string;
-  public devCacheDir: string;
 
+  private address: URL | undefined;
+  public get addressUrl(): URL {
+    if (!this.address) {
+      throw new Error(
+        'Invalid access to `addressUrl` because `start` has not yet populated `this.address`.'
+      );
+    }
+    return this.address;
+  }
+
+  public devCacheDir: string;
   private currentDevCommand?: string;
   private caseSensitive: boolean;
   private apiDir: string | null;
@@ -175,7 +184,6 @@ export default class DevServer {
     this.systemEnvValues = options.systemEnvValues || [];
     this.projectEnvs = options.projectEnvs || [];
     this.files = {};
-    this.address = '';
     this.originalProjectSettings = options.projectSettings;
     this.projectSettings = options.projectSettings;
     this.caseSensitive = false;
@@ -714,7 +722,7 @@ export default class DevServer {
         this.projectEnvs || [],
         this.systemEnvValues || [],
         this.projectSettings?.autoExposeSystemEnvs,
-        new URL(this.address).host
+        this.addressUrl.host
       );
 
       allEnv = { ...cloudEnv };
@@ -849,7 +857,7 @@ export default class DevServer {
   injectSystemValuesInDotenv(env: Env): Env {
     for (const name of Object.keys(env)) {
       if (name === 'VERCEL_URL') {
-        env['VERCEL_URL'] = new URL(this.address).host;
+        env['VERCEL_URL'] = this.addressUrl.host;
       } else if (name === 'VERCEL_REGION') {
         env['VERCEL_REGION'] = 'dev1';
       }
@@ -922,9 +930,9 @@ export default class DevServer {
       }
     }
 
-    this.address = address
-      .replace('[::]', 'localhost')
-      .replace('127.0.0.1', 'localhost');
+    this.address = new URL(
+      address.replace('[::]', 'localhost').replace('127.0.0.1', 'localhost')
+    );
 
     const vercelConfig = await this.getVercelConfig();
     const devCommandPromise = this.runDevCommand();
@@ -1028,7 +1036,7 @@ export default class DevServer {
 
     await devCommandPromise;
 
-    this.output.ready(`Available at ${link(this.address)}`);
+    this.output.ready(`Available at ${link(this.addressUrl.toString())}`);
   }
 
   /**
@@ -1576,7 +1584,7 @@ export default class DevServer {
               const rewriteUrlParsed = new URL(rewritePath);
 
               // `this.address` already has localhost normalized from ip4 and ip6 values
-              const devServerParsed = new URL(this.address);
+              const devServerParsed = this.addressUrl;
               if (devServerParsed.origin === rewriteUrlParsed.origin) {
                 // remove origin, leaving the path
                 req.url = rewritePath.slice(rewriteUrlParsed.origin.length);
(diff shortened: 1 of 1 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -135,9 +135,18 @@ export default class DevServer {
...
 
+  private address: URL | undefined;
+  public get addressUrl(): URL {
+    if (!this.address) {
+      throw new Error(
+        'Invalid access to `addressUrl` because `start` has not yet populated `this.address`.'
+      );
+    }
+    return this.address;
+  }
```
EndangeredMassa, line 147: I wanted to have only the property:

```ts
private address: URL;
```

but we can't because this property isn't definitely set in the constructor. It's set in `start`.

If we leave it as:

```ts
private address: URL | undefined;
```

then we have to check for `undefined` before using `this.address`, which is annoying.

I think this end result works best, but definitely open to other options.

packages/cli/src/util/dev/server.ts
```diff
@@ -1028,7 +1036,9 @@ export default class DevServer {
 
     await devCommandPromise;
 
-    this.output.ready(`Available at ${link(this.address)}`);
+    // log address without trailing slash to maintian backwards compatibility
+    const addressFormatted = this.addressUrl.toString().replace(/\/$/, '');
+    this.output.ready(`Available at ${link(addressFormatted)}`);
```
EndangeredMassa, line 1044: This part is necessary because if we logged out `this.address.toString()`, there'd be an extra trailing slash. This may not matter, but just in case someone is parsing it and expecting consistent output, we shouldn't change that output.

packages/cli/src/util/dev/server.ts
TooTallNate: You can define it as `private address!: URL;` to make it always be defined.

packages/cli/src/util/dev/server.ts
```diff
@@ -135,9 +135,18 @@ export default class DevServer {
   public proxy: httpProxy;
   public envConfigs: EnvConfigs;
   public files: BuilderInputs;
-  public address: string;
-  public devCacheDir: string;
 
+  private address: URL | undefined;
+  public get addressUrl(): URL {
```
styfle: ~~Let call this `addressURL` with uppercase URL~~

Actually we don't need the getter if we do this: https://github.com/vercel/vercel/pull/8538/files#r967401450

packages/cli/src/util/dev/server.ts
styfle: You can also assign an initial value like it was doing before one line 178

```js
this.addressURL = new URL('http://localhost');
```

packages/cli/src/util/dev/server.ts
```diff
@@ -1028,7 +1036,9 @@ export default class DevServer {
 
     await devCommandPromise;
 
-    this.output.ready(`Available at ${link(this.address)}`);
+    // log address without trailing slash to maintian backwards compatibility
```
```suggestion
    // log address without trailing slash to maintain backwards compatibility
```

packages/cli/src/util/dev/server.ts
styfle: Is there any chance the address could end up being something like `http://localhost:3000/foo/` where the trailing slash would have meaning?

packages/cli/src/util/dev/server.ts
EndangeredMassa: The casing is inconsistent in this repo. It seems that for this specific case, `somethingUrl` (376 results) is more common than `somethingURL` (50 results).

Is there a defined team convention on casing that we're trying to stick to? I'd be happy to always use one or the other if there is.

packages/cli/src/util/dev/server.ts
EndangeredMassa: I try to avoid the non-null assertion if I can. It is more brittle than the other approaches.

I'm afraid an initial value would cause confusing errors (or lack of error) if it wasn't overwritten with the proper value. The current code in this PR will throw an error in that case.

packages/cli/src/util/dev/server.ts
styfle: I'm basing it on the web standards

- URL
- URLSearchParams
- URLPattern
- URL.createObjectURL()

styfle: That being said, I prefer removing it in favor of https://github.com/vercel/vercel/pull/8538/files#r967401450

packages/cli/src/util/dev/server.ts
EndangeredMassa: It's technically possible, depending on what value comes in from here: https://github.com/vercel/vercel/blob/a7dbd9649ba21478a1384011a1823573fd93a3b7/packages/cli/src/commands/dev/dev.ts#L30

I added a check that the path is only `/` and tested it out. [`4db6594` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/4db65942e55326d260eeb6d8b7c19f95f8b99577)

EndangeredMassa: Hrm. I tried it out and I get an error: `Error: listen ENOTDIR: not a directory localhost:3001/foo/`

But, I like this change anyway in case some refactoring later moves stuff around. This is a more robust replacement of the trailing slash.

packages/cli/src/util/dev/server.ts
EndangeredMassa: Updated casing: [`735b0b6` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/735b0b67cdef3d838e75f6328537d3f215d154d0)

packages/cli/src/util/dev/server.ts
styfle: There appears to be a lot of different ways to listen, its not always a URL.

It could be `unix:/home/user/server.sock` or `tcp://127.0.0.1:5000

See https://github.com/vercel/vercel/blob/0f720472c4ead9350bc57a190cc5390ee5823d52/packages/cli/test/unit/util/dev/parse-listen.test.ts

EndangeredMassa: Ah, neat!

I don't think the dev server will work with non-http anyway, but I can restrict this check to that protocol to be sure. [`8283c6d` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/8283c6db631ac34faedf52f68e5a59434077564a)

TooTallNate: `unix:` socket _used to_ work, but it seems currently it fails when doing `new URL(this.address)` in various parts of the codebase. I guess we don't have tests for that behavior 🙃

packages/cli/src/util/dev/server.ts
```diff
@@ -1576,7 +1592,7 @@ export default class DevServer {
               const rewriteUrlParsed = new URL(rewritePath);
 
               // `this.address` already has localhost normalized from ip4 and ip6 values
-              const devServerParsed = new URL(this.address);
+              const devServerParsed = this.addressURL;
```
styfle: You can remove `devServerParsed` now that is only used on this line

packages/cli/src/util/dev/server.ts
```diff
@@ -2319,7 +2335,7 @@ export default class DevServer {
 
     this.output.debug(`Spawning dev command: ${command}`);
 
-    const devPort = new URL(this.address).port;
+    const devPort = this.addressURL.port;
```
styfle: Same here, you can remove `devPort`

packages/cli/src/util/dev/server.ts
styfle: In that case maybe we should call the possible undefined on `_address` so we don't accidentally try to use it

EndangeredMassa: Styfle and I chatted about this some more. We decided that if we are going to add another variable to this giant class in this giant file, we should make it very clearly related to the getter.

Updated: [`18783d9` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/18783d9bbc0feebec68cf34800fb5e60ade4e596)

packages/cli/src/util/dev/server.ts
EndangeredMassa: Good call: [`ae1e1ee` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/ae1e1ee0b3f9d68ebbf71e0036e742d81844c1de)

packages/cli/src/util/dev/server.ts
EndangeredMassa: Good call: [`ae1e1ee` (#8538)](https://github.com/vercel/vercel/pull/8538/commits/ae1e1ee0b3f9d68ebbf71e0036e742d81844c1de)

styfle (approved):

cb1kenobi (approved):

TooTallNate (approved):

PR #8512, opened 2022-09-03
[frameworks] Fix Sanity default dev command

unitof:
### Related Issues

> Fixes https://github.com/vercel/vercel/discussions/6639 (**only for Sanity**)
> Proposed in https://github.com/vercel/vercel/discussions/8511

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8540, opened 2022-09-09
[frameworks] Detect Sanity v3 config

unitof:
Sanity v3 is currently in [developer preview](https://www.sanity.io/blog/sanity-studio-v3-developer-preview), and has deprecated `sanity.json` in the project root, replacing it with `sanity.config.js`. This should detect either.

All commands remain the same in both v2 and v3.

(Note though there's currently a bug affecting both versions' `dev` command on macOS: #8512)

### Related Issues

_none_

#### Tests

  - _Though I had to add `--gitRemote upstream` to `packages/edge` `build:docs` command, otherwise GH links are updated to my fork and docs tests fail_

#### Code Review

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1792,6 +1792,9 @@ export const frameworks = [
         {
           path: 'sanity.config.js',
         },
+        {
+          path: 'sanity.config.ts',
+        },
       ],
     },
     settings: {
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1785,9 +1785,12 @@ export const frameworks = [
     website: 'https://www.sanity.io',
     envPrefix: 'SANITY_STUDIO_',
     detectors: {
-      every: [
+      some: [
         {
           path: 'sanity.json',
+        },
+        {
+          path: 'sanity.config.js',
         },
       ],
     },


cb1kenobi:
LGTM

styfle (approved):
Perfect, thanks! 🥇

packages/frameworks/src/frameworks.ts
```diff
@@ -1785,10 +1785,13 @@ export const frameworks = [
...
     envPrefix: 'SANITY_STUDIO_',
     detectors: {
-      every: [
+      some: [
         {
           path: 'sanity.json',
         },
+        {
+          path: 'sanity.config.js',
+        },
```
styfle, line 1794:
```suggestion
        {
          path: 'sanity.config.js',
        },
        {
          path: 'sanity.config.ts',
        },
```

The linked document says `.ts` is supported as well so lets add that

kmelve: I wonder if we should include `.tsx` as well. There are entry points in the config API that assumes React components, so I wouldn't be surprised if people try to stick them into this file.

geball: Especially since [we say](https://beta.sanity.io/docs/platform/studio/config) that "All these are valid file suffixes for the studio configuration file: .js , .jsx , .ts , .tsx."

unitof: On it!

PR #8384, opened 2022-08-12
specify vite host parameter + pass env vars to Python runtime

oscartbeaumont:
Vite 3 changes the default bind address from `0.0.0.0` to `localhost` and this seems to cause issues with the vercel dev server.

Refer to [Vite changelog](https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#300-2022-07-13).

This PR also fixes a bug where environment variables on the host system are not correctly passed to the Python serverless runtime when using Vercel dev.

### Related Issues

> Related to #8121

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [x] Issue from task tracker has a link to this PR

PR #8542, opened 2022-09-09
[cli] Poll the Dev Command server on both IPv4 and IPv6 in `vc dev`

### LINKED ISSUE #8121: `vercel dev` in Vite app causes port timeout
I get the following error after a while running vercel dev in a Vite/SvelteKit app:

```
Error! Detecting port 62323 timed out after 300000ms
Error! An unexpected error occurred!
Error: Detecting port 62323 timed out after 300000ms
    at checkForPort (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243842:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at DevServer.runDevCommand (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243643:9)
    at DevServer._start (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243267:9)
Error! An unexpected error occurred!
Error: Detecting port 62323 timed out after 300000ms
    at checkForPort (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243842:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at DevServer.runDevCommand (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243643:9)
    at DevServer._start (/Users/kabir/Library/pnpm/global/5/.pnpm/vercel@27.0.0/node_modules/vercel/dist/index.js:243267:9)
```

TooTallNate:
This change will allow the downstream Dev Command server be able to listen on either of `127.0.0.1` for IPv4 or `[::1]` for IPv6.

- Fixes #6639
- Fixes #8511
- Fixes #8121
- Closes #8512
- Closes #8384

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -153,7 +153,7 @@ export default class DevServer {
   private filter: (path: string) => boolean;
   private podId: string;
   private devProcess?: ChildProcess;
-  private devProcessPort?: number;
+  private devProcessOrigin?: string;
   private devServerPids: Set<number>;
   private originalProjectSettings?: ProjectSettings;
   private projectSettings?: ProjectSettings;
@@ -1014,14 +1014,14 @@ export default class DevServer {
     // Configure the server to forward WebSocket "upgrade" events to the proxy.
     this.server.on('upgrade', async (req, socket, head) => {
       await this.startPromise;
-      if (!this.devProcessPort) {
+      if (!this.devProcessOrigin) {
         this.output.debug(
           `Detected "upgrade" event, but closing socket because no frontend dev server is running`
         );
         socket.destroy();
         return;
       }
-      const target = `http://127.0.0.1:${this.devProcessPort}`;
+      const target = this.devProcessOrigin;
       this.output.debug(`Detected "upgrade" event, proxying to ${target}`);
       this.proxy.ws(req, socket, head, { target });
     });
@@ -1820,8 +1820,8 @@ export default class DevServer {
 
     if (!match) {
       // If the dev command is started, then proxy to it
-      if (this.devProcessPort) {
-        const upstream = `http://127.0.0.1:${this.devProcessPort}`;
+      if (this.devProcessOrigin) {
+        const upstream = this.devProcessOrigin;
         debug(`Proxying to frontend dev server: ${upstream}`);
 
         // Add the Vercel platform proxy request headers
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -2641,15 +2633,29 @@ function needsBlockingBuild(buildMatch: BuildMatch): boolean {
...
+  }
+  return host;
+}
+
+async function getReachableHostOnPort(port: number): Promise<string | false> {
+  const optsIpv4 = { host: '127.0.0.1' };
+  const optsIpv6 = { host: '::1' };
+  const results = await Promise.all([
+    isPortReachable(port, optsIpv4).then(r => r && optsIpv4.host),
+    isPortReachable(port, optsIpv6).then(r => r && `[${optsIpv6.host}]`),
```
styfle, line 2655: Should the brackets be includes in the optsIpv6 on line 2653 above?

packages/cli/src/util/dev/server.ts
```diff
@@ -2641,15 +2633,29 @@ function needsBlockingBuild(buildMatch: BuildMatch): boolean {
...
+  ]);
+  return results.find(Boolean) || false;
```
styfle, line 2658: What about if both are found? Looks like this will favor ipv4 over ipv6, is that intended?

packages/cli/src/util/dev/server.ts
TooTallNate: No because you don't use the brackets in the `host` parameter of `net.connect()`/`http.request()`/etc. The brackets are only required in a URL.

packages/cli/src/util/dev/server.ts
TooTallNate: If both are reachable then it shouldn't really matter. But we could swap to prefer IPv6.

styfle: I think since Node.js prefers ipv6 by default, we probably should too

styfle (commented):
We also have a similar implementation in `@vecel/static-build` but I'm not sure if its still used (perhaps if `builds` is defined)

https://github.com/vercel/vercel/blob/2b57e12ad39687fd9225e5390da29c342c55130f/packages/static-build/src/index.ts#L52

TooTallNate:
Ya the `@vercel/static-build` version won't be used in most cases. We should probably remove it anyways.

styfle (approved):
Thanks! 🎉

PR #8545, opened 2022-09-10
[cli] Switch to `npm` and add more context to install errors during Builder import

TooTallNate:
Before:

Screen Shot 2022-09-09 at 7 03 01 PM

After:

Screen Shot 2022-09-12 at 12 27 31 PM

This is a precursor to https://github.com/vercel/vercel/pull/8485, since the last remaining test failure there is related to specialized messaging we had for these same cases.

### DIFF
--- a/packages/cli/src/util/build/import-builders.ts
+++ b/packages/cli/src/util/build/import-builders.ts
@@ -1,3 +1,4 @@
+import { URL } from 'url';
 import npa from 'npm-package-arg';
 import { satisfies } from 'semver';
 import { dirname, join } from 'path';
@@ -13,6 +14,9 @@ import { VERCEL_DIR } from '../projects/link';
 import { Output } from '../output';
 import readJSONFile from '../read-json-file';
 import { CantParseJSONFile } from '../errors-ts';
+import { errorToString, isErrnoException, isError } from '../is-error';
+import cmd from '../output/cmd';
+import code from '../output/code';
 
 export interface BuilderWithPkg {
   path: string;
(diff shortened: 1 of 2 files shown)


packages/cli/src/util/build/import-builders.ts
```diff
@@ -201,15 +205,46 @@ async function installBuilders(
...
+  } catch (err: unknown) {
+    if (isError(err)) {
+      if (isErrnoException(err) && err.code === 'ENOENT') {
+        // `yarn` is not installed
+        err.message = `Please install ${cmd('yarn')} before continuing`;
+        (err as any).link = 'https://classic.yarnpkg.com/lang/en/docs/install';
+      } else {
+        const message = errorToString(err);
+        const notFound = /"(.*): Not found"/.exec(message);
+        (err as any).link = 'https://vercel.link/npm-install-failed-dev';
```
styfle: This redirect doesnt go to the correct section

packages/cli/src/util/build/import-builders.ts
```diff
@@ -201,15 +205,46 @@ async function installBuilders(
...
+  output.log(`Installing Builders: ${Array.from(buildersToAdd).join(', ')}`);
+  try {
+    await spawnAsync('yarn', ['add', '@vercel/build-utils', ...buildersToAdd], {
+      cwd: buildersDir,
+      stdio: 'pipe',
+    });
+  } catch (err: unknown) {
+    if (isError(err)) {
+      if (isErrnoException(err) && err.code === 'ENOENT') {
+        // `yarn` is not installed
```
styfle: Should we use npm instead since that is more likely to be available on the user's system?

packages/cli/src/util/build/import-builders.ts
TooTallNate: It does, but we should use a different link here to be more generic. That's still TODO here.

packages/cli/src/util/build/import-builders.ts
TooTallNate: Ya possibly. I don't remember why I used `yarn` here originally. I'll see how easy it is to change.

styfle: Probably because thats how the build container did it. 

We can guarantee yarn in the build container but we can't guarantee on the user's local machine.

I think the `vc dev` builders use npm so we should probably consolidate that logic.

cb1kenobi: It would be neat if it detected a `yarn.lock` or `pnpm-lock.yaml` and spawned the corresponding package manager, otherwise fallback to `npm`.

TooTallNate: In this case it's related to the installation of Builders which happens in `.vercel/builders` (so we control it and there is no lockfile). But in general, ya I agree, we should have all of this logic in a `addDependencies()` function in `build-utils` similar to how we have the `installDependencies()` function.

packages/cli/src/util/build/import-builders.ts
TooTallNate: Using new link (https://vercel.link/builder-dependencies-install-failed) that points to updated docs section.

packages/cli/src/util/build/import-builders.ts
TooTallNate: Switched to `npm`.

EndangeredMassa (approved):

PR #8546, opened 2022-09-12
[frameworks] Update framework logos to be self hosted

styfle:
This removes the reliance on raw github hosting and instead relies on the Vercel deployment hosting the logo images.

Since we already have static files in each deployment (tarballs), it makes sense to start adding static images too.


cb1kenobi (approved):
CR'd and looks good. Will need to verify the URLs once the site has been deployed.

PR #8547, opened 2022-09-12
[cli] Replace "0.0.0.0" with "localhost" in `vc dev`

TooTallNate:
Instead of replacing "127.0.0.1" in `vc dev`, replace "0.0.0.0" when rendering the URL to access the `vc dev` server.

The thought here is that the IPv6 form is replacing the catch-all host name (`::`), but the IPv4 version is only replacing the explicit "127.0.0.1" version, so there's an inconsistency.

Considering the catch-all host names are not usually intended to be routable anyways, we will fix the inconsistency by only replacing the catch-all versions instead of localhost.

Also, since Node.js will default to binding to the catch-all address when not explicitly specifying a host, this replacement will only happen in that case. If the user is explicitly specifing a host to bind to (i.e. `vc dev -l 127.0.0.1:8080` or `vc dev -l tcp://[::1]:8080`) then it makes more sense to print explicitly what the user specified.

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -103,6 +103,7 @@ import {
 } from '../is-error';
 import isURL from './is-url';
 import { pickOverrides } from '../projects/project-settings';
+import { replaceLocalhost } from './parse-listen';
 
 const frontendRuntimeSet = new Set(
   frameworkList.map(f => f.useRuntime?.use || '@vercel/static-build')
@@ -930,9 +931,7 @@ export default class DevServer {
       }
     }
 
-    this._address = new URL(
-      address.replace('[::]', 'localhost').replace('0.0.0.0', 'localhost')
-    );
+    this._address = new URL(replaceLocalhost(address));
 
     const vercelConfig = await this.getVercelConfig();
     const devCommandPromise = this.runDevCommand();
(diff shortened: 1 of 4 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -931,7 +931,7 @@ export default class DevServer {
     }
 
     this._address = new URL(
-      address.replace('[::]', 'localhost').replace('127.0.0.1', 'localhost')
+      address.replace('[::]', 'localhost').replace('0.0.0.0', 'localhost')
```
styfle: Should we replace both 127.0.0.1 and 0.0.0.0?

TooTallNate: I don't think so, because the only case where it will be 127.0.0.1 is if the user explicitly specified that, in which case we should show the explicit value.

styfle: Ok seems good, just need a test

cb1kenobi (commented):
LGTM, and I like it. Can this be unit tested?

TooTallNate:
Tests added in 63a37bce629ef9a9e1574a5e158650ad9894b2ac.

EndangeredMassa (approved):

PR #8549, opened 2022-09-12
[fs-detectors] Add Nx Workspace

chloetedder:
### Related Issues

https://nx.dev/more-concepts/folder-structure#integrated-repo-folder-structure

Nx monorepo has an option to use Nx workspaces.
Nx workspace is defined within the root `workspace.json` file
Within this `workspace.json` file the workspace packages are under projects
```{
  "$schema": "./node_modules/nx/schemas/workspace-schema.json",
  "version": 2,
  "projects": {
    "myblog": "apps/myblog",
    "svelte-app": "apps/svelte-app",
  }
}
```
Within `getNxWorkspacePackagePaths` get the projects object values for the paths
Nx is listed as the last workspace manager within `workspace-managers.ts` because other workspace managers could exist to check for before Nx workspaces because the `workspace.json` could exist but not be the correct workspace manager 

Nx workspace file can exist when yarn/npm workspaces exist. When this happens, the workspace.json file is empty with no projects so it will not add any package paths to the list to look through for projects. 

#### Tests

#### Code Review


tknickman (approved):
👍🏼

styfle (commented):
CI is failing

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -103,6 +106,18 @@ async function getPackageJsonWorkspacePackagePaths({
   return getPackagePaths(packages, fs);
 }
 
+async function getNxWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  let packages: string[] = [];
+  const workspaceJsonAsBuffer = await fs.readFile('workspace.json');
+  const { workspaces } = JSON.parse(workspaceJsonAsBuffer.toString());
```
agadzik: we should handle bad JSON here in a try/catch

packages/fs-detectors/test/fixtures/42-npm-workspace-with-nx/workspace.json
```diff
@@ -0,0 +1,4 @@
+{
+  "version": 2,
+  "projects": {}
```
agadzik, line 3: is this a valid scenario to have an empty workspace.json?

chloetedder: Yes it is. It is when you initialize the project with `npx create-nx-workspace --package-manager=npm`

agadzik (approved):

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -87,29 +90,51 @@ async function getPackagePaths(
...
+  try {
+    const pnpmWorkspaceAsBuffer = await fs.readFile('pnpm-workspace.yaml');
+    const { packages = [] } = yaml.load(
+      pnpmWorkspaceAsBuffer.toString()
+    ) as PnpmWorkspaces;
+
+    return getPackagePaths(packages, fs);
+  } catch {
+    return getPackagePaths([], fs);
+  }
```
styfle: Is there a difference between `return getPackagePaths([], fs)` and `return []`?

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -87,29 +90,51 @@ async function getPackagePaths(
...
-    packageJsonAsBuffer.toString()
-  ) as PackageJsonWithWorkspace;
-
-  let packages: string[] = [];
-
-  if (Array.isArray(workspaces)) {
-    packages = workspaces;
-  } else {
-    packages = workspaces?.packages ?? [];
+  try {
```
styfle: In what case would this throw? It seems bad to consume all errors in the catch without any logging.

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: No...I should just return empty array instead 💯

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: If the file doesn't exist it would throw

chloetedder: BUT actually I don't think this would ever happen because we are already checking for the file existence when checking which type of monorepo manager it is

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -103,6 +106,16 @@ async function getPackageJsonWorkspacePackagePaths({
   return getPackagePaths(packages, fs);
 }
 
+async function getNxWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  const workspaceJsonAsBuffer = await fs.readFile('workspace.json');
+  const { projects } = JSON.parse(workspaceJsonAsBuffer.toString());
+
+  const packages: string[] = Object.values(projects) ?? [];
```
styfle:
```suggestion
  const workspace = JSON.parse(workspaceJsonAsBuffer.toString());
  const packages: string[] = Object.values(workspace?.projects ?? {});
```

This needs a test for when `workspace.json` is not the shape you are expecting

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -103,6 +106,16 @@ async function getPackageJsonWorkspacePackagePaths({
   return getPackagePaths(packages, fs);
 }
 
+async function getNxWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  const workspaceJsonAsBuffer = await fs.readFile('workspace.json');
+  const { projects } = JSON.parse(workspaceJsonAsBuffer.toString());
```
Ethan-Arrowood: I recommend not destructuring from something that can be malformed as the resulting error can be quite annoying and hard to track down sometimes. 

Without prop JSON validation we should use logic like:

```js
const workspaceJSON = JSON.parse(workspaceJsonAsBuffer.toString());
if (workspaceJSON.projects === undefined) {
  throw Error('...')
}
// ...
```

Also seeing styfle's comment now just below; similar affect and reasoning. Either solution is fine with me

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: The shape I am not expecting is empty projects array. But I covered that within the regex when looking for the workspaces. But I'll add a test specifically in this file

Ethan-Arrowood (commented):
lgtm. agree with Styfle's review; lets test when its malformed

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: Oh wait actually `'42-npm-workspace-with-nx'` this is the test case for the odd shape

chloetedder: But ill add one for if `projects` doesn't exist

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: `packages/fs-detectors/test/fixtures/43-nx-json-misshaped/workspace.json` I added this example when no projects exists

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: `packages/fs-detectors/test/fixtures/43-nx-json-misshaped/workspace.json` I added this example when no projects exists

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: `packages/fs-detectors/test/fixtures/43-nx-json-misshaped/workspace.json` I added this example when no projects exists

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -103,6 +106,17 @@ async function getPackageJsonWorkspacePackagePaths({
...
 }
 
+async function getNxWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  const nxWorkspaceJsonAsBuffer = await fs.readFile('workspace.json');
+
+  const { projects } = JSON.parse(nxWorkspaceJsonAsBuffer.toString());
+
+  const packages: string[] = Object.values(projects) ?? [];
```
styfle: In what case can `Object.values(projects)` be falsy? Seems like the `??` is not used.

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
styfle: What about if `workspace.json` is the following:

```json
{
  "projects": "hello"
}
```

chloetedder: Let me add that in

chloetedder: `'44-nx-json-string'` added test case

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: Just tested and that is the case that `??` is not used. Updated

styfle (approved):

PR #8550, opened 2022-09-12
[build-utils] Add flag for `npm install --legacy-peer-deps`

styfle:
Prior to `npm@8.6.0`, running `npm install --legacy-peer-deps` to generate a lock file and then later running `npm install` would work fine.

After `npm@8.6.0`, running `npm install` with that same lock file will fail with `code ERESOLVE` and `Conflicting peer dependency`.

Steps to reproduce:

```sh
corepack enable npm
echo '{"packageManager":"npm@8.5.5","dependencies":{"next":"12.3.0","react":"16.8.0"}}' > package.json
npm install --legacy-peer-deps
npm install # works
echo '{"packageManager":"npm@8.6.0","dependencies":{"next":"12.3.0","react":"16.8.0"}}' > package.json
npm install # fails with code ERESOLVE
```

This PR introduces a flag so we can safely upgrade npm to a newer version without breaking thousands of deployments.

The [npm docs](https://docs.npmjs.com/cli/v8/using-npm/config#legacy-peer-deps) say the flag is not recommended so we also make sure that once we upgrade to Node.js 18, we stop flagging `--legacy-peer-deps` because its safe to introduce breaking changes.


EndangeredMassa (approved):
I like this approach. It's fine if `npm` considered this a bug fix instead of a breaking change, but on our platform, this is a breaking change. This allows us to treat it as such.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -464,6 +464,18 @@ export async function runNpmInstall(
       commandArgs = args
         .filter(a => a !== '--prefer-offline')
         .concat(['install', '--no-audit', '--unsafe-perm']);
+      if (
+        nodeVersion?.major === 16 &&
```
cb1kenobi, line 469: Does this need to check the `npm` version? Do we even need to do an `if` at all? I just tested `npm@6.14.2` with `--legacy-peer-deps` and it succeeded without error or any warnings. Was this `npm` issue just with 8.6.0? The `npm` version that ships with Node 18 won't be an issue?

styfle: @cb1kenobi We don't have the npm version here unfortunately. The issue is with 8.6.0 or newer (see code comment).

You can read more details in slack: https://vercel.slack.com/archives/C041Z5U9C5B

Also see this document: https://www.notion.so/npm-install-exited-with-1-54368abd9798467f8296dfed3cc33258

cb1kenobi: The problem is the Node.js version doesn't necessarily correspond to the npm version.

Getting the npm version is pretty quick and easy:

```javascript
const [ major, minor ] = execFileSync('npm', ['--version'], { encoding: 'utf-8' }).trim().split('.').map(n => ~~n);
if (major > 8 || (major === 8 && minor >= 6)) {
  commandArgs.push('--legacy-peer-deps');
}
```

It's relatively cheap compared to the `npm install` that will happen next in the code. I measure about 185ms on my machine.

Or, just always set `--legacy-peer-deps` as it doesn't seem to cause issue with npm 6. :)

styfle: It took 322 ms on my machine. 

We would also need to pass in `opts.env` or else that would select the wrong npm.

styfle: > Or, just always set --legacy-peer-deps as it doesn't seem to cause issue with npm 6

I think that is the best solution however I'm using node 16 as a way to prevent patching this when we release node 18 as my comment mentions.

Since users have to opt into node 18 (when the time comes) we will no longer need this patch because its safe to have "breaking changes" in semver major

TooTallNate (approved):

Ethan-Arrowood (approved):

cb1kenobi (approved):

packages/build-utils/test/unit.run-npm-install.test.ts
```diff
@@ -1,15 +1,23 @@
...
-import { runNpmInstall } from '../src';
+import { runNpmInstall, cloneEnv } from '../src';
 const spawnMock = jest.spyOn(require('child_process'), 'spawn');
 
 afterEach(() => {
   spawnMock.mockClear();
 });
 
+function getTestSpawnOpts(env: Record<string, string>) {
+  return { env: cloneEnv(process.env, env) };
```
cb1kenobi, line 25: ❤️

cb1kenobi (approved):

cb1kenobi:
Dunno if CI is gonna work, but locally the new mock doesn't work. Same error with `arg[0]` != `"npm"`.

If I add a `console.log(args)` to the top of the mock, I get:

```
[
      'C:\\Windows\\system32\\cmd.exe',
      [
        '/d',
        '/s',
        '/c',
        '"npm ^"install^" ^"--no-audit^" ^"--unsafe-perm^""'
      ],
      {
        stdio: 'inherit',
        cwd: 'C:\\Users\\chris\\projects\\vercel\\packages\\build-utils\\test\\fixtures\\20-npm-7',
        env: {
          PATH: <snip>
        },
        prettyCommand: 'npm install',
        windowsVerbatimArguments: true
      }
]
```

cb1kenobi (approved):

PR #8552, opened 2022-09-13
[tests] Fix tarball lookup with sha contains 0

styfle:
This PR fixes an issue when the sha begins with `0` and causes [an error](https://vercel.com/vercel/vercel/EP7fVcXKsoodzWRy3fwG8AoeQcs6) when looking up the tarball:

> Error: ENOENT: no such file or directory, 
> copyfile '/vercel/path0/packages/build-utils/vercel-build-utils-v5.4.2-0251253.tgz' 
>  -> '/vercel/path0/public/tarballs/@vercel/build-utils.tgz'

The fix is to no longer rely on the exact tarball name because we can't guarantee the name from `yarn pack` and instead rely on a pattern for the tarball name.

### DIFF
--- a/api/_lib/script/build.ts
+++ b/api/_lib/script/build.ts
@@ -1,6 +1,5 @@
 import fs from 'fs/promises';
 import { join, dirname } from 'path';
-import execa from 'execa';
 import { getExampleList } from '../examples/example-list';
 import { mapOldToNew } from '../examples/map-old-to-new';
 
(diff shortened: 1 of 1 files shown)


api/_lib/script/build.ts
```diff
@@ -61,12 +56,21 @@ async function main() {
       'utf-8'
     );
     const packageJson = JSON.parse(packageJsonRaw);
-    const tarballName = `${packageJson.name
-      .replace('@', '')
-      .replace('/', '-')}-v${packageJson.version}-${sha.trim()}.tgz`;
+    const files = await fs.readdir(fullDir);
+    const tarballName = files.find(f => /^vercel-(.+)\.tgz$/.test(f));
```
TooTallNate: Not a big deal but you don't need the capture group.

```suggestion
    const tarballName = files.find(f => /^vercel-.+\.tgz$/.test(f));
```

TooTallNate (approved):

PR #8554, opened 2022-09-13
[tests] Add more tests for `importBuilders()` scenarios

TooTallNate:
Covers:
  - Importing a built-in Builder with `@latest` should use the CLI's version
  - Importing a built-in Builder with `@canary` should use the CLI's version
  - Importing a built-in Builder with an explicit version should install to `.vercel/builders`


packages/cli/test/unit/util/build/import-builders.test.ts
```diff
@@ -27,7 +27,67 @@ describe('importBuilders()', () => {
...
+    expect(builders.get('@vercel/next@latest')?.pkg).toMatchObject(
+      vercelNextPkg
+    );
+    expect(typeof builders.get('@vercel/node@latest')?.builder.build).toEqual(
+      'function'
+    );
+    expect(typeof builders.get('@vercel/next@latest')?.builder.build).toEqual(
+      'function'
+    );
+  });
```
styfle, line 60: Can we verify that no "npm install" was run?

Perhaps check `client.stderr`?

```js
expect(client.stderr).not.toMatch('Installing Builder');
```

TooTallNate: The "not" case depends on a timeout so I prefer not to use that if necessary. I'd rather check the imported file path. Will add a case for that.

cb1kenobi (approved):

packages/cli/test/unit/util/build/import-builders.test.ts
TooTallNate: dc7d3a6b30c3a8d24ca4d8f91c6778f36d0d4877

styfle (approved):
Great, thanks!

PR #8555, opened 2022-09-13
Add "rush" workspace

chloetedder:
### Related Issues

Rush does not intrinsically have its own version of workspaces. But it does allow for a user to set `useWorkspaces: true` within the `rush.json`
This will utilize pnpm workspaces BUT the catch is the yaml file for this is found within the defaulted `/common/temp` folder.

[As stated in the documentation](https://rushjs.io/pages/configs/rush_json/) for rush.json, the workspace feature is experimental

#### Tests

#### Code Review

### DIFF
--- a/packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
+++ b/packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
@@ -33,6 +33,12 @@ export async function getWorkspacePackagePaths({
     case 'pnpm':
       results = await getPnpmWorkspacePackagePaths({ fs: workspaceFs });
       break;
+    case 'rush':
+      // /common/temp is the location that Rush defaults adding the pnpm workspace file
+      results = await getPnpmWorkspacePackagePaths({
+        fs: fs.chdir('/common/temp'),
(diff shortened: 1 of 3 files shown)


packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -33,6 +33,12 @@ export async function getWorkspacePackagePaths({
     case 'pnpm':
       results = await getPnpmWorkspacePackagePaths({ fs: workspaceFs });
       break;
+    case 'rush':
+      // /common/temp is the location that Rush defaults adding the pnpm workspace file
+      results = await getPnpmWorkspacePackagePaths({
+        fs: fs.chdir('/common/temp'),
```
gaspar09, line 39: Seeing how this can be used, it my be more robust to use projects enumerated in `rush.json`

https://github.com/microsoft/rushstack/blob/8671d79a48b2a87bcf139567ab8281e50d00c9b7/libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts#L126-L137

PR #8565, opened 2022-09-14
[fs-detectors] Rush json workspace

chloetedder:
### Related Issues

Add in Rush workspace

#### Tests

#### Code Review


agadzik (approved):
LGTM!

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -113,3 +124,21 @@ async function getPnpmWorkspacePackagePaths({
...
 }
+
+async function getRushWorkspacePackagePaths({
+  fs,
+}: GetPackagePathOptions): Promise<string[]> {
+  const rushWorkspaceAsBuffer = await fs.readFile('rush.json');
+
+  const { projects = [] } = JSON.parse(
+    rushWorkspaceAsBuffer.toString()
+  ) as RushWorkspaces;
```
EndangeredMassa, line 149: **suggestion (blocking):** throw better error message on parse failure

Can we try-catch this JSON parse and throw a better error message if it fails? Something like `Failed to parse "rush.json": ${error.message}` would be helpful.

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -113,3 +124,21 @@ async function getPnpmWorkspacePackagePaths({
...
+
+  if (Array.isArray(projects)) {
```
EndangeredMassa, line 151: **suggestion (blocking):** validate `rush.json`

Can we validate the parsed result against the `rush.json` schema? That way we'll get better error messages when it's invalid.

Right now, if `projects` is not an array, no message is given to the user about what is wrong.

Ethan-Arrowood (approved):
lgtm

EndangeredMassa (changes requested):

gaspar09:
Looks good. Same suggestions as @EndangeredMassa

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: I don't think we want to thrown an error because we still want to return something for the user. But I could do a `console.error`

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: ^^ Same as above, would it be acceptable to return the empty array but to also log the error? Or do you think we should throw an error

packages/fs-detectors/test/unit.get-workspaces-package-paths.test.ts
```diff
@@ -14,6 +14,9 @@ describe.each<[string, string[]]>([
     ['/backend/c', '/backend/d', '/frontend/a', '/frontend/b'],
   ],
   ['22-pnpm', []],
+  ['40-rush-monorepo', ['/apps/my-app', '/apps/my-second-app']],
+  ['41-rush-monorepo-empty', []],
+  ['42-rush-json-invalid', []],
 ])('`getWorkspacesPackagePaths()`', (fixturePath, packagePaths) => {
```
EndangeredMassa: While we're here, can we fix the typo in the description?
```suggestion
])('`getWorkspacePackagePaths()`', (fixturePath, packagePaths) => {
```

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
EndangeredMassa: From a library/package perspective, I think `fs-detectors` should throw an error in this case. There's a problem somewhere and not throwing an error will make it harder for someone to figure out where the problem is.

It looks like the API uses this. If we make `fs-detectors` throw an error here, then handle that error in the API endpoint and return an empty set there, I think that's reasonable. We'd log the error somewhere from the API endpoint and could still track down what happened.

- [`getWorkspacePackagePaths` at `api/utils/detect-eligible-projects/get-eligible-projects.ts`](https://github.com/vercel/api/blob/de56ea3e1414769c6f517c7622ad55056d14e530/utils/detect-eligible-projects/get-eligible-projects.ts#L53-L56)
- [`getEligibleProjects` at `api/utils/detect-eligible-projects/detect-eligible-projects.ts`](https://github.com/vercel/api/blob/de56ea3e1414769c6f517c7622ad55056d14e530/utils/detect-eligible-projects/detect-eligible-projects.ts#L54)
- [`detectEligibleProjects` at `api/utils/detect-eligible-projects/detect-eligible-projects.ts `](https://github.com/vercel/api/blob/de56ea3e1414769c6f517c7622ad55056d14e530/services/api-integrations/src/endpoints/detect-eligible-projects.ts#L89)



I'm less familiar with this area of the codebase. There could definitely be context I'm missing, though.

chloetedder: Yeah we could catch it in the api for `detectEligibleProjects` 👍 I'll make that change

chloetedder: Since this would apply to all the calls not just for this PR I am going to follow up with a separate PR for throwing errors if that works

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
styfle: Depends if we think there are cases where `rush.json` is not a corresponding Rush workspace (like a user just happened to name it that).

If we know its always a Rush workspace, then we should show the error to the user. If not we should just consume it and move onto the next file

chloetedder: It would at this point that we know they have a `rush.json` which means they should have rush workspaces. If `projects` is not an array that would mean the user has their `rush.json` configured incorrectly (this should be rare because most cases it shows the user errors for the format so if they truly are using Rush they wouldn't be able to run it with an incorrectly formatted `rush.json`

styfle: Theres also the case when `projects` exists but `projectFolder` doesnt.

styfle: > at this point that we know they have a rush.json which means they should have rush workspaces.

Why is that? Couldn't a user just have an arbitrary file named `rush.json` and store whatever json they want in there for their own purpose?

This is different than `yarn.lock` which is a very established filename that is unlikely to collide with user's files.

chloetedder: Correct they could have an arbitrary file named that.

So do you think we should be checking for `projects: [{ projectFolder:...}]` here and not just for `projects` as an array?

Then if we checked for `rush.json` to contain array of `projects` with `projectFolder` and if it doesn't match that then send an error

EndangeredMassa: I think it's reasonable to look only for `projects: [{ projectFolder:...}]` for now, but I'd like to iterate on schema validation for better error messages.

Given that this file is not a common standard where we can assume it's a Rush config file, I think we should debug log any errors we find here. This would give someone the ability to figure out why their `rush.json` didn't work properly when they were using Rush. It would also hide these details completely when they were not using Rush.

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
EndangeredMassa: A separate PR for this sounds good.

EndangeredMassa (approved):

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
chloetedder: @EndangeredMassa For `vercel` where do the logs go? Would `console.error` go to the customer logs?

EndangeredMassa: Hrm. Because we're in `fs-detectors`, I don't think there's a clear way to log this. Ideally, `getWorkspacePackagePaths` would return a complicated response that includes what workspace package managers it tried to parse and what the results were. Then the API could log that itself. It could go further and return that more complicated response as some metadata in the API response, where the caller could log it if it made sense. That's way out of scope for this PR, though.

Maybe instead when we expand Vercel CLI to better support monorepos, we can use `fs-detectors` to parse workspace package manager configs and show validation failures. That would give the user a path to figuring out what's breaking.

For this PR, I think we're good to go.

styfle:
@chloetedder The tests added here are failing so I reverted in #8591

PR #8379, opened 2022-08-12
[cli] Add "image" optimization configuration support in `vercel.json`

theMosaad:
Continuation of https://github.com/vercel/vercel/pull/8217

PR #8566, opened 2022-09-14
[cli] Add "images" optimization configuration support in `vercel.json`

TooTallNate:
This is a (further 😄) continuation of #8379 by @theMosaad, so that the tests can run.

I also added a change to make `images` _not_ be sent to the create deployment API endpoint, since it fails validation there.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -522,7 +522,12 @@ export default async (client: Client): Promise<number> => {
       quiet,
       wantsPublic: argv['--public'] || localConfig.public,
       type: null,
-      nowConfig: localConfig,
+      nowConfig: {
+        ...localConfig,
+        // `images` is allowed in "vercel.json" and processed
+        // by `vc build`, but don't send it to the API endpoint
+        images: undefined,
```
styfle, line 529: If we dont send it to the API, how do we make it work with the deployment?

Are you thinking `.vercel/output/config.json` is going to be read by the API instead?

TooTallNate: It's still in the `vercel.json` source file, which is read during build by `vc build`.

styfle (approved):

packages/cli/src/commands/build.ts
```diff
@@ -555,7 +555,150 @@ async function doBuild(
     builds: builderRoutes,
   });
 
-  const mergedImages = mergeImages(buildResults.values());
+  const images = vercelConfig?.images
+  if (images) {
```
EndangeredMassa, line 559: **question (non-blocking):** Should we do most of this with schema validation instead?

**question (non-blocking):** Can we extract this into a function like `validateImages`? It would make it easier to add tests for the validation error cases.

EndangeredMassa (approved):

packages/cli/src/commands/build.ts
TooTallNate: Sounds good on both counts. I think this logic was lifted from `@vercel/next`.

Ethan-Arrowood (approved):
this lgtm, i'd like to see the validation logic tested more but wont block on that

packages/cli/src/commands/build.ts
styfle: We have some vercel.json schema validation for `vc dev` here.
So we should probably do the same for `vc build` (and add images schema 🙂 )

https://github.com/vercel/vercel/blob/e5131d55575096beb764e11ae2ff5e19ecaf940e/packages/cli/src/util/dev/validate.ts#L19-L34

TooTallNate:
Tagging for merge. We can address validation in a follow-up PR.

PR #8568, opened 2022-09-15
[cli] Clean up after create-git-meta test

cb1kenobi:
The `create-git-meta` unit test will modify the git index file causing the `vercel` repo dirty. By copying the git metadata which will be mutated to a subdirectory, we can easily cleanup after the test runs.

#### Tests

#### Code Review


packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -182,12 +182,13 @@ describe('createGitMeta', () => {
   });
   it('detects not dirty commit', async () => {
     const directory = fixture('not-dirty');
+    const dest = join(directory, 'repo', '.git');
     try {
-      await fs.rename(join(directory, 'git'), join(directory, '.git'));
-      const dirty = await isDirty(directory, client.output);
+      await fs.copy(join(directory, 'git'), dest);
```
styfle, line 187: The heirarchy here is a little weird. Can we do the same without separating `git` and `repo` into two directories?

cb1kenobi: I don't think it's weird. I'm still keeping all the create-git-meta fixture files in the same directory. Using a single directory for the fixture files as well as the mutated git repo files is what caused this problem in the first place. An alternative solution would be to create a temp directory and copy the test, or at the very least the `git/index` file, there before running. This was a simple solution.

TooTallNate: What I don't understand is why the repo is being modified in the first place. This test (the `isDirty()` function) basically just runs `git status` 🤔

cb1kenobi: It gits weirder. When you run `git status`, it refreshes/optimizes the index in the background (https://git-scm.com/docs/git-status#_background_refresh). I have created a PR to disable the index updates which we should probably accept in favor of this PR: https://github.com/vercel/vercel/pull/8569.

cb1kenobi:
Closing in favor of something else... possibly #8569.

PR #8569, opened 2022-09-15
[cli] Disable git status background index refresh/optimization

cb1kenobi:
The `create-git-meta` file contains a `isDirty()` function which calls `git status` and subsequently causes one of the test fixture's git index to be updated. Turns out that `git status` will update/optimize the index in the background when invoked: https://git-scm.com/docs/git-status#_background_refresh.

By disabling this using the `--no-optional-locks` flag, `isDirty()` will no longer update the index when checking the status.

From my testing, the `--no-optional-locks` flag must be set after `git` and before `status`.

This PR is likely a better alternative to https://github.com/vercel/vercel/pull/8568 as this PR eliminates the problem all together.

#### Tests

#### Code Review


packages/cli/src/util/create-git-meta.ts
```diff
@@ -78,16 +78,20 @@ function getLastCommit(directory: string): Promise<git.Commit> {
...
   return new Promise(resolve => {
-    exec('git status -s', { cwd: directory }, function (err, stdout, stderr) {
-      let debugMessage = `Failed to determine if Git repo has been modified:`;
-      if (err || stderr) {
-        if (err) debugMessage += `\n${err}`;
-        if (stderr) debugMessage += `\n${stderr.trim()}`;
-        output.debug(debugMessage);
-        return resolve(false);
+    exec(
+      'git --no-optional-locks status -s',
```
EndangeredMassa, line 85: **suggestion (non-blocking):** add comment about why `--no-optional-locks` is necessary

Can we add a comment about this flag? The flag name and its [official description](https://git-scm.com/docs/git/2.22.1#Documentation/git.txt---no-optional-locks) don't do a great job explaining what the actual effects are.

EndangeredMassa (approved):

packages/cli/src/util/create-git-meta.ts
cb1kenobi: Good idea! Done!

EndangeredMassa (approved):

Ethan-Arrowood (approved):
lgtm!

packages/cli/src/util/create-git-meta.ts
styfle: If we're worried about the order of the flags, we could use the `GIT_OPTIONAL_LOCKS=0` env var as an alternative

styfle (approved):

cb1kenobi:
@styfle We have another option... we could set `GIT_OPTIONAL_LOCKS=1` just for the unit test. That way we don't need to set the `--no-optional-locks` flag in production and we don't have to worry about the test mutating files. Thoughts?

styfle:
@cb1kenobi I think its worth keeping for improved performance for our users too. We just want the quickest way to know if the working tree is dirty or not.

TooTallNate (approved):

PR #8590, opened 2022-09-20
[tests] Skip test 01-normalize-routes temporarily

styfle:
The  `01-normalize-routes` test is failing on windows so we need to temporarily disable until the root cause can be corrected.


EndangeredMassa (approved):

cb1kenobi:
The test still fails. I don’t think the lack of a lock file is the problem.

PR #8594, opened 2022-09-20
[tests] Update workflows to use global `env`

styfle:
Follow up to #8586 because [that syntax](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable) never worked on windows


TooTallNate (approved):

cb1kenobi (approved):
Cleaner, very nice

ijjk (approved):

PR #8596, opened 2022-09-20
[build-utils][next] Remove Path once copied to PATH.

cb1kenobi:
Windows doesn't like it when the env has both a `PATH` and a `Path`, so `cloneEnv()` must remove `Path` after copying it to `PATH`.

Renamed the next builder `01-normalize-routes` test to `01-normalize-paths` as it's test normalization of output paths, then re-enabled the test for Windows.

#### Tests

#### Code Review


packages/build-utils/src/clone-env.ts
```diff
@@ -16,14 +16,19 @@ export function cloneEnv(...envs: (Env | undefined)[]): Env {
...
-    // the system path is called `Path` on Windows and Node.js will
-    // automatically return the system path when accessing `PATH`,
-    // however we lose this proxied value when we destructure and
-    // thus we must explicitly copy it
-    if (hasOwnProperty.call(env, 'PATH') || hasOwnProperty.call(env, 'Path')) {
-      obj.PATH = env.PATH;
+    // mixin the env first
+    obj = Object.assign(obj, env);
+
+    if (hasOwnProperty.call(env, 'Path')) {
```
EndangeredMassa, line 22: **question (non-blocking):** Should we only do this on Windows?

Now that it deletes, it's not really a no-op for other platforms. But maybe we'd rather have it work the same for all platforms anyway--no one gets to access `Path`.

packages/build-utils/src/clone-env.ts
```diff
@@ -16,14 +16,19 @@ export function cloneEnv(...envs: (Env | undefined)[]): Env {
...
+      // the system path is called `Path` on Windows and Node.js will
+      // automatically return the system path when accessing `PATH`,
+      // however we lose this proxied value when we destructure and
+      // thus we must explicitly copy it, but we must also remove the
+      // `Path` property since we can't have both a `PATH` and `Path`
+      obj.PATH = obj.Path;
+      delete obj.Path;
```
EndangeredMassa, line 29: **question (non-blocking):** Is it OK that `Path` will never be in the result?

Will all of the uses of the result of `cloneEnv` always work when `Path` is not there?

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
```diff
@@ -1,184 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
styfle, line 1: Lets keep the lock file because it should be faster, right?

packages/build-utils/src/clone-env.ts
cb1kenobi: It should be harmless for all platforms. While it's true that `Path` should only exist on Windows, there's nothing stopping someone from setting `process.env.Path` as `process.env` is mutable.

Not removing `Path` allowing both `Path` and `PATH` is ultimately what causes the issue. I suspect that `spawn()` will happily pass both `Path` and `PATH` along to the system call, but Win32 will use `Path` or `PATH`, meanwhile if `yarn` looks at the path, it will do the same as `vercel` and look for `PATH`.

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
cb1kenobi: I can't speak to the performance, but the `package.json` has all deps as `"latest"` and since that can change, creating a lock file for a single point in time may cause issues down the road. On the other hand, 6 of the 90-some next builder fixtures have lock files. I really don't think we need to weigh down the repo with this file, but I'm indifferent.

packages/build-utils/src/clone-env.ts
cb1kenobi: Just to circle back, we can't have both `Path` and `PATH` since that's what causes this issue. A quick grep of the code yields no match where we're referencing the `PATH` as spelled `Path`. We should be good to go.

styfle: I think this is is the purpose of the `cloneEnv()` function. After cloning, we will read `.PATH` and never `.Path`

Ethan-Arrowood (approved):

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
styfle: @cb1kenobi This test is not really testing Next.js though, its testing the Builder so I would prefer it be faster

packages/build-utils/src/clone-env.ts
EndangeredMassa: If we need to remove `Path`, we need to remove it, sure. I'm wondering if there would be other issues caused by it being forcibly deleted when that wasn't happening before.

On Windows, it doesn't seem like it would matter. On another OS, it would only matter if you were explicitly expecting `Path` to show up somewhere, which seems unlikely.

packages/next/test/unit/fixtures/01-normalize-routes/yarn.lock
cb1kenobi: Good idea! Here's the PR to add the lock file back: https://github.com/vercel/vercel/pull/8600.

PR #8600, opened 2022-09-21
[tests] Added lock file for normalize paths test fixture

cb1kenobi:
Added a lock file to the normalize paths text fixture to speed up install dependencies.

#### Tests

#### Code Review


styfle (approved):

styfle:
@cb1kenobi How about adding assertions for `routes` here?

https://github.com/vercel/vercel/blob/48eb72055603c954b7e399e7783474fd0e87316b/packages/next/test/unit/fixtures/01-normalize-paths/index.test.js#L27

cb1kenobi:
> @cb1kenobi How about adding assertions for `routes` here?
> 
> https://github.com/vercel/vercel/blob/48eb72055603c954b7e399e7783474fd0e87316b/packages/next/test/unit/fixtures/01-normalize-paths/index.test.js#L27

I looked into it, but `routes` doesn't contain any "paths" per se. It's basically a bunch of catch-all patterns, so I didn't think there was anything related to the path normalization to test in `routes`.

TooTallNate (approved):

styfle:
@cb1kenobi Right, they are not file paths they are routes so they should always contain `/` and never `\` unless escaping regex in `src`. But `dest` should never have `\`.

I was pretty sure these were incorrectly using `path.join()` before and it was changed to use `path.posix.join()`

cb1kenobi:
@styfle You are correct. Added some assertions here: https://github.com/vercel/vercel/pull/8609.

PR #8603, opened 2022-09-21
[fs-detectors] Rush json workspace

chloetedder:
### Related Issues

Add in getting the package paths for Rush workspaces.
First get the contents of the `rush.json` file. If the `projects` property is an array, map through that array and return the `projectFolder` as the package path.

#### Tests

#### Code Review


packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
```diff
@@ -127,3 +138,23 @@ async function getPnpmWorkspacePackagePaths({
...
+  const rushWorkspaceAsBuffer = await fs.readFile('rush.json');
+
+  const { projects = [] } = JSON.parse(
+    rushWorkspaceAsBuffer.toString()
+  ) as RushWorkspaces;
+
+  if (Array.isArray(projects)) {
+    const packages = projects
+      .filter(proj => proj.projectFolder)
+      .map(project => project.projectFolder);
```
EndangeredMassa, line 154: This is where the [previous PR](https://github.com/vercel/vercel/pull/8565/files#diff-b318a167de8653a693a7d6029d515e14dadae456c9e5834debb60c40e9077fa7R152) had a preceding `/` added to the project directory, which caused the later `glob` to return results with the windows prefix (like `C:`), which was failing the test assertions.

EndangeredMassa (approved):

packages/fs-detectors/src/workspaces/get-workspace-package-paths.ts
styfle: @chloetedder Any reason this is using `proj` for the filter but `project` for the map?

packages/fs-detectors/test/fixtures/40-rush-monorepo/apps/my-app/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "a",
```
styfle: I think the name in each package.json should be unique

chloetedder: Oh yes let me update it to match the path name (I had been doing that for most)

TooTallNate (approved):

PR #8605, opened 2022-09-21
[cli][dev] Use `browser` platform when bundling edge functions for `vc dev`

EndangeredMassa:
When bundling edge functions for `vc dev`, we were targeting a `node14` platform. This meant that the output would have `require` statements left in for core node libraries, which would fail when executed in `edge-runtime`, which does not allow for that.

This change sets the platform to `browser`, which changes the bundling behavior to replace `require('crypto').whatever` with `crypto.whatever`, using the global `crypto`. This is expected for `edge-runtime`.

This PR also changes the target to `node14`, `node16`, or whatever major node version is currently running. This ensures that the VM spawned by the current version of node can execute the generated syntax.


cramforce (approved):

TooTallNate (approved):

packages/node/src/dev-server.ts
```diff
@@ -84,6 +84,14 @@ import esbuild from 'esbuild';
 import fetch from 'node-fetch';
 import { createEdgeWasmPlugin, WasmAssets } from './edge-wasm-plugin';
 
+const NODE_VERSION_MAJOR = process.version.match(/^v(\d+)\.\d+/)?.[1];
```
styfle, line 87: Another alternative is:

```js
process.versions.node.split('.')[0];
```

because the `v` is excluded there

styfle (approved):

PR #8611, opened 2022-09-22
[next] Update to route app rsc paths correctly

ijjk:
### Related Issues

x-ref: [slack thread](https://vercel.slack.com/archives/C035J346QQL/p1663799417151099)

Failing probes are unrelated, more info can be seen in [this thread](https://vercel.slack.com/archives/C035J346QQL/p1663820032810519?thread_ts=1663775935.504379&cid=C035J346QQL)

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1131,21 +1131,21 @@ export async function serverBuild({
     });
   }
 
-  if (appDir) {
+  if (appPathRoutesManifest) {
     // create .rsc variant for app lambdas and edge functions
     // to match prerenders so we can route the same when the
     // __flight__ header is present
-    Object.values(appPathRoutesManifest || {}).forEach(route => {
-      route = path.posix.join('./', route === '/' ? '/index' : route);
+    const edgeFunctions = middleware.edgeFunctions as Record<string, any>;
 
-      const edgeFunctions = middleware.edgeFunctions as Record<string, any>;
+    for (let route of Object.values(appPathRoutesManifest)) {
+      route = path.posix.join('./', route === '/' ? '/index' : route);
 
       if (lambdas[route]) {
         lambdas[`${route}.rsc`] = lambdas[route];
       } else if (edgeFunctions[route]) {
         edgeFunctions[`${route}.rsc`] = edgeFunctions[route];
       }
-    });
+    }
   }
 
   return {
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1135,7 +1135,7 @@ export async function serverBuild({
     // create .rsc variant for app lambdas and edge functions
     // to match prerenders so we can route the same when the
     // __flight__ header is present
-    const edgeFunctions = middleware.edgeFunctions as Record<string, any>;
+    const edgeFunctions = middleware.edgeFunctions;
 
     for (let route of Object.values(appPathRoutesManifest)) {
       route = path.posix.join('./', route === '/' ? '/index' : route);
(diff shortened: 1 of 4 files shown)


packages/next/src/server-build.ts
```diff
@@ -1131,6 +1131,23 @@ export async function serverBuild({
     });
   }
 
+  if (appDir) {
+    // create .rsc variant for app lambdas and edge functions
+    // to match prerenders so we can route the same when the
+    // __flight__ header is present
+    Object.values(appPathRoutesManifest || {}).forEach(route => {
```
cb1kenobi: Not sure if performance is a concern, but a for-of loop would be more efficient as V8 wouldn't have to allocate a stack frame for each iteration.

packages/next/src/server-build.ts
```diff
@@ -1131,6 +1131,23 @@ export async function serverBuild({
     });
   }
 
+  if (appDir) {
```
cb1kenobi: `appDir` is `null` unless `appPathRoutesManifest` is truthy, so you could check `appPathRoutesManifest` and then you don't need the `|| {}` below.

packages/next/src/server-build.ts
```diff
@@ -1131,6 +1131,23 @@ export async function serverBuild({
...
+    // create .rsc variant for app lambdas and edge functions
+    // to match prerenders so we can route the same when the
+    // __flight__ header is present
+    Object.values(appPathRoutesManifest || {}).forEach(route => {
+      route = path.posix.join('./', route === '/' ? '/index' : route);
+
+      const edgeFunctions = middleware.edgeFunctions as Record<string, any>;
```
cb1kenobi: It looks like `edgeFunctions` could be defined outside before the loop.

packages/next/src/server-build.ts
ijjk: Updated in https://github.com/vercel/vercel/pull/8611/commits/81ad6360e2bb6afa8cb05cd96aa80095bd906420

cb1kenobi (approved):

huozhi (approved):

packages/next/test/fixtures/00-app-dir/vercel.json
```diff
@@ -12,15 +12,26 @@
       "mustContain": "hello from app/dashboard"
     },
     {
-      "path": "/dashboard/another",
+      "path": "/dashboard",
```
styfle, line 15: Lets changes this file from `vercel.json` to `probes.json` and remove the `builds` prop

packages/next/src/server-build.ts
```diff
@@ -1131,6 +1131,23 @@ export async function serverBuild({
     });
   }
 
+  if (appPathRoutesManifest) {
+    // create .rsc variant for app lambdas and edge functions
+    // to match prerenders so we can route the same when the
+    // __flight__ header is present
+    const edgeFunctions = middleware.edgeFunctions as Record<string, any>;
```
styfle: Lets fix the type here so that it always returns `Record<string, EdgeFunction>`

https://github.com/vercel/vercel/blob/7323beea8e52de946756103af76880770717231a/packages/next/src/utils.ts#L2430

styfle: Fixed in [e941a76](https://github.com/vercel/vercel/pull/8611/commits/e941a762dbc28c44fed4ddc1ac949e7c04c02f21)

packages/next/test/fixtures/00-app-dir/vercel.json
ijjk: We need the builds for the integration test still

packages/next/src/server-build.ts
styfle: I might have borked this. Trying again without JS changes, only TS 🕵️

PR #8613, opened 2022-09-22
[tests] Add timeout for broken `actions/setup-node@v3`

styfle:
This PR will make sure that CI fails fast if there is a network issue when restoring the cache.

This has been a known issue for 3 months and no resolution:

- https://github.com/actions/cache/issues/810


cb1kenobi (approved):
Cool, I think I ran into this the other day.

PR #8616, opened 2022-09-22
[cli][dev] Add strict mode to `vc dev` edge function handlers

EndangeredMassa:
Add strict mode to `vc dev` edge function handlers. This is behind a flag in production, but that flag has been at 100% for a while. So, it seems safe to include it here unconditionally.

Also remove legal comments.

These changes bring `vc dev` edge function support closer to production.


EndangeredMassa:
I'll ensure this is targeting `main` after https://github.com/vercel/vercel/pull/8615 merges.

TooTallNate (approved):

cb1kenobi (approved):

PR #8620, opened 2022-09-23
[next] Add caching for static font files

hanneslund:
Adds caching for static font files emitted by `@next/font`.

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -877,7 +877,7 @@ export const build: BuildV2 = async ({
           src: path.posix.join(
             '/',
             entryDirectory,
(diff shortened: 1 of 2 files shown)


styfle (commented):
Seems correct but CI is failing.

I'm also wondering if we should put fonts into the existing `media` directory. Because we used to use `/_next/static/images` but then it was changed to `/_next/static/media` but I don't recall why. 

Maybe @ijjk or @sokra knows the reason.

https://nextjs.org/docs/deployment#:~:text=This%20output%20is%20generated%20inside%20the%20.next%20folder%3A

Update: I found this PR https://github.com/vercel/next.js/pull/30168

hanneslund:
I think that makes sense, closing this and changing the output path. Thanks.
https://github.com/vercel/next.js/pull/40868

PR #8622, opened 2022-09-23
[cli] Add `vercel.json` validation to `vercel build`

styfle:
We were doing this validation in `vercel dev` but not `vercel build`.

This PR adds `vercel.json` validation to `vercel build` too.

Note I am calling this a patch because invalid `vercel.json` was already failing when passed to the API so this allows a nice error message earlier in the process.


packages/cli/src/commands/build.ts
```diff
@@ -54,6 +55,7 @@ import { importBuilders } from '../util/build/import-builders';
 import { initCorepack, cleanupCorepack } from '../util/build/corepack';
 import { sortBuilders } from '../util/build/sort-builders';
 import { toEnumerableError } from '../util/error';
+import { validateConfig } from '../util/dev/validate';
```
TooTallNate: Let's move this file?

TooTallNate:
We need to add validation for `images` config at this point.

packages/cli/src/commands/build.ts
styfle: Done in [f838dfe](https://github.com/vercel/vercel/pull/8622/commits/f838dfe09772061357e6f1e5327be4b5cd961732)

styfle:
I changed the `images` validation to use a schema in [e39de6c](https://github.com/vercel/vercel/pull/8622/commits/e39de6c32e523ea5c36f9e9d877c31a5b81786e1)

packages/cli/src/util/validate-config.ts
```diff
@@ -7,7 +7,7 @@ import {
   rewritesSchema,
   trailingSlashSchema,
 } from '@vercel/routing-utils';
-import { VercelConfig } from './types';
+import { VercelConfig } from './dev/types';
```
TooTallNate, line 10: The version in `dev/types` is just re-exported from client.

```suggestion
import { VercelConfig } from '@vercel/client';
```

TooTallNate (approved):
Good stuff! Thanks 👍

packages/cli/src/util/validate-config.ts
styfle: I found a few of these so I fixed in [af0e12d](https://github.com/vercel/vercel/pull/8622/commits/af0e12de588e5c8b18227d8328027223a700af24)

PR #8623, opened 2022-09-23
[cli] Fix invalid vercel.json config error serialization

styfle:
Follow up to #8622 since we should be throwing errors so the error is correctly serialized to `builds.json`.

### DIFF
--- a/packages/build-utils/src/errors.ts
+++ b/packages/build-utils/src/errors.ts
@@ -61,14 +61,14 @@ export function getPrettyError(obj: {
     }
 
     return new NowBuildError({
-      code: 'DEV_VALIDATE_CONFIG',
+      code: 'INVALID_LOCAL_CONFIG',
       message: message,
       link: prop ? `${docsUrl}#project/${prop.toLowerCase()}` : docsUrl,
       action: 'View Documentation',
     });
   } catch (e) {
     return new NowBuildError({
-      code: 'DEV_VALIDATE_CONFIG',
+      code: 'INVALID_LOCAL_CONFIG',
       message: `Failed to validate configuration.`,
       link: docsUrl,
       action: 'View Documentation',
--- a/packages/build-utils/src/errors.ts
+++ b/packages/build-utils/src/errors.ts
@@ -61,14 +61,14 @@ export function getPrettyError(obj: {
     }
 
     return new NowBuildError({
-      code: 'INVALID_LOCAL_CONFIG',
+      code: 'invalid_vercel_config',
       message: message,
       link: prop ? `${docsUrl}#project/${prop.toLowerCase()}` : docsUrl,
       action: 'View Documentation',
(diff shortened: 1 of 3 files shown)


packages/build-utils/src/errors.ts
```diff
@@ -61,14 +61,14 @@ export function getPrettyError(obj: {
     }
 
     return new NowBuildError({
-      code: 'DEV_VALIDATE_CONFIG',
+      code: 'INVALID_LOCAL_CONFIG',
```
TooTallNate: Convention is usually "noun - verb" for error naming. How about this?

```suggestion
      code: 'VERCEL_CONFIG_INVALID',
```

styfle: When I search this repo, I found many existing error codes that begin with `INVALID_` such as

```
invalid_build_specification
INVALID_DOMAIN
INVALID_DEPLOYMENT_ID
INVALID_ALIAS
invalid_email
invalid_function
INVALID_PIPFILE_LOCK 
```

TooTallNate (approved):

packages/build-utils/src/errors.ts
styfle: I'll change it to `invalid_vercel_config` [126695e](https://github.com/vercel/vercel/pull/8623/commits/126695e7a6b78e98afce808e7dc01052fc244b5e)

packages/build-utils/src/errors.ts
```diff
@@ -61,14 +61,14 @@ export function getPrettyError(obj: {
     }
 
     return new NowBuildError({
-      code: 'DEV_VALIDATE_CONFIG',
+      code: 'invalid_vercel_config',
```
cb1kenobi: Aw, I liked `INVALID_VERCEL_CONFIG` better.

cb1kenobi (approved):

packages/build-utils/src/errors.ts
styfle: Sadly, this code is already used for a different error here: https://github.com/vercel/vercel/blob/cfae7ec3c291d1802fb6eec6642a0ebf792fb71d/packages/cli/src/util/errors.ts#L38

cb1kenobi: Yeah, I accidentally copy-n-pasted `INVALID_LOCAL_CONFIG` when I meant `INVALID_VERCEL_CONFIG` as you had it. I just like my `code` in `CAPS`.

styfle: I guess Node.js error codes are uppercase so lets keep it uppercase [e4a89b7](https://github.com/vercel/vercel/pull/8623/commits/e4a89b72d967a08d0e28799c6c6b067067018d7a)

cb1kenobi (approved):

PR #8629, opened 2022-09-25
[next] read `regions` from middleware-manifest output

Schniz:
### Related Issues

- Needs https://github.com/vercel/next.js/pull/40881 for this to be useful
- Resolves EC-238

#### Tests

#### Code Review

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2202,7 +2202,10 @@ interface BaseEdgeFunctionInfo {
   page: string;
   wasm?: { filePath: string; name: string }[];
   assets?: { filePath: string; name: string }[];
-  userConfig?: Record<string, unknown>;
+  userConfig?: {
+    regions?: 'auto' | string[] | 'all' | 'default';
+    [key: string]: unknown;
+  };
 }
 
 interface EdgeFunctionInfoV1 extends BaseEdgeFunctionInfo {
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2202,6 +2202,7 @@ interface BaseEdgeFunctionInfo {
   page: string;
   wasm?: { filePath: string; name: string }[];
   assets?: { filePath: string; name: string }[];
+  userConfig?: Record<string, unknown>;
 }
 
 interface EdgeFunctionInfoV1 extends BaseEdgeFunctionInfo {
@@ -2341,6 +2342,11 @@ export async function getMiddlewareBundle({
                   ...wasmFiles,
                   ...assetFiles,
                 },
+                regions:
+                  typeof edgeFunction.userConfig === 'object' &&
+                  'regions' in edgeFunction.userConfig
+                    ? (edgeFunction.userConfig.regions as any)
+                    : 'auto',
                 entrypoint: 'index.js',
                 envVarsInUse: edgeFunction.env,
                 assets: (edgeFunction.assets ?? []).map(({ name }) => {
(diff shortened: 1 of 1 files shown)


packages/next/src/utils.ts
```diff
@@ -2341,6 +2342,11 @@ export async function getMiddlewareBundle({
                   ...wasmFiles,
                   ...assetFiles,
                 },
+                regions:
+                  typeof edgeFunction.userConfig === 'object' &&
+                  'regions' in edgeFunction.userConfig
+                    ? (edgeFunction.userConfig.regions as any)
+                    : 'auto',
```
nuta: Why not `edgeFunction.userConfig?.regions ?? 'auto'`? Do we need to accept `undefined` as a valid value?

Schniz: Maybe I was too strict here, I wanted to make sure that `edgeFunction.userConfig` is indeed an object, but you're right. I'll change <picture data-single-emoji=":relieved-telegram:" title=":relieved-telegram:">:relieved-telegram:</picture>

packages/next/src/utils.ts
```diff
@@ -2202,6 +2202,7 @@ interface BaseEdgeFunctionInfo {
   page: string;
   wasm?: { filePath: string; name: string }[];
   assets?: { filePath: string; name: string }[];
+  userConfig?: Record<string, unknown>;
```
styfle: Can we add regions to the type here?

packages/next/src/utils.ts
```diff
@@ -2341,6 +2342,7 @@ export async function getMiddlewareBundle({
                   ...wasmFiles,
                   ...assetFiles,
                 },
+                regions: (edgeFunction.userConfig?.regions ?? 'auto') as any,
```
styfle: Do you need to update build-utils too?

packages/next/src/utils.ts
Schniz: Sure

Schniz: I am torn between having `unknown` aka “I don’t trust the input and must parse everything” and having the expected type and trusting the input 🥲

packages/next/src/utils.ts
Schniz: It was already updated I believe? Typecheck passes

styfle: Why `as any`?

packages/next/src/utils.ts
styfle: In that case, we could throw here if the shape is invalid. Another option is to pass it through to the API and let it respond with 400 if the payload is invalid.

packages/next/src/utils.ts
Schniz: Because `unknown` does not match the expected type. Related to the other conversation

packages/next/src/utils.ts
Schniz: The latter is already implemented, so I’ll type it as the API expects

```suggestion
  userConfig?: {
    regions?: 'auto' | string[] | 'all' | 'default';
    [key: string]: unknown;
  };
```

This will make TS happy and then you can remove `as any` below

styfle (approved):

ijjk (changes requested):
Holding for changes as discussed in https://vercel.slack.com/archives/C03S8ED1DKM/p1664188823380889

Schniz:
@ijjk it's <picture data-single-emoji=":done:" title=":done:">:done:</picture> can you please re-review 😁

ijjk (approved):

PR #8633, opened 2022-09-26
[go] remove unused, breaking watchlist from Go builder

EndangeredMassa:
Generating the list of files to watch was causing Go serverless functions to fail in some situations. The file watch list isn't being used anyway. So this PR removes it from the Go builder.

A test was added for the failing scenario:
- entrypoint "api/one.go" has a handler name like `Handler2`
- second entrypoint "api/two.go" has a handler name that's a subset of the first, like `Handler`
- when hitting `/api/one`, the Go build fails because part of `anaylze.go` was looking up related files based on `Handler`

I can't tell exactly why this happens only when the handler names are similar, but it does.


packages/go/util/analyze.go
```diff
@@ -104,84 +103,6 @@ func main() {
...
-				if strings.Contains(se, ed) {
-					// find relative path of related file
-					var basePath string
-					if modPath == "" {
-						basePath = filepath.Dir(fileName)
-					} else {
-						basePath = modPath
-					}
-
-					rel, err := filepath.Rel(basePath, file)
```
EndangeredMassa, line 176: This `basePath` is a relative path, but it should be absolute. Rather than fix this and wait to discover other edge cases in this unused logic, it has been removed.

packages/go/util/analyze.go
```diff
@@ -104,84 +103,6 @@ func main() {
...
-	// Add entrypoint to watchlist
-	relFileName, err := filepath.Rel(filepath.Dir(fileName), fileName)
-	if err != nil {
-		log.Fatal(err)
-	}
-	relatedFiles = append(relatedFiles, relFileName)
-
-	// looking for all go files that have export func
-	// using in entrypoint
-	err = filepath.Walk(filepath.Dir(fileName), visit(&files))
```
styfle, line 121: I'm wondering if we should keep this logic, not for file watching but for tracing the ast to find `readFile()` calls the same way we do in `@vercel/node`.

cc @TooTallNate since we recently found out `@vercel/go` doesnt include assets in the lambda

EndangeredMassa: I suggest we make a card to fix that behavior and restore this logic when it can be adapted for that purpose. The changes as written fixes a different issue.

styfle (approved):

packages/go/util/analyze.go
EndangeredMassa: I made a draft card for this in our board at the top of the backlog.

PR #8634, opened 2022-09-26
[tests] replace spinner messages with normal output during tests

EndangeredMassa:
Convert spinner output to simple prints during test runs. This makes it easier to write tests against the output of commands.


packages/cli/test/unit/commands/list.test.ts
```diff
@@ -38,11 +38,11 @@ describe('list', () => {
 
       await list(client);
 
-      const output = await readOutputStream(client, 4);
```
EndangeredMassa, line 41: Because of the timing of the spinner used in these areas of the code, the message wasn't showing up in the tested output before. Not that it is, these line indexes had to be adjusted.

Before:

```
> Deployments for with-team under caroline [1ms]
> To list deployments for a project, run `vercel ls [project]`.

  Age      Deployment            Status      Duration
  44ms     https://pubafe.aw     ● Ready     30s
```

After:

```
Retrieving project…
Fetching deployments in ina
> Deployments for with-team under ina [1ms]
> To list deployments for a project, run `vercel ls [project]`.

  Age      Deployment         Status      Duration
  44ms     https://zaz.im     ● Ready     30s
```

TooTallNate (approved):

PR #8657, opened 2022-09-30
[next] Use OS path separator to match serverless file references

cb1kenobi:
When running `vc build` for a Next.js app, the Next builder will execute the server build which performs several steps. One of the steps is to trace each serverless function for any referenced files, then the raw list of files is scrubbed and filtered. The filtering uses OS specific file path comparisons to see if a file is of interest. Since it's comparing OS specific paths, we need to use OS specific path separators.

During testing on Windows, the traced serverless functions file list was always empty.

#### Tests

#### Code Review


EndangeredMassa (approved):
Is there a test we can add for this?

styfle (commented):
How about a test?

cb1kenobi:
Good idea. Here's the test: https://github.com/vercel/vercel/pull/8661.

PR #8661, opened 2022-09-30
[tests] Adding test for next builder OS path separator for serverless file refs

cb1kenobi:
Here's the test for https://github.com/vercel/vercel/pull/8657.

#### Tests

#### Code Review


packages/next/test/unit/fixtures/01-normalize-paths/index.test.js
```diff
@@ -38,9 +39,15 @@ describe(`${__dirname.split(path.sep).pop()}`, () => {
...
-      );
+      if (page === 'index') {
+        const { files, type } = output[page];
+        expect(type).toEqual('Lambda');
+        expect(files).toHaveProperty([path.join('data', 'strings.json')]);
+      } else {
+        expect(path.resolve(output[page].fsPath)).toEqual(
+          path.join(pagesDir, `${page}.html`)
+        );
+      }
```
styfle, line 50:
```suggestion
      expect(path.resolve(output[page].fsPath)).toEqual(
        path.join(pagesDir, `${page}.html`)
      );
```

Remove the `else`? Seems like this should always be true

cb1kenobi: I don't think will work. When `page` is `index`, `output[page]` is a `NodejsLambda` instance which does not have a `fsPath` property.

cb1kenobi: Gave it a try and indeed `fsPath` is undefined:

```
    TypeError: The "path" argument must be of type string. Received undefined

      45 |         expect(files).toHaveProperty([path.join('data', 'strings.json')]);
      46 |       }
    > 47 |       expect(path.resolve(output[page].fsPath)).toEqual(
         |                   ^
      48 |         path.join(pagesDir, `${page}.html`)
      49 |       );
      50 |     }

      at Object.<anonymous> (test/unit/fixtures/01-normalize-paths/index.test.js:47:19)
```

styfle (approved):

packages/next/test/unit/fixtures/01-normalize-paths/index.test.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs-extra');
 const ms = require('ms');
 const path = require('path');
-const { build } = require('../../../../src');
+const { build } = require('../../../../dist');
```
styfle, line 4: Was this change intended?

cb1kenobi: Yes. I don't know why it worked before, but if I leave it as `src`, the tests fail with:

```
ENOENT: no such file or directory, open 'C:\Users\chris\projects\vercel\packages\next\src\server-launcher.js'
```

So, I changed it to `dist` and it works as long as you `yarn run build` before running the test. Do you have any idea why?

ijjk (approved):

packages/next/test/unit/fixtures/01-normalize-paths/index.test.js
styfle: Probably because this is a `.test.js` instead of `.test.ts`

cb1kenobi: Just for fun, I renamed the test to `.ts`, changed the `require()`s to imports, changed `dist` to `src`, and it still blows up with the same "ENOENT: no such file or directory" error above.

The code that references `server-launcher.js` is in `server-build.ts`:

```
    const launcherData = await fs.readFile(
      path.join(__dirname, 'server-launcher.js'),
      'utf8'
    );
```

`server-launcher.js` will only ever exist in the `dist` directory, so it doesn't make any sense why it apparently worked before. In any case, I think referencing `dist` is required.

styfle: I prefer `dist` anyway. Best to test the code we distribute because thats the code that users run.

PR #8669, opened 2022-10-03
[cli] Fix `vc build` to error early when runtime is discontinued

styfle:
This moves an existing error from the build container to `vercel build`.

Its rare, but [Vercel Runtimes](https://vercel.com/docs/runtimes) might target a discontinued [AWS Lambda Runtime](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) so we should fail fast when we know this has happened in `vercel build`.

A test has been added to demonstrate the failure using an old PHP version.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -3,6 +3,7 @@ import chalk from 'chalk';
 import dotenv from 'dotenv';
 import { join, normalize, relative, resolve } from 'path';
 import {
+  getDiscontinuedNodeVersions,
   normalizePath,
   Files,
   FileFsRef,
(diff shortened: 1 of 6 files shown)


TooTallNate (approved):

packages/cli/src/commands/build.ts
```diff
@@ -467,6 +468,25 @@ async function doBuild(
       );
       const buildResult = await builder.build(buildOptions);
 
+      if (
+        buildResult &&
+        'output' in buildResult &&
+        'runtime' in buildResult.output &&
+        'type' in buildResult.output &&
+        buildResult.output.type === 'Lambda'
```
cb1kenobi, line 476: `buildResult` should always be ok since the code currently doesn't seem to check if it's not ok. If `buildResult` could possibly be not ok, then it should be asserted.

Additionally, optional chaining may make the code more readable:

```
if (buildResult.runtime && buildResult.output?.type === 'Lambda') {
```

styfle: TS isnt smart enough to use your solution so it needs some help with the type guards. I think if we bump to latest TS it might be slightly better

PR #8671, opened 2022-10-03
[tests] Replace `cancel-workflow-action` with native `cancel-in-progress`

styfle:
This removes the `styfle/cancel-workflow-action` in favor of native GitHub Actions `cancel-in-progress`.

The cancel key is workflow+branch but we don't want to cancel on the `main` branch.

https://docs.github.com/en/actions/using-jobs/using-concurrency


styfle:
You can see it working by checking the first commit and then the second commit. Since the second commit was pushed while the first was in progress, the first was canceled.

ijjk (approved):

PR #8677, opened 2022-10-04
[cli] Improve error message when using legacy `@now` builders

styfle:
This PR adds a helpful error message when using a legacy `@now` builder.

This mimics the behavior of the build-container.


TooTallNate (approved):

cb1kenobi (approved):

styfle:
Adding the new test was causing others to fail.

- New test [passes](https://github.com/vercel/vercel/actions/runs/3972382914/jobs/6810233278#step:10:3557)
- Unrelated test [fails](https://github.com/vercel/vercel/actions/runs/3972382914/jobs/6810233278#step:10:3631)

Removing the new test works and now all tests pass again, likely because `@now/node-server` is modifying files outside its own working directory.

PR #8681, opened 2022-10-04
[build-utils] Allow file-ref sema to be controlled through env flag

pvdz:
My IDE tells me `process` is unknown but mentions something about package.json so that may just be a superficial issue. I guess CI/CD will tell me soon enough.

This adds an env flag to override the file ref sema's so we can experiment with setting a higher sema.

One potential problem I'm seeing is that this is a generic sema for all the things that use this class. Not sure if that's going to work out as intended but in that case we'll have to find a different way :)

### DIFF
--- a/packages/build-utils/src/file-fs-ref.ts
+++ b/packages/build-utils/src/file-fs-ref.ts
@@ -5,7 +5,9 @@ import path from 'path';
 import Sema from 'async-sema';
 import { FileBase } from './types';
 
-const semaToPreventEMFILE = new Sema(20);
+const semaToPreventEMFILE = new Sema(
+  parseInt(process.env.VERCEL_INTERNAL_FILE_FS_REF_SEMA, 10) || 20
+);
 
 interface FileFsRefOptions {
   mode?: number;
--- a/packages/build-utils/src/file-fs-ref.ts
+++ b/packages/build-utils/src/file-fs-ref.ts
@@ -5,8 +5,12 @@ import path from 'path';
 import Sema from 'async-sema';
 import { FileBase } from './types';
 
+const DEFAULT_SEMA = 20;
 const semaToPreventEMFILE = new Sema(
-  parseInt(process.env.VERCEL_INTERNAL_FILE_FS_REF_SEMA, 10) || 20
+  parseInt(
+    process.env.VERCEL_INTERNAL_FILE_FS_REF_SEMA || String(DEFAULT_SEMA),
+    10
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/file-fs-ref.ts
```diff
@@ -5,7 +5,9 @@ import path from 'path';
 import Sema from 'async-sema';
 import { FileBase } from './types';
 
-const semaToPreventEMFILE = new Sema(20);
+const semaToPreventEMFILE = new Sema(
+  parseInt(process.env.VERCEL_INTERNAL_FILE_FS_REF_SEMA, 10) || 20
+);
```
styfle:
```suggestion
const semaToPreventEMFILE = new Sema(
  parseInt(process.env.VERCEL_INTERNAL_FILE_FS_REF_SEMA || '20')
);
```

packages/build-utils/src/file-ref.ts
```diff
@@ -12,7 +12,7 @@ interface FileRefOptions {
   mutable?: boolean;
 }
 
-const semaToDownloadFromS3 = new Sema(5);
+const semaToDownloadFromS3 = new Sema(parseInt(process.env.VERCEL_INTERNAL_FILE_REF_SEMA,10), 5);
```
styfle:
```suggestion
const semaToDownloadFromS3 = new Sema(parseInt(process.env.VERCEL_INTERNAL_FILE_REF_SEMA || '5'));
```

styfle (commented):
I think the suggestion I made will make TS happy

packages/build-utils/src/file-fs-ref.ts
pvdz: Not sure why TS was yelling but the second arg to `parseInt` is the radix and it's a best practice to always use it to prevent unexpected behavior for some inputs. The result should be a number so I'll check the error in detail in the morning.

styfle: The error is that the env var can be undefined and it doesnt like that

packages/build-utils/src/file-ref.ts
pvdz: fixed

packages/build-utils/src/file-fs-ref.ts
pvdz: yup. fixed

styfle (approved):

PR #8687, opened 2022-10-05
[node] Update `edge-runtime` to v1.1.0-beta.37

TooTallNate:
Fixes error:

```
ENOENT: no such file or directory, open 'querystring'
```

Unfortunately this issue would only manifest when installed externally. I.e. our tests didn't catch this since the `querystring` module is presumably present in the monorepo.


styfle (approved):

cb1kenobi (approved):

styfle:
> Unfortunately this issue would only manifest when installed externally. I.e. our tests didn't catch this since the querystring module is presumably present in the monorepo.

Might be time to switch to `pnpm` since it should catch these types of errors.

PR #8688, opened 2022-10-05
[fs-detectors] Add setting placeholders for monorepos

chloetedder:
### Related Issues

Add in placeholder settings for monorepos

#### Tests

#### Code Review


TooTallNate (approved):

api/_lib/script/build.ts
```diff
@@ -18,6 +18,12 @@ async function main() {
     { recursive: true, force: true }
   );
 
+  await fs.cp(
+    join(repoRoot, 'packages', 'fs-detectors', 'src', 'monorepos', 'logos'),
```
styfle: Lets move this out of `src` since these images are not source files

agadzik (approved):

packages/frameworks/src/types.ts
```diff
@@ -135,11 +135,11 @@ export interface Framework {
     /**
      * Default Build Command or a placeholder
      */
-    buildCommand: SettingValue;
+    buildCommand: Setting;
     /**
      * Default Development Command or a placeholder
      */
-    devCommand: SettingValue;
+    devCommand: Setting;
```
styfle: This is causing TS to fail. Any reason why it was changed?

chloetedder: Yes - but I updated this to just return `null` for the value because it shouldn't be defined here because it is project based

api/_lib/script/build.ts
chloetedder: 👍

styfle (approved):

PR #8718, opened 2022-10-11
[cli][next] Removed --forceExit from jest args

cb1kenobi:
Remove `--forceExit` and wait for the child `vc dev` process to close before continuing.

Listen to the `'close'` event instead of `'exit'` to ensure stdio has been flushed and closed before continuing.

Some tests cause `vc dev` to spawn child processes that in turn spawn nested child processes that do not exit when the parent exits, so a `nukeProcessTree()` helper was added to ensure all spawned processes are cleaned up. Furthermore, some of the nested child processes don't respond well to `SIGTERM` and we need to use `SIGKILL` if `SIGTERM` doesn't do the job.

I uncovered a larger issue exposed by removing the `--forceExit` flag. The go builder uses `go run` to build and run the go-based serverless function. Turns out that `go run` will build the executable, then spawn it with a parent process id (ppid) of `1` (launchd) on macOS. This means that the go serverless function executable is not detected as a child process of `vc dev` and won't be cleaned up, yet has inherited `stdout` and `stderr` keeping Jest alive. The solution is to explicitly `go build` the serverless executable, then run it.

#### Tests

#### Code Review


packages/cli/src/util/dev/server.ts
```diff
@@ -211,6 +211,8 @@ export default class DevServer {
     this.podId = Math.random().toString(32).slice(-5);
 
     this.devServerPids = new Set();
+
+    process.on('SIGTERM', () => this.stop());
```
TooTallNate: I suggest doing this in `commands/dev/dev.ts` instead.

packages/cli/test/dev/utils.js
```diff
@@ -260,8 +275,11 @@ async function testFixture(directory, opts = {}, args = []) {
   });
 
   dev._kill = dev.kill;
-  dev.kill = async (...args) => {
-    dev._kill(...args);
+  dev.kill = async () => {
+    // kill the entire process tree for the child as some tests
+    // spawn child processes that we don't have the pid for
+    await nukeProcessTree(dev.pid);
```
TooTallNate, line 281: I don't think this is necessary anymore, since `stop()` should be taking care of teardown when the main dev process is killed.

packages/static-build/package.json
```diff
@@ -48,5 +48,8 @@
     "node-fetch": "2.6.7",
     "rc9": "1.2.0",
     "typescript": "4.3.4"
+  },
+  "dependencies": {
+    "tree-kill": "1.2.2"
```
TooTallNate: Should be a dev dependency since we ncc.

packages/cli/test/dev/utils.js
cb1kenobi: Turns out this `nukeProcessTree()` is still needed. For whatever reason when we run the `invalid-env-var-name` integration test on Linux and it spawns `vc dev` with `{ shell: true }`, killing the process leaves the shell process in a "defunct" state. I guess "defunct" processes are a Linux thing and happen when the process is corrupt, killed, or in this case has a child process `start.js` that hasn't exited.

TooTallNate (approved):

Ethan-Arrowood (approved):

packages/cli/test/dev/utils.js
```diff
@@ -490,30 +508,94 @@ function testFixtureStdio(
...
+      pid = parseInt(pid);
+      const recurse = Object.prototype.hasOwnProperty.call(pids, pid);
+      pids[parentPid].push(pid);
+      pids[pid] = [];
+      if (recurse) {
+        await ps(pid, pids);
+      }
+    }
+  } catch (e) {
+    // squelch
```
EndangeredMassa: **suggestion (non-blocking):** debug log error

Can we debug log this error?

packages/cli/test/dev/utils.js
```diff
@@ -490,30 +508,94 @@ function testFixtureStdio(
...
+    return;
+  }
+
+  await sleep(250);
+
+  try {
+    // check if killed
+    process.kill(pid, 0);
+  } catch (e) {
+    return;
```
EndangeredMassa, line 564: **suggestion (non-blocking):** debug log error

Can we debug log this error?

packages/cli/test/dev/utils.js
```diff
@@ -490,30 +508,94 @@ function testFixtureStdio(
...
+    console.log(`pid ${pid} won't die, giving up`);
+    return;
+  }
+
+  // kill the process
+  try {
+    process.kill(pid, signal);
+  } catch (e) {
+    // process does not exist
+    return;
```
EndangeredMassa, line 554: **suggestion (non-blocking):** debug log error

Can we debug log this error?

EndangeredMassa (approved):
🎉

packages/cli/test/dev/utils.js
cb1kenobi: Fixed!

packages/cli/test/dev/utils.js
cb1kenobi: Fixed!

packages/cli/test/dev/utils.js
cb1kenobi: Fixed!

PR #8720, opened 2022-10-11
[build-utils] Cleanup getNodeVersion

pvdz:
- `meta.isDev` is unconditionally read after the first check (even though it could technically still be `null`, ignoring typescript)
- `meta.isDev` is explicitly asserted to be falsy, since the first `if` will return early if it's truthy, so the later checks are all redundant
- added `?.` for the `packageJson?.engines?.node` read

The rest is auto formatting (and if my local eslint warnings are correct then I'm just gonna close this PR...)


TooTallNate (approved):

styfle (approved):

pvdz:
Guess lint is happy. This repo will be a ball of fun once the style guide kicks in, eh.

PR #8733, opened 2022-10-17
[node] allow to extract Edge Function regions from source code

Schniz:
This allows to export `config = { runtime: 'auto' | 'all' | 'default' | string[] }` for edge functions.

### Related Issues

- Resolves [EC-156](https://linear.app/vercel/issue/EC-156/make-nodejs-builder-extract-preferred-regions-from-source-code)

#### Tests

#### Code Review

### DIFF
--- a/packages/hydrogen/src/build.ts
+++ b/packages/hydrogen/src/build.ts
@@ -125,10 +125,7 @@ export const build: BuildV2 = async ({
         const project = new Project();
         const config = getConfig(project, edgeFunctionFiles['index.js'].fsPath);
         return config?.regions;
-      } catch (err) {
-        console.warn(
-          `Could not parse the entrypoint file to get the regions: ${err}`
-        );
+      } catch {
         return undefined;
       }
     })(),
--- a/packages/hydrogen/src/build.ts
+++ b/packages/hydrogen/src/build.ts
@@ -15,6 +15,8 @@ import {
   scanParentDirs,
 } from '@vercel/build-utils';
 import type { BuildV2, PackageJson } from '@vercel/build-utils';
+import { getConfig } from '@vercel/static-config';
+import { Project } from 'ts-morph';
 
 export const build: BuildV2 = async ({
   entrypoint,
@@ -118,6 +120,18 @@ export const build: BuildV2 = async ({
     deploymentTarget: 'v8-worker',
     entrypoint: 'index.js',
     files: edgeFunctionFiles,
+    regions: (() => {
+      try {
+        const project = new Project();
+        const config = getConfig(project, edgeFunctionFiles['index.js'].fsPath);
+        return config?.regions;
+      } catch (err) {
+        console.warn(
+          `Could not parse the entrypoint file to get the regions: ${err}`
+        );
+        return undefined;
+      }
+    })(),
   });
 
   // The `index.html` file is a template, but we want to serve the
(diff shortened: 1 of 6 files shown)


javivelasco (approved):

packages/node/src/index.ts
```diff
@@ -454,6 +454,8 @@ export const build: BuildV3 = async ({
       entrypoint: handler,
       files: preparedFiles,
```
styfle: This needs to be added to every builder that calls `new EdgeFunction()`:

- `@vercel/node`
- `@vercel/next`
- `@vercel/hydrogen`
- `@vercel/static-build`

packages/node/src/index.ts
```diff
@@ -454,6 +454,8 @@ export const build: BuildV3 = async ({
...
 
+      ...(staticConfig?.regions && { regions: staticConfig?.regions }),
```
```suggestion
      regions: staticConfig?.regions
```

Also, I think you can simplify this

Schniz:
@styfle 

* `@vercel/node` is the one I added first and it works 😺 
* `@vercel/next` was implemented and merged previously <picture data-single-emoji=":just-right2:" title=":just-right2:">:just-right2:</picture> 
* `@vercel/hydrogen` was implemented and tested too (with `hydrogen build && echo "..." >> dist/worker/index.js`)
* `@vercel/static-build` was updated but it shouldn't really affect it. Anyway should work <picture data-single-emoji=":relieved-telegram:" title=":relieved-telegram:">:relieved-telegram:</picture> if you have an idea how I can test it lmk

packages/hydrogen/src/build.ts
```diff
@@ -118,6 +120,18 @@ export const build: BuildV2 = async ({
...
     files: edgeFunctionFiles,
+    regions: (() => {
+      try {
+        const project = new Project();
+        const config = getConfig(project, edgeFunctionFiles['index.js'].fsPath);
+        return config?.regions;
+      } catch (err) {
+        console.warn(
+          `Could not parse the entrypoint file to get the regions: ${err}`
+        );
```
styfle: Lets remove this warning since its not actionable by the user

styfle (approved):

packages/hydrogen/src/build.ts
Schniz:
```suggestion
      } catch {
```

packages/hydrogen/src/build.ts
```diff
@@ -118,6 +120,15 @@ export const build: BuildV2 = async ({
     deploymentTarget: 'v8-worker',
     entrypoint: 'index.js',
     files: edgeFunctionFiles,
+    regions: (() => {
+      try {
+        const project = new Project();
+        const config = getConfig(project, edgeFunctionFiles['index.js'].fsPath);
+        return config?.regions;
+      } catch {
+        return undefined;
```
EndangeredMassa, line 129: Can we debug log the warning that was here? I imagine trying to debug an issue around regions not working and having nothing that points me to this chunk of code.

Schniz: I can add this in a different PR but tbh hydrogen being a single edge function deployment might not be the best approach for regional invocations? <picture data-single-emoji=":surprised-wide-eyed:" title=":surprised-wide-eyed:">:surprised-wide-eyed:</picture>

PR #8735, opened 2022-10-17
[cli] Allow `ls` to specify a limit up to 100

andymckay:
### Related Issues

Allow the `ls` commands to have a `--limit` option that allows a user to fetch up to 100 records per page. Currently the default is 20, but the API allows up to 100. This keeps the default, if unspecified at 20. 

Fixes: https://linear.app/vercel/issue/VCCLI-244/add-limit-option-all-the-ls-subcommands

This adds in `ls --limit` into:
- [x] `alias ls`
- [x] `certs ls`
- [x] `dns ls`
- [x] `domains ls`

I note that `env` has an `ls` command, but it doesn't have a pagination command and [looking at the API](https://vercel.com/docs/rest-api#endpoints/projects/retrieve-the-environment-variables-of-a-project-by-id-or-name) it doesn't support pagination or limit.

Wasn't sure if I should add in `-L` as a short cut to `--limit`, seems like a good idea.

~Couldn't find any tests that cover this API, but I could be looking in the wrong place, this is the first pull request, so my apologies if I've missed it. But I'll take another look tomorrow and leave it in the draft state for now.~

Added in unit tests for each of the commands and mocks for those unit tests, which is has caused this PR to get quite a bit larger, sorry about that.

Of note for reviewers, there were a few places where I changed `console.log` to `output.log` to get the output passed to the tests - as far as I can tell everything works on the command line and in tests.

#### Tests

#### Code Review


packages/cli/src/commands/certs/ls.ts
```diff
@@ -38,10 +40,20 @@ async function ls(
     return 1;
   }
 
+  if (
+    typeof nextTimestamp !== undefined &&
+    (Number.isNaN(limit) || limit! > 100)
```
TooTallNate: We try to avoid using `!` when possible:

```suggestion
    typeof nextTimestamp !== undefined &&
    typeof limit === 'number' &&
    (Number.isNaN(limit) || limit > 100)
```

andymckay: Thanks 🙇🏾  there's a few of those, I'll update across the whole PR.

packages/cli/src/commands/alias/ls.ts
```diff
@@ -22,13 +23,23 @@ export default async function ls(
...
+  const { '--limit': limit } = opts;
   const { contextName } = await getScope(client);
 
   if (typeof nextTimestamp !== undefined && Number.isNaN(nextTimestamp)) {
     output.error('Please provide a number for flag --next');
     return 1;
   }
 
+  if (
+    typeof nextTimestamp !== undefined &&
```
EndangeredMassa: **suggestion (blocking):** Remove the check for `nextTimestamp`

The current condition `typeof nextTimestamp !== undefined` will always be true because if `nextTimestamp` is undefined, `typeof nextTimestamp` is the string `"undefined"`, not the value `undefined`.

I don't think we want to require `--next` for this `--limit` validation to show an error. So, I think we should remove this check entirely from each occurrence in this PR.

packages/cli/src/commands/alias/index.ts
```diff
@@ -38,6 +38,7 @@ const help = () => {
     -S, --scope                         Set a custom scope
     -N, --next                          Show next page of results
     -y, --yes                           Skip the confirmation prompt when removing an alias
+    --limit                             Number of results to return per page
```
EndangeredMassa: **suggestion (non-blocking):** add default and max

Should we add the default and max limit values for each of these commands? We could say:

```suggestion
    --limit                             Number of results to return per page (default 20, max 100)
```

packages/cli/test/unit/commands/alias.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    await expect(client.stderr).toOutput('dummy-1.app');
+  });
+
+  it('should throw an error if limit not valid', async () => {
+    useUser();
+    for (let limit of ['abc', '101']) {
+      client.setArgv('alias', 'ls', '--limit', limit);
+      const exitCodePromise = alias(client);
+      await expect(exitCodePromise).resolves.toEqual(1);
+    }
```
EndangeredMassa: **suggestion (non-blocking):** split this into two tests

I'm generally not a fan of generating tests at runtime. It makes them easier to write, sometimes harder to read, and often harder to debug.

packages/cli/src/commands/certs/ls.ts
```diff
@@ -38,10 +40,21 @@ async function ls(
...
+
   // Get the list of certificates
-  const { certs, pagination } = await getCerts(client, nextTimestamp).catch(
-    err => err
-  );
+  const { certs, pagination } = await getCerts(
+    client,
+    nextTimestamp,
+    limit
+  ).catch(err => err);
```
EndangeredMassa, line 49: **suggestion (non-blocking):** remove `.catch(err => err)`

I know this wasn't changed in the PR, but while we're here, can re remove this exception handling `.catch(err => err)`? If `getCerts` throws, it looks like the result will be `certs` and `pagination` will be undefined, because those don't exist on an `Error` instance. Then we'll get a failure below when accessing `certs.length`, obfuscating the root error.

This also updates the types for `certs` and `pagination` here to be their true types instead of `any`, giving us more type protection.

EndangeredMassa (changes requested):
I think the `nextTimestamp` check should be removed. It's not really a bug because it essentially adds a `true AND ...` to the condition, but it will probably confuse someone later on.

packages/cli/src/commands/alias/ls.ts
andymckay: Oops as you pointed out, `nextTimestamp` shouldn't even be in this check because this is a check for limit. Sorry about that and thanks for spotting.

packages/cli/src/commands/alias/index.ts
andymckay: Good idea, thanks!

packages/cli/src/commands/certs/ls.ts
andymckay: Would you mind if I filed an issue for that and did that in a seperate PR? This one has been dragging much longer than I originally hoped and smaller focused PRs are easier to write and merge.

EndangeredMassa: Certainly!

packages/cli/src/commands/alias/index.ts
```diff
@@ -38,6 +38,7 @@ const help = () => {
     -S, --scope                         Set a custom scope
     -N, --next                          Show next page of results
     -y, --yes                           Skip the confirmation prompt when removing an alias
+    --limit                             Number of results to return per page (default: 20, max: 100)
```
cb1kenobi: Seems the convention is if the option has a value, specify a `=HINT`  with the hint being bold and underlined. The hint usual describes the thing or data type, so the hint could be `=NUM` or `=COUNT` or `=VALUE` or whatever you want.

```suggestion
    --limit${chalk.bold.underline(
    'VALUE'
  )}                             Number of results to return per page (default: 20, max: 100)
```

packages/cli/src/commands/alias/ls.ts
```diff
@@ -22,13 +23,19 @@ export default async function ls(
 ) {
   const { output } = client;
   const { '--next': nextTimestamp } = opts;
+  const { '--limit': limit } = opts;
```
cb1kenobi: I know you're copying `--next`, but passing `--next` and `--limit` around seems wrong. It would be cool if those were resolved into `nextTimestamp` and `limit`, then passed into `ls()`.

packages/cli/src/commands/alias/ls.ts
```diff
@@ -22,13 +23,19 @@ export default async function ls(
...
   const { contextName } = await getScope(client);
 
   if (typeof nextTimestamp !== undefined && Number.isNaN(nextTimestamp)) {
     output.error('Please provide a number for flag --next');
     return 1;
   }
 
+  if (typeof limit === 'number' && (Number.isNaN(limit) || limit > 100)) {
```
cb1kenobi: Probably should also check that limit > 0.

cb1kenobi (changes requested):

packages/cli/src/commands/alias/ls.ts
andymckay: Yeah and maybe also decimals. The API is actually interesting what it does with the input:

```
URL: https://api.vercel.com/v3/now/aliases?limit=-5
Status Code: 200
Results: 1
URL: https://api.vercel.com/v3/now/aliases?limit=0
Status Code: 200
Results: 10
URL: https://api.vercel.com/v3/now/aliases?limit=0.1
Status Code: 200
Results: 1
URL: https://api.vercel.com/v3/now/aliases?limit=9.1
Status Code: 500
Error: An unexpected internal error occurred
```

Mostly I was just trying to catch and help the user, with what felt like a common use case "of the limit isn't enough and I don't want to do pagination, let's try increasing it".

packages/cli/src/commands/alias/ls.ts
andymckay: Agreed would be good to refactor this and reduce the amount of copied code, also would reduce the amount of boilerplate in the tests. I'm 50/50 if I want to do this in this PR or just ship this and add that to the backlog at this point.

packages/cli/src/commands/alias/index.ts
andymckay: Screen Shot 2022-10-27 at 9 57 39 AM

Looks much better, thank you 🙇🏾

packages/cli/src/commands/alias/ls.ts
cb1kenobi: After discussing with Andy, I didn't realize that `opts` is actually `argv`, thus the existing code is good to go.

packages/cli/src/commands/alias/ls.ts
andymckay: Thank you for the feedback @cb1kenobi, I've added all these checks into the function.

packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -0,0 +1,23 @@
+interface Options {
```
cb1kenobi: I think the function name is good, but `Options` type name is a little ambiguous. What kind of options are they?
```suggestion
interface PaginationOptions {
```

packages/cli/src/commands/alias/ls.ts
```diff
@@ -4,28 +4,27 @@ import table from 'text-table';
...
   opts: Options,
   args: string[]
 ) {
   const { output } = client;
-  const { '--next': nextTimestamp } = opts;
   const { contextName } = await getScope(client);
 
-  if (typeof nextTimestamp !== undefined && Number.isNaN(nextTimestamp)) {
-    output.error('Please provide a number for flag --next');
+  let validated;
```
cb1kenobi: `validated` is ambiguous and doesn't describe what is validated.
```suggestion
  let paginationOptions;
```

packages/cli/src/commands/alias/ls.ts
```diff
@@ -46,10 +45,11 @@ export default async function ls(
   const { aliases, pagination } = await getAliases(
     client,
     undefined,
-    nextTimestamp
+    validated.nextTimestamp,
+    validated.limit
```
cb1kenobi: Using the spread operator would make it easier to add/remove future pagination related options.
```suggestion
    ...paginationOptions
```

packages/cli/src/commands/certs/index.ts
```diff
@@ -50,6 +50,9 @@ const help = () => {
       'FILE'
     )}                      CA certificate chain file
     -N, --next                     Show next page of results
+    --limit=${chalk.bold.underline(
+      'VALUE'
+    )}                       Number of results to return per page (default: 20, max: 100)
```
cb1kenobi: I think there are 5 too many spaces between the option and the description.

packages/cli/src/commands/dns/index.ts
```diff
@@ -38,6 +38,9 @@ const help = () => {
   )}        Login token
     -S, --scope                    Set a custom scope
     -N, --next                     Show next page of results
+    --limit=${chalk.bold.underline(
+      'VALUE'
+    )}                       Number of results to return per page (default: 20, max: 100)
```
cb1kenobi: Likewise, I think there are 5 too many spaces here between the option and description.

packages/cli/src/commands/domains/index.ts
```diff
@@ -45,6 +45,9 @@ const help = () => {
   )}        Login token
     -S, --scope                    Set a custom scope
     -N, --next                     Show next page of results
+    --limit=${chalk.bold.underline(
+      'VALUE'
+    )}                       Number of results to return per page (default: 20, max: 100)
```
cb1kenobi: 5 too many spaces between the option and description.

packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  if (
+    typeof limit === 'number' &&
+    (!Number.isInteger(limit) || limit > 100 || limit < 1)
+  ) {
+    throw new Error('Please provide a number up to 100 for flag --limit');
+  }
+
+  return { nextTimestamp: nextTimestamp, limit: limit };
```
cb1kenobi: This can be simplified to:
```suggestion
  return { nextTimestamp, limit };
```

packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export { Options, getPaginationOpts };
```
cb1kenobi: Not sure if we have adopted a convention to whether we should export declarations inline or group exports. Poking around the codebase I'd say tend to export things inline.

cb1kenobi (changes requested):
The CLI integration tests were failing earlier and it wasn't immediately clear to me why. I see they are re-running, so hopefully they'll pass this time.

packages/cli/src/commands/alias/ls.ts
andymckay: It does, but it returning an iterator leaves us vulnerable to someone changing the order of the iterator and breaking things, but that's probably ok.

cb1kenobi (approved):
LGTM

EndangeredMassa (approved):

andymckay:
@anatrajkovska or @mglagola would you be able to take a quick look please? This adds in more options for the CLI, but doesn't affect certs directly, just how we list them.

packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  if (typeof nextTimestamp !== undefined && Number.isNaN(nextTimestamp)) {
+    throw new Error('Please provide a number for flag --next');
+  }
+
+  if (
+    typeof limit === 'number' &&
+    (!Number.isInteger(limit) || limit > 100 || limit < 1)
+  ) {
+    throw new Error('Please provide a number up to 100 for flag --limit');
```
styfle:
```suggestion
    throw new Error('Please provide an integer from 1 to 100 for option --limit');
```

We call this options (not flags) 

See https://vercel.com/docs/cli/alias#unique-options

packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -0,0 +1,21 @@
+export interface PaginationOptions {
+  '--next'?: number;
+  '--limit'?: number;
+}
+
+export function getPaginationOpts(opts: PaginationOptions) {
+  const { '--next': nextTimestamp, '--limit': limit } = opts;
+
+  if (typeof nextTimestamp !== undefined && Number.isNaN(nextTimestamp)) {
+    throw new Error('Please provide a number for flag --next');
```
styfle:
```suggestion
    throw new Error('Please provide a number for option --next');
```

We call this options (not flags) 

See https://vercel.com/docs/cli/alias#unique-options

anatrajkovska (approved):

PR #8736, opened 2022-10-17
[tests] Update workflow to examine both labels and use annotations

andymckay:
### Related Issues

I couldn't find an issue related to this, but just noticed it when I submitted a request. I got a build failing because I failed to add in an `area` label, so added that label and it failed again this time on `semver`. This change makes it so the workflow looks at both labels, rather than just one and exiting early. This might save some people doing 2 builds, like I did 😄 

Also changed it to use Actions annotations, which does colourization for you. Docs on that are here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-debug-message

But the output is now something like this ([test repo](https://github.com/andymckay/test-config)). Missing both labels:

Screen Shot 2022-10-17 at 4 01 35 PM

And in the log:

Screen Shot 2022-10-17 at 4 01 41 PM

And if you've got the labels:

Screen Shot 2022-10-17 at 4 02 59 PM

#### Tests

#### Code Review

### DIFF
--- a/.github/workflows/required-pr-label.yml
+++ b/.github/workflows/required-pr-label.yml
@@ -10,13 +10,17 @@ jobs:
         uses: actions/github-script@v6
         with:
           script: |
+            let missing = false;
             const labels = context.payload.pull_request.labels.map(l => l.name);
             if (labels.filter(l => l.startsWith('area:')).length === 0) {
-              console.error('\u001b[31mMissing label: Please add at least one "area" label.');
-              process.exit(1);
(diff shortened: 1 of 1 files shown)


styfle (approved):
Thanks!

styfle:
@andymckay The Vercel deployment failed because your account is not part of the `vercel` team or the email address you used to `git commit` doesn't match the GitHub account linked to your Vercel account.

andymckay:
Thank for the heads up @styfle I'll update that email address

PR #8747, opened 2022-10-19
[build-utils] allow EdgeFunction constructor to receive any string as a region

Schniz:
### Related Issues

This will allow a shorthand form of `export const config = { regions: 'iad1' }`

#### Tests

#### Code Review

### DIFF
--- a/packages/build-utils/src/edge-function.ts
+++ b/packages/build-utils/src/edge-function.ts
@@ -39,7 +39,12 @@ export class EdgeFunction {
   assets?: { name: string; path: string }[];
 
   /** The regions where the edge function will be executed on */
-  regions?: 'auto' | string[] | 'all' | 'default';
+  regions?:
+    | 'auto'
+    | string[]
+    | 'all'
+    | 'default'
+    | (string & Record<never, unknown>);
 
   constructor(params: Omit<EdgeFunction, 'type'>) {
     this.type = 'EdgeFunction';
--- a/packages/build-utils/src/edge-function.ts
+++ b/packages/build-utils/src/edge-function.ts
@@ -39,12 +39,7 @@ export class EdgeFunction {
   assets?: { name: string; path: string }[];
 
   /** The regions where the edge function will be executed on */
-  regions?:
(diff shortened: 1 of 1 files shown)


packages/build-utils/src/edge-function.ts
```diff
@@ -39,7 +39,12 @@ export class EdgeFunction {
   assets?: { name: string; path: string }[];
 
   /** The regions where the edge function will be executed on */
-  regions?: 'auto' | string[] | 'all' | 'default';
+  regions?:
+    | 'auto'
+    | string[]
+    | 'all'
+    | 'default'
+    | (string & Record<never, unknown>);
```
styfle: How about `regions?: string[] | string`?

Schniz: Then we won't have autocomplete when using it. It's also kinda nice that we show the different options

What do you think?

styfle: Its already not showing autocomplete for your example: `regions: 'iad1'`

Schniz: but it is for `all` and `default` 😁

styfle: Better to add JSDoc since its not even clear what "default" is or how to find which regions "all" goes to

styfle: Also what happens if the caller uses a string array like this?

```ts
{
  regions: ['auto', 'all', 'default', 'iad1']
}
```

Schniz: <picture data-single-emoji=":ack:" title=":ack:">:ack:</picture> <picture data-single-emoji=":sunglasses_+1:" title=":sunglasses_+1:">:sunglasses_+1:</picture>

styfle (approved):

cb1kenobi (approved):
LGTM

feugy (approved):
<picture data-single-emoji=":shipit_parrot:" title=":shipit_parrot:">:shipit_parrot:</picture>

PR #8751, opened 2022-10-19
[ENV] Use new env/pull endpoint for vc env pull, vc pull

baruchadi:
### Related Issues

We introduced a new endpoint: `/env/pull/:projectId/:target?/:gitBranch?` which will return a complete list of key/value pairs (environment variables) for the proper target environment, using the same logic as we use for deployments. 

This means that we don't need to fetch env variables from multiple sources (system, project, shared) and stitch them on the client anymore. 

- removes unused logic for fetching & merging environment variables on the CLI client
- offloads env variables generation to the API
  - new: shared env variables are now supported in vc env pull! 

#### Tests

#### Code Review


baruchadi:
Current dependencies on/for this PR:
* main
  * **PR #8751** <a href="https://app.graphite.dev/github/pr/vercel/vercel/8751" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/vercel/8751?utm_source=stack-comment).

packages/cli/src/util/env/get-env-records.ts
```diff
@@ -49,3 +49,40 @@ export default async function getEnvRecords(
...
+  if (target) {
+    const gitBranchPath = gitBranch ? `/${gitBranch}` : '';
+    url += `/${target}${gitBranchPath}`;
+  }
+
+  if (source) {
+    query.set('source', source);
+  }
+
+  url += `?${query}`;
```
EndangeredMassa: **suggestion (non-blocking):**

```suggestion
  if (Array.from(query).length) {
    url += `?${query}`;
  }
```

Let's not add the question mark if there's nothing that will come after.

packages/cli/test/mocks/project.ts
```diff
@@ -1,42 +1,49 @@
...
-    target: ['development'],
-    gitBranch: null,
+    target: [ProjectEnvTarget.Development],
+    gitBranch: undefined,
     configurationId: null,
     updatedAt: 1557241361445,
     createdAt: 1557241361445,
   },
-];
+] as ProjectEnvVariable[];
```
EndangeredMassa: **suggestion (blocking):** use type instead of assertion

Can we change this to the following?

```
const envs: ProjectEnvVariable[] = [ ... ];
```

That provides type protection whereas the type assertion `as ProjectEnvVariable[]` does not. For example, if you add an unrecognized key to one of the objects in this list, the typed version will throw an error and the type assertion version will not.

packages/cli/test/mocks/project.ts
```diff
@@ -206,6 +213,35 @@ export function useProject(project: Partial<Project> = defaultProject) {
...
+            ? envs.filter(env =>
+                (env.target as ProjectEnvTarget[]).includes(target)
+              )
+            : envs,
+          systemEnvs.map(env => env.key),
+          project.autoExposeSystemEnvs,
+          undefined,
+          target
+        )
+      );
```
EndangeredMassa: **suggestion (soft-blocking):**

```suggestion
      let projectEnvs = envs;
      if (target) {
        projectEnvs = projectEnvs.filter(env => {
          if (typeof env.target === 'string') {
            return env.target === target;
          }
          return env.target.includes(target);
        });
      }

      const allEnvs = Object.entries(
        exposeSystemEnvs(
          projectEnvs,
          systemEnvs.map(env => env.key),
          project.autoExposeSystemEnvs,
          undefined,
          target
        )
      );
```

We know that `envs` only has the array option for `target` on the individual objects. But can we check for the non-array option here in this or a follow-up PR? That keeps this type checked instead of using the type assertion, making this logic more resilient to change.

packages/cli/test/mocks/project.ts
```diff
@@ -223,9 +259,14 @@ export function useProject(project: Partial<Project> = defaultProject) {
...
-    const target = _req.query.target;
+    const target: ProjectEnvTarget | undefined =
+      typeof _req.query.target === 'string'
+        ? parseEnvironment(_req.query.target)
+        : undefined;
     if (typeof target === 'string') {
-      const targetEnvs = envs.filter(env => env.target.includes(target));
+      const targetEnvs = envs.filter(env =>
+        (env.target as ProjectEnvTarget[]).includes(target)
+      );
```
EndangeredMassa: Same as above regarding type assertions.

EndangeredMassa (changes requested):
I gave this an initial review. Can you resolve the [test failure](https://github.com/vercel/vercel/actions/runs/3284179690/jobs/5409821266) and type assertions?

EndangeredMassa (commented):
✅ this looks good to me, but I'd like to get another team member's approval before shipping it

EndangeredMassa (approved):

packages/cli/src/util/env/get-env-records.ts
```diff
@@ -49,3 +49,41 @@ export default async function getEnvRecords(
...
+    `Fetching Environment Variables of project ${projectId} and target ${target}`
+  );
+  const query = new URLSearchParams();
+
+  let url = `/v1/env/pull/${projectId}`;
+
+  if (target) {
+    const gitBranchPath = gitBranch ? `/${gitBranch}` : '';
+    url += `/${target}${gitBranchPath}`;
+  }
```
Ethan-Arrowood: How are you ensuring that `gitBranch` or `target` don't already have a `/` character at the front of them? I'd rather see something like `path.join` be used or utilize the `URL` class

packages/cli/src/util/env/get-env-records.ts
```diff
@@ -49,3 +49,41 @@ export default async function getEnvRecords(
...
+  client: Client,
+  projectId: string,
+  source: EnvRecordsSource,
+  {
+    target,
+    gitBranch,
+  }: {
+    target?: ProjectEnvTarget | string;
+    gitBranch?: string;
+  } = {}
```
Ethan-Arrowood: nit: pull out the type into an interface for readability purposes.

packages/cli/src/util/env/get-env-records.ts
```diff
@@ -49,3 +49,41 @@ export default async function getEnvRecords(
...
+
+  if (target) {
+    const gitBranchPath = gitBranch ? `/${gitBranch}` : '';
+    url += `/${target}${gitBranchPath}`;
+  }
+
+  if (source) {
+    query.set('source', source);
+  }
+  if (Array.from(query).length) {
```
Ethan-Arrowood: nit: don't rely on implicit type conversion. add a conditional
```suggestion
  if (Array.from(query).length > 0) {
```

packages/cli/test/mocks/project.ts
```diff
@@ -206,6 +213,43 @@ export function useProject(project: Partial<Project> = defaultProject) {
     Object.assign(project, req.body);
     res.json(project);
   });
+  client.scenario.get(
+    `/v1/env/pull/${project.id}/:target?/:gitBranch?`,
```
Ethan-Arrowood, line 217: trailing `?` necessary?

packages/cli/test/mocks/project.ts
```diff
@@ -223,14 +267,25 @@ export function useProject(project: Partial<Project> = defaultProject) {
     }
   );
   client.scenario.get(`/v8/projects/${project.id}/env`, (_req, res) => {
-    const target = _req.query.target;
-    if (typeof target === 'string') {
-      const targetEnvs = envs.filter(env => env.target.includes(target));
-      res.json({ envs: targetEnvs });
-      return;
+    const target: ProjectEnvTarget | undefined =
+      typeof _req.query.target === 'string'
```
Ethan-Arrowood: lets rename `_req` to `req`. The preceding `_` is meant to indicate "unused"

Ethan-Arrowood (approved):
some minor nits. approving since tests are passing too

packages/cli/test/mocks/project.ts
baruchadi: on it! Gotta get out of that habit, I'm used to it meaning "private/protected" scope-wise 😅

packages/cli/test/mocks/project.ts
baruchadi: yes, without it the tests don't pass when you don't provide a git branch - it marks both target & gitbranch paths as optional, and we have the same configuration in routes.ts on API

PR #8724, opened 2022-10-13
[@vercel/edge] Support overriding request headers

nuta:
This PR adds a feature in middleware to add, modify, or delete request headers. This feature is quite useful to pass data from middleware to Serverless/Edge API routes.

### New APIs

Adds a new option `request.headers` to the `MiddlewareResponseInit` parameter in `NextResponse.next()` and `NextResponse.rewrite()`. It's a [`Header`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object holding *all* request headers. Specifically:

```ts
interface ExtraResponseInit extends ResponseInit {
  request?: {
    headers?: Headers
  }
}
```

### Example

```ts
// api/hello.ts
export default (req, res) => {
  const valueFromMiddleware = req.headers['x-hello-from-middleware']
  return res.send(valueFromMiddleware)
}

// middleware.ts
import { next } from '@vercel/edge'

export default function middleware(request: NextRequest) {
  // Clone request headers
  const headers = new Headers(request.headers);
  // Add a new request header
  headers.set('x-hello-from-middleware', 'foo');
  // Delete a request header from the client
  headers.delete('x-from-client');

  return next({
    request: {
      headers
    }
  });
}
```

### New middleware headers

- `x-middleware-override-headers`: A comma separated list of *all* request header names. Headers not listed will be deleted.
- `x-middleware-request-<name>`: A new value for the header `<name>`.

### Related Issues

- Next.js' implementation: https://github.com/vercel/next.js/pull/41380

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8752, opened 2022-10-20
[dev] Support request headers override in middleware

nuta:
Implements request headers override in middlewares.

#### New middleware headers

- `x-middleware-override-headers`: A comma separated list of *all* request header names. Headers not listed will be deleted.
- `x-middleware-request-<name>`: A new value for the header `<name>`.

### Related Issues

- #8724: Add helper functions for non-Next.js middlewares
- https://github.com/vercel/next.js/pull/41380: Next.js' implementation

#### Tests

#### Code Review


packages/cli/src/util/dev/server.ts
```diff
@@ -1478,6 +1478,29 @@ export default class DevServer {
...
+              if (newValue) {
+                req.headers[name] = newValue;
+              } else {
+                delete req.headers[name];
+              }
+
+              middlewareRes.headers.delete(valueKey);
+            }
+            middlewareRes.headers.delete('x-middleware-override-headers');
+          }
```
Schniz: do we want to extract this into a simple to understand function, that can easily be unit testable? `applyOverriddenHeaders(middlewareRes.headers)` <picture data-single-emoji=":eyeses:" title=":eyeses:">:eyeses:</picture>

EndangeredMassa: I agree. Some unit tests for the new logic would be great in this or a follow-up PR.

EndangeredMassa (approved):

javivelasco (approved):

packages/cli/src/util/dev/headers.ts
```diff
@@ -16,3 +16,60 @@ export function nodeHeadersToFetchHeaders(
...
+  respHeaders: Headers,
+  skippedHeaders: Set<string>
+) {
+  if (!respHeaders.has('x-middleware-override-headers')) {
+    return;
+  }
+
+  const overriddenHeaders: Set<string> = new Set();
+  for (const key of respHeaders
+    .get('x-middleware-override-headers')!
```
Ethan-Arrowood: Is there an edge case where `respHeaders.get('x-middleware-override-headers')` could return `null` or `undefined` or some that isn't a string?

Ethan-Arrowood (approved):
lgtm - see potential edge case comment. might just need a simple guard.

packages/cli/src/util/dev/headers.ts
nuta: No there was a if guard to ensure it exists. Refactored that part so that we don't need to use `!` 👍

PR #8795, opened 2022-10-27
[node-bridge] Support streaming response for Serverless Function

TooTallNate:
Adds streaming response support for React Server Components with Next 13.


ijjk (approved):

packages/node-bridge/bridge.js
```diff
@@ -1,4 +1,10 @@
+const { URL } = require('url');
```
styfle, line 1: I think you can use the global `URL` instead

packages/node-bridge/package.json
```diff
@@ -23,6 +23,8 @@
   "devDependencies": {
     "@types/aws-lambda": "8.10.19",
     "@types/node": "*",
+    "jsonlines": "0.1.1",
+    "test-listen": "1.1.0",
```
styfle, line 27: Lets publish a new version of this without all the bloat

EndangeredMassa:
I'm missing a bunch of context to be able to review this in detail. If others have enough, no worries, but if there's somewhere I can read more about the technical details, I'd be happy to dig through it.

packages/node-bridge/bridge.js
TooTallNate: TypeScript doesn't seem to like it, so seems like something isn't configured correctly.

cb1kenobi: I also ran into not being able to use the global `URL`. Maybe our `@types/node` is out of date?

styfle (approved):

ijjk (approved):

PR #8811, opened 2022-10-31
[node] upgrade edge-runtime to v2.0.0

Kikobeats:
It drops node12 support


EndangeredMassa (approved):
I believe this is good to merge. The CLI already only supports Node.js 14+ and builds for deployment don't use the `edge-runtime`.

I took off automerge to give others a chance to weigh in, just to be sure.

cb1kenobi (approved):

styfle (approved):

styfle:
I changed to semver patch since we already dropped support for Node.js 12 in a previous PR

PR #8814, opened 2022-10-31
[routing-utils] Allow to appendRoutes add null phase

tobiaslins:
### Related Issues

For Analytics V2 we need to append routes to the `null` phase instead of `filesystem`.

#### Tests

#### Code Review

### DIFF
--- a/packages/routing-utils/src/types.ts
+++ b/packages/routing-utils/src/types.ts
@@ -122,6 +122,7 @@ export interface AppendRoutesToPhaseProps {
   newRoutes: Route[] | null;
   /**
    * The phase to append the routes such as `filesystem`.
+   * If the phase is `null`, the routes will be appended prior to the first handle being found.
    */
   phase: HandleValue | null;
 }
--- a/packages/routing-utils/src/types.ts
+++ b/packages/routing-utils/src/types.ts
@@ -123,5 +123,5 @@ export interface AppendRoutesToPhaseProps {
   /**
    * The phase to append the routes such as `filesystem`.
    */
-  phase: HandleValue;
+  phase: HandleValue | null;
 }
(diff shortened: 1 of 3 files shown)


packages/routing-utils/src/types.ts
```diff
@@ -123,5 +123,5 @@ export interface AppendRoutesToPhaseProps {
   /**
    * The phase to append the routes such as `filesystem`.
```
dglsparsons, line 124: Might be worth expanding this comment with an explanation of the behaviour of passing `null`

packages/routing-utils/test/append.spec.ts
```diff
@@ -128,3 +128,41 @@ test('appendRoutesToPhase one routes before, two routes in phase, two routes in
...
+  const phase = null;
+  const actual = appendRoutesToPhase({ routes, newRoutes, phase });
+  const expected = [
+    { src: '/first', dest: '/one' },
+    { src: '/second', dest: '/two' },
+    { src: '/new', dest: '/to' },
+  ];
+
+  deepEqual(actual, expected);
+});
```
styfle, line 168: Lets add one more test when `const routes = []`

packages/routing-utils/test/append.spec.ts
```diff
@@ -128,3 +128,41 @@ test('appendRoutesToPhase one routes before, two routes in phase, two routes in
...
+
+test('appendRoutesToPhase to null phase with no handle', () => {
+  const routes: Route[] = [
+    { src: '/first', dest: '/one' },
+    { src: '/second', dest: '/two' },
+  ];
+  const newRoutes = [{ src: '/new', dest: '/to' }];
```
styfle, line 158: Also lets and another test when `newRoutes` has two routes

packages/routing-utils/src/types.ts
styfle, line 124: Lets updating the JSDoc comment here to mention that `null` represents the phase prior to any `handle` being found.

dglsparsons: :spiderman-pointing:

styfle (approved):
Thanks!

PR #8742, opened 2022-10-18
[cli][fs-detectors] Improve `vc build` monorepo support

Ethan-Arrowood:
### Related Issues

Improves how `vc build` handles monorepos. In short, this introduces monorepo manager detection logic and then some helpful defaults so users don't have to manually specify a `buildCommand` or `installCommand` when linking a project within a monorepo.

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8819, opened 2022-10-31
[errors] Add @vercel/errors private package

Ethan-Arrowood:
### Related Issues

While working on #8742 i found some useful error checking code buried in the CLI. This PR adds a new private package `@vercel/errors` that moves those helpful utilities into its own package so it can be used throughout the monorepo.

#### Tests

#### Code Review


packages/errors/src/index.ts
```diff
@@ -25,7 +25,6 @@ export const isError = (error: unknown): error is Error => {
   // Walk the prototype tree until we find a matching object.
   while (error) {
     if (Object.prototype.toString.call(error) === '[object Error]') return true;
-    // eslint-disable-next-line no-param-reassign -- TODO: Fix eslint error following @vercel/style-guide migration
     error = Object.getPrototypeOf(error);
   }
```
Ethan-Arrowood: This block is the only path I can't figure out how to test. 

I tried using:

```js
const err = Object.assign(new Error(), {});
t.ok(isError(err));
```

But the `err` returns true for `error instanceof Error` and never enters the while loop.

I need to figure out how to define an object that has a nested Error prototype.

packages/errors/tsconfig.json
```diff
@@ -0,0 +1,106 @@
+{
```
styfle: Lets use the same tsconfig.json as our other packages

packages/errors/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "build": "tsc",
+    "test": "tap --ts"
+  },
+  "author": "Ethan Arrowood <ethan.arrowood@vercel.com>",
+  "license": "MIT",
+  "devDependencies": {
+    "@types/node": "16.11.7",
+    "@types/tap": "^15.0.7",
+    "tap": "^16.3.0",
```
styfle: Lets use jest like our other tests in this monorepo

packages/errors/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@vercel/errors",
```
styfle: Lets call this `@vercel/error-utils`

I'm also wondering if you plan to  import it from builders. In which case, it should probably go in `@vercel/build-utils` instead.

Ethan-Arrowood: Hmm I could move it to `@vercel/build-utils`. It is going to be used in `@vercel/fs-detectors` next (https://github.com/vercel/vercel/pull/8742#discussion_r1006370174)

packages/errors/src/index.ts
Ethan-Arrowood: Ahh I think I got it:

```js
class Foo {}
const err = new Error();
Object.setPrototypeOf(err, Foo.prototype);
```

Ethan-Arrowood:
whoops - recreated: https://github.com/vercel/vercel/pull/8828

Issue #8644, opened 2022-09-27
Removing emojis from CLI output


bound-variable:
Yellow on white is hard on the eyes for many. Green on white might be even worse for colorblind users (like myself). Can we have the option to remove the emojis from the output, please?

### Screenshot one
![image](https://user-images.githubusercontent.com/51193876/192440551-f48755c4-bbae-40a5-ba74-dcb991254a5b.png)

### Screenshot two
![13310](https://user-images.githubusercontent.com/51193876/192441244-6b19a4e5-ee69-4986-aafa-f840098b1a38.png)

For more info, [see here](https://www.uxmatters.com/mt/archives/2007/01/applying-color-theory-to-digital-displays.php).

Thanks

techoi:
https://github.com/vercel/vercel/discussions/8775
If maintainers answer it, I want to make option for you.

PR #8826, opened 2022-11-01
[cli] Add `--no-color` mode

techoi:
### Related Issues
https://github.com/vercel/vercel/issues/8644

discussion https://github.com/vercel/vercel/discussions/8775

### Feature
- [x] Add no color feature which is disable chalk color and remove emoji in outputs.
- [x] Support env var FORCE_COLOR=0
- [x] Support env var NO_COLOR=1
- [x] Support flag --no-color

### Outputs
with emoji
withEmoji

without emoji
with --no-color

with NO_COLOR

#### Tests

#### Code Review


techoi:
I do not have permission to vercel/vercel so i forked and make pull request.

packages/cli/src/util/output/create-output.ts
```diff
@@ -35,6 +39,16 @@ export class Output {
   };
 
   print = (str: string) => {
+    if (this.textOnly) {
+      const emojis = Object.values(emojiLabels).map(emoji => `${emoji}  `);
+      const regExp = new RegExp(emojis.join('|'), 'gi');
+      const textOnlyStr = str.replace(regExp, '');
```
Ethan-Arrowood: I like how you are accounting for the two spaces we add in `prependEmoji`, but I'd rather match strictly on the emoji and then trim leading whitespace from the string.

Ethan-Arrowood (commented):
looking good

packages/cli/src/util/output/create-output.ts
techoi: @Ethan-Arrowood Thank you for comment. I think that's much better and safety, I will apply it.

techoi: [be185fb](https://github.com/vercel/vercel/pull/8826/commits/be185fb0a71a1e8a84a87775beee1b323bc8eb63) applied it.

TooTallNate:
I wonder if it would make more sense to have this be configurable via an env var or something more permanent. If the `--text-only` flag is needed to be specified for every command I think that could become annoying.

packages/cli/src/util/output/create-output.ts
```diff
@@ -17,15 +19,17 @@ export interface LogOptions {
...
   stream: WritableTTY;
   debugEnabled: boolean;
+  textOnly: boolean;
   private spinnerMessage: string;
   private _spinner: StopSpinner | null;
 
   constructor(
     stream: WritableTTY,
-    { debug: debugEnabled = false }: OutputOptions = {}
+    { debug: debugEnabled = false, textOnly = false }: OutputOptions = {}
```
techoi: @TooTallNate
> I wonder if it would make more sense to have this be configurable via an env var or something more permanent. If the `--text-only` flag is needed to be specified for every command I think that could become annoying.

That makes sense.
What about use env var name as `TEXT_ONLY_MODE`?
If someone set `TEXT_ONLY_MODE=on` then default value of textOnly will be `process.env.TEXT_ONLY_MODE === 'on'`

techoi: Like this

```suggestion
    { debug: debugEnabled = false, textOnly = process.env.TEXT_ONLY_MODE === 'on' }: OutputOptions = {}
```

Maybe `const IS_TEXT_ONLY_MODE = process.env.TEXT_ONLY_MODE === 'on';` then `{ debug: debugEnabled = false, textOnly = IS_TEXT_ONLY_MODE }: OutputOptions = {}` is better but...

Ethan-Arrowood: It could also be a global option for folks that prefer that configuration pattern

techoi: [ef35bed](https://github.com/vercel/vercel/pull/8826/commits/ef35bed3ab3c096aa7d9eec336c2601b1131cf4a)

applied it

Is there any good way to guide this global option to folks that prefer this? l(ike doc or --help)

EndangeredMassa:
@techoi thanks for updating this!

We're having an internal discussion about how exactly we'd like this to be implemented. We'll come back with some more feedback soon.

EndangeredMassa:
Because this comes from an issue with color, we think it would make the most sense for this PR to combine the removal of emoji with the removal of color.

The CLI already respects an environment variable of `FORCE_COLOR=0` to remove color from the text, but that's because we use Chalk to render that color.

For Vercel CLI, we'd prefer to adopt the [`NO_COLOR` standard](https://no-color.org/). So, requirements for this feature are:

- support environment variables: `FORCE_COLOR=0` or `NO_COLOR=1`
- support global flag `--no-color`
- if in no color mode (for any of those three reasons)
    - disable Chalk color (probably by setting the `FORCE_COLOR=0` env var)
    - remove all emoji (in the [emoji()](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/emoji.ts) function)

TooTallNate:
Note that Chalk can be configured by setting `chalk.level`:

```
> chalk.red('hi')
'\x1B[31mhi\x1B[39m'

> chalk.level = 0
0

> chalk.red('hi')
'hi'
```

cb1kenobi:
It's perhaps worth noting that setting `chalk.level = 0` will globally set the level which affects all dependencies too. Generally, I'd imagine this is desired behavior, though it has caused issues for me in specific use cases.

packages/cli/src/util/input/list.ts
```diff
@@ -54,7 +54,7 @@ export default async function list(
     eraseFinalAnswer = false, // If true, the line with the final answer that inquirer prints will be erased before returning
   }: ListOptions
 ): Promise<string> {
-  require('./patch-inquirer-legacy');
+  require('./patch-inquirer');
```
techoi: I changed `patch-inquirer-legacy` to `patch-inquirer` to override chalk that inquirer list directly use.

If there is any concern please leave a comment.

packages/cli/test/unit/commands/login.test.ts
```diff
@@ -47,19 +47,44 @@ describe('login', () => {
...
+      await expect(client.stderr).toOutput(
+        `Success! Email authentication complete for ${user.email}`
+      );
+
+      await expect(client.stderr).not.toOutput(emoji('tip'));
+
+      await expect(exitCodePromise).resolves.toEqual(0);
+    });
+    it('should remove image the `NO_COLOR` env var with 1', async () => {
+      process.env.NO_COLOR = '1';
```
EndangeredMassa: **suggestion (blocking):** set and unset `NO_COLOR` in `before`/`after` hooks

To avoid test pollution, can we set `process.env.NO_COLOR = '1';` in a `before` hook and unset it in an `after` hook?

packages/cli/src/util/output/create-output.ts
```diff
@@ -43,7 +39,7 @@ export class Output {
   };
 
   print = (str: string) => {
-    if (this.textOnly) {
```
EndangeredMassa: **suggestion (blocking):** implement emoji removal in `emoji` function

Can we implement the emoji removal in [emoji.ts](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/emoji.ts) directly instead of here?

Not all output goes through this Output class. We could miss cases that need removal by implementing it here.

EndangeredMassa (changes requested):
Great changes! I just have a couple suggestions.

packages/cli/src/util/output/create-output.ts
techoi: That's good idea.
~~I used NO_COLOR env var to remove emoji in `emoji.ts`~~
I create removeEmoji function in emoji.ts 
1810189

techoi: [d8651cc](https://github.com/vercel/vercel/pull/8826/commits/d8651ccad7c1dcc1ffabb9eaab952e93e830efa9)

packages/cli/src/util/emoji.ts
```diff
@@ -15,7 +15,7 @@ export function emoji(label: EmojiLabel) {
 }
 
 export function prependEmoji(message: string, emoji?: string): string {
-  if (emoji && process.stdout.isTTY) {
+  if (emoji && process.stdout.isTTY && process.env.NO_COLOR !== '1') {
```
TooTallNate: We shouldn't rely on the env var here, we should add a 3rd parameter instead.

packages/cli/src/index.ts
```diff
@@ -109,8 +110,22 @@ const main = async () => {
...
+  ) {
+    process.env.NO_COLOR = '1';
+    chalk.level = 0;
+  }
+
   const isDebugging = argv['--debug'];
-  const output = new Output(process.stderr, { debug: isDebugging });
+
+  const output = new Output(process.stderr, {
+    debug: isDebugging,
```
TooTallNate, line 109: I think we should store the `noColor` param here on the Output object, so that we can properly test the scenarios in the unit tests without messing around with the env vars.

packages/cli/src/util/input/list.ts
TooTallNate: Ideally this would be a separate PR to keep the diff noise down in this one.

packages/cli/test/mocks/client.ts
```diff
@@ -132,6 +132,10 @@ export class MockClient extends Client {
 
   setArgv(...argv: string[]) {
     this.argv = [process.execPath, 'cli.js', ...argv];
+
```
TooTallNate, line 140: This won't be necessary if we have `output.noColor` instead.

TooTallNate (commented):
This is getting there, nice work! Left a few comments.

techoi:
> Ideally this would be a separate PR to keep the diff noise down in this one.
@TooTallNate I separate PR https://github.com/vercel/vercel/pull/8919

packages/cli/src/util/input/list.ts
techoi: I separated PR https://github.com/vercel/vercel/pull/8919

techoi:
@EndangeredMassa Hi, Is there any update about accepting this pr? If there is any suggestion please let me know. Thank you.

techoi:
yarn -> pnpm migrate and fix the conflict [c428d6b](https://github.com/vercel/vercel/pull/8826/commits/c428d6b8b3d1a25d1b69896422e5da2d428a42ba)

techoi:
> Because this comes from an issue with color, we think it would make the most sense for this PR to combine the removal of emoji with the removal of color.
> 
> The CLI already respects an environment variable of `FORCE_COLOR=0` to remove color from the text, but that's because we use Chalk to render that color.
> 
> For Vercel CLI, we'd prefer to adopt the [`NO_COLOR` standard](https://no-color.org/). So, requirements for this feature are:
> 
> * support environment variables: `FORCE_COLOR=0` or `NO_COLOR=1`
> * support global flag `--no-color`
> * if in no color mode (for any of those three reasons)
>   
>   * disable Chalk color (probably by setting the `FORCE_COLOR=0` env var)
>   * remove all emoji (in the [emoji()](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/emoji.ts?rgh-link-date=2022-11-14T21%3A56%3A15Z) function)

@EndangeredMassa

Please check the PR, I apply your suggestions. Thanks

EndangeredMassa:
@techoi thanks! Will review soon. Sorry for the delay.

EndangeredMassa:
@techoi I made some small changes:

- [small refactor in `create-output` with some links to more details on the standards](https://github.com/vercel/vercel/pull/8826/files#diff-edcc89c00cb919143e9d7bada97aa314c17f57d96bc92f56235d60bce99b81f7R48)
- [added an assertion that we see the emoji in tests where we don't set any of the flags](https://github.com/vercel/vercel/pull/8826/files?diff=split&w=1#diff-58faa58fd55b336efd8bb7ca08442525b30711178fe79fe139c31a959498dfbeR47)
- [extract env var manipulation in tests to not cause test pollution](https://github.com/vercel/vercel/pull/8826/files?diff=split&w=1#diff-58faa58fd55b336efd8bb7ca08442525b30711178fe79fe139c31a959498dfbeR77-L48)

[Commit `36924c3` (#8826)](https://github.com/vercel/vercel/pull/8826/commits/36924c34a2c753b068f54531bb06c3ca62bb72cf)

What do you think?

packages/cli/src/commands/env/rm.ts
```diff
@@ -19,6 +19,7 @@ import { isAPIError } from '../../util/errors-ts';
 
 type Options = {
   '--debug': boolean;
+  '--no-color': boolean;
```
cb1kenobi: This doesn't appear to be necessary.

packages/cli/src/commands/build.ts
```diff
@@ -118,6 +118,7 @@ const help = () => {
       --output [path]                Directory where built assets should be written to
       --prod                         Build a production deployment
       -d, --debug                    Debug mode [off]
+      --no-color                     No color mode [off]
```
cb1kenobi, line 121: Since `--no-color` is global and we're adding it to the help for `build`, `deploy`, `login`, and `pull` commands, then we should probably add it to all commands.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -91,6 +91,7 @@ export default async (client: Client): Promise<number> => {
       '--prod': Boolean,
       '--archive': String,
       '--yes': Boolean,
+      '--no-color': Boolean,
```
cb1kenobi: This doesn't appear to be necessary.

packages/cli/src/index.ts
```diff
@@ -94,6 +94,7 @@ const main = async () => {
         '-v': '--version',
         '--debug': Boolean,
         '-d': '--debug',
+        '--no-color': Boolean,
```
cb1kenobi: I can't figure out if this is actually necessary. It seems to make sense to have it, but removing it appears to have no side effects.

packages/cli/src/util/emoji.ts
```diff
@@ -21,3 +21,10 @@ export function prependEmoji(message: string, emoji?: string): string {
 
   return message;
 }
+
+export function removeEmoji(message: string): string {
+  const regExp = new RegExp(Object.values(emojiLabels).join('|'), 'gi');
```
cb1kenobi: Since `emojiLabels` is const, the `regExp` can be initialized and cached outside `removeEmoji()` so that the regex doesn't need to be reconstructed and compiled every time `log()` is called. Should probably also give it a better name like `stripEmojiRegex` or something.

packages/cli/src/util/output/create-output.ts
```diff
@@ -33,20 +36,32 @@ export class Output {
...
+    }
   }
 
   isDebugEnabled = () => {
     return this.debugEnabled;
   };
 
   print = (str: string) => {
+    if (this.colorDisabled) {
+      const strWithoutEmoji = removeEmoji(str);
```
cb1kenobi: @EndangeredMassa would it be acceptable to set `str = removeEmoji(str);` and then not need to have two `stopSpinner()` calls and two `this.stream.write()` calls?

cb1kenobi (changes requested):

packages/cli/src/util/output/create-output.ts
EndangeredMassa: Yeah, that would be cool with me

techoi: I reflected reviews.

EndangeredMassa (dismissed):

cb1kenobi (dismissed):
Looks great! Thanks!

EndangeredMassa:
The assertion that the tip emoji is rendered normally only passes locally because there's an `isTTY` check. In CI, that emoji is not rendered. So, I removed the assertion: [`a1eee08` (#8826)](https://github.com/vercel/vercel/pull/8826/commits/a1eee08492a3b5dd80f8f2d47da37bb045a906bf)

EndangeredMassa (approved):

EndangeredMassa:
I verified that CI is passing with these changes.

EndangeredMassa:
I verified that CI is passing with these changes.

PR #8828, opened 2022-11-01
[error-utils] Create `@vercel/error-utils` package

Ethan-Arrowood:
### Related Issues

While working on #8742 i found some useful error checking code buried in the CLI. This PR adds a new private package `@vercel/errors` that moves those helpful utilities into its own package so it can be used throughout the monorepo.

#### Tests

#### Code Review


packages/error-utils/babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', { targets: { node: 'current' } }],
+    '@babel/preset-typescript',
+  ],
+};
```
styfle: Lets remove babel. ~~We use swc~~

In fact, we don't need it at all because TS has `target` so transpilation is already handled.

Ethan-Arrowood: Jest errored without this and told me to add it. Ill look how other packages are doing it and fix

styfle: Probably needs a jest config file like this
https://github.com/vercel/vercel/blob/51fe09d5e96df72877e5daf21929aa7ec43cd382/packages/fs-detectors/jest.config.js

packages/error-utils/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@babel/preset-env": "7.19.4",
+    "@babel/preset-typescript": "7.18.6",
+    "@types/jest": "29.2.1",
+    "@types/node": "16.11.7",
+    "babel-jest": "29.2.2",
+    "jest": "29.2.2",
+    "typescript": "^4.8.4"
+  },
+  "dependencies": {
+    "jest": "29.2.2"
```
styfle: This should be `devDependencies`

cb1kenobi (approved):
LGTM

packages/error-utils/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "type": "git",
+    "url": "https://github.com/vercel/vercel.git",
+    "directory": "packages/error-utils"
+  },
+  "scripts": {
+    "build": "tsc",
+    "test": "jest --coverage --env node --verbose",
+    "test-unit": "yarn test"
+  },
+  "author": "Ethan Arrowood <ethan.arrowood@vercel.com>",
```
styfle: Lets remove `author`

packages/cli/package.json
```diff
@@ -42,6 +42,7 @@
   },
   "dependencies": {
     "@vercel/build-utils": "5.5.5",
+    "@vercel/error-utils": "1.0.0",
```
styfle: This wont work if you plan to publish private. It would need to be `devDependencies`

Ethan-Arrowood: Unless we publish the cli with deps bundled or whatever then I'm making error-utils public so it can be required/imported correctly.

styfle (changes requested):
Lets move it to `devDependencies` because cli already bundles its dependencies.

packages/cli/package.json
styfle: Lets move it to `devDependencies` regardless because cli already bundles its dependencies.

styfle (approved):
Thanks!

Issue #853, opened 2017-09-15
now switch no longer accepts token


sholladay:
Since Now 8,  it seems I cannot `now switch` to a team from a script, because `--token` is no longer respected.

I love the fact that `--team` was introduced 💕 but it is annoying and less maintainable to have to specify `--team` with every single command in the script (e.g. if our team name changes, more chance for mistakes). As far as I can tell, this shouldn't be necessary. I consider this a regression, personally, unless I am missing something. Let's keep it DRY.

## Before

Now 7.4.0:
```console
$ now switch my-team --token=$ZEIT_TOKEN
> Success! The team My Team is now active!
```

## After

Now 8.3.0:
```console
$ now switch my-team --token=$ZEIT_TOKEN;
> Error! This command doesn't work with "--token". Please use "--team".
```

leo:
This is expected. We introduced `--team` because it's much more explicit. In terms of CI, you shouldn't use `now switch` - it's only meant to be used manually.

Also great to hear that you like the new flag! 😍

PR #8835, opened 2022-11-02
[next] Prepend basePath to edge function destination

cb1kenobi:
For Next.js apps, a custom `basePath` in the `next.config.js` file was being prepended to all paths except edge functions of which this PR resolves.

### Related Issues

> Fixes [#853](https://github.com/vercel/customer-issues/issues/853)

#### Tests

#### Code Review


packages/next/test/unit/fixtures/02-edge-function-basepath/next.config.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  reactStrictMode: true,
+  swcMinify: true,
```
styfle: You can remove swcMinify since its default enabled

packages/next/test/unit/fixtures/02-edge-function-basepath/pages/_app.tsx
```diff
@@ -0,0 +1,7 @@
+import type { AppProps } from 'next/app'
+
+function MyApp({ Component, pageProps }: AppProps) {
+  return <Component {...pageProps} />
+}
+
+export default MyApp
```
styfle: You can remove _app since its basically a no-op here

packages/next/test/unit/fixtures/02-edge-function-basepath/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "incremental": true
+  },
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "exclude": ["node_modules"]
+}
```
styfle, line 20: Maybe remove TS too?

packages/next/test/unit/fixtures/02-edge-function-basepath/yarn.lock
```diff
@@ -0,0 +1,191 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@next/env@13.0.1":
```
styfle, line 5: We might want to remove this lockfile too so we always test latest Next.js

packages/next/src/utils.ts
```diff
@@ -2445,6 +2445,12 @@ export async function getMiddlewareBundle({
           shortPath.replace(/^app\//, '').replace(/(^|\/)page$/, '') || 'index';
       }
 
+      if (routesManifest?.basePath) {
+        shortPath = path.posix
+          .join(routesManifest.basePath, shortPath)
+          .replace(/^\//, '');
```
styfle, line 2451: I'm curious about this `.replace(/^\//, '')` because I'm not sure how its related to basePath.

Is this a different bug you found?

packages/next/test/unit/fixtures/02-edge-function-basepath/pages/_app.tsx
cb1kenobi: Done!

packages/next/test/unit/fixtures/02-edge-function-basepath/next.config.js
cb1kenobi: Done!

packages/next/test/unit/fixtures/02-edge-function-basepath/tsconfig.json
cb1kenobi: While I'm sure TypeScript has nothing to do with the test, it's nice to match the supplied reproduction test case for which this issue was discovered.

packages/next/test/unit/fixtures/02-edge-function-basepath/yarn.lock
cb1kenobi: This contradicts the request to add a lockfile to the normalize paths PR: https://github.com/vercel/vercel/pull/8600. I think we should either keep the `yarn.lock` for the same reason it was added to the normalize paths PR or remove the lock file for both tests.

styfle: Hmm good point, the lock file is going to be faster

packages/next/src/utils.ts
cb1kenobi: It appears the `routesManifest.basePath` contains the unaltered value from the `next.config.js` and thus a leading slash is valid. The build system is quite happy with or without the leading slash, however every key in the `edgeFunctions` object is a path that does _not_ start with a leading slash, so for consistency's sake I remove the leading slash and things still work as they were.

I could have simply stripped the leading slash from `routesManifest.basePath` before passing it into `join()`, but I felt this looks easier to read.

styfle (approved):
Thanks!

PR #8842, opened 2022-11-03
[cli][client][error-utils][frameworks][fs-detectors][hydrogen][next][node-bridge][node][redwood][remix][routing-utils][static-config] update @types/node to v14 across repo

Ethan-Arrowood:
### Related Issues

Updates @types/node to the latest version within the v14 major (based on `npm view @types/node`)

```
❯ npm view @types/node@'>=14.0.0 <15.0.0' version | tail -1
@types/node@14.18.33 '14.18.33'
```

This PR also fixes the various necessary type changes

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -49,6 +49,12 @@ export const isJSONObject = (v: any): v is JSONObject => {
   return v && typeof v == 'object' && v.constructor === Object;
 };
 
+// https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+const ABSOLUTE_URL_SCHEME = /^[a-zA-z][a-zA-Z0-9+\-.]*?:/;
+
+// https://www.rfc-editor.org/rfc/rfc3986#section-4.3
+const isAbsoluteURI = (url: string) => ABSOLUTE_URL_SCHEME.test(url);
+
 export default class Client extends EventEmitter implements Stdio {
   argv: string[];
   apiUrl: string;
@@ -85,8 +91,10 @@ export default class Client extends EventEmitter implements Stdio {
     });
   }
 
-  private _fetch(_url: string, opts: FetchOptions = {}) {
-    const parsedUrl = new URL(_url);
+  private _fetch(urlOrPath: string, opts: FetchOptions = {}) {
+    const parsedUrl = isAbsoluteURI(urlOrPath)
+      ? new URL(urlOrPath)
+      : new URL(urlOrPath, this.apiUrl);
     const apiUrl = parsedUrl.host
       ? `${parsedUrl.protocol}//${parsedUrl.host}`
       : '';
@@ -105,7 +113,7 @@ export default class Client extends EventEmitter implements Stdio {
         query.set('teamId', this.config.currentTeam);
       }
 
-      _url = `${apiUrl}${parsedUrl.pathname}?${query}`;
+      urlOrPath = `${apiUrl}${parsedUrl.pathname}?${query}`;
     }
 
     const headers = new Headers(opts.headers);
@@ -122,7 +130,7 @@ export default class Client extends EventEmitter implements Stdio {
       body = opts.body;
     }
 
-    const url = `${apiUrl ? '' : this.apiUrl}${_url}`;
+    const url = `${apiUrl ? '' : this.apiUrl}${urlOrPath}`;
     const requestId = this.requestIdCounter++;
     return this.output.time(res => {
       if (res) {
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -93,7 +93,8 @@ export default class Client extends EventEmitter implements Stdio {
       : '';
 
     if (opts.accountId || opts.useCurrentTeam !== false) {
-      const query = new URLSearchParams(parsedUrl.query);
+      // TODO - not sure what the correct TypeScript fix is here
+      const query = new URLSearchParams(parsedUrl.query as any);
 
       if (opts.accountId) {
         if (opts.accountId.startsWith('team_')) {
--- a/packages/cli/src/util/client.ts
+++ b/packages/cli/src/util/client.ts
@@ -1,7 +1,7 @@
 import { bold } from 'chalk';
 import inquirer from 'inquirer';
 import { EventEmitter } from 'events';
-import { URL, URLSearchParams } from 'url';
+import { URL } from 'url';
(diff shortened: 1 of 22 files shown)


packages/cli/src/util/dev/parse-query-string.ts
```diff
@@ -38,9 +38,9 @@ export function parseQueryString(
  */
 export function formatQueryString(
   query: Record<string, string[]> | undefined
-): string | undefined {
+): string | null {
```
TooTallNate, line 41: How come this change was necessary?

Ethan-Arrowood: The URL property the result of this function was being assigned too requires string | null instead of undefined

packages/cli/src/util/client.ts
```diff
@@ -93,7 +93,8 @@ export default class Client extends EventEmitter implements Stdio {
       : '';
 
     if (opts.accountId || opts.useCurrentTeam !== false) {
-      const query = new URLSearchParams(parsedUrl.query);
+      // TODO - not sure what the correct TypeScript fix is here
+      const query = new URLSearchParams(parsedUrl.query as any);
```
cb1kenobi: What happens if you do `as string`? I see `URLSearchParams` constructor can accept a string argument.

packages/cli/src/util/dev/builder.ts
```diff
@@ -87,8 +87,12 @@ async function createBuildProcess(
 
   return new Promise((resolve, reject) => {
     // The first message that the builder process sends is the `ready` event
-    buildProcess.once('message', ({ type }) => {
-      if (type !== 'ready') {
+    buildProcess.once('message', data => {
+      if (
+        data !== null &&
+        typeof data === 'object' &&
+        (data as { type: string }).type !== 'ready'
```
cb1kenobi, line 94: Is it possible to put declare the type on `data` in the function signature?

packages/cli/src/util/dev/router.ts
```diff
@@ -57,8 +57,9 @@ export async function devRouter(
   phase?: HandleValue | null
 ): Promise<RouteResult> {
   let result: RouteResult | undefined;
-  let { pathname: reqPathname = '/', search: reqSearch } = url.parse(reqUrl);
-  const reqQuery = parseQueryString(reqSearch);
+  let { pathname: reqPathname, search: reqSearch } = url.parse(reqUrl);
+  reqPathname ??= '/';
```
cb1kenobi, line 61: ❤️

packages/cli/src/util/dev/router.ts
```diff
@@ -57,8 +57,9 @@ export async function devRouter(
...
+  const reqQuery = parseQueryString(reqSearch ?? undefined);
```
cb1kenobi: `parseQueryString()` can accept a `string`, `undefined`, or `null`, so you don't need to do a nullish check.

packages/cli/src/util/client.ts
styfle: This might be a real bug if the constructor doesn’t handle the case with multiple keys. 

Might be better to change this to “new URL(_url)”

packages/cli/src/util/dev/parse-query-string.ts
styfle: This seems like a bug. We treat null and undefined differently in this function

cb1kenobi (changes requested):

packages/cli/src/util/dev/parse-query-string.ts
Ethan-Arrowood: I believe its a bug with this function. All usage of `formatQueryString` is for assigning a value to a `url.UrlWithStringQuery` instance `search` property. Which is typed as `string | null`, not `string | undefined`. The easiest (and what I believe the correct change here), is that `formatQueryString` should be changed to return `string | null`. 
 
Screen Shot 2022-11-04 at 12 13 16 PM

Furthermore, the `parseQueryString` function is always called with properties typed as `string | null`, hence that change. That function already has logic checking the truthiness of the parameter so it was an easy change.

Screen Shot 2022-11-04 at 12 15 15 PM

Ethan-Arrowood: Another solution is to add `?? null` to all of the `formatQueryString` calls i.e. `origUrl.search = formatQueryString(origQuery) ?? null` and `?? undefined` to all of the parseQueryString calls i.e. `const origQuery = parseQueryString(origUrl.search ?? undefined)`

packages/cli/src/util/client.ts
Ethan-Arrowood: The actual TS error here is: 

```
Argument of type 'ParsedUrlQuery' is not assignable to parameter of type 'string | URLSearchParams | Record<string, string | readonly string[]> | Iterable<[string, string]> | readonly [string, string][] | undefined'.
  Type 'Dict<string | string[]>' is missing the following properties from type 'readonly [string, string][]': length, concat, join, slice, and 19 more.
```

I believe `ParsedUrlQuery` is an alias for `Dict<string | string[]>`, which should satisfy the `Record<string | readonly string[]>` option but doesn't. 

Here is some type definitions from `@types/node`:
```
interface ParsedUrlQuery extends NodeJS.Dict<string | string[]> { }
interface Dict<T> {
  [key: string]: T | undefined;
}
```

Ethan-Arrowood: not sure why your comments didn't thread but see my response here: https://github.com/vercel/vercel/pull/8842/files#r1014334769

packages/cli/src/util/dev/builder.ts
Ethan-Arrowood: No, technically yes you can and TypeScript will treat it as a type cast, but i don't like that pattern. `data` is already receiving the `Serializable` type based on the definition for the event handler. By doing some type guarding within the conditional here it is a safer type cast than just saying `(data: { type: string }) => {...}`

packages/cli/src/util/dev/parse-query-string.ts
cb1kenobi: At a higher level, this function could, and possibly should, always return a `string`. If there's no `query`, then the result would simply be `""`. Then the body of the function could then be simplified down to:

```
return new URLSearchParams(query).toString();
```

TooTallNate (approved):

cb1kenobi (approved):

PR #8670, opened 2022-10-03
[cli] Use ensureLink() in `link` and `pull` commands

cb1kenobi:
The `link` and `pull` commands have essentially duplicate code from `ensureLink()`. This

Card: https://linear.app/vercel/issue/VCCLI-224/adopt-ensurelink-across-commands

The `dev` command should also be updated to use `ensureLink()`, however that may interfere with this in progress PR: https://github.com/vercel/vercel/pull/8666.

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [x] Issue from task tracker has a link to this PR

PR #8845, opened 2022-11-03
[cli] Allow `vc link` to overwrite existing link

TooTallNate:
There was a regression in #8670 which caused `vc link` to be a no-op when there was already a link to a Vercel project in the `.vercel` directory.

### DIFF
--- a/packages/cli/src/util/link/ensure-link.ts
+++ b/packages/cli/src/util/link/ensure-link.ts
@@ -34,7 +34,7 @@ export async function ensureLink(
 ): Promise<LinkResult | number> {
   let link = await getLinkedProject(client, cwd);
 
-  if (link.status === 'not_linked') {
+  if (opts.forceDelete || link.status === 'not_linked') {
     link = await setupAndLink(client, cwd, opts);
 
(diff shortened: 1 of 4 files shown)


styfle (approved):
Thanks! How about adding a test?

packages/cli/src/util/link/ensure-link.ts
```diff
@@ -34,7 +34,7 @@ export async function ensureLink(
 ): Promise<LinkResult | number> {
   let link = await getLinkedProject(client, cwd);
 
-  if (link.status === 'not_linked') {
+  if (opts.forceDelete || link.status === 'not_linked') {
```
cb1kenobi: It's possible that `link.status` will equal `"error"`, so you need to explicitly check the link status is `"linked"` in addition to `opts.forceDelete`:

```suggestion
  if ((link.status === 'linked' && opts.forceDelete) || link.status === 'not_linked') {
```

cb1kenobi (changes requested):

cb1kenobi (approved):
Looks good!

PR #8848, opened 2022-11-04
[edge] fix broken link on npmjs.org

dferber90:
Before this change clicking on "follow the documentation" from https://www.npmjs.com/package/@vercel/edge would lead to a 404 as the docs link is relative.

### Related Issues

To reproduce
1. Open https://www.npmjs.com/package/@vercel/edge
2. Click on "follow the documentation"
3. You will see a 404 page

### Note

Merging this change will not actually fix npm yet, as npm only updates the README with new releases. But at least this will be fixed in the next one then.

### DIFF
--- a/packages/edge/README.md
+++ b/packages/edge/README.md
@@ -1,4 +1,4 @@
 # `@vercel/edge`
 
 A set of utilities to help you deploy any framework on the Edge using Vercel.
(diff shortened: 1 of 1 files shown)


TooTallNate:
Thanks!

PR #8864, opened 2022-11-07
[cli] Add node_modules/.bin to PATH instead of running npx/yarn run

cb1kenobi:
`runDevCommand()` assumes the dev command is an npm package and thus uses `npx` or `yarn run` to execute it. In the case of a Hugo-based app, there is no npm package, so we want spawn to find Hugo in the `PATH`. Then for Node-based apps, instead of `npx`, spawn should find the command since `node_modules/.bin` has been added to the `PATH`.

### Related Issues

> https://github.com/vercel/customer-issues/issues/871

#### Tests

#### Code Review


TooTallNate:
Current tests seem ok with it. We should add a new test as well.

cb1kenobi:
@TooTallNate what do you think the new test would test that the existing tests aren't covering?

EndangeredMassa:
Can we add a hugo test to show this working?

cb1kenobi:
Closing in favor of https://github.com/vercel/vercel/pull/8890 due to several Hugo related files being updated thanks to prettier.

PR #8887, opened 2022-11-10
[build-utils][node][next][remix] add `operationType` to Lambda

EndangeredMassa:
These changes are necessary to support: https://github.com/vercel/api/pull/15237/files

Please keep comments on feature design in that PR.

This was tested with Build Output API, but more work needs to be done to get the builders updated.


packages/build-utils/src/lambda.ts
```diff
@@ -123,7 +137,10 @@ export class Lambda {
         '"regions" is not a string Array'
       );
     }
+
     this.type = 'Lambda';
+    this.operationType = operationType;
+    this.bundledFunctions = bundledFunctions;
```
EndangeredMassa: **question:** is there a test I should write around this?

It doesn't sound super useful to test that the constructor can assign these values, but that's about the only logic in this PR.

packages/build-utils/src/lambda.ts
```diff
@@ -24,6 +32,8 @@ export interface LambdaOptionsBase {
   supportsMultiPayloads?: boolean;
   supportsWrapper?: boolean;
   experimentalResponseStreaming?: boolean;
+  operationType?: OperationType;
```
TooTallNate: How would `operationType: 'API'` be different than `bundledFunction: { API: 1 }`? Is it to differentiate the fact that bundling happened at all or not?

ijjk: Do we have a doc discussing the goals for these fields? `bundledFunction: { API: 1 }` seems a bit vague/limiting unless I'm misunderstanding, it could provide additional context with a different shape/fields

EndangeredMassa: I think exposing bundled functions is out of scope for this change. I am going to open a separate discussion about if and how we'd want to handle this.

I've updated the PR to use only single values.

packages/build-utils/src/lambda.ts
```diff
@@ -47,6 +48,7 @@ interface GetLambdaOptionsFromFunctionOptions {
 
 export class Lambda {
   type: 'Lambda';
+  operationType?: string;
```
cb1kenobi, line 56: Non-blocking: I got to thinking that if `operationType` is not required, then what is the default when the value is `undefined`? Maybe add a hint or something?

```suggestion
  operationType?: string; // default "API"
```

cb1kenobi (approved):
Looks good!

packages/build-utils/src/lambda.ts
```diff
@@ -47,6 +48,7 @@ interface GetLambdaOptionsFromFunctionOptions {
 
 export class Lambda {
   type: 'Lambda';
+  operationType?: string; // defaults to "API" (or "ISR" for prerenders)
```
TooTallNate: > (or "ISR" for prerenders)

Should we set this in the `Prerender` constructor?

cb1kenobi (approved):

packages/build-utils/src/prerender.ts
```diff
@@ -37,7 +37,9 @@ export class Prerender {
     this.expiration = expiration;
 
     this.lambda = lambda;
-    this.lambda.operationType = this.lambda.operationType || 'ISR';
+    if (this.lambda) {
```
cb1kenobi: Can `lambda` be undefined/null?

EndangeredMassa: I saw an error saying so, but I now realize that was is a less-typed test passing it in. I'll update this.

EndangeredMassa: I removed the guard and fixed the test that caused this issue: [`e50c3ac` (#8887)](https://github.com/vercel/vercel/pull/8887/commits/e50c3acf74dd6ddd8d69ab5b47aeb52527102a8f)

packages/build-utils/src/lambda.ts
styfle: Should the type be more strict to avoid typos and make it self documenting?

```ts
operationType?: 'API' | 'ISR'
```

packages/build-utils/src/lambda.ts
```diff
@@ -123,7 +126,9 @@ export class Lambda {
         '"regions" is not a string Array'
       );
     }
+
     this.type = 'Lambda';
+    this.operationType = operationType;
```
styfle, line 136: Should this default to API when undefined?

packages/remix/src/build.ts
```diff
@@ -278,6 +278,7 @@ async function createRenderFunction(
     runtime: nodeVersion.runtime,
     shouldAddHelpers: false,
     shouldAddSourcemapSupport: false,
+    operationType: 'SSR',
```
styfle, line 281: Whats the difference between API and SSR?

packages/build-utils/src/lambda.ts
EndangeredMassa: We could, but the current plan was to default it only for uploading to deployment metadata: https://github.com/vercel/api/pull/15237/files#diff-ce539d6a5c1ccfc0866021b6133828ca803f22791586c1e79037bbcab134cbabR130-R135

I liked keeping the original value (`undefined`, potentially) up until it was displayed. That way, if we use this for anything else, we can know if it was intentionally set or not.

packages/remix/src/build.ts
EndangeredMassa: I think the intent is that `SSR` returns HTML and `API` returns something else. These labels can be whatever makes the most sense for a given framework.

In that sense, "SSR" seemed like a reasonable default label for this. We could change it to a more Remix-specific value, but I'm not sure what that would be here. This is the main render function, right? Maybe we call this "Render".

@TooTallNate: you added this builder originally. What do you think makes sense here?

TooTallNate: SSR is appropriate in this case I think.

TooTallNate (approved):

packages/remix/src/build.ts
styfle: Do we need to document those differences here or will it be in the BOA v3 docs?

The comment above is confusing to me since its not exhaustive:

```
// defaults to "API" (or "ISR" for prerenders)
```

EndangeredMassa: That's fair. In this repo, that comment doesn't make sense.

There will be a note about this in BOAv3, but this PR's code isn't about BOA.

How about we generically say this here:

```
// framework-specific lambda type label
```

[`958f1fb` (#8887)](https://github.com/vercel/vercel/pull/8887/commits/958f1fb70e6fac69ec7e3f2ca4933e599bb22432)

styfle: Lets remove the comment. We should either add JSDoc for all of these or none of these properties.

packages/next/src/index.ts
```diff
@@ -1072,6 +1072,7 @@ export const build: BuildV2 = async ({
           handler: '___next_launcher.cjs',
           runtime: nodeVersion.runtime,
           ...lambdaOptions,
+          operationType: 'SSR',
```
ijjk, line 1075: Is this just being added for testing? This specific line is for legacy Next.js apps (`< 7.0.2`)

EndangeredMassa: It was added for completeness. If we don't add this and someone deploys an old next app, the richer deployments output will show an "API" label instead.

packages/remix/src/build.ts
EndangeredMassa: I don't agree that we should all-or-nothing these. Some information is better than none. I do agree that jsdoc'ing all of them would be good.

I can make this one a jsdoc for now. [`a0b30ee` (#8887)](https://github.com/vercel/vercel/pull/8887/commits/a0b30ee6fb4c3047d899468cff5e0f38926e20c1)

packages/build-utils/src/lambda.ts
```diff
@@ -47,6 +48,10 @@ interface GetLambdaOptionsFromFunctionOptions {
 
 export class Lambda {
   type: 'Lambda';
+  /**
+   * framework-specific lambda type label
```
styfle: Can we document the possible options here? The current comment doesnt make any sense. Imagine you're seeing this for the first time.

EndangeredMassa: Yeah, I can add some examples: [`76a9761` (#8887)](https://github.com/vercel/vercel/pull/8887/commits/76a9761708ac0ea924ac04b6fab119acbea5cf36)

packages/next/src/index.ts
styfle: What about existing deployments that are empty? Is there going to be a migration or just leave them empty?

styfle (approved):

packages/next/src/index.ts
EndangeredMassa: Good question. I asked internally.

PR #8888, opened 2022-11-10
[cli] Remove `--runInBand` from Jest

TooTallNate:


cb1kenobi:
Do you think this going to increase stability?

TooTallNate:
I don't think it would affect stability, but would make running the tests faster. Granted, Unit tests are not the slowest part of our CI, but still seems like a win to me.

cb1kenobi:
@TooTallNate Looks this the CLI tests are getting hung up on the same Gatsby test that my PRs are running into: https://github.com/vercel/vercel/blob/main/packages/cli/test/integration.js#L3400-L3431.

After the first deploy, it waits a minute, then checks for a `.cache` directory which doesn't exist. Any idea why this all the sudden is breaking?

cb1kenobi (approved):

PR #8894, opened 2022-11-11
[static-build] support gatsby-config as typescript

dglsparsons:
### Related Issues

A customer issue was raised that web-vitals analytics were not working for their Gatsby application, even though this is meant to be zero config. 

It turns out the issue is due to their `gatsby-config` file being declared as a `ts` file, rather than `js`. This is perfectly valid and supported in Gatsby. 

However, the static-build modifications that are made to automatically add the `gatsby-plugin-vercel` only apply to existing `js` files. 

This lead to their deployments containing both a `js` and `ts` version of `gatsby-config`. Luckily, the `ts` version has higher precedence, so _only_ web-vitals were affected. 

Closes https://linear.app/vercel/issue/FLA-364/investigate-gatsby-and-nuxt-data-issues

#### Tests

- [X] The code changed/added as part of this PR has been covered with tests
- [X] All tests pass locally with `yarn test-unit`

#### Code Review

- [X] This PR has a concise title and thorough description useful to a reviewer
- [X] Issue from task tracker has a link to this PR


packages/static-build/src/utils/gatsby.ts
```diff
@@ -8,50 +8,106 @@ import {
...
+    await updateGatsbyTsConfig(gatsbyConfigPathTs);
+    return;
+  }
 
-  const gatsbyConfigPath = path.join(dir, gatsbyConfigName);
   console.log(
-    `Injecting Gatsby.js analytics plugin "${gatsbyPluginPackageName}" to \`${gatsbyConfigPath}\``
+    `Injecting Gatsby.js analytics plugin "${GATSBY_PLUGIN_PACKAGE_NAME}" to \`${gatsbyConfigPathJs}\``
   );
+  await addGatsbyPackage(dir);
```
EndangeredMassa, line 40: **suggestion (non-blocking):** only call `addGatsbyPackage` once higher up in function

It looks like we always need to call `addGatsbyPackage`. Can we move it up to line ~26 to make it clearer that this always happens?

packages/static-build/src/utils/gatsby.ts
```diff
@@ -8,50 +8,106 @@ import {
...
     await writePackageJson(dir, pkgJson);
   }
+}
 
-  if (await fileExists(gatsbyConfigPath)) {
-    await fs.rename(
-      gatsbyConfigPath,
-      gatsbyConfigPath + '.__vercel_builder_backup__.js'
-    );
+async function updateGatsbyTsConfig(configPath: string): Promise<void> {
```
EndangeredMassa, line 66: **suggestion (soft-blocking):** extract and re-use common body of file template

In this or a follow-up PR: It looks like the the only differences between the TS and JS versions of this function are the imports/exports. Can we extract the bodies to make it (1) clearer that they are the same and (2) easier to keep in sync if we need to change it later?

We could do this in a single function, like:

```ts
updateGatsbyConfig(path: string, moduleType: 'commonjs' | 'module') {
  const header = ''; // require- or import-based code
  const footer = ''; // export- or module.exports-based code
	const fileBody = `
    ${header}

    // current body

    ${footer}
  `;

  // write file
}
```

But separate functions would work well too.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -8,50 +8,106 @@ import {
...
+
+  // https://github.com/gatsbyjs/gatsby/blob/a6ecfb2b01d761e8a3612b8ea132c698659923d9/packages/gatsby/src/services/initialize.ts#L113-L117
+  preferDefault(userConfig)
+);
+if (!vercelConfig.plugins) {
+  vercelConfig.plugins = [];
+}
+
+const hasPlugin = vercelConfig.plugins.find(
+  (p: any) =>
```
EndangeredMassa: **suggestion (non-blocking):** replace `any` with types

While we're here, we should get rid of `any` where possible.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -8,50 +8,106 @@ import {
...
 
-    await fs.writeFile(
-      gatsbyConfigPath,
-      `const userConfig = require("./gatsby-config.js.__vercel_builder_backup__.js");
+  await fs.writeFile(
+    configPath,
+    `import userConfig from "./gatsby-config.ts.__vercel_builder_backup__.ts";
+
+// https://github.com/gatsbyjs/gatsby/blob/354003fb2908e02ff12109ca3a02978a5a6e608c/packages/gatsby/src/bootstrap/prefer-default.ts
+const preferDefault = (m: any) => (m && m.default) || m;
```
EndangeredMassa, line 75: **suggestion (non-blocking):** replace `any` with types

While we're here, we should get rid of `any` where possible.

EndangeredMassa (approved):

packages/static-build/src/utils/gatsby.ts
dglsparsons: The types are also different, which is why I opted for having them totally separate - it felt a bit mad having the shared parts templated in, and lost a bit of clarity around what was going on. Happy to change if you disagree though.

packages/static-build/src/utils/gatsby.ts
dglsparsons: I duplicated it as I wanted it to happen after the `console.log` statement - as `addGatsbyPackage` also writes logs. 

I tweaked the `console.log` statement (L31, L39) to reflect whether it was updating `gatsby-config.ts` or `gatsby-config.js`. Thought it was nicer to duplicate `addGatsbyPackage` than duplicating the `ts/js` condition.

packages/static-build/src/utils/gatsby.ts
dglsparsons: This one is taken from the Gatsby source-code (above link), which annoyingly, uses `any`. 
https://github.com/gatsbyjs/gatsby/blob/354003fb2908e02ff12109ca3a02978a5a6e608c/packages/gatsby/src/bootstrap/prefer-default.ts

EndangeredMassa:
@dglsparsons sounds good on those points. We're waiting to get this [other PR](https://github.com/vercel/vercel/pull/8911) out, then we can merge this PR and release it.

EndangeredMassa:
It'll be one or the other of these that need to get in first:

- https://github.com/vercel/vercel/pull/8911
- https://github.com/vercel/vercel/pull/8913

PR #8913, opened 2022-11-15
[cli] add support for JSON5 and skip when no rootDirectory

Ethan-Arrowood:
### Related Issues

Adds support for comments in JSON configs and skips operation when root directory is null or `'.'`

#### Tests

#### Code Review


packages/cli/src/commands/build.ts
```diff
@@ -304,7 +304,12 @@ async function doBuild(
     ...pickOverrides(localConfig),
   };
 
-  await setMonorepoDefaultSettings(cwd, workPath, projectSettings, output);
+  if (
+    projectSettings.rootDirectory !== null &&
+    projectSettings.rootDirectory !== '.'
+  ) {
+    await setMonorepoDefaultSettings(cwd, workPath, projectSettings, output);
+  }
```
EndangeredMassa, line 312: Why exactly do we need to skip in this case?

EndangeredMassa (approved):

packages/cli/src/commands/build.ts
Ethan-Arrowood: We thought about it like in the case of `vercel/vercel`. There is no way to really know what the user wants to output and we just should rely on whatever their build script is in this instance. As for example, we use turbo build, which then eventually calls `packages/cli/scripts/build.ts` which outputs the output of cli build into the root `public` directory.

packages/cli/test/unit/commands/build.test.ts
```diff
@@ -8,6 +8,7 @@ import { defaultProject, useProject } from '../../mocks/project';
 import { useTeams } from '../../mocks/team';
 import { useUser } from '../../mocks/user';
 import { setupFixture } from '../../helpers/setup-fixture';
+import JSON5 from 'json5';
 // import execa from 'execa';
```
TooTallNate, line 12:
```suggestion
```

cb1kenobi (approved):

PR #8921, opened 2022-11-16
[fs-detectors] Remove `limitedRoutes`

TooTallNate:
These were only used for FS API v2, and are no longer needed.


cb1kenobi (approved):
Awesome!

Ethan-Arrowood (approved):

PR #8926, opened 2022-11-16
[cli] support package.json based turbo config in vc build automatic build/install command feature

Ethan-Arrowood:
### Related Issues

Adds support for package.json based turbo configuration. 

Includes test.

Confirmed with Turbo that `turbo.json` takes precedence over `package.json` based config so that is how it is processed here.

#### Tests

#### Code Review

### DIFF
--- a/packages/cli/src/util/build/monorepo.ts
+++ b/packages/cli/src/util/build/monorepo.ts
@@ -69,9 +69,9 @@ export async function setMonorepoDefaultSettings(
         hasBuildPipeline = true;
       }
     } else if (packageJsonBuf !== null) {
-      const packageJson = JSON.parse(packageJsonBuf.toString('utf-8'));
+      const packageJSON = JSON.parse(packageJsonBuf.toString('utf-8'));
 
-      if (packageJson?.turbo?.pipeline?.build) {
+      if (packageJSON?.turbo?.pipeline?.build) {
         hasBuildPipeline = true;
       }
     }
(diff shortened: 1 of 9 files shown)


EndangeredMassa (approved):

TooTallNate (approved):

cb1kenobi (approved):

Ethan-Arrowood:
Removing auto merge and adding another fix to this PR :)

packages/cli/src/util/build/monorepo.ts
```diff
@@ -55,21 +55,40 @@ export async function setMonorepoDefaultSettings(
...
-        'Missing required `build` pipeline in turbo.json. Skipping automatic setting assignment.'
+        'Missing required `build` pipeline in turbo.json or package.json Turbo configuration. Skipping automatic setting assignment.'
       );
       return;
     }
 
     setCommand(
       'buildCommand',
-      `cd ${relativeToRoot} && npx turbo run build --filter=${projectName}...`
+      `cd ${relativeToRoot} && npx turbo run build --filter={./${relative(
```
TooTallNate, line 88: Do we need to worry about Windows here and have `.\\`?

Ethan-Arrowood: Great question - i assume CI will fail if thats the case. But let me ask Turbo right now too

Ethan-Arrowood: https://github.com/vercel/vercel/actions/runs/3483582850/jobs/5827228329#step:8:2934 ironically CI timeout on an Nx test, but all the Turbo ones seemed to pass on Windows

EndangeredMassa: I think we should always `path.join` unless we specifically have a good reason not to. It's more likely to save us the trouble, right?

TooTallNate: It's kinda annoying, but `path.join()` will strip the `./`:

```
> path.join('.', 'foo')
'foo'
```

PR #8927, opened 2022-11-16
[cli] Create symlinks for duplicate edge functions in `vc build`

TooTallNate:
Similar to how Lambda instances are de-duped using symlinks, do the same for EdgeFunction instances.

TODO: backend needs to support this first, and also add tests here.


TooTallNate:
Superseded by https://github.com/vercel/vercel/pull/11027.

PR #8931, opened 2022-11-17
[cli] Add `vc link --repo` flag

TooTallNate:
Adding a new `--repo` flag to `vc link` which is a new linking style to link to a Git URL rather then directly to a single Vercel Project. This allows for multiple Projects to be linked simultaneously, which is useful for monorepo setups.

Utilization of this new linking style in other commands will be done in a follow-up PR.


TooTallNate:
Added "alpha" status warnings in a few places. As discussed today, let's merge this one and continue iterating.

packages/cli/src/commands/link/index.ts
```diff
@@ -12,6 +14,7 @@ const help = () => {
   ${chalk.dim('Options:')}
 
     -h, --help                     Output usage information
+    -r, --repo                     Link multiple projects based on Git repository
```
EndangeredMassa: **question (non-blocking):** ‏Should we remove this so that a random user won't see it and try to use it?

If we keep it, can we add the `(alpha)` label here too?

packages/cli/src/util/link/repo.ts
```diff
@@ -0,0 +1,269 @@
...
+  let { rootPath, repoConfig, repoConfigPath } = await getRepoLink(cwd);
+  if (rootPath) {
+    output.debug(`Found Git repository root directory: ${rootPath}`);
+  } else {
+    throw new Error('Could not determine Git repository root directory');
+  }
+
+  // If `repoPath` is defined then this should always be defined
+  if (!repoConfigPath) {
+    throw new Error('`repoConfigPath` not defined');
```
EndangeredMassa: **suggestion (soft-blocking):** add details to error

Does TypeScript not recognize this path is unreachable?

I think we should add the root path for context, just in case this is ever hit for some reason.

```suggestion
    throw new Error(`'repoConfigPath' not found in ${rootPath}`);
```

packages/cli/src/util/link/repo.ts
```diff
@@ -0,0 +1,269 @@
...
+    const stat = await lstat(gitConfigPath).catch(err => {
+      if (err.code !== 'ENOENT') throw err;
+    });
+    if (stat) {
+      return current;
+    }
+  }
+}
+
+export function* traverseDirectories(start: string) {
```
EndangeredMassa: **suggestion (non-blocking):** clarify direction‏

```suggestion
export function* traverseUpDirectories(start: string) {
```

packages/cli/src/util/link/repo.ts
```diff
@@ -0,0 +1,269 @@
...
+  } else {
+    throw new Error('Could not determine Git repository root directory');
+  }
+
+  // If `repoPath` is defined then this should always be defined
+  if (!repoConfigPath) {
+    throw new Error('`repoConfigPath` not defined');
+  }
+
+  if (!repoConfig) {
```
EndangeredMassa, line 78: **suggestion (non-blocking):** break into more functions

This function (and this condition block specifically) is getting long. I suggest breaking this up into a few functions that will make the logical flow clearer.

packages/cli/src/util/link/repo.ts
```diff
@@ -0,0 +1,269 @@
...
+    // Not yet linked, so prompt user to begin linking
+    let shouldLink =
+      yes ||
+      (await confirm(
+        client,
+        `Link to Git repository at ${chalk.cyan(
+          `“${toHumanPath(rootPath)}”`
+        )}?`,
```
EndangeredMassa: **suggestion (non-blocking):** ‏make it clear here too that this supports multiple projects

```suggestion
        )} to your Project(s)?`,
```

packages/cli/src/util/link/repo.ts
```diff
@@ -0,0 +1,269 @@
...
+    output.print(
+      prependEmoji(
+        `Linked to ${link(repoUrl)} under ${chalk.bold(
+          org.slug
+        )} (created ${VERCEL_DIR}${
+          isGitIgnoreUpdated ? ' and added it to .gitignore' : ''
+        })`,
+        emoji('link')
+      ) + '\n'
+    );
```
EndangeredMassa, line 199: **suggestion (non-blocking):** ‏Until we implement this, can we log that the user should do it?

```suggestion
    );
    console.log(`Created ${repoConfigPath}. Ensure ${VERCEL_DIR} is added to your ".gitignore".`);
```

packages/cli/src/commands/link/index.ts
```diff
@@ -68,17 +80,36 @@ export default async function main(client: Client) {
...
+    );
+  } else {
+    cwd = process.cwd();
+  }
 
-  if (typeof link === 'number') {
-    return link;
+  if (argv['--repo']) {
+    client.output.warn(
+      `The ${cmd('--repo')} flag is in alpha, please report issues`
```
EndangeredMassa, line 98: **suggestion (non-blocking):** ‏Can we note here (and wherever else we say `(alpha)`) that this linking style will only work with specific commands?

EndangeredMassa (dismissed):

packages/cli/src/commands/link/index.ts
TooTallNate: It's in the examples below as well, so let's just mark alpha here.

packages/cli/src/util/link/repo.ts
TooTallNate: Fixed the types to remove this condition

packages/cli/src/util/link/repo.ts
TooTallNate: I added the gitignore logic now.

EndangeredMassa (dismissed):

PR #8942, opened 2022-11-17
[cli] Add `vc rollback` command

cb1kenobi:
The `vc rollback` command provides the ability to redeploy a previous deployment and check the status of a rollback request.

#### Requesting a rollback

    vc rollback <id | url>

Upon requesting a rollback, the command will being a status polling loop for 3 minutes. This timeout can be adjusted via the `--timeout <value>` option which accepts time formats such as `30s` or `2m`. A timeout of `0` (zero) will skip the status polling and immediately exit after successfully requesting a rollback.

#### Querying rollback status 

    vc rollback
    vc rollback status

The `status` action will return the most recent rollback info within the last 3 minutes.

### Related Issues

> https://linear.app/vercel/issue/HIT-117/cli-add-support-for-vc-rollback-deployid
> https://linear.app/vercel/issue/HIT-118/cli-add-support-for-vc-rollback-[status]

#### Tests

#### Code Review


packages/cli/src/util/link/ensure-link.ts
```diff
@@ -38,6 +38,7 @@ export async function ensureLink(
     (link.status === 'linked' && opts.forceDelete) ||
     link.status === 'not_linked'
   ) {
+    opts.link = link;
     link = await setupAndLink(client, cwd, opts);
```
EndangeredMassa: **question (non-blocking):** What is this change doing, exactly?

It seems that `opts.link` is populated so that `setupAndLink` can later not call `getLinkedProject`. In this case (from `ensure-link.ts`), that means there's no change because in this condition `link` will always be truthy.

In the other place where `setupAndLink` is called (from `commands/dev/dev.ts`), `link` is not passed in, which will cause `getLinkedProject to be called. Should that invocation be updated too?

If so, then all invocations of `setupAndLink` would be called with a truthy `link`, making the change in `setup-and-link.ts` effectively a no-op.

What's the intent behind this change?

packages/cli/src/util/rollback/get-deployment-info.ts
```diff
@@ -0,0 +1,36 @@
...
+  contextName: string,
+  deployId: string
+): Promise<Deployment> {
+  const deployment = handleCertError(
+    client.output,
+    await getDeploymentByIdOrHost(client, contextName, deployId)
+  );
+
+  if (deployment === 1) {
+    throw new Error('Failed to get deployment');
```
EndangeredMassa: **suggestion (soft-blocking):** add context to error

In this or a follow-up PR: Can we add some context to this error? Maybe it should say `Failed to get deployment "${deployId}" in scope "${contextName}"`

packages/cli/src/util/rollback/render-alias-status.ts
```diff
@@ -0,0 +1,19 @@
+import chalk from 'chalk';
+
+/**
+ * Stylize the alias status label.
+ * @param {AliasStatus} status - The status label
+ * @returns {string}
+ */
+export default function renderAliasStatus(status: string): string {
```
EndangeredMassa, line 8: **question (non-blocking):** Is there a difference between alias status and rollback status?

packages/cli/src/util/rollback/request-rollback.ts
```diff
@@ -0,0 +1,87 @@
...
+    `Fetching deployment "${deployId}" in ${chalk.bold(contextName)}…`
+  );
+
+  let deployment;
+  try {
+    deployment = await getDeploymentInfo(client, contextName, deployId);
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      output.error(err.toString());
+    }
```
EndangeredMassa: **suggestion (soft-blocking):** Log error regardless of type

If the this condition is false, some non-Error value was thrown? Why do we not want to log it as well?

packages/cli/src/util/rollback/status.ts
```diff
@@ -0,0 +1,198 @@
...
+import renderAliasStatus from './render-alias-status';
+import sleep from '../sleep';
+
+interface RollbackAlias {
+  alias: {
+    alias: string;
+    deploymentId: string;
+  };
+  id: string;
+  status: 'completed' | 'in-progress' | 'pending' | 'failed';
```
EndangeredMassa, line 23: **question (non-blocking):** Should this refer to the `RollbackJobStatus` type or no because it's only a subset?

packages/cli/src/util/rollback/status.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+interface RollbackAliasesResponse {
+  aliases: RollbackAlias[];
+  pagination: PaginationOptions;
+}
+
+/**
+ * Continuously checks a deployment status until it has succeeded, failed, or
+ * taken longer than 3 minutes.
```
EndangeredMassa: **suggestion (non-blocking):**

```suggestion
 * taken longer than the timeout (default 3 minutes).
```

packages/cli/src/util/rollback/status.ts
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      // if we've done our first poll and not rolling back, then print the
+      // requested at date/time
+      if (counter++ === 0 && !deployment) {
+        spinnerMessage += ` requested at ${formatDate(requestedAt)}`;
+      }
+      output.spinner(`${spinnerMessage}…`);
+
+      await sleep(500);
```
EndangeredMassa: **question (non-blocking):** what's the "right" amount to wait here?

I don't know. Responding at most (ish) a half second to a rollback reaching a terminal state is pretty fast, but if we could do it faster, why not?

I wonder if this should be ~250ms.

packages/cli/src/util/rollback/status.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+  } finally {
+    output.stopSpinner();
+  }
+}
```
EndangeredMassa, line 152: **suggestion (soft-blocking):** refactor to make logic clearer

In this or a follow-up PR: This logic includes many loops (some of which are infinite) and different logical steps in one function. I suggest refactoring this (mostly with function extraction) to make it clearer what is happening. I'd be happy to pair on this.

packages/cli/test/mocks/deployment.ts
```diff
@@ -140,6 +140,51 @@ function setupDeploymentEndpoints() {
...
+      created: deployment.createdAt,
+      ready: deployment.ready,
+      buildingAt: deployment.buildingAt,
+      creator: {
+        uid: deployment.creator?.uid,
+        username: deployment.creator?.username,
+      },
+      target: deployment.target,
+      ownerId: undefined, // ?
+      projectId: undefined, // ?
```
EndangeredMassa, line 181: **question (non-blocking):** did it not seem to matter what these values are in test?

EndangeredMassa (approved):
Non of my comments are blocking, but I think getting more reviews is worth it before merging.

packages/cli/src/util/link/ensure-link.ts
cb1kenobi: What I found was this error https://github.com/vercel/vercel/blob/49f453742b190c989447bedf3c85947287086688/packages/cli/src/util/projects/link.ts#L130-L136 was being rendered twice because `getLinkedProject()` was getting called twice. I figured, if we already fetched the project, no sense doing it twice, so I pass the `link` through.

I did notice `commands/dev/dev.ts` calls `setupAndLink()`, but I missed that `getLinkedProject()` is called right before.

I believe we should make `link` a required property of `SetupAndLinkOptions`, pass `link` into `setupAndLink()` in `commands/dev/dev.ts`, and remove the `getLinkedProject()` call in `setup-and-link.ts`.

packages/cli/src/util/rollback/get-deployment-info.ts
cb1kenobi: Good idea!

packages/cli/src/util/rollback/render-alias-status.ts
cb1kenobi: Yeah, a rollback request has a status like `succeeded` or `failed` and that encompasses the entire rollback state. When a rollback occurs, it also rolls back aliases and each of those are distinct operations with their own statuses.

packages/cli/src/util/rollback/request-rollback.ts
cb1kenobi: Fixed. Removed the condition, but favor calling `toString()` as many of the errors caught are error messages and we'd rather not show a stack trace.

packages/cli/src/util/rollback/status.ts
cb1kenobi: No, these alias status are different than the job statuses.

packages/cli/src/util/rollback/status.ts
cb1kenobi: Fixed.

packages/cli/src/util/rollback/status.ts
cb1kenobi: I had 250 at one point, but then bumped it to 500 since since it was basically hitting the endpoint 4 times a second and that seems like a lot.

packages/cli/test/mocks/deployment.ts
cb1kenobi: Good question. I made these `undefined` because the rollback command doesn't use them and production returned them as `undefined` in my testing. So, we should be ok.

packages/cli/src/util/rollback/status.ts
```diff
@@ -0,0 +1,255 @@
...
+      contextName,
+      toDeploymentId
+    );
+    deploymentInfo = `${chalk.bold(deployment.url)} (${toDeploymentId})`;
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      output.debug(
+        `Failed to get deployment url for ${toDeploymentId}: ${err.toString()}`
+      );
+    }
```
EndangeredMassa: **suggestion (non-blocking):** remove `instanceof` check here as well?

EndangeredMassa (approved):

packages/cli/src/util/rollback/status.ts
cb1kenobi: Fixed. I do this to make TypeScript happy, but I can make TypeScript happy without the instanceof.

packages/cli/src/commands/rollback.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  ${chalk.gray('–')} Show the status of any current pending rollbacks
+
+    ${chalk.cyan(`$ ${getPkgName()} rollback`)}
+    ${chalk.cyan(`$ ${getPkgName()} rollback status`)}
+    ${chalk.cyan(`$ ${getPkgName()} rollback status --timeout 30s`)}
+
+  ${chalk.gray('–')} Rollback a deployment using id or url
+
+    ${chalk.cyan(`$ ${getPkgName()} rollback <deploymnent id/url>`)}
```
andymckay:
```suggestion
    ${chalk.cyan(`$ ${getPkgName()} rollback <deployment id/url>`)}
```

andymckay: I've noticed in that most other places we use square brackets to denote a variable to entered and use `-` in spaces. I'm not sure if we have standard for `or` though. Examples:

https://vercel.com/docs/cli/alias 👉🏾 `vercel alias set [deployment-url] [custom-domain]`
https://vercel.com/docs/cli/list 👉🏾 `vercel list [project-name] [--meta foo=bar]`

So I think for consistency this could be:

`[deployment-id or url]`

But I'm not sure about the `or`, but it feels more clear than a `/`

cb1kenobi: Angle brackets denotes required and square brackets denotes optional. When `vc rollback` is executed without a deployment id/url, it polls the status and in that context deployment id/url is optional.

However in this particular case, the context is an _example_ of how to perform a rollback in which case the deployment id/url is required and hence angle brackets.

cb1kenobi: Good catch, will fix.

andymckay: Ah cool thank you @cb1kenobi 🙇🏾  It's interesting that the docs say:

```
vercel alias set [deployment-url] [custom-domain]
```

And the CLI for `vc alias --help` says:

```
 set   <deployment> <alias>   Create a new alias
```

I was only looking at the docs and not the other help files. Perhaps this when we get to the idea of syncing our help text with docs this will all get fixed up 🙇🏾

cb1kenobi: I'm of the opinion that the CLI is correct and the docs are not.

andymckay (approved):
🚢 it, looks great!

PR #8949, opened 2022-11-18
[cli] Fix `vc build --cwd` with a relative path

TooTallNate:
`--cwd` is handled globally in the CLI entrypoint file, so if a relative path is used then it would be applied twice in `vc build` since it was _also_ handling `--cwd`. So remove `vc build`'s logic for it.

Fixes https://github.com/vercel/vercel/discussions/8948.


Ethan-Arrowood (approved):

cb1kenobi (approved):
Good find!

cb1kenobi:
It's funny, @EndangeredMassa and I were just chatting about the `process.chdir()` in the `index.js` the other day and I didn't know the `build` command also did a `chdir()`.

PR #8950, opened 2022-11-18
[cli] Fix failing tests after Node 18 default change

TooTallNate:


packages/cli/test/dev/integration-3.test.ts
```diff
@@ -70,27 +70,6 @@ test(
...
-/*
-test(
-  '[vercel dev] 05-gatsby',
-  testFixtureStdio('05-gatsby', async (testPath: any) => {
-    await testPath(200, '/', /Gatsby Default Starter/m);
-  })
-);
-*/
-test(
-  '[vercel dev] 06-gridsome',
```
TooTallNate: Removed this test since Gridsome hasn't had a new release in over 2 years, and the current version does not build on Node 18.

cb1kenobi: 👍

packages/cli/test/fixtures/unit/commands/build/lambda-with-128-memory/.vercel/project.json
```diff
@@ -2,6 +2,7 @@
   "orgId": ".",
   "projectId": ".",
   "settings": {
+    "nodeVersion": "16.x",
```
styfle: I would prefer this go in package.json engines

I'm also curious why this test needs to be pinned?

packages/cli/test/dev/integration-3.test.ts
styfle: Could it be pinned to Node.js 16 using package.json engines?

TooTallNate: The problem then is that our tests are using Node 14 still, so then `yarn install` fails when testing `vc dev`.

TooTallNate:
The remaining test failure is on the infra side, so will merge this and get a release out to fix that.

packages/cli/test/dev/integration-3.test.ts
styfle: Lets pin it to node 14 then 🙂

packages/cli/test/fixtures/unit/commands/build/lambda-with-128-memory/.vercel/project.json
styfle: I see now, the assertion was checking for `nodejs16.x` already. I think that can be changed to a wildcard with object.match() and that way it doesnt need to be pinned (since this test is about checking memory and not runtime)

styfle (approved):
Thanks!

packages/cli/test/dev/integration-3.test.ts
TooTallNate: Done.

packages/cli/test/fixtures/unit/commands/build/lambda-with-128-memory/.vercel/project.json
TooTallNate: Removed the runtime check entirely from this test.

Ethan-Arrowood (approved):
great work on this!

PR #8961, opened 2022-11-22
[fs-detectors][cli] Implement project selector

Ethan-Arrowood:
### Related Issues

Implements the project selector to be used by the `vc link --repo` command. Also, fixes a bug in fs-detectors

#### Tests

#### Code Review


packages/cli/test/helpers/setup-fixture.ts
```diff
@@ -46,6 +46,23 @@ export function setupFixture(fixtureName: string) {
   return cwd;
 }
 
+export function setupFixtureFromTemplate(
+  fixture: string,
+  template: string,
+  nestedPaths?: string[]
+) {
+  const cwd = setupFixture(fixture);
+  const templateFixture = path.join(fixturesRoot!, 'unit', template);
```
EndangeredMassa, line 55: **suggestion (soft-blocking):** ensure `fixturesRoot` is populated

Are we sure that `fixturesRoot` is populated here? It looks like it's only populated if `setupFixture` has been called, but if `packages/cli/test/unit/util/input/project-selector.test.ts` is run alone, that never happens.

It seems like this would work in a full run, but fail if you isolate and run this test file.

In this or a follow-up PR: Can we ensure that `fixturesRoot` is populated here?

packages/fs-detectors/src/detectors/local-file-system-detector.ts
```diff
@@ -44,7 +44,7 @@ export class LocalFileSystemDetector extends DetectorFilesystem {
     };
     return dir.map(dirent => ({
       name: dirent.name,
-      path: path.join(dirPath, dirent.name),
```
EndangeredMassa, line 47: **question (non-blocking):** what bug(s) are being fixed with the `fs-detectors` changes?

Would it make sense to extract `fs-detectors` changes and deploy them before the project selector?

packages/cli/src/util/input/project-selector.ts
```diff
@@ -0,0 +1,69 @@
...
+    const packagesPaths = await Promise.all(
+      workspaces.map(workspace =>
+        getWorkspacePackagePaths({ fs: localFS, workspace })
+      )
+    );
+    const packages = await Promise.all(
+      packagesPaths.flat().map(async packagePath => {
+        const packagePackageJSONBuffer = await localFS
+          .readFile(path.join(packagePath, 'package.json'))
+          .catch(() => null);
```
EndangeredMassa, line 41: **suggestion (soft-blocking):** debug log error

In this or a follow-up PR: Can we debug log any error that comes from this? I'm always skeptical of ignoring errors.

packages/cli/test/unit/util/input/project-selector.test.ts
```diff
@@ -0,0 +1,88 @@
...
+        client = new MockClient();
+      });
+
+      describe('when repo.json is not present', () => {
+        it('should return list of projects based on workspace list', async () => {
+          const projectSelectorPromise = projectSelector(client, cwd);
+          await expect(client.stderr).toOutput('Select projects');
+          client.stdin.write('a\n');
+          const actual = await projectSelectorPromise;
+          expect(actual.sort()).toStrictEqual(expected?.sort());
```
EndangeredMassa, line 45: **suggestion (non-blocking):**

```suggestion
          expect(actual.sort()).toStrictEqual(expected.sort());
```

It looks like `expected` will alway be an array of strings. I don't think we need to use `?.` here.

EndangeredMassa (changes requested):
It looks like CI is failing, but the code looks good to me. I left some suggestions.

agadzik (commented):
I keep getting notified every day to review this 😂 

What's the status of this PR?

Ethan-Arrowood:
it is very old and out of date. closing - at this point if we want this we should reimplement.

PR #8965, opened 2022-11-23
[build-utils] make `lambda` optional in `Prerender` constructor

EndangeredMassa:
It turns out that there is [a logic flow](https://github.com/vercel/api/pull/15237/files#diff-54171844caccedbadf06e6f6f408676d4cd200358b55b62ca6f9818e113aeabbL144) in the build container that will not populate `Prerender#lambda` right away. The type error was being ignored with `@ts-expect-error`.

The `Prerender` class needs to handle this case.

---

This will be used in https://github.com/vercel/api/pull/15237


cb1kenobi (approved):

styfle (approved):
Seems okay but I can’t think of a reason why a Prerender wouldn’t have a Lambda associated

EndangeredMassa:
@styfle I meant to link it. That happens here: https://github.com/vercel/api/pull/15237/files#diff-54171844caccedbadf06e6f6f408676d4cd200358b55b62ca6f9818e113aeabbL144

PR #8984, opened 2022-11-28
[build-utils] expand build result type for framework version

EndangeredMassa:
Augments the `BuildResultV2Typical` type to support `framework` and `frameworkVersion` for use in Richer Deployment Outputs.

I added this to a new `meta` object because it contains data about the result, not really part of the result.

---

Supports a related builder container update.


cb1kenobi (approved):

packages/build-utils/src/types.ts
```diff
@@ -425,6 +425,10 @@ export interface BuildResultV2Typical {
     domain: string;
     value: string;
   }>;
+  meta?: {
```
styfle: I'm not a fan of `meta` here since we already have `meta` passed into the `build()` function so its strange to have a different `meta` in the return type of that function.

Can we move these two props up to the top level instead?

EndangeredMassa: Yeah, that's confusing.

I do like having these values in a separate space because they aren't really part of the build result, they are about the build result. That distinction isn't super important, especially if we do end up using these values for more things later, but that was the intent here in keeping them in `meta`.

I can certainly move them to be top-level. I think that's a reasonable change here.

styfle: Are these just labels? What are the implications if the value of framework changes? Is there a RFC I can read?

cb1kenobi: Instead of `meta`, maybe name it `framework`:

```
framework?: {
  name: string;
  version: string;
};
```

TooTallNate: I think I like Chris' suggestion

styfle: Is it `name` or is it `slug`?

EndangeredMassa: Good question. I writing out a response and realized that in my effort to ensure this change had general application (including BOA use cases), I may have been overreaching the scope of the task.

As currently written in these related PRs, `meta.framework` can be a name (arbitrary text) or a slug. If it happens to be a slug, the deployment inspect page will use it to look up the proper name and logo for the framework. Otherwise, the `meta.framework` value is used as the display name and no logo is shown.

The intent was to allow unsupported frameworks who produce BOA to show their own framework name in the deployment inspect page. However, this is a display-only value. If a deployment `meta.framework` was "rando framework" and the project setting was "Next.js", the deployment inspect page would claim "rando framework", but project settings (like build command) would come from "Next.js" defaults (unless overridden). That's not great.

We could allow this custom framework name to show only when the project settings framework is set to "Other" or some new value like "Custom", but that's a stretch to include here.

You're right that we'd need an RFC to approach the broader uses and impacts of declaring framework here.

---

So, I'm going to cut these PRs back to only use the framework version. The framework name will still come from project settings. Anything beyond that would need to be discussed in an RFC.

It seems weird to have `frameworkVersion` and no identifier for a framework here, but adding one only breaks things. We could add it and not use it, but BOA implementers may assume it does something and ask us why it doesn't, regardless of how it is documented. We could add it and use it to look up framework for display in the deployment inspect page, but then we'd have the same problem as before where the deployment can say one thing and the project settings says another. So, I think it's best to leave it off for now.

---

That leaves us with only one value and a decision about where to put it. I suggest:

```
{
  // ...
  framework: {
    version: string;
  }
}
```

This leaves the `framework` key open for expansion later.

styfle: To be clear, `framework.version` is a label that can be any string?

EndangeredMassa: Yes, it would still allow any string. It will hit the validation in the build container before we use it, though.

styfle (approved):

PR #8986, opened 2022-11-28
[node] Add caching to TS compile per tsconfig.json

styfle:
Since introducing `vc build`, we no longer have isolated builds. This means we can share the cache across builds and improve TS compile across multiple builds.

For example, `api/foo.js` and `api/bar.js` are two different builds but they will likely share the same `tsconfig.json`. So this PR caches the initialization used before running the TS compiler so that it can be shared between builds of the same deployment.

In one customer build, we saw deployment time drop from 14 min to 7 min.

- Related to https://github.com/vercel/customer-issues/issues/925

### DIFF
--- a/packages/node/src/typescript.ts
+++ b/packages/node/src/typescript.ts
@@ -114,6 +114,11 @@ function cachedLookup<T>(fn: (arg: string) => T): (arg: string) => T {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


cb1kenobi (approved):
Nice

PR #8987, opened 2022-11-28
[node] Add caching for TS compile per file

styfle:
As a follow up to #8986, we can reduce the TS compile time even more by caching TS compile per file so we never compile the same file twice.  This brings down the total deployment time of a large app with many `api/*.ts` from 7 min to 5 min.

Note: review this PR with [whitespace disabled](https://github.com/vercel/vercel/pull/8987/files?w=1).

- Related to https://github.com/vercel/customer-issues/issues/925


packages/node/src/typescript.ts
```diff
@@ -189,37 +189,54 @@ export function register(opts: Options = {}): Register {
...
+    configFileName = '',
+    skipTypeCheck?: boolean
+  ): GetOutputFunction {
+    let getOutput: GetOutputFunction;
+    let outFiles = new Map<string, SourceOutput>();
+    const cache = configFileToBuildMap.get(configFileName);
+
+    if (cache) {
+      getOutput = cache.getOutput;
+      outFiles = cache.outFiles;
```
cb1kenobi:
```suggestion
      return cache.getOutput;
```

It doesn't appear that `outFiles` is actually being used after this. The `else` is getting pretty large, so it would be easier for me to read if we could return sooner.

cb1kenobi (changes requested):

packages/node/src/typescript.ts
styfle: @cb1kenobi It is used several times below. Take a look with whitespace removed https://github.com/vercel/vercel/pull/8987/files?w=1

EndangeredMassa: I agree with @cb1kenobi here. I think that the `else` is unnecessary and adds makes the resulting code (not just the diff) a bit harder to read.

cb1kenobi: @styfle I opened this file up in VSCode so I could collapse the code and it doesn't appear that `outFiles` is being referenced in a meaningful way. I guess I don't understand where it's being used.

styfle: @cb1kenobi @EndangeredMassa You both are correct. Since there is a closure, there is no need to cache the `outFiles`. I just pushed [a2faf98](https://github.com/vercel/vercel/pull/8987/commits/a2faf9865e425047ec4f1182536867a8e2650d56) with your suggestion 👍

cb1kenobi: Change looks good! Now we just need to make the tests happy.

styfle: @cb1kenobi The failing tests seem unrelated to my PR since I only changed `@vercel/node` and those are passing

cb1kenobi (approved):
LGTM

PR #8990, opened 2022-11-30
edge: format MB with sepace separator

Kikobeats:
The Edge Function size limit is declared in KiB, not in KB:

https://github.com/vercel/vercel/blob/d7654e2252660f6e3fd6f1de375ab8237dc6e178/packages/next/src/edge-function-source/constants.ts#L8

And `pretty-bytes` doesn't support base-2 as input:

```js
require('pretty-bytes')(1024 * 1024* 4) // => '4.19 MB' :(
```

Instead, `bytes` is used:

```js
require('bytes')(1024 * 1024* 4, { unitSeparator: ' ' }) // => '4 MB' 🙂
```

### DIFF
--- a/packages/next/src/edge-function-source/get-edge-function-source.ts
+++ b/packages/next/src/edge-function-source/get-edge-function-source.ts
@@ -6,7 +6,7 @@ import { join } from 'path';
 import { EDGE_FUNCTION_SIZE_LIMIT } from './constants';
 import zlib from 'zlib';
 import { promisify } from 'util';
-import { prettyKibibytes } from '../utils';
+import { prettyBytes } from '../utils';
 
 // @ts-expect-error this is a prebuilt file, based on `../../scripts/build-edge-function-template.js`
 import template from '../../dist/___get-nextjs-edge-function.js';
@@ -85,9 +85,9 @@ async function validateSize(script: string, wasmFiles: string[]) {
   const gzipped = await gzip(content);
   if (gzipped.length > EDGE_FUNCTION_SIZE_LIMIT) {
     throw new Error(
-      `Exceeds maximum edge function size: ${prettyKibibytes(
+      `Exceeds maximum edge function size: ${prettyBytes(
         gzipped.length
-      )} / ${prettyKibibytes(EDGE_FUNCTION_SIZE_LIMIT)}`
+      )} / ${prettyBytes(EDGE_FUNCTION_SIZE_LIMIT)}`
     );
   }
 }
--- a/packages/next/src/edge-function-source/get-edge-function-source.ts
+++ b/packages/next/src/edge-function-source/get-edge-function-source.ts
@@ -6,7 +6,7 @@ import { join } from 'path';
 import { EDGE_FUNCTION_SIZE_LIMIT } from './constants';
 import zlib from 'zlib';
 import { promisify } from 'util';
-import bytes from 'pretty-bytes';
+import { prettyKibibytes } from '../utils';
 
 // @ts-expect-error this is a prebuilt file, based on `../../scripts/build-edge-function-template.js`
 import template from '../../dist/___get-nextjs-edge-function.js';
@@ -44,7 +44,9 @@ export async function getNextjsEdgeFunctionSource(
    * We validate at this point because we want to verify against user code.
    * It should not count the Worker wrapper nor the Next.js wrapper.
    */
-  const wasmFiles = (wasm ?? []).map(({ filePath }) => join(outputDir, filePath));
+  const wasmFiles = (wasm ?? []).map(({ filePath }) =>
+    join(outputDir, filePath)
+  );
   await validateSize(text, wasmFiles);
 
   // Wrap to fake module.exports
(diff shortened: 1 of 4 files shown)


packages/next/src/utils.ts
```diff
@@ -39,9 +39,12 @@ import { getNextjsEdgeFunctionSource } from './edge-function-source/get-edge-fun
 import type { LambdaOptionsWithFiles } from '@vercel/build-utils/dist/lambda';
 import { stringifySourceMap } from './sourcemapped';
 import type { RawSourceMap } from 'source-map';
+import bytes from 'bytes';
 
 type stringMap = { [key: string]: string };
 
+export const prettyKibibytes = (n: number) => bytes(n, { unitSeparator: ' ' });
```
feugy: **suggestion(non blocking):** can we name this `prettyBytes` so it is consistent with this [PR](https://github.com/vercel/edge-functions/pull/419/files#diff-71a2dfec0f0613059ce4d092de5620a4982dc9d4c9edeee3358d981a74836138R9) on `edge-functions`?

feugy (approved):

packages/next/src/edge-function-source/get-edge-function-source.ts
```diff
@@ -6,7 +6,7 @@ import { join } from 'path';
 import { EDGE_FUNCTION_SIZE_LIMIT } from './constants';
 import zlib from 'zlib';
 import { promisify } from 'util';
-import bytes from 'pretty-bytes';
+import { prettyByte } from '../utils';
```
styfle:
```suggestion
import { prettyBytes } from '../utils';
```

packages/next/src/edge-function-source/get-edge-function-source.ts
```diff
@@ -83,9 +85,9 @@ async function validateSize(script: string, wasmFiles: string[]) {
   const gzipped = await gzip(content);
   if (gzipped.length > EDGE_FUNCTION_SIZE_LIMIT) {
     throw new Error(
-      `Exceeds maximum edge function size: ${bytes(
+      `Exceeds maximum edge function size: ${prettyByte(
```
styfle:
```suggestion
      `Exceeds maximum edge function size: ${prettyBytes(
```

packages/next/src/edge-function-source/get-edge-function-source.ts
```diff
@@ -83,9 +85,9 @@ async function validateSize(script: string, wasmFiles: string[]) {
...
         gzipped.length
-      )} / ${bytes(EDGE_FUNCTION_SIZE_LIMIT)}`
+      )} / ${prettyByte(EDGE_FUNCTION_SIZE_LIMIT)}`
```
styfle:
```suggestion
      )} / ${prettyBytes(EDGE_FUNCTION_SIZE_LIMIT)}`
```

packages/next/src/utils.ts
```diff
@@ -39,9 +39,12 @@ import { getNextjsEdgeFunctionSource } from './edge-function-source/get-edge-fun
 import type { LambdaOptionsWithFiles } from '@vercel/build-utils/dist/lambda';
 import { stringifySourceMap } from './sourcemapped';
 import type { RawSourceMap } from 'source-map';
+import bytes from 'bytes';
 
 type stringMap = { [key: string]: string };
 
+export const prettyByte = (n: number) => bytes(n, { unitSeparator: ' ' });
```
```suggestion
export const prettyBytes = (n: number) => bytes(n, { unitSeparator: ' ' });
```

styfle (approved):
Missing the "s"

styfle (changes requested):
Line 64 of package.json should remove `pretty-bytes` and I think there are usages instances that need to be updated to the new function assuming Lambda uses the same base 2 instead of base 10

Kikobeats:
@styfle The `pretty-bytes` dependency is correctly used for AWS Lambda constants

styfle (approved):

packages/next/src/utils.ts
```diff
@@ -39,9 +39,13 @@ import { getNextjsEdgeFunctionSource } from './edge-function-source/get-edge-fun
 import type { LambdaOptionsWithFiles } from '@vercel/build-utils/dist/lambda';
 import { stringifySourceMap } from './sourcemapped';
 import type { RawSourceMap } from 'source-map';
+import bytes from 'bytes';
 
 type stringMap = { [key: string]: string };
 
+const _prettyBytes = (n: number) => bytes(n, { unitSeparator: ' ' });
+export { _prettyBytes as prettyBytes }
```
cb1kenobi, line 47: This could be simplified by exporting a named function which has the added benefit of showing the function name in the unlikely event of a stack trace. This file also contains several other exported functions that match the following style:

```suggestion
export function prettyBytes(n: number) {
  return bytes(n, { unitSeparator: ' ' });
}
```

Kikobeats: so much better; thank you 🙂

Kikobeats: unfortunately, the change is causing a conflict ☹️

CleanShot 2022-11-30 at 17 10 18@2x

cb1kenobi (approved):
Looks great!

PR #8991, opened 2022-11-30
[next] Use MIB for Lambda constants

Kikobeats:
According to AWS Lambda official documentation:

> The Lambda documentation, log messages, and console use the abbreviation MB (rather than MiB) to refer to 1024 KB.

https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html

This is also a convenient thing for us since CF Limits are in base-2 (even they are printed as base-10).


styfle (commented):
Looks like CI is failing because the error message says

> Max serverless function size of 47.68 MB compressed or 250 MB uncompressed

cb1kenobi:
Should these https://github.com/vercel/vercel/blob/7fbb3d7e993d24cde55bc26f4b4f9a01643529c2/packages/next/src/utils.ts#L1545-L1547 also use `MIB`?

packages/next/src/utils.ts
```diff
@@ -1501,7 +1502,7 @@ export const outputFunctionFileSizeInfo = (
     .forEach(depKey => {
       const dep = dependencies[depKey];
 
-      if (dep.compressed < 100 * 1000 && dep.uncompressed < 500 * 1000) {
+      if (dep.compressed < 100 * MIB && dep.uncompressed < 500 * MIB) {
```
ijjk:
```suggestion
      if (dep.compressed < 100 * KIB && dep.uncompressed < 500 * KIB) {
```

Kikobeats: ouch, good catch

packages/next/src/utils.ts
```diff
@@ -1296,9 +1297,9 @@ export type LambdaGroup = {
   pseudoLayerUncompressedBytes: number;
 };
 
-export const MAX_UNCOMPRESSED_LAMBDA_SIZE = 250 * 1000 * 1000; // 250MB
-const LAMBDA_RESERVED_UNCOMPRESSED_SIZE = 2.5 * 1000 * 1000; // 2.5MB
-const LAMBDA_RESERVED_COMPRESSED_SIZE = 250 * 1000; // 250KB
+export const MAX_UNCOMPRESSED_LAMBDA_SIZE = 250 * MIB;
+const LAMBDA_RESERVED_UNCOMPRESSED_SIZE = 2.5 * MIB;
+const LAMBDA_RESERVED_COMPRESSED_SIZE = 250 * MIB;
```
cb1kenobi:
```suggestion
const LAMBDA_RESERVED_COMPRESSED_SIZE = 250 * KIB;
```

cb1kenobi (changes requested):

packages/next/src/utils.ts
Kikobeats: oh, thanks!

cb1kenobi (approved):
LGTM!

TooTallNate (approved):

PR #8993, opened 2022-12-01
[tests] add eslintcache

Simon-He95:


styfle (approved):
Thanks!

We might need to cache the results in GH Actions too but that can be a separate PR

cb1kenobi (approved):
This seems like a good idea. I did some testing and it appears to be relatively safe, so let's give it a go.

PR #8999, opened 2022-12-02
[node-bridge] Add `res.waitUntil` for lambdas and wait for handler to finish

karaggeorge:
An attempt to add `waitUntil` to node lambdas and to prevent code execution from getting cut after `res.end()`

**DISCLAIMER**: this is my first time reading into the node/node-bridge packages, and I haven't found a good way to run/test this locally, so I might have missed some edge cases

From reading into node-bridge, I realized the reason why the execution of the handler can get cut-off is because in the bridge, we only wait for the response to be streamed back to end the event, but not for the actual handler to finish.

I tried to add a `waitUntil` helper that gets attached to the response backed by the node-bridge for a specific request by id.
It also adds the handler itself, so it can be awaited before finishing execution.

There's a few areas that I'm unsure about:

### 1

```
const req = request(
        { hostname: '127.0.0.1', port, path, method },
        socket && url && cipher
          ? getStreamResponseCallback({ url, socket, cipher, resolve, reject })
          : getResponseCallback({ isApiGateway, resolve, reject })
      );
```

Here, I'm not sure when each case happens, but reading through it, it seems like if `getStreamResponseCallback` is used, this will work as expected and stream the response back, while waiting to finish the tasks afterwards. But if `getResponseCallback` is used, then it'll still wait for the tasks to finish but it'll also hold the response until those are done.

### 2

In `node-bridge/launcher.js`, in `getListener` there's a 3rd case for exporting an entire Server object, or even having a server listen automatically. In those cases this wrapper won't apply, since it wraps the listener.

I _could_ try to make it work by wrapping the prototype of `ServerResponse` like we do for the `Server`, but I wasn't sure if this is something legacy that we don't need to worry about before I go down that route.

### 3

I _think_ this will automatically work for Next.js as well, since from what I see the lambda just uses this launcher as well, but it'll probably need updated types and actual implementation of a similar functionality in the next.js repo as well, to handle the dev/prod server.

#### Tests

#### Code Review

### DIFF
--- a/packages/node-bridge/bridge.js
+++ b/packages/node-bridge/bridge.js
@@ -140,10 +140,17 @@ class Bridge {
     this.shouldStoreEvents = shouldStoreEvents;
     this.launcher = this.launcher.bind(this);
     this.reqIdSeed = 1;
+    this.waitUntilIdSeed = 1;
     /**
      * @type {{ [key: string]: import('./types').VercelProxyRequest }}
      */
     this.events = {};
+    /**
+     * Each request will have one promise for the handler and any extra promises that were added using `res.waitUntil`
+     *
+     * @type {{ [key: string]: { handler?: Promise<void>, extra: Promise<any>[] } }}
+     */
+    this.waitUntilPromises = {};
 
     this.listening = new Promise(resolve => {
       this.resolveListening = resolve;
@@ -203,6 +210,25 @@ class Bridge {
         resolveListening(addr);
       }
     );
+  }
+
+  /**
+   * @param {string} waitUntilId
+   * @param {Promise<any>} promise
+   * @param {boolean} isHandler
+   */
+  waitUntil(waitUntilId, promise, isHandler) {
+    if (!this.waitUntilPromises[waitUntilId]) {
+      this.waitUntilPromises[waitUntilId] = {
+        extra: [],
+      };
+    }
+
+    if (isHandler) {
+      this.waitUntilPromises[waitUntilId].handler = promise;
+    } else {
+      this.waitUntilPromises[waitUntilId].extra.push(promise);
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/node-bridge/launcher.js
```diff
@@ -87,19 +87,29 @@ function getVercelLauncher({
...
           Server.prototype.listen = originalListen;
+
           if (shouldAddHelpers) {
             bridge.setStoreEvents(true);
-            import(helpersPath).then(helper => {
-              const h = helper.default || helper;
-              const server = h.createServerWithHelpers(listener, bridge);
+          }
+
+          import(helpersPath).then(helper => {
```
karaggeorge, line 95: I added the new wrapper in `helpers.ts` mostly to add type-safety, but this means we have to import this in both cases of `shouldAddHelpers`. I could move that function in this file to avoid having to do that

packages/node-bridge/bridge.js
```diff
@@ -205,6 +212,25 @@ class Bridge {
     );
   }
 
+  /**
+   * @param {string} waitUntilId
+   * @param {Promise<any>} promise
+   * @param {boolean} isHandler
+   */
+  waitUntil(waitUntilId, promise, isHandler) {
```
ijjk, line 220: I wonder if this should be allowed in non-streaming mode as it may cause confusion why the response is delayed 🤔

ijjk (commented):
Is there a new thread related to this? The last one that we discussed this in was here https://vercel.slack.com/archives/C0457JK07KP/p1665720487686199 not sure if things have change since then.

karaggeorge:
@ijjk had no idea about that thread/project, was just something I've been needing more recently, and decided to try implementing for DYF

packages/node-bridge/bridge.js
karaggeorge: Yeah, I noticed that too. What are the parameters that determine which scenario happens? Streaming vs not? It wasn't clear in the code

ijjk: It's only streaming when `responseCallbackUrl` is present currently.

EndangeredMassa:
Revamped in: https://github.com/vercel/vercel/pull/9745

PR #9007, opened 2022-12-05
[examples] Restrict node version for umi example

ofhouse:
The umi example currently uses `umi@2` which uses `webpack@4` under the hood to build.
Because of this issue https://github.com/webpack/webpack/issues/14532 Webpack@4 is no longer able to run on Node.js 17+.
Since we use [Node.js 18 as default](https://vercel.com/changelog/node-js-18-lts-is-now-available) for new projects, this example currently breaks when it gets deployed.

Temporary fix is here to ensure that a Node.js version below 17 is used in the meantime.


lucleray (approved):
Thanks! I successfully tested the fix, the deployment was successful with the updated `engines`: https://vercel.com/lucleray/umijs2/31uEGx53iF3yRYYkyBdyVU7BRE34 🎉

EndangeredMassa (approved):

leerob (approved):

PR #9009, opened 2022-12-05
[cli][frameworks][fs-detectors][next] detect framework versions

EndangeredMassa:
This PR:

- updates `packages/frameworks` to have most supported frameworks specify which dependency version should reflect the overall framework version
- updates `packages/fs-detectors` to allow framework detection that returns the full `Framework` record instead of just the slug
- updates `packages/next` to return the detected Next.js version in the build result
- updates `packages/cli` to leverage these changes so that `vc build` can add `framework: { version: string; }` to `config.json` output

The result is that Build Output API and supported frameworks will return their framework version in the build result of `vc build` when possible, which is used by the build container  when creating the deployment. The dashboard later retrieves this value to display in richer deployment outputs.

Supports:

- https://github.com/vercel/api/pull/15601
- https://github.com/vercel/front/pull/18319

---

With the related build container updates, we get to see Next.js version in the build output. You'll see this with BOA+Prebuilt or a normal deploy:

Screen Shot 2022-12-09 at 2 48 12 PM

---

### The Path to this PR

I went through all the supported frameworks and figured out how to best determine their versions. For most of them, we can check a known dependency's installed version number. 

We can get most of the way only checking npm. For a handful, we'd have to support Go/Ruby/Rust/Whatever dependencies.

I started with a more complex method signature to allow for later expansion without changing the signature. It looked like this, in practice:

```
async getVersion(dependencies: DependencyMap) => depedencies['next']
```

However, after checking all currently supported frameworks, I don't think this will end up being necessary. It also has the constraint that all dependencies have to be gathered and presented to the function even though it only needs to check for one or two. That's not a huge deal if we have them already where we need them, but we don't. We could use a variant here where this function does its own lookups, but this seemed unnecessary and would beg for duplication and small variances that could cause bugs.

Further, if we only look at `package.json`, we're going to either see a specific version of a version range. To be precise, we have to look at the installed version of the package. That means checking one of the various types of lockfiles that can exist or poking into node_modules.

If we poke into node_modules to detect the installed version, we introduce another point where Yarn 3 (default mode) will not be supported. If we read lockfiles, we have to potentially parse `npm`, `pnpm`, and `yarn` lockfiles.

If we use `npm ls <package-name>`, that also fails in Yarn 3 (default mode). We could accept that and go forward anyway, which would look like:

```
const args = `ls ${packageName} --depth=0 --json`.split(' ');
const { stdout } = await execa('npm', args, { cwd });
const regex = new RegExp(String.raw`${packageName}@([\.\d]+)`);
const matches = stdout.match(regex);
if (matches) {
  return matches[1];
}
```

But it turns out there's a `--json` option! That's what I ended up using, for now.

We could explore the lockfile route more, but after some initial digging, it' non-trivial. There are 3 main lockfiles we'd want to check for (npm, pnpm, and yarn) and there are different lockfile versions that put necessary data in different places. I looked for existing tools that parse this, but I didn't find any. We could certainly go down this path, but the effort doesn't seem worth it when `npm ls` gets us really close.

---

### Follow-up Versioning

Now that we know how to determine version per framework, we can vary configuration by version. In a future PR, we could allow a given value to vary by version number:

```
name: (version) => {
  if (semver.gt(version, '9.8.7')) {
    return 'some-framework-2''
  }

  return 'some-framework';
}
```

However, it may still be easier to differentiate significant versions by adding multiple entries in the list.


packages/cli/src/commands/build.ts
```diff
@@ -538,8 +561,8 @@ async function doBuild(
 
   // Merge existing `config.json` file into the one that will be produced
   const configPath = join(outputDir, 'config.json');
-  // TODO: properly type
```
EndangeredMassa, line 541: I resolved this TODO because I was working with this anyway.

The type `BuildOutputConfig` is duplicated in the api repo, though. Should we move this to a central place, like `built-utils`?

packages/cli/src/commands/build.ts
```diff
@@ -434,7 +457,7 @@ async function doBuild(
   // Execute Builders for detected entrypoints
   // TODO: parallelize builds (except for frontend)
   const sortedBuilders = sortBuilders(builds);
-  const buildResults: Map<Builder, BuildResult> = new Map();
+  const buildResults: Map<Builder, BuildResult | BuildOutputConfig> = new Map();
```
EndangeredMassa, line 458: Because we add a `BuildOutputConfig` object to the `buildResults` map later in this code, we have to expand the allowed types here.

This had a couple of knock-on effects on related functions, but the type overlaps work out just fine.

packages/cli/src/commands/build.ts
```diff
@@ -557,6 +580,21 @@ async function doBuild(
...
+    fs: new LocalFileSystemDetector(cwd),
+    frameworkList: frameworks,
+  });
+  for (const [build, buildResult] of buildResults.entries()) {
+    if (
+      'framework' in buildResult &&
+      build.use === detectedFramework?.useRuntime?.use
+    ) {
+      framework = buildResult.framework;
+      break;
```
EndangeredMassa: I think `break` can be a code smell, but because I wanted the key and value in this loop, this structure seemed the most straight-forward. I'm open to suggestions, though.

packages/cli/src/commands/build.ts
```diff
@@ -585,15 +623,20 @@ async function doBuild(
   const mergedOverrides: Record<string, PathOverride> =
     overrides.length > 0 ? Object.assign({}, ...overrides) : undefined;
 
+  if (!framework) {
+    // if no `buildResult` identified a framework version, try to determine it here
+    framework = await determineFrameworkVersion(detectedFramework, cwd, output);
+  }
```
EndangeredMassa: Builders can return a framework version, but if they don't, we run framework detection to find the framework record, then leverage the new `versionDependencies` key to look up installed framework version.

packages/cli/src/commands/build.ts
```diff
@@ -608,6 +651,59 @@ async function doBuild(
...
+        `Could not parse JSON output of \`npm ls\` for ${packageName}${errorMessage}`
+      );
+      return;
+    }
+
+    return lsOutput.dependencies?.[packageName]?.version;
+  } catch (error) {
+    output.debug(`Error during lookupInstalledVersion: ${error}`);
+  }
+}
```
EndangeredMassa: This currently only works when `npm ls` can retrieve the installed package, which works in npm, pnpm, and yarn 1 projects, at least.

I want to follow-up with some logic that will use the proper package manage to run `${packageManager} ls --depth=0` (and `--json` where supported), then parse out the result. This will be less brittle.

packages/cli/src/util/read-json.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (!jsonRaw) {
+    return;
+  }
+
+  try {
+    return JSON.parse(jsonRaw);
+  } catch (error: any) {
+    return new CantParseJSON(error.message, jsonRaw);
+  }
+}
```
EndangeredMassa: This new function was added because the existing `readJSONFile` reads an actual file.

We could make that function use this one, but the errors thrown are different. It didn't seem worth the minimal re-use.

packages/frameworks/src/frameworks.ts
```diff
@@ -51,6 +51,7 @@ export const frameworks = [
       },
     },
     getOutputDirName: async () => 'public',
+    versionDependencies: ['blitz'],
```
EndangeredMassa: The new `versionDependencies` key is different from `dependency` because the latter is mostly unused and intended to be used to find the CLI dependency, not necessarily the framework dependency.

It is sometimes the case that CLI and framework version match, but it's not a guarantee.

For this PR, we want the framework version.

---

I did look briefly into removing `dependency` first, but it wasn't clear how exactly to replace the the behavior where it's currently used. I'd be happy to pair on that to achieve a clear direction on removing it, though.

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -90,3 +95,18 @@ export async function detectFramework({
...
+  const result = await Promise.all(
+    frameworkList.map(async frameworkMatch => {
+      if (await matches(fs, frameworkMatch)) {
+        return frameworkMatch;
+      }
+      return null;
+    })
+  );
+  return result.find(res => res !== null) ?? null;
+}
```
EndangeredMassa, line 219: We could probably make `detectFramework` use `detectFrameworkRecord`, but I was having trouble reconciling the need for different types between them. It's `BaseFramework` vs. `Framework`.

packages/frameworks/src/frameworks.ts
```diff
@@ -1446,6 +1473,7 @@ export const frameworks = [
         dest: '/200.html',
       },
     ],
+    versionDependencies: ['nuxt3', 'nuxt3-edge'],
```
styfle: Can we reuse the the existing `detectors` logic to get the matching version instead of this new property?

packages/cli/src/commands/build.ts
```diff
@@ -608,6 +651,59 @@ async function doBuild(
   );
 }
 
+export async function determineFrameworkVersion(
+  frameworkRecord: UnboxPromise<ReturnType<typeof detectFrameworkRecord>>,
```
```suggestion
  frameworkRecord: Framework,
```

Can we simplify the type here?

packages/cli/src/commands/build.ts
```diff
@@ -608,6 +651,59 @@ async function doBuild(
...
+}
+
+async function lookupInstalledVersion(
+  cwd: string,
+  packageName: string,
+  output: Output
+) {
+  try {
+    const args = `ls ${packageName} --depth=0 --json`.split(' ');
+    const { stdout } = await execa('npm', args, { cwd });
```
styfle: How about reading package.json instead?

```js
return require(`${packageName}/package.json`).version
```

packages/cli/src/util/errors-ts.ts
```diff
@@ -679,6 +679,23 @@ export class CantParseJSONFile extends NowError<
   }
 }
 
+export class CantParseJSON extends NowError<
```
styfle, line 682: Can we reuse `CantParseJSONFile` instead of making a new error?

packages/cli/src/commands/build.ts
```diff
@@ -608,6 +651,59 @@ async function doBuild(
   );
 }
 
+export async function determineFrameworkVersion(
```
styfle: Should this be `detectFrameworkVersion` to match the other existing function names?

packages/cli/src/commands/build.ts
styfle: Since you already have `Array.from(buildResults.entries())` above, you can store that as a new variable and try this:

```js
const buildResultsArray = Array.from(
  buildResults.entries()
)

const framework = buildResultsArray.find(([build]) => build.use === detectedFramework?.useRuntime?.use)?[1]?.framework
```

packages/frameworks/src/frameworks.ts
styfle: We shouldn't touch `dependency` since its legacy. New frameworks shouldn't add that property because it was used before project linking existed.

packages/fs-detectors/src/detect-framework.ts
styfle: I'm thinking we can completely drop `detectFramework()` in favor of `detectFrameworkRecord` but that can be a future PR since it will be a major semver.

Might be nice to add `// TODO` for this so the next person reading knows our intentions 😃

packages/frameworks/src/frameworks.ts
EndangeredMassa: I did look at using `detectors`, but as written currently they do not identify an actual package we can use to get the version. We could replace them all to use packages only, but I'm concerned that we'd break detection with such a big change. There may be specific frameworks where we really need to rely on that config file detection--not sure. We could changes here incrementally, though.

Or, were you thinking of some other approach?

packages/cli/src/commands/build.ts
EndangeredMassa: Huh, yeah. I never reach for dynamic require as a tool, but I should definitely consider it more often. This would work regardless of package manager, which would be great.

packages/cli/src/commands/build.ts
EndangeredMassa: We'd have to add in a type check as well to differentiate `BuildResultV2Typical` from `BuildResultBuildOutput` in `BuildResultV2` to appease typescript for that.

I think that may be more confusing to read than the PR as written.

packages/cli/src/commands/build.ts
EndangeredMassa: Testing out [a `require` strategy`](https://github.com/vercel/vercel/pull/9009#discussion_r1045090544) instead.

packages/cli/src/commands/build.ts
EndangeredMassa: Updated in [`fbf562e` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/fbf562eb51e1fb296ecbde2e4fdab807bb069fad).

This does seem to work. I [tested it in staging](https://github.com/vercel/api/pull/15601#issuecomment-1347192835).

I could not write a test for this, though. I would always get package not found. I tried manipulating `module.paths`, setting `NODE_PATH`, and using `process.chdir`, but nothing worked for me.

packages/cli/src/commands/build.ts
EndangeredMassa: Updated in https://github.com/vercel/vercel/pull/9009#:~:text=Resolve%20conversation-,PR%20feedback,-fbf562e

packages/cli/src/commands/build.ts
EndangeredMassa: Updated in [`fbf562e` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/fbf562eb51e1fb296ecbde2e4fdab807bb069fad)

packages/fs-detectors/src/detect-framework.ts
EndangeredMassa: Updated in [`fbf562e` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/fbf562eb51e1fb296ecbde2e4fdab807bb069fad)

packages/cli/src/commands/build.ts
EndangeredMassa: Updated in [`fbf562e` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/fbf562eb51e1fb296ecbde2e4fdab807bb069fad)

EndangeredMassa:
working on the weird windows-only test failure

packages/cli/src/commands/build.ts
styfle: That number is the same as this PR 😄 

Is that a copy/pasta error?

packages/frameworks/src/frameworks.ts
styfle: Can you give an example of one that wouldn't work with `detectors`? I'm not sure I understand the problem.

packages/cli/src/util/create-git-meta.ts
```diff
@@ -40,20 +40,28 @@ export async function createGitMeta(
...
-    getLastCommit(directory).catch(err => {
-      output.debug(
-        `Failed to get last commit. The directory is likely not a Git repo, there are no latest commits, or it is corrupted.\n${err}`
-      );
-      return;
-    }),
-    isDirty(directory, output),
+  const [commitResult, dirtyResult] = await Promise.allSettled([
+    getLastCommit(directory),
+    isDirty(directory),
```
EndangeredMassa: These changes are being extracted into a separate PR: https://github.com/vercel/vercel/pull/9071

packages/fs-detectors/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/minimatch": "3.0.5",
     "@types/node": "14.18.33",
     "@types/semver": "7.3.10",
-    "@vercel/build-utils": "4.2.0",
+    "@vercel/build-utils": "5.7.0",
```
EndangeredMassa: moving this over to the "fix CI" PR: https://github.com/vercel/vercel/pull/9071

EndangeredMassa:
This PR now depends on: https://github.com/vercel/vercel/pull/9071

packages/cli/src/commands/build.ts
EndangeredMassa: Huh, yeah. This is the right place: [`fbf562e` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/fbf562eb51e1fb296ecbde2e4fdab807bb069fad)

I updated the other instances as well.

packages/frameworks/src/frameworks.ts
EndangeredMassa: The current `blitzjs` detector looks for `path: 'blitz.config.(js|ts)'`. That will identify a given project as a Blitz.js project, but it doesn't give us any hint as to the version number. We'd still need to know that the package is `blitz` in order to lookup the installed version number.

Options:

1. have a separate property that points to packages to find version numbers: this is what the PR currently does
2. update the detectors to look only for packages, then we know where to find the version numbers: not sure if this is possible for all `npm` packages, but definitely would have to have some flexibility for non-`npm` packages like `Hugo`
3. do something else?

I'd be happy to talk this through.

package.json
```diff
@@ -44,11 +44,11 @@
     "publish-canary": "git checkout main && git pull && lerna version prerelease --preid canary --message \"Publish Canary\" --exact",
     "publish-from-github": "./utils/publish.sh",
     "changelog": "node utils/changelog.js",
-    "build": "node utils/gen.js && turbo run build",
+    "build": "node utils/gen.js && turbo run build --force",
```
TooTallNate: Remove?

```suggestion
    "build": "node utils/gen.js && turbo run build",
```

package.json
```diff
@@ -44,11 +44,11 @@
...
     "vercel-build": "yarn build && yarn run pack && cd api && node -r ts-eager/register ./_lib/script/build.ts",
     "pre-commit": "lint-staged",
     "test": "jest --rootDir=\"test\" --testPathPattern=\"\\.test.js\"",
-    "test-unit": "yarn test && node utils/gen.js && turbo run test-unit",
+    "test-unit": "yarn test && node utils/gen.js && turbo run test-unit --force",
```
TooTallNate:
```suggestion
    "test-unit": "yarn test && node utils/gen.js && turbo run test-unit",
```

packages/cli/src/commands/build.ts
TooTallNate: `build-utils` would be fine I think.

EndangeredMassa: I can move this in a follow-up

packages/frameworks/src/frameworks.ts
styfle: I prefer option 3: Add a `frameworkVersion` prop to the `frameworks.ts` which would override the version found in package.json (if any)

That way you only need to add it to blitz, sveltkit, or others with two entries.

EndangeredMassa: @styfle Hrm. I think we're misunderstanding each other. I want the actual, specific version that the project happens to use. We can't know that until we detect it per deployment.

styfle: @EndangeredMassa Ok in that case, I think its fine to ignore the frameworks that don't have a `package.json` listed in `detectors`. They don't need to list the version.

The fewer props in frameworks.ts, the better 😄

EndangeredMassa: To clarify, your suggestion is:

- leverage existing detectors, no new properties
- if a detector hits package.json
    - have it return the version range as well
    - if the version range is a specific version
        - use it
    - else
        - extract the matched package from the detector
        - lookup the installed version for the matched package
        - use it

Is that right?

It looks like 31/41 frameworks reference `package.json` dependencies. This would be reasonably good coverage.

NOTE: Remix, Sanity, and Hydrogen would not get version support given this change because they do not reference `package.json` dependencies in detectors.

styfle: Yes that algorithm sounds right. If we can do framework detection and version detection in the same pass, that would be ideal.

31/41 is a bit lower than expected but still reasonable. I think it is showing a deficiency in our detectors. For example, does remix always have a `remix.config.js`? Perhaps that should also have a dependency check.

TooTallNate (approved):

packages/cli/src/commands/build.ts
```diff
@@ -608,6 +650,33 @@ async function doBuild(
...
+        version,
+      };
+    }
+  }
+}
+
+function lookupInstalledVersion(packageName: string, output: Output) {
+  try {
+    const pkgJson = require(`${packageName}/package.json`) as PackageJson;
+    return pkgJson.version;
```
cb1kenobi: Using `require()` causes the following error to be thrown:

```
Error during lookupInstalledVersion: Error: Cannot find module 'gatsby/package.json'
```

The culprit is `ncc` because it resolves `'gatsby/package.json'` to the module index in the bundle, not the actual file.

There's probably multiple ways to solve this, but here's a one liner that will work:

```suggestion
return spawnSync(process.execPath, [
      '-p',
      `require('${packageName}/package.json').version`,
    ], {
      encoding: 'utf-8',
    }).stdout.trim();
```

cb1kenobi (changes requested):

packages/cli/src/commands/build.ts
TooTallNate: `eval('require')`

cb1kenobi: Good catch! I forgot about that.

EndangeredMassa: Updated in [`057d0ed` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/057d0ed269a14b272266b292c8bc8776496a5593)

packages/frameworks/src/frameworks.ts
EndangeredMassa: Updated in:
- [`6774f99` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/6774f990b472b1e7c4298427084c6025deccd424)
- [`6ca6b39` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/6ca6b397c9e9bd230fb75f46678ef87ba6866486)

I'm going to open a different PR to see if we can update some of the detectors for the frameworks that do not currently reference package names.

EndangeredMassa:
Removed the new property in `frameworks.ts` in favor of leveraging detectors to retrieve version number. The new detector property is called `matchPackage`. Using this will pluck out the version metadata.

```
detectors: {
  some: [
    {
      matchPackage: 'react-scripts',
    },
    {
      matchPackage: 'react-dev-utils',
    },
  ],
},
```

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -90,3 +150,83 @@ export async function detectFramework({
...
+  }
+
+  return;
+}
+
+function lookupInstalledVersion(packageName: string): string | undefined {
+  try {
+    // we can't use "require(`${packageName}/package.json`)" here because
+    // ncc bundling will make it look in the wrong place
+    const script = `eval('require')('${packageName}/package.json').version`;
```
Ethan-Arrowood: `eval` worries me. Lets be extra diligent this is not introducing some kind of code injection vulnerability

Ethan-Arrowood (commented):
this is looking good to me. Will need to take another look later

packages/frameworks/src/frameworks.ts
EndangeredMassa: New PR to update some of the detectors to use package names: https://github.com/vercel/vercel/pull/9167

packages/fs-detectors/src/detect-framework.ts
cb1kenobi: `eval()` is necessary to get the global Node.js `require()` and not the webpack version. Webpack's `require()` can only return modules that are bundled. In this case, the code needs `package.json` which is not included in the bundle, thus Node.js' `require()` will easily resolve it.

Ethan-Arrowood: Yeah that makes sense. Just calling it out here that we need to be extra careful this doesn't introduce a vulnerability. I'm not an expert in that field but `eval` always sets off the alarm bells for me.

EndangeredMassa: I believe `eval('required')` is necessary because of some ncc processing.
https://github.com/vercel/vercel/pull/9009#discussion_r1051235678

cb1kenobi: This is actually a bit more interesting. I thought the `spawnSync()` was a hack to get around the bundler and thought it could be removed in favor of the `eval('require')`, but it turns out the `eval('require')('<package>/package.json')` trick doesn't work! You have to spawn a subprocess.

I did some testing and both the bundled output and the Node.js REPL error executing `require('gatsby/package.json')`. However when I run `node -p "require('gatsby/package.json').version"` as you have it, it works! I have no idea why. They shouldn't be any different, right?

Additionally, it doesn't appear that the `eval()` is necessary after all. For the sake of clarity, I vote to remove the `eval()`.

cb1kenobi (changes requested):

EndangeredMassa:
@cb1kenobi I fixed the version lookup and did some more testing. This does seem to work the way we want. Thanks! [`c0d63d8` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/c0d63d8017178a123bf5f4bd29d9fe593bb29ec3)

cb1kenobi (approved):
Looks good! Reviewed and tested with Next.js, Redwood, Hugo, Gatsby, and simple Node.js project and all work.

EndangeredMassa:
I'd like to get at least one more approval before merging

packages/fs-detectors/test/unit.framework-detector-record.test.ts
```diff
@@ -0,0 +1,194 @@
...
+          nuxt3: '2.0.0',
+          'nuxt3-edge': '3.0.0',
+        },
+      }),
+    });
+
+    const framework = await detectFrameworkRecord({ fs, frameworkList });
+    expect(framework?.slug).toBe('nuxtjs');
+    expect(framework?.detectedVersion).toBe('1.0.0');
+  });
```
Ethan-Arrowood: When you say "first" does this mean _first_ dependency in `dependencies` or _first_ detector in the frameworks list? This is pretty gnarly edge case since if the user had  all of these actually listed in their package.json - they either have a bug in their project somewhere or shouldn't be relying on our auto detection logic.

packages/fs-detectors/test/unit.framework-detector-record.test.ts
```diff
@@ -0,0 +1,194 @@
...
+        dependencies: {
+          next: '9.0.0',
+          gatsby: '4.18.0',
+        },
+      }),
+    });
+
+    const framework = await detectFrameworkRecord({ fs, frameworkList });
+    expect(framework?.slug).toBe('nextjs');
+  });
```
Ethan-Arrowood, line 83: Another order related concern. Not that it is within scope of this PR, but this makes me think if theres a better way to do this. I'd assume we check in alphabetical order, but I know we intentionally stick `next.js` at the top of the list so that its "first".

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -90,3 +150,75 @@ export async function detectFramework({
...
+    })
+  );
+  const frameworkRecord = result.find(res => res !== null) ?? null;
+
+  return frameworkRecord;
+}
+
+export async function detectFrameworkVersion(
+  frameworkRecord: Framework
+): Promise<string | undefined> {
```
Ethan-Arrowood:
```suggestion
export function detectFrameworkVersion(
  frameworkRecord: Framework
): string | undefined {
```

Don't think this needs to be async anymore

Ethan-Arrowood (changes requested):

packages/fs-detectors/test/unit.framework-detector-record.test.ts
EndangeredMassa: "First" means the first detected version based on the order of the detectors, which I now realize isn't exactly tested by this test setup.

I don't love this either. The alternative is to return no version if we find more than one. I think that would be fine. We could log that it happened, too.

packages/fs-detectors/test/unit.framework-detector-record.test.ts
EndangeredMassa: This is a test that `detectFrameworkRecord` works the same as `detectFramework` did. If we want to change how order works in the framework list, we should discuss that as a team separately, yeah.

packages/fs-detectors/src/detect-framework.ts
EndangeredMassa: Thanks! [`956af74` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/956af74b3f516e463bcf668d477bce56271e0bd4)

packages/fs-detectors/test/unit.framework-detector-record.test.ts
EndangeredMassa: Or we could go back to creating a separate key that identifies the package we should use to pluck version number. Then, the detector order wouldn't matter.

Previous discussion: https://github.com/vercel/vercel/pull/9009#discussion_r1045090088

EndangeredMassa: We could use a team decision on the strategy to take with using detectors to determine version of detected frameworks. I've implemented it different ways and there's not yet a consensus on the path to take.

Options:

- **(1)** create new property on framework entries like `versionDependency: 'nuxt3'` at the top-level of a framework entry
    - **(1a)** allow only a single package to be listed; if you want multiple, create multiple framework entries
    - **(1b)** allow an array of packages, preferring the first matched package to return version
    - **pro:** clearer place to determine version 
    - **con:** could be confusing (no matter what we name it) next to the existing (and mostly deprecated) `dependency` property
- **(2)** leverage detectors that already look at package.json dependencies like `matchPackage: 'nuxt3'` in a detector
    - **(2a)** if multiple detectors match, use the first version number
    - **(2b)** if multiple detectors match, use none of the version numbers
    - **(2c)** if multiple detectors look at package.json dependencies, throw error saying to create multiple framework entries instead
    - **pro:** we already scan package.json dependencies for specific packages; plucking out version can be done in the same action
    - **con:** less clear that this is happening

*Note:* There are cases where our framework config is the same for different packages that could match, such as `nuxt`/`nuxt3`/`nuxt-edge`.

Having implemented both options, I think option **(1b)** offers the best balance of flexibility and clarity.

styfle: I would vote for `(2a) if multiple detectors match, use the first version number` and put nuxt3 first

Ethan-Arrowood: Agree with @styfle - this seems like a unique enough edge case that our users probably wont run into, however if there is a legitimate scenario we at least can be deterministic

cb1kenobi: I like option 2a. I'd like to see the order of the detectors be grouped by framework, but within the group ordered by specificity.

I also like the idea of `matchPackage` and `matchContent` be a `string | RegExp` just in case we have a use case come up where the current mechanism isn't opinionated enough.

EndangeredMassa: `matchPackage` is used to build the regex. So, it could include regex syntax. That's not clear from the type, but I'm also not (edit)**sure that**(/edit) passing in a regex to build another regex will be worth making this harder to debug.

cb1kenobi: Right, I'm suggesting build the regex if `matchPackage` is a string and not already a regex.

EndangeredMassa:
I'll flesh out more test coverage of the `frameworks.ts` detectors in: https://github.com/vercel/vercel/pull/9167

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -90,3 +150,75 @@ export async function detectFramework({
...
+
+  const version = lookupInstalledVersion(
+    process.execPath,
+    firstMatchPackage.matchPackage
+  );
+  if (version) {
+    return version;
+  }
+
+  return;
```
Ethan-Arrowood:
```suggestion
  const version = lookupInstalledVersion(
    process.execPath,
    firstMatchPackage.matchPackage
  );
  
  return version;
```

Suggestion: This should be fine unless `version` can be returned as some kinda falsy value and you want to explicitly return `undefined` _*cough* *cough* explicit conditionals *cough* *cough*_

Ethan-Arrowood (approved):

packages/fs-detectors/src/detect-framework.ts
EndangeredMassa: Done! [`91f922b` (#9009)](https://github.com/vercel/vercel/pull/9009/commits/91f922b1d03f6721da02529a71b473e1b6cb8d1e)

PR #9016, opened 2022-12-06
feat(node,cli): allows 'edge' as possible runtime value

feugy:
### 📖 What's in there?

As part of Edge function GA, we're allowing `edge` as a runtime value.
This is adding to the allowed list, so we stay backward compatible.

[Linear ticket](https://linear.app/vercel/issue/EC-481/add-edge-to-the-list-of-allowed-runtimes)

I've kept one instance of `experimental-edge` in the test fixtures to ensure we still supports it.

####  🧪 Tests

#### Code Review

### DIFF
--- a/packages/node/src/dev-server.ts
+++ b/packages/node/src/dev-server.ts
@@ -74,7 +74,7 @@ import { VercelProxyResponse } from '@vercel/node-bridge/types';
 import { Config } from '@vercel/build-utils';
 import { getConfig } from '@vercel/static-config';
 import { Project } from 'ts-morph';
-import { logError } from './utils';
+import { EdgeRuntimes, isEdgeRuntime, logError } from './utils';
 import { createEdgeEventHandler } from './edge-functions/edge-handler';
 import { createServerlessEventHandler } from './serverless-functions/serverless-handler';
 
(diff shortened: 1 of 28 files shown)


packages/node/src/dev-server.ts
```diff
@@ -86,18 +86,17 @@ function listen(server: Server, port: number, host: string): Promise<void> {
...
-const validRuntimes = ['experimental-edge'];
 function parseRuntime(
   entrypoint: string,
   entryPointPath: string
 ): string | undefined {
   const project = new Project();
   const staticConfig = getConfig(project, entryPointPath);
   const runtime = staticConfig?.runtime;
-  if (runtime && !validRuntimes.includes(runtime)) {
+  if (runtime && !isEdgeRuntime(runtime)) {
```
Kikobeats, line 96:
```suggestion
  if (!isEdgeRuntime(runtime)) {
```

Kikobeats (approved):
just dropping an unnecessary condition, the rest looks fine 🙂

TooTallNate (approved):

packages/node/src/dev-server.ts
Kikobeats: Since `isEdgeRuntime` is actually taking care about the nullish value

feugy: That's a good point...in JS. Typescript complains about `includes()` not allowing `undefined` as a parameter:
image

PR #9028, opened 2022-12-06
[cli] Add `--git-branch` option to `vc pull` and `vc env pull`

TooTallNate:
Allows for the the Git branch name to be specified when pulling Preview environment variables.

```
$ vc pull --branch feat/new-header --environment preview
…
✅  Updated .vercel/.env.preview.local file [236ms]
…

$ cat .vercel/.env.preview.local
# Created by Vercel CLI
A="a"

$ vc pull --branch remove/something --environment preview
…
✅  Updated .vercel/.env.preview.local file [340ms]
…

$ cat .vercel/.env.preview.local
# Created by Vercel CLI
B="b"
```


packages/cli/src/commands/env/pull.ts
```diff
@@ -65,6 +66,15 @@ export default async function pull(
     return 1;
   }
 
+  const gitBranch = opts['--branch'];
+
+  if (gitBranch && environment !== 'preview') {
+    output.error(
+      `The \`--branch\` flag can only be used with \`--environment preview\``
```
styfle: I think we should throw this error in the API instead right?

packages/cli/src/commands/env/pull.ts
```diff
@@ -100,6 +110,7 @@ export default async function pull(
   const records = (
     await pullEnvRecords(output, client, project.id, source, {
       target: environment || ProjectEnvTarget.Development,
+      gitBranch,
```
styfle, line 105: Was the API updated to merge env vars? Previously we needed two API calls:

1. fetch all preview env vars
2. fetch preview env vars for the git branch

Then the client had to merge.

Does the API do the merging now? It might be good to add a test to confirm.

packages/cli/src/commands/env/pull.ts
TooTallNate: API now handles this:

```
 $ vc pull --branch old --environment production
Vercel CLI 28.7.2
> Overwriting existing .vercel/.env.production.local file
> Downloading `production` Environment Variables for Project n8.io
Error: Invalid request: `target` must be "preview" when specifying a `gitBranch`
```

packages/cli/src/commands/env/pull.ts
TooTallNate: Yes it merges.

packages/cli/src/commands/env/pull.ts
styfle: @TooTallNate Lets remove this `if` statement then?

And perhaps change `--branch` to `--gitBranch`

styfle (approved):
Might want to use “—gitBranch” to match the API

cb1kenobi (approved):

Belrestro (approved):
Please merge this change, I was specifically looking for such feature 🥳

EndangeredMassa (approved):
I think `--gitBranch` would make it less consistent with our other options. I prefer `--git-branch` here.

PR #9030, opened 2022-12-07
feat: Script + workflow for upgrading SvelteKit example repo

elliott-with-the-longest-name-on-github:
DO NOT MERGE UNTIL SK LAUNCH. Because the template currently uses the "next" tag, it'll always open a PR.

Adds a script and associated workflow to automatically keep the SvelteKit example up-to-date. This script checks three packages:

- `@sveltejs/kit`
- `@sveltejs/adapter-vercel`
- `create-svelte`

If either of the first two have changed, we can just rerun `create-svelte`. If `create-svelte` has changed, we have to bump its package version (it's installed in the workspace root), re-import it, and then run it. We also have to do a find-and-replace in `/examples/sveltekit/package.json` and `/examples/sveltekit/svelte.config.js` for `adapter-auto` => `adapter-vercel`.

### Related Issues

N/A that I'm aware of -- discussed via Slack with @leerob and @Rich-Harris 

### 📋 Checklist
 
- [x] Add workflow to fire version check logic
- [x] Write version check logic
- [ ] Who should be marked as reviewers? @ijjk and @styfle are, currently, as they were on the Next.js script, but I'd be happy to instead list myself and (potentially, not volunteering them on their behalf), @Rich-Harris and/or @dummdidumm.

#### Tests

#### Code Review

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@typescript-eslint/parser": "5.21.0",
     "async-retry": "1.2.3",
     "buffer-replace": "1.0.0",
+    "create-svelte": "2.0.0-next.198",
     "eslint": "8.14.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-jest": "26.1.5",
--- a/utils/update-sveltekit.js
+++ b/utils/update-sveltekit.js
@@ -66,23 +66,21 @@ module.exports = async ({ github, context }) => {
 
   // replace the default adapter-static with adapter-vercel
   // fs paths are root-relative
-  const packageJsonContents = JSON.parse(
-    fs.readFileSync(PACKAGE_JSON_PATH.slice(1), 'utf8')
-  );
-  const svelteKitConfigContents = fs.readFileSync(
-    SVELTEKIT_CONFIG_PATH.slice(1),
-    'utf8'
-  );
+  const pjResolved = require.resolve(PACKAGE_JSON_PATH);
+  const skResolved = require.resolve(SVELTEKIT_CONFIG_PATH);
+
+  const packageJsonContents = JSON.parse(fs.readFileSync(pjResolved, 'utf8'));
+  const svelteKitConfigContents = fs.readFileSync(skResolved, 'utf8');
 
   delete packageJsonContents.devDependencies['@sveltejs/adapter-static'];
   packageJsonContents.devDependencies['@sveltejs/adapter-vercel'] = newVercel;
 
   fs.writeFileSync(
-    PACKAGE_JSON_PATH.slice(1),
+    pjResolved,
     JSON.stringify(packageJsonContents, null, '\t') + '\n'
   );
   fs.writeFileSync(
-    SVELTEKIT_CONFIG_PATH.slice(1),
+    skResolved,
     svelteKitConfigContents.replace('adapter-auto', 'adapter-vercel')
   );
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "@typescript-eslint/parser": "5.21.0",
     "async-retry": "1.2.3",
     "buffer-replace": "1.0.0",
+    "create-svelte": "2.0.0-next.198",
```
styfle, line 26: Would it be better to remove this exact version and instead run `npm create svelte@latest` like the docs say here?

https://www.npmjs.com/package/create-svelte

That way it would also use the latest version each run of cron?

elliott-with-the-longest-name-on-github: Yes and no -- if you read the script, you'll see that it upgrades `create-svelte` if a newer version is available (and includes that upgrade as part of the PR) -- but the problem is, `create-svelte` is an interactive prompt and doesn't support command-line flags. Instead, it also exports a programmatic interface. I actually asked Rich about this and offered to add support, and some of the other maintainers dug up this: https://github.com/sveltejs/kit/pull/6117#issuecomment-1221323822

Which TBH I actually agree with.

Rich-Harris: Does this mean someone needs to manually bump that version each time?

Since the programmatic API will be subject to semver once we come out of pre-release mode, we could pin it to a major?

utils/update-sveltekit.js
```diff
@@ -0,0 +1,112 @@
...
+  const newCreate = newVersion('create-svelte');
+  const branch = `sveltekit-${`${newKit}-${newVercel}-${newCreate}`.replaceAll(
+    '.',
+    '-'
+  )}`;
+
+  if (oldKit === newKit && oldVercel === newVercel && oldCreate === newCreate) {
+    console.log(`SvelteKit package versions did not change; skipping update.`);
+    return;
+  }
```
Rich-Harris: minor nit but we don't use `branch` until we've potentially exited

```suggestion

  if (oldKit === newKit && oldVercel === newVercel && oldCreate === newCreate) {
    console.log(`SvelteKit package versions did not change; skipping update.`);
    return;
  }

  const branch = `sveltekit-${`${newKit}-${newVercel}-${newCreate}`.replaceAll(
    '.',
    '-'
  )}`;
```

utils/update-sveltekit.js
```diff
@@ -0,0 +1,112 @@
...
+  if (
+    exec('git', ['ls-remote', '--heads', 'origin', branch]).toString().trim()
+  ) {
+    console.log(`Branch ${branch} already exists, skipping update.`);
+    return;
+  }
+
+  if (oldCreate !== newCreate) {
+    exec('yarn', ['upgrade', 'create-svelte']);
+    const module = require('create-svelte');
```
Rich-Harris, line 53: I'm surprised this works — `create-svelte` is ESM. we don't need to `await import(...)`?

utils/update-sveltekit.js
elliott-with-the-longest-name-on-github: ty

package.json
elliott-with-the-longest-name-on-github: No, no manual version-bumping -- it'll upgrade itself prior to running if there's a newer version.

utils/update-sveltekit.js
elliott-with-the-longest-name-on-github: Ah shoot, we might. I'll do some testing and see.

dummdidumm:
Re "PR reviewers" - you can put me on that list

PR #9041, opened 2022-12-07
testing a theory in CI

EndangeredMassa:
DO NOT MERGE

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -12,6 +12,10 @@ try {
   }
 }
 
+// testing kodiakhq
+
+
+
 import { join } from 'path';
 import { existsSync } from 'fs';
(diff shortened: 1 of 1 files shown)


styfle (approved):
Approving to see if kodiak will merge

EndangeredMassa:
Findings: There is a situation where Github can show a Reviewers list with no approvals that KodiakHQ will still merge.

* open a PR
* get an approval
* re-request review from your one approver
* add automerge label
* kodiakhq merges PR

In this scenario, KodiakHQ is merging even though you re-requested a review from your only approver. This is happening because Approval state is separate from Review state in Github PRs. Submitting a review with an approval does both, separate things.

If you want to re-request a review and not allow a KodiakHQ automerge until you get another approval, you can "Dismiss Review" on the approval instead of only re-requesting their review.

PR #9047, opened 2022-12-08
[build-utils] Add meta buildEnv to spawn env

khuezy:
**Related Issues**
I'd like to be able to pass env variables to `build`, eg:

```ts
await nextBuild({
   meta: {
      buildEnv: { NEXT_PRIVATE_STANDLONE: 'true' }
   }
})
```

📋 Checklist

**Tests**

- [ ]  The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with yarn test-unit

Code Review
 
### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -218,7 +218,7 @@ export function getSpawnOptions(
   nodeVersion: NodeVersion
 ): SpawnOptions {
   const opts = {
-    env: cloneEnv(process.env),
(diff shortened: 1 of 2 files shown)


TooTallNate:
I'm not sure I understand. You can just set env vars on `process.env` and the child processes will inherit those env vars.

khuezy:
Hi @TooTallNate Thanks for responding. The use case would be for a library to execute the build process on behalf of the "user". So the user doesn't have to set the `output: "standalone"` in their config or set their environmental variables.

Feel free to close.

khuezy:
I didn't realize it was that simple... thanks!

PR #9056, opened 2022-12-09
[tests] Update to use `actions/cache@v3`

styfle:
Use `actions/cache` directly instead of relying on `actions/setup-node` to see if this solves [the hanging restore](https://github.com/vercel/vercel/actions/runs/3660219547/jobs/6187125554).

- Related to https://github.com/actions/cache/issues/810
- Related to https://github.com/Azure/azure-sdk-for-js/issues/22321

This also shaves a minute off cache restore time since we are caching 250MB instead of multiple GB.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,7 +42,13 @@ jobs:
       timeout-minutes: 5 # See https://github.com/actions/cache/issues/810
       with:
         node-version: 14
-        cache: 'yarn'
+    - name: Cache
+      if: ${{ steps.check-release.outputs.IS_RELEASE == 'true' }}
+      uses: actions/cache@v3
+      with:
+        path: '**/node_modules'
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -42,7 +42,13 @@ jobs:
       timeout-minutes: 5 # See https://github.com/actions/cache/issues/810
       with:
         node-version: 14
-        cache: 'yarn'
+    - name: Cache
```
styfle, line 45:
```suggestion
    - name: Cache // testing rerun with a comment
```

.github/workflows/publish.yml
```diff
@@ -42,7 +42,13 @@ jobs:
       timeout-minutes: 5 # See https://github.com/actions/cache/issues/810
       with:
         node-version: 14
-        cache: 'yarn'
+    - name: Cache // testing rerun with a comment
```
styfle:
```suggestion
    - name: Cache
```

TooTallNate (approved):

PR #9070, opened 2022-12-12
[examples] Upgrade SvelteKit template to 1.0.0

elliott-with-the-longest-name-on-github:
🎉 Kit is live! 🎉 

Merge prior to #9030. Manually updates Kit to latest (will need to have a commit pushed after 1.0 goes live on NPM).


examples/sveltekit/README.md
```diff
@@ -1,30 +1,38 @@
...
+npm run dev -- --open
 ```
 
 ## Building
 
 To create a production version of your app:
 
 ```bash
-pnpm run build
+npm run build
```
leerob: Should we keep using pnpm? 🙏 That's our preference.

examples/sveltekit/playwright.config.js
```diff
@@ -0,0 +1,10 @@
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
```
leerob: Do all templates start with playwright now?

package.json
```diff
@@ -23,6 +23,7 @@
     "@typescript-eslint/parser": "5.21.0",
     "async-retry": "1.2.3",
     "buffer-replace": "1.0.0",
+    "create-svelte": "2.0.0-next.202",
```
leerob: Is this needed?

examples/sveltekit/README.md
elliott-with-the-longest-name-on-github: I would prefer that as well but keep in mind this is going to get overwritten every time the cron job from the other PR runs 😬

package.json
elliott-with-the-longest-name-on-github: It's needed as a prerequisite to the cron PR.

examples/sveltekit/playwright.config.js
elliott-with-the-longest-name-on-github: We could generate it without, but it's the "blessed integration testing framework" from Kit's perspective.

examples/sveltekit/README.md
elliott-with-the-longest-name-on-github: Probably best to leave it -- this README is the one generated by the `create-svelte` CLI, so anything we do to it is going to get overridden, and we can't guarantee it's not going to just change on us whenever.

package.json
styfle: @tcc-sejohnson I think we should used `npx` instead so it always uses the latest.

See this example for updating Next.js https://github.com/vercel/vercel/blob/5120689bf2b6201ace3f2349c0438f451ff13209/utils/update-next.js#L35

styfle (changes requested):
We can't change the commands for an existing framework. It must be a new entry

elliott-with-the-longest-name-on-github:
> We can't change the commands for an existing framework. It must be a new entry

Ahh, that should've been obvious. My bad. I'll add a new one.

package.json
elliott-with-the-longest-name-on-github: @styfle 

We cannot. You can't pass CLI arguments to `npx create svelte@latest`. See https://github.com/vercel/vercel/pull/9030#discussion_r1042641320

The cron script I adapted from the next script does keep it up-to-date, though.

packages/frameworks/src/frameworks.ts
```diff
@@ -989,8 +989,8 @@ export const frameworks = [
     ],
   },
   {
-    name: 'SvelteKit',
-    slug: 'sveltekit',
+    name: 'SvelteKit 1',
+    slug: 'sveltekit-1',
```
styfle: We cannot rename existing framework slugs

styfle (changes requested):
We cannot rename existing framework slugs. It will break existing projects.

Also tests are failing.

examples/sveltekit/README.md
```diff
@@ -1,30 +1,38 @@
...
 ```
 
 You can preview the production build with `npm run preview`.
+
+> To deploy your app, you may need to install an [adapter](https://kit.svelte.dev/docs/adapters) for your target environment.
```
styfle: Is this still the case? I thought we handle it automatically?

elliott-with-the-longest-name-on-github: This is the README generated by `create-svelte` for all Svelte projects. The adapter _is_ installed automatically, but if we want to rewrite the README, we're also going to have to deal with re-overwriting it every time we rerun `create-svelte` for new versions.

package.json
```diff
@@ -33,6 +34,7 @@
     "node-fetch": "2.6.7",
     "npm-package-arg": "6.1.0",
     "prettier": "2.6.2",
+    "prettier-plugin-svelte": "2.9.0",
```
styfle: Whats this for? Can we remove it?

packages/frameworks/src/frameworks.ts
```diff
@@ -1027,6 +1027,45 @@ export const frameworks = [
...
+    slug: 'sveltekit-2',
+    demo: 'https://sveltekit-template.vercel.app',
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/svelte.svg',
+    screenshot:
+      'https://assets.vercel.com/image/upload/v1647366075/front/import/sveltekit.png',
+    tagline:
+      'SvelteKit is a framework for building web applications of all sizes.',
+    description: 'A SvelteKit app optimized Edge-first.',
+    website: 'https://kit.svelte.dev',
+    envPrefix: 'VITE_',
```
styfle: I thought you said the newer version of SveltKit no longer needs a prefix to inject env vars in the client bundle?

packages/frameworks/src/frameworks.ts
```diff
@@ -1027,6 +1027,44 @@ export const frameworks = [
...
+        },
+      ],
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install`, `pnpm install`, or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`pnpm run build``',
+        value: 'pnpm run build',
```
styfle: The `value` should be the command from the package.json

packages/frameworks/src/frameworks.ts
```diff
@@ -1027,6 +1027,44 @@ export const frameworks = [
...
+      },
+      devCommand: {
+        value: 'pnpm run dev',
+        placeholder: '`pnpm run dev`',
```
styfle: Same here for `value`, should be comamdn from package.json and it should include port flag

packages/frameworks/src/frameworks.ts
```diff
@@ -1027,6 +1027,44 @@ export const frameworks = [
...
+    tagline:
+      'SvelteKit is a framework for building web applications of all sizes.',
+    description: 'A SvelteKit app optimized Edge-first.',
+    website: 'https://kit.svelte.dev',
+    detectors: {
+      every: [
+        {
+          path: 'package.json',
+          matchContent:
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"@sveltejs\\/kit":\\s*".*?"[^}]*}',
```
```suggestion
            '"(dev)?(d|D)ependencies":\\s*{[^}]*"@sveltejs\\/kit":\\s*".+?"[^}]*}',
```

styfle (approved):
Thanks!

styfle (changes requested):
This is failing dev

```
> Running Dev Command “svelte-kit dev --port $PORT”

  ERROR
    Invalid command: dev

  Run `$ svelte-kit --help` for more info.
```

styfle (approved):
Oh my mistake, that was using the old project setting. The new one works.

PR #9071, opened 2022-12-12
[tests] fix flakey tests

EndangeredMassa:
Fixes flakey tests / CI:

- git metadata test for corrupted `.git` directory
- version identifier for `build-utils` being using in `fs-detectors`'s `devDependencies`
- bad import from `../dist/..`

### DIFF
--- a/packages/cli/src/util/create-git-meta.ts
+++ b/packages/cli/src/util/create-git-meta.ts
@@ -99,7 +99,7 @@ export function isDirty(directory: string): Promise<boolean> {
         if (err) {
           return reject(err);
         }
-        if (stderr?.trim()) {
+        if (stderr !== undefined && stderr.trim().length > 0) {
           return reject(new Error(stderr));
         }
 
--- a/packages/cli/src/util/create-git-meta.ts
+++ b/packages/cli/src/util/create-git-meta.ts
@@ -5,7 +5,7 @@ import git from 'git-last-commit';
 import { exec } from 'child_process';
 import { GitMetadata, Project } from '../types';
 import { Output } from './output';
-import { errorToString } from '@vercel/error-utils';
+import { errorToString, normalizeError } from '@vercel/error-utils';
 
 export async function createGitMeta(
   directory: string,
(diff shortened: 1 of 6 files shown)


packages/cli/src/util/create-git-meta.ts
```diff
@@ -40,20 +40,28 @@ export async function createGitMeta(
...
-      output.debug(
-        `Failed to get last commit. The directory is likely not a Git repo, there are no latest commits, or it is corrupted.\n${err}`
-      );
-      return;
-    }),
-    isDirty(directory, output),
+  const [commitResult, dirtyResult] = await Promise.allSettled([
+    getLastCommit(directory),
+    isDirty(directory),
   ]);
```
EndangeredMassa, line 46: This change was made to make the log output of these two tasks deterministic.

packages/cli/src/util/create-git-meta.ts
```diff
@@ -68,31 +76,35 @@ function getLastCommit(directory: string): Promise<git.Commit> {
   return new Promise((resolve, reject) => {
     git.getLastCommit(
       (err, commit) => {
-        if (err) return reject(err);
+        if (err) {
+          return reject(normalizeError(err));
```
EndangeredMassa, line 80: `normalizeError` is necessary here because the `git-last-commit` package will throw a `string` instead of an `Error` in some cases.

I submitted [a small PR](https://github.com/seymen/git-last-commit/pull/32) to fix this in that library.

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -277,13 +277,6 @@ describe('createGitMeta', () => {
         `Failed to get last commit. The directory is likely not a Git repo, there are no latest commits, or it is corrupted.`
       );
 
-      // skip next line
-      await lines.next();
-
-      line = await lines.next();
-      expect(line.value).toContain(
-        `Failed to determine if Git repo has been modified:`
-      );
```
EndangeredMassa, line 286: This output doesn't happen anymore. It seems better to only render out the most relevant error if both of these errors occur.

It's not clear how to reproduce this second issue without triggering the first. If anyone can think of one, I'll add another test for it.

styfle (dismissed):
Thanks!

TooTallNate (dismissed):

cb1kenobi (dismissed):

packages/cli/src/util/create-git-meta.ts
```diff
@@ -68,31 +76,35 @@ function getLastCommit(directory: string): Promise<git.Commit> {
...
-          if (err) debugMessage += `\n${err}`;
-          if (stderr) debugMessage += `\n${stderr.trim()}`;
-          output.debug(debugMessage);
-          return resolve(false);
+        if (err) {
+          return reject(err);
         }
+        if (stderr?.trim()) {
+          return reject(new Error(stderr));
+        }
```
Ethan-Arrowood: Nit: I think the `?.` lets the conditional work as you expect it too, but I much prefer explicit conditionals.
```suggestion
        if (stderr !== undefined) {
          return reject(new Error(stderr.trim()));
        }
```

Ethan-Arrowood (dismissed):

packages/cli/src/util/create-git-meta.ts
EndangeredMassa: The `trim` in the version as written in the PR also allows whitespace-only values to not reject. We could make this a more explicit check, though, certainly.

Updated: [`abbce0f` (#9071)](https://github.com/vercel/vercel/pull/9071/commits/abbce0f48d3fb46b0bb10415e59e495dbf8ba333)

EndangeredMassa:
Windows Unit Tests have been running into the timeout. We could look into making them take less time, but to mitigate right now, I think we should bump this timeout up.

[`0742d77` (#9071)](https://github.com/vercel/vercel/pull/9071/commits/0742d77d3e4129f9466bdbbf51ae046159b43bcb)

styfle (approved):

PR #9077, opened 2022-12-13
[frameworks] Add quotes to match vitepress dep

cb1kenobi:
Added quotes to get the `vitepress` detector to correctly detect the dependency.

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1732,7 +1732,7 @@ export const frameworks = [
         {
           path: 'package.json',
           matchContent:
-            '"(dev)?(d|D)ependencies":\\s*{[^}]*vitepress:\\s*".+?"[^}]*}',
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"vitepress":\\s*".+?"[^}]*}',
         },
       ],
     },


EndangeredMassa:
Tests for this are in `fs-detectors`: https://github.com/vercel/vercel/blob/main/packages/fs-detectors/test/unit.framework-detector.test.ts#L336

packages/frameworks/src/frameworks.ts
```diff
@@ -1732,7 +1732,7 @@ export const frameworks = [
         {
           path: 'package.json',
           matchContent:
-            '"(dev)?(d|D)ependencies":\\s*{[^}]*vitepress:\\s*".+?"[^}]*}',
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"vitepress":\\s*".+?"[^}]*}',
```
EndangeredMassa, line 1735: I ran a quick test and it looks like this works:

```
> !!('"dependencies": { "vitepress": "1.2.3" }'.match(new RegExp( '"(dev)?(d|D)ependencies":\\s*{[^}]*vitepress:\\s*".+?"[^}]*}' )));
false

> !!('"dependencies": { "vitepress": "1.2.3" }'.match(new RegExp( '"(dev)?(d|D)ependencies":\\s*{[^}]*"vitepress":\\s*".+?"[^}]*}' )));
true
```

EndangeredMassa (approved):

PR #9081, opened 2022-12-15
feat(edge): adds json() response helper

feugy:
### 🔖 What's in there?

Because Typescript's `libdom` does not have [`static Response.json()`](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1444) (which modern browsers and edge runtimes are supporting), Typescript users can't use easily use it.

This helper fills the gap.

### 🧪 How to test?

It's covered with unit tests

### DIFF
--- a/packages/edge/src/response.ts
+++ b/packages/edge/src/response.ts
@@ -1,0 +1,22 @@
+/**
+ * Builds a response object from a serializable JavaScript object:
+ * - sets the 'Content-Type' response header to 'application/json'
+ * - sets the response body from provided data
+ *
+ * @see {@link https://fetch.spec.whatwg.org/#dom-response-json}
+ * @param data serialized data
+ * @param init optional custom response status, statusText and headers
+ *
+ * @example
+ * <caption>Building a JSON response</caption>
+ *
+ * ```ts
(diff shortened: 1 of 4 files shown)


packages/edge/src/response.ts
```diff
@@ -0,0 +1,22 @@
...
+ * <caption>Building a JSON response</caption>
+ *
+ * ```ts
+ * import { json } from '@vercel/edge';
+ *
+ * const response = json({ notification: { success: true, content: 'worked' } }, { headers: {'x-custom': '1' }})
+ * ```
+ */
+export function json(data: any, init?: ResponseInit): Response {
+  // @ts-expect-error This is not in lib/dom right now, and we can't augment it.
```
javivelasco, line 20: We can also implement it with `new Response`

javivelasco (approved):

packages/edge/src/response.ts
feugy: But then I would have to set the content type header, and invoke JSON.stringify(). Lazyness is good sometimes 😅 

(unless there's a strong reason to not use native `Response.json()` of course)

PR #9098, opened 2022-12-19
[cli] Replace update-notifier dependency with build in

cb1kenobi:
This PR replaces the `update-notifier` dependency with a custom implementation.

There are a few reasons: the dependency is quite large, it requires ESM in order to update, can sometimes suggest an update to an older version, and used dependencies with known security issues.

The result looks like:

image

Note: This PR is the successor to https://github.com/vercel/vercel/pull/8090.


packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Spawn the worker, wait for the worker to report it's ready, then signal the
+ * worker to fetch the latest version.
+ */
+function spawnWorker(
+  payload: GetLatestWorkerPayload,
+  output: Output | undefined
+) {
+  // we need to find the update worker script since the location is
+  // different based on production vs tests
```
styfle, line 103: This part is concerning. We should be testing the actual built output

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,226 @@
...
+ * This file is spawned in the background and checks npm for the latest version
+ * of the CLI, then writes the version to the cache file.
+ *
+ * NOTE: Since this file runs asynchronously in the background, it's possible
+ * for multiple instances of this file to be running at the same time leading
+ * to a race condition where the most recent instance will overwrite the
+ * previous cache file resetting the `notified` flag and cause the update
+ * notification to appear for multiple consequetive commands. Not the end of
+ * the world, but something to be aware of.
+ */
```
styfle, line 15: This comment should also mention that no 3rd part dependencies can be used in this file

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,226 @@
...
+
+const https = require('https');
+const { mkdirSync, writeFileSync } = require('fs');
+const {
+  access,
+  mkdir,
+  readFile,
+  unlink,
+  writeFile,
+} = require('node:fs/promises');
```
styfle: Any reason you're using `node:` prefix here but not the others?

Perhaps it should be used on all of them to make it clear these are all built-ins.

packages/cli/src/util/get-latest-version/index.ts
cb1kenobi: It does test the actual build output. When running the test, `__filename` resolves `packages/cli/src/util/get-latest-version/index.ts` and when running from production, `__filename` resolves to `dist/index.js`. Then it will walk up the directory tree until it hits root (shouldn't happen) or finds `dist/get-latest-worker.js`.

I agree this is gross, but I seem to recall there was an issue with attempting to find `get-latest-worker.js` relative to `get-latest-version/index.ts`.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: Fixed

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: Copy and pasted from the Node.js docs. Didn't really give it much thought. I removed the `node:`, but I think we should eventually add it everywhere.

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -0,0 +1,224 @@
...
+    // fetch the latest version from npm
+    const agent = new https.Agent({
+      keepAlive: true,
+      maxSockets: 15, // See: `npm config get maxsockets`
+    });
+    const headers = {
+      accept:
+        'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
+    };
+    const url = `https://registry.npmjs.org/${name}`;
```
styfle: I realize now that this returns 2MB of json. Since it only runs once a week thats probably okay but that is not really nice for anyone on low data or if we later decide to check more often.

cb1kenobi: Fixed! d240f8fefa3c7b1bb7ae9a890b4127a5c15dcbce

EndangeredMassa (approved):

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  updateCheckInterval = 1000 * 60 * 60 * 24 * 7, // 1 week
+}: GetLatestVersionOptions): string | undefined {
+  if (
+    !pkg ||
+    typeof pkg !== 'object' ||
+    !pkg.name ||
+    typeof pkg.name !== 'string'
+  ) {
+    throw new TypeError('Expected package to be an object with a package name');
+  }
```
Ethan-Arrowood, line 52: Is this not handled by TypeScript?

cb1kenobi: This was based on one of my first PRs and I have the habit of validating arguments despite living in a TypeScript world.

PR #9101, opened 2022-12-19
[fs-detectors][cli] move monorepo default setting code to fs-detectors

Ethan-Arrowood:
Makes the monorepo default setting logic reusable by API and includes a `commandForIgnoringBuildStep` setting for Turbo projects.


packages/cli/src/util/build/monorepo.ts
```diff
@@ -1,40 +1,28 @@
...
-      frameworkList: packageManagers,
-    }),
-  ]);
-
   const projectName = basename(workPath);
   const relativeToRoot = relative(workPath, cwd);
 
   const setCommand = (
-    command: 'buildCommand' | 'installCommand',
+    command: 'buildCommand' | 'installCommand' | 'commandForIgnoringBuildStep',
```
EndangeredMassa, line 25: **suggestion (non-blocking):** this PR is described as moving logic, but it also appears to add logic around `commandForIgnoringBuildStep`. In the future, can you either use multiple PRs or mention changes like this in the description?

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (!hasBuildPipeline) {
+      throw new MissingBuildPipeline();
+    }
+
+    return {
+      monorepoManager: 'turbo',
+      buildCommand: `cd ${relativeToRoot} && npx turbo run build --filter=${projectName}...`,
+      installCommand: `cd ${relativeToRoot} && ${packageManager} install`,
+      commandForIgnoringBuildStep: `cd ${relativeToRoot} && npx turbo-ignore`,
+    };
```
EndangeredMassa, line 72: **suggestion (non-blocking):** extract condition bodies into functions.

I think this would be easier to read if each condition delegated to a function for that purpose.

```
switch (monorepoManager) {
  case 'turbo':
    return getTurboDefaultSettings(detectorFilesystem);
  case 'nx':
    return getNxDefaultSettings(detectorFilesystem);
}

return null;
```

EndangeredMassa (approved):

packages/cli/src/util/build/monorepo.ts
Ethan-Arrowood: yep my bad. @chloetedder and i were on the same page about this but I forgot to update things appropriately.

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
Ethan-Arrowood: definitely a good change. I'd like to land this asap so going to punt this recommendation to the next time. I'll add a ticket to track so I improve it later.

packages/fs-detectors/test/fixtures/get-monorepo-default-settings/turbo/turbo.json
```diff
@@ -0,0 +1,2 @@
+// TEST COMMENT TO VERIFY JSON5 SUPPORT
```
chloetedder, line 1: Comment still needed?

chloetedder (approved):

packages/fs-detectors/test/fixtures/get-monorepo-default-settings/turbo/turbo.json
Ethan-Arrowood: yes - if we don't use a JSON5 parser then this comment will cause things to fail (and that is intentional)

PR #9102, opened 2022-12-19
[cli] Fix pagination next when using --limit

andymckay:
When using `--limit` the helpful next text was not being displayed.

This text:

```
➜  status-via-edge git:(main) ✗ vc alias ls
Vercel CLI 28.8.0

aliases found under andymckay [343ms]
source                                              url                                                age
test-some-files-i51lqtzbf-andymckay.vercel.app      test-some-files-andymckay.vercel.app                5d
….

To display the next page run vercel alias ls --next 1669404182512
```

The reason is the check for a hard coded `20`. Instead this sets the default in `getPaginationOptions` to `20` and then checks the value in that. The result is that it will show the next text if you hit the default `20` or whatever you specified in `limit`


packages/cli/src/util/get-pagination-opts.ts
```diff
@@ -14,8 +14,10 @@ export function getPaginationOpts(opts: PaginationOptions) {
...
     (!Number.isInteger(limit) || limit > 100 || limit < 1)
   ) {
-    throw new Error('Please provide an integer from 1 to 100 for option --limit');
+    throw new Error(
+      'Please provide an integer from 1 to 100 for option --limit'
+    );
   }
 
-  return [nextTimestamp, limit];
+  return [nextTimestamp, limit || 20];
```
styfle: Lets change this to an object so it can be referenced by property name

```suggestion
  return { nextTimestamp, limit: limit || 20 };
```

cb1kenobi: +1

andymckay: Rightio, changed that, but then changed the input to `getAliases` to an object. This change means changing a few more functions and all the calls to them, so thought I'd get a quick 👍🏾  at that before changing a bunch more stuff. 

https://github.com/vercel/vercel/pull/9102/commits/361d261be0f8f89f55ac71477ed809ae94685515

cb1kenobi: So far so good!

packages/cli/src/commands/alias/ls.ts
```diff
@@ -45,15 +45,14 @@ export default async function ls(
   output.spinner(`Fetching aliases under ${chalk.bold(contextName)}`);
 
   // Get the list of alias
-  const { aliases, pagination } = await getAliases(
-    client,
-    undefined,
-    ...paginationOptions
-  );
+  const { aliases, pagination } = await getAliases({
+    client: client,
```
styfle:
```suggestion
    client,
```

packages/cli/src/commands/certs/ls.ts
```diff
@@ -58,7 +58,7 @@ async function ls(
     output.log(formatCertsTable(certs));
   }
 
-  if (pagination && pagination.count === 20) {
+  if (pagination && pagination.count === paginationOptions[1]) {
```
styfle, line 61: Is this a typo?

andymckay: I haven't updated the 4 other functions yet, because I wanted to make sure the way I did `alias` was ok. There's a few updates if `alias` works 👍🏾

PR #9103, opened 2022-12-19
[build-utils] Support empty directory entries for `glob()` and `download()`

TooTallNate:
There are some edge-case situations where a dependency or build process is expecting an empty directory to exist, and this expectation would fail currently because `glob()` does not return directory entries.

So update `glob()` to return entries for empty directories, which are also re-created properly when passed to the `download()` function.


packages/build-utils/src/fs/download.ts
```diff
@@ -2,18 +2,23 @@ import path from 'path';
...
-import { remove, mkdirp, readlink, symlink } from 'fs-extra';
+import { remove, mkdirp, readlink, symlink, Stats } from 'fs-extra';
 import streamToBuffer from './stream-to-buffer';
 
 export interface DownloadedFiles {
   [filePath: string]: FileFsRef;
 }
 
-const S_IFMT = 61440; /* 0170000 type of file */
-const S_IFLNK = 40960; /* 0120000 symbolic link */
```
styfle, line 13: Were these consts not working?

I liked the previous code better because it didn't share the Stat object which seems error-prone.

TooTallNate: They work fine, but I figured sharing the logic with Node core is more proper.

cb1kenobi (commented):
LGTM

cb1kenobi (approved):

packages/build-utils/src/fs/download.ts
EndangeredMassa: I think the current code operates correctly, sure. I agree that sharing `STAT` is error-prone, though.  I can imagine a future code change using `STAT` without understanding how that shared value is manipulated, creating a hard to debug error.

Is creating an instance of `Stats` really slow? If this is a hot-path optimization, can we add comments about that?

PR #9107, opened 2022-12-20
Update frameworks.ts for Next.js 13

daveodipo:
When deploying a Next.js app with Vercel, the output seems to point to **public** as the default directory instead of **pages**. This results in a `404: NOT_FOUND` error when displaying the page. This is as confirmed by the **placeholder** value in **./packages/frameworks/src/frameworks.ts** for Next.js which is set to **Next.js default**


styfle (changes requested):
Hi Dave,

Please share the source code for your Next.js app that reproduces the problem.

If you don't feel comfortable sharing here, you can create a support ticket: https://vercel.com/help#issues

Thanks!

daveodipo:
Hi Steven, 

Sure thing.

[source code](https://github.com/daveodipo/tali-consultancy.git)

styfle:
@daveodipo Looks like that project wasn't configured to use Next.js. 

1. Visit your [dashboard](https://vercel.com/dashboard)
2. Select the desired project
3. Select "Settings"
4. Scroll down to the "Build & Development Settings"
5. Find "Framework Preset" and select "Next.js"
6. Deploy again

daveodipo:
Hi @styfle,

Alright. Thanks for the reply

PR #9157, opened 2023-01-03
[build-utils] Revert "Support empty directory entries for `glob()` and `download()`"

EndangeredMassa:
Reverts vercel/vercel#9103

We believe the original PR is the cause of errors like:

> Error: EISDIR: illegal operation on a directory, read


styfle (approved):
Confirmed this fixes the bug with `01-cowsay` ruby test.

Tests were skipped on the [original PR](https://github.com/vercel/vercel/actions/runs/3735932688/jobs/6339745177) (note the 0s) but it correctly ran tests on the [revert PR](https://github.com/vercel/vercel/actions/runs/3833369834/jobs/6524792854) (note the 2m 39s)

We need to figure out why the tests were skipped in that case.

PR #9158, opened 2023-01-03
[frameworks] Fix vitepress screenshot

andymckay:
When you are on this page, creating a project the image here is the screenshot of an example project in the `Vercel Examples` enterprise. However that was not working:

Screen Shot 2023-01-03 at 2 26 58 PM

It does this by hitting the API. In this case the problem is that there is a trailing slash on the domain name, causing the API to fail:

Screen Shot 2023-01-03 at 2 29 48 PM

So this removes that trailing, but I'll see if I can tighten this up as well so in the future it doesn't matter if there is a `/` there.

FYI: this passes unit tests in framework, because it passes the URL in demo through to `URL` and grabs the hostname, which strips out the `protocol` and trailing `/`. Perhaps we should do the same on front.


leerob (approved):
Wow, nice catch!

cb1kenobi (approved):

PR #9163, opened 2023-01-04
[build-utils] Add fix for windows git bash to check path is set

NALZA:


NALZA:
Links to thread [https://github.com/vercel/vercel/discussions/8762](https://github.com/vercel/vercel/discussions/8762)

packages/build-utils/src/clone-env.ts
```diff
@@ -25,8 +25,13 @@ export function cloneEnv(...envs: (Env | undefined)[]): Env {
       // however we lose this proxied value when we destructure and
       // thus we must explicitly copy it, but we must also remove the
       // `Path` property since we can't have both a `PATH` and `Path`
-      obj.PATH = obj.Path;
-      delete obj.Path;
+
+      // while using git bash obj.path is undefined, but obj.PATH is the correct path
+      if(obj.Path !== undefined){
```
cb1kenobi: Instead of a nested `if`, please combine this condition with the `hasOwnProperty` condition.

cb1kenobi (changes requested):

styfle (commented):
Can you add a unit test to this file?

    packages/build-utils/test/unit.clone-env.test.ts

NALZA:
> Can you add a unit test to this file?
> 
> ```
> packages/build-utils/test/unit.clone-env.test.ts
> ```

Yes no worries have added.

packages/build-utils/src/clone-env.ts
```diff
@@ -19,13 +19,16 @@ export function cloneEnv(...envs: (Env | undefined)[]): Env {
...
       // the system path is called `Path` on Windows and Node.js will
       // automatically return the system path when accessing `PATH`,
       // however we lose this proxied value when we destructure and
       // thus we must explicitly copy it, but we must also remove the
       // `Path` property since we can't have both a `PATH` and `Path`
+
       obj.PATH = obj.Path;
+
       delete obj.Path;
     }
```
cb1kenobi: I did some testing and I discovered things are a little tricky. `cloneEnv()` was designed to receive `process.env` as well as arbitrary objects and mix them into a new object called `obj`.

Turns out that `process.env.hasOwnProperty('PATH')` and `process.env.hasOwnProperty('Path')` are `true` on Windows, including Command Prompt, PowerShell, and git-bash terminals.

On Windows, `Object.getOwnPropertyNames(process.env)` will list `Path`, but not `PATH`, however you can still call `process.env.PATH` and will return `Path`. There's some magic going on.

This means this `hasOwnPoperty.call(env, 'Path')` will always be `true` on Windows. When we clone, we want to make sure `PATH` is set and `Path` is unset since the intention is to pass the result into a `spawn()` call.

I believe this code needs an outer `if` that checks if `Path` is present in the cloned `obj` and an inner `if` to only clobber `obj.PATH` if `obj.Path` is `undefined`. Regardless the value of `obj.Path`, we must delete it.

```suggestion
    // while using git bash obj.path is undefined, but obj.PATH is the correct path
    if (hasOwnProperty.call(obj, 'Path') {
      // the system path is called `Path` on Windows and Node.js will
      // automatically return the system path when accessing `PATH`,
      // however we lose this proxied value when we destructure and
      // thus we must explicitly copy it, but we must also remove the
      // `Path` property since we can't have both a `PATH` and `Path`

      if (obj.Path !== undefined) {
          obj.PATH = obj.Path;
      }
      delete obj.Path;
    }
```

The test looks correct at first glance, but it doesn't account for the magic since the input object is a plain object and secondly the `if` is only deleting `obj.Path` if it is not `undefined`. We should change/add a test that clones 2 args: `{ PATH: 'baz' }` and a `Proxy` for `Path=undefined`, then check that it equals `{ PATH: 'baz' }`. Here's a proxy example: https://github.com/vercel/vercel/blob/fc0e6872e51d53fffc137658c95acf2a249ec926/packages/build-utils/test/unit.clone-env.test.ts#L68-L80

cb1kenobi (changes requested):

cb1kenobi (approved):
I've tested this fix by running `vc dev` using the Command Prompt, PowerShell, and git-bash. `vc dev` now works as expected. Thank you @NALZA!

Ethan-Arrowood (approved):

EndangeredMassa (approved):

packages/build-utils/src/clone-env.ts
```diff
@@ -25,7 +25,11 @@ export function cloneEnv(...envs: (Env | undefined)[]): Env {
       // however we lose this proxied value when we destructure and
       // thus we must explicitly copy it, but we must also remove the
       // `Path` property since we can't have both a `PATH` and `Path`
-      obj.PATH = obj.Path;
+
+      if (obj.Path !== undefined) {
+        obj.PATH = obj.Path;
+      }
+
       delete obj.Path;
```
styfle, line 33: Should we move the `delete obj.Path` inside the if statement or should it always be deleted even when its undefined?

cb1kenobi: We have to delete `obj.Path` here, even if it's `undefined`. The "outer" if will detect the property exists, so we need to always remove it as we can't have both `PATH` and `Path`.

It is technically possible for `PATH` to not exist and `Path` to exist, but be `undefined` in which case the result will be neither `PATH` or `Path` exist. This would probably the same as `PATH` or `Path` existing, but being `undefined`. Regardless, we can't have both `PATH` and `Path` coexisting or bad things happen.

PR #9164, opened 2023-01-04
[build-utils] Support empty directory entries for `glob()` and `download()`

Ethan-Arrowood:
Reopening #9103 since it was reverted.


packages/build-utils/src/fs/glob.ts
```diff
@@ -64,5 +72,20 @@ export default async function glob(
     }
   }
 
+  // Add empty directory entries
+  for (const relativePath of dirs) {
+    if (dirsWithEntries.has(relativePath)) continue;
```
EndangeredMassa, line 77: **suggestion (non-blocking):** Should this look for the dirname of the path or am I misreading this?
 
```suggestion
    if (dirsWithEntries.has(path.dirname(relativePath))) continue;
```

EndangeredMassa (approved):

cb1kenobi:
How do I test this? I tried creating an empty directory in my test project, then I've tried `vc deploy` and my empty directory is not "uploaded".

I tried `vc build` and there's no mention of my empty directory. I did some quick digging and I'm wondering if I need to use BOA v2, but that sounds weird.

I checked the source from my deployment in the web UI and my empty directory isn't there.

I did not test a git-based project and I did not test these changes in the context of the build container.

The code looks good, the tests pass, and that's fine enough for me, but just thought there might be a way to actually test this.

packages/build-utils/src/fs/glob.ts
TooTallNate: We already did dirname in the loop above that populates this set, so this is correct.

TooTallNate:
@cb1kenobi The tests included here should be sufficient IMO. This change isn't going to affect the CLI deployment itself yet, but it will be relevant for `vc build` and also the globbing that is done in the vc-builder inside the build-container.

For CLI itself to upload empty directories there's yet another PR, which still needs tests: https://github.com/vercel/vercel/pull/9118

TooTallNate (approved):

PR #9166, opened 2023-01-05
[frameworks] fix blitzjs detection

EndangeredMassa:
The framework detectors do not support this syntax:

```
path: 'blitz.config.(js|ts)',
```

This PR fixes framework detection for "Blitz.js (Legacy)".

### DIFF
--- a/packages/fs-detectors/test/unit.framework-detector.test.ts
+++ b/packages/fs-detectors/test/unit.framework-detector.test.ts
@@ -420,5 +420,13 @@ describe('DetectorFilesystem', () => {
 
       expect(await detectFramework({ fs, frameworkList })).toBe('zola');
     });
+
+    it('Detect Blitz.js (Legacy)', async () => {
+      const fs = new VirtualFilesystem({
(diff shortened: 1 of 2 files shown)


styfle (approved):
Thanks!

packages/fs-detectors/test/unit.framework-detector.test.ts
```diff
@@ -420,5 +420,13 @@ describe('DetectorFilesystem', () => {
 
       expect(await detectFramework({ fs, frameworkList })).toBe('zola');
     });
+
+    it('Detect Blitz.js (Legacy)', async () => {
```
cb1kenobi, line 424: Out of curiosity, what makes this legacy? Did they switch to `.ts`? Do we have a test for `blitz.config.ts`?

EndangeredMassa: I don't know the history.

I didn't add a test for the `ts` version. Should we flesh out the tests to cover all the cases for all the detectors? I'm not opposed.

PR #8073, opened 2022-07-02
[frameworks] Add "hydrogen" framework preset

TooTallNate:
Adds "hydrogen" framework preset for Shopify Hydrogen framework to allow for zero-config usage of the `@vercel/hydrogen` Builder.

PR #7761, opened 2022-05-05
[frameworks] Update Remix detection to check for "remix.config.js"

mcansh:
we don't encourage the "remix" magic package anymore, and anyone who doesn't have it installed will have their app framework default to "other" and cause it to 404

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [x] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #9167, opened 2023-01-05
[frameworks] update framework detectors with packages

EndangeredMassa:
In [a different PR](https://github.com/vercel/vercel/pull/9009), detecting frameworks by package name will also provide framework version metadata to the build. Should we update these framework detectors to look up their respective packages or were they not doing that already for a reason?

I left the old detectors in place as fallbacks, which looks like:

```
some: [
  {
    path: 'package.json',
    matchContent:
      '"(dev)?(d|D)ependencies":\\s*{[^}]*"remix":\\s*".+?"[^}]*}',
  },
  {
    path: 'remix.config.js',
  },
],
```

Please review carefully.


styfle (changes requested):
Blitz legacy doesn’t look for the package intentionally I believe. 

These need to be checked against all the examples. Also check git blame to find original PRs adding those frameworks.

EndangeredMassa:
@styfle good call! I instead added a note about why we don't detect package: [`2da2506` (#9167)](https://github.com/vercel/vercel/pull/9167/commits/2da2506ad06ea811ae79e6f33ac51e0a4823ba40)

cb1kenobi (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -204,7 +206,12 @@ export const frameworks = [
     useRuntime: { src: 'package.json', use: '@vercel/remix' },
     ignoreRuntimes: ['@vercel/node'],
     detectors: {
-      every: [
+      some: [
+        {
+          path: 'package.json',
+          matchContent:
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"remix":\\s*".+?"[^}]*}',
+        },
```
EndangeredMassa: Remix should not use package name: https://github.com/vercel/vercel/pull/7761

packages/frameworks/src/frameworks.ts
```diff
@@ -1942,6 +1964,11 @@ export const frameworks = [
     useRuntime: { src: 'package.json', use: '@vercel/hydrogen' },
     detectors: {
       some: [
+        {
+          path: 'package.json',
+          matchContent:
+            '"(dev)?(d|D)ependencies":\\s*{[^}]*"@shopify/hydrogen":\\s*".+?"[^}]*}',
+        },
```
EndangeredMassa: @TooTallNate was there a reason this looked for the config file and not the package? Do you think it's OK to add the package as an optional lookup as well?

I ask you because you originally added the framework to the list: https://github.com/vercel/vercel/pull/8073

packages/frameworks/src/frameworks.ts
```diff
@@ -205,6 +207,8 @@ export const frameworks = [
     ignoreRuntimes: ['@vercel/node'],
     detectors: {
       every: [
+        // Intentionally does not detect a package name
+        // https://github.com/vercel/vercel/pull/7761
```
EndangeredMassa, line 207: Because we have a [Remix builder](https://github.com/vercel/vercel/tree/main/packages/remix), we can still probably get version info out of it. Builders can optionally return identified versions. We'd have to figure out how to identify a version there.

Ethan-Arrowood (approved):
Akin to @styfle comment, as long as the CI is passing with these new detectors I'm content with merging this. Would maybe be interested in seeing some additional examples for the frameworks where like maybe the config file is missing, but the package is specified? A lot of these now use `some` where maybe `every` should be used instead since like maybe a brunch project requires both `brunch-config.js` as well as `brunch` as a dependency.

EndangeredMassa:
@Ethan-Arrowood The detectors I changed to `some` should indeed be `some`, in my opinion. I wanted to keep the detection as backwards-compatible as possible when messing with frameworks I'm less familiar with.

If there are some we know for sure should be `every`, then let's change it.

EndangeredMassa:
I tested all `./examples` on `main` and this branch. In both branches, "hydrogen", "sveltekit", and "vuepress" were not detected properly. We need to look into that, but this PR didn't break detection of examples.

I'm starting a separate PR to dig into those failures: https://github.com/vercel/vercel/pull/9197

packages/frameworks/src/frameworks.ts
TooTallNate: I don't recall, but package name lookup would probably be fine.

EndangeredMassa:
Added more tests

styfle (approved):
Thanks!

PR #9173, opened 2023-01-05
[fs-detectors] Remove placeholders for monorepo managers

chloetedder:
We are no longer going to be displaying placeholders for monorepo managers (Turbo, Nx, Rush). This has caused a lot of confusion from users.


cb1kenobi:
LGTM, but not sure exactly how `placeholder` is referenced. Waiting for another reviewer to take a peek.

styfle (approved):

styfle:
It would be good to include a screenshot of this change before/after

agadzik (approved):

PR #9174, opened 2023-01-06
Add in `VERCEL_ENV` for `vc deploy`

andymckay:
It would be nice in vc pull if an environment variable could be used. So that it can be done once and is consistent for multiple commands. This adds in `VERCEL_ENV` as a environment variable.

For other commands you can use the `--environment` variable. For `deploy` it assumed to be preview unless you specify `--prod`.

This again has `VERCEL_ENV` as a fallback, that is overridden by setting the `--prod` argument.

- [x] `--prebuilt` support
- [x] support when `--prebuilt` is not specified

### DIFF
--- a/packages/cli/src/util/deploy/parse-target.ts
+++ b/packages/cli/src/util/deploy/parse-target.ts
@@ -37,11 +37,24 @@ export default function parseTarget(
     return 'production';
   }
 
-  if (process.env.VERCEL_ENV) {
+  const VERCEL_ENV = process.env.VERCEL_ENV;
+  if (VERCEL_ENV) {
+    if (!['preview', 'staging', 'production'].includes(VERCEL_ENV)) {
+      output.error(
+        `The specified environment variable ${param('VERCEL_ENV')} ${code(
+          VERCEL_ENV
+        )} is not valid.`
+      );
+      return 1;
+    }
     output.debug(
       `Setting target to ${process.env.VERCEL_ENV} using VERCEL_ENV environment variable.`
     );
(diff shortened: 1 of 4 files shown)


packages/cli/src/util/deploy/parse-target.ts
```diff
@@ -37,5 +37,12 @@ export default function parseTarget(
     return 'production';
   }
 
+  if (process.env.VERCEL_ENV) {
+    output.debug(
+      `Setting target to ${process.env.VERCEL_ENV} using VERCEL_ENV environment variable.`
+    );
+    return process.env.VERCEL_ENV;
```
styfle: Should this env var be validated first?

andymckay: Yep, thanks. It's in a later commit 👇🏾 😄

packages/cli/src/util/deploy/parse-target.ts
```diff
@@ -37,5 +37,25 @@ export default function parseTarget(
     return 'production';
   }
 
+  const VERCEL_ENV = process.env.VERCEL_ENV;
+  if (VERCEL_ENV) {
+    if (!['preview', 'staging', 'production'].includes(VERCEL_ENV)) {
```
styfle, line 42: I don't think we should allow `staging` here. 

Also cc @lucleray who knows more about this target.

EndangeredMassa:
Closing this for now. If anyone needs these, please open a [discussion](https://github.com/vercel/vercel/discussions).

PR #9176, opened 2023-01-06
Allow `VERCEL_ENV` for `vc build`

andymckay:
It would be nice in vc env if an environment variable could be used instead of --environment. So that it can be done once and is consistent for multiple commands. This adds in VERCEL_ENV as a environment variable.

Setting --environment on the command line will override it.


EndangeredMassa:
Closing this for now. If anyone needs these, please open a [discussion](https://github.com/vercel/vercel/discussions).

PR #9183, opened 2023-01-06
[static-build] add support for gatsby esm config

Ethan-Arrowood:
Adds support for auto injection into Gatsby's new ESM config format. Only injects if the `gatsby-config.mjs` is present, and continues to default generation of `gatsby-config.js`


packages/static-build/test/builds.test.js
```diff
@@ -290,6 +290,54 @@ it(
...
+    );
+
+    const contents = await fs.readdir(workPath);
+
+    expect(contents.some(name => name === 'gatsby-config.js')).toBeFalsy();
+    expect(contents.some(name => name === 'gatsby-config.ts')).toBeFalsy();
+    expect(contents.some(name => name === 'gatsby-config.mjs')).toBeTruthy();
+    // using `import` causes a seg fault.
+    expect(await fs.readFile(path.join(workPath, 'gatsby-config.mjs'), 'utf-8'))
+      .toBe(`import userConfig from "./gatsby-config.mjs.__vercel_builder_backup__.mjs";
```
EndangeredMassa, line 307: **suggestion (non-blocking):** make assertion less specific

This test assertion is brittle when we want to change the injected code. What do you think about asserting only the first line, which captures the core of the difference between this type of injection vs. the others?

EndangeredMassa (approved):

packages/static-build/test/builds.test.js
Ethan-Arrowood: This matches the other test which utilize snapshots. This stuff rarely changes so I'm fine with leaving as is. Maybe using some other form of snapshotting will work better too

packages/static-build/test/build-fixtures/14-gatsby-with-esm-config/vercel.json
```diff
@@ -0,0 +1,13 @@
+{
+  "version": 2,
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@vercel/static-build",
+      "config": {
+        "zeroConfig": true,
+        "framework": "gatsby"
```
styfle, line 9: I believe vercel.json can be simplified to a single line now that we support framework

https://vercel.com/docs/project-configuration#project-configuration/framework

```json
{
  "framework": "gatsby"
}
```

styfle (approved):

packages/static-build/test/build-fixtures/14-gatsby-with-esm-config/vercel.json
Ethan-Arrowood: <picture data-single-emoji=":true-dat:" title=":true-dat:">:true-dat:</picture> I'm going to merge as is since this copies the other ones and I don't want to gold plate in this PR. Going to be modifying static-build including tests soon so I'll make sure to incorporate this recommended change then

PR #9191, opened 2023-01-09
Migrate gatsby-plugin-vercel to @vercel/gatsby-plugin-vercel-analytics

Ethan-Arrowood:
This PR migrates https://github.com/vercel/gatsby-plugin-vercel to this repo so that we can eventually integrate into our monorepo tooling. 

Since it currently requires Node.js v12 to be built, I've manually built, copied, and committed the package as it exists now so that we can publish a stable v1 of `@vercel/gatsby-plugin-vercel-analytics`. 

After work completes on adding support for Gatsby v4, I will come back to this package and update it to Node.js v16 so that it can be integrated properly into the repository's tooling.

🚨 DO NOT SQUASH MERGE 🚨


Ethan-Arrowood:
Interesting, looks like you can build this project with node 16, but in minor 18 (https://github.com/vercel/gatsby-plugin-vercel/actions/runs/3875718651/jobs/6609774745#step:3:7) . I tried having this install and build using 16.19.x and it didn't work. Its up to other folks if we want to try solving the build issue before landing this but IMO we land as is and focus on reworking plugin later. I also noticed the web-vital dependency is severely out of date (we have v0.x and v3.x is available).

styfle (approved):
Looks good. Remember to do a proper merge (instead of squash) if you want to keep all the git history.

Also remember to deprecate the old npm package and old repo

PR #9194, opened 2023-01-09
[static-build] @vercel/static-build to use @vercel/gatsby-plugin-vercel-analytics

Ethan-Arrowood:
Updates the `static-build` injector to use the new plugin.

Still need to verify somehow that the newly published plugin is working as expected. It should be fine since it was a copy-paste from the previous plugin repo, but always good to verify before we break everything! 

This PR also updates the README in `@vercel/gatsby-plugin-vercel-analytics`


packages/static-build/test/builds.test.js
```diff
@@ -321,13 +321,13 @@ if (!vercelConfig.plugins) {
 
 const hasPlugin = vercelConfig.plugins.find(
   (p) =>
-    p && (p === "gatsby-plugin-vercel" || p.resolve === "gatsby-plugin-vercel")
+    p && (p === "@vercel/@vercel/gatsby-plugin-vercel-analytics-analytics" || p.resolve === "@vercel/@vercel/gatsby-plugin-vercel-analytics-analytics")
```
styfle: This looks like a bug since it has `@vercel/@vercel`

styfle (approved):

PR #9196, opened 2023-01-10
[next] Add Operation Types to Next.js Lambdas

EndangeredMassa:
In order to have Next.js Lambdas show their operation types more specifically in the build output in the dashboard, the builder needs to return the Lambdas with `operationType` set to the appropriate value.

This PR adds those values. This allows the Richer Deployment Outputs to show the different types of serverless functions:
Screenshot 2023-02-03 at 3 49 42 PM


packages/next/src/server-build.ts
```diff
@@ -869,6 +869,7 @@ export async function serverBuild({
           ),
           '___next_launcher.cjs'
         ),
+        operationType: 'SSR', // TODO: Are these only ever SSR?
```
ijjk: This can be inferred from the `group.isPrerenders` and then can add a flag to the API group to signal that as well then otherwise it would be SSR.

packages/next/src/index.ts
```diff
@@ -1826,6 +1827,7 @@ export const build: BuildV2 = async ({
                 path.relative(baseDir, entryPath),
                 '___next_launcher.cjs'
               ),
+              operationType: 'SSR', // TODO: What should these be?
```
ijjk: This would always be SSR or API as it predates ISR.

packages/next/src/index.ts
```diff
@@ -1082,7 +1082,7 @@ export const build: BuildV2 = async ({
           handler: '___next_launcher.cjs',
           runtime: nodeVersion.runtime,
           ...lambdaOptions,
-          operationType: 'SSR',
+          operationType: 'SSR', // TODO: Is this ever ISR?
```
ijjk: This is always SSR as it's for legacy Next.js version.

packages/next/src/index.ts
```diff
@@ -1807,6 +1807,7 @@ export const build: BuildV2 = async ({
                 path.relative(baseDir, entryPath),
                 '___next_launcher.cjs'
               ),
+              operationType: 'SSR', // TODO: What should these be?
```
ijjk: These can be SSR, API or ISR. To infer ISR here would have to look at the `prerender-manifest`

packages/next/src/utils.ts
```diff
@@ -2666,3 +2668,50 @@ function transformSourceMap(
...
+    const routePaths = Object.keys({
+      ...prerenderManifest?.staticRoutes,
+      ...prerenderManifest?.blockingFallbackRoutes,
+      ...prerenderManifest?.fallbackRoutes,
+    });
+
+    if (routePaths.includes(pageFileName)) {
+      return 'ISR';
+    }
+  }
```
EndangeredMassa, line 2714: Is this the right way to check the `prerenderManifest`?

ijjk: Only `blockingFallbackRoutes` or `fallbackRoutes` would match `pageFileName` here static routes are the normal path not the page e.g. `/blog/first` instead of `/blog/[slug]`

EndangeredMassa:
@cb1kenobi tested this branch against a bunch of Next.js projects. It seems to work in all of the test cases we could muster.

Can sone folk from the Next.js team take a look?

packages/next/test/fixtures/00-server-build/index.test.js
```diff
@@ -22,7 +22,7 @@ async function checkForChange(url, initialValue, getNewValue) {
 const ctx = {};
 
 describe(`${__dirname.split(path.sep).pop()}`, () => {
-  it('should deploy and pass probe checks', async () => {
```
TooTallNate: Was this change intentional?

EndangeredMassa: Yes. It mitigates some test pollution I noticed while running these tests to make sure changes didn't break anything.

ijjk: if you change this to `beforeAll()` and it fails the tests will hang indefinitely

EndangeredMassa: @ijjk I simulated an error: [`c9c4225` (#9196)](https://github.com/vercel/vercel/pull/9196/commits/c9c4225361fe63bdc1e9cf647fab5361797256e8)

and it seemed to fail fine: https://github.com/vercel/vercel/actions/runs/4105864790/jobs/7083376282#step:9:465

but I'm happy to revert if this is an issue.

EndangeredMassa: reverted for now: [`ba22f16` (#9196)](https://github.com/vercel/vercel/pull/9196/commits/ba22f16676c2a62ad9f8e280911973ef063821b5)

cb1kenobi (dismissed):

EndangeredMassa:
@ijjk @styfle can you take another look at this PR?

styfle (commented):
Please add tests.

For example: https://github.com/vercel/vercel/blob/d8fb3def8a664139cd7e9d4699f89e709395b49b/packages/next/test/integration/index.test.js

ijjk (dismissed):
Looks good to me, tests would be nice as mentioned above

EndangeredMassa:
I'm working on tests for this and may have found an issue. Digging deeper.

cb1kenobi (dismissed):
LGTM

packages/next/src/utils.ts
```diff
@@ -2677,3 +2679,49 @@ function transformSourceMap(
...
+
+  if (group?.isPrerenders) {
+    return 'ISR';
+  }
+
+  if (pageFileName) {
+    const routePaths = Object.keys({
+      ...prerenderManifest?.blockingFallbackRoutes,
+      ...prerenderManifest?.fallbackRoutes,
+    });
```
styfle: How many `routePaths` do we expect here? Is it bounded or can it grow to N?

It could be a lot faster to do an early return like this:

```ts
for (const key of Object.keys(blockingFallbackRoutes)) {
  if (key === pageFileName) return 'ISR'
}

for (const key of Object.keys(fallbackRoutes)) {
  if (key === pageFileName) return 'ISR'
}
```

styfle: If the same `prerenderManifest` is passed to this function more than once, its probably best to hoist it up and precompute a Set that is then passed into this function so you can do `set.has(pageFileName)`

EndangeredMassa:
It turns out my test setup was off. The code works as it did before, but now there are assertions that different operation types make it into the output. [`64753f8` (#9196)](https://github.com/vercel/vercel/pull/9196/commits/64753f89c4ef1fc60a47c3b66e2d2a59329922d0)

packages/next/src/utils.ts
cb1kenobi: Yeah, I thought about commenting on this too, but didn't think of breaking it into two different tests. Maybe skip the `Object.keys()` call and  allocation and use `Object.prototype.hasOwnProperty.call(blockingFallbackRoutes, pageFileName)`?

EndangeredMassa: Did the initial optimization: [`ecf4712` (#9196)](https://github.com/vercel/vercel/pull/9196/commits/ecf4712d8ea8ed352620e625caa55a00f070ec3d)

Looking into a better optimization.

styfle: I think precomputing the Set is actually best. See existing implementation here:

https://github.com/vercel/vercel/blob/ecf4712d8ea8ed352620e625caa55a00f070ec3d/packages/next/src/index.ts#L501-L503

cb1kenobi: For performance, `Object.prototype.hasOwnProperty.call()` is going to be faster and more efficient since there's no memory being allocated, but it does look ugly.

```
if (pageFileName) {
  const {
    blockingFallbackRoutes = {},
    fallbackRoutes = {},
  } = prerenderManifest || {};
  const hasOwn = Object.prototype.hasOwnProperty;
  if (hasOwn.call(blockingFallbackRoutes, pageFileName) || hasOwn.call(fallbackRoutes, pageFileName)) {
    return 'ISR';
  }
}
```

How about using `in`?

```
if (pageFileName) {
  const {
    blockingFallbackRoutes = {},
    fallbackRoutes = {},
  } = prerenderManifest || {};
  if (pageFileName in blockingFallbackRoutes || pageFileName in fallbackRoutes) {
    return 'ISR';
  }
}
```

There used to be performance issues with `in`, but it seems to be ok these days.

If sticking with `Object.keys()`, then perhaps `if (Object.keys(<obj>).find(k => k === pageFileName)) { }`?

styfle: Yeah “in” could work too 👍

EndangeredMassa: I did some [quick perf testing](https://gist.github.com/EndangeredMassa/89f4ea35d5365537c7cb1d1f22a42595). Let me know if any of that looks wrong. If correct:

- looping with short circuiting: slow
- in operator: very fast
- building up a set to use: very slow
- checking that set: very fast

Yeah, `in` looks like a win here. [`184ef3d` (#9196)](https://github.com/vercel/vercel/pull/9196/commits/184ef3d6b5185b3068a74d22d7cf57d668058e46)

packages/next/src/utils.ts
```diff
@@ -2677,3 +2679,49 @@ function transformSourceMap(
...
+    return 'API';
+  }
+
+  if (group?.isPrerenders) {
+    return 'ISR';
+  }
+
+  if (pageFileName) {
+    const { blockingFallbackRoutes = {}, fallbackRoutes = {} } =
+      prerenderManifest || {};
```
styfle: Nit pick: 

```suggestion
  if (pageFileName && prerenderManifest) {
    const { blockingFallbackRoutes = {}, fallbackRoutes = {} } =
      prerenderManifest;
```

styfle (dismissed):

cb1kenobi (approved):

cb1kenobi (approved):

PR #9197, opened 2023-01-10
[examples][frameworks] add tests for all examples being detected

EndangeredMassa:
This PR adds tests (under `packages/fs-detectors`) that ensure are `./examples` get detected as the appropriate framework.


packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1934,7 @@ export const frameworks = [
     ],
   },
   {
+    // TODO: fix detected as "vite"
```
EndangeredMassa: Should "Hydrogen" be moved above "Vite"?

PR where Hydrogen was added: https://github.com/vercel/vercel/pull/8073

examples/sveltekit/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "devDependencies": {
     "@sveltejs/adapter-auto": "next",
-    "@sveltejs/kit": "next",
+    "@sveltejs/kit": "next TODO ",
```
EndangeredMassa: `sveltekit` detection (as opposed to `sveltekit-1` detection) looks for a package version that looks like this part of the regex `"1\\.0\\.0-next\\.(\\d+)"`. Should we update the `./examples/sveltekit` example to match?

This is the PR where it was changed: https://github.com/vercel/vercel/pull/9070/files#diff-837505f1cd8d9a3b7e5986eb08b96e256bf9e5d8d74ce1ac973fe0db4a4961f9L1008

EndangeredMassa: @tcc-sejohnson @leerob what do you think?

packages/frameworks/src/frameworks.ts
EndangeredMassa: @TooTallNate @leerob what do you think?

examples/sveltekit/package.json
styfle: cc @tcc-sejohnson

packages/frameworks/src/frameworks.ts
styfle: I think the answer is yes.

Also cc @TooTallNate

TooTallNate: I think yes.

EndangeredMassa: done: [`f880f35` (#9197)](https://github.com/vercel/vercel/pull/9197/commits/f880f355022a32c845fc589a188c5399d2a63d8b)

examples/sveltekit/package.json
dummdidumm: Makes sense to do this - if in the future we do `2.0.0-next` just `next` would point to that, so this should be changed to `1.0.0-next.589` (the last prerelease version)

EndangeredMassa: Done! [`56e2857` (#9197)](https://github.com/vercel/vercel/pull/9197/commits/56e2857bcad8eadbd768a3f04a40234405174aa5)

cb1kenobi (approved):
LGTM!

examples/sveltekit/package.json
styfle: Lets add a lock file if there isn't already one

TooTallNate (approved):

leerob (approved):

cb1kenobi (approved):

PR #9198, opened 2023-01-10
[tests] Migrate from yarn to pnpm

Ethan-Arrowood:
<picture data-single-emoji=":pnpm:" title=":pnpm:">:pnpm:</picture> 

yarn has become increasingly more difficult to use as the v1 we rely on no longer receives updates. pnpm is faster and is actively maintained. 

This PR migrates us to pnpm.


package.json
```diff
@@ -3,27 +3,19 @@
   "version": "0.0.0",
   "private": true,
   "license": "Apache-2.0",
-  "packageManager": "yarn@1.22.19",
-  "workspaces": {
```
TooTallNate, line 7: We don't need this anymore? I think perhaps lerna relies on it.

packages/cli/package.json
```diff
@@ -49,8 +49,10 @@
     "@vercel/python": "3.1.33",
     "@vercel/redwood": "1.0.44",
     "@vercel/remix": "1.1.6",
+    "@vercel/routing-utils": "workspace:2.1.3",
     "@vercel/ruby": "1.3.49",
     "@vercel/static-build": "1.1.0",
+    "i": "^0.3.7",
```
TooTallNate: What's this?

Ethan-Arrowood: that looks like a mistake. i was running a lot of `pnpm i` commands to get everything building

package.json
Ethan-Arrowood: not sure - lerna is used for publishing packages only right? I can keep this in here, but it seemed like they were `yarn` configurations. I wonder if since I configured lerna to `pnpm` it will look at the `pnpm-workspace.yaml` instead

packages/cli/package.json
```diff
@@ -49,8 +49,10 @@
     "@vercel/python": "3.1.33",
     "@vercel/redwood": "1.0.44",
     "@vercel/remix": "1.1.6",
+    "@vercel/routing-utils": "workspace:2.1.3",
```
TooTallNate: Is `workspace:` necessary here?

packages/cli/package.json
cb1kenobi: Ha, I've done this before :)

packages/cli/package.json
Ethan-Arrowood: Ahh thats a pnpm thing. https://pnpm.io/cli/add#install-from-the-workspace

Might have to update all them to use this `workspace:` syntax

package.json
cb1kenobi: @TooTallNate is correct, `lerna` still needs `workspaces` in the `package.json`.

```
$ lerna changed
lerna notice cli v6.0.3
lerna ERR! EWORKSPACES Workspaces need to be defined in the root package.json.
lerna ERR! EWORKSPACES See: https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces
```

.github/workflows/publish.yml
```diff
@@ -17,51 +17,51 @@ jobs:
...
-        GA_TRACKING_ID: ${{ secrets.GA_TRACKING_ID }}
-        SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
-    - name: Publish
-      if: ${{ steps.check-release.outputs.IS_RELEASE == 'true' }}
-      run: yarn publish-from-github
-      env:
-        NPM_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
-        GA_TRACKING_ID: ${{ secrets.GA_TRACKING_ID }}
-        SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+      - name: Checkout
```
styfle: Is this whitespace change correct? Curious why its changing now?

Ethan-Arrowood: Might be some auto formatting on my end - will revert

packages/cli/package.json
cb1kenobi: Related/unrelated, FWIW when you create a new Turbo repo, the template uses `workspace:`: https://github.com/vercel/turbo/blob/main/examples/basic/packages/ui/package.json#L14.

.github/workflows/publish.yml
```diff
@@ -17,51 +17,51 @@ jobs:
...
+      - name: Cache
+        if: ${{ steps.check-release.outputs.IS_RELEASE == 'true' }}
+        uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: pnpm-${{ matrix.os }}-${{ matrix.node }}-${{ hashFiles('pnpm-lock.yaml') }}
+          restore-keys: pnpm-${{ matrix.os }}-${{ matrix.node }}
+      - name: Install
+        if: ${{ steps.check-release.outputs.IS_RELEASE == 'true' }}
+        run: pnpm install --check-files --frozen-lockfile --network-timeout 1000000
```
styfle: Before each `pnpm install` step, there should be a `corepack enable` step.

.github/workflows/publish.yml
Ethan-Arrowood: Unless its a correct fix? I trust the auto formatter more than myself lol tho yaml is yaml so 🤷‍♂️

styfle: I'm guessing you have a vs code plugin thats formatting and its not something our repo automatically does

packages/cli/package.json
styfle: Yeah that seems correct for `turbo`. However, does `lerna` know how to handle `workspace:`?

packages/gatsby-plugin-vercel-analytics/README.md
```diff
@@ -21,7 +21,7 @@ npm i @vercel/gatsby-plugin-vercel-analytics
 or
 
 ```bash
-yarn add @vercel/gatsby-plugin-vercel-analytics
+pnpm add @vercel/gatsby-plugin-vercel-analytics
```
styfle: Lets either add all options (npm/yarn/pnpm) to the readme, or just have npm only

packages/next/package.json
```diff
@@ -61,6 +61,7 @@
     "get-port": "5.0.0",
     "nanoid": "3.3.4",
     "ndjson": "2.0.0",
+    "pretty-bytes": "^6.0.0",
```
styfle: Seems like the `save-prefix: ""` setting didn't work because the up-caret is still there.

Perhaps we should use `save-exact: true` instead.

packages/redwood/package.json
```diff
@@ -27,6 +27,9 @@
     "@types/aws-lambda": "8.10.19",
     "@types/node": "14.18.33",
     "@types/semver": "6.0.0",
-    "@vercel/build-utils": "5.7.4"
+    "@vercel/build-utils": "5.7.4",
+    "execa": "3.2.0",
+    "fs-extra": "^11.1.0",
```
styfle: Same here: unexpected `^`

utils/publish.sh
```diff
@@ -30,9 +30,9 @@ else
 fi
 
 # Sometimes this is a false alarm and blocks publish
-git checkout yarn.lock
+git checkout pnpm-lock.yaml
```
styfle, line 33: I think we can remove this checkout now since pnpm defaults to frozen lockfile in CI, so it should never happen.

Although we can circle back to it after this PR lands.

packages/cli/package.json
Ethan-Arrowood: https://lerna.js.org/docs/recipes/using-pnpm-with-lerna it supports it

package.json
Ethan-Arrowood: It seems lerna _should_ be able to use pnpm-workspace.yaml https://lerna.js.org/docs/recipes/using-pnpm-with-lerna investigating now

Ethan-Arrowood: Ahh it must only support it in like v5+ - we are using v3 :(

.github/workflows/publish.yml
Ethan-Arrowood: Yeah, but my other comment still stands. Is this change okay if its _better_ yaml format? Other workflows have this indentation. @andymckay do you know if this indentation is okay to keep or do I need to revert and disable my auto-formatter for yaml files?

package.json
cb1kenobi: Can we update Lerna? The last v3 release was 2-1/2 years ago. Lerna (again) is being maintained and it makes most sense to keep current. I'm not sure how the release process would be affected.

Ethan-Arrowood: imo we should migrate off lerna. use pnpm+turbo instead

Ethan-Arrowood: I can try updating it now to atleast v5 

v6 uses Nx lol

cb1kenobi: Sigh, figures. +1 for pnpm+turbo. Are there any projects out there that uses pnpm+turbo for publishing/changelog that we can look to for inspiration?

Ethan-Arrowood: https://github.com/vercel/turbo lol they've got a whole changeset workflow we can clone

Ethan-Arrowood: oh nvm they moved to something else: https://github.com/vercel/turbo/blob/main/.github/workflows/publish-npm.yml

but we should still consult them.

Ethan-Arrowood: yay lerna5 works! try it out locally and let me know if you hit some edge case I didn't

packages/cli/package.json
Ethan-Arrowood: It does in v5! Migrating all workspace packages to the prefix

.github/workflows/test-unit.yml
```diff
@@ -33,22 +33,24 @@ jobs:
           go-version: '1.13.15'
       - uses: actions/checkout@v3
         with:
-            fetch-depth: 2
+          fetch-depth: 2
       - uses: actions/setup-node@v3
         timeout-minutes: 5 # See https://github.com/actions/cache/issues/810
         with:
-            node-version: ${{ matrix.node }}
+          node-version: ${{ matrix.node }}
```
Ethan-Arrowood: @styfle this spacing fix seems correct. If anything we probably need to be enforcing format on these files

.github/workflows/publish.yml
Ethan-Arrowood: If you compare this spacing change to the other files you can see how in the `test-*.yml` files they have indents after `steps:`, but its in this file they are all inline with `steps:`.

packages/next/package.json
Ethan-Arrowood: save-exact is not supported by pnpm so just adding save-prefix which is

utils/publish.sh
Ethan-Arrowood: lets circle back after PR

packages/next/package.json
styfle: Oddly enough, `save-exact=true` work for me with pnpm even though its not documented

Ethan-Arrowood: are you putting that in `.npmrc` ? I tried it and it didn't work for some reason

styfle: Yes in `.npmrc` with `pnpm@7.22.0`

https://github.com/styfle/links-awakening/blob/main/.npmrc

Oddly enough, I couldn't get `save-prefix` to work when I tried it last.

.github/workflows/test-unit.yml
styfle: I would prefer the spacing is fixed in a different PR if possible and added to prettier (since it seems like its currently not auto-formatted)

The reason being: this PR is changing a lot so we should reduce the lines touched where possible

packages/next/package.json
Ethan-Arrowood: lol really? that is interesting. let me try again

.github/workflows/test-unit.yml
Ethan-Arrowood: Fair enough - ill revert and make sure to follow up

packages/next/package.json
Ethan-Arrowood: okay both seem to work. ill stick with save-exact=true

packages/python/package.json
```diff
@@ -16,15 +16,18 @@
...
     "@types/execa": "^0.9.0",
     "@types/jest": "27.4.1",
-    "@vercel/build-utils": "5.7.4",
+    "@vercel/build-utils": "workspace:5.7.4",
     "@vercel/ncc": "0.24.0",
     "execa": "^1.0.0",
     "typescript": "4.3.4"
+  },
+  "dependencies": {
+    "@types/node": "14.18.33"
```
styfle: This should be in `devDependencies`

.github/workflows/publish.yml
Ethan-Arrowood: @TooTallNate and I think that the `corepack enabled` is now disabling yarn from executing within the CLI and causing some tests to fail. Ideas/thoughts? Maybe we don't use corepack and instead just install pnpm manually or use [pnpm/action-setup](https://github.com/pnpm/action-setup) ?

styfle:
In the future we, should use `COREPACK_ENABLE_STRICT=0` however that version of corepack with this feature hasn't shipped in Node.js 14 yet. So might as well install pnpm directly.

- https://github.com/nodejs/corepack/issues/195

.github/workflows/test.yml
```diff
@@ -39,9 +39,11 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: '**/node_modules'
-          key: yarn-${{ matrix.os }}-${{ matrix.node }}-${{ hashFiles('yarn.lock') }}
-          restore-keys: yarn-${{ matrix.os }}-${{ matrix.node }}
-      - run: yarn install --network-timeout 1000000 --frozen-lockfile
+          key: pnpm-${{ matrix.os }}-${{ matrix.node }}-${{ hashFiles('pnpm-lock.yaml') }}
+          restore-keys: pnpm-${{ matrix.os }}-${{ matrix.node }}
+      - name: install pnpm
+        run: npm i -g pnpm
```
styfle: Lets make sure to use `7.24.2` for all our installs so it matches what corepack would have used

packages/build-utils/test/fixtures/15-npm-8-legacy-peer-deps/package-lock.json
```diff
@@ -7,6 +7,9 @@
       "dependencies": {
         "react": "16.8.0",
         "swr": "1.3.0"
+      },
+      "engines": {
+        "node": "16.x"
```
Ethan-Arrowood, line 12: @styfle you were the last to touch this one. Not sure why this is getting generated but wasn't before. I've tried reverting, but it generates when I run the unit tests locally.

packages/build-utils/test/unit.test.ts
```diff
@@ -474,21 +474,21 @@ it('should support initialHeaders and initialStatus correctly', async () => {
 });
 
 it('should support require by path for legacy builders', () => {
-  const index = require('@vercel/build-utils');
+  const index = require('../dist');
```
Ethan-Arrowood: @TooTallNate not sure if theres a better way to do this, but these paths work and now require the project being build first since it can't rely on hoisting lol

packages/build-utils/test/unit.walk.test.ts
```diff
@@ -2,15 +2,15 @@ import { walkParentDirs } from '../src';
 import { strict } from 'assert';
 import { join } from 'path';
 import { promises } from 'fs';
-const { deepEqual, notDeepEqual, fail } = strict;
```
Ethan-Arrowood, line 5: This was a weird one, was getting this error that seemed only solvable by doing this change. Must be connected to finally using `@types/ndoe@14`

```
Assertions require every name in the call target to be declared with an explicit type annotation.ts(2775)
unit.walk.test.ts(5, 9): 'deepEqual' needs an explicit type annotation.
```

packages/edge/docs/README.md
```diff
@@ -36,7 +36,7 @@ City of the original client IP as calculated by Vercel Proxy.
 
 #### Defined in
 
-[src/edge-headers.ts:4](https://github.com/vercel/vercel/blob/main/packages/edge/src/edge-headers.ts#L4)
+[packages/edge/src/edge-headers.ts:4](https://github.com/vercel/vercel/blob/main/packages/edge/src/edge-headers.ts#L4)
```
Ethan-Arrowood, line 39: Not sure why these had to change. But the tests were failing and said to re run the docs:build script 🤷‍♂️

packages/build-utils/test/fixtures/15-npm-8-legacy-peer-deps/package-lock.json
styfle: Its probably fine if the tests pass.

The test is checking to see if `npm install` fails so I'm a little surprised if the lock file updates

See https://github.com/vercel/vercel/pull/8598

packages/build-utils/test/unit.test.ts
styfle: This test was explicitly designed to check the package boundary since we used to have builders that did this:

```
require('@vercel/build-utils/fs/get-writable-directory.js');
```

packages/build-utils/test/unit.walk.test.ts
styfle: Lets bump the types then?

packages/edge/docs/interfaces/ExtraResponseInit.md
```diff
@@ -66,4 +66,4 @@ Omit.statusText
 
 #### Defined in
 
-node_modules/typescript/lib/lib.dom.d.ts:1579
+node_modules/.pnpm/typescript@4.7.4/node_modules/typescript/lib/lib.dom.d.ts:1579
```
styfle: Might want to get approval from the edge team on this change since this looks unexpected

packages/build-utils/test/unit.walk.test.ts
Ethan-Arrowood: If i'm not mistaken bumping types to 16 causes a bunch of other issues. the whole repo is on 14 so we should keep it that way

api/_lib/util/error-handler.ts
```diff
@@ -13,6 +13,7 @@ function initSentry() {
   sentryInitDone = true;
 
   init({
+    // @ts-ignore
```
EndangeredMassa, line 17: **suggestion (non-blocking):** add comment about why

Can you add a comment here about why this was ignored?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -137,6 +137,7 @@ export function execAsync(
       child.on('close', (code, signal) => {
         if (code === 0 || opts.ignoreNon0Exit) {
           return resolve({
+            // @ts-ignore
```
EndangeredMassa, line 142: **suggestion (non-blocking):** add comment about why

Can you add a comment here about why this was ignored?

packages/build-utils/package.json
```diff
@@ -13,8 +13,8 @@
   "scripts": {
     "build": "node build",
     "test": "jest --env node --verbose --runInBand --bail",
-    "test-unit": "yarn test test/unit.*test.*",
-    "test-integration-once": "yarn test test/integration.test.ts"
+    "test-unit": "turbo build --filter=@vercel/build-utils && pnpm test test/unit.*test.*",
```
TooTallNate: Why was the `turbo build` part necessary?

api/package.json
```diff
@@ -16,9 +16,9 @@
     "unzip-stream": "0.3.0"
   },
   "devDependencies": {
-    "@types/node": "14.18.33",
+    "@types/node": "16.18.11",
```
EndangeredMassa, line 18: **question (non-blocking):** Is everything supposed to be node@14 or no?

https://github.com/vercel/vercel/pull/9198/files#r1067359556

packages/build-utils/test/unit.test.ts
TooTallNate: Maybe this is slightly better. Personally I wouldn't mind deleting this test + legacy behavior entirely, but idk might be too risky.

```suggestion
  const index = require('../');
```

packages/build-utils/package.json
Ethan-Arrowood: Since these tests now require the built module, we must ensure its being built every time tests are run. turbo here lets that build hit the cache if nothing has changed. Can change to just `pnpm build` but then no caching will happen

packages/build-utils/test/unit.walk.test.ts
TooTallNate: I'd prefer to switch to `expect`

packages/cli/package.json
TooTallNate: I'm curious how this works with the published package. Does lerna change the version numbers when publishing?

packages/client/README.md
```diff
@@ -10,8 +10,6 @@ Firstly, install the package:
 
 ```bash
 npm install @vercel/client
-# or
-yarn add @vercel/client
```
EndangeredMassa, line 14: **question (non-blocking):** why was this removed?

packages/build-utils/test/unit.walk.test.ts
Ethan-Arrowood: I consider that out of scope for this PR. We can update tests in packages in other PRs since that is its own effort

packages/edge/scripts/fix-links.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+const path = require('path');
+
+const docs = path.join(__dirname, '..', 'docs');
+const interfaces = path.join(docs, 'interfaces');
+
+for (const dir of [docs, interfaces]) {
```
EndangeredMassa, line 9: **suggestion (non-blocking):** add comment about why this is necessary

Can we add a comment here about what exactly is happening and why? It could link to the issue you created: https://github.com/TypeStrong/typedoc/issues/2140

packages/gatsby-plugin-vercel-analytics/README.md
```diff
@@ -18,12 +18,6 @@ This plugin sends [Core Web Vitals](https://web.dev/vitals/) to Vercel Analytics
 npm i @vercel/gatsby-plugin-vercel-analytics
 ```
 
-or
-
-```bash
-yarn add @vercel/gatsby-plugin-vercel-analytics
-```
-
```
EndangeredMassa, line 26: **question (non-blocking):** why was this removed?

utils/chunk-tests.js
```diff
@@ -20,6 +20,7 @@ async function getChunkedTests() {
       `--cache-dir=.turbo`,
       '--output-logs=full',
       '--',
+      '--',
```
EndangeredMassa: **suggestion (non-blocking):** remove?

```suggestion
```

Is this necessary? If it is, can we add a comment about why?

EndangeredMassa:
I tested this out locally and it seems to work pretty well. 🎉

packages/gatsby-plugin-vercel-analytics/README.md
Ethan-Arrowood: @styfle previous comment about our docs should either have `yarn/npm/pnpm` or just `npm`

packages/build-utils/package.json
styfle: @Ethan-Arrowood Something doesn't seem right.

I would expect `turbo build` to automatically be run when you do `turbo test-unit` because of our config here:

https://github.com/vercel/vercel/blob/3f47587a8b204ecc5782fde77dabcee5f1d6e7e7/turbo.json#L24-L25

utils/chunk-tests.js
Ethan-Arrowood: necessary because how  pnpm handles argument parsing

packages/client/README.md
Ethan-Arrowood: @styfle previous comment about our docs should either have yarn/npm/pnpm or just npm

api/package.json
Ethan-Arrowood: yeah but not api since we deploy this with Node 16

packages/client/README.md
styfle: Probably because of my comment here https://github.com/vercel/vercel/pull/9198#discussion_r1066224797

In hindsight, we should leave revert the changes to `packages/*/README.md` because they are for the consumer and therefore not relevant to this PR. They can be updated in a future PR.

packages/gatsby-plugin-vercel-analytics/README.md
styfle: Probably because of my comment here https://github.com/vercel/vercel/pull/9198#discussion_r1066224797

In hindsight, we should leave revert the changes to `packages/*/README.md` because they are for the consumer and therefore not relevant to this PR. They can be updated in a future PR.

packages/build-utils/package.json
Ethan-Arrowood: <picture data-single-emoji=":hmm:" title=":hmm:">:hmm:</picture>

Ethan-Arrowood: Yeah so this does work when `turbo` is used to invoke `test-unit`, but if you call `pnpm test-unit` it wont run

styfle: Yeah but thats expected. You should run `pnpm build` and then `pnpm test-unit` if you're running from the current directory without turbo.

packages/build-utils/package.json
```diff
@@ -13,8 +13,8 @@
...
+    "test-integration-once": "pnpm test test/integration.test.ts"
```
styfle, line 17: This is wrong. It should be only running the integration tests, not the unit tests

Ethan-Arrowood: Huh? It is only running integration tests

styfle: Oh you're right. I misread that as `pnpm test-unit`. Ignore this comment 🙈

api/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "API for the vercel/vercel repo",
   "main": "index.js",
   "scripts": {
-    "//TODO": "We should add this pkg to yarn workspaces"
+    "//TODO": "We should add this pkg to pnpm workspaces"
```
styfle:
```suggestion
```

You already did this TODO so we can remove it 🙂

utils/pack.ts
```diff
@@ -21,6 +21,9 @@ async function main() {
     const dir = path.join(rootDir, task.directory);
     const packageJsonPath = path.join(dir, 'package.json');
     const originalPackageObj = await fs.readJson(packageJsonPath);
+    if (originalPackageObj.name === 'api') {
+      continue;
```
styfle, line 26: Lets add a comment here explaining why

I think its because turbo is finding this by parsing the new pnpm-workspace.yml file but we don't want to create a tarball for it since its a Vercel Deployment already?

packages/cli/package.json
```diff
@@ -41,16 +41,19 @@
...
-    "@vercel/static-build": "1.1.1"
+    "@vercel/build-utils": "workspace:5.7.5",
+    "@vercel/go": "workspace:2.2.24",
+    "@vercel/hydrogen": "workspace:0.0.38",
+    "@vercel/next": "workspace:3.3.9",
+    "@vercel/node": "workspace:2.8.6",
+    "@vercel/python": "workspace:3.1.34",
+    "@vercel/redwood": "workspace:1.0.45",
+    "@vercel/remix": "workspace:1.1.7",
+    "@vercel/routing-utils": "workspace:2.1.3",
```
styfle: I think `@vercel/routing-utils` should be in devDependencies right? Since it wasn't there before?

packages/cli/package.json
```diff
@@ -41,16 +41,19 @@
...
+    "@vercel/ruby": "workspace:1.3.50",
+    "@vercel/static-build": "workspace:1.1.1",
+    "qs": "6.11.0",
```
styfle: Where is `qs` used?

packages/cli/package.json
```diff
@@ -41,16 +41,19 @@
...
+    "update-notifier": "5.1.0"
```
styfle:
```suggestion
```

@Ethan-Arrowood Pretty sure we just removed `update-notifier`. 

cc @cb1kenobi

packages/edge/package.json
```diff
@@ -8,17 +8,17 @@
...
     "@edge-runtime/jest-environment": "2.0.0",
     "@types/jest": "27.4.1",
     "ts-node": "8.9.1",
     "tsup": "6.1.2",
-    "typedoc": "0.23.10",
-    "typedoc-plugin-markdown": "3.13.4",
-    "typedoc-plugin-mdn-links": "2.0.0",
+    "typedoc": "0.23.24",
+    "typedoc-plugin-markdown": "3.14.0",
+    "typedoc-plugin-mdn-links": "2.0.2",
```
styfle: Any reason why these deps are updating?

packages/cli/package.json
Ethan-Arrowood: https://github.com/vercel/vercel/blob/main/packages/cli/src/util/git/connect-git-provider.ts#L2

packages/edge/package.json
Ethan-Arrowood: Trying to fix the edge docs issue. I can revert now that I have the replacer script

packages/next/package.json
```diff
@@ -7,11 +7,11 @@
   "scripts": {
     "build": "node build.js",
     "build-dev": "node build.js --dev",
-    "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "yarn test test/unit/",
-    "test-next-local": "jest --env node --verbose --bail --testTimeout=360000 test/integration/*.test.js test/integration/*.test.ts",
-    "test-next-local:middleware": "jest --env node --verbose --bail --useStderr --testTimeout=360000 test/integration/middleware.test.ts",
```
styfle, line 13: Why was `useStderr` removed?

Ethan-Arrowood: these tests get passed `-- --listTests` by the chunk script and thus can't have any other arguments already in there. they seem to still run and pass fine without it so we went with it

packages/cli/package.json
styfle: :oof: ok we need to remove that in a follow up PR

EndangeredMassa (approved):

package.json
```diff
@@ -3,53 +3,51 @@
...
-      "**/@types/**",
-      "**/typedoc",
-      "**/typedoc-plugin-markdown",
-      "**/typedoc-plugin-mdn-links"
-    ]
-  },
+  "packageManager": "pnpm@7.24.2",
   "dependencies": {
-    "lerna": "3.16.4"
+    "lerna": "5.6.2"
```
styfle, line 8: Any reason why this version of lerna was chosen?

The latest is 6.4.0 https://www.npmjs.com/package/lerna

Ethan-Arrowood: 6 uses Nx lol didn't want to muddy with Turbo. But we need a different major than 3 for `workspace:` support

package.json
```diff
@@ -3,53 +3,51 @@
...
   },
   "devDependencies": {
+    "@types/node": "14.18.33",
     "@typescript-eslint/eslint-plugin": "5.21.0",
     "@typescript-eslint/parser": "5.21.0",
+    "@vercel/build-utils": "workspace:5.7.5",
+    "@vercel/fun": "1.0.4",
```
styfle: I think `@vercel/fun` should be in `packages/cli/package.json` instead

package.json
```diff
@@ -3,53 +3,51 @@
...
+    "@vercel/ncc": "0.24.0",
+    "@vercel/next": "workspace:3.3.9",
```
styfle, line 16: Curious why `@vercel/next` is here. Where is it used?

package.json
Ethan-Arrowood: This chunk somehow needs it at the root level: 

https://github.com/vercel/vercel/blob/4c3bc0532216f21f354623ada76b2315ef8ba519/packages/cli/scripts/build.ts#L69-L76

(Remove, run install, run build, and you'll see the full error)

package.json
Ethan-Arrowood: Ahh this ones tricky. The `@vercel/next` test suite invokes the `test/lib/run-build-lambda.js` which does some funky `require(build.use)` call. `build.use` returns `'@vercel/next'` and so it has to be included in the root devDep list

TooTallNate (approved):

package.json
styfle: @Ethan-Arrowood Pretty sure that path was written that way because of yarn hoisting. So you can move `@vercel/fun` to the cli package and then update the `build.ts` to read the new path from `./node_modules/@vercel/fun/dist/src/runtimes` I believe.

@TooTallNate Do you agree?

package.json
styfle: Oh interesting. Perhaps we should have this script change directory into the package running it or update `use` to be the path to the current package.

@ijjk Do you have thoughts on this one?

https://github.com/vercel/vercel/blob/5b88f673f89c44b9e91e44247e29be6bed43805a/test/lib/run-build-lambda.js#L35

package.json
Ethan-Arrowood: ill give that a shot

Ethan-Arrowood: seems to build fine!

package.json
Ethan-Arrowood: this modification feels like scope creep to me. its more of a "update `@vercel/next` tests" kinda change

styfle: You're right. I just noticed its use for both `@vercel/next` and `@vercel/static-build` so lets save it for a future PR.

styfle (approved):
Great work, thanks! 🎉

javivelasco (approved):

PR #8483, opened 2022-08-29
[build-utils] Replace `npm bin` with custom implementation

styfle:
This PR removes `npm bin` in favor of a custom implementation since the command will be removed in npm 9.

- Related to https://github.com/npm/statusboard/issues/537

PR #9200, opened 2023-01-10
[build-utils] Removed unused execAsync() function

cb1kenobi:
This function can sometimes give a TypeScript error if @types/node says 'code' should be a `number | null` and yet the code is returns as a `number`:

```
@vercel/build-utils:build: src/fs/run-user-scripts.ts(140,13): error TS2322: Type 'number \| null' is not assignable to type 'number'.
@vercel/build-utils:build:   Type 'null' is not assignable to type 'number'.
```

I'm not sure if there are any other projects that depend on `@vercel/build-utils`, but it doesn't appear that the Vercel CLI nor the build container use this `execAsync()` function. If I'm mistaken, feel free to close this PR.

As this removal is a breaking change, setting the semver to major and not auto-merging unless others are in agreement.


styfle (changes requested):
Please do a search across all 3rd party community builders since removing public apis could break those.

cb1kenobi:
I searched all 52 dependents (https://www.npmjs.com/browse/depended/@vercel/build-utils) of `@vercel/build-utils` and none of them appear to use `execAsync()`.

@styfle Is there a list of all 3rd party community builders?

styfle:
@cb1kenobi I sent you a message in slack last week that should give a good idea of builders that might be using this. Did you find any usages?

cb1kenobi:
@styfle thanks for the follow up. I was working on a script that downloads the packages, but got sidetracked. I'll get back on it so that we can verify if this affects any of the builders.

styfle:
@cb1kenobi It looks like this function was used by getNpmBin() however that was removed in #8483

I also considered using it for #9289 but removed it in favor of execa pkg

cb1kenobi:
Upon investigation, our analytics data shows there are 107 builders. Several of them are not valid builders such as `"bash"` and `"start"` packages.

Of the ones that are valid, only 6 actually reference `execAsync()`. 5 of them are reference the definition of this function from within a bundle and 1 of them just happens declare `const execAsync = promisify(exec);`.

Here's the grep output:

```
$ grep -R "execAsync(" .
./@nuxtjs/vercel-builder/0.24.0/lib/index.js:function execAsync(command, args, opts = {}) {
./@nuxtjs/vercel-builder/0.23.0/lib/index.js:function execAsync(command, args, opts = {}) {
./@now/next/2.3.18/dist/index.js:function execAsync(command, args, opts = {}) {
./@now/next/2.3.18/dist/dev-server.js:function execAsync(command, args, opts = {}) {
./@now/node/1.4.1/dist/index.js:function execAsync(command, args, opts = {}) {
./@vercel/edge/0.2.6/test/docs.test.ts:  await execAsync(`pnpm build:docs`, { cwd });
./@vercel/edge/0.2.6/test/docs.test.ts:  const result = await execAsync(`git status --short docs`, {
./@vercel/edge/0.2.6/test/docs.test.ts:    const diff = await execAsync(`git diff docs`, { cwd, encoding: 'utf8' });
./@vercel/build-utils/5.9.0/dist/index.js:function execAsync(command, args, opts = {}) {
./@vercel/build-utils/5.9.0/dist/fs/run-user-scripts.d.ts:export declare function execAsync(command: string, args: string[], opts?: SpawnOptionsExtended): Promise<{
./@vercel/build-utils/5.9.0/dist/fs/run-user-scripts.js:function execAsync(command, args, opts = {}) {
```

It's pretty safe to say that this function is not used and can be removed.

styfle (approved):
Thanks for doing the research here 👍

EndangeredMassa (approved):

PR #9201, opened 2023-01-10
[build-utils] Support directory entries in `Lambda#createZip()`

TooTallNate:
Adds support for empty directory entries in the Lambda `createZip()` function.

### DIFF
--- a/packages/build-utils/test/unit.lambda.test.ts
+++ b/packages/build-utils/test/unit.lambda.test.ts
@@ -1,0 +1,89 @@
+import path from 'path';
+import { tmpdir } from 'os';
+import fs from 'fs-extra';
+import { createZip } from '../src/lambda';
+import { FileBlob, glob, spawnAsync } from '../src';
+
+describe('Lambda', () => {
+  it('should create zip file with symlinks', async () => {
+    if (process.platform === 'win32') {
+      console.log('Skipping test on windows');
+      return;
+    }
+    const files = await glob('**', path.join(__dirname, 'symlinks'));
+    expect(Object.keys(files)).toHaveLength(4);
+
+    const outFile = path.join(__dirname, 'symlinks.zip');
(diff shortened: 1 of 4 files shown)


packages/build-utils/test/unit.lambda.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    expect(linkStat.isSymbolicLink()).toEqual(true);
+    expect(linkDirStat.isSymbolicLink()).toEqual(true);
+    expect(aStat.isFile()).toEqual(true);
+  });
+
+  it('should create zip file with empty directory', async () => {
+    if (process.platform === 'win32') {
+      console.log('Skipping test on windows');
+      return;
+    }
```
EndangeredMassa, line 43: **suggestion (non-blocking):** remove windows guard

Can this run on Windows?

packages/build-utils/test/unit.lambda.test.ts
```diff
@@ -0,0 +1,94 @@
...
+          data: 'inside dir b',
+          mode: 33188,
+        }),
+        c: new FileBlob({
+          data: '',
+          mode: 16877,
+        }),
+        'c/a': new FileBlob({
+          data: 'inside dir c',
+          mode: 33188,
```
EndangeredMassa: **suggestion (non-blocking):** label mode values

Can we label these mode values (with comments or named variables) to make it clear what each one means?

EndangeredMassa (approved):
Can you explain the how this compares to the [the previous PR](https://github.com/vercel/vercel/pull/9103)?

packages/build-utils/test/unit.lambda.test.ts
TooTallNate: No cause the test relies on the GNU `unzip` command.

PR #9203, opened 2023-01-10
Implement potentiallyLongRunningResponse helper in `@vercel/edge`

cramforce:
This helper produces an immediate response, which writes the data once the given promise is resolved.

### DIFF
--- a/packages/edge/src/response.ts
+++ b/packages/edge/src/response.ts
@@ -25,7 +25,7 @@ export function json(data: any, init?: ResponseInit): Response {
  * Builds a response for returning data based on promise that take many seconds to resolve.
  * The response is returned immediately, but data is only written to it when the promise resolves.
  *
- * @param dataPromise Promise for data. Supported types of data are string and Uint Uint8Array
+ * @param dataPromise Promise for data to be sent as the response body
  * @param init optional custom response status, statusText and headers
  *
  * @example
--- a/packages/edge/src/response.ts
+++ b/packages/edge/src/response.ts
@@ -20,3 +20,41 @@ export function json(data: any, init?: ResponseInit): Response {
   // @ts-expect-error This is not in lib/dom right now, and we can't augment it.
   return Response.json(data, init);
 }
+
+/**
+ * Builds a response for returning data based on promise that take many seconds to resolve.
+ * The response is returned immediately, but data is only written to it when the promise resolves.
+ *
+ * @param dataPromise Promise for data. Supported types of data are string and Uint Uint8Array
+ * @param init optional custom response status, statusText and headers
+ *
+ * @example
+ * ```ts
+ * import { potentiallyLongRunningResponse } from '@vercel/edge';
+ *
+ * export default () => {
+ *   const slowPromise = new Promise((resolve) => setTimeout(() => resolve("Done"), 20000));
+ *   return potentiallyLongRunningResponse(slowPromise);
+ * };
+ * ```
+ */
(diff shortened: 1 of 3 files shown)


packages/edge/src/response.ts
```diff
@@ -20,3 +20,41 @@ export function json(data: any, init?: ResponseInit): Response {
   // @ts-expect-error This is not in lib/dom right now, and we can't augment it.
   return Response.json(data, init);
 }
+
+/**
+ * Builds a response for returning data based on promise that take many seconds to resolve.
+ * The response is returned immediately, but data is only written to it when the promise resolves.
+ *
+ * @param dataPromise Promise for data. Supported types of data are string and Uint Uint8Array
```
Schniz: maybe we don't need to tell the data types? docs are generated from typescript so they have a `Promise<string | Uint8Array>` mentioned in them automatically 😄 

```suggestion
 * @param dataPromise Promise for data to be sent as the response body
```

cramforce:
Added error handling

packages/edge/src/response.ts
```diff
@@ -20,3 +20,51 @@ export function json(data: any, init?: ResponseInit): Response {
...
+            } else {
+              controller.enqueue(data);
+            }
+            controller.close();
+          })
+          .catch(error => {
+            console.log(
+              `Error in 'potentiallyLongRunningResponse' dataPromise: ${error}`
+            );
+            controller.enqueue(new TextEncoder().encode('ERROR'));
```
feugy, line 63: **suggestion:** shouldn't we return the error text instead of a fixed string?

packages/edge/test/response.test.ts
```diff
@@ -32,3 +32,50 @@ describe('json', () => {
...
+        status: 400,
+        headers: {
+          'content-type': 'text/custom',
+        },
+      }
+    );
+    expect(await response.text()).toBe('test');
+    expect(response.status).toBe(400);
+    expect(response.headers.get('content-type')).toBe('text/custom');
+  });
```
feugy, line 65: **praise:** great testing here. I was wondering whether we could return JSON content, and this test proves that user can still tweak the response headers.

packages/edge/src/response.ts
cramforce: The normal behavior of a last resort error handler is to return a 500 without a specific error message (which might leak stuff). If folks want something better they need to handle the error themselves.

packages/edge/test/response.test.ts
cramforce: I don't follow

Schniz: This is just Damien being positive in code reviews <picture data-single-emoji=":lol:" title=":lol:">:lol:</picture> <picture data-single-emoji=":sunglasses_+1:" title=":sunglasses_+1:">:sunglasses_+1:</picture>

packages/edge/src/response.ts
Schniz: > If folks want something better they need to handle the error themselves

<picture data-single-emoji=":sgtm:" title=":sgtm:">:sgtm:</picture>

If only we had a typed error in Promse in TypeScript (`Promise<Ok, Err>`), we could ask for a `Promise<Ok, never>` in this function to explain this using types <picture data-single-emoji=":wen:" title=":wen:">:wen:</picture> <picture data-single-emoji=":p0:" title=":p0:">:p0:</picture>

PR #9205, opened 2023-01-12
[python] Cope with duplicate header values

### LINKED ISSUE #9132: Python ASGI functions crashing on incoming headers
## Issue

When deploying a Python ASGI application (e.g. using FastAPI, Sanic) with Vercel Serverless, the function is built successfully but invoking the URL the function leads to a crash with a 500 error.

Here is the backend log from Vercel:
```
[ERROR] AttributeError: 'list' object has no attribute 'encode'
  File "/var/task/vc__handler__python.py", line 288, in vc_handler
      'headers': [[k.lower().encode(), v.encode()] for k, v in headers.items()],
  File "/var/task/vc__handler__python.py", line 288, in <listcomp>
      'headers': [[k.lower().encode(), v.encode()] for k, v in headers.items()],
```

## Investigations

It seems there is an issue with incoming complex headers parsing for ASGI functions:

- Python WSGI applications (e.g. using Flask) do not show this error
- Invocations of ASGI functions using `curl` or `httpie` works perfectly, no errors
- Bisecting headers from the web browser (Firefox, Chrome), here is a faulty header causing the crash:
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
```
- Invocations of ASGI functions using `curl` or `httpie` using the above header are triggering the error

## Reproduction

A simple hello world app with FastAPI or Sanic in a module name `app.py`:

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
def index():
    return {"message": "Hello World"}
```

```python
from sanic import Sanic
from sanic.response import json

app = Sanic(name=__name__)

@app.route("/")
async def index(request):
    return json({"message": "Hello World"})
```

`vercel.json`:
```
{
  "builds": [
    {
      "src": "app.py",
      "use": "@vercel/python"
    }
  ],
  "routes": [
    {
      "src": "/(.*)",
      "dest": "app.py"
    }
  ]
}
```

## Related issues

- https://github.com/vercel/vercel/issues/2526

andymckay:
There are times when a request can arrive for a Python function with headers as a list. One of those examples is this header `x-vercel-proxied-for` which apparently is set twice. Example:

`[b'x-vercel-proxied-for', [b'207.81.134.243', b'172.71.147.74']]`

I took a quick scan through the other Python server implementations and I don't think any of them manipulate the value of the HTTP headers, the way the ASGI one does so I think we are good there.

To reproduce:

`curl https://..../ -H "foo: bar" -H "foo: bar"`

Will fail.

Fixes: https://github.com/vercel/vercel/issues/9132


andymckay:
This works on production, but I'd like to see why the tests are passing since there's specifically a test for `multivalue-header` that tests ASGI.

styfle (commented):
Please add a test fixture to `./packages/python/test/fixtures/`

andymckay:
> Please add a test fixture to `./packages/python/test/fixtures/`

I currently cannot recreate a test failure to write a meaningful test, even though `./packages/python/test/fixtures/20-multivalue-header/cookie-asgi.py` claims to test this exact case.

TooTallNate (approved):

andymckay:
Trying to add in some tests, not that using a `fetch` does not reproduce the problem whereas `curl` can.

packages/python/test/fixtures/32-fail-duplicate-header/probe.js
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = async function ({ deploymentUrl, fetch }) {
+  const probeUrl = `https://${deploymentUrl}`;
+  const result = await execa('curl', [
+    probeUrl,
+    '-s',
+    '-H',
+    'foo: bar',
+    '-H',
+    'foo: bar',
```
styfle, line 11: Does the issue only reproduce with `curl` or can you also reproduce with `fetch`?

styfle: Oh yeah because `fetch` can't set multiple headers, it combines them into a single key with CSV value 🤦

styfle: Another option is to try with `require('https')` instead of `fetch` because I think that supports multiple headers

andymckay: I tried a couple to no avail, so far. For example `https` requires an object as headers, it then iterates through them and adds the headers in. But you can't specify multiple keys in a object, JS just uses the last one defined. I tried setting it to a list as in `'a': ['a', 'b']` and that didn't fail. So far only `curl` has been able to easily send something like:

```
> user-agent: curl/7.79.1
> accept: */*
> foo: bar
> foo: bar
```

Although technically I could start diving further into one of the http implementations to get it to set headers like that.

styfle (approved):

cb1kenobi (approved):

styfle (approved):

styfle (approved):

PR #9206, opened 2023-01-12
[cli] Remove `qs` dependency

styfle:
There is no need for `qs` because the query string for `teamId` is already handled here:

https://github.com/vercel/vercel/blob/5b88f673f89c44b9e91e44247e29be6bed43805a/packages/cli/src/util/client.ts#L99

### DIFF
--- a/packages/cli/src/commands/git/index.ts
+++ b/packages/cli/src/commands/git/index.ts
@@ -34,7 +34,7 @@ const help = () => {
   )} Connect your Vercel Project to your Git repository defined in your local .git config
 
     ${chalk.cyan(`$ ${getPkgName()} git connect`)}
-  
+
   ${chalk.gray(
     '–'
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/git/index.ts
```diff
@@ -96,6 +96,7 @@ export default async function main(client: Client) {
   }
 
   const { org, project } = linkedProject;
+  client.config.currentTeam = org.type === 'team' ? org.id : undefined;
```
TooTallNate, line 99: Is this change related?

styfle, line 99: @TooTallNate Yes, this is used for all client fetch calls here

https://github.com/vercel/vercel/blob/5b88f673f89c44b9e91e44247e29be6bed43805a/packages/cli/src/util/client.ts#L99

TooTallNate (approved):

PR #9216, opened 2023-01-12
[fs-detectors] Add support for `api/**/*.tsx` zero config detection

styfle:
There are cases where you would want to use `.tsx` with Serverless Functions, such as OG Image Generation.

This PR adds zero config detection for `.tsx` file extensions and also consolidates the glob pattern into a single string matching all valid `@vercel/node` cases.

https://twitter.com/hasparus/status/1593136849404694528

https://linear.app/vercel/issue/VCCLI-322


styfle:
Converting to a draft for now because there must be changes to `@vercel/node` in order to support `.jsx`.

Also, not sure if we should default to a `tsconfig.json` and somehow install react, or assume the user configured their project correctly already?

styfle:
I think we shouldn't support `.jsx` for zero config because its not clear what it should output. Should it emit react, or preact, or something else? The default react behavior won't even work with `@vercel/og`.

With `.tsx`, it makes more sense because you can configure `"jsx": "react"` in `tsconfig.json` or even in a comment.

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -453,9 +453,7 @@ function getApiMatches() {
 
   return [
     { src: 'middleware.[jt]s', use: `@vercel/node`, config },
-    { src: 'api/**/*.js', use: `@vercel/node`, config },
-    { src: 'api/**/*.mjs', use: `@vercel/node`, config },
-    { src: 'api/**/*.ts', use: `@vercel/node`, config },
+    { src: 'api/**/*.+(js|mjs|ts|tsx)', use: `@vercel/node`, config },
```
cb1kenobi, line 456: It looks like `src` is a glob pattern. Is `+` supported in our system because it doesn't appear to be valid in bash:

image

```suggestion
    { src: 'api/**/*.(js|mjs|ts|tsx)', use: `@vercel/node`, config },
```

styfle: @cb1kenobi As you can see the tests are passing. This is because we don't use bash, we use `minimatch`

https://github.com/isaacs/minimatch#usage

cb1kenobi: Ah, it's an extglob. I saw the tests were passing, just didn't see it in the code. OK, cool, looks good!

cb1kenobi (approved):

PR #9222, opened 2023-01-13
[client] Add sliding window delay when polling deployment complete

styfle:
The previous delay of 1500ms was causing some users to hit the API rate limits. This doesn't normally happen with a single deployment, but it can happen with several concurrent deployments (for example a monorepo with many projects).

We don't need to be polling so often, so this PR changed the polling delay to the following:

- During 0s-15s: check every 1 second
- During 15s-60s: check every 5 seconds
- During 1m-5m: check every 15 seconds
- During 5m-10m: check every 30 seconds


packages/client/src/check-deployment-status.ts
```diff
@@ -155,6 +156,9 @@ export async function* checkDeploymentStatus(
       };
     }
 
-    await sleep(ms('1.5s'));
+    const elapsed = Date.now() - startTime;
+    const duration = getPollingDelay(elapsed);
+    console.log('sleeping for', duration, 'ms'); // TODO: remove
```
styfle:
```suggestion
```

Ethan-Arrowood (approved):

cb1kenobi (approved):

PR #9223, opened 2023-01-13
[tests] adjust root and api deps to use worskpace:*

Ethan-Arrowood:
Fixes error after lerna publish

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/cli/package.json
```

https://github.com/vercel/vercel/actions/runs/3914506649/jobs/6691645690


EndangeredMassa (approved):

TooTallNate (approved):

cb1kenobi (approved):

styfle (approved):
Thanks!

We should try to remove those dependencies from the root package.json in a future PR

PR #9227, opened 2023-01-13
[cli] Revert some tables back to stdout

andymckay:
In https://github.com/vercel/vercel/pull/8735 some output was sent to `stderr`, this sends them `stdout` instead.


packages/cli/src/commands/alias/ls.ts
```diff
@@ -51,7 +51,7 @@ export default async function ls(
     ...paginationOptions
   );
   output.log(`aliases found under ${chalk.bold(contextName)} ${lsStamp()}`);
-  output.log(printAliasTable(aliases));
+  client.stdout.write(printAliasTable(aliases));
```
styfle, line 54: I think this is missing a `\n`

Have you considered using `console.log()` instead?

TooTallNate: Agree, we probably want to add a `\n`. However we don't want to use `console.log()` so that we can assert in the tests.

andymckay: In my tests it has a newline at the end, just not two newlines. Happy to add in `\n` but I'm not sure if we need two newlines?

Existing:

```bash
➜  vercel git:(andymckay/back-to-stdout) ✗ cat -e stdout.txt
  source                                                     url                                                        age$
...[snip]
  15-binary-handler-bbp1gxgo8-andymckay.vercel.app           15-binary-handler.vercel.app                                6d$
  14-unicode-handler-f7eg54s4k-andymckay.vercel.app          14-unicode-handler-andymckay.vercel.app                     6d$
$
$
```

New:

```bash
➜  vercel git:(andymckay/back-to-stdout) ✗ cat -e stdout-dev.txt
  source                                                     url                                                        age$
...[snip]
  15-binary-handler-bbp1gxgo8-andymckay.vercel.app           15-binary-handler.vercel.app                                6d$
  14-unicode-handler-f7eg54s4k-andymckay.vercel.app          14-unicode-handler-andymckay.vercel.app                     6d$
$
```

cb1kenobi: There's no need for a `\n` since `printAliasTable()` already appends and returns `\n\n`: https://github.com/vercel/vercel/pull/9227/files#diff-f60b8447303c4253e468e88965c478eaee1a8c04feaa48e371c98ceb78d87c27R86.

cb1kenobi (approved):

mglagola (approved):

styfle (approved):

PR #9245, opened 2023-01-17
[build-utils] Add `includeDirectories` option to `glob()`

TooTallNate:
Instead of always including empty directories in `glob()`, make it an opt-in behavior because technically it's a breaking change to include them by default.


cb1kenobi (approved):

styfle (approved):

styfle:
I'm going to force merge since it seems some of the e2e tests are affected by this bug

PR #9246, opened 2023-01-18
[cli] Fix pagination next when using `--limit`

andymckay:
When using `--limit` the helpful next text was not being displayed.

This text:

```
➜  status-via-edge git:(main) ✗ vc alias ls
Vercel CLI 28.8.0

aliases found under andymckay [343ms]
source                                              url                                                age
test-some-files-i51lqtzbf-andymckay.vercel.app      test-some-files-andymckay.vercel.app                5d
….

To display the next page run vercel alias ls --next 1669404182512
```

The reason is the check for a hard coded `20`. Instead this sets the default in `getPaginationOptions` to `20` and then checks the value in that. The result is that it will show the next text if you hit the default `20` or whatever you specified in `limit`

This PR is quite a bit bigger than planned, because of the change to use an object for params, so we could use spread operators as suggested in this comment here: https://github.com/vercel/vercel/pull/9102#discussion_r1063579957

This is a second PR from https://github.com/vercel/vercel/pull/9102, because getting that landed involved a rebase and that rebase had so many problems I started to hate my life. It was quicker to make a new branch and cherry pick over.


packages/cli/src/util/dns/get-domain-dns-records.ts
```diff
@@ -8,29 +8,32 @@ type Response = {
...
-  limit = 20
+  args: getDomainDNSRecordsArgs
 ) {
-  output.debug(`Fetching for DNS records of domain ${domain}`);
+  args.output.debug(`Fetching for DNS records of domain ${args.domain}`);
   try {
-    let url = `/v${apiVersion}/domains/${encodeURIComponent(
-      domain
-    )}/records?limit=${limit}`;
+    let url = `/v4/domains/${encodeURIComponent(args.domain)}/records?limit=${
```
andymckay, line 24: Note: that the entry point has the domains set to `v4` here: https://github.com/vercel/vercel/pull/9246/files#diff-91299e67aed9eb69236ea2a90f38d7a8c9a55d45bdf9fb9dfd7d3d8544de93d8L54. But if not set, goes back to `v3` which seems odd. Unless there's a good reason too I think its easier to not have to pass around the `apiVersion` and pin it to a version.

EndangeredMassa:
@styfle may take over this PR and https://github.com/vercel/front/pull/18134

PR #8793, opened 2022-10-27
[remix] add support for `.cjs` and `.mjs` extensions for remix.config

mcansh:
adds support for `remix.config.mjs` and `remix.config.cjs` and also updates the example/fixtures to the latest version of remix

https://github.com/remix-run/remix/pull/3675

### Related Issues

> Fixes #1
> Related to #2

#### Tests

- [x] The code changed/added as part of this PR has been covered with tests
- [x] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #9249, opened 2023-01-18
[remix] Fix config file dynamic import

TooTallNate:
Follow-up to https://github.com/vercel/vercel/pull/8793, which breaks dynamic import of the config file, causing some issues for remix users: https://github.com/orgs/vercel/discussions/1282

The underlying error is `MODULE_NOT_FOUND`.

`import()` should work with `file://` URIs, however, since it's using typescript, the `import` gets compiled to `require()`, which does not support `file://` protocol scheme.

Supersedes https://github.com/vercel/vercel/pull/9248.


packages/remix/package.json
```diff
@@ -26,6 +25,6 @@
     "@types/jest": "27.5.1",
     "@types/node": "14.18.33",
     "@vercel/build-utils": "5.9.0",
-    "typescript": "4.6.4"
+    "typescript": "4.9.4"
```
TooTallNate, line 28: Thought this was necessary at first. This isn't the case, but doesn't hurt either.

packages/remix/package.json
```diff
@@ -12,8 +12,7 @@
   "scripts": {
     "build": "node build.js",
     "test-integration-once": "pnpm test test/test.js",
-    "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "pnpm test test/build.test.ts"
```
TooTallNate, line 16: The unit tests were deleted since they test basically the same as the integration tests. However, `await import()` breaks Jest currently, so they needed to be removed.

packages/remix/src/build.ts
```diff
@@ -169,9 +168,7 @@ export const build: BuildV2 = async ({
 
   try {
     if (remixConfigFile) {
-      const remixConfigModule = await import(
-        pathToFileURL(remixConfigFile).href
-      );
+      const remixConfigModule = await eval('import(remixConfigFile)');
```
TooTallNate, line 171: `eval` is required to prevent TypeScript from compiling the import into require.

swarnava (approved):

packages/remix/src/build.ts
styfle: I would be surprised if this works since I didn’t think eval works for ESM

styfle (approved):
Can you add a test?

packages/remix/src/build.ts
styfle: A better solution is to update tsconfig.json with `module: nodenext`

TooTallNate:
The existing integration test should already be testing for this. I don't really understand how it was passing before, but possibly just a caching issue 😕

PR #9255, opened 2023-01-18
[next] Fix normalize route with middleware and basePath

ijjk:
This ensures we properly match the index route when normalizing routes with `x-nextjs-data` for middleware. 

x-ref: [slack thread](https://vercel.slack.com/archives/C03S8ED1DKM/p1674034533265989)

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1060,9 +1060,9 @@ export async function serverBuild({
           {
             src: path.posix.join(
               '^/',
-              entryDirectory,
-              trailingSlash ? '/' : '',
-              '$'
+              entryDirectory !== '.'
+                ? `${entryDirectory}${trailingSlash ? '/$' : '$'}`
+                : '$'
             ),
             has: [
               {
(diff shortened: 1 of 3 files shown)


packages/next/src/server-build.ts
```diff
@@ -1060,9 +1060,9 @@ export async function serverBuild({
           {
             src: path.posix.join(
```
cb1kenobi, line 1061: This is a little complicated looking with the inline ternary condition. It reads clearer for me if the `join()` is completely skipped if the `entryDirectory` is `'.'`. I realize this deviates from the pattern used elsewhere.

```suggestion
            src: entryDirectory === '.'
              ? '^/$'
              : path.posix.join('^', entryDirectory, '$'),
```

cb1kenobi (changes requested):

packages/next/src/server-build.ts
ijjk: This re-introduces the bug, the ternary is still necessary for the trailing slash handling.

cb1kenobi: Oh, I see now. Sometimes you want the trailing slash and sometimes you don't.

cb1kenobi (approved):
LGTM

PR #9256, opened 2023-01-18
[tests] verify package before publish

EndangeredMassa:
In order to make sure that the package is installable, this PR adds a "Verify Package" step before publishing.

### DIFF
--- a/utils/verify-package.sh
+++ b/utils/verify-package.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+# everything merged to 'main' is published to this URL
+packageUrl="https://api-frameworks.zeit.sh/tarballs/vercel.tgz"
+
+
+# make a temp dir (works on mac and linux)
+tmpDir=$(mktemp -d 2>/dev/null || mktemp -d -t 'tmpDir')
+cd $tmpDir
+echo ""
+echo "Testing installs in: $tmpDir"
+
+
+# Test 'npm' Install
+echo ""
+echo "Verifying: 'npm install $packageUrl --no-save'"
+npm install $packageUrl --no-save
+status=$?
+if [ $? -ne 0 ]
+then
+  echo "ERROR: 'npm install' failed with exit code $status"
+  cd -
+  exit $status
+fi
+rm -rf node_modules
+
+
+# Test 'pnpm' Install
+echo ""
+echo "Verifying: 'pnpm add $packageUrl'"
+pnpm add $packageUrl
+status=$?
+if [ $? -ne 0 ]
+then
+  echo "ERROR: 'pnpm add' failed with exit code $status"
+  cd -
+  exit $status
+fi
+rm -rf node_modules pnpm-lock.yaml package.json
+
+
+# Test 'npm' Install
+echo ""
+echo "Verifying: 'yarn add $packageUrl'"
+yarn add $packageUrl
+status=$?
+if [ $? -ne 0 ]
+then
+  echo "ERROR: 'yarn add' failed with exit code $status"
+  cd -
+  exit $status
+fi
+rm -rf node_modules yarn.lock package.json
+
+
+cd -
+rm -rf $tmpDir
+echo ""
+echo "Package Verified!"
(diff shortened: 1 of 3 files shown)


utils/verify-package.sh
```diff
@@ -0,0 +1,60 @@
...
+  cd -
+  exit $status
+fi
+rm -rf node_modules yarn.lock package.json
+
+
+cd -
+rm -rf $tmpDir
+echo ""
+echo "Package Verified!"
```
EndangeredMassa, line 60: I tested this manually. It seems to work as we need!

Ethan-Arrowood (approved):

utils/verify-package.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+# everything merged to 'main' is published to this URL
+packageUrl="https://api-frameworks.zeit.sh/tarballs/vercel.tgz"
```
styfle, line 5: I think we need to wait for the url from the specific commit because this script might run before the Vercel deployment completes. Unless that is intentional?

EndangeredMassa: Ah, sounds like we could really use: https://linear.app/vercel/issue/VCCLI-261/create-cli-command-to-wait-for-deployment

styfle: @EndangeredMassa We have that already

https://github.com/vercel/vercel/blob/aaceeef604ffe432af15d5f8f315c8ebc5db98f5/.github/workflows/test.yml#L23-L27

TooTallNate, line 5: We'd just have to be careful to make sure latest main has the Vercel deployment completed already. This will probably usually be the case already, but something to keep in mind.

An alternative would be to wait for the Vercel deployment of the associated commit to complete and then use the unique URL from that, but that would slow down the Publish step quite a bit.

styfle: I'm actually not even sure if we can catch this category of bug because we don't know what `lerna` will do. So we kinda have to test against the published package.

Perhaps a better solution is to update CI to include a rollback mechanism.

Basically `npm publish`, `npm install vercel@28.0.2`, then `npm dist-tag vercel@28.0.1` to previous version if that failed to install.

EndangeredMassa: As is, this PR is more protection than we had before. I think it's mergable, but we should keep in mind that the latest main CI has to have passed (and that deployment needs to be ready) before we can publish with certainty.

We can revise this by waiting for that deployment after making the CLI command for it.

EndangeredMassa: Indeed. Styfle had the same thought here: https://github.com/vercel/vercel/pull/9256#discussion_r1080572275

styfle: How will we know which tarball was tested? Might give a false sense of security right now.

EndangeredMassa: Ah, I thought the issue existed before the actual publish, but that wasn't true. If I test `npm install` on the tarball we create vs. the published version of `vercel@28.12.3`, the issue only appears in the published version.

```
$  npm i -g https://vercel-ibsxg4abt.vercel.sh/tarballs/vercel.tgz

changed 150 packages, and audited 151 packages in 5s

13 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

~ took 5s
```

```
$  npm i -g vercel@28.12.3
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:5.8.2

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/smassa/.npm/_logs/2023-01-18T22_29_03_860Z-debug-0.log
```

So yes, we could implement something like the current PR, but it's not going to protect us from the specific issue we're targeting.

---

I like the auto-rollback. I'll try that out.

styfle: Great, thanks for verifying my theory 👍 

I do think the tarball test you wrote could be somewhat useful for running against every `main` commit and maybe catch a different class of bug earlier (perhaps turbo or ncc misbehaving)

So it could be pulled out into its own workflow file or added as a step after "Find Changes" but only `if` ref main

PR #9258, opened 2023-01-19
[node] Fix TypeScript built-in compiler log message check

TooTallNate:
This check started breaking probably after the pnpm switch, so switch to a simpler solution that doesn't require comparing the paths.


cb1kenobi (approved):

styfle (commented):
Please add a test

PR #9259, opened 2023-01-19
[static-build] run pnpm install --lockfile-only after injecting gatsby plugins

Ethan-Arrowood:
Fixes a bug where after injecting plugins and modifying the users package.json, pnpm would fail normal install


cb1kenobi (approved):
LGTM

packages/static-build/src/index.ts
```diff
@@ -388,7 +391,20 @@ export const build: BuildV2 = async ({
...
+          entrypointDir
+        );
+
+        if (injectedPlugins) {
+          const packageManager = await detectFramework({
+            fs: localFileSystemDetector,
+            frameworkList: packageManagers,
+          });
+          if (packageManager === 'pnpm') {
+            await execCommand('pnpm install --lockfile-only');
```
styfle, line 405: I think this is okay but I dont think we use detectFramework when running `pnpm install` in build-utils currently. So it might have slightly different detection logic.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -35,35 +35,34 @@ export async function injectPlugins(
...
   if (pluginsToInject.length === 0) {
-    return;
+    return false;
   }
 
   await addGatsbyPackage(dir, pluginsToInject);
 
   const gatsbyConfigPathTs = path.join(dir, `${GATSBY_CONFIG_FILE}.ts`);
+  const gatsbyConfigPathMjs = path.join(dir, `${GATSBY_CONFIG_FILE}.mjs`);
+  const gatsbyConfigPathJs = path.join(dir, `${GATSBY_CONFIG_FILE}.js`);
```
styfle, line 45: Do we need to consider `.cjs` as well?

packages/static-build/src/utils/gatsby.ts
```diff
@@ -35,35 +35,34 @@ export async function injectPlugins(
...
-        plugins: pluginsToInject,
-      })}`
-    );
+    printInjectingPlugins(pluginsToInject, gatsbyConfigPathJs);
+    if (await fileExists(gatsbyConfigPathJs)) {
+      await updateGatsbyJsConfig(gatsbyConfigPathJs, pluginsToInject);
+    } else {
+      await fs.writeFile(
+        gatsbyConfigPathJs,
+        `module.exports = ${JSON.stringify({
```
styfle, line 59: Perhaps we should use `.cjs` extension for `module.exports` (or use the `.mjs` with `export default`

Ethan-Arrowood: We'd have to check if gatsby supports the .cjs extension for its config file

packages/static-build/src/utils/gatsby.ts
Ethan-Arrowood: Not that I'm aware of. It's not in gatsby docs at least

packages/static-build/src/utils/gatsby.ts
Ethan-Arrowood: And we do use .mjs with export syntax

styfle: @Ethan-Arrowood What I mean is that `.js` is ambiguous and depends on the customer's `package.json`

I think we should be explicit when writing to a new file.

Ethan-Arrowood: this is a gatsby thing unfortunately https://www.gatsbyjs.com/docs/reference/config-files/gatsby-config/

PR #9261, opened 2023-01-19
[node] Add build logs probe for TypeScript usage messages

TooTallNate:
Follow-up to https://github.com/vercel/vercel/pull/9258, even though that issue seemed to only happen when linked to the monorepo locally. In any case, this test will ensure those log messages are intact for any other change around that part of the codebase in the future.


styfle (approved):
Thanks!

cb1kenobi (approved):

PR #9252, opened 2023-01-18
[static-build] add gatsby builder plugin injection

Ethan-Arrowood:
Modifies the `@vercel/static-build` injection logic in order to now inject both the analytics and builder gatsby plugins.

PR #9262, opened 2023-01-19
[tests] Pass in builder to `runBuildLambda()`

TooTallNate:
Follow-up for [this comment](https://github.com/vercel/vercel/pull/9252#discussion_r1080768761). Basically this just removes the need for `next`/`static-build` to be present in the root `package.json` file.

View with whitespace hidden: https://github.com/vercel/vercel/pull/9262/files?diff=unified&w=1

### DIFF
--- a/packages/next/test/unit/export.test.js
+++ b/packages/next/test/unit/export.test.js
@@ -1,6 +1,0 @@
-describe('export', () => {
-  it('should require by path main', async () => {
-    const main = require('@vercel/next');
-    expect(main).toBeDefined();
-  });
-});
--- a/test/lib/run-build-lambda.js
+++ b/test/lib/run-build-lambda.js
@@ -4,75 +4,56 @@ const fs = require('fs-extra');
 const json5 = require('json5');
 const { glob } = require('@vercel/build-utils');
 
-function runAnalyze(wrapper, context) {
-  if (wrapper.analyze) {
-    return wrapper.analyze(context);
-  }
+exports.createRunBuildLambda = async function (builder) {
+  return async inputPath => {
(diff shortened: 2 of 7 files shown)


packages/next/test/unit/export.test.js
```diff
@@ -1,6 +0,0 @@
-describe('export', () => {
```
TooTallNate, line 1: This test was not very useful. And it started failing because `@vercel/next` was removed from the root level `package.json`.

test/lib/run-build-lambda.js
```diff
@@ -4,75 +4,56 @@ const fs = require('fs-extra');
...
-  }
-  inputFiles[entrypoint].digest =
-    'this-is-a-fake-digest-for-non-default-analyze';
-  const wrapper = require(build.use);
-
-  const analyzeResult = runAnalyze(wrapper, {
-    files: inputFiles,
-    entrypoint,
-    config: build.config,
-  });
```
EndangeredMassa, line 41: **question (non-blocking):** Was `analyzeResult` not being used anywhere?

EndangeredMassa (approved):

test/lib/run-build-lambda.js
TooTallNate: Yes, Builders used to have an `analyze()` function but that's no longer the case.

ijjk (approved):

packages/next/test/unit/export.test.js
styfle: Hmm, I feel like this was added because of a regression but I suppose its not the valuable since no one runs the builders directly anymore so it should be implicitly tested by `vercel build` already 👍

PR #9266, opened 2023-01-19
[static-build] adjust gatsby auto-inject semver range to exclude v5+

Ethan-Arrowood:
Modify the semver range the auto injection operates in so that we only add the plugin for v4 projects


Ethan-Arrowood:
Plugin works with v5 - disregard this change

PR #9269, opened 2023-01-20
[next] Ensure outputDirectory is normalized

ijjk:
This ensures we normalize an absolute `outputDirectory` passed into the Next.js builder as this can cause incorrect resolving. We should probably normalize this higher up as well but this is a starting point. 

Fixes: [slack thread](https://vercel.slack.com/archives/C0289CGVAR2/p1674176040821579?thread_ts=1674168748.257019&cid=C0289CGVAR2)


javivelasco (approved):

cb1kenobi (approved):
Nice!

PR #9275, opened 2023-01-20
[examples] Restrict node version for gridsome example

ofhouse:
Gridsome was not updated for [more than 2 years](https://www.npmjs.com/package/gridsome) and still relies on Webpack 4 what makes it incompatible with >= Node.js 17.
To make the example still deployable on Vercel, this adds an engine restriction to the `package.json` for that example.

Same as #9007.


leerob (approved):
Thank you!

PR #9286, opened 2023-01-23
[node-bridge][node][build-utils] add support for isomorphic signature serverless functions

feugy:
### 📖 What's in there?

Introduces support for web-compliant signatures (aka isomorphic) `(request: Request) => MaybePromise<Response>` signature in serverless functions.

This PR covers prebuilt apps (Build Output API documentation update to come) and Vercel applications (both locally and in production, tested with build-container).

[RFC](https://www.notion.so/vercel/Isomorphic-Functions-part-1-0e89cf379f79445086112f09659bb978#ba806fed7d1b4a828a2b2a82c334251f)

This covers:
- feat(node-bridge): based on configured `launcherType: WinterCG-experimental` property, wraps the user-provided isomorphic function into a node.js compliant function
- feat(node): sets `launcherType` based on explicit runtime configured in the user function (`export const config = { runtime: 'nodejs-experimental' }`), both when building and developing
- feat(node): throws error when developing when executing an isomorphic serverless function on node <18
- feat(node): throws error while building when the targeted node.js version of an isomorphic serverless function is <18

### 🧪 How to test locally

One can set a local vc app and build it locally.
1. build all packages locally `pnpm i && pnpm build`
1. create a new folder with this content:
   ```sh
   // file api/serverless.js:
   export default async function handler() {
     return Response.json({ works: 'just right from serverless' });
   }

   export const config = { runtime: 'nodejs-experimental' }
   ```
1. run it locally with node 18+: `<path to vercel repo>/packages/cli/dist/index.js dev` and browse to localhost:3000/api/serverless
   > it should serve the json payload
1. run it locally with node <18: `<path to vercel repo>/packages/cli/dist/index.js dev` and browse to localhost:3000/api/serverless
   > the request fails and the appropriate message is displayed in terminal
1. build your app: `<path to vercel repo>/packages/cli/dist/index.js build`
   > inspect `vercel/output/functions/api/serverless.func/.vc-config.json`: its launcher type should be `WinterCG-experimental`
1. add a package.json file to your app with `"engines": { "node": "16.x.x" }` and rebuild 
   > build will fails and the appropriate message on incompatibility


packages/node-bridge/build.js
```diff
@@ -4,56 +4,51 @@ const execa = require('execa');
...
     { stdio: 'inherit' }
   );
   await fs.rename(
-    join(sourceMapSupportDir, 'index.js'),
-    join(__dirname, 'source-map-support.js')
+    join(outputDir, 'index.js'),
+    join(__dirname, `${finalName}.js`)
   );
-  await fs.remove(sourceMapSupportDir);
+  console.error(`ncc: Writing ${finalName}.js`);
```
craigandrews: *question* is this intentional or leftover debug?

packages/build-utils/src/nodejs-lambda.ts
```diff
@@ -1,25 +1,28 @@
...
 interface NodejsLambdaOptions extends LambdaOptionsWithFiles {
   shouldAddHelpers: boolean;
   shouldAddSourcemapSupport: boolean;
   awsLambdaHandler?: string;
+  launcherType?: LauncherType;
 }
 
 export class NodejsLambda extends Lambda {
-  launcherType: 'Nodejs';
+  launcherType: NodejsLambdaOptions['launcherType'];
```
craigandrews, line 12: *thought* TIL subscripting interfaces

packages/node-bridge/build.js
feugy: It is intentional. It's using stderr because ncc does so, to make it the build output easier to figure out.
With it:
```sh
﻿❯ pnpm --filter node-bridge build

> @vercel/node-bridge@3.1.10 build /Users/damien/dev/vercel/packages/node-bridge
> node build.js

ncc: Version 0.24.0
ncc: Compiling file index.js
ncc: Using typescript@4.3.4 (local user-provided)
 1kB  helpers/helpers.d.ts
 2kB  helpers/types.d.ts
22kB  helpers/index.js
25kB  [1330ms] - ncc 0.24.0
ncc: Writing helpers.js
ncc: Version 0.24.0
ncc: Compiling file index.js
122kB  source-map-support/index.js
122kB  [244ms] - ncc 0.24.0
ncc: Writing source-map-support.js
ncc: Version 0.24.0
ncc: Compiling file index.js
ncc: Using typescript@4.3.4 (local user-provided)
 0kB  web-handler/web-handler.d.ts
 1kB  web-handler/helpers.d.ts
 2kB  web-handler/types.d.ts
10kB  web-handler/index.js
13kB  [1246ms] - ncc 0.24.0
ncc: Writing web-handler.js
```

Without:
```sh
❯ pnpm --filter node-bridge build

> @vercel/node-bridge@3.1.10 build /Users/damien/dev/vercel/packages/node-bridge
> node build.js

ncc: Version 0.24.0
ncc: Compiling file index.js
ncc: Using typescript@4.3.4 (local user-provided)
 1kB  helpers/helpers.d.ts
 2kB  helpers/types.d.ts
22kB  helpers/index.js
25kB  [1157ms] - ncc 0.24.0
ncc: Version 0.24.0
ncc: Compiling file index.js
122kB  source-map-support/index.js
122kB  [236ms] - ncc 0.24.0
ncc: Version 0.24.0
ncc: Compiling file index.js
ncc: Using typescript@4.3.4 (local user-provided)
 0kB  web-handler/web-handler.d.ts
 1kB  web-handler/helpers.d.ts
 2kB  web-handler/types.d.ts
10kB  web-handler/index.js
13kB  [1252ms] - ncc 0.24.0
```

I don't have strong opinion on this, I can rollback.

craigandrews: leave it, it's useful information :)

craigandrews (dismissed):

packages/build-utils/src/types.ts
```diff
@@ -312,13 +312,16 @@ export interface Builder {
   config?: Config;
 }
 
+export type LauncherType = 'Nodejs' | 'edge-light';
```
styfle: This doesn't follow the convention of PascalCase. I'm thinking it should be `EdgeLight` but not sure. It might even be `WinterCG` since this isnt a runtime but an API signature, right?

@TooTallNate @EndangeredMassa Do you have a preference here?

packages/node/package.json
```diff
@@ -29,6 +29,7 @@
     }
   },
   "dependencies": {
+    "@edge-runtime/feature-detector": "1.0.0",
```
styfle: I think these should be devDependencies since ncc will bundle it, right?

packages/node/src/dev-server.ts
```diff
@@ -68,6 +73,11 @@ async function createEventHandler(
   return createServerlessEventHandler(entrypointPath, {
     shouldAddHelpers: options.shouldAddHelpers,
     useRequire,
+    launcherType: detectServerlessLauncherType(
+      entrypoint!,
+      process.versions.node
```
styfle: This won't match the same syntax as runtime

- `process.versions.node = '16.19.0'`
- `runtime = 'nodejs16.x'`

packages/node/src/utils.ts
```diff
@@ -79,3 +81,27 @@ export function isEdgeRuntime(runtime?: string): runtime is EdgeRuntimes {
...
+function checkLauncherCompatibility(
+  entrypoint: string,
+  launcherType: LauncherConfiguration['launcherType'],
+  runtime: string
+) {
+  if (launcherType === 'edge-light') {
+    const nodeMajor = Number(runtime.replace('nodejs', '').split('.')[0]);
+    if (nodeMajor < 18) {
+      throw new Error(
+        `${entrypoint}: ${launcherType} launcher type can only be used with node.js 18 and later`
```
styfle: Is this an error message that a user will see at build time? Seems like we need to use `NowBuilderError()` and add a link to the documentation. Example:

https://github.com/vercel/vercel/blob/5b88f673f89c44b9e91e44247e29be6bed43805a/packages/build-utils/src/fs/node-version.ts#L64-L70

packages/node/test/detection-fixtures/node/index.ts
```diff
@@ -0,0 +1,3 @@
+export default function (req, res) {
```
styfle: Should we add types here to ensure its detected with real TS code?

styfle: Also another test with `req` but not `res`?

packages/node/src/index.ts
```diff
@@ -475,10 +476,12 @@ export const build: BuildV3 = async ({
     const experimentalResponseStreaming =
       staticConfig?.experimentalResponseStreaming === true ? true : undefined;
 
+    const { runtime } = nodeVersion;
```
styfle: We already have `nodeVersion.major` so I think we should use that instead to pass into `detectServerlessLauncherType()`

packages/build-utils/src/types.ts
EndangeredMassa: I'm out of the loop on Edge Light, but I agree that this value should be `EdgeLight` unless there are constraints from other parts of the system.

feugy: Really not a strong opinion.
I meant to use the WinterCG key because when running this signature, we aim to provide our runtime (even if not using edge-runtime directly here).
In my first experiment, I used `Web` as a key, but it didn't felt good.

Happy to switch to `EdgeLight`, unless a better proposal arises.

packages/node/package.json
feugy: Good point, thank you!

packages/node/src/index.ts
feugy: Thanks! I'll pass `detectServerlessLauncherType` the `runtime.nodeVersion` since it's available.
I need this function because it's used in two places: here and in the dev server.

packages/node/src/dev-server.ts
feugy: Indeed. `detectServerlessLauncherType()` was handling the both cases. But based on your other remark, I'll extract the version here, to pass it as a number instead.

packages/node/src/utils.ts
feugy: Yes, good call! I have a similar error in my build-container PR.

packages/node/test/detection-fixtures/node/index.ts
feugy: It makes no difference to the feature detector (which has its [own tests](https://github.com/vercel/edge-runtime/blob/main/packages/feature-detector/test/has-edge-signature.test.ts)), but it can only be beneficial to add more. Doing it.

packages/build-utils/src/types.ts
EndangeredMassa: Matching the WinterCG key sounds reasonable. I'd rather a code search of `edge-light` show all references than maintain casing style in this type.

packages/node-bridge/launcher.js
```diff
@@ -8,10 +8,13 @@ const { Bridge } = require('./bridge.js');
  */
 function makeVercelLauncher(config) {
   const {
+    launcherType = 'Nodejs',
     entrypointPath,
     bridgePath,
     helpersPath,
+    webHandlerPath,
```
styfle: Since this is a new variable, we can't guarantee it will be defined. Think about legacy Builders and anyone targeting BOA v3.

https://vercel.com/docs/build-output-api/v3#vercel-primitives/serverless-functions/configuration/node-js-config

packages/node-bridge/tsconfig.json
```diff
@@ -11,6 +11,6 @@
     "module": "commonjs",
     "skipLibCheck": true
   },
-  "include": ["helpers.ts", "bridge.js", "launcher.js"],
+  "include": ["helpers.ts", "bridge.js", "launcher.js", "web-handler.js"],
```
styfle: Can you check the size before and after of thse files?

We need to make sure they are extremely small because any increase could put new deployments over the limit of 50MB.

packages/node-bridge/launcher.js
feugy: Yes, and it is ok. The launcher code only uses `webHandlerPath` if the incoming launcher type is `EdgeLight`, and having it `undefined` will not create any issue.

The build container is the only one using `makeVercelLauncher`, and will stay consistent: prior versions will not have `webHandlerPath`, and will also forbid `EdgeLight` as launcher type.

Same with BOAv3 compliant apps: since they won't use `EdgeLight`, they won't need `webHandlerPath`.
Let's have a call if I'm missing something.

packages/node-bridge/tsconfig.json
feugy: Good call.
Ncc output (I trimmed out some of the lines):

Main:
```
 1kB  helpers/helpers.d.ts
 2kB  helpers/types.d.ts
22kB  helpers/index.js
122kB  source-map-support/index.js
```

This branch:
```
 1kB  helpers/helpers.d.ts
 2kB  helpers/types.d.ts
22kB  helpers/index.js
122kB  source-map-support/index.js
 0kB  web-handler/web-handler.d.ts
 1kB  web-handler/helpers.d.ts
 2kB  web-handler/types.d.ts
10kB  web-handler/index.js
```

So it's 13kB more

packages/node/src/dev-server.ts
```diff
@@ -35,14 +41,7 @@ function parseRuntime(
   const project = new Project();
   const staticConfig = getConfig(project, entryPointPath);
   const runtime = staticConfig?.runtime;
-  if (runtime && !isEdgeRuntime(runtime)) {
-    throw new Error(
-      `Invalid function runtime "${runtime}" for "${entrypoint}". Valid runtimes are: ${JSON.stringify(
-        Object.values(EdgeRuntimes)
-      )}. Learn more: https://vercel.link/creating-edge-functions`
-    );
-  }
```
feugy, line 32: **question:** since this code was very close to [this one](https://github.com/vercel/vercel/pull/9286/files#diff-b80a0ec8c1900042621cb5f358ad6f6d54d114269c292d28b57146fe0ecf4b92L409), I merged them in an utility. Is it worthy?

In particular, the error message differ. Previously, in dev mode:
>  Invalid function runtime "bla" for "api/edge.js". Valid runtimes are: ["edge","experimental-edge"]. Learn more: https://vercel.link/creating-edge-functions`

Previously, at build time:
> Unsupported "runtime" property in `config`: "bla" (must be one of: ["nodejs","edge","experimental-edge"])

Now they both are:
> api/edge.js: unsupported "runtime" property in \`config\`: "bla" (must be one of: ["nodejs-experimental","edge","experimental-edge"]). Learn more: https://vercel.link/creating-edge-functions

packages/node/src/utils.ts
```diff
@@ -73,9 +75,54 @@ export enum EdgeRuntimes {
...
+
+export function checkLauncherCompatibility(
+  entrypoint: string,
+  launcherType: LauncherConfiguration['launcherType'],
+  nodeMajorVersion: number
+) {
+  if (launcherType === 'EdgeLight' && nodeMajorVersion < 18) {
+    throw new NowBuildError({
+      code: 'INVALID_RUNTIME_FOR_LAUNCHER',
+      message: `${entrypoint}: configured runtime "${IsomorphicRuntime}" can only be used with node.js 18 and later`,
```
```suggestion
      message: `${entrypoint}: configured runtime "${IsomorphicRuntime}" can only be used with Node.js 18 or newer`,
```

packages/node/src/utils.ts
```diff
@@ -73,9 +75,54 @@ export enum EdgeRuntimes {
...
     Object.values(EdgeRuntimes).includes(runtime as EdgeRuntimes)
   );
 }
+
+export function detectServerlessLauncherType(
+  config: { runtime?: string } | null | undefined
+): LauncherType {
+  // The final execution environment is always node.js, but the signature differ:
+  // - for BuildOutAPI, `launcherType: EdgeLight` is web-compliant signature, `launcherType: Nodejs` is node-compliant signature
+  // - for vc api files, `runtime: edge` is web-compliant signature on Edge, `runtime: nodejs` is web-compliant signature on node.js, nothing is legacy node-compliant on node.js
```
styfle: Lets make sure all of this is documented here

https://vercel.com/docs/build-output-api/v3

```suggestion
  // - for vc api files, `runtime: edge` is web-compliant signature on Edge, `runtime: nodejs` is web-compliant signature on node.js, undefined is legacy node-compliant on node.js
```

packages/node/src/utils.ts
```diff
@@ -73,9 +75,54 @@ export enum EdgeRuntimes {
   ExperimentalEdge = 'experimental-edge',
 }
 
+export const IsomorphicRuntime = 'nodejs';
```
styfle: Perhaps this should be 

```js
export const NodejsRuntimes = {
  Legacy = undefined,
  Isomorphic = 'nodejs',
}
```

packages/node-bridge/tsconfig.json
styfle: Looks good. Basically we'll swap out `helpers.js` with `web-handler.js` when `export const config` is used so should be much smaller 👍 

image

packages/build-utils/src/types.ts
```diff
@@ -312,13 +312,16 @@ export interface Builder {
   config?: Config;
 }
 
+export type LauncherType = 'Nodejs' | 'EdgeLight';
```
styfle: I think `EdgeLight` is misleading since thats a runtime and not an API signature.

Perhaps `WinterCG` or `WHATWG` might be better here since it implies Request/Response from fetch.

Also cc @TooTallNate @EndangeredMassa @Ethan-Arrowood who since this affects BOA.

packages/node/src/utils.ts
feugy: Typescript does not accept `undefined` in a string enum, but I agree it's better code-as-documentation.
I've relaxed the compiler and included it.

packages/build-utils/src/types.ts
feugy: I've changed to `WinterCG`

packages/node/src/utils.ts
```diff
@@ -73,9 +75,69 @@ export enum EdgeRuntimes {
   ExperimentalEdge = 'experimental-edge',
 }
 
+export enum NodejRuntime {
```
styfle: Should this be plural to match the enum above?

```suggestion
export enum NodejsRuntimes {
```

packages/node/src/utils.ts
```diff
@@ -73,9 +75,69 @@ export enum EdgeRuntimes {
...
+  // @ts-ignore typescript does not allow undefined as a value
+  Legacy = undefined,
+  WinterCG = 'nodejs-experimental', // TODO remove -experimental once ready
+}
+
+export const ALLOWED_RUNTIMES = [
+  NodejRuntime.WinterCG as string,
```
styfle: Now you can change this to `Object.values(NodejsRuntimes)` right?

feugy: We actually can not, because `NodejsRuntime` contains `Legacy`, which is `undefined`. It will look quite bad in error message, where we display items of thie array.

packages/node/src/utils.ts
feugy: <picture data-single-emoji=":facepalm:" title=":facepalm:">:facepalm:</picture> good catch!

packages/node-bridge/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowJs": true,
     "checkJs": true,
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "DOM"],
```
EndangeredMassa: **question (non-blocking):** The "DOM" lib here will make TS believe it's OK to use all DOM APIs, but Edge Functions don't support all of them. Could this become an issue?

packages/node-bridge/web-handler.ts
```diff
@@ -0,0 +1,27 @@
+import type { BuildDependencies } from '@edge-runtime/node-utils';
+import { buildToNodeHandler } from '@edge-runtime/node-utils';
+
+export function transformToNodeHandler(webHandler: any) {
```
EndangeredMassa: **question (non-blocking):** `transformToNodeHandler` is often called with a second paramter (`runtime`), but that's not used. Should it be used in some way or removed from call sites?

packages/node/src/utils.ts
```diff
@@ -73,9 +75,69 @@ export enum EdgeRuntimes {
...
+        `${entrypoint}: \`config.runtime: "nodejs"\` semantics will evolve soon. Please remove the \`runtime\` key to keep the existing behavior.`
+      );
+    }
+    if (!ALLOWED_RUNTIMES.includes(runtime)) {
+      throw new Error(
+        `${entrypoint}: unsupported "runtime" property in \`config\`: ${JSON.stringify(
+          runtime
+        )} (must be one of: ${JSON.stringify(
+          ALLOWED_RUNTIMES
+        )}). Learn more: https://vercel.link/creating-edge-functions`
```
EndangeredMassa, line 101: **question (non-blocking):** Do we even want to say that `experimental-edge` is a valid value here? Maybe we hide that from this list when it's displayed.

EndangeredMassa:
Resolved merge conflicts except for `node-bridge`. A PR will come for changes there.

EndangeredMassa:
@Kikobeats is taking over this effort

PR #9289, opened 2023-01-23
[cli] add missing deprecation warnings for builders/runtimes

styfle:
We used to print deprecations on the old build pipeline but this was lost when switching to `vercel build`.

This PR ensures that `vercel build` prints deprecation warnings if available.

## Example

https://npmjs.com/package/@now/node

image

### DIFF
--- a/packages/cli/src/util/build/import-builders.ts
+++ b/packages/cli/src/util/build/import-builders.ts
@@ -5,12 +5,8 @@ import { satisfies } from 'semver';
 import { dirname, join } from 'path';
 import { mkdirp, outputJSON, readJSON, symlink } from 'fs-extra';
 import { isStaticRuntime } from '@vercel/fs-detectors';
-import {
-  BuilderV2,
-  BuilderV3,
-  PackageJson,
-  spawnAsync,
-} from '@vercel/build-utils';
+import { BuilderV2, BuilderV3, PackageJson } from '@vercel/build-utils';
+import execa from 'execa';
 import * as staticBuilder from './static-builder';
 import { VERCEL_DIR } from '../projects/link';
 import { Output } from '../output';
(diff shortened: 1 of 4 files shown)


packages/cli/test/unit/util/build/import-builders.test.ts
```diff
@@ -105,6 +105,7 @@ describe('importBuilders()', () => {
       await expect(client.stderr).toOutput(
         '> Installing Builder: @vercel/node'
       );
+      await expect(client.stderr).not.toOutput('npm WARN deprecated');
```
TooTallNate, line 108: I'd prefer not to use `.not.toOutput()` if possible, as it relies on a few seconds timeout and will make the tests start to take long as these add up. Not saying we can't use it here, but something to keep in mind.

styfle: @TooTallNate The thing is, we should really be testing to make sure we don't print warnings when we don't intend to. If you can think of another way to test for that, I'm all ears. But I don't think its a good idea to remove these assertions.

packages/cli/src/util/build/import-builders.ts
```diff
@@ -213,14 +209,21 @@ async function installBuilders(
...
       'npm',
       ['install', '@vercel/build-utils', ...buildersToAdd],
       {
         cwd: buildersDir,
         stdio: 'pipe',
+        reject: true,
       }
     );
+    stderr
+      .split('\n')
```
cb1kenobi: Once upon a time I ran into a problem on Windows where the output contained both carriage returns and new lines. Ever since then, I've always split on `/\r?\n/`. I'm not sure if Node.js guards against this these days or if Windows got better. I wish I could recall the exact scenario, but it was probably PowerShell related. In any case, it's probably harmless as is.

packages/cli/test/unit/util/build/import-builders.test.ts
cb1kenobi: You may have to resort to adding a listener to `client.stderr` and manually append the chunks into a buffer (like this https://stackoverflow.com/a/49428486) that can be asserted after the command has finished.

TooTallNate: I think it's appropriate in this case then, let's keep it. Let's just be mindful when adding these.

cb1kenobi (approved):
LGTM

packages/cli/src/util/build/import-builders.ts
styfle: Added in [00ae98d](https://github.com/vercel/vercel/pull/9289/commits/00ae98d1948869b0f710020316bf0eba2218d5aa)

PR #9291, opened 2023-01-23
[static-build] Remove `VERCEL_GATSBY_BUILDER_PLUGIN` env var check

TooTallNate:
Keep as a draft for now until we're ready to remove the opt-in/out completely.


TooTallNate:
This happened in https://github.com/vercel/vercel/pull/9766.

PR #9297, opened 2023-01-24
[static-build] Use tarball URLs for Gatsby plugin E2E tests

TooTallNate:
When running the static-build integration tests, use the tarball URL for the Gatsby plugins, so that any changes made to the plugin(s) in the PR will be reflected in the test deployments without having to publish to npm first.

Example screenshots of a deployment's build logs before I removed the debugging:

Screenshot 2023-01-23 at 7 35 32 PM
Screenshot 2023-01-23 at 7 35 41 PM

### DIFF
--- a/packages/static-build/src/utils/gatsby.ts
+++ b/packages/static-build/src/utils/gatsby.ts
@@ -22,8 +22,8 @@ const PLUGIN_VERSIONS = new Map<PluginName, string>([
 ]);
 
 // For E2E tests, ensure the same version of the plugin is used as the source code
-const { NODE_ENV, VERCEL_CLI_VERSION } = process.env;
-if (NODE_ENV === 'test' && VERCEL_CLI_VERSION) {
+const { VERCEL_CLI_VERSION } = process.env;
+if (VERCEL_CLI_VERSION?.startsWith('https://')) {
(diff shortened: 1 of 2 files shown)


packages/static-build/src/utils/gatsby.ts
```diff
@@ -2,39 +2,55 @@ import { PackageJson } from '@vercel/build-utils';
...
+const PLUGINS = [
+  '@vercel/gatsby-plugin-vercel-analytics',
+  '@vercel/gatsby-plugin-vercel-builder',
+] as const;
+type PluginName = typeof PLUGINS[number];
+
+const PLUGIN_VERSIONS = new Map<PluginName, string>([
+  ['@vercel/gatsby-plugin-vercel-analytics', 'latest'],
+  ['@vercel/gatsby-plugin-vercel-builder', 'latest'],
+]);
```
styfle: Small nit pick: I think you can remove this Map and the initialization here instead lazily create the URL on line 112

TooTallNate: SGTM 718930ca48e1a1d605770a23d8ad86496e355df8

styfle (approved):
Thanks!

PR #9298, opened 2023-01-24
[static-build] Add Gatsby plugins as dev dependencies

TooTallNate:
So that the test runs get invalidated by Turbo when code changes within one of the plugin files. This is to avoid a PR breaking something in a plugin, but we don't notice it because the static-build tests still "pass" due to cache hit.


cb1kenobi (approved):
LGTM. I wish JSON supported comments so the context of these dependencies wouldn't get lost.

PR #9301, opened 2023-01-24
[tests] Added cron workflow to update turbo

cb1kenobi:
Linear: https://linear.app/vercel/issue/VCCLI-450/add-cron-update-turbo-workflow

Based on the Next.js update script, this PR adds a cron job that checks for the latest Turbo canary release, then updates the `package.json` and `pnpm-lock.yaml` files.

### DIFF
--- a/.github/workflows/cron-update-turbo.yml
+++ b/.github/workflows/cron-update-turbo.yml
@@ -1,0 +1,27 @@
+name: Cron Update Turbo
+
+on:
+  # Run every 4 hours https://crontab.guru/every-4-hours
+  schedule:
+    - cron: '0 */4 * * *'
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        # 0 means fetch all commits so we can commit and push in the script below
+        with:
+          fetch-depth: 0
+      - name: install pnpm@7.26.0
+        run: npm i -g pnpm@7.26.0
+      - name: Create Pull Request
+        uses: actions/github-script@v6
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        # See https://github.com/actions/github-script#run-a-separate-file-with-an-async-function
+        with:
+          script: |
+            const script = require('./utils/update-turbo.js')
+            await script({ github, context })
--- a/.github/workflows/cron-update-turbo.yml
+++ b/.github/workflows/cron-update-turbo.yml
@@ -1,9 +1,9 @@
 name: Cron Update Turbo
 
 on:
-  # Run every 4 hours https://crontab.guru/every-4-hours
+  # Run every week https://crontab.guru/every-week
   schedule:
-    - cron: '0 */4 * * *'
+    - cron: '0 0 * * 0'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/cron-update-turbo.yml
```diff
@@ -0,0 +1,27 @@
+name: Cron Update Turbo
+
+on:
+  # Run every 4 hours https://crontab.guru/every-4-hours
+  schedule:
+    - cron: '0 */4 * * *'
```
styfle: I'm thinking we might want to do this less often. The reason being that turbo bugs might not show up until we do a release and it could be hard to track down if we update turbo every day.

How about once per week?

utils/update-turbo.js
```diff
@@ -0,0 +1,56 @@
...
+  const output = execFileSync(cmd, args, opts).toString().trim();
+  console.log({ output });
+  console.log();
+  return output;
+}
+
+module.exports = async ({ github, context }) => {
+  const pkgJson = JSON.parse(readFileSync('package.json', 'utf-8'));
+  const oldVersion = pkgJson.devDependencies.turbo;
+  const newVersion = exec('npm', ['view', 'turbo', 'dist-tags.canary']);
```
styfle: Should this be `pnpm` instead?

utils/update-turbo.js
```diff
@@ -0,0 +1,56 @@
...
+  const { repo, owner } = context.repo;
+
+  await github.rest.pulls.create({
+    owner,
+    repo,
+    head: branch,
+    base: 'main',
+    title: `[tests] Upgrade Turbo to version ${newVersion}`,
+    body: `This auto-generated PR updates Turbo to version ${newVersion}`,
+  });
```
styfle, line 55: Lets also add the labels automatically too

.github/workflows/cron-update-turbo.yml
cb1kenobi: Done!

utils/update-turbo.js
cb1kenobi: Done

utils/update-turbo.js
cb1kenobi: Done

styfle (approved):
Perfect, thanks! 🥇

.github/workflows/cron-update-turbo.yml
```diff
@@ -0,0 +1,27 @@
...
+  create-pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        # 0 means fetch all commits so we can commit and push in the script below
+        with:
+          fetch-depth: 0
+      - name: install pnpm@7.26.0
+        run: npm i -g pnpm@7.26.0
```
Ethan-Arrowood, line 18: is `npm` available in actions without setting up node first?

Ethan-Arrowood (approved):

.github/workflows/cron-update-turbo.yml
cb1kenobi: Not that I'm aware of. I copied this from https://github.com/vercel/vercel/blob/main/.github/workflows/test-unit.yml#L46-L47.

PR #9302, opened 2023-01-24
[tests] Disable Turbo's update notifier

cb1kenobi:
Since switching to the canary releases of Turbo, we are seeing the update notification often. We have a cron job (https://github.com/vercel/vercel/pull/9301) that will keep this up-to-date for us, so we don't need to have Turbo notify us of updates.

image


Ethan-Arrowood (dismissed):
is there some kind of global config for turbo so we don't have to pass it in every command?

cb1kenobi:
> is there some kind of global config for turbo so we don't have to pass it in every command?

Not that I'm aware of. Based on the code, it appears the environment variable is the only option: https://github.com/vercel/turbo/blob/5232e179a408f24da2ecc7da6d6d42c6985a9a67/crates/turbo-updater/src/lib.rs#L80-L88.

styfle (commented):
This doesn't work on windows https://github.com/vercel/vercel/actions/runs/3999082420/jobs/6862537077#step:8:15

We could use cross-env but it feels strange to put this env var in every script

cb1kenobi:
@styfle could we ask the turbo team to add a flag to disable the update notifier?

package.json
```diff
@@ -42,14 +42,14 @@
...
     "pre-commit": "lint-staged",
     "test": "jest --rootDir=\"test\" --testPathPattern=\"\\.test.js\"",
-    "test-unit": "pnpm test && node utils/gen.js && turbo run test-unit",
-    "test-integration-cli": "node utils/gen.js && turbo run test-integration-cli",
-    "test-integration-once": "node utils/gen.js && turbo run test-integration-once",
-    "test-integration-dev": "node utils/gen.js && turbo run test-integration-dev",
+    "test-unit": "pnpm test && node utils/gen.js && NO_UPDATE_NOTIFIER=1 turbo run test-unit",
+    "test-integration-cli": "node utils/gen.js && NO_UPDATE_NOTIFIER=1 turbo run test-integration-cli",
+    "test-integration-once": "node utils/gen.js && NO_UPDATE_NOTIFIER=1 turbo run test-integration-once",
+    "test-integration-dev": "node utils/gen.js && NO_UPDATE_NOTIFIER=1 turbo run test-integration-dev",
```
EndangeredMassa: This will fail on windows, won't it? Maybe we should use [cross-env](https://github.com/kentcdodds/cross-env#readme) or some other solution.

EndangeredMassa (changes requested):

cb1kenobi:
`turbo` now has support for a `--no-update-notifier` flag! https://github.com/vercel/turbo/pull/4003

Thanks @tknickman!

styfle (dismissed):

cb1kenobi:
Yikes, I think we should wait for the `turbo` dependency to be updated first.

EndangeredMassa (approved):

PR #9303, opened 2023-01-24
[examples] Upgrade Angular to version 15.1.1

sp90:
Upgrade the angular example to use the latest stable version of angular


jessycormier:
We would probably want to have this updated to the latest v17 of angular with SSR/SSG enabled.

PR #9304, opened 2023-01-24
[gatsby-plugin-vercel-builder] Get rid of the `_ssr` function

TooTallNate:
Right now we create the SSR serverless function at path `_ssr`, and then symlink all the other pages to that function.

Instead just make the first page encountered be the "real" function, and symlink all the other pages to that endpoint.

### DIFF
--- a/packages/gatsby-plugin-vercel-builder/src/helpers/functions.ts
+++ b/packages/gatsby-plugin-vercel-builder/src/helpers/functions.ts
@@ -10,29 +10,15 @@ import {
 } from '../handlers/build';
 import type { GatsbyFunction, GatsbyPage } from '../schemas';
 
+/**
+ * Gatsby SSR/DSG on Vercel is enabled through Vercel Serverless Functions.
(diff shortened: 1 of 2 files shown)


EndangeredMassa:
@TooTallNate what are the pros/cons here? Without more details, the current (pre-PR) version makes more sense to me.

TooTallNate:
This PR is the more correct behavior. `GET /_ssr` should not invoke the Serverless Function - it should 404. Without this change [it's a 500](https://gatsby-starter-default-63l88z8i8-tootallnate.vercel.app/_ssr/).

The current way would also prevent a user page (`src/pages/_ssr.js`) from working correctly (edge-case, but still valid).

It's also just annoying [seeing `_ssr` in the Functions tab](https://vercel.com/tootallnate/gatsby-starter-default/FYSrfwkYaq3vCmJNRAhKaUrxvShU/functions) 🙈

Ethan-Arrowood (approved):

PR #9309, opened 2023-01-25
[examples] Pin Node version to 16.x for Parcel build

swarnava:
Looks like our Parcel examples are currently failing in Node 18.x. Pinning Node version to 16.x till we update the example


swarnava:
created #9310

PR #9310, opened 2023-01-25
[examples] Update Parcel version to support Node 18

swarnava:
Looks like our Parcel examples are currently failing in Node 18.x. Updating the version which also Support Node 18.x

image


styfle (approved):
Great, thanks!

PR #9312, opened 2023-01-25
[static-build] Cleanup gatsby file changes after static-build completes

Ethan-Arrowood:
Adds a new function that restores the user's `gatsby-config.(j|t|mj)s` and `gatsby-node.(j|t|mj)s` files


packages/static-build/src/utils/gatsby.ts
```diff
@@ -335,3 +335,27 @@ module.exports = gatsbyNode;
...
+      const generatedFilePath = path.join(dir, generatedFile);
+
+      if (
+        (await fileExists(backupFilePath)) &&
+        (await fileExists(generatedFilePath))
+      ) {
+        await fs.rm(generatedFilePath);
+        await fs.rename(backupFilePath, generatedFilePath);
+      } else if (await fileExists(generatedFilePath)) {
+        await fs.rm(generatedFilePath);
```
TooTallNate: This one seems dangerous. How about if the user existing config file was not modified?

I think we may want to add a comment to the top of the generated files and check for it before deleting.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -92,6 +92,8 @@ async function updateGatsbyConfig(dir: string, plugins: string[]) {
   }
 }
 
+const GENERATED_FILE_COMMENT = `// This file was generated by @vercel/static-build`;
```
TooTallNate, line 95: I would suggest something more generic like:

```
This file was auto-generated by Vercel
```

Not a huge preference though.

TooTallNate (approved):

packages/static-build/src/utils/gatsby.ts
```diff
@@ -303,6 +311,36 @@ module.exports = gatsbyNode;
   );
 }
 
+export async function cleanupGatsbyFiles(dir: string) {
+  const backup = '.__vercel_builder_backup__';
+  const fileEndings = ['.js', '.ts', '.mjs'];
+
+  for (const fileName of [GATSBY_CONFIG_FILE, GATSBY_NODE_FILE]) {
+    for (const fileEnding of fileEndings) {
```
EndangeredMassa, line 319: **question (non-blocking):** Should this try-catch around each operation so the rest of the operations can continue if one errored?

EndangeredMassa (approved):

PR #9313, opened 2023-01-25
[static-build] Include Gatsby plugins as regular dependencies

TooTallNate:
Right now, static-build will add the necessary Gatsby plugins to the project's `package.json` at build-time, which has been bothersome for package managers when using a frozen lockfile.

Another issue with it is that we install `latest` version of the plugin, so the version used becomes disjointed from the CLI version itself, which leads to unpredictability when trying to debug issues or help users roll back to a previous behavior if something breaks.

So instead of patching `package.json` directly, include the plugins as deps of static-build itself, and create symlinks to those paths into the project's `node_modules` directory.

### DIFF
--- a/packages/static-build/package.json
+++ b/packages/static-build/package.json
@@ -15,8 +15,8 @@
   "scripts": {
     "build": "node build",
     "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "pnpm test -- test/build.test.ts test/gatsby.test.ts test/prepare-cache.test.ts",
-    "test-integration-once": "pnpm test -- test/integration-*.test.js"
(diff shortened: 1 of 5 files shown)


Ethan-Arrowood (approved):

packages/static-build/package.json
```diff
@@ -28,6 +28,10 @@
       }
     }
   },
+  "dependencies": {
+    "@vercel/gatsby-plugin-vercel-analytics": "1.0.7",
+    "@vercel/gatsby-plugin-vercel-builder": "1.0.1"
+  },
```
styfle, line 34: I think you also need to mark as external in ncc script, right?

PR #9314, opened 2023-01-25
[next] Remove hardcoded NODE_OPTIONS

styfle:
The `NODE_OPTIONS` env var value is wrong here and we shouldn't be overriding the user's setting to begin with.

Typically, it will be assigned in a build container or project settings.

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -45,7 +45,6 @@ import {
   remove,
   writeFile,
 } from 'fs-extra';
-import os from 'os';
 import path from 'path';
 import resolveFrom from 'resolve-from';
 import semver from 'semver';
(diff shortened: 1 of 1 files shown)


ijjk (approved):

packages/next/src/index.ts
```diff
@@ -408,8 +407,6 @@ export const build: BuildV2 = async ({
   }
 
   const env: typeof process.env = { ...spawnOpts.env };
-  const memoryToConsume = Math.floor(os.totalmem() / 1024 ** 2) - 128;
-  env.NODE_OPTIONS = `--max_old_space_size=${memoryToConsume}`;
```
mknichel, line 412: Does it make sense to set this if env.NODE_OPTIONS is not already defined?

styfle: No, I don't think so. It should be assigned in a build container or project settings.

lucleray (approved):

mknichel (approved):

PR #9315, opened 2023-01-25
[fs-detectors] getMonorepoDefaultSettings: Fix settings

chloetedder:
1. `commandForIgnoringBuildStep` should be run at the project directory level not the monorepo root level
2. Simplifying the `installCommand` because doesn't need the relative root unless it is `npm`


Ethan-Arrowood (dismissed):

Ethan-Arrowood:
https://github.com/vercel/vercel/actions/runs/4009863452/jobs/6885709398#step:10:1776 looks like there are some unit tests to fix

packages/fs-detectors/test/fixtures/get-monorepo-default-settings/turbo-npm/packages/app-1/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "app-1",
```
styfle: Lets use a more specific name so it doesn't collide

styfle (changes requested):
We have tons of warnings due to the repeated names

- See related https://github.com/vercel/vercel/pull/9317

Lets either use a unique `name`, or if its not needed use `private: true`

styfle (dismissed):

Ethan-Arrowood (approved):

PR #9318, opened 2023-01-25
[tests] add gatsby related unit tests to static-build

Ethan-Arrowood:
Adds some unit tests for the gatsby injection logic. 

These are net new since we've heavily changed the injection logic. The tests and fixtures in `build-fixtures` and `builds.test.js` are seemingly not executed. We may delete those here too.

### DIFF
--- a/packages/static-build/package.json
+++ b/packages/static-build/package.json
@@ -15,7 +15,7 @@
   "scripts": {
     "build": "node build",
     "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "pnpm test test/build.test.ts test/prepare-cache.test.ts",
+    "test-unit": "pnpm test test/build.test.ts test/gatsby.test.js test/prepare-cache.test.ts",
     "test-integration-once": "pnpm test test/integration-*.test.js"
   },
   "jest": {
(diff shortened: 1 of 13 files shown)


packages/static-build/package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "build": "node build",
     "test": "jest --env node --verbose --bail --runInBand",
-    "test-unit": "pnpm test test/build.test.ts test/prepare-cache.test.ts",
+    "test-unit": "pnpm test test/build.test.ts test/gatsby.test.js test/prepare-cache.test.ts",
```
styfle, line 18: I just noticed there is a `builds.test.js` that is being ignored. Do we know why?

https://github.com/vercel/vercel/blob/a7a5bf1a1293b23360833a8dc5e9286f3b41c6e9/packages/static-build/test/builds.test.js

Ethan-Arrowood: I mentioned that in the PR description. No idea when or why its ignored, including the plethora of fixtures it references. I actually added a test to that file previously thinking it was being executed lol

regardless, the fixtures are out of date and so are the tests which is why this is all net-new

TooTallNate (approved):
We'll need to adjust these tests for https://github.com/vercel/vercel/pull/9312 and https://github.com/vercel/vercel/pull/9313, but let's merge this first to ensure that we keep everything green.

PR #9320, opened 2023-01-26
[go] Improved 'go build' error handling

cb1kenobi:
The `go` builder doesn't properly handle when `go build` fails. There are 2 failure scenarios:

1. The `go` binary is not found in the `PATH`. There is no message that the `go` executable was not found. Worse, the builder continues and fails trying to run the compiled executable:

image

The solution is to switch from `spawnSync()` to `execFileSync()` so that if `go` is not found, it will throw like this:

image
Note: I temporarily changed the code to spawn `gooooo` to simulate `go` not being found. Either way, `execFileSync()` will throw and we get a helpful message.

2. The `go build` fails due to some actual build error. This should never happen, but better safe than sorry.

The solution here is to allow `go build` to inherit stdout/stderr and if `go build` should fail, we can see the output:

image


packages/go/index.ts
```diff
@@ -694,9 +694,10 @@ Learn more: https://vercel.com/docs/runtimes#official-runtimes/go`
   }`;
 
   debug(`SPAWNING go build -o ${executable} ./... CWD=${tmp}`);
-  spawnSync('go', ['build', '-o', executable, './...'], {
+  execFileSync('go', ['build', '-o', executable, './...'], {
     cwd: tmp,
     env,
+    stdio: 'inherit',
```
styfle, line 700: I think we only want stderr to print right? In the successful case we don't want to print anything?

cb1kenobi: On successful builds, `go build` has no output.

EndangeredMassa (approved):

PR #9322, opened 2023-01-26
[build-utils] Rewrite rename() to be more efficient

cb1kenobi:
Around 6 months ago, @styfle brought to my attention how `rename()` in build-utils used `reduce()` and could be written better. So, I rewrote it.

Before, the code would create a new `Files` object and copy the contents of the previous `Files` object. This caused heavy garbage collection and memory thrashing.

Instead, I created a single `Files` object, then add the files to it.

Results:

| # Files | Before | After |
|---|---|---|
| 1,000 | 75 ms | 1 ms |
| 10,000 | 10.6 s | 7 ms |
| 20,000 | 44.6 s | 16 ms |
| 30,000 | 105.5 s | 22 ms |
| 100,000 | Too long | 73 ms |


styfle (approved):
Can we delete this function? I don't think its used.

cb1kenobi:
I don't think we should remove the `rename()` function. It's currently being used by the `vercel-php` builder: https://github.com/vercel-community/php/blob/master/src/index.ts#L83-L102.

There's another builder called `now-custom-runtime` that uses it. There could be other builders, who knows?

packages/build-utils/test/unit.test.ts
```diff
@@ -451,3 +454,26 @@ it('should retry npm install when peer deps invalid and npm@8 on node@16', async
...
   ]);
 });
+
+describe('rename', () => {
+  it('should rename keys of files map', () => {
+    const before: Files = {};
+    const expected: Files = {};
+    const reverser = (s: string) => s.split('').reverse().join('');
+
+    for (let i = 0; i < 20000; i++) {
```
styfle: I'm curious why 20000 was chosen here.

This takes 804ms for CI so seems like we could reduce this down to speed up CI. I think we should keep unit tests to 100ms or less if we can.

https://github.com/vercel/vercel/actions/runs/4012696635/jobs/6891556522#step:10:1498

packages/build-utils/test/unit.test.ts
```diff
@@ -451,3 +454,26 @@ it('should retry npm install when peer deps invalid and npm@8 on node@16', async
...
+      let key = '';
+      for (let j = 0; j < 5; j++) {
+        key += String.fromCharCode(
+          Math.floor(Math.random() * (122 - 97 + 1) + 97)
```
styfle: Lets not use random for a test because its not deterministic. Sequential keys is fine for this test.

EndangeredMassa (approved):
So fast!

packages/build-utils/test/unit.test.ts
cb1kenobi: Yeah, good call. I was benchmarking with 20k.

packages/build-utils/test/unit.test.ts
cb1kenobi: I changed the test from reversing random generated strings to uppercasing letters instead.

EndangeredMassa (approved):

PR #9326, opened 2023-01-26
[frameworks] add `envPrefix` to `hydrogen` framework

stipsan:
Hydrogen uses the `PUBLIC_` env prefix, [it also supports `VITE_`](https://github.com/Shopify/hydrogen/blob/8168cd2dd850b59fb051e610d1db918f7e4ef327/packages/hydrogen/src/framework/plugins/vite-plugin-hydrogen-config.ts#L127) but [the docs only talk about `PUBLIC_`](https://shopify.dev/custom-storefronts/hydrogen/environment-variables#public-variables).

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1698,6 +1698,7 @@ export const frameworks = [
     description: 'React framework for headless commerce',
     website: 'https://hydrogen.shopify.dev',
     useRuntime: { src: 'package.json', use: '@vercel/hydrogen' },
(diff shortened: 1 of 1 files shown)


styfle (dismissed):
Thanks!

EndangeredMassa:
We're doing some testing to make sure this works in production.

We may also need a change in [packages/hydrogen](https://github.com/vercel/vercel/tree/main/packages/hydrogen) to support this properly.

We'll report back.

styfle:
I'm going to merge this one and create a new PR to make sure `@vercel/hydrogen` applies the value

PR #9331, opened 2023-01-27
[docs] Replace references to @now/next in error docs

ijjk:
It's been long enough I think we can change these to refer to `@vercel/next` instead. 

x-ref: [slack thread](https://vercel.slack.com/archives/C03DQ3QFV7C/p1674849801196389)


styfle (approved):
I think its an alias but lets do it

PR #9333, opened 2023-01-28
[docs] Add troubleshooting step for no serverless pages error

biiishal:
Adds an additional troubleshooting option to resolve this build step error:

```
Error: No serverless pages were built
```


styfle (commented):
Thanks!

I think this error should not happen since Next.js 12 so there might be something else going on.

I'll defer to @ijjk to see how we can improve this doc.

ijjk (approved):
This seems like a valid debug step so seems like a good addition here, thanks!

PR #9353, opened 2023-01-31
[fs-detectors] Add fix for if monorepo and proj at root

chloetedder:
If there is a monorepo that has one project at the root level we want to return commands without the relative to root prefixes


packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
```diff
@@ -64,15 +64,28 @@ export async function getMonorepoDefaultSettings(
       throw new MissingBuildPipeline();
     }
 
-    return {
+    const rootSettings = {
```
cb1kenobi: Instead of creating multiple variables, what do you think of checking `if (projectPath === '/')` and returning early? In the case when `projectPath === '/'`, then it wouldn't need to evaluate `nonRootSettings`.

chloetedder: Agreed 💯

packages/cli/src/util/build/monorepo.ts
```diff
@@ -54,8 +54,10 @@ export async function setMonorepoDefaultSettings(
       )} monorepo manager. Attempting to assign default settings.`
     );
 
-    setCommand('buildCommand', commands.buildCommand);
-    setCommand('installCommand', commands.installCommand);
+    if (commands.buildCommand)
```
cb1kenobi: The coding style doesn't require curly braces for all block scopes? I think for consistency we should have curly braces around all block scopes.

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
```diff
@@ -21,12 +21,19 @@ export class MissingBuildTarget extends Error {
   }
 }
 
+type MonorepoDefaultSettings = {
+  buildCommand: string | null;
+  installCommand: string | null;
```
cb1kenobi: Should we also allow `undefined` just in case?

cb1kenobi (changes requested):

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
chloetedder: Yeah I am good with having that for backup 👍

cb1kenobi (approved):
Looks awesome! Thank you!

cb1kenobi (approved):

PR #9354, opened 2023-01-31
[static-build] Add Gatsby `GENERATED_FILE_COMMENT` for non-existing config file cases

TooTallNate:
Follow-up to #9312 which missed these two cases.


packages/static-build/test/gatsby.test.ts
```diff
@@ -65,9 +65,10 @@ describe('gatsby utilities', () => {
     const [gatsbyConfig] = await Promise.all([
       fs.readFile(path.join(fixture, 'gatsby-config.js'), 'utf-8'),
     ]);
-    expect(gatsbyConfig).toMatchInlineSnapshot(
-      `"module.exports = {\\"plugins\\":[\\"@vercel/gatsby-plugin-vercel-analytics\\"]}"`
-    );
+    expect(gatsbyConfig).toMatchInlineSnapshot(`
+      "// This file was generated by @vercel/static-build
+      module.exports = {\\"plugins\\":[\\"@vercel/gatsby-plugin-vercel-analytics\\"]}"
```
cb1kenobi, line 70: Now that this is a multiline string, should the leading whitespace be trimmed?
```suggestion
module.exports = {\\"plugins\\":[\\"@vercel/gatsby-plugin-vercel-analytics\\"]}"
```

TooTallNate: It's generated directly by Jest so I guess not.

cb1kenobi: I'm not sure I follow. What is Jest generating?

It looks like the code in `gatsby.ts` is writing a multiline string with no leading whitespace to the `gatsby-config.js` file. I would expect the test to read that file containing no leading whitespace which should error when it finds leading whitespace, unless `toMatchInlineSnapshot()` isn't strict?

TooTallNate: Ya I think Jest does some formatting on the result string. The value here was generated by `jest -u` is what I mean though, and the unit tests are happy with the value, so I'm not gonna touch it.

cb1kenobi (approved):

PR #9356, opened 2023-01-31
[gatsby-plugin-vercel-builder] Move "public" dir instead of copying

TooTallNate:
We are getting reports of some larger Gatsby projects running out of disk space during build which seem to be the result of copying the static directory multiple times.

So move instead of copying, which doesn't fully solve the problem, but it does kick the can down the road a bit. Very large projects would probably still hit the disk limit at some point. Anyways, moving is faster than copying so this is still a net positive.


packages/gatsby-plugin-vercel-builder/src/handlers/build.ts
```diff
@@ -88,10 +87,6 @@ export async function copyFunctionLibs({
         src: join('.cache', 'page-ssr'),
         dest: join(functionDir, '.cache', 'page-ssr'),
       },
-      // {
-      //   src: join(functionDir, '.cache', 'query-engine', 'assets'),
-      //   dest: join(functionDir, 'assets'),
-      // },
```
styfle, line 94: Should this comment be removed?

TooTallNate: Ya, was just cleaning up some dead code while in here.

cb1kenobi: According to https://github.com/vercel/vercel/commit/ae89b8b8bee068cf9499867d6c794cb2ccc47c24#diff-3b89128206f28715de44e9a753ba57ac89f9ea0b484035ca0d74c0a3f8d9204bR98-R101, these lines have always been commented out. Sounds like they're not needed.

cb1kenobi (dismissed):

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
```diff
@@ -1,14 +1,20 @@
...
     'static',
     prefix ?? ''
   );
-  await ensureDir(targetDir);
-  await copy(join(process.cwd(), 'public'), targetDir);
+  await ensureDir(dirname(targetDir));
+  try {
+    await move(publicDir, targetDir);
+  } catch (err: any) {
+    console.log('failed to move "public" dir', err);
```
styfle:
```suggestion
    console.error('failed to move "public" dir', err);
```

styfle (dismissed):

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
```diff
@@ -1,14 +1,20 @@
...
     'static',
     prefix ?? ''
   );
-  await ensureDir(targetDir);
-  await copy(join(process.cwd(), 'public'), targetDir);
+  await ensureDir(dirname(targetDir));
+  try {
+    await move(publicDir, targetDir);
+  } catch (err: any) {
+    console.error('failed to move "public" dir', err);
```
EndangeredMassa: **suggestion (non-blocking):**

```suggestion
    console.error(`Failed to move "public" dir from "${publicDir}" to "${targetDir}". Copying instead.`, err);
```

EndangeredMassa (dismissed):

Ethan-Arrowood (approved):

cb1kenobi (approved):

PR #9358, opened 2023-01-31
[cli] Merge build output contents instead of copying

TooTallNate:
When Root Directory setting is used, and/or `vc build --output` is used, do an intelligent "merge" (move) operation rather than copying the contents of the build output directory to the destination. This should overall be faster and avoid disk space issues for larger projects.


cb1kenobi (dismissed):
LGTM

cb1kenobi (approved):

PR #9359, opened 2023-01-31
[tests] Remove `actions/cache` since its actually slower than no cache

styfle:
This PR removes `actions/cache` since it is actually **slower** than no cache.

## ubuntu-latest
- [with cache](https://github.com/vercel/vercel/actions/runs/4059017971/jobs/6986950123): 21s + 3s + 1m4s = 88s
- [without cache](https://github.com/vercel/vercel/actions/runs/4059216913/jobs/6987026470): 44s

## macos-latest
- [with cache](https://github.com/vercel/vercel/actions/runs/4059017971/jobs/6986950275): 2m55s + 8s + 3m20s = 383s
- [without cache](https://github.com/vercel/vercel/actions/runs/4059216913/jobs/6987026554): 2m21s = 141s

## windows-latest
- [with cache](https://github.com/vercel/vercel/actions/runs/4059017971/jobs/6986950359): 1m10s + 6s + 3m29s = 285s
- [without cache](https://github.com/vercel/vercel/actions/runs/4059216913/jobs/6987026661): 1m25s = 85s


ijjk (dismissed):

TooTallNate (dismissed):

cb1kenobi (approved):
Unbelievable.

TooTallNate (approved):

tsamantanis:
I think this change has increased the build times of one of my medium sized projects (couple hundred pages). After this change was released, the build times more than doubled from around 4 minutes to 9+. The major problem is that the site utilizes a source plugin and there are tasks such as `Running gatsby.IMAGE_CDN jobs` and `gatsby-plugin-sharp.IMAGE_PROCESSING` which are very time consuming and need to run again. 

Is it possible to specify the version of `@vercel/gatsby-plugin-vercel-builder` used for the build? I'd like to somehow test using `v1.0.3`. I tried adding the package in `package.json` but then it runs twice and throws an error.

PR #9344, opened 2023-01-30
[cli] Add `crons` to build output

AndyBitz:
- [x] Updates `build-utils` to define the `crons` type
- [x] Updates `vercel build` to consider the `cron` property and output the `crons` config
- [x] Updates  `next` to consider the crons manifest file
- [x] Updates `node` to consider the `cron` property
- [x] Updates `static-config` to consider the `cron` property
- [x] Add tests

PR #9360, opened 2023-01-31
[cli] Add crons to build output

AndyBitz:
Alternative implementation for https://github.com/vercel/vercel/pull/9344

Ensures that Node.js Serverless Functions and Edge Functions correctly output the `cron` property for use by the Build Output API.


packages/build-utils/src/types.ts
```diff
@@ -428,6 +430,7 @@ export interface BuildResultV2Typical {
   framework?: {
     version: string;
   };
+  crons?: Cron[];
```
TooTallNate: I think we want to remove this one? Can be added on the server-side if it's needed there.

```suggestion
```

packages/build-utils/src/types.ts
```diff
@@ -436,6 +439,7 @@ export interface BuildResultV3 {
   // TODO: use proper `Route` type from `routing-utils` (perhaps move types to a common package)
   routes?: any[];
   output: Lambda | EdgeFunction;
+  cron?: string;
```
TooTallNate: This is directly on `Lambda`/`EdgeFunction`, ya?

```suggestion
```

packages/cli/src/util/build/write-build-result.ts
```diff
@@ -221,10 +234,7 @@ async function writeBuildResultV3(
       })
     : {};
 
-  const ext = extname(src);
-  const path = stripDuplicateSlashes(
-    build.config?.zeroConfig ? src.substring(0, src.length - ext.length) : src
-  );
+  const path = getBuildV3Path(build);
```
TooTallNate: I'd recommend removing the helper function if it's only being used once here, to keep the diff less noisy.

TooTallNate (changes requested):

TooTallNate (approved):

PR #9362, opened 2023-02-01
[gatsby-plugin-vercel-builder] Use `pathPrefix` from BuildArgs

TooTallNate:
Instead of using `pathPrefix` from the Redux store, grab the value directly from the "BuildArgs" passed to `onPostBuild()`. This value seems to properly reflect whether or not `--prefix-path` mode was enabled during the Gatsby build process.


Ethan-Arrowood (dismissed):

cb1kenobi (approved):
LGTM

Ethan-Arrowood (approved):

PR #9364, opened 2023-02-01
[cli] add charset to content-type headers for `vc dev`

Ethan-Arrowood:
Adds `charset: utf8` to the `content-type` header


packages/cli/test/dev/integration-5.test.ts
```diff
@@ -447,3 +447,13 @@ test(
     }
   })
 );
+
+test(
+  '[vercel dev] should set charset in content-type header',
+  testFixtureStdio('test-tomato', async (_: any, port: any) => {
+    const response = await fetch(`http://localhost:${port}/`);
+    expect(response.headers.get('content-type')).toBe(
+      'text/html; charset=utf-8'
```
styfle: Can we add this assertion to an existing test so we can avoid adding a new deployment (which will slow down our tests)

TooTallNate (approved):

PR #9368, opened 2023-02-02
[build-utils][node][static-config] Extend `cron` property

AndyBitz:
Extends the `cron` property so it can  be an object with additional options or an array of objects to create multiple cronjobs per function.


packages/cli/test/fixtures/unit/commands/build/with-cron/api/custom.js
```diff
@@ -0,0 +1,20 @@
+export const config = {
+  runtime: 'edge',
+  cron: {
```
vvo: I have vague memories of people saying we should move from `export const config` to independent `export const runtime`, `export const cron` and so on. Perhaps it's no more planned to do so?

packages/cli/test/fixtures/unit/commands/build/with-cron/api/custom.js
```diff
@@ -0,0 +1,20 @@
...
+    expression: '* * * * *',
+    path: '/api/custom',
```
vvo: I still don't get why there's a need to provide a path.
When would there be a case where you want your cron to be triggered on a different path than the main route to the filename?

i.e. if 100% of the time that path will be the $root/[pages]/$path then I am unsure why we need it, aside from internal implementation details we may not want to surface to the customer API level. Happy to discuss this in a quick sync.

PR #9371, opened 2023-02-03
[examples] Update Astro example to v2

swarnava:


styfle (commented):
This is swapping from pnpm to yarn, is that intentional?

I think we should stick with pnpm since thats what it was using before.

swarnava:
For some reason, I pnpm-lock file isn't being generated when I run `pnpm install`. any idea why?

styfle:
Could be corepack. You could try it in a separate directory first and copy the lockfile and package.json over. Also could try updating pnpm.

styfle (approved):
Thanks!

PR #9373, opened 2023-02-03
[node] Add log for unused config runtime

styfle:
Both `runtime: undefined` and `runtime: nodejs` have the same behavior but we want to change that in the future.

The first step is to see if `runtime: nodejs` is currently used at all.

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -413,6 +413,11 @@ export const build: BuildV3 = async ({
         )} (must be one of: ${JSON.stringify(ALLOWED_RUNTIMES)})`
       );
     }
+    if (staticConfig.runtime === 'nodejs') {
+      console.log(
+        `Detected unused static config runtime "nodejs" in "${entrypointPath}"`
+      );
+    }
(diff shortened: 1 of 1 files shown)


packages/node/src/index.ts
```diff
@@ -413,6 +413,11 @@ export const build: BuildV3 = async ({
         )} (must be one of: ${JSON.stringify(ALLOWED_RUNTIMES)})`
       );
     }
+    if (staticConfig.runtime === 'nodejs') {
```
EndangeredMassa, line 416: **suggestion (non-blocking):** Just in case we forget to come back to remove this, do you want to add a comment?

```suggestion
    // TODO: Remove this after we learn how often `nodejs` is
    // explicitly set as a `runtime` value
    if (staticConfig.runtime === 'nodejs') {
```

EndangeredMassa (approved):

packages/node/src/index.ts
styfle: I actually think we should turn it into a warning or error depending on the results we find

PR #9374, opened 2023-02-03
[next][build-utils] Add support for cron to vercel.json

AndyBitz:
Extends `vercel.json` to support setting `cron` on `functions`.

Supports Serverless Functions and Edge Functions.


packages/build-utils/src/schemas.ts
```diff
@@ -31,6 +31,11 @@ export const functionsSchema = {
           type: 'string',
           maxLength: 256,
         },
+        cron: {
+          type: 'string',
+          minLength: 9,
```
Ethan-Arrowood, line 36: Does this represent "* * * * *"? Is that the minimum possible cron string?

packages/next/test/unit/build.test.js
```diff
@@ -0,0 +1,49 @@
+const path = require('node:path');
+const rimraf = require('rimraf');
```
Ethan-Arrowood: Don't we have fs-extra for this?

Ethan-Arrowood (commented):
This is looking really good! Excited for CRON job support 😁

packages/build-utils/src/schemas.ts
AndyBitz: Exactly, 5 fields + 4 spaces is the minimum

packages/next/test/unit/build.test.js
AndyBitz: Ah, good point, forgot about that 👀

Ethan-Arrowood (approved):

packages/next/test/unit/build.test.js
```diff
@@ -0,0 +1,44 @@
+const path = require('node:path');
```
vvo, line 1: The name of the test is generic (= no cron specific), was that a missing test for next?

vvo (approved):

packages/next/test/unit/build.test.js
AndyBitz: It is sort of a new way to test the next builder. The other unit tests, test some specific functions from the package, while the integration tests create actual Deployments on Vercel. This is sort of a middle ground, where it tests the main build function from the package, but without creating a Deployment.

PR #9011, opened 2022-12-05
[remix] use `@remix-run/dev` version when `@remix-run/vercel` package is missing

mcansh:
deploying a remix app without explicitly having the `@remix-run/vercel` package installed will cause an older version (1.6.1) to be installed which then results in some runtime issues as we end up with multiple copies of remix packages due to us using a pinned version in each of our packages

this PR removes the hardcoded 1.6.1 version, and instead will get the version from `@remix-run/dev`

Signed-off-by: Logan McAnsh <logan@mcan.sh>

### Related Issues

> closes #9010 
> closes https://github.com/remix-run/remix/issues/4759

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #8784, opened 2022-10-26
[remix] add edge runtime support

theMosaad:
### Feature
Support deploying Remix to Vercel's edge functions.

### Demo
This example is built locally with this patch: https://github.com/theMosaad/remix-vercel-edge

This also fixes https://github.com/vercel/vercel/issues/8788

P.S. Please advise if there's a more accurate way of emulating Vercel's production edge runtime locally.

PR #9375, opened 2023-02-03
[remix] Support per-page Edge functions and replace "@remix-run/vercel"

TooTallNate:
Adds support for server-side rendered Remix using Edge Functions. This runtime can be enabled on a per-page basis, by adding the following to a page within the `app/routes` directory:

```js
export const config = {
  runtime: 'edge'
};
```

Additionally, this PR further supersedes the `@remix-run/vercel` runtime adapter, because we will always inject our own server entrypoint. So the logic to ensure that package exists in the project's `package.json` has been removed (so this closes #9011). The only requirement is that the Remix project has `@remix-run/node` as a dependency, which is the case for the vanilla Remix template so I think that's a fair assumption.

To make Edge Functions work, we need to ensure that `remix build` is executed with a few specific configuration values in place, so this change wraps the existing `remix.config.js` file and adds our own to make sure those values are in place (and then cleans itself up after the build command is executed).

Finally, the reading of the Remix config logic was simplified by using the `readConfig()` function from the `@remix-run/dev` package, which also includes the routes manifest, so the hacky `vm` running logic to retrieve the manifest was able to be removed.

Closes #8784.
Closes #9011.

---

# To test this out:

1. Ensure that Remix dependencies are running _at least version `1.5.0`_ (which is when `writeReadableStreamToWritable()` was added to `@remix-run/node`).

2. Add the following line of code to any page in your Remix application’s `app/routes/*` directory:
    
    ```jsx
    export const config = { runtime: 'edge' };
    ```
    
3. Make a deployment, either by running `vercel deploy` in the CLI, or by pushing a Git commit to your repository which has a Vercel Git integration enabled.


packages/remix/server-edge.mjs
```diff
@@ -0,0 +1,4 @@
+import { createRequestHandler } from '@remix-run/server-runtime';
+import build from './index.js';
+//export default createRequestHandler(build, process.env.NODE_ENV);
```
Ethan-Arrowood:
```suggestion
```

Ethan-Arrowood (dismissed):

Ethan-Arrowood (dismissed):

packages/remix/server-edge.mjs
```diff
@@ -0,0 +1,3 @@
+import { createRequestHandler } from '@remix-run/server-runtime';
+import build from './index.js';
+export default createRequestHandler(build);
```
cramforce, line 3: Beautiful

theMosaad:
FYI, when deploying with this, I always get the below warning and the build time takes a lot longer (3-6 minutes compared to 1 minute with the old adapter)

```
Building Remix app in production mode...
--
11:28:48.166 | Built in 625ms
11:29:00.343 | Build Completed in /vercel/output [18s]
11:30:14.763 | Warning: The log length exceeded the limit of 4 MB. Learn more: https://vercel.link/build-log-size-limit
```

Here are my app's routes (generated via `npx remix routes`):

```
<Routes>
  <Route file="root.tsx">
    <Route path="forms/icon-stroke-width" file="routes/forms/icon-stroke-width.tsx" />
    <Route path="forms/color-scheme" file="routes/forms/color-scheme.tsx" />
    <Route path="forms/icon-color" file="routes/forms/icon-color.tsx" />
    <Route path="forgot-password" file="routes/forgot-password.tsx" />
    <Route path="reset-password" file="routes/reset-password.tsx" />
    <Route file="routes/__dashboard.tsx">
      <Route path="favorites" index file="routes/__dashboard/favorites/index.tsx" />
      <Route path="icon-set" file="routes/__dashboard/icon-set.tsx">
        <Route path=":iconSetSlug" file="routes/__dashboard/icon-set/$iconSetSlug.tsx" />
      </Route>
    </Route>
    <Route path="icon/:slug" file="routes/icon.$slug.tsx" />
    <Route path="onboarding" file="routes/onboarding.tsx" />
    <Route path="logout" file="routes/logout.tsx" />
    <Route path="signup" file="routes/signup.tsx" />
    <Route index file="routes/index.tsx" />
    <Route path="login" file="routes/login.tsx" />
  </Route>
</Routes>
```

cb1kenobi:
Still running into issues with my Remix 1.4.3 site:

```
2023-02-06T16:37:27.535Z	036b1a48-6d0d-40c6-9076-367558f7c19f	ERROR	file:///var/task/build/server-node.mjs:6
  writeReadableStreamToWritable,
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'writeReadableStreamToWritable' not found. The requested module '@remix-run/node' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
import pkg from '@remix-run/node';
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:438:15)
    at async getListener (/var/task/___vc/__launcher.js:37:11)
RequestId: 036b1a48-6d0d-40c6-9076-367558f7c19f Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

Note that I'm doing `vc build --prod && vc deploy --prod --prebuilt`. Also note that my site's `package.json` does NOT have `@remix-run/node` as a dependency.

TooTallNate:
@theMosaad If you wanna share a deployment URL with me then we can investigate the build times further.

TooTallNate:
After doing some investigation with @cb1kenobi, we discovered that `@remix-run/node` needs to be at least on version v1.5.0. I've updated the testing instructions in the PR description to account for this.

theMosaad:
@TooTallNate here's a deployment with the old adapter: https://vercel.com/notion-icons/notionicons/7TuMpCn42zkaUiyWKkRA2V76BEF9 which took 1m 4s
and here's a deployment with this adapter: https://vercel.com/notion-icons/notionicons/FeA3owjJXJSMV2CUnxbwDtGbq7hS which took 6m 41s

Happy to create a minimal reproducible example with a public repo if needed. Can also share the current private repo with you and the team.

Ethan-Arrowood (approved):

WesleyYue:
Can you describe how the `runtime: "edge"` config works behind the scenes? We need fine-grain control over the exact response served to [workaround hydration bugs and for Chakra](https://github.com/remix-run/remix/discussions/5244#discussioncomment-5029201) (same with any emotion based kits). I can see that the source in `remix-entry-server` is pretty straightforward. Does Vercel just try to execute the entry.server.[node|edge].ts files based on this config?

ie. can I copy the source from remix-entry-server, make the modifications I need to the exact response, and put them at app/entry.client.[node|edge].ts and have it work? Or is it hardcoded to @vercel/remix-entry-server?

wonu:
@TooTallNate This change causes deploy problem when any empty paths exist.

For example,
```ts
<Routes>
  <Route file="root.tsx">
    <Route index file="routes/index.tsx" />
    <Route file="routes/__foo.tsx" />
  </Route>
</Routes>
```
```ts
{
  path: undefined,
  index: undefined,
  caseSensitive: undefined,
  id: 'routes/__foo',
  parentId: 'root',
  file: 'routes/__foo.tsx'
}
```
such route will have path as `''` from `getPathFromRoute()`, and `output` will be:
```ts
{
  '404': NodejsLambda,
  ...,
  '': NodejsLambda // <--
}
```

It should make symlink on `.../.func`, but `stripDuplicateSlashes('')` returns `'.'` so the symlink is created on `.../..func`, which causes the following error:
```sh
[Error: ENOENT: no such file or directory, readlink '/vercel/output/functions/.func'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'readlink',
  path: '/vercel/output/functions/.func'
}
```
(related [discussion#1587](https://github.com/vercel/community/discussions/1587#discussion-4880305))

Before this PR(`v28.15.5`), the empty path became `'index'` as a fallback in [`getSsrRoutes()`](https://github.com/vercel/vercel/blob/vercel%4028.15.4/packages/remix/src/build.ts#L322), so it was no problem.

P.S. The error has changed since `v28.16.5` as in #9524, but the above problem seems not related to this so it should be treated as a separate issue I think.

TooTallNate:
@WesleyYue The `@vercel/remix-entry-server` needs to be a standalone package within node_modules. This is due to how the Remix compiler bundles files directly in the project, but _does not_ bundle node_modules. This allows Vercel to utilize different entrypoints in the package based on running in Node vs. Edge.

So to answer your question, no you wouldn't be able to just copy those files into your project for modification, since Remix doesn't know which one to use during its compilation process. You could submit a PR to `@vercel/remix-entry-server` for any potential fixes though.

TooTallNate:
@wonu Let's continue the discussion in https://github.com/vercel/community/discussions/1587. If you have a reproduction, that would be helpful as well.

WesleyYue:
@TooTallNate gotcha, so sounds like if we have a monorepo setup, we should be able to put our own modified entry.server.[node|edge].ts files in a separate package, and set up the appropriate main/browser settings in package.json so that remix/vercel pulls in the right one?

I don't think you would want the changes we have in entry.server, since it's specific to Chakra, and the hydration workaround has non-obvious trade-offs.

TooTallNate:
Yes, using a monorepo would work for that case.

TrySound:
Does this mean `@vercel/remix` still will be auto installed on deploy?

I get this warning on commit build and after I have install with --frozen-lockfile which fails because vercel added the  dependency.

```
Warning: Adding "@remix-run/vercel" v1.6.1 to `dependencies`. You should commit this change.
```

PR #9377, opened 2023-02-03
[node] Make NFT resolve "worker"/"browser" exports for Edge Functions

TooTallNate:
This makes `react-dom` work as expected from within Edge Functions. Otherwise, NFT will only select the Node.js version of the files which do not work with react-dom within an Edge Function.


packages/node/src/index.ts
```diff
@@ -180,13 +180,16 @@ async function compile(
     return source;
   }
 
+  const conditions = isEdgeFunction ? ['worker', 'browser'] : undefined;
```
styfle, line 183: Where did you find these conditions defined? I thought we should be using `edge-light` according to the standard https://runtime-keys.proposal.wintercg.org/

cc @Ethan-Arrowood who wrote the standard

TooTallNate: Ya we can add `edge-light`. Needs to be supported on the server-side still though, so we may want to wait for that and do in another PR.

Ethan-Arrowood: :+1: we should do that in another PR. We also need to update edge docs to include that that key should be added

Ethan-Arrowood (approved):

PR #9388, opened 2023-02-06
[build-utils] Add passQuery field to Prerender

ijjk:
This adds a new `passQuery` field to `Prerender` outputs to allow opting into new behavior for passing query/route matches in the URL instead of the `x-now-route-matches` header. 

x-ref: [slack thread](https://vercel.slack.com/archives/C02K2HCH5V4/p1675704226492169)
x-ref: [slack thread](https://vercel.slack.com/archives/C02K2HCH5V4/p1660070208792189)

### DIFF
--- a/packages/build-utils/src/prerender.ts
+++ b/packages/build-utils/src/prerender.ts
@@ -10,6 +10,7 @@ interface PrerenderOptions {
   allowQuery?: string[];
   initialHeaders?: Record<string, string>;
   initialStatus?: number;
+  passQuery?: boolean;
 }
 
 export class Prerender {
@@ -22,6 +23,7 @@ export class Prerender {
   public allowQuery?: string[];
   public initialHeaders?: Record<string, string>;
   public initialStatus?: number;
+  public passQuery?: boolean;
 
   constructor({
     expiration,
@@ -32,6 +34,7 @@ export class Prerender {
     allowQuery,
     initialHeaders,
     initialStatus,
+    passQuery,
   }: PrerenderOptions) {
     this.type = 'Prerender';
     this.expiration = expiration;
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/prerender.ts
```diff
@@ -52,6 +55,17 @@ export class Prerender {
     }
     this.group = group;
 
+    if (passQuery === true) {
+      this.passQuery = true;
+    } else if (
+      typeof passQuery !== 'boolean' &&
+      typeof passQuery !== 'undefined'
+    ) {
+      throw new Error(
```
cb1kenobi, line 64: This should throw a `TypeError`.

It would be rad to have a test that passes in something that isn't `undefined`, `true`, or `false`.

cb1kenobi (dismissed):

packages/build-utils/src/prerender.ts
ijjk: Hmm not sure if we should use `TypeError` specifically as that deviates from the existing checks, can add an assertion to make sure it throws as expected though.

ijjk: Added in https://github.com/vercel/vercel/pull/9388/commits/47f939fcb599d76c248b46d321ea58993db2dd0f

cb1kenobi: A `TypeError` is specifically for type errors which is exactly what this code is testing for.

For example, checking that the `bypassToken` has a length < 32 is not a `TypeError` because it's validating the variable contents, not its type.

Technically `fallback` and `initialStatus` should also throw `TypeError`. At the end of the day, `TypeError` is still an instanceof `Error` and shouldn't cause an issue.

ijjk: Would prefer to keep changes like this in a separate PR to reduce risk as there can be unexpected side effects from `Error` or `TypeError` being serialized/deserialized. 

Additional risk is being assumed as we aren't currently using `TypeError` anywhere in `build-utils` so no prior guarantees this just works throughout the pipeline.

styfle (approved):

styfle:
@cb1kenobi Ideally, we wouldn't be doing this validation in the class at all. It should validate in the API. The validation in the class is just for convenience if you're writing a builder and have a typo. But most users already use TS so these type checks are useless.

PR #9392, opened 2023-02-07
[cli] Update notification redesign

cb1kenobi:
This PR does a couple fantastic things:

1. Increases latest version check from once a week to once a day
2. Increases latest version notification from once a week to once every 3 days unless an even newer version is available
3. Update changelog link to all releases and make text an actual link
4. Redesign of the update notification to be more visible by wrapping in a box
5. Replace `boxen` (500KB) (used by `vc bisect`) with slimmed down built-in version (3.4KB)
6. Update notification appears at the end of the command regardless if an error occurred

Regular notification: displayed once every 3 days or sooner if a new release first day:

image

Same as above, but formats the notification to adapt to narrow terminals:

image

Whenever any command (aside from `help`) returns a non-zero exit code, show an additional message to encourage updating:

image

Linear: https://linear.app/vercel/issue/VCCLI-504/investigate-old-update-notifier-behavior


packages/cli/src/index.ts
```diff
@@ -716,7 +695,39 @@ process.on('unhandledRejection', handleRejection);
...
+                  'may'
+                )} fix any errors that occurred.`
+              )
+            : '';
+        output.print(
+          box(
+            `Update available! ${chalk.gray(`v${pkg.version}`)} ≫ ${chalk.green(
+              `v${latest}`
+            )}
+Changelog: ${output.link(changelog, changelog, { fallback: () => changelog })}
```
EndangeredMassa: **suggestion (soft-blocking):** remove fallback

The fallback doesn't need to be specified if the text is the same as the original.

```suggestion
Changelog: ${output.link(changelog, changelog)}
```

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -163,12 +163,27 @@ process.once('message', async msg => {
...
+    try {
+      const old = JSON.parse(await readFile(cacheFile, 'utf-8'));
+      notifyAt =
+        old?.version && old.version === version ? old?.notifyAt : undefined;
+    } catch (err) {
+      // cache does not exist or malformed
+      if (err.code !== 'ENOENT') {
+        output.debug(`Error reading latest package cache file: ${err}`);
+      }
+    }
```
EndangeredMassa: **suggestion (non-blocking):** extract function

I think this file could benefit from some extracted functions, but at least for this addition, can we extract a function?

Something like:

```
const notifyAt = await getNotifyAt(output, cacheFile, version);
```

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -67,25 +69,26 @@ export default function getLatestVersion({
...
+    (!cache.notifyAt || cache.notifyAt < Date.now()) &&
     pkg.version &&
+    cache.version &&
     semver.lt(pkg.version, cache.version)
   ) {
-    cache.notified = true;
+    cache.notifyAt = Date.now() + notifyInterval;
     outputJSONSync(cacheFile, cache);
     return cache.version;
   }
```
EndangeredMassa: **suggestion (non-blocking):** simplify conditional expression

This conditional has a complex expression already. I think it will be easier to understand if we extract some named things, like:

```js
function isPkgOld(pkg, cache) {
  if (!pkg?.version || !cache?.version) {
    return false;
  }
  
  return semver.lt(pkg.version, cache.version);
}

const currentPkgIsOld = isPkgOld(pkg, cache);
const shouldNotify = !cache.notifyAt || Date.now() > cache.notifyAt;

if (currentPkgIsOld && shouldNotify) {
  cache.notifyAt = Date.now() + notifyInterval;
  outputJSONSync(cacheFile, cache);
  return cache.version;
}
```

packages/cli/test/unit/util/output/box.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(stripAnsi(result)).toEqual(
+      `
+──────────────────────────────────────────
+
+This is a really, really long line of text
+
+──────────────────────────────────────────
+    `.trim()
+    );
+  });
```
EndangeredMassa, line 123: **requirement (blocking):** when text is longer than the number of terminal columns, the text should wrap within the narrow rules at the top and bottom

What seems to happen when the text is too long is that the top/bottom rules keep going, also wrapping:

Screenshot 2023-02-07 at 2 17 18 PM

We should add a test for this case and make sure it works.

packages/cli/src/util/output/box.ts
```diff
@@ -0,0 +1,109 @@
...
+            leftPadding += maxLine - len;
+          } else if (textAlignment === 'left') {
+            rightPadding += maxLine - len;
+          }
+        }
+
+        return `${borderColorFn(left)}${' '.repeat(
+          leftPadding
+        )}${line}${' '.repeat(rightPadding)}${borderColorFn(right)}`;
+      })
```
EndangeredMassa: **suggestion (non-blocking):** extract function

Can we extract a function for this `map` to use to make this return expression clearer?

Something like:

```js
function generateLineProcessor(left, right) {
  return ([line, len]) => {
    let leftPadding = 0;
    let rightPadding = 0;

    if (!slim) {
      leftPadding = sidePadding;
      rightPadding = sidePadding;

      if (textAlignment === 'center') {
        leftPadding += Math.floor((maxLine - len) / 2);
        rightPadding += maxLine - len - leftPadding + sidePadding;
      } else if (textAlignment === 'right') {
        leftPadding += maxLine - len;
      } else if (textAlignment === 'left') {
        rightPadding += maxLine - len;
      }
    }

    return `${borderColorFn(left)}${' '.repeat(
      leftPadding
    )}${line}${' '.repeat(rightPadding)}${borderColorFn(right)}`;
  }
}

const processedLines = lines
  .map(generateLineProcessor(left, right))
  .join('\n');

return (
  borderColorFn(`${topLeft}${hr.repeat(innerWidth)}${topRight}`) +
  '\n' +
  spacerRow +
  processedLines +
  '\n' +
  spacerRow +
  borderColorFn(`${bottomLeft}${hr.repeat(innerWidth)}${bottomRight}`)
);
```

packages/cli/src/util/get-latest-version/index.ts
```diff
@@ -67,25 +69,26 @@ export default function getLatestVersion({
     }
   }
 
-  if (!cache || cache.expireAt < Date.now()) {
+  if (!cache || !cache.expireAt || cache.expireAt < Date.now()) {
     spawnWorker(
```
EndangeredMassa, line 73: **question (soft-blocking):** do we need `spawnWorker` to pass along the No Color env vars and CLI argument in some form to make sure the border of the box is not colored when color is disabled?

That PR was only recently merged in: https://github.com/vercel/vercel/pull/8826

packages/cli/test/unit/util/get-latest-version.test.ts
```diff
@@ -130,6 +130,42 @@ describe('get latest version', () => {
...
+    // 2. get the latest version
+    let latest = getLatestVersion({
+      cacheDir,
+      pkg,
+    });
+    expect(latest).toEqual('28.0.0');
+
+    // we need to wait up to 20 seconds for the cacheFile to be updated
+    for (let i = 0; i < 80; i++) {
+      await sleep(250);
```
EndangeredMassa, line 153: **suggestion (non-blocking):** reduce sleep

For this to work, we only need to relinquish control flow, right? Can we `sleep(0)` instead?

packages/cli/test/unit/util/get-latest-version.test.ts
```diff
@@ -130,6 +130,42 @@ describe('get latest version', () => {
...
+      try {
+        const cache = await fs.readJSON(cacheFile);
+        if (cache.version !== '28.0.0') {
+          break;
+        }
+      } catch {
+        // cacheFile has not been updated yet
+      }
+    }
```
EndangeredMassa, line 165: **suggestion (soft-blocking):** throw explicit error on timeout

If the timeout expires, we already know that we've failed. Can we throw a clearer error message?

EndangeredMassa (changes requested):
This looks awesome! I think we have some changes to make before merging, though.

packages/cli/test/unit/util/output/box.test.ts
cb1kenobi: This is by design. I think it's outside the scope to implement word wrapping. Proper word wrapping requires unicode handling and should handle hyphenation. In the context of an update available notification, any custom word wrapping solution would unlikely improve the actual messaging.

packages/cli/src/util/get-latest-version/index.ts
cb1kenobi: Cool. I simplified it without needing to do a function.

packages/cli/src/index.ts
cb1kenobi: Fixed! Due to the implementation, I need to set `fallback: false`. Simply passing in `undefined` won't work.

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: Normally I would agree, but this is a `.js` file, not a `.ts` file, and this `get-latest-worker.js` is not bundled, but rather gets copied at build time into the dist directory. Because of this and because of past issues on dependencies, `get-latest-worker.js` is designed to not have any non-Node.js dependencies, for better or worse.

packages/cli/src/util/output/box.ts
cb1kenobi: Awesome, applied a similar simplification. Check it out.

packages/cli/src/util/get-latest-version/index.ts
cb1kenobi: `get-latest-worker.js` no longer uses colors because it no longer has any external dependencies. There shouldn't be any instance where colorized text is rendered to stdio.

packages/cli/test/unit/util/get-latest-version.test.ts
cb1kenobi: This isn't kernel level `sleep()`. This is the `sleep()` function in the `cli/src/util/sleep.ts` file and it's just a promise with a `setTimeout()`. During the `sleep()`, it's making an `fetch()` which is known to take up to 10 seconds in CI.

packages/cli/test/unit/util/get-latest-version.test.ts
cb1kenobi: Done!

packages/cli/test/unit/util/output/box.test.ts
EndangeredMassa: I think we can get a bit closer by constraining the border width to the column count. I tested this out:

```ts
const maxLine = getMaxWidth(lines, cols);

// ...

/**
 * Find the max width of box border characters we should render based on
 * the width of the content and the terminal columns, whichever is smaller.
 */
function getMaxWidth(lines: [string, number][], terminalColumns: number) {
  const maxFromContent = lines.reduce((p, [, len]) => Math.max(p, len), 0);
  return Math.min(terminalColumns, maxFromContent);
}
```

Screenshot 2023-02-07 at 4 52 07 PM

packages/cli/src/util/get-latest-version/get-latest-worker.js
EndangeredMassa: I was suggesting extracting a function. It can live in this file. I don't we need a dependency here.

cb1kenobi: Ah, yes! I misunderstood, but now I see the 💡!

Not only is is this a good idea, I decided to bust the dist tag fetching out into a separate function too.

packages/cli/test/unit/util/output/box.test.ts
cb1kenobi: Gotcha! Fixed!

EndangeredMassa (approved):
nice work!

PR #9402, opened 2023-02-09
[next] Update env variable for output trace root

ijjk:
Since turbo and nx only consider `NEXT_PUBLIC_` prefixed env variables for the cache key this ensures we expose variables that affect the build in that format as well so that an incompatible build cache isn't attempted to be leveraged. No additional tests were added here as this related to build cache which can't be reliably tested here. 

x-ref: [slack thread](https://vercel.slack.com/archives/C03S8ED1DKM/p1675821643786319)

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -415,8 +415,13 @@ export const build: BuildV2 = async ({
     // this helps us catch cases where we can't locate the next.config.js
     // correctly
     env.NEXT_PRIVATE_TARGET = target;
+    env.NEXT_PUBLIC_TARGET = target;
   }
+  // Only NEXT_PUBLIC_ is considered for turbo/nx cache keys
+  // and caches may not have the correct trace root so we
+  // need to ensure this included in the cache key
   env.NEXT_PRIVATE_OUTPUT_TRACE_ROOT = baseDir;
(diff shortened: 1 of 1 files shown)


styfle (approved):

tknickman (approved):

tknickman:
Just for reference - here's a quick summary of how turbo infers its framework env vars. 

We include env vars with known framework prefixes from here (Next.js as an example):
https://github.com/vercel/turbo/blob/main/cli/internal/inference/inference.go#L35

However, we don't blanket include ALL env vars with that prefix. we support another var `TURBO_CI_VENDOR_ENV_KEY` that allows a CI vendor to set a prefix that we will exclude. This is done to avoid auto including vars like `NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA ` that would result in uncachable builds. 

For Vercel, we set that here:
https://github.com/vercel/vercel/blob/main/packages/build-utils/src/get-prefixed-env-vars.ts#L29 (in the Next.js case, this will be `NEXT_PUBLIC_VERCEL_`

And we're consuming it here:
https://github.com/vercel/turbo/blob/main/cli/internal/env/env.go#L60

So to wrap it all up, when a Next.js app is deployed on Vercel, all `NEXT_PUBLIC_` env's are included, but all `NEXT_PUBLIC_VERCEL_` env vars are excluded.

PR #9403, opened 2023-02-09
[tests] Add tests for `examples`

styfle:
These tests will run if one of the conditions is met:

- any change to `./examples`
- any changes to `./packages/frameworks`

The test ensures that examples can build and deploy without errors.

In a future PR, we could improve these tests to make requests to the test deployment (like we do for probes).

### DIFF
--- a/examples/__tests__/test-utils.ts
+++ b/examples/__tests__/test-utils.ts
@@ -2,6 +2,6 @@ import { basename, join } from 'path';
 import { testDeployment } from '../../test/lib/deployment/test-deployment.js';
 
 export async function deployExample(filename: string) {
-  const example = basename(filename).slice(0, -8);
+  const example = basename(filename).replace(/\.test\.ts$/, '');
   await testDeployment(join(filename, '..', '..', '..', example));
 }
--- a/examples/__tests__/test-utils.ts
+++ b/examples/__tests__/test-utils.ts
@@ -1,0 +1,7 @@
+import { basename, join } from 'path';
+import { testDeployment } from '../../test/lib/deployment/test-deployment.js';
+
+export async function deployExample(filename: string) {
+  const example = basename(filename).slice(0, -8);
+  await testDeployment(join(filename, '..', '..', '..', example));
+}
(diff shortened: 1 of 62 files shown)


examples/sveltekit/package.json
```diff
@@ -11,8 +11,8 @@
     "format": "prettier --write ."
   },
   "devDependencies": {
-    "@sveltejs/adapter-auto": "next",
-    "@sveltejs/kit": "1.0.0-next.589",
+    "@sveltejs/adapter-auto": "1.0.0-next.65",
+    "@sveltejs/kit": "1.0.0-next.428",
```
TooTallNate, line 15: Why was this one seemingly downgraded?

TooTallNate (dismissed):

examples/__tests__/test-utils.ts
```diff
@@ -0,0 +1,7 @@
+import { basename, join } from 'path';
+import { testDeployment } from '../../test/lib/deployment/test-deployment.js';
+
+export async function deployExample(filename: string) {
+  const example = basename(filename).slice(0, -8);
```
cb1kenobi: There's something about slicing that scares me. I don't immediately know what is being sliced.

```suggestion
  const example = basename(filename).replace(/\.test\.ts$/, '');
```

cb1kenobi (dismissed):

examples/sveltekit/package.json
styfle: Because it was failing. Someone upgraded it and it never worked. So I downgraded back to the last working version by looking at `git blame`. Previously, we didn't have tests for examples so it was possible to merge a PR without testing if it actually worked.

examples/__tests__/test-utils.ts
styfle: Done [862b1fa](https://github.com/vercel/vercel/pull/9403/commits/862b1fa6efa3ec5916d4b5a6f807977018fe94a8)

examples/sveltekit/package.json
styfle: Note: this change here is to the "Legacy Sveltekit" example, not to be confused with the new 1.0.0 example.

(This legacy example has the original `sveltekit` slug and the 1.0.0 example has the `sveltekit-1` slug)

cb1kenobi (approved):
LGTM

PR #9408, opened 2023-02-10
[next] Read rscContentType from routes manifest

timneutkens:
Related Next.js change: https://github.com/vercel/next.js/pull/45783


ijjk (approved):

ijjk:
Might be good to re-open as a branch on the repo instead of fork so tests can run here

timneutkens:
Landed it as-is as it doesn't affect the tests, searched for `vary` which is not tested.

PR #9409, opened 2023-02-10
[node] Fix incorrect stack trace of TS error

styfle:
### Description

TS has an option, `noEmitOnError`, that is used to either fail the build by throwing when `true` or just print the error and continuing to build when `false`.

This PR fixes the case when `noEmitOnError: false`. Previously showing an error stack trace but now it correctly shows the error message without the stack.

### Before

image

### After

image

### DIFF
--- a/packages/node/test/fixtures/46-noEmitOnError-false/vercel.json
+++ b/packages/node/test/fixtures/46-noEmitOnError-false/vercel.json
@@ -4,7 +4,8 @@
   "probes": [
     {
       "path": "/",
-      "mustContain": "no-emit-on-error-false:RANDOMNESS_PLACEHOLDER"
+      "mustContain": "no-emit-on-error-false:RANDOMNESS_PLACEHOLDER",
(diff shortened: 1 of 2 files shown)


packages/node/test/fixtures/46-noEmitOnError-false/vercel.json
```diff
@@ -4,7 +4,8 @@
   "probes": [
     {
       "path": "/",
-      "mustContain": "no-emit-on-error-false:RANDOMNESS_PLACEHOLDER"
+      "mustContain": "no-emit-on-error-false:RANDOMNESS_PLACEHOLDER",
+      "logmustContain": "error TS2339: Property 'thisDoesNotExist' does not exist on type 'IncomingMessage'"
```
TooTallNate:
```suggestion
      "logMustContain": "error TS2339: Property 'thisDoesNotExist' does not exist on type 'IncomingMessage'"
```

TooTallNate (changes requested):

TooTallNate (approved):

cb1kenobi (approved):

PR #9410, opened 2023-02-10
[cli][node][node-bridge] improve edge/serverless function error messages

EndangeredMassa:
The error messages shown during `vc dev` when an Edge or Serverless functions returns a promise that ends up being rejected could be better.

Main changes:

- removed "socket hang up" error messages because they were not helpful to the user
- changed serverless function error handling to log explicitly and exit
- changed serverless function error message to include the request path
- changed edge function error message to include request path and a faked (but useful) stack trace

### In Production

**Serverless Function:** In production, for a serverless function rejected promise, you'll see this in the logs:

```
Unhandled Promise Rejection 	{
  "errorType": "Runtime.UnhandledPromiseRejection",
  "errorMessage": "Error: intentional break!",
  "reason": {
    "errorType": "Error",
    "errorMessage": "intentional break!",
    "stack": [
      "Error: intentional break!",
      "    at handler (/var/task/api/node.js:3:9)",
      "    at Server.<anonymous> (/var/task/___vc/__helpers.js:813:19)",
      "    at Server.emit (node:events:527:28)",
      "    at parserOnIncoming (node:_http_server:956:12)",
      "    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)"
    ]
  },
  "promise": {},
  "stack": [
    "Runtime.UnhandledPromiseRejection: Error: intentional break!",
    "    at process.<anonymous> (file:///var/runtime/index.mjs:1194:17)",
    "    at process.emit (node:events:539:35)",
    "    at process.emit (/var/task/___vc/__sourcemap_support.js:559:21)",
    "    at emit (node:internal/process/promises:140:20)",
    "    at processPromiseRejections (node:internal/process/promises:274:27)",
    "    at processTicksAndRejections (node:internal/process/task_queues:97:32)"
  ]
}
Unknown application error occurred
Runtime.Unknown
```

**Edge Function:** In production, for an edge function rejected promise, you'll see this in the logs:

```
Error: intentional break!
    at (api/edge.js:10:10)
```

In both cases, in the browser, you see the "This Serverless/Edge Function has crashed." template with no error message or stack trace.

### In `vc dev`

**Serverless Function:** In `vc dev`, for a serverless function rejected promise, you'll see this in the output:

```
Unhandled rejection: Error: intentional break!
    at handler (/Users/smassa/source/demo/edge-errors/api/node.js:3:9)
    at Server.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:813:19)
[...]

PR #9411, opened 2023-02-10
[cli] show active logs during `vc dev`

EndangeredMassa:
Updating `vc dev` user experience to:

- error messages render red
- request/response log lines
- log is aligned on icon, method/status, then message

EDIT: This was changed to only show request/response logs if the `--info` arg was passed.

Check it out:

CleanShot 2023-02-13 at 18 18 16@2x

---

### Old Design Draft

The `vc dev` experience could be better with some simple request log output. I tried out an implementation that looks pretty good, but has some limitations.

```
$ vc dev
> Ready! Available at http://localhost:3000
GET /api/node-bad :: Rejected Promise returned from /api/node-bad: Error: intentional break!
    at handler (/Users/smassa/source/demo/edge-errors/api/node-bad.js:3:9)
    at Server.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:824:19)
    at Server.emit (node:events:513:28)
    at Server.emit (node:domain:489:12)
    at parserOnIncoming (node:_http_server:998:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
GET /api/node-meh :: 400
GET /api/node-good :: 200
```

In action, you can see how a single line shows the request, then waits, then shows the response:

CleanShot 2023-02-10 at 12 38 34

### Issue 1: Can't show 500 status code before error message is logged

Because of how errors are logged right now, there timing of information available goes (1) request, (2) error if any, (3) response code. It would take some refactoring to capture the error and propagate it back to where the request was initiated in the `DevServer`. There, the log output could be properly ordered.

Ideally, the example above would show the error log like:

```
GET /api/node-bad :: 500
Rejected Promise returned from /api/node-bad: Error: intentional break!
    at handler (/Users/smassa/source/demo/edge-errors/api/node-bad.js:3:9)
    at Server.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:824:19)
    at Server.emit (node:events:513:28)
    at Server.emit (node:domain:489:12)
    at parserOnIncoming (node:_http_server:998:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
```

**Impact:** I think this PR is shippable without making a change here. It would just be nice to show the status code as well without requiring a second line.

### Issue 2: Output gets mangled when handling simultaneous requests

If you make several requests at the same time, the log output gets messed up:

```
GET /api/node-good :: 200
GET /api/node-good :: GET /api/node-good :: GET /api/node-good :: GET /api/node-good :: 200
```

Ideally, the logs for multiple requests would show something like:

```
GET /api/node-good :: 200
GET /api/node-good :: 200
GET /api/node-good :: ...
GET /api/node-good :: ...
```

which would fill in status codes as responses came in. It would take some doing to figure out what to do when an error was logged out.

This happens because the 2-step log strategy for each line assumes nothing else gets logged before the response comes back.

To solve this, we'd have to make the log output more sophisticated. It would have to remember a relative position of lines in the past to render out the proper status code.

Or, we can skip the 2-step log approach for something else.

**Impact:** This isn't good enough to ship as-is. Either we come up with a good solution or go with an alternative.

### Alternative 1: Show two log lines per request

We could instead show a request line and a response line. This is what many request logs look like. It handles concurrency pretty well.

That could look like:

```
GET /api/node-good
200 /api/node-good
```

And some concurrent requests could look like:

```
GET /api/node-good
GET /api/node-bad
Rejected Promise returned from /api/node-bad: Error: intentional break!
    at handler (/Users/smassa/source/demo/edge-errors/api/node-bad.js:3:9)
    at Server.<anonymous> (/Users/smassa/source/vercel/vercel/packages/node-bridge/helpers.js:824:19)
    at Server.emit (node:events:513:28)
    at Server.emit (node:domain:489:12)
    at parserOnIncoming (node:_http_server:998:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
500 /api/node-bad
200 /api/node-good
```

---

Depends on: https://github.com/vercel/vercel/pull/9410


packages/cli/src/util/dev/server.ts
```diff
@@ -1280,9 +1280,20 @@ export default class DevServer {
...
+    this.output.print(`${method} ${chalk.bold(req.url)} :: `);
+
     try {
       const vercelConfig = await this.getVercelConfig();
       await this.serveProjectAsNowV2(req, res, requestId, vercelConfig);
+
+      // wait until the response completes so that the `statusCode` has the real response value
+      res.on('close', () => {
+        // an error message was already logged if the status code is 500
+        if (res.statusCode !== 500) {
```
styfle: What about 501, 502, etc?

packages/cli/src/util/dev/server.ts
```diff
@@ -1280,9 +1280,20 @@ export default class DevServer {
...
 
+    // `print` without a newline so that we can finish the line later with an error or a status code
+    this.output.print(`${method} ${chalk.bold(req.url)} :: `);
+
     try {
       const vercelConfig = await this.getVercelConfig();
       await this.serveProjectAsNowV2(req, res, requestId, vercelConfig);
+
+      // wait until the response completes so that the `statusCode` has the real response value
+      res.on('close', () => {
```
styfle: Should this listener be added before the `try` instead?

EndangeredMassa: I don't want the callback to be executed unless we've called `serveProjectAsNowV2` because it will think the `res.statusCode` is `200`, which would be a very false positive. I had some issues with this during development.

I think moving this would be functionality equivalent right now, but when future changes are made, the current position makes it clearer that there's a dependency of timing here.

packages/cli/src/util/dev/server.ts
EndangeredMassa: Yeah, might as well include those too. [`ddcd4c4` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/ddcd4c4fb741234f6bd6b305bb6c744399ce1df6)

EndangeredMassa:
I think we have to go with "Alternative 1: Show two log lines per request" for now. It's far less work and the risk of unknown edge cases is much lower.

EndangeredMassa:
Switching to Alternative 1. It looks like this for typical requests/responses:

```
→ GET /api/node-good
← 200 /api/node-good
→ GET /api/node-meh
← 400 /api/node-meh
```

and this for errors:

```
→ GET /api/edge
~ ERR /api/edge
intentional break!
    at (api/edge.ts)
← 500 /api/edge
```

I can't easily format the error message with `chalk` because it's not a dependency of `node-bridge` or `node` and I'm not sure if we should add it or not. 

Example of requests/responses with an error in the mix:

CleanShot 2023-02-13 at 15 39 36@2x

Example of simultaneous requests trying to make it clear what error goes with with request:

CleanShot 2023-02-13 at 15 55 00@2x

EndangeredMassa:
Now there's a bit more alignment with common verbs (GET and POST):

CleanShot 2023-02-13 at 18 18 16@2x

packages/cli/src/util/dev/server.ts
```diff
@@ -1278,7 +1278,21 @@ export default class DevServer {
     }
 
     const method = req.method || 'GET';
-    this.output.debug(`${chalk.bold(method)} ${req.url}`);
+    const space = method === 'GET' ? '  ' : ' '; // make GET and POST take up the same width
+    this.output.debug(`${chalk.bold(method)}${space}${req.url}`);
```
cb1kenobi: The `method` and `space` definitions can essentially be combined.

```suggestion
    const method = (req.method || 'GET').padEnd(4);
    this.output.debug(`${chalk.bold(method)} ${req.url}`);
```

I dunno, but I think this is a bit more readable since I don't need to count spaces. Plus the above supports `PUT` just in case.

packages/cli/src/util/dev/server.ts
```diff
@@ -1278,7 +1278,21 @@ export default class DevServer {
...
+
+    // `print` without a newline so that we can finish the line later with an error or a status code
+    this.output.print(
+      `${chalk.bold('→')} ${method}${space}${chalk.bold(req.url)}\n`
+    );
+
+    // wait until the response completes so that the `statusCode` has the real response value
+    res.on('close', () => {
+      const responseLine = `${chalk.bold('←')} ${res.statusCode}  ${chalk.bold(
```
cb1kenobi: I really like this.

packages/node-bridge/helpers.ts
```diff
@@ -309,3 +310,7 @@ export function createServerWithHelpers(
 
   return server;
 }
+
+function colorRed(content: string) {
+  return `\x1b[31m${content}\x1b[0m`;
```
cb1kenobi: Non-blocking: Can `node-bridge` pull in `chalk`? It should be free assuming `ncc` bundles it, right?

cb1kenobi (changes requested):
A few minor nitpicks that may be worth investigating, but shouldn't block this nice UI enhancement.

packages/node-bridge/helpers.ts
EndangeredMassa: I don't know the full impact of adding dependencies there. Maybe someone else can speak to it. I'd be happy to pull in chalk here.

packages/cli/src/util/dev/server.ts
EndangeredMassa: I didn't realize `padEnd` was built-in. Sounds good to me! [`bf390cc` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/bf390cc13b8b2e56d9b1e5b2b02931dfba16f81c)

packages/node-bridge/helpers.ts
cb1kenobi: I guess another benefit of using `chalk` is the whole `--no-colors` thing, ya?

styfle: No we should not add dependencies because any increase in size could put users over the lambda limit of 50MB

styfle (commented):
Does this log `index.html` and `favicon.ico` every page visit?

I'm not seeing any tests for the non-error cases.

It seems like it will get really noisy though 😬

EndangeredMassa:
Updates:

- only log functions, not static files: [`ac3acf1` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/ac3acf1a5ea6c2a5451de89c82aee03deffde01d)
- support `NO_COLOR` where we don't use chalk: [`bbbdd4a` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/bbbdd4a0ffc4e590fe20a889f346eb010c030bee)
- add test for logging functions, but not static files: [`2ba4f6c` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/2ba4f6c6ed3b48c553c5013cac560e7478e7f575)

No Color example:

CleanShot 2023-02-14 at 10 48 20@2x

cb1kenobi (approved):
I think this looks good!

EndangeredMassa:
I'm updating this to not show successful responses. I'm also adding more tests.

cb1kenobi (dismissed):
I have tested this on macOS and Windows and it everything renders beautifully.

EndangeredMassa:
Expanded the tests: [`c20097f` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/c20097f5a4e96f1f0cab4a1a32a2b14ad29fdf03)

packages/node-bridge/helpers.ts
```diff
@@ -298,7 +298,8 @@ export function createServerWithHelpers(
       try {
         await handler(req, res);
       } catch (err) {
-        console.log(`Error from API Route ${req.url}: ${err.stack}`);
+        const coloredStack = colorRed(err.stack);
+        console.error(`! ERR  ${req.url}\n${coloredStack}`);
```
EndangeredMassa: I changed these to `stderr` to match the rest of the log output

cb1kenobi (dismissed):

packages/cli/src/util/dev/server.ts
```diff
@@ -2082,6 +2085,27 @@ export default class DevServer {
     }
   };
 
+  private logRequestResponse(
+    req: http.IncomingMessage,
+    res: http.ServerResponse
+  ) {
+    const method = (req.method || 'GET').padEnd(4);
```
styfle: Nit pick: there are [methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) like `OPTIONS` that are 7 chars long

```suggestion
    const method = (req.method || 'GET').padEnd(7);
```

packages/cli/test/dev/integration-1.test.ts
```diff
@@ -90,6 +92,65 @@ test(
   })
 );
 
+test('[vercel dev] only logs request/response for functions', async () => {
```
styfle: Nit pick: add this to `integration-5.test.ts` instead since that is currently the fastest

packages/cli/src/util/dev/server.ts
EndangeredMassa: I picked 4 because get/post seem most likely. If a longer one comes along it will no longer be aligned, but I think that's fine over every GET having 4 extra spaces.

I could go either way, though.

packages/cli/test/dev/integration-1.test.ts
EndangeredMassa: Good call. [`5703795` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/570379572c58762be8e363f0149a1f73b3421c8d)

packages/cli/src/util/dev/server.ts
styfle: Seems like it should either be aligned or not. So either drop pad or make it work for all methods. There are so few methods its going to look like a bug if its aligned on some but not others. You'll see OPTIONS for any CORS request which I think is common.

EndangeredMassa: Not aligned at all:

CleanShot 2023-02-16 at 16 11 13@2x

Aligned to 7 spaces:
CleanShot 2023-02-16 at 16 16 28@2x

---

I like 7. What do you think? [`795c863` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/795c86390522856732a625e663fd5bcd34470568)

styfle: Agreed lets go with 7.

Also, are we still planning to log for success? I though we decided to only log on error?

EndangeredMassa: We don't know a request will be successful until it responds. If we wait for the response to show any log message, it will feel slow. I think we want to see the request as soon as it was triggered.

We were logging all requests and all responses before, but I changed it earlier to:

- request
    - lambda or edge: log
    - else: hide
- response
    - lambda or edge:
        - <400: hide
        - 404: hide
        - else: log
    - else: hide
    
Relevant commits:

- [`ac3acf1` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/ac3acf1a5ea6c2a5451de89c82aee03deffde01d)
- [`289e6cb` (#9411)](https://github.com/vercel/vercel/pull/9411/commits/289e6cb3c84a8d5de83b0e3eade8a3e692d77d27)

styfle: Is it going to be too noise and hide the user's `console.log()`? Is there a way to turn it off if the user wants to only see their own logs?

EndangeredMassa: Potentially.

Option 1 - Dim: We could dim all the log lines except the exceptional ones. That would make the default text of a user's `console.log()` stand out.

Option 2 - Flag: We could add an option to disable this behavior. It may make sense to go fully into the realm of setting log levels.

Option 3 - Error Only: We could not add these request/response logs in this way at all.

cb1kenobi: I'm +1 for Option 2 - implementing a log level system. It would default to `warn` which would also display errors. A new global `--log-level <level>` option makes sense, but I would really like to be able to control the log level using an environment variable. Users could set the env var globally or in the `.env` file.

styfle: What would be the possible values of `--log-level`? And what would be the default value?

cb1kenobi: Traditionally the value would be a string containing one of:
* `'error'` - Errors only
* `'warn'` - Errors and warnings
* `'info'` - General useful info (URLs, ports, something has started)
* `'debug'` - Dumping variables, general debug logging

I've also see `'trace'` that would be ultra detailed output such as `stdout` from spawned processes, etc, but I don't think we need this.

cb1kenobi: Chrome's Dev Tools console has `'error'`, `'warn'`, `'info'`, and `'verbose'`. I kinda like `'verbose'` which essentially means no filtering, show everything.

styfle: Since we already have —debug, perhaps that existing feature can be merged with this new feature

EndangeredMassa: Yeah, I would fold in `--debug` as a shorthand for `--loglevel=debug`.

Should we block this PR on implementing better log level management? I'd be fine with that.

EndangeredMassa: We discussed this on a call:

- put these logs behind `--info` so that they are not shown by default
- this is compatible with a future change where logs levels are managed more explicitly
- restore the function path in the error messages in case other log lines (perhaps from frontend server) end up between request/response log lines

styfle: If I use `--debug`, do I also see `--info` logs plus more?

EndangeredMassa: Right now, that would mean you'd see two different representations of a request log line, but that's probably OK until we revise log levels later. When we add log levels, we could audit what is shown in the `debug` level and possibly remove the `debug` request log line.

For this PR, I think it'd be fine to show both in `--debug`.

EndangeredMassa: We can leave them separate for minimal disruption, then when we implement full log level support we do make them cascade. I'm good with either, though.

PR #9412, opened 2023-02-10
[build-utils] Filter out files outside of `cwd` when glob is in "follow" mode

TooTallNate:
Enables `glob()` to operate in "follow" mode, but filters out any values where a symlink points to a target that lives outside of the `cwd.

### DIFF
--- a/packages/build-utils/src/fs/glob.ts
+++ b/packages/build-utils/src/fs/glob.ts
@@ -63,7 +63,7 @@ export default async function glob(
       (isSymlink || (await lstat(fsPath)).isSymbolicLink())
     ) {
       const target = await readlink(absPath);
-      const absTarget = path.join(path.dirname(absPath), target);
+      const absTarget = path.resolve(path.dirname(absPath), target);
       if (path.posix.relative(options.cwd, absTarget).startsWith('../')) {
         continue;
       }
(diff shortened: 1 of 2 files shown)


styfle (dismissed):

packages/build-utils/src/fs/glob.ts
```diff
@@ -45,16 +45,33 @@ export default async function glob(
...
+
+    // When `follow` mode is enabled, ensure that the entry is not a symlink
+    // that points to outside of `cwd`
+    if (
+      options.follow &&
+      (isSymlink || (await lstat(fsPath)).isSymbolicLink())
+    ) {
+      const target = await readlink(absPath);
+      const absTarget = path.join(path.dirname(absPath), target);
+      if (path.posix.relative(options.cwd, absTarget).startsWith('../')) {
```
cb1kenobi: I think there's a bug on Windows. No matter the link target, relative or absolute, `path.posix.relative()` will ALWAYS return a path that begins with `../` on Windows.

If you use `path.relative()`, then it will resolve the correct relative path, however on Windows, relative paths always starts with `..\\`.

```suggestion
      if (path.relative(options.cwd, absTarget).startsWith(`..${path.sep}`)) {
```

cb1kenobi (changes requested):

cb1kenobi (dismissed):

cb1kenobi (approved):

cb1kenobi (approved):

EndangeredMassa (approved):

PR #9415, opened 2023-02-10
[tests] Fix examples tests with shared util

styfle:
Fix examples tests with shared util and also made sure to exclude `node_modules`

### DIFF
--- a/examples/__tests__/test-utils.ts
+++ b/examples/__tests__/test-utils.ts
@@ -1,7 +1,28 @@
 import { basename, join } from 'path';
-import { testDeployment } from '../../test/lib/deployment/test-deployment.js';
+import { lstatSync, readdirSync } from 'fs';
 
 export async function deployExample(filename: string) {
+  const testDeployment = require('../../test/lib/deployment/test-deployment.js');
   const example = basename(filename).replace(/\.test\.ts$/, '');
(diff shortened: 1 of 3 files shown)


EndangeredMassa (dismissed):

TooTallNate (dismissed):

TooTallNate (approved):

examples/__tests__/test-utils.ts
```diff
@@ -1,7 +1,27 @@
 import { basename, join } from 'path';
-import { testDeployment } from '../../test/lib/deployment/test-deployment.js';
+import { lstatSync, readdirSync } from 'fs';
 
 export async function deployExample(filename: string) {
+  const { testDeployment } = require('../../test/lib/deployment/test-deployment.js');
```
cb1kenobi, line 5: Non-blocking: it might be a good idea to drop a comment as to why this is being required inside `deployExample()` instead of a top-level `import`.

EndangeredMassa (approved):

examples/__tests__/test-utils.ts
styfle: Seems like jest had a problem with it, perhaps because of the package boundary. Or maybe it had a typo, not sure. I can check again when I’m not AFK

EndangeredMassa:
merging because this is blocking other PRs

PR #9417, opened 2023-02-10
[remix-entry-server] Add initial implementation

TooTallNate:
This adds a new package to the monorepo: `@vercel/remix-entry-server`

The purpose of this package is to provide most of the implementation of the Remix `app/entry.server.tsx` file with proper support for the Vercel Serverless and Edge runtimes, specifically in regards to React 18 streaming.

The reason that this package is necessary, as opposed to just updating our Remix template, is due to the fact that Serverless and Edge runtimes require a different implementation. This is because the `react-dom/server` package exports different functions for the Node (`renderToPipeableStream()`) vs. Edge (`renderToReadableStream()`) bundles, and thus this package also has two different implementations (by utilizing the `main` and `browser` fields in `package.json`).

Usage of this package in a Remix application's `app/entry.server.tsx` file looks like:

```tsx
import handleRequest from "@vercel/remix-entry-server";
import { RemixServer } from "@remix-run/react";
import type { EntryContext } from "@remix-run/server-runtime";

export default function (
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext
) {
  const remixServer = <RemixServer context={remixContext} url={request.url} />;
  return handleRequest(request, responseStatusCode, responseHeaders, remixServer)
}
```

Once this package is published then we can update our own Remix template to utilize it, enabling React 18 streaming for both Vercel runtimes.


Ethan-Arrowood (dismissed):

packages/remix-entry-server/README.md
```diff
@@ -0,0 +1,28 @@
+# `@vercel/remix-entry-server`
+
+This package is meant for use within Remix applications when deploying to Vercel. It provides implementations for the `app/entry.server.tsx` file for both the Node.js Serverless Runtime and the Edge Runtime. The implementations are configured to [handle streaming responses](https://remix.run/docs/en/v1/guides/streaming).
+
+## Usage
+
```
EndangeredMassa, line 6: **suggestion (non-blocking):** be more explicit

Can we write something here to be more explicit about how to use this? Maybe something like:

```suggestion

Make sure `@vercel/remix-entry-server` is installed with your favorite package manager, then replace your `app/entry.server.tsx` file with the following:

```

EndangeredMassa (dismissed):

packages/remix-entry-server/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "@remix-run/node": "1.12.0",
+    "isbot": "3.6.5",
+    "react-dom": "18.2.0"
+  },
+  "devDependencies": {
+    "@types/react-dom": "18.0.10",
```
cb1kenobi, line 26: For safety, it's probably a good idea to add `execa` and `fs-extra` to the `devDependencies`.

cb1kenobi (changes requested):
`serveTheBots()` and `serveBrowsers()` are so similar. It seems `onAllReady()` and `onShellReady()` are identical aside from `didError` which seems harmless for the bot path. I dunno how much more time you want to put into this, but there seems like room for some cute code reuse.

TooTallNate:
@cb1kenobi The code is actually direct copy+paste from the Remix Streaming docs page, I wanted to keep it "official".

cb1kenobi (approved):

EndangeredMassa (approved):

Ethan-Arrowood (approved):

PR #9428, opened 2023-02-13
[cli] add slash before url for edge function errors

EndangeredMassa:
Edge Functions and Serverless Functions were both updated to show the path of their entry points, but it turns out Serverless Function paths started with a slash and Edge Function paths did not.

The [related code for Serverless Functions](https://github.com/vercel/vercel/blob/d62888094236c6f2afdc34a5db2f098adf575470/packages/node-bridge/helpers.ts) does not need to be updated.

---

Before:

CleanShot 2023-02-13 at 13 23 42@2x

---

After: 

CleanShot 2023-02-13 at 13 20 22@2x


cb1kenobi (dismissed):

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -169,7 +169,7 @@ export async function createEdgeEventHandler(
       const fakeStackTrace = `    at (${entrypointRelativePath})`;
       const urlPath = extractUrlPath(entrypointRelativePath);
```
styfle: Should `extractUrlPath()` return with the forward slash instead of modifying the log?

EndangeredMassa: Yes, that better represents what `urlPath` is: [`6b362d9` (#9428)](https://github.com/vercel/vercel/pull/9428/commits/6b362d9ee9f62ed37dbf7a04c216002d52bf6365)

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -192,6 +192,11 @@ function extractUrlPath(entrypointRelativePath: string) {
   if (parts.length === 1) {
     return entrypointRelativePath;
```
styfle: This early return isnt account for. How about this:

```js
const parts = entrypointRelativePath.split('.');
if (parts.length > 1) {
  parts.pop();
}
return `/${parts.join('.')}`;
```

cb1kenobi (dismissed):

packages/node/src/edge-functions/edge-handler.ts
EndangeredMassa: definitely going too fast here, yeah, I'll update this

EndangeredMassa: @styfle @cb1kenobi or should I use `entrypointToOutputPath`? It doesn't return the `/` at the start, but this function does already essentially exist.

https://github.com/vercel/vercel/blob/5b88f673f89c44b9e91e44247e29be6bed43805a/packages/node/src/utils.ts#L49

cb1kenobi: I do prefer the implementation of `entrypointToOutputPath()` better, I'm concerned about the potential breakage from touching something shared. In theory the tests would catch it, so I'd say give it a shot. Add `/` to the result of `entrypointToOutputPath()` and replace instances of `extractUrlPath()` with `entrypointToOutputPath()`.

styfle: How about this?

```ts
function entrypointToRequestPath(entrypoint: string, zeroConfig?: boolean) {
  return `/${entrypointToOutputPath(entrypoint, zeroConfig)}`;
}
```

EndangeredMassa: How about this? [`0e8d8b1` (#9428)](https://github.com/vercel/vercel/pull/9428/commits/0e8d8b1694c1762cb9dbab9bbb9dfa30109fb2a9)

styfle (approved):

cb1kenobi (approved):

PR #9432, opened 2023-02-13
[tests] Fix monorepo tests

chloetedder:


cb1kenobi (dismissed):

package.json
```diff
@@ -45,7 +45,7 @@
     "build": "node utils/gen.js && turbo run build",
     "vercel-build": "pnpm build && pnpm run pack && cd api && node -r ts-eager/register ./_lib/script/build.ts",
     "pre-commit": "lint-staged",
-    "test": "jest --rootDir=\"test\" --testPathPattern=\"\\.test.js\"",
+    "test": "jest --rootDir=\"test\" --testPathPattern=\"\\.test.ts\"",
```
styfle: I don't think this is correct

styfle (changes requested):
Did you intend to change the root `package.json`?

packages/cli/test/fixtures/unit/commands/build/monorepo-detection/nx-package-config/packages/app-2/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "app-2",
   "version": "0.0.1",
-  "main": "dist/index.js",
+  "main": "script.js",
```
styfle: What's the purpose of this change?

I would have expected `main` be the dist output instead of the source file.

cb1kenobi: ~There is no `dist/index.js` in this monorepo test fixture.~

package.json
cb1kenobi: Agree w/ @styfle. There are no `.ts` files in the root `test` directory.

packages/cli/test/fixtures/unit/commands/build/monorepo-detection/nx-package-config/packages/app-2/package.json
styfle: @cb1kenobi If you look at the build script, it emits a `dist/index.js` which is typically how npm packages work: `main` points to the compiled output file

https://github.com/vercel/vercel/blob/main/packages/cli/test/fixtures/unit/commands/build/monorepo-detection/nx-package-config/packages/app-2/script.js

packages/cli/test/fixtures/unit/commands/build/monorepo-detection/turbo-package-config/packages/app-2/package.json
```diff
@@ -1,10 +1,6 @@
 {
   "name": "app-2",
   "version": "0.0.1",
-  "main": "dist/index.js",
-  "files": [
-    "dist"
-  ],
   "scripts": {
```
styfle, line 4: Can you explain why this would affect the test? Do we read “main” or “files” at all?

PR #9436, opened 2023-02-14
[ruby] fix: HEAD requests

marcgreenstock:
When WEBrick receives `HEAD` requests it discards the body (i.e. `req.body.nil? => true`), this causes Vercel to throw a `BODY_NOT_A_STRING_FROM_FUNCTION` since it is expecting the serverless function to respond with a string in the body.


TooTallNate:
Thanks! Could you add a test fixture for this as well?

marcgreenstock:
@TooTallNate of course

styfle (approved):
Great work, thanks! 🥇

PR #9442, opened 2023-02-14
[remix] Support "regions", "memory" and "maxDuration" in static config

TooTallNate:
Apply the `regions` configuration (for both Edge and Node) and `memory`/`maxDuration` (only for Node) in a page's static config export, i.e.

```js
export const config = { runtime: 'edge', regions: ['iad1'] }

// or for Node
export const config = { runtime: 'nodejs', regions: ['iad1'], maxDuration: 5, memory: 3008 }
```

Similar to `runtime`, these config values can be inherited from a parent layout route to apply to all sub-routes. Routes with common config settings get placed into a common server bundle, meaning that there may now be more than 2 functions created (previously was one Edge, one Node), allowing for more granularity between the server build bundles.


packages/remix/test/fixtures/01-remix-basics/app/b.server.ts
```diff
@@ -8,10 +8,13 @@ import { json } from '@remix-run/node';
 
 export async function loader() {
   const hi = await new Promise<string>((resolve, reject) => {
-    exec('echo hi from the B page', (err, stdout) => {
-      if (err) return reject(err);
-      resolve(stdout);
-    });
+    exec(
+      `echo hi from the B page running in ${process.env.VERCEL_REGION}`,
```
cb1kenobi, line 12: **question (non-blocking):** Why spawn `echo`? If Node already sees `process.env.VERCEL_REGION`, then I'd think you can just resolve it, or use `setImmediate()` if you need to force async.

cb1kenobi (approved):
CR'd and LGTM. I did not functional test.

packages/remix/src/utils.ts
```diff
@@ -18,6 +34,60 @@ export function findConfig(dir: string, basename: string): string | undefined {
...
+  let regions: ResolvedRouteConfig['regions'];
+  let maxDuration: ResolvedNodeRouteConfig['maxDuration'];
+  let memory: ResolvedNodeRouteConfig['memory'];
+
+  for (const currentRoute of getRouteIterator(route, routes)) {
+    const staticConfig = configs.get(currentRoute);
+    if (staticConfig) {
+      if (typeof runtime === 'undefined' && staticConfig.runtime) {
+        runtime = isEdgeRuntime(staticConfig.runtime) ? 'edge' : 'nodejs';
+      }
```
EndangeredMassa, line 56: **question (non-blocking):** where should we validate that `maxDuration` and `memory` are only used when `runtime === 'edge'`?

We could do it here, but it probably makes more sense in `static-config`'s `getConfig` function.

EndangeredMassa (approved):

packages/remix/test/fixtures/01-remix-basics/app/b.server.ts
TooTallNate: The point of this test was to verify that `spawn` could be used in a Node route (and wouldn't subsequently cause the Edge bundle to fail).

In this PR I'm just piggy backing off that existing test to also check that `VERCEL_REGION` is set to the expected value.

packages/remix/src/utils.ts
TooTallNate: Currently those values would just be ignored. We could improve upon that in the future if needed though.

packages/remix/test/fixtures/01-remix-basics/app/b.server.ts
cb1kenobi: Ah, that makes sense. Is that worth mentioning in a comment?

packages/remix/src/utils.ts
EndangeredMassa: Right. Someone could configure them and wonder why they aren't doing anything. It's a UX bug, in my opinion.

PR #9445, opened 2023-02-14
[test] Use execFileSync() to get processes

cb1kenobi:
`spawnSync()` does not throw if the command can't be found in the PATH or if an error occurs. If we use `execFileSync()`, it will throw and that was likely the desired behavior in this test utility function.

### DIFF
--- a/packages/cli/test/dev/utils.js
+++ b/packages/cli/test/dev/utils.js
@@ -522,7 +522,7 @@ async function ps(parentPid, pids = {}) {
       : ['ps', '-o', 'pid', '--no-headers', '--ppid', parentPid];
 
   try {
-    const { stdout: buf } = execFileSync(cmd[0], cmd.slice(1), {
+    const buf = execFileSync(cmd[0], cmd.slice(1), {
       encoding: 'utf-8',
     });
     for (let pid of buf.match(/\d+/g)) {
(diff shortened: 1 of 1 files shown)


packages/cli/test/dev/utils.js
```diff
@@ -522,7 +522,7 @@ async function ps(parentPid, pids = {}) {
       : ['ps', '-o', 'pid', '--no-headers', '--ppid', parentPid];
 
   try {
-    const { stdout: buf } = spawnSync(cmd[0], cmd.slice(1), {
+    const { stdout: buf } = execFileSync(cmd[0], cmd.slice(1), {
```
EndangeredMassa: **requirement (blocking):** support different return value

`execFileSync` has a different return value than `spawnSync`. The response is stdout, not an object containing a `stdout` key.

```
> spawnSync('echo', ['hey', 'there']).stdout.toString()
'hey there\n'

> execFileSync('echo', ['hey', 'there']).toString()
'hey there\n'
```


I think this resolves it:

```suggestion
    const buf = execFileSync(cmd[0], cmd.slice(1), {
```

EndangeredMassa (changes requested):

packages/cli/test/dev/utils.js
cb1kenobi: Good catch!

EndangeredMassa (approved):
Assuming CI passes, this looks good to me!

styfle (approved):

PR #9446, opened 2023-02-14
Add template key to frameworks package

S3Prototype:
We pull data from the `frameworks` package in the docs to create a list of the frameworks we support. We'd like to be able to also embed a link to a vercel template associated with each framework (if one exists). Doing so manually made less sense than just adding that info to the `frameworks` package to me.

This PR:

- Adds an optional `templateUrl` key to the `Framework` type
- Fills in the templateUrl property on each Framework item where one exists


styfle (changes requested):
This will create circular dependencies. 

Since vercel/front already depends on @vercel/frameworks and this PR is effectively making @vercel/frameworks depend on vercel/front

PR #9447, opened 2023-02-14
[build-utils] Add framework to build output

lubakravche:
Adds framework to Lambda and edge build outputs so that we can distinguish which framework they originated from when certain features should be applied to specific frameworks. 

ticket: ED-131

x-ref: [slack channel](https://vercel.slack.com/archives/C042LHPJ1NX)


packages/next/src/index.ts
```diff
@@ -1094,6 +1094,7 @@ export const build: BuildV2 = async ({
           shouldAddHelpers: false,
           shouldAddSourcemapSupport: false,
           supportsMultiPayloads: !!process.env.NEXT_PRIVATE_MULTI_PAYLOAD,
+          framework: 'nextjs',
```
ijjk: Might be helpful to have the version for additional context e.g. `nextjs@${detectedVersion}`

cb1kenobi:
Is there a related issue/ticket for this? What's the context? Where's the description?

lubakravche:
@cb1kenobi updated description with context, thanks @ijjk!

packages/remix/src/build.ts
```diff
@@ -334,6 +334,7 @@ async function createRenderNodeFunction(
     shouldAddSourcemapSupport: false,
     operationType: 'SSR',
     experimentalResponseStreaming: true,
+    framework: 'remix',
```
styfle, line 337: Should this also be added to `@vercel/static-build` which is where most frameworks are used?

I'll defer to @EndangeredMassa who worked on a similar feature recently.

lubakravche:
breaking up into 2 parts https://github.com/vercel/vercel/pull/9448 https://github.com/vercel/vercel/pull/9449

PR #9448, opened 2023-02-14
[build-utils] add framework slug and version to build output

lubakravche:
Adds framework to Lambda and edge build outputs so that we can distinguish which framework they originated from when certain features should be applied to specific frameworks.

Breaking up https://github.com/vercel/vercel/pull/9447 into two parts:
This PR introduces the frameworks type. Part 2: https://github.com/vercel/vercel/pull/9449

ticket: ED-131

x-ref: [slack channel](https://vercel.slack.com/archives/C042LHPJ1NX)

### DIFF
--- a/packages/build-utils/src/edge-function.ts
+++ b/packages/build-utils/src/edge-function.ts
@@ -1,4 +1,4 @@
-import type { Cron, Files } from './types';
+import type { Cron, Files, Framework } from './types';
 
 /**
  * An Edge Functions output
@@ -45,10 +45,7 @@ export class EdgeFunction {
   cron?: Cron;
 
   /** The framework */
-  framework?: {
-    slug: string;
-    version?: string;
-  };
+  framework?: Framework;
 
   constructor(params: Omit<EdgeFunction, 'type'>) {
     this.type = 'EdgeFunction';
--- a/packages/build-utils/src/edge-function.ts
+++ b/packages/build-utils/src/edge-function.ts
@@ -44,6 +44,9 @@ export class EdgeFunction {
   /** Cronjob definition for the edge function */
   cron?: Cron;
 
+  /** The framework */
+  framework?: string;
+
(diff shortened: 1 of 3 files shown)


packages/build-utils/src/edge-function.ts
```diff
@@ -44,6 +44,9 @@ export class EdgeFunction {
   /** Cronjob definition for the edge function */
   cron?: Cron;
 
+  /** The framework */
+  framework?: string;
```
ijjk: Should this be `{ slug: string, version?: string }` to match the others?

packages/build-utils/src/edge-function.ts
```diff
@@ -44,6 +44,12 @@ export class EdgeFunction {
   /** Cronjob definition for the edge function */
   cron?: Cron;
 
+  /** The framework */
```
TooTallNate, line 47:
```suggestion
  /** The framework which created this function */
```

ijjk (dismissed):
Thanks!

packages/build-utils/src/edge-function.ts
TooTallNate: Actually, can we stick this in `types.ts`? That way we can share the same interface (and doc comment) with both `EdgeFunction` and `Lambda`.

TooTallNate (changes requested):

packages/build-utils/src/edge-function.ts
lubakravche: `Framework` or `FunctionFramework`?

TooTallNate: Let's go with `FunctionFramework` since there's already a `Framework` type in `@vercel/frameworks`.

TooTallNate (approved):

TooTallNate:
Thanks Luba!

PR #9454, opened 2023-02-15
[build-utils][cli][client][node][next][static-config]: forward `crons` from vercel.json to config.json

vvo:
This PR changes the way cron jobs are being created in the build output API. This is my first time contributing here. If you see something unusual, let me know.

✅ Good for review

Our goal is to:
- Allow creating cron jobs via the `crons` property of `vercel.json` for end users
- Allow framework authors to create cron jobs on Vercel via the `crons` property of the Build Output API configuration

---

As you can see, we removed the previous implementation where cron jobs could be configured at the function code level (export const cron = ""), on top of vercel.json `functions` property. Here's why:

- All frameworks would have to implement the configure at the function code level
- Not all frameworks can easily map a path to a specific function (example: SvelteKit) and would have to bail on bundling functions inside the same lambda
- Configuring a path + scheduler provides a better mapping of what cron jobs are as of today: API routes on a schedule and not functions on a schedule
- Dynamic routes Cron Jobs will be supported: /api/crons/sync-slack-team/230
- Query parameters will be supported support: /api/crons/sync-slack-team/230?secret=32k13l2k13lk21 (= securing cron jobs v0)
- 100% frameworks compatibility from day one

Next.js and other frameworks may choose to implement their own cron jobs feature that will then need to be configured through the `crons` property of `config.json` (build output API).

cc @timneutkens @Rich-Harris 

Internal thread: https://vercel.slack.com/archives/C04DWF5HB6K/p1676366892714349


packages/cli/src/util/validate-config.ts
```diff
@@ -93,6 +93,28 @@ const imagesSchema = {
...
+        maxLength: 512,
+      },
+      schedule: {
+        type: 'string',
+        minLength: 1,
+        maxLength: 256,
+      },
+    },
+  },
+};
```
styfle, line 117: Is there a test for this?

vvo: Will add one 👍

packages/cli/test/fixtures/unit/commands/build/with-cron/vercel.json
```diff
@@ -1,10 +1,8 @@
 {
-  "functions": {
-    "api/overwrite/serverless.js": {
-      "cron": "0 10-20 * * *"
-    },
-    "api/overwrite/edge.js": {
-      "cron": "10 * * * *"
+  "crons": [
+    {
+      "path": "/api/cron-job",
```
styfle, line 4: Does that path work with static files too, not just functions?

vvo: Yes, it can be anything indeed.
This path is a free-form path. You can input whatever you want. We won't be checking if the path exists, for now at least.

packages/cli/test/integration.js
```diff
@@ -2635,7 +2635,7 @@ test('next unsupported functions config shows warning link', async t => {
   t.is(output.exitCode, 0, formatOutput(output));
   t.regex(
     output.stderr,
-    /Ignoring function property `runtime`\. When using Next\.js, only `memory`, `maxDuration`, and `cron` can be used\./gm,
+    /Ignoring function property `runtime`\. When using Next\.js, only `memory`, and `maxDuration` can be used\./gm,
```
styfle: I think this needs to change in the API first before you update this test

packages/cli/src/util/validate-config.ts
styfle: Lets add multiple:

- path without schedule
- path and length but exceeding 512
- path and length but empty strings

packages/build-utils/src/types.ts
```diff
@@ -411,7 +410,11 @@ export interface BuildResultBuildOutput {
   buildOutputPath: string;
 }
 
-export type Cron = string;
+export interface Cron {
+  path: string;
+  schedule: string;
```
styfle, line 415: Do we also need a `region` property here?

A function might be deployed to multiple regions so its not clear if all regions are invoked or just one?

styfle:
I discussed with @vvo and we decided this is not considered semver major because the feature was not enabled so changing the public API is not going to change behavior (it didn't do anything yet).

packages/build-utils/src/types.ts
AndyBitz: The requests for Cron Jobs are made from one region at the moment over public internet, so it doesn't matter in this case.

packages/cli/src/util/validate-config.ts
vvo: Added tests here: 12f076fb97087bbd01fce130365c29a59d838e24

packages/cli/test/integration.js
vvo: is there a test failing because of this already?

styfle: Yes this test is failing as you can see. You'll need to change it back for tests to pass

https://github.com/vercel/vercel/actions/runs/4187893545/jobs/7258336358#step:8:1134

vvo: (I am fine reverting, I am just wondering which test is currently not passing because of this)

vvo: Done

vvo:
There are failing tests, but I am not sure they are related.

packages/cli/test/unit/util/dev/validate.test.ts
```diff
@@ -285,4 +285,78 @@ describe('validateConfig', () => {
...
+
+  it('should error when schedule is too short', () => {
+    const error = validateConfig({
+      crons: [{ path: '/', schedule: '* * * * ' }],
+    });
+    expect(error!.message).toEqual(
+      'Invalid vercel.json - `crons[0].schedule` should NOT be shorter than 9 characters.'
+    );
+    expect(error!.link).toEqual(
+      'https://vercel.com/docs/concepts/projects/project-configuration#crons'
```
styfle, line 359: Let's remember to update the documentation after this lands

styfle (dismissed):

packages/cli/test/unit/util/dev/validate.test.ts
vvo: Planned, it had been done already for the previous API in a PR, and we will update that PR

AndyBitz (dismissed):

7koma95 (approved):

PR #9455, opened 2023-02-15
[examples] update remix template

Ethan-Arrowood:
Updates our remix template so that it works with our new Remix changes


examples/remix/app/root.tsx
```diff
@@ -16,6 +16,7 @@ export const meta: MetaFunction = () => ({
 });
 
 export default function App() {
+  inject();
```
Ethan-Arrowood: Is this correct? Using `Analytics` component wasn't working

TooTallNate (dismissed):

leerob (approved):

PR #9458, opened 2023-02-16
fix(remix): splat routes

mcansh:
pathToRegexp blows up on splat routes `app/routes/$.tsx` where path is `*`, it's fine for nested splat routes `app/routes/nested/$.tsx` as the path is `nested/*`

note the new splat regex was made by GitHub copilot so feel to tweak if desired


mcansh:
closing in favor of #9457

PR #9464, opened 2023-02-16
[cli]: fix merging of vercel.json and build result crons

vvo:
Ensures that existing crons and crons from vercel.json are merged together correctly.


AndyBitz (dismissed):

packages/cli/src/commands/build.ts
```diff
@@ -756,7 +756,7 @@ function mergeCrons(
 ): BuildOutputConfig['crons'] {
   for (const result of buildResults) {
     if ('crons' in result && result.crons) {
-      crons = Object.assign({}, crons, result.crons);
+      crons = crons?.concat(result.crons);
```
AndyBitz: Hmm, I guess you actually want it to be result.crons.concat, no?

AndyBitz (approved):

7koma95 (approved):

PR #9440, opened 2023-02-14
[build-utils][cli] Remove 64 increment limit for serverless functions

ofhouse:
We no longer require the memory to be provided in steps of 64mb for serverless functions.
Instead the memory can now be chosen freely from `128mb` to `3008mb` in `1mb increments`.

Updates the `vercel.json` schema to reflect that change.

PR #9465, opened 2023-02-16
[fs-detectors] Remove increments of 64 limit for function memory

ofhouse:
Missed this occurrence so it still prevents the upload of serverless functions that have a mem value that is not dividable by 64.
Should be the last place before we can ship the documentation update.

#### Related PRs
- #9440


styfle (approved):
Looks good. Although we should probably change this from manual validation to using the schema.

cb1kenobi (approved):

PR #9469, opened 2023-02-16
[remix] Add unit tests

TooTallNate:
Moves parts of the `@vercel/remix` builder into util functions that have isolated unit tests. No functionality changes.

### DIFF
--- a/packages/remix/package.json
+++ b/packages/remix/package.json
@@ -11,8 +11,9 @@
   },
   "scripts": {
     "build": "node build.js",
-    "test-e2e": "pnpm test test/integration.test.ts",
-    "test": "jest --env node --verbose --bail --runInBand"
+    "test": "jest --env node --verbose --bail --runInBand",
+    "test-unit": "pnpm test test/unit.*test.*",
+    "test-e2e": "pnpm test test/integration.test.ts"
   },
   "files": [
     "dist",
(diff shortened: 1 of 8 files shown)


packages/remix/src/utils.ts
```diff
@@ -12,3 +17,39 @@ export function findConfig(dir: string, basename: string): string | undefined {
 
   return undefined;
 }
+
+export function isLayoutRoute(
+  routeId: string,
+  routes: Pick<ConfigRoute, 'id' | 'parentId'>[]
```
cb1kenobi, line 23: Out of curiosity, is the `Pick` necessary? If `id` is not used, they what's the harm in just declaring `routes: ConfigRoute`?

cb1kenobi (approved):
Looks good!

packages/remix/src/utils.ts
TooTallNate: It's just cause TypeScript complains in the tests without all the necessary props, which aren't directly necessary for this function.

packages/remix/package.json
```diff
@@ -11,8 +11,9 @@
   },
   "scripts": {
     "build": "node build.js",
-    "test-e2e": "pnpm test test/integration.test.ts",
-    "test": "jest --env node --verbose --bail --runInBand"
+    "test": "jest --env node --verbose --bail --runInBand",
+    "test-unit": "pnpm test test/unit.*test.*",
```
styfle, line 15: In the future, lets move all unit tests into a new directory: `test/unit/**`

It makes it easier to avoid typos when adding new tests 🙂

PR #9471, opened 2023-02-17
feat(cli): get config from `vercel` property in package.json

vedantmgoyal9:
## Description

This change allows Vercel CLI to get config from `vercel` property in package.json. 

I don't know how the Vercel CLI's codebase is structured. I performed a GitHub search of `vercel.json` and I was able to make out that this is the file that resolves the path and parses configuration from `vercel.json` or `now.json`. Here, I have added some lines of code which parses `package.json`, and if `packageJson.vercel !== null`, the config object is returned, otherwise not.

I would kindly request the maintainers to test this since I don't know how to compile the package and test it locally.

## Motivation

Mainly, it came from a YouTube video: [WTF are all these config files for?](https://www.youtube.com/watch?v=14WanxTD2O4) uploaded on the Beyond Fireship channel.

This idea is also inspired by the famous code-formatter prettier which allows putting the config inside a `prettier` field in package.json (https://prettier.io/docs/en/configuration.html).

The same is also done by ESLint which reads config from package.json's `eslintConfig` property.

### DIFF
--- a/packages/cli/src/util/get-config.ts
+++ b/packages/cli/src/util/get-config.ts
@@ -82,6 +82,19 @@ export default async function getConfig(
     return config;
   }
 
+  // Then check whether `vercel` field exists in package.json
+  const packageJsonPath = path.resolve(localPath, 'package.json');
+  const packageJson = await readJSONFile<any>(packageJsonPath);
+  if (packageJson instanceof CantParseJSONFile) {
+    return packageJson;
+  }
+  if (packageJson.vercel !== null) {
+    output.debug(`Found 'vercel' property in package.json (${packageJsonPath})`);
+    config = packageJson.vercel;
+    config[fileNameSymbol] = 'package.json';
+    return config;
+  } 
+
   // If we couldn't find the config anywhere return error
-  return new CantFindConfig([vercelFilePath, nowFilePath].map(humanizePath));
+  return new CantFindConfig([vercelFilePath, nowFilePath, packageJsonPath].map(humanizePath));
 }


styfle (changes requested):
Thanks for the PR!

We will not be accepting this feature at this time.

1. Unlike Prettier or ESlint, Vercel CLI is meant to be installed globally.
2. Since we don't own `package.json`, we don't want to add new keys to it.
3. Since `package.json` is published to npm, you don't want to publish these settings to the registry.
4. Adding a new way to define configuration would need to be updated across the entire platform – from the Git integration, to documentation, to any 3rd party tools, etc (not just Vercel CLI).
5. Vercel strives to be "zero configuration" so this is contrary to that goal.

vedantmgoyal9:
image

How can we set these through Project Settings in Vercel? If Vercel really needs to be a zero-configuration framework, then we shouldn't need any `vercel.json` at all and all the settings should be available at Project Settings.

styfle:
Makes sense! Would you mind opening a support ticket for this feature request? Let us know which props you want to assign in the dashboard. Thanks!

https://vercel.com/help#issues

vedantmgoyal9:
> 1. Unlike Prettier or ESlint, Vercel CLI is meant to be installed globally.

I don't see this as a reason. As Vercel gets configuration from `vercel.json` while being installed globally, the same way, it will get from `package.json`. There's no difference. Well, I'm not completely familiar with the large codebase of Vercel CLI, but based on file names, project structure, and references in other functions, I have added the lines of code, where it seemed best to me.

> 2. Since we don't own `package.json`, we don't want to add new keys to it.

Why will Vercel add keys to it? I haven't seen any Vercel command that auto-creates a vercel.json file with some value in it, and every time I have manually added any type of config to vercel.json. The user/developer should have at least a choice here whether to configure things though `vercel.json` or `package.json`, and it will happen when it's supported by Vercel CLI itself.

> 3. Since `package.json` is published to npm, you don't want to publish these settings to the registry.

Same as 2nd one. It's upon the will of the developer, whether he's OK with publishing to the registry or not. Also, I don't think the source code of a website is published to npm, is it? I am new to JavaScript/Node.js/NPM (or whatever you call it) ecosystem, so I don't know really.

> 4. Adding a new way to define configuration would need to be updated across the entire platform – from the Git integration, to documentation, to any 3rd party tools, etc (not just Vercel CLI).

This can be done by taking baby steps at a time. We would publish this on documentation as we have completed the feature implementation across other related tools and integrations. We can also add something like `Setting configuration though vercel.json is the recommended way, but you can also put any of the values under a "vercel" property of project's package.json. Remember to not use both of them at the same time.`

> 5. Vercel strives to be "zero configuration" so this is contrary to that goal.

I recently came to know about Vercel Cron jobs, so my thought of completely eliminating a config file wouldn't happen at any cost. On thinking about this deeper, I came across that default settings wouldn't apply to everyone, so there will be a need for some config file to share the settings so that contributors can easily test/verify their changes before making a PR to a project. We certainly don't want anyone contributing should take additional steps like going to Project settings in Vercel, changing this setting to that, and then you can set up a Dev environment for testing your changes.

vedantmgoyal9:
cc @styfle

PR #9473, opened 2023-02-17
[remix] Allow `runtime` to be defined from a parent route

TooTallNate:
Make it so that the `export const config = { runtime }` value is "inherited" from a parent route, if it is defined.

For example, "edge" can be defined in `app/root.tsx`, and so all routes will use the Edge runtime by default, unless `"runtime": "nodejs"` is used more specifically in a route deeper in the route hierarchy.


Ethan-Arrowood (approved):
<picture data-single-emoji=":lgtm:" title=":lgtm:">:lgtm:</picture>

PR #9481, opened 2023-02-18
[next] Add initial `vary` header to all prerendered pages when RSC is enabled

shuding:
This PR changes the fallback headers that relate to RSC to the defaults that Next.js currently uses. Also, it sets the initial `vary` header to all prerendered routes when RSC is enabled (`routesManifest?.rsc`), even for the pages directory. 

That's because although the pages directory won't return any RSC payload, it can still be used in a project that contains app routes. When the app route requests a page route for RSC data, it's important for the browser to not accidentally cache that result hence we need the `vary` header to set there as well.

More related discussions can be found [here](https://linear.app/vercel/issue/NEXT-382/add-vary-rsc-etc-header-to-all-responses-to-ensure-browser-caching).

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1149,7 +1149,7 @@ export async function serverBuild({
     }
   }
 
-  const rscHeader = routesManifest.rsc?.header?.toLowerCase() || '__rsc__';
+  const rscHeader = routesManifest.rsc?.header?.toLowerCase() || 'rsc';
   const completeDynamicRoutes: typeof dynamicRoutes = [];
 
   if (appDir) {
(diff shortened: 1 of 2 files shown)


packages/next/src/server-build.ts
```diff
@@ -1149,7 +1149,7 @@ export async function serverBuild({
     }
   }
 
-  const rscHeader = routesManifest.rsc?.header?.toLowerCase() || '__rsc__';
+  const rscHeader = routesManifest.rsc?.header?.toLowerCase() || 'rsc';
```
```suggestion
  const rscHeader = routesManifest.rsc?.header?.toLowerCase() || '__rsc__';
```
Shouldn't this still be `__rsc__` for backwards compat as the new value is read from the manifest?

ijjk (commented):
I think we also need to add `headers: { vary: rscVaryHeader }` to these routes as well https://github.com/vercel/vercel/blob/38f40f1c15417d8da70a9270d2d62486272360f3/packages/next/src/server-build.ts#L1414

ijjk (approved):
Thanks!

PR #9487, opened 2023-02-19
[node] fix TypeScript + ESM in Windows

Schniz:
ESM on Windows versions of Node.js require `--loader PATH` to be a `file://` protocol.
This PR allows `vc dev` to be used with both TypeScript and ESM

### DIFF
--- a/packages/node/src/fork-dev-server.ts
+++ b/packages/node/src/fork-dev-server.ts
@@ -1,0 +1,93 @@
+import once from '@tootallnate/once';
+import { cloneEnv, Config, Meta } from '@vercel/build-utils';
+import { ChildProcess, fork, ForkOptions } from 'child_process';
+import { join } from 'path';
+
+export function forkDevServer(options: {
+  tsConfig: any;
+  config: Config;
+  maybeTranspile: boolean;
+  workPath: string | undefined;
+  isTypeScript: boolean;
+  isEsm: boolean;
+  require_: NodeRequire;
+  entrypoint: string;
+  meta: Meta;
+  devServerPath?: string;
+}) {
+  let nodeOptions = process.env.NODE_OPTIONS;
+  const tsNodePath = options.require_.resolve('ts-node');
+  // const esmLoader = url.pathToFileURL(join(tsNodePath, '..', '..', 'esm.mjs'));
+  const esmLoader = join(tsNodePath, '..', '..', 'esm.mjs');
+  const cjsLoader = join(tsNodePath, '..', '..', 'register', 'index.js');
+  const devServerPath =
+    options.devServerPath || join(__dirname, 'dev-server.js');
+
+  if (options.maybeTranspile) {
+    if (options.isTypeScript) {
+      if (options.isEsm) {
+        nodeOptions = `--loader ${esmLoader} ${nodeOptions || ''}`;
+      } else {
+        nodeOptions = `--require ${cjsLoader} ${nodeOptions || ''}`;
+      }
+    } else {
+      if (options.isEsm) {
+        // no transform needed because Node.js supports ESM natively
+      } else {
+        nodeOptions = `--require ${cjsLoader} ${nodeOptions || ''}`;
+      }
+    }
+  }
+
+  const forkOptions: ForkOptions = {
+    cwd: options.workPath,
+    execArgv: [],
+    env: cloneEnv(process.env, options.meta.env, {
+      VERCEL_DEV_ENTRYPOINT: options.entrypoint,
+      VERCEL_DEV_IS_ESM: options.isEsm ? '1' : undefined,
+      VERCEL_DEV_CONFIG: JSON.stringify(options.config),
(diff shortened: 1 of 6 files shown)


packages/node/src/fork-dev-server.ts
```diff
@@ -0,0 +1,102 @@
...
+  meta: Meta;
+
+  /**
+   * A path to the dev-server path. This is used in tests.
+   */
+  devServerPath?: string;
+}) {
+  let nodeOptions = process.env.NODE_OPTIONS;
+  const tsNodePath = options.require_.resolve('ts-node');
+  const esmLoader = pathToFileURL(join(tsNodePath, '..', '..', 'esm.mjs'));
```
EndangeredMassa, line 25: **question (blocking):** what does this fix, exactly?

I tried latest `vc` on some example functions that were written in ESM/MJS style on Windows, but they worked fine. How do I reproduce the issue?

packages/node/test/dev.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    config: {},
+    isEsm: true,
+    isTypeScript: false,
+    meta: {},
+    require_: require,
+    tsConfig: undefined,
+    workPath: resolve(__dirname, './dev-fixtures'),
+    entrypoint: './esm-module.mjs',
+    devServerPath: resolve(__dirname, '../dist/dev-server.js'),
+  });
```
EndangeredMassa, line 17: **suggestion (non-blocking):** extract function

I think we could make it clearer what's being tested if we extracted this like:

```
function testForkDevServer(entrypoint: string) {
  const ext = extname(entrypoint);
  const isTypeScript = ext === '.ts';
  const isEsm = ext === '.mjs';
  return forkDevServer({
    maybeTranspile: true,
    config: {},
    isEsm,
    isTypeScript,
    meta: {},
    require_: require,
    tsConfig: undefined,
    workPath: resolve(__dirname, './dev-fixtures'),
    entrypoint,
    devServerPath: resolve(__dirname, '../dist/dev-server.js'),
  });
}
```

then calling it here would look like:

```
const child = testForkDevServer('./esm-module.mjs');
```

What do you think?

EndangeredMassa (approved):
The changes look good, but I can't reproduce the error to make sure these changes fix it. Can you provide a clearer example of the failure case?

EndangeredMassa (changes requested):
wrong button

Schniz:
The test I added reproduced on our Windows CI

Schniz:
Wrong button ;(

packages/node/src/fork-dev-server.ts
Schniz: It must be Windows + TypeScript for it to happen. If I understand correctly, when running on ESM, it is expected to have the loader in the `file://` format. Unix paths are allowed too, but full windows path is being parsed wrongly, where the drive is the protocol. So the error says “c: is not a valid protocol” 😩

makes sense?

packages/node/test/dev.test.ts
Schniz: Looks good hope I have time tomorrow to do that

packages/node/src/fork-dev-server.ts
Schniz: Again the first commits were to reproduce. Once I had a single failing test I fixed it :) so this ensures this very line fixes the issue

EndangeredMassa: I can't run this repo on windows for some reason (looking into it elsewhere), but I did create an example project that has files that look like the fixture in this PR on my windows machine. They both respond just fine with and without this PR's changes.

Has a user reported this issue or have we only seen it in Windows CI? We can certainly merge this even if it's just in Windows CI, but if it happens on user machines, I'd like to be able to reproduce to verify the fix.

Schniz: I ran it locally with a windows vm. But it originated from a user issue on Twitter and I got nerdsniped by @cramforce:

https://twitter.com/jkmartindale/status/1627072622965174273?s=46&t=8TjNUXg9oCEpSpQ4hafq_A

EndangeredMassa: I got my windows install working and was able to reproduce this with the test. This looks good to me. Thanks!

EndangeredMassa (approved):

packages/node/test/dev.test.ts
EndangeredMassa: PR: https://github.com/vercel/vercel/pull/9509

packages/node/src/index.ts
```diff
@@ -603,52 +593,24 @@ export const startDevServer: StartDevServer = async opts => {
     // In prod, we emit outputs to the filesystem.
     // In dev, we don't emit because we use ts-node.
     tsConfig.compilerOptions.noEmit = true;
```
styfle, line 595: Why was this one left here but the rest were moved?

PR #9489, opened 2023-02-20
[next] Handle prerender-manifest v4

ijjk:
Updates to handle the updated fields in the `prerender-manifest@v4` which is required for new app ISR handling. Tests added in https://github.com/vercel/next.js/pull/46133

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -849,16 +849,18 @@ export type NextPrerenderedRoutes = {
   staticRoutes: {
     [route: string]: {
       initialRevalidate: number | false;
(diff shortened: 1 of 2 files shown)


packages/next/src/utils.ts
```diff
@@ -1696,6 +1734,7 @@ export const onPrerenderRouteInitial = (
       };
     }
 
+    console.log('adding nonLambda', route);
```
timneutkens: Debug line?

timneutkens:
```suggestion
```

timneutkens (approved):

PR #9506, opened 2023-02-22
[remix] Restore support for a custom `server.js` entrypoint file

TooTallNate:
When the project defines a custom `server.js` file in `remix.config.js`, then don't override that with the bundled
`server-node.mjs`/`server-edge.mjs` file. This allows for a custom `getLoadContext()` function to remain supported.

In order for Edge runtime to be supported, a [companion PR](https://github.com/remix-run/remix/pull/5537) for `@remix-run/vercel` adapter has been created with a dedicated entrypoint for when Edge runtime is being used. In order to support Edge on previous versions of the adapter, we may end up patching the package to enable support, but that will happen in a follow-up PR.

Related to https://github.com/orgs/vercel/discussions/1596.


packages/remix/src/build.ts
```diff
@@ -339,17 +345,21 @@ async function createRenderNodeFunction(
...
+  let handler = relative(rootDir, serverBuildPath);
+  let handlerPath = join(rootDir, handler);
+  if (!serverEntryPoint) {
+    handler = join(dirname(handler), 'server-edge.mjs');
+    handlerPath = join(rootDir, handler);
 
-  // Copy the `server-edge.mjs` file into the "build" directory
-  const sourceHandlerPath = join(__dirname, '../server-edge.mjs');
-  await fs.copyFile(sourceHandlerPath, handlerPath);
+    // Copy the `server-node.mjs` file into the "build" directory
```
EndangeredMassa: **suggestion (soft-blocking):** revert this line

```suggestion
    // Copy the `server-edge.mjs` file into the "build" directory
```

Was this changed in error? The code below this comment copies `server-edge.mjs`

packages/remix/test/fixtures/02-remix-basics-mjs/remix.config.mjs
```diff
@@ -2,9 +2,13 @@
  * @type {import('@remix-run/dev').AppConfig}
  */
 export default {
-  ignoredRouteFiles: ['**/.*'],
+  ignoredRouteFiles: ["**/.*"],
+  // When running locally in development mode, we use the built-in remix
+  // server. This does not understand the vercel lambda module format,
+  // so we default back to the standard build output.
+  server: process.env.NODE_ENV === "development" ? undefined : "./server.ts",
+  //serverBuildPath: "api/index.js",
```
EndangeredMassa: **suggestion (non-blocking):** fix spacing

```suggestion
  // serverBuildPath: "api/index.js",
```

EndangeredMassa (dismissed):

packages/remix/src/build.ts
TooTallNate: Ya good catch

EndangeredMassa (approved):

packages/remix/src/build.ts
```diff
@@ -339,17 +345,21 @@ async function createRenderNodeFunction(
...
-  serverBuildPath: string
+  serverBuildPath: string,
+  serverEntryPoint: string | undefined
 ): Promise<EdgeFunction> {
   const files: Files = {};
 
-  const relativeServerBuildPath = relative(rootDir, serverBuildPath);
-  const handler = join(dirname(relativeServerBuildPath), 'server-edge.mjs');
-  const handlerPath = join(rootDir, handler);
+  let handler = relative(rootDir, serverBuildPath);
```
lauchness, line 353: Does this change break support for Remix apps?

TooTallNate: I don't understand the question, can you elaborate? This PR restores support for specifying an explicit `server.js` file in your Remix app.

PR #9508, opened 2023-02-22
[cli][dev] remove link to 404 FAQ

EndangeredMassa:
In a previous PR (https://github.com/vercel/vercel/pull/8976/files#diff-bd86ad03c67b58fd8df71509a96d1e303343068bbe3addc19cd3cb7247db98b4R10), a link was added to the 404 template that directed the user to https://vercel.link/404. The information at that link talks about deployments, which is confusing working from `vc dev` locally.

This appears to be the only reference to https://vercel.link/404. Can anyone find others?

Should we make a new FAQ entry we can link to or just remove this link? If we want specific documentation about this, should it live under https://vercel.com/docs/cli ?

---

- [Slack Discussion](https://vercel.slack.com/archives/C03F2CMNGKG/p1677081506021239)


packages/cli/src/util/dev/templates/error_404.jst
```diff
@@ -6,6 +6,4 @@
     {{?}}
     <span class="devinfo-line">ID: <code>{{! it.request_id }}</code>
   </p>
-  
-  <a href="https://vercel.link/404"><div class="note">Click here to learn more about this error.</div></a>
```
styfle, line 10: I believe this code is generated so it would probably end up here again

EndangeredMassa: I believed these were manually updated by us from time to time

styfle (approved):

cb1kenobi (approved):

PR #9510, opened 2023-02-22
[tests] Add cron job to update gatsby fixtures

Ethan-Arrowood:
PR to add a cron job and script that automatically updates gatsby fixture versions.


Ethan-Arrowood:
Another option to doing the whole `npm view` check is to use: `npm i gatsby@<tag> --save-exact --package-lock-only` which will update the `gatsby` package version to the right semver value (i.e. `4.24.4`) and update the package lock file all in one...without actually installing all the dependencies I think

utils/update-gatsby-fixtures.js
```diff
@@ -0,0 +1,111 @@
+const path = require('path');
```
TooTallNate, line 1: Can we do TypeScript for these files and invoke with `ts-node`?

Ethan-Arrowood: The github action thing we use to execute doesn't have typescript support it seems https://github.com/actions/github-script/issues/294

Is this a blocker? I we use this pattern in vercel/api as well as the other cron scripts here in this repo so figuring out typescript support seems a bit out of scope

utils/update-gatsby-fixtures.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * invoked by ../.github/workflows/cron-test-and-update-gatsby.yml
+ * @param {{ github: ReturnType<import('@actions/github').getOctokit>, context: import('@actions/github').context }} param0 Defined by https://github.com/actions/github-script
+ * @returns
+ */
+module.exports = async ({ github, context }) => {
+  exec('git', ['config', '--global', 'user.email', 'infra+release@vercel.com']);
+  exec('git', ['config', '--global', 'user.name', 'vercel-release-bot']);
+  exec('git', ['checkout', branch]);
```
TooTallNate:
```suggestion
  try {
    // Branch may exist if there's already an existing PR
    exec('git', ['checkout', branch]);
  } catch {
    exec('git', ['checkout', '-b', branch]);
  }
```

utils/update-gatsby-fixtures.js
```diff
@@ -0,0 +1,111 @@
...
+        `gatsby version ${newVersion} did not change for fixture ${fixture}, skipping update.`
+      );
+      continue;
+    }
+
+    packageJSON.dependencies.gatsby = newVersion;
+
+    fs.writeFileSync(
+      packageJSONPath,
+      JSON.stringify(packageJSON, null, 2),
```
TooTallNate:
```suggestion
      JSON.stringify(packageJSON, null, 2) + '\n',
```

utils/update-gatsby-fixtures.js
```diff
@@ -0,0 +1,111 @@
...
+    );
+
+    // update lockfiles
+    if (fixture.includes('pnpm')) {
+      exec('pnpm', ['install', '--lockfile-only']);
+    } else {
+      exec('npm', ['install', '--package-lock-only']);
+    }
+  }
+
```
TooTallNate, line 77: Should check if there were any updates here, and bail if not.

utils/update-gatsby-fixtures.js
EndangeredMassa: If there's an existing branch, do we need to merge main before continuing?

TooTallNate (approved):

utils/update-gatsby-fixtures.js
```diff
@@ -0,0 +1,123 @@
...
+ * invoked by ../.github/workflows/cron-test-and-update-gatsby.yml
+ * @param {{ github: ReturnType<import('@actions/github').getOctokit>, context: import('@actions/github').context }} param0 Defined by https://github.com/actions/github-script
+ * @returns
+ */
+module.exports = async ({ github, context }) => {
+  exec('git', ['config', '--global', 'user.email', 'infra+release@vercel.com']);
+  exec('git', ['config', '--global', 'user.name', 'vercel-release-bot']);
+  try {
+    // Branch may exist if there's already an existing PR
+    exec('git', ['checkout', branch]);
```
EndangeredMassa, line 17: **question (non-blocking):** should we also merge main in this case?

Ethan-Arrowood: i assume itd happen automatically cause of automerge or our manual input. if it becomes an issue Ill modify

PR #9511, opened 2023-02-22
[remix] Handle escaped characters in route names

TooTallNate:
Remix allows to use `[]` brackets to escape the characters within the brackets in the URL path. So we need to escape those characters as well (and remove the brackets) from the RegExp that gets created for `routes`.

See: https://remix.run/docs/en/v1/guides/resource-routes#url-escaping

Fixes: https://github.com/orgs/vercel/discussions/1594 (probably)


styfle (approved):

TooTallNate:
This is not the correct approach. Remix already removes the brackets by the time `getRegExpFromPath()` is invoked.

PR #9520, opened 2023-02-23
[node-bridge] Fix multiple `Set-Cookie` headers in streaming mode

TooTallNate:
Multiple `Set-Cookie` headers are being concat'd instead of being sent individually when in streaming mode.

**Non-streaming:**

```
$ curl -v https://01-remix-basics-eqztqyq3y.vercel-support.app/set-cookie 2>&1 | grep cookie
* h2h3 [:path: /set-cookie]
> GET /set-cookie HTTP/2
< set-cookie: foo=bar
< set-cookie: sessionId=38afes7a8
```

**Streaming:**

```
$ curl -v https://01-remix-basics-9b18alizh.vercel-support.app/set-cookie 2>&1 | grep cookie
* h2h3 [:path: /set-cookie]
> GET /set-cookie HTTP/2
< set-cookie: foo=bar,sessionId=38afes7a8
```


ijjk (approved):

styfle (approved):
Thanks! It would be good to add a test 👍

cb1kenobi (approved):

PR #9527, opened 2023-02-23
[cli][dev] warn and timeout on bad serverless function response

EndangeredMassa:
When writing an Edge Function, you may forget to:

```
export const config = {
  runtime: 'edge',
};
```

leaving a Serverless Function (by default) that looks like:

```
export default (req: Request) => {
  return new Response("not the correct signature");
};
```

In current `vc dev`, this will hang forever and provide no feedback.

After this PR, `vc dev` will show a warning and eventually time out the request.

---

CleanShot 2023-02-23 at 09 54 14@2x

CleanShot 2023-02-23 at 09 54 24@2x


packages/node-bridge/launcher.js
```diff
@@ -3,6 +3,11 @@ const { createServer, Server } = require('http');
 const { isAbsolute } = require('path');
 const { Bridge } = require('./bridge.js');
 
+// TODO: is a 900-second timeout even useful for most cases?
+// Set to max timeout for ENT plans because we don't
+// know what plan the current user is on
+const SERVERLESS_FUNCTION_TIMEOUT = 900 * 1000; // 900 seconds
```
EndangeredMassa, line 8: We don't currently know the plan type when running `vc dev`. So, I set this timeout the max possible, which is 900 seconds for ENT.

Having a timeout at all is kinda nice, but I imagine this not being super useful until we get plan types in here.

packages/node-bridge/launcher.js
```diff
@@ -132,6 +142,54 @@ function getVercelLauncher({
   };
 }
 
+/**
+ * @param {Function} listenerHandler
+ * @returns {(req: import('./types').VercelRequest, res: import('./types').VercelResponse) => void}
+ */
+function wrapListener(listenerHandler) {
+  // TODO: how can we best test the timeout and warning behavior?
```
EndangeredMassa: We could pass the timeout value through an env var to `vc dev` from an integration test

packages/node-bridge/launcher.js
```diff
@@ -132,6 +142,54 @@ function getVercelLauncher({
...
+
+/**
+ * @param {Function} listenerHandler
+ * @param {import('./types').VercelRequest} req
+ * @param {import('./types').VercelResponse} res
+ */
+async function validateResponse(listenerHandler, req, res) {
+  const resp = await listenerHandler(req, res);
+  if (typeof resp !== 'undefined') {
+    // TODO: add vercel.link ?
```
EndangeredMassa: We could make a vercel.link (or use an existing one, if it exists) that points to https://vercel.com/docs/concepts/functions/serverless-functions

packages/node-bridge/launcher.js
```diff
@@ -132,6 +142,54 @@ function getVercelLauncher({
...
+    // TODO: only show this once per invocation of this function?
```
EndangeredMassa: It could get really annoying if you wanted to return a value from your serverless function, for some reason. I don't know when you'd want this, though.

We could ship this and see if anyone complains.

packages/node-bridge/launcher.js
EndangeredMassa: Created https://vercel.link/serverless-function-quickstart

EndangeredMassa:
With the migration of node-bridge, we have to restart this PR to make it work

PR #9528, opened 2023-02-23
[tests] add tests to edge handler template

EndangeredMassa:
The Edge Handler wrapping logic we add during `vc dev` was not testable. This PR refactors it to be testable and adds some basic tests.


TooTallNate (dismissed):

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -69,13 +69,28 @@ async function compileUserCode(
 
       // user code
       ${compiledFile.text};
+      const userEdgeHandler = module.exports.default;
```
cb1kenobi, line 72: **question (non-blocking):** I follow that the ref to `module.exports.default` was moved from the template into here, but I don't get why it needed to be moved here. In theory, `registerFetchListener()` would be eval'd in the same context and thus resolve the same `module.exports.default`.

cb1kenobi (dismissed):
LGTM, tests ran locally.

packages/node/src/edge-functions/edge-handler.ts
EndangeredMassa: The `module.exports.default` was looking for the user's code from `compiledFile`. I put it right after the `compiledFile` interpolation to make that relationship visual.

The `edgeHandlerTemplate` is not related to this `moduled.exports.default`. The template comes later.

cb1kenobi (approved):

TooTallNate (approved):

PR #9531, opened 2023-02-23
[remix] Ignore `serverBuildPath` from Remix config

### LINKED ISSUE #9524: Error: File /vercel/path0/api/index.js does not exist
**Vercel deployment for remix is not working after the latest update.**

No build errors but throwing error while deployment - `Error: File /vercel/path0/api/index.js does not exist.`
Tried deploying a new branch or redeployed already built branch, failed with the above error. This happened after today's latest update.

image

TooTallNate:
We are explicitly overriding `serverBuildPath` so there is no need to use the value that `readConfig()` gives to us.

Fixes #9524.

### DIFF
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -96,6 +96,7 @@ export const build: BuildV2 = async ({
   spawnOpts.env.NODE_ENV = 'production';
 
   let remixConfig = await chdirAndReadConfig(entrypointFsDirname);
+  console.log(remixConfig);
   const { serverEntryPoint } = remixConfig;
 
   // We need to patch the `remix.config.js` file to force some values necessary
@@ -168,6 +169,7 @@ module.exports = config;`;
       }
     }
     remixConfig = await chdirAndReadConfig(entrypointFsDirname);
+    console.log(remixConfig);
   } finally {
     // Clean up our patched `remix.config.js` to be polite
     if (remixConfigPath && renamedRemixConfigPath) {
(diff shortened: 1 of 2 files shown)


packages/remix/src/build.ts
```diff
@@ -167,17 +168,13 @@ module.exports = config;`;
         });
       }
     }
-    remixConfig = await chdirAndReadConfig(entrypointFsDirname);
```
TooTallNate, line 170: This second read `readConfig()` call was not actually doing anything, since `readConfig()` uses `await import()` and thus the module was cached and didn't pick up our patches.

packages/remix/test/fixtures/02-remix-basics-mjs/remix.config.mjs
```diff
@@ -3,6 +3,7 @@
  */
 export default {
   ignoredRouteFiles: ["**/.*"],
+  serverBuildTarget: "vercel",
```
cb1kenobi, line 6: Admittedly I'm a bit out-of-the-loop, but Remix docs say `serverBuildTarget` has been deprecated. Is this line just testing that `config.serverBuildTarget = undefined;` scrubs the config value?

TooTallNate: It is deprecated, but still exists in projects out in the wild. This PR is basically for backwards compat.

cb1kenobi (approved):

Ethan-Arrowood (approved):

styfle (approved):

PR #9538, opened 2023-02-24
[tests] Properly test for multiple headers with same name in probes

TooTallNate:
Updates the `responseHeaders` probe checks to properly test for multiple headers with the same name.

Previously the probes were using `headers.get()` which concats multiple headers into a single string, which results in the test not really checking if there are in fact multiple headers with the same name. Using `headers.raw()` allows us to properly test for that.

A couple of Python tests that were already checking for multiple `set-cookie` headers needed to be updated to match the full value, since the check now properly validates the full string match of each header (before it was basically just doing a `string.includes()` check).

This is a precursor for https://github.com/vercel/vercel/pull/9533.

### DIFF
--- a/test/lib/deployment/test-deployment.js
+++ b/test/lib/deployment/test-deployment.js
@@ -222,24 +222,46 @@ async function runProbe(probe, deploymentId, deploymentUrl, ctx) {
 
   if (probe.responseHeaders) {
     // eslint-disable-next-line no-loop-func
+    const rawHeaders = resp.headers.raw();
     Object.keys(probe.responseHeaders).forEach(header => {
-      const actual = resp.headers.get(header);
-      const expected = probe.responseHeaders[header];
-      const isEqual = Array.isArray(expected)
-        ? expected.every(h => actual.includes(h))
-        : typeof expected === 'string' &&
-          expected.startsWith('/') &&
-          expected.endsWith('/')
-        ? new RegExp(expected.slice(1, -1)).test(actual)
-        : expected === actual;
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood (approved):

cb1kenobi (approved):

test/lib/deployment/test-deployment.js
```diff
@@ -220,26 +220,47 @@ async function runProbe(probe, deploymentId, deploymentUrl, ctx) {
...
-        : typeof expected === 'string' &&
-          expected.startsWith('/') &&
-          expected.endsWith('/')
-        ? new RegExp(expected.slice(1, -1)).test(actual)
-        : expected === actual;
-      if (!isEqual) {
-        const headers = Array.from(resp.headers.entries())
-          .map(([k, v]) => `  ${k}=${v}`)
-          .join('\n');
+      const actualArr = rawHeaders[header.toLowerCase()];
```
styfle, line 230: Are raw headers always lowercase keys?

styfle (approved):

test/lib/deployment/test-deployment.js
TooTallNate: They are not.

styfle: So why is the `header.toLowerCase()` there? Are we assuming the server will always lowercase the key before sending the response?

TooTallNate: I guess so. Some existing tests were using lowercase.

PR #9542, opened 2023-02-24
[build-utils] Handle error during `readConfigFile()`

styfle:
This fixes a confusing error:

```
$ vercel build
Vercel CLI 28.16.6
Error: duplicated mapping key in "/vercel/path0/pnpm-lock.yaml" at line 1215, column -164:
      /@react-dnd/asap/4.0.1:
      ^
```

### DIFF
--- a/packages/build-utils/src/fs/read-config-file.ts
+++ b/packages/build-utils/src/fs/read-config-file.ts
@@ -25,12 +25,16 @@ export async function readConfigFile<T>(
 
     if (data) {
       const str = data.toString('utf8');
-      if (name.endsWith('.json')) {
-        return JSON.parse(str) as T;
-      } else if (name.endsWith('.toml')) {
-        return (toml.parse(str) as unknown) as T;
-      } else if (name.endsWith('.yaml') || name.endsWith('.yml')) {
-        return yaml.safeLoad(str, { filename: name }) as T;
+      try {
+        if (name.endsWith('.json')) {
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):

packages/build-utils/src/fs/read-config-file.ts
```diff
@@ -25,12 +25,16 @@ export async function readConfigFile<T>(
...
+      try {
+        if (name.endsWith('.json')) {
+          return JSON.parse(str) as T;
+        } else if (name.endsWith('.toml')) {
+          return toml.parse(str) as unknown as T;
+        } else if (name.endsWith('.yaml') || name.endsWith('.yml')) {
+          return yaml.safeLoad(str, { filename: name }) as T;
+        }
+      } catch (error: unknown) {
+        console.log(`Error while parsing config file: "${name}"`);
```
pvdz, line 37: Should it not tell the user why it was trying to do this, and/or why it can't continue the build now?
Do we not want to print the actual error that was thrown as a way for the user to debug/see what's wrong?

pvdz (approved):
:yay:

packages/build-utils/src/fs/read-config-file.ts
styfle: It will continue the build. The idea is that whatever tool uses the config file later will have a hard error such as `pnpm install`

PR #9549, opened 2023-02-24
[cli] Added fetch dist-tags http status validation

cb1kenobi:
If for some reason the `get-latest-worker.js` fails to get the dist-tags, it doesn't check the status code and proceeds to `JSON.parse()` the response. If for example npm returns 401 Unauthorized, you get a cryptic JSON parse error message.

### DIFF
--- a/packages/cli/src/util/get-latest-version/get-latest-worker.js
+++ b/packages/cli/src/util/get-latest-version/get-latest-worker.js
@@ -247,6 +247,14 @@ async function fetchDistTags(name) {
         });
         res.on('end', () => {
           try {
+            if (res.statusCode && res.statusCode >= 400) {
+              return reject(
+                new Error(
+                  `Fetch dist-tags failed ${res.statusCode} ${res.statusMessage}`
+                )
+              );
+            }
+
             resolve(JSON.parse(buf));
           } catch (err) {
             reject(err);


packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -247,6 +247,14 @@ async function fetchDistTags(name) {
         });
         res.on('end', () => {
           try {
+            if (res.statusCode && res.statusCode >= 400) {
+              return reject(
```
Ethan-Arrowood, line 251: Don't think you need the return here

Ethan-Arrowood (approved):

packages/cli/src/util/get-latest-version/get-latest-worker.js
cb1kenobi: This `return` will prevent the rejected promise from attempting to resolve/reject later. To my knowledge, `reject()` does not stop the execution of the current scope.

packages/cli/src/util/get-latest-version/get-latest-worker.js
```diff
@@ -247,6 +247,14 @@ async function fetchDistTags(name) {
         });
         res.on('end', () => {
           try {
+            if (res.statusCode && res.statusCode >= 400) {
```
styfle, line 250: Can you use (!res.ok) instead?

cb1kenobi: This is using Node's built-in `https.get()` in which the handler receives a `ServerResponse` and that response object doesn't have an `ok` property.

styfle (approved):

packages/cli/src/util/get-latest-version/get-latest-worker.js
Ethan-Arrowood: You're correct, it doesn't stop the execution, but I don't believe you can resolve after a promise rejects

EndangeredMassa: I'd rather not allow further execution that happens to not do anything. I think it's more correct and less brittle to return here.

PR #9550, opened 2023-02-24
[build-utils] Change `debug()` to accept a single string param

styfle:
This is technically a semver major change but will improve our logging because we can optionally turn on debugging in the build container.

- Before: `debug(obj)`
- After: `debug(JSON.stringify(obj))`

If there are any newlines then they will be replaced to avoid accidentally printing when you don't intend to.

### DIFF
--- a/packages/build-utils/src/debug.ts
+++ b/packages/build-utils/src/debug.ts
@@ -1,9 +1,6 @@
-import { getPlatformEnv } from './get-platform-env';
-
-export default function debug(message: string, ...additional: any[]) {
-  if (getPlatformEnv('BUILDER_DEBUG')) {
-    console.log(message, ...additional);
-  } else if (process.env.VERCEL_DEBUG_PREFIX) {
-    console.log(`${process.env.VERCEL_DEBUG_PREFIX}${message}`, ...additional);
+export default function debug(message: string) {
+  if (process.env.VERCEL_DEBUG_PREFIX) {
+    const line = message.replace(/\r?\n/g, ' ');
+    console.log(`${process.env.VERCEL_DEBUG_PREFIX}${line}`);
   }
 }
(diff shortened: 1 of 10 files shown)


packages/build-utils/src/debug.ts
```diff
@@ -1,9 +1,6 @@
-import { getPlatformEnv } from './get-platform-env';
-
-export default function debug(message: string, ...additional: any[]) {
-  if (getPlatformEnv('BUILDER_DEBUG')) {
-    console.log(message, ...additional);
-  } else if (process.env.VERCEL_DEBUG_PREFIX) {
-    console.log(`${process.env.VERCEL_DEBUG_PREFIX}${message}`, ...additional);
+export default function debug(message: string) {
+  if (process.env.VERCEL_DEBUG_PREFIX) {
+    const line = message.replace(/\r?\n/g, ' ');
```
cb1kenobi, line 3: I think this is a bit too restrictive. I would recommend continuing to accept `message` and `additional`, but pass them into `util.format()`: https://nodejs.org/dist/latest-v18.x/docs/api/util.html#utilformatformat-args. Then split on lines and call `console.log()` for each line:

```
const lines = util.format(message, ...additional).split(/\r?\n/);
for (const line of lines) {
  console.log(`${process.env.VERCEL_DEBUG_PREFIX}${line}`);
}
```

packages/go/go-helpers.ts
```diff
@@ -189,7 +188,7 @@ async function parseGoVersion(modulePath: string): Promise<string> {
     } else {
       console.log(`Warning: Unknown Go version in ${file}`);
     }
-  } catch (err) {
+  } catch (err: any) {
```
Ethan-Arrowood, line 191: Make this `unknown` and use `isError` from `@vercel/error-utils`

Ethan-Arrowood (changes requested):

PR #9551, opened 2023-02-24
[hydrogen] Add missing envPrefix

styfle:
- Follow up to #9326


TooTallNate:
Let's add a test? Do we need to apply these at runtime as well?

cb1kenobi (approved):

PR #9555, opened 2023-02-25
[next] Fix deploying with cached build

ijjk:
This leverages the relative app dir field in the required files manifest when available so that we don't rely on absolute paths from a cached build since the cache can be restored in a separate context where the value no longer applies.

x-ref: https://github.com/vercel/next.js/pull/45864
x-ref: https://github.com/vercel/next.js/pull/46393
x-ref: https://github.com/vercel/next.js/discussions/39432#discussioncomment-4914549
x-ref: [slack thread](https://vercel.slack.com/archives/C03S8ED1DKM/p1675821643786319)


styfle (approved):

packages/next/test/fixtures/00-cached-build/vercel.json
```diff
@@ -0,0 +1,30 @@
...
+      "src": "package.json",
+      "use": "@vercel/next",
+      "config": {
+        "rootDirectory": "apps/web",
+        "buildCommand": "cd ../../ && yarn build",
+        "installCommand": "yarn"
+      }
+    }
+  ],
+  "probes": [
```
styfle, line 13: Note that you can move this into a separate probes.json file

PR #9556, opened 2023-02-25
[tests] Use `probes.json` instead of `vercel.json` for e2e tests

TooTallNate:
Makes it easier to test fixtures directly via CLI deployments.

Coincidentally, this also revealed that auto-detection of Remix apps wasn't working correctly if there's a `remix.config.mjs` file. So `@vercel/frameworks` is updated to account for that case as well.


packages/frameworks/src/frameworks.ts
```diff
@@ -202,12 +202,13 @@ export const frameworks = [
...
-      every: [
-        // Intentionally does not detect a package name
-        // https://github.com/vercel/vercel/pull/7761
+      some: [
         {
           path: 'remix.config.js',
         },
+        {
+          path: 'remix.config.mjs',
+        },
```
styfle: Let's separate this change into a different PR because the changes won't take affect until its updated in `vercel/api`

TooTallNate: Sounds good.

TooTallNate: https://github.com/vercel/vercel/pull/9558

styfle (approved):
Thanks! One of these days we need to run a script to update all of the fixtures in the repo 😄

PR #9558, opened 2023-02-25
[frameworks] Check for `remix.config.mjs` for "remix" auto-detection

TooTallNate:
Moving this to a separate PR based on this comment: https://github.com/vercel/vercel/pull/9556#discussion_r1117939290

This will need to be merged and released before the relevant E2E test in #9556 will pass.

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -202,11 +202,12 @@ export const frameworks = [
     useRuntime: { src: 'package.json', use: '@vercel/remix' },
     ignoreRuntimes: ['@vercel/node'],
     detectors: {
-      every: [
-        // Intentionally does not detect a package name
-        // https://github.com/vercel/vercel/pull/7761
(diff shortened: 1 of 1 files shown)


styfle (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -202,12 +202,13 @@ export const frameworks = [
     useRuntime: { src: 'package.json', use: '@vercel/remix' },
     ignoreRuntimes: ['@vercel/node'],
     detectors: {
-      every: [
-        // Intentionally does not detect a package name
-        // https://github.com/vercel/vercel/pull/7761
```
EndangeredMassa, line 207: Why was this comment removed? It's a note that makes sure we don't think we can replace this with something like `matchPackage: 'remix'` later.

I don't think it's necessarily required, but if we're removing these comments, let's remove the other comments about this from this file.

PR #9565, opened 2023-02-27
[cli] Remove unncessary warning when monorepo setting override

styfle:
This warning is not actionable by the user so this PR removes it.


EndangeredMassa (approved):

styfle:
Going to force merge since it looks like GH Actions is having an outage causing some network requests to fail (therefore unrelated to this PR)

PR #9566, opened 2023-02-27
[tests] resolve sharp to a dummy small package to avoid install errors

EndangeredMassa:
Try setting:

```
"pnpm": {
  "overrides": {
    "gatsby-sharp>sharp": "npm:is-odd-num@1.1.1"
  }
},
```

to avoid install errors with `sharp`. We don't need it.

Uses [pnpm overrides](https://pnpm.io/package_json#pnpmoverrides).

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,11 @@
     "ts-jest": "28.0.5",
     "turbo": "1.7.4"
   },
+  "pnpm": {
+    "overrides": {
+      "gatsby-sharp>sharp": "npm:is-odd-num@1.1.1"
+    }
+  },
   "scripts": {
     "lerna": "lerna",
     "version": "pnpm install && git add pnpm-lock.yaml",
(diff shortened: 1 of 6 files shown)


TooTallNate (dismissed):

package.json
```diff
@@ -34,6 +34,11 @@
     "ts-jest": "28.0.5",
     "turbo": "1.7.4"
   },
+  "pnpm": {
+    "overrides": {
+      "gatsby-sharp>sharp": "npm:is-odd-num@1.1.1"
```
styfle, line 39: This package is actually quite large https://packagephobia.com/result?p=is-odd-num

styfle: I think this is going to lead to more confusion if someone ever tries to add a test for gatsby and expects sharp to work (for example `vc dev` tests)

styfle (changes requested):
I think this is going to lead to more confusion if someone ever tries to add a test for gatsby and expects sharp to work (for example vc dev tests).

package.json
cb1kenobi: Why `is-odd-num` needed to add `mocha` as a `dependency` instead of a `devDependency` is puzzling.

Issue #9393, opened 2023-02-07
Vercel Go Serverless Functions don't deploy


vedantmgoyal9:
### Discussed in https://github.com/vercel/vercel/discussions/9381

<div type='discussions-op-text'>

<sup>Originally posted by **vedantmgoyal2009** February  5, 2023</sup>
Repository: https://github.com/vedantmgoyal2009/winget-manifests-manager/

Vercel deployment always fails with this error:

![image](https://user-images.githubusercontent.com/83997633/216812849-a17ca915-fbb5-4147-9aeb-5e564c4d463c.png)

According to the message, it looks like it is because of `@vercel` of path `<dir>/@vercel/go/dist`. This needs to be patched at the earliest.</div>

vedantmgoyal2009:
I'm using `replace` directive in go.mod.

![image](https://user-images.githubusercontent.com/83997633/217918151-69794c2a-997d-4f34-9b26-3d2e62b01433.png)

vedantmgoyal9:
* https://github.com/vercel/vercel/pull/9576
* https://github.com/vercel/vercel/issues/9393
* https://github.com/vercel/vercel/pull/9367

vedantmgoyal9:
@cb1kenobi this is still not fixed. https://github.com/vedantmgoyal2009/winget-manifests-manager

cb1kenobi:
@vedantmgoyal2009 I was able to look at your project and created a simplified reproduction test case. I found and fixed a few issues that I found when running `vc dev`. `vc build` seemed to work. PR is here: https://github.com/vercel/vercel/pull/9708. Please take a peek and let me know if I'm on target or if I need to fix something else.

vedantmgoyal9:
![image](https://user-images.githubusercontent.com/83997633/227399014-a8ede146-5a0a-45df-8a82-e063b4b3aa08.png)

cb1kenobi:
@vedantmgoyal2009 I [saw](https://github.com/vercel/vercel/pull/9708#issuecomment-1482109650) in the latest PR that you found the CLI link. Is the above screenshot using that build?

```
vc deploy -b VERCEL_CLI_VERSION=https://vercel-3bkhxtt1s.vercel.sh/tarballs/vercel.tgz --prod
```

vedantmgoyal9:
@cb1kenobi yes 👍🏻

cb1kenobi:
[...]

PR #9576, opened 2023-02-28
[go] Go builder improvements

cb1kenobi:
This PR fixes a handful of Go builder issues all related to the selected Golang version being used to build the function:

- `go.mod` version ignored for `vc build` and `vc dev`, uses system `PATH` version only
- `vc dev` fails if `go.mod` does not exist
- If the analyze bin doesn’t exist, downloads golang into `.vercel/cache/golang` instead of a global shared dir
- When running `vc dev`, doesn’t reuse go build code/common settings
- go tidy fails when `go.mod` set to 1.19 or 1.20, but 1.18 or older is installed
- `vc build` builds wrong arch on Apple Silicon/arm64
- `vc build` on Windows doesn't properly resolve "builds" in `vercel.json` due to posix separator issue
- `vc build` on Windows fails with `package <pkg/name> is not in GOROOT` due to posix separator issue
- Removed `actions/setup-go` from all test workflows

I added a test that tests the `go tidy` issue.


vedantmgoyal2009:
Please merge this so that I can resolve issues with the project at my end.

vedantmgoyal9 (dismissed):

packages/go/go-helpers.ts
```diff
@@ -112,72 +112,146 @@ class GoWrapper {
...
+  // check we have the desired `go` version cached
+  if (await pathExists(goGlobalBinDir)) {
+    // check if `go` has already been downloaded and that the version is correct
+    env.GOROOT = goGlobalDir;
+    env.PATH = `${goGlobalBinDir}${delimiter}${env.PATH}`;
+    const { failed, stdout } = await execa('go', ['version'], {
+      env,
+      reject: false,
+    });
+    if (!failed) {
```
EndangeredMassa: **suggestions (non-blocking):** Can we debug log when this does fail?

packages/go/go-helpers.ts
```diff
@@ -112,72 +112,146 @@ class GoWrapper {
...
+  // we need to download and cache the desired `go` version
+  const url = getGoUrl(goSelectedVersion, platform, arch);
+  debug(`Downloading go: ${url}`);
+  const res = await fetch(url);
+
+  if (!res.ok) {
+    throw new Error(`Failed to download: ${url} (${res.status})`);
+  }
+
+  // TODO: use a zip extractor when `ext === "zip"`
```
EndangeredMassa: **question (non-blocking):** when will this be a `.zip`?

EndangeredMassa (dismissed):
There's a lot in here, but it looks good to me. I did some manual testing locally with some go projects. Everything seemed to work.

Thanks!

packages/go/go-helpers.ts
```diff
@@ -1,16 +1,19 @@
...
-  ['1.19', '1.19.5'],
-  ['1.18', '1.18.1'],
-  ['1.17', '1.17.3'],
-  ['1.16', '1.16.10'],
-  ['1.15', '1.15.8'],
+  ['1.19', '1.19.6'],
+  ['1.18', '1.18.10'],
+  ['1.17', '1.17.13'],
+  ['1.16', '1.16.15'],
+  ['1.15', '1.15.15'],
```
styfle, line 29: I would prefer the version bump be a separate PR

packages/go/test/fixtures/10-go-mod-newer/README.md
```diff
@@ -0,0 +1,3 @@
+Build container has Go 1.18 and this test has a `go.mod` that says to use
```
styfle: I don't believe the build container has Go installed at all. It should install and cache on the first deployment.

packages/go/go-helpers.ts
```diff
@@ -32,46 +35,39 @@ const getGoUrl = (version: string, platform: string, arch: string) => {
   return `https://dl.google.com/go/go${version}.${goPlatform}-${goArch}.${ext}`;
 };
 
+const goGlobalCachePath = join(XDGAppPaths('com.vercel.cli').cache(), 'golang');
```
styfle: Whats' the difference between this cache path and the `cacheDir` on line 22?

packages/go/go-helpers.ts
cb1kenobi: Turns out that all Go distributions for Windows are `.zip` files and it apparently has never been supported. I added the ability to extract zip files and now Windows is happy.

packages/go/go-helpers.ts
cb1kenobi: Originally, I had logging in there, but it was making the logs too verbose. In this case, it doesn't really matter that it failed because it'll still try the system PATH branch before downloading the correct version.

packages/go/go-helpers.ts
cb1kenobi: My thought with bumping them here is that they will actually get tested, which I have done for 1.18 and 1.19.

packages/go/test/fixtures/10-go-mod-newer/README.md
cb1kenobi: Hmm, I believe you are correct. I'll fix this messaging.

packages/go/go-helpers.ts
cb1kenobi: The `cacheDir` is in the project directory and is used by `vc dev` to temporarily build and run the Go handlers.

`goGlobalCachePath` is where the Golang SDKs are installed and they are shared across projects as well as used to build the `analyze` binary on first use. The `goGlobalCachePath` is in the `com.vercel.cli` user data directory where package update metadata is stored.

EndangeredMassa (dismissed):

packages/go/index.ts
```diff
@@ -289,18 +289,20 @@ export async function build({
...
-          cwd: entrypointDirname,
-          stdio: 'inherit',
+    const go = await createGo({
+      goPath,
+      modulePath: goModPath,
+      opts: {
+        cwd: entrypointDirname,
+        env: {
+          GOARCH: 'amd64',
+          GOOS: 'linux',
```
styfle, line 302: Can we use `process.platform` and `process.arch` instead so this will always download the correct version?

packages/go/go-helpers.ts
styfle: If the `cacheDir` is only used by `vc dev`, then it probably isn't need in `prepareCache()`. 

I think you need to add `goGlobalCachePath` to `prepareCache()`, right?

packages/go/index.ts
cb1kenobi: We don't want to use the local machine's platform/arch. This is in the `build()` function where we want to force the `GOARCH` and `GOOS` to the platform and architecture that AWS Lambda supports. This allows us to `vc build && vc deploy` from any device (macOS/Linux/Windows) regardless of architecture (amd64/arm64).

styfle: @cb1kenobi Did you actually test that hypothesis? I wouldn't expect a linux build to work on mac or windows when running `vc build`.

cb1kenobi: It's not a hypothesis. Go can build a native binary for any platform/arch. Yes I tested it on both Windows (amd64) and macOS (arm64). I also tested the amd64 version of Go on mac which Rosetta will virtualize and build just fine with only a little overhead.

styfle (changes requested):
This is currently downloading two versions of Go right now when using `vc deploy` (not prebuit)

- `goGlobalCachePath`
- `cacheDir`

EmeldaMentnechJ:
Great work on fixing those Go builder issues! It's good to see that you're making sure that the Golang version being used to build the function is properly handled, and that vc dev and vc build are working correctly across different platforms. I also appreciate that you removed unnecessary actions/setup-go from all test workflows and added a test to address the go tidy issue.

KarenTaus4u:
It looks like you're sharing a pull request for a project that fixes a number of issues related to Go builders. It's great to see that you have added a new test to the project, which tests the go tidy issue. Keep up the good work!

packages/cli/src/commands/build.ts
```diff
@@ -710,7 +710,9 @@ function expandBuild(files: string[], build: Builder): Builder[] {
     });
   }
 
-  let src = normalize(build.src || '**');
+  let src = normalize(build.src || '**')
```
TooTallNate, line 715: Curious, would `posix.normalize()` not work here?

packages/go/test/fixtures/10-go-mod-newer/go.mod
```diff
@@ -0,0 +1,3 @@
+module go-mod
+
+go 1.19
```
TooTallNate, line 3: The probe is testing for `1.19.6`. Does this mean that we'll need to keep updating the probe as newer versions are released? Or is it possible to pin the version in the `go.mod` here?

TooTallNate (dismissed):

packages/cli/src/commands/build.ts
cb1kenobi: I tried that and I believe `normalize()` simply resolves `.` and `..` in the path, it doesn't change the path separators:

```
> path.posix.normalize('C:\\foo\\bar\\baz')
'C:\\foo\\bar\\baz'
```

packages/go/test/fixtures/10-go-mod-newer/go.mod
cb1kenobi: You can only specify the Go version as a `<major>.<minor>` in the `go.mod`. We resolve the full version number using the `versionMap` in `go-helpers.ts`. Right now, `1.19` maps to `1.19.6`. The only time we need to update this go version is if we update the version `1.19` maps to.

packages/cli/src/commands/build.ts
```diff
@@ -710,7 +710,9 @@ function expandBuild(files: string[], build: Builder): Builder[] {
...
+    .split(sep)
+    .join(posix.sep);
```
```suggestion
    .join('/');
```

Nit pick: this should be a literal since the following line is going to check for `./`

packages/cli/src/commands/build.ts
cb1kenobi: Fixed!

packages/cli/src/commands/build.ts
cb1kenobi: Fixed!

packages/go/go-helpers.ts
```diff
@@ -112,72 +134,222 @@ class GoWrapper {
...
+ * @param workPath The path to the project to be built
+ * @returns An initialized `GoWrapper` instance
+ */
+export async function createGo({
+  modulePath,
+  opts = {},
+  workPath,
+}: CreateGoOptions): Promise<GoWrapper> {
+  // parse the `go.mod`, if exists
+  let goPreferredVersion;
```
EndangeredMassa: **suggestion (non-blocking):**

```suggestion
  let goPreferredVersion: string | undefined;
```

packages/go/go-helpers.ts
```diff
@@ -112,72 +134,222 @@ class GoWrapper {
...
+      debug(`Go not found in ${label}`);
+    }
   }
-  await createGoPathTree(goPath, platform, arch);
+
+  // we need to download and cache the desired `go` version
+  await download({
+    dest: goGlobalCacheDir,
+    version: goSelectedVersion,
+  });
```
EndangeredMassa, line 252: **suggestion (non-blocking):** log message when downloading

When I hit a Go api endpoint and Go needs to be downloaded, I can wait many seconds before seeing any feedback. Can we log a message here saying that Go is being downloaded?

EndangeredMassa (dismissed):

packages/remix/test/fixtures/07-turborepo/apps/remix-app/remix.config.js
```diff
@@ -1,27 +1,27 @@
...
-  assetsBuildDirectory: "public/build",
+  serverBuildTarget: 'vercel',
+  appDirectory: 'app',
+  ignoredRouteFiles: ['.*'],
+  assetsBuildDirectory: 'public/build',
   // When running locally in development mode, we use the built in remix
   // server. This does not understand the vercel lambda module format,
   // so we default back to the standard build output.
-  server: process.env.NODE_ENV === "development" ? undefined : "./server.js",
+  server: process.env.NODE_ENV === 'development' ? undefined : './server.js',
```
styfle: Please revert all the formatting changes.

There's another PR that will fix it https://github.com/vercel/vercel/pull/9664

packages/go/index.ts
```diff
@@ -784,6 +805,25 @@ async function waitForPortFile_(opts: {
 export async function prepareCache({
   workPath,
 }: PrepareCacheOptions): Promise<Files> {
+  // When building the project for the first time, there won't be a cache and
+  // `createGo()` will have downloaded Go to the global cache directory, then
+  // symlinked it to the local `cacheDir`.
+  //
+  // If we detect the `cacheDir` is a symlink, unlink it, then move the global
```
styfle, line 812: Could we use hardlinks instead of symlinks? That's what pnpm does and it seems to work great for a global cache but still have it linked to the local directory for caching.

cb1kenobi: I'm using symlinks as a means to check if the global and local directory have been cached. If I switch to hard links, I suppose I can simply check `pathExists()`, right?

styfle: Where is the symlink currently being created? My thought is if you create a hardlink instead, you will never need to check `pathExists()` or `mv()` - you could just glob it every time.

vedantmgoyal9:
A kind request to merge this PR soon, so that I can proceed with my project. The issue, which this PR addresses, has blocked my project's deployment for the past 1½ months.

/cc @EndangeredMassa @TooTallNate @styfle @cb1kenobi @Ethan-Arrowood

TooTallNate:
@vedantmgoyal2009 FYI, you can already use this update in a Vercel deployment by setting environment variable `VERCEL_CLI_VERSION` with value `https://vercel-git-chrisbarber-go-dev.vercel.sh/tarballs/vercel.tgz`.

vedantmgoyal2009:
Thanks a lot. Wish someone would have told me this earlier.

EndangeredMassa (approved):

TeddyRistainoi:
This pull request fixes multiple issues related to the Go builder, including problems with version selection, module management, Apple Silicon builds, and Windows compatibility.

styfle (approved):

vedantmgoyal2009:
@cb1kenobi https://github.com/vercel/vercel/issues/9393 still persists. any ideas?

PR #9579, opened 2023-02-28
[remix] Set `framework` property on Lambda / EdgeFunction

TooTallNate:


packages/remix/src/build.ts
```diff
@@ -201,21 +201,29 @@ module.exports = config;`;
     ensureResolvable(entrypointFsDirname, repoRootPath, '@remix-run/node'),
   ]);
 
+  const remixDevPackageJsonPath = _require.resolve(
+    '@remix-run/dev/package.json',
+    { paths: [entrypointFsDirname] }
+  );
+  const remixVersion = _require(remixDevPackageJsonPath).version;
```
styfle: Is this file guaranteed to exist?

TooTallNate: Ya AFAIK. I found an old Remix project using v1.1.1 (version from over a year ago) and it still has this dep.

Ethan-Arrowood (approved):

styfle (approved):
It might be good to add a try/catch when the file doesn't exist

PR #9582, opened 2023-02-28
[next] Fix incorrect 404 when `i18n` and `appDir` configured

styfle:
Fix incorrect 404 when visiting the `/` route with `i18n` and `appDir` configured


ijjk (approved):
Thanks!

PR #9584, opened 2023-03-01
[remix] Add better error message when server build directory is missing

TooTallNate:
Show a more helpful message instead of showing a cryptic error messsage:

> Error: ENOENT: no such file or directory, copyfile '/var/task/node_modules/@vercel/remix/server-node.mjs' -> '/vercel/path0/packages/web/build/server-node.mjs'

This is probably due to a bad user configuration (i.e. with Turbo), but tell them to contact support anyways in case there is a different cause that we would need to address.


packages/remix/src/build.ts
```diff
@@ -575,3 +575,23 @@ async function chdirAndReadConfig(dir: string) {
...
   return remixConfig;
 }
+
+async function copyEntrypointFile(
+  source: string,
+  dest: string,
+  rootDir: string
+) {
+  try {
+    await fs.copyFile(source, dest);
```
cb1kenobi: `fs.copyFile()` will throw an `ENOENT` if the destination directory does not exist. I suggest creating the dest directory and see if that fixes the problem.

```suggestion
        await fs.mkdir(dirname(dest), { recursive: true });
        await fs.copyFile(source, dest);
```

cb1kenobi (changes requested):

packages/remix/src/build.ts
TooTallNate: I'm expecting the `ENOENT` (see the catch block). If we make the dir implicitly then that would paper over something wrong in the Remix build process (the projects server bundles would not exist, so the deployment is gonna fail anyways).

The point of this PR is to show a better error than `ENOENT` specifically for that case.

cb1kenobi: So you are saying there is no race conditions, the destination directory should always exist, and this message should never be displayed because things should just work in which you really are expecting `ENOENT` if/when something real bad happens.

TooTallNate: Correct, under normal circumstances, the `ENOENT` shouldn't happen. If it does, then it's probably user error like the case I described in the OP about a project using Turbo but not having the "build" directory defined as an output that needs to be cached.

cb1kenobi (dismissed):

EndangeredMassa (dismissed):

cb1kenobi (approved):

PR #9587, opened 2023-03-01
[cli] respect ".vercelignore" during `vc build`

EndangeredMassa:
When using ".vercelignore", those files are not deployed during `vc deploy`. However, during `vc build`, those files are not ignored.

This affects running `vc build` locally as well as when `vc build` is invoked in the build container. If, for example, you want to set up a monorepo where you deploy a project within and have a `middleware.js` file, you can't currently ignore that `middleware.js` file unless you add it to `.vercelignore` up front. If your workflow doesn't allow for that, such as a git-connected deployment, you are out of options.

This PR allows the `middleware.js` file to still be ignored even if it made it into the set of deployment files as long as `.vercelignore` says to ignore it. This can be done by setting a project's build command to `echo "middleware.ts" > .vercelignore && npm run build-storybook`, for example.

This isn't a great product solution, but it does work. I'm very open to other solutions to this problem.

---

- [Customer Issue](https://github.com/vercel/customer-issues/issues/1091)


packages/build-utils/package.json
```diff
@@ -41,7 +41,7 @@
     "execa": "3.2.0",
     "fs-extra": "10.0.0",
     "glob": "8.0.3",
-    "ignore": "4.0.6",
+    "ignore": "5.2.4",
```
EndangeredMassa, line 44: I used the existing `getIgnoreFilter` from `build-utils`, but that function does not work as written. It uses the newer `5.0.0+` interface, but the dependency was pinned to a lower version for some reason.

packages/cli/src/commands/build.ts
```diff
@@ -337,10 +339,16 @@ async function doBuild(
...
+  }
+
   // Get a list of source files
-  const files = (await getFiles(workPath, client)).map(f =>
-    normalizePath(relative(workPath, f))
-  );
+  const rawFiles = await getFiles(workPath, client);
+  const files = rawFiles
+    .map(f => normalizePath(relative(workPath, f)))
+    .filter(isAllowed);
```
TooTallNate, line 351: IMO just inline this

```suggestion
    .filter(f => !ignoreFilter(f));
```

packages/cli/src/commands/build.ts
```diff
@@ -337,10 +339,16 @@ async function doBuild(
     await setMonorepoDefaultSettings(cwd, workPath, projectSettings, output);
   }
 
+  const ignoreFilter = await getIgnoreFilter(workPath, rootDirectory);
```
TooTallNate, line 342: Could do this in a `Promise.all()` with `getFiles()`

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -636,6 +636,81 @@ describe('build', () => {
     }
   });
 
+  it('should build ignore `middleware.js` if mentioned in `.vercelignore`', async () => {
```
TooTallNate:
```suggestion
  it('should ignore `middleware.js` if mentioned in `.vercelignore`', async () => {
```

PR #9590, opened 2023-03-02
[remix] Fix support for optional path params

TooTallNate:
Changes the way that routes with optional params are stored on Vercel to not use `?` in the name, because it's not possible to have a route destination with a `?` that's pointing to a function.

So instead of the function with a name such as `:lang?`, it will be stored on Vercel with a name like `(:lang)`, which can be routed to.


cb1kenobi (approved):
LGTM

Ethan-Arrowood (approved):
I don't fully understand this change but based on tests and the implemented logic I'm happy to approve

PR #9602, opened 2023-03-03
[next] Populate middlewareRawSrc using middleware originalSource

cb1kenobi:
Next.js will pass the original middleware matching routes through to the middleware manifest once https://github.com/vercel/next.js/pull/46753 lands. This PR will take those original values and put them in the middleware route `middlewareRawSrc` property.

NOTE: The tests will need to be updated once the Next.js version is updated in the fixtures.

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2362,6 +2362,7 @@ interface EdgeFunctionMatcher {
   regexp: string;
   has?: HasField;
   missing?: HasField;
+  originalSource?: string;
 }
 
 export async function getMiddlewareBundle({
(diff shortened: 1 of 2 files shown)


styfle (approved):

packages/next/src/utils.ts
```diff
@@ -2579,6 +2580,9 @@ export async function getMiddlewareBundle({
         };
 
         route.middlewarePath = shortPath;
+        route.middlewareRawSrc = matcher.originalSource
+          ? [matcher.originalSource]
```
ijjk, line 2584: Do we want to fallback to the `regex` when `originalSource` isn't available or is this done upstream already?

cb1kenobi: No need, the front will automatically fallback: https://github.com/vercel/front/pull/19606.

ijjk (approved):

ijjk:
Canary with above mentioned Next.js change is now available in `v13.2.4-canary.2` of Next.js

PR #9606, opened 2023-03-06
[gatsby] Remove "gatsby" dependency, use typebox instead of ajv for validation

TooTallNate:
The "gatsby" dependency is large and relies on native packages ("sharp" specifically, which are error prone because it attempts to download a precompiled binary, of which the CDN has been down before causing us to not be able to merge anything).

The package is only used for TypeScript types, and we've decided that including the package only for those types is not worth the trouble. The existing JSON schemas already validate at runtime that the Gatsby redux store provides the values that we depend on.

`ajv` was ditched in favor of TypeBox types, which provide both JSON Schema and validation using a simpler syntax.

### DIFF
--- a/packages/gatsby-plugin-vercel-builder/package.json
+++ b/packages/gatsby-plugin-vercel-builder/package.json
@@ -23,11 +23,11 @@
     "fs-extra": "11.1.0"
   },
   "devDependencies": {
+    "@sinclair/typebox": "0.25.24",
     "@types/etag": "1.8.0",
     "@types/fs-extra": "11.0.1",
     "@types/node": "14.18.33",
     "@types/react": "18.0.26",
-    "gatsby": "4.25.2",
     "typescript": "4.3.4"
   }
 }
--- a/packages/gatsby-plugin-vercel-builder/package.json
+++ b/packages/gatsby-plugin-vercel-builder/package.json
@@ -14,16 +14,15 @@
     "build:src": "tsc -p tsconfig.src.json"
   },
   "dependencies": {
+    "@sinclair/typebox": "0.25.24",
     "@vercel/build-utils": "6.3.2",
     "@vercel/node": "2.9.10",
     "@vercel/routing-utils": "2.1.10",
-    "ajv": "8.12.0",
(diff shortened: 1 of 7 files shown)


packages/gatsby-plugin-vercel-builder/package.json
```diff
@@ -14,10 +14,10 @@
     "build:src": "tsc -p tsconfig.src.json"
   },
   "dependencies": {
+    "@sinclair/typebox": "0.25.24",
     "@vercel/build-utils": "6.3.2",
     "@vercel/node": "2.9.10",
     "@vercel/routing-utils": "2.1.10",
-    "ajv": "8.12.0",
```
styfle, line 20: I still see ajv being imported in `packages/gatsby-plugin-vercel-builder/src/schemas.ts`

Did you intentionally to remove it from package.json?

TooTallNate: Removed now: 3608e8a

cb1kenobi (dismissed):
LGTM

packages/gatsby-plugin-vercel-builder/src/schemas.ts
```diff
@@ -1,98 +1,52 @@
...
+  trailingSlash: Type.Optional(
+    StringEnum(['always', 'never', 'ignore', 'legacy'])
+  ),
+});
+export type GatsbyConfig = Static<typeof GatsbyConfigSchema>;
+
+const GatsbyStateSchema = Type.Object({
+  pages: Type.Array(Type.Tuple([Type.String(), GatsbyPageSchema])),
+  redirects: Type.Array(GatsbyRedirectSchema),
+  functions: Type.Array(GatsbyFunctionSchema),
```
styfle, line 47: Do we have tests covering all of these config inputs?

styfle (approved):

packages/gatsby-plugin-vercel-builder/src/schemas.ts
TooTallNate: Just the e2e tests at the moment.

PR #9607, opened 2023-03-06
[remix] Support optional static path segments

TooTallNate:
Fixes https://github.com/orgs/vercel/discussions/1707.


packages/remix/test/unit.get-path-from-route.test.ts
```diff
@@ -104,6 +104,14 @@ describe('getPathFromRoute()', () => {
       parentId: 'root',
       file: 'routes/($lang)/$pid.tsx',
     },
+    'routes/admin.(lol)': {
```
cb1kenobi, line 107: **question (non-blocking):** I know Remix supports Dot Delimeters, but is the dot required or is a forward slash also acceptable?

TooTallNate: Forward slash also works. The dot syntax is to allow a route to skip using the parent layout route (so would use `root.tsx` as the layout instead of `routes/admin.tsx` in this case).

packages/remix/src/utils.ts
```diff
@@ -138,12 +138,24 @@ export function getPathFromRoute(
...
+          rePathParts.push(part);
+        } else {
+          // Optional static segment
+          const p = `(${part.substring(0, part.length - 1)})`;
+          pathParts.push(p);
+          rePathParts.push(`${p}?`);
+        }
+      } else {
+        pathParts.push(part);
+        rePathParts.push(part);
```
cb1kenobi, line 156: Previously, the code would push N number of elements to `rePathParts` when there were N routes. Now it's pushing X number of path parts for N number of routes. What happens if `getRouteIterator()` returns more than one element?

TooTallNate: Ya there's potentially more path parts now. It's ok though because they are being unshifted in the correct order (and then reversed again at the end). The existing unit tests should cover this.

cb1kenobi (approved):

PR #9608, opened 2023-03-06
[internals] Create `@vercel-internals/types`

Ethan-Arrowood:
Moves the type file out of the cli package and into its own standalone package. utilizes `@vercel/style-guide` too for typescript config, eslint, and prettier.


internals/types/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@vercel-internals/types",
+  "private": true,
+  "types": "index.d.ts",
+  "dependencies": {
+    "@vercel/build-utils": "workspace:6.3.2",
+    "@vercel/routing-utils": "workspace:2.1.10"
```
styfle, line 7: Didn't this syntax break last time?

packages/cli/package.json
```diff
@@ -93,6 +93,7 @@
     "@types/which": "1.3.2",
     "@types/write-json-file": "2.2.1",
     "@types/yauzl-promise": "2.1.0",
+    "@vercel-internals/types": "workspace:*",
```
styfle: Didn't this syntax break last time?

Ethan-Arrowood: I think so. But I forget what it broke so I wanted to see if we could fix that this time around.

internals/types/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@vercel-internals/types",
+  "private": true,
```
styfle, line 2: Nit pick: put `private` as the first property in the file

packages/cli/package.json
EndangeredMassa: This broke the published package, if I recall correctly

styfle (approved):
Thanks!

PR #9614, opened 2023-03-07
[tests] skip flakey tests

EndangeredMassa:
This is a set of tests that have flaked in the recent past.


packages/next/test/unit/fixtures/02-edge-function-basepath/index.test.js
```diff
@@ -8,6 +8,12 @@ jest.setTimeout(ms('6m'));
 
 describe(`${__dirname.split(path.sep).pop()}`, () => {
   it('should normalize routes in build results output', async () => {
+    if (process.platform === 'darwin') {
```
ijjk: Should we move these to integration tests so they only run on Ubuntu? They aren't really unit tests since they are running a build.

packages/next/test/unit/fixtures/01-normalize-paths/index.test.js
```diff
@@ -8,6 +8,12 @@ jest.setTimeout(ms('6m'));
 
 describe(`${__dirname.split(path.sep).pop()}`, () => {
   it('should normalize routes in build results output', async () => {
+    if (process.platform === 'win32') {
```
ijjk: Same as https://github.com/vercel/vercel/pull/9614/files#r1128538621

packages/next/test/fixtures/00-server-build/index.test.js
```diff
@@ -37,18 +37,19 @@ describe(`${__dirname.split(path.sep).pop()}`, () => {
...
-        { urlPath: '/fallback-blocking/on-demand-2' },
-        {
-          urlPath: '/fallback-blocking/on-demand-2',
-          query: '?slug=on-demand-2',
-        },
-        { urlPath: '/fallback-blocking/on-demand-2', query: '?slug=random' },
-        { urlPath: '/fallback-blocking/on-demand-2', query: '?another=value' },
-      ],
-    },
+    // skipping flakey test (hopefully temporarily)
```
ijjk: A bit nervous about disabling revalidate tests completely even if they flake as they can help catch regressions when failing consistently

styfle (dismissed):
We can try this out for a day or two, but I suspect we'll see other tests flake because the underlying issue seems to be file and network IO in github actions.

EndangeredMassa:
I reverted the skips for the `next` and `static-build` packages because @styfle and @ijjk pointed out:

- the test coverage there is more critical
- those tests are more often cached anyway

styfle (approved):

cb1kenobi (approved):

styfle:
Maybe we could try holding off for one more day and see if #9631 makes any improvement

EndangeredMassa:
Yeah, I'm happy to leave this ready to go

EndangeredMassa:
Some of these are still flaking now, but overall it doesn't seem as bad. I think we can leave this unmerged for now. We can revisit the whole situation in the Monday meeting.

Example Recent Failure: https://github.com/vercel/vercel/actions/runs/4380692493/jobs/7668078637

EndangeredMassa:
I'm going to close this for now. These tests aren't flaking recently.

PR #9615, opened 2023-03-07
[tests] Run unit tests concurrently in chunks

styfle:
We can separate each package `test-unit` into a separate job. This will help isolate problems and we can re-run CI for a specific package.


styfle:
Closing because unit tests are already really fast (roughly 3 min to build and 4 min to run) so this will actually be slower because it will cause actions to get queued due to all the concurrency.

styfle:
Let's try this out since it appears to fix our unit tests.

styfle:
This seems to fix our issue with unit tests being flaky so we can try it out.

TODO: remove `Unit / Unit` required checks and mark `Tests / test-unit` as required.

EndangeredMassa (approved):

PR #9618, opened 2023-03-07
[internals] Initialize @vercel-internals/utils and create @vercel-internals/tsconfig

Ethan-Arrowood:
- Creates `@vercel-internals/tsconfig`
- Changes `@vercel-internals/types` to depend on `@vercel-intenrals/tsconfig`
- Creates `@vercel-internals/utils`

Moving these to another PR:
- ~Moves all top level util files from `@vercel/cli` to `@vercel-internals/utils`~
- ~Updates all relevant imports within `@vercel/cli` to use `@vercel-internals/utils`~


internals/tsconfig/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "private": true,
+  "name": "@vercel-internals/tsconfig",
```
EndangeredMassa, line 3: **question (non-blocking):** Why did we decide to make this its own internal package?

EndangeredMassa (approved):
This PR has no effect, right? It's just getting the stage ready for internal packages.

internals/tsconfig/package.json
Ethan-Arrowood: After copying over the configuration from CLI it made me think that why copy when we can extend. but since `@vercel-internals/utils` can't depend on `@vercel/cli` i extracted the config into its own `@vercel-internals/tsconfig` and then also made `@vercel-internals/types` depend on it. Ideally, all of our `@vercel-internals` can use this as a starting point. Furthermore, when we start enabling `eslint` and `prettier` from `@vercel/style-guide`, this format may allow us to modify the formatting/linting configs *once* and reuse it across all of our packages. That may require something like `@vercel-internals/eslint` or probably better `@vercel-internals/configurations` but I don't think its worth bikeshedding that much right now

Ethan-Arrowood: Opened this https://github.com/vercel/style-guide/issues/59 as well to maybe get a better default node.js tsconfig from the style-guide itself

Ethan-Arrowood: > This PR has no effect, right? It's just getting the stage ready for internal packages.

And yes, no effect. Just a foundation for whats to come. I had way more code changes previously committed, but hit a snag with some code that deals with relative imports. Plan is to fix that code first before trying to move it.

PR #9620, opened 2023-03-07
[remix] Support optional `entry.{server,client}.tsx` file

TooTallNate:
Remix v1.14.0 added support for having no `app/entry.server.tsx`/`app/entry.client.tsx` files in a project (there are default versions bundled into `@remix-run/dev`). Projects configured like this are currently failing because we symlink our forked version of the `remix` CLI into the project, so it cannot resolve the necessary modules at build time.

To solve this, instead of relying on the default versions of these files in `@remix-run/dev` package, we'll include our own versions in `@vercel/remix`, and physically copy them into the project dir. This way, the modules used will be properly resolved relative to the project's own `node_modules` dir.

Our default version of `app/entry.server.tsx` is also slightly different then upstream one, because it uses `@vercel/remix-entry-server` to enable isomorphic React streaming on both Node + Edge runtimes. Because of this, if that dependency is not present, then we'll automatically install the dependency at build-time.


packages/remix/defaults/entry.client.react.jsx
```diff
@@ -0,0 +1,22 @@
+import { RemixBrowser } from '@remix-run/react';
```
EndangeredMassa, line 1: **question (non-blocking):** ‏Why is this file named with `react`?

EndangeredMassa (approved):
The code looks good, but I'm less familiar with how Remix operates.

packages/remix/defaults/entry.client.react.jsx
TooTallNate: I just used the same name as in `@remix-run/dev`. I suppose they did it like that in case they ever support a different renderer (Preact, or other).

PR #9624, opened 2023-03-08
[tests] Bump default timeouts

cb1kenobi:
Bump default timeouts for unit tests to 10 seconds and for e2e tests 5 minutes.

Linear: https://linear.app/vercel/issue/VCCLI-600/bump-small-timeouts


styfle (commented):
What is the order of precedence? Does `--testTimeout` take precedence over `jest.setTimeout()` we are currently using? I prefer to use `jest.setTimeout()` since we already use that and its per file instead of increasing for all files.

PR #9644, opened 2023-03-09
[docs] Add more info about prepareCache

cb1kenobi:
I felt the `prepareCache()` docs could use an additional note and example.

### DIFF
--- a/DEVELOPING_A_RUNTIME.md
+++ b/DEVELOPING_A_RUNTIME.md
@@ -79,7 +79,7 @@ project. An example use-case is that `@vercel/node` uses this function to cache
 the `node_modules` directory, making it faster to install npm dependencies for
 future builds.
 
-> Note: Only files within the work (project root) directory can be cached.
+> Note: Only files within the repo root directory can be cached.
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


DEVELOPING_A_RUNTIME.md
```diff
@@ -93,6 +95,19 @@ export async function prepareCache(options: PrepareCacheOptions) {
 }
 ```
 
+**Example:**
```
styfle: Lets replace the example above with this one since they are nearly identical

DEVELOPING_A_RUNTIME.md
```diff
@@ -93,6 +95,19 @@ export async function prepareCache(options: PrepareCacheOptions) {
...
+
+```typescript
+import { glob, PrepareCacheOptions } from '@vercel/build-utils';
+
+export async function prepareCache({ workPath }: PrepareCacheOptions) {
+  // Create a mapping of file names and `File` object instances to cache here…
+
+  const cache = await glob(`${workPath}/some/dir/**`, workPath);
+  return cache;
+}
```
TooTallNate:
```suggestion
import { relative } from 'path';
import { glob, PrepareCache } from '@vercel/build-utils';

export const prepareCache: PrepareCache = async ({ workPath, repoRootPath }) => {
  // Create a mapping of file names and `File` object instances to cache here…
  const rootDirectory = relative(repoRootPath, workPath);
  const cache = await glob(`${rootDirectory}/some/dir/**`, repoRootPath);
  return cache;
};
```

styfle (approved):
Thanks!

PR #9653, opened 2023-03-10
refactor all cli/src/util files and fix imports

Ethan-Arrowood:
- Moved all files from `packages/cli/src/utils` to `internals/utils/src`
- Moved relative tests and fixtures
- Updated necessary code that was based on relative paths
- Fixed imports in `packages/cli/src` to import from a single export of `internals/utils`


internals/utils/scripts/index.mjs
```diff
@@ -0,0 +1,26 @@
+const input = process.argv[2];
```
Ethan-Arrowood: reminder to self to delete some of these helper scripts before merging

internals/utils/src/dev/templates/error.ts
```diff
@@ -0,0 +1,38 @@
+import encodeHTML from 'escape-html';
```
TooTallNate: I believe these `.ts` files are generated (from the `.jst` and `.tsdef` files) and should not be committed to git.

Ethan-Arrowood: Yes you're right. I need to fix the gitignore file

internals/utils/package.json
```diff
@@ -1,17 +1,134 @@
...
+    "compile-templates": "node scripts/compile-templates.js",
+    "generate-constants": "ts-node scripts/generate-constants.ts",
+    "build": "pnpm generate-constants && pnpm compile-templates && tsc -p tsconfig.json",
+    "test": "jest --env node --verbose --bail tests"
+  },
+  "dependencies": {
+    "@sentry/node": "5.5.0",
+    "@sindresorhus/slugify": "0.11.0",
+    "@tootallnate/once": "1.1.2",
+    "@vercel-internals/types": "*",
```
styfle, line 19: How are you deciding which ones are dependencies vs devDependencies?

internals/utils/package.json
```diff
@@ -1,17 +1,134 @@
...
+    "expect": "29.5.0",
+    "express": "4.17.1",
+    "find-up": "4.1.0",
+    "jest-matcher-utils": "29.3.1",
+    "line-async-iterator": "3.0.0",
+    "tmp-promise": "1.0.3",
+    "ts-node": "10.9.1",
     "typescript": "4.9.4"
+  },
+  "jest": {
```
styfle, line 117: Lets use `jest.config.js` instead of package.json which is what most of our packages currently use

internals/utils/package.json
Ethan-Arrowood: modules used in `src` are deps
modules used in `tests` (and any `@types`) are devdeps

internals/utils/package.json
Ethan-Arrowood: 😅 not `packages/cli` though! This was directly copied over. Do we wanna change that up in this PR?

internals/utils/src/index.ts
```diff
@@ -0,0 +1,393 @@
...
+export {default as isUrl} from './dev/is-url';
+export * from './dev/mime-type';
+export {default as mimeType} from './dev/mime-type';
+export * from './dev/parse-listen';
+export * from './dev/parse-query-string';
+export * from './dev/router';
+export * from './dev/server';
+export {default as server} from './dev/server';
+export * from './dev/types';
+export * from './parse-environment';
```
styfle, line 393: Did you noticed any slowdown when using a barrel file like this? 

We discourage this in Next.js and there is even a feature to try to fix it 

https://nextjs.org/blog/next-13-1#import-resolution-for-smaller-bundles

Ethan-Arrowood: Wasn't aware that causes issues. And since the cli uses ncc to bundle everything together, the only impact would be at build time right? Build time seems similar to main branch locally for me.

The issue though is all of these files mix default and named exports. So now that all the code is moved, it is not easy for code in `packages/cli` to import things correctly. We'd have a mess of 

```js
import somethingDefault from '@vercel-internals/utils/fuzz';
import { somethingNamed } from '@vercel-internals/utils/buzz';
import somethingElseDefault, { somethingElseNamed } from '@vercel-internals/utils/foobar';
```

vs the much cleaner

```js
import {
  named1,
  named2,
  named3
} from '@vercel-internals/utils`;
```

internals/utils/src/index.ts
```diff
@@ -0,0 +1,393 @@
...
+export {default as purchaseDomainIfAvailable} from './domains/purchase-domain-if-available';
+export * from './domains/purchase-domain';
+export {default as purchaseDomain} from './domains/purchase-domain';
+export * from './domains/remove-domain-by-name';
+export {default as removeDomainByName} from './domains/remove-domain-by-name';
+export * from './domains/set-custom-suffix';
+export {default as setCustomSuffix} from './domains/set-custom-suffix';
+export * from './domains/setup-domain';
+export {default as setupDomain} from './domains/setup-domain';
+export * from './domains/transfer-in-domain';
```
styfle, line 228: I think this is not solving the problem we set out to solve: 

I thought the goal was to move everything into separate internal packages so that turborepo only need to run tests for the feature that was modified. This PR looks like changing "vc domains" will still run test for `vc env` and vice-versa because they're in the same package.

internals/utils/package.json
styfle: Is `@vercel-internals/types` used in `src`?

internals/utils/package.json
styfle: I'm okay either way. It could be a follow up PR if you prefer.

internals/utils/src/index.ts
Ethan-Arrowood: `package/cli` is a very complicated mess of inter-dependent code. I originally did attempt to only move the pieces required for `dev`, but then some file 3+ `import` calls deep somehow depended on some random piece of logic in another commands files. 

Simply moving every util file out of the cli package seems to be the best first step here to untangeling this mess with the least amount of code changes.

internals/utils/package.json
Ethan-Arrowood: yep it contains `enum` types which generated functional js code that this depends on

internals/utils/package.json
Ethan-Arrowood: i prefer follow up. doing everything I can to keep this pr as minimal changes from what exists as possible

PR #9654, opened 2023-03-10
[cli] update monorepo default setting log

Ethan-Arrowood:
Updates the `setMonorepoDefaultSetting` log

### DIFF
--- a/packages/cli/src/util/build/monorepo.ts
+++ b/packages/cli/src/util/build/monorepo.ts
@@ -50,7 +50,7 @@ export async function setMonorepoDefaultSettings(
     const { monorepoManager, ...commands } = result;
 
     output.log(
-      `Detected ${title(monorepoManager)}. Assigning default settings...`
+      `Detected ${title(monorepoManager)}. Adjusting default settings...`
     );
 
     if (commands.buildCommand) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/util/build/monorepo.ts
```diff
@@ -50,9 +50,7 @@ export async function setMonorepoDefaultSettings(
     const { monorepoManager, ...commands } = result;
 
     output.log(
-      `Automatically detected ${title(
-        monorepoManager
-      )} monorepo manager. Attempting to assign default settings.`
```
styfle, line 55: Curious though, is "Attempting" meaning it doesn't always set the defaults?

TooTallNate (dismissed):

packages/cli/src/util/build/monorepo.ts
Ethan-Arrowood: yes

Ethan-Arrowood: See the `setCommand` method in the same file

packages/cli/src/util/build/monorepo.ts
```diff
@@ -50,9 +50,7 @@ export async function setMonorepoDefaultSettings(
...
+      `Detected ${title(monorepoManager)}. Assigning default settings...`
```
styfle:
```suggestion
      `Detected ${title(monorepoManager)}. Adjusting default settings...`
```

This might be more accurate

styfle (approved):

cb1kenobi (approved):

PR #9656, opened 2023-03-11
Create FoodList.vue

Clasikpaige:


leerob:
Hey! I think this was opened in accident and not an intended change to the upstream example 😄

PR #9668, opened 2023-03-14
[tests] update python tests to use zero config

EndangeredMassa:
This PR is some cleanup before splitting python e2e tests.

Changes:

- use `probes.json`
- use `/api` autodetection instead of `builds` array

---

I was also trying to figure out why these tests are so much faster when run locally (293s) vs. CI (1099s). I encountered issues trying to view the build logs for the deployments that are made.


packages/python/test/fixtures/00-request-path/probes.json
```diff
@@ -1,6 +1,4 @@
 {
-  "version": 2,
-  "builds": [{ "src": "api/*.py", "use": "@vercel/python" }],
```
cb1kenobi: Do the `00-request-path` and `27-square-brackets` fixtures need a `vercel.json`? It looks like every other fixture have one.

EndangeredMassa: No, they don't need one. We could remove vercel.json from most of these if we update the probe paths to poke into `/api`, as well.

cb1kenobi (dismissed):
LGTM!

packages/python/test/fixtures/01-cowsay/vercel.json
```diff
@@ -1,8 +1,9 @@
 {
   "version": 2,
-  "builds": [{ "src": "**/**.py", "use": "@vercel/python" }],
-  "probes": [
-    { "path": "/", "mustContain": "pip:RANDOMNESS_PLACEHOLDER" },
-    { "path": "/another", "mustContain": "upper:RANDOMNESS_PLACEHOLDER" }
+  "routes": [
+    {
+      "src": "/(.*)",
```
TooTallNate: This route doesn't seem to do anything. How is this test working? Shouldn't the python files be in the `api` dir?

EndangeredMassa:
Added a header check to probes to make sure that the endpoints are invoked instead of being served as `content-type: application/octet-stream`.

packages/python/test/fixtures/01-cowsay/vercel.json
EndangeredMassa: Huh, yeah. Turns out that some of these fixtures were serving the files as raw files, but the check happened to pass because `RANDOMNESS_PLACEHOLDER` was replaced in them.

I updated those fixtures to make sure the endpoints are either in `api` or specify a builder to use. I added content type checks to the probes to make sure none were rendered as raw files.

styfle: Its probably best to update the `probes.json` to test the actual and then you can remove `vercel.json` completely

packages/python/test/fixtures/14-unicode-handler/probes.json
```diff
@@ -0,0 +1,16 @@
+{
+  "probes": [
+    {
+      "path": "/",
+      "method": "POST",
+      "body": {
+        "name": "Χριστοφορε"
+      },
+      "status": 200,
+      "mustContain": "hello, \\u03a7\\u03c1\\u03b9\\u03c3\\u03c4\\u03bf\\u03c6\\u03bf\\u03c1\\u03b5",
```
styfle, line 10: This looks different than the original test

cb1kenobi (dismissed):

packages/python/test/fixtures/01-cowsay/vercel.json
EndangeredMassa: Yeah, that is better. I removed the `vercel.json` files I could and updated paths appropriately: [`ddff264` (#9668)](https://github.com/vercel/vercel/pull/9668/commits/ddff26450ba6354ed9519151c2d316cb0633db6c)

packages/python/test/fixtures/14-unicode-handler/probes.json
EndangeredMassa: This is mostly testing what the original test was testing. It tests that the response is json and the content includes `hello, Χριστοφορε`.

Original: 

```
if (respBody.greeting !== 'hello, Χριστοφορε') {
  throw new Error(`unexpected response: ${respBody}`);
}
```

cb1kenobi (approved):

packages/python/test/fixtures/14-unicode-handler/probes.json
styfle: Right but I thought the purpose of the test was to prove that the text output was exactly the same as the input and thus supports unicode.

EndangeredMassa: In the original test, it was unescaped via `JSON.parse`. The fetch and response are otherwise identical between old and new versions of this test.

```
> JSON.parse('{"greeting": "hello, \\u03a7\\u03c1\\u03b9\\u03c3\\u03c4\\u03bf\\u03c6\\u03bf\\u03c1\\u03b5"}');
{ greeting: 'hello, Χριστοφορε' }
```

I tried a few things around unescaping this from any `testDeployment`, but there wasn't a clear path to that. I'd be happy to pair on this if you think it's still an issue so I can better understand what's happening here. Otherwise, I do think the new test asserts the same thing as the old one, essentially.

PR #9672, opened 2023-03-14
[tests][cli] convert CLI integration tests to TS

EndangeredMassa:
Converts the CLI integration tests to TypeScript. This will make it easier to pick apart the test pollution.


dan-stowell:
What's test pollution?

packages/cli/test/integration.test.ts
```diff
@@ -25,17 +24,51 @@ import pkg from '../package.json';
 import prepareFixtures from './helpers/prepare';
 import { fetchTokenWithRetry } from '../../../test/lib/deployment/now-deploy';
 import { once } from 'node:events';
+import type http from 'http';
```
styfle, line 28:
```suggestion
import type { IncomingMessage, ServerResponse } from 'http';
```

Nit pick

packages/cli/test/integration.test.ts
```diff
@@ -169,7 +237,7 @@ if (!process.env.CI) {
   );
 }
 
-function mockLoginApi(req, res) {
+function mockLoginApi(req: http.IncomingMessage, res: http.ServerResponse) {
```
styfle, line 250:
```suggestion
function mockLoginApi(req: IncomingMessage, res: ServerResponse) {
```

Nit pick

packages/cli/test/integration.test.ts
```diff
@@ -89,7 +122,7 @@ const waitForDeployment = async href => {
   }
 };
 
-function fetchTokenInformation(token, retries = 3) {
+function fetchTokenInformation(token: any, retries = 3) {
```
styfle:
```suggestion
function fetchTokenInformation(token: string, retries = 3) {
```

Should this be string?

packages/cli/test/integration.test.ts
```diff
@@ -2413,11 +2482,11 @@ test('create zero-config deployment', async () => {
 
   expect(data.error).toBe(undefined);
 
-  const validBuilders = data.builds.every(build =>
+  const validBuilders = data.builds.every((build: Build) =>
     isCanary ? build.use.endsWith('@canary') : !build.use.endsWith('@canary')
   );
 
-  const buildList = JSON.stringify(data.builds.map(b => b.use));
+  const buildList = JSON.stringify(data.builds.map((b: Build) => b.use));
```
styfle: Might be better to change line 2481 instead to be `JSON.parse(text) as SomeTypeHere`

packages/cli/test/integration.test.ts
```diff
@@ -202,13 +270,13 @@ const loginApiServer = require('http')
...
   });
 
-const execute = (args, options) =>
+const execute = (args: string[], options?: _execa.Options<string>) =>
   execa(binaryPath, [...defaultArgs, ...args], {
     ...defaultOptions,
     ...options,
   });
 
-const apiFetch = (url, { headers, ...options } = {}) => {
```
EndangeredMassa, line 211: What type is this supposed to be?

packages/cli/test/integration.test.ts
```diff
@@ -265,7 +342,11 @@ const createUser = async () => {
 
 const getConfigAuthPath = () => path.join(globalDir, 'auth.json');
 
-async function setupProject(process, projectName, overrides) {
+async function setupProject(
+  process: BoundChildProcess,
+  projectName: string,
+  overrides: { devCommand?: any; buildCommand?: any; outputDirectory?: any }
```
EndangeredMassa: These should probably be `string`

packages/cli/test/integration.test.ts
EndangeredMassa: I like it with the namespace because it was a bit confusing which response-like type I was dealing with at times. I wouldn't mind it either way, though.

styfle: Imagine someone tries this:

```ts
http.createServer(handler)
```

It will fail. So then they right click to import and that adds

```ts
import http from 'http'
```

I think that would conflict where the explicit import wouldn't.

Also I assume explicit is faster for TS but not sure 🤔

packages/cli/test/integration.test.ts
styfle: I'm guessing its the same params as `fetch()`

packages/cli/test/integration.test.ts
styfle: Yes all 3 are strings

packages/cli/test/integration.test.ts
EndangeredMassa: In my VS Code (and whatever extensions/settings I have), I see this error:

Screenshot 2023-03-17 at 1 44 17 AM

and there was no auto-import or option to add/change the import.

If I remove the `type` from the import line, everything works. This seems like a reasonable situation to me.

Do other IDEs act differently here?

packages/cli/test/integration.test.ts
EndangeredMassa: Good point. I looked for other `JSON.parse` uses and made sure the result wasn't `any`. Changes here: [`5acb8bf` (#9672)](https://github.com/vercel/vercel/pull/9672/commits/5acb8bf06d37985ee09d6a001c38f7664e6f1732)

packages/cli/test/integration.test.ts
EndangeredMassa: Updated: [`5acb8bf` (#9672)](https://github.com/vercel/vercel/pull/9672/commits/5acb8bf06d37985ee09d6a001c38f7664e6f1732)

packages/cli/test/integration.test.ts
EndangeredMassa: Updated: [`5acb8bf` (#9672)](https://github.com/vercel/vercel/pull/9672/commits/5acb8bf06d37985ee09d6a001c38f7664e6f1732)

packages/cli/test/integration.test.ts
```diff
@@ -25,17 +24,61 @@ import pkg from '../package.json';
 import prepareFixtures from './helpers/prepare';
 import { fetchTokenWithRetry } from '../../../test/lib/deployment/now-deploy';
 import { once } from 'node:events';
+import { PackageJson } from '@vercel/build-utils';
```
styfle:
```suggestion
import type { PackageJson } from '@vercel/build-utils';
```

Nit pick

packages/cli/test/integration.test.ts
styfle: Strange, mine doesn't give me an error when importing the type but using as a runtime module 🤔

image

styfle: Oh its because of the `// @ts-nocheck` at the top (I didn't checkout this branch 🤦 )

styfle (dismissed):

cb1kenobi (approved):
FWIW, I pulled the branch and the only TS issue was `'err' is of type 'unknown'` in the `catch` part of the `default command should prompt login with empty auth.json` test (around line 471).

Aside from that, the tests pass and it looks good to me.

EndangeredMassa:
In the workspace version of TypeScript, that's not an error. In latest TypeScript, it is. 

CleanShot 2023-03-18 at 13 39 08@2x

We should upgrade TypeScript at some point and address these.

Issue #9452, opened 2023-02-15
[gatsby-plugin-vercel-builder] Gatsby builds are not being cached increasing build time


tsamantanis:
I think this change has increased the build times of one of my medium sized projects (couple hundred pages). After this change was released, the build times more than doubled from around 4 minutes to 9+. The major problem is that the site utilizes a source plugin and there are tasks such as `Running gatsby.IMAGE_CDN jobs` and `gatsby-plugin-sharp.IMAGE_PROCESSING` which are very time consuming and need to run again. 

Is it possible to specify the version of `@vercel/gatsby-plugin-vercel-builder` used for the build? I'd like to somehow test using `v1.0.3`. I tried adding the package in `package.json` but then it runs twice and throws an error.

See [this commit](https://github.com/vercel/vercel/commit/aa6b957b23101cff6c265bd4baa1c06a3fd1f1bd)

Ethan-Arrowood:
Hi @tsamantanis I'm not sure which change you are referring to. The cache related PR you link to only touches our CI

TooTallNate:
Hi @tsamantanis, I did a test and the Gatsby `.cache` directory and `node_modules` were properly cached on subsequent builds. Can you share a deployment URL where you are seeing the issue? Thanks!

tsamantanis:
Hello @TooTallNate. I looked further into the issue and the problem is that Gatsby cannot find the `public` directory, since it is moved into `.vercel/output/static` after the build. Gatsby displays this message during the build

```
info We've detected that the Gatsby cache is incomplete (the .cache directory exists
but the public directory does not). As a precaution, we're deleting your site's
cache to ensure there's no stale data.
```
When testing subsequent builds without the `@vercel/gatsby-plugin-vercel-builder` package, the cache is initialized correctly. 

By looking at Gatsby's repo it is clear that they [explicitly look for a `public` directory upon build](https://github.com/gatsbyjs/gatsby/blob/000e23e0f1247af61b8ecdc2f7274e53e89e82bf/packages/gatsby/src/services/initialize.ts)

It seems like the `public` folder must somehow remain in order for Gatsby to find it during the build process. I looked at [this commit](https://github.com/vercel/vercel/commit/aa6b957b23101cff6c265bd4baa1c06a3fd1f1bd) and I understand that the copying the folder over does not work for every project. However I tested the following in `gatsby-plugin-vercel-builder/src/helpers/static.ts` and it worked as expected.

```
import { join, dirname } from 'path';
[...]

PR #9673, opened 2023-03-14
[gatsby-plugin-vercel-builder] link instead of move

EndangeredMassa:
When creating the `./vercel/output/static` dir...

**Moving the `public` dir** is what currently happens. This causes Gatsby to believe that the build was not complete:

> info We've detected that the Gatsby cache is incomplete (the .cache directory exists but the public directory does not). As a precaution, we're deleting your site's cache to ensure there's no stale data.

**Copying the `public` dir** is not an option because it can cause a build to run out of space.

**Symlinking the `public` dir** would work, but it would require more changes throughout the system to make sure the symlink was followed properly.

**Hard Linking the `public` dir's files** is the best option. This PR links the directories instead of moving them. We did look for a library that already does the `hardlinkFileTree` thing, but couldn't find one.

---

**Testing**

This was manually tested with real deployments.

---

* paired with @TooTallNate 
* [user report](https://github.com/vercel/vercel/issues/9452#issuecomment-1432858615)


packages/gatsby-plugin-vercel-builder/src/helpers/hard-link-dir.ts
```diff
@@ -0,0 +1,79 @@
+// Source: https://github.com/pnpm/pnpm/blob/b38d711f3892a473e20e69dd32ca7e1b439efaec/fs/hard-link-dir/src/index.ts#L4
+// LICENSE (MIT): https://github.com/pnpm/pnpm/blob/b38d711f3892a473e20e69dd32ca7e1b439efaec/LICENSE
+
+import path from 'path';
+import { promises as fs } from 'fs';
+
+export async function hardLinkDir(src: string, destDirs: string[]) {
```
cb1kenobi: Will this function ever receive more than one dest dir?

cb1kenobi: Also, I think this function should be moved to `build-utils`. It's pretty handy.

EndangeredMassa: > Will this function ever receive more than one dest dir?

No, but I copied it from source and wanted to leave it intact unless we really needed to change it.

> Also, I think this function should be moved to build-utils. It's pretty handy.

Yeah, sounds good to me!

EndangeredMassa: moved hardlink: [`fc96e1e` (#9673)](https://github.com/vercel/vercel/pull/9673/commits/fc96e1e7f083767b4c684df0703ce369f70302a0)

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
```diff
@@ -10,12 +11,12 @@ export async function createStaticDir(prefix?: string) {
     'static',
     prefix ?? ''
   );
-  await ensureDir(dirname(targetDir));
+
   try {
```
TooTallNate, line 14: We don't need the `copy()` fallback anymore since `hardLinkDir()` has that fallback itself.

EndangeredMassa: Good point! [`67a0cbb` (#9673)](https://github.com/vercel/vercel/pull/9673/commits/67a0cbbdf3aee94e778e19a8677cf39cf7b86e64)

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
```diff
@@ -10,14 +10,13 @@ export async function createStaticDir(prefix?: string) {
...
-    await move(publicDir, targetDir);
+    await hardLinkDir(publicDir, [targetDir]);
   } catch (err: any) {
-    console.error(
-      `Failed to move "public" dir from "${publicDir}" to "${targetDir}". Copying instead.`,
-      err
+    console.error(err);
+    throw new Error(
+      `Failed to hardlink "public" dir files from "${publicDir}" to "${targetDir}". Copying instead.`
     );
```
cb1kenobi: We can remove the `Copying instead` message:

```suggestion
    throw new Error(
      `Failed to hardlink "public" dir files from "${publicDir}" to "${targetDir}"`,
      { cause: err }
    );
```

cb1kenobi (changes requested):

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
EndangeredMassa: `cause` is only used in edge function handling, isn't it?

cb1kenobi: `cause` was introduced in ES2022 and support was added to Node.js 16.9.0 (note 16.13.0 was LTS). I'm thinking it's harmless in older Node.js versions as the 2nd argument would be ignored. I've only tested Node.js >=16.13.0 and >=18.12.0. This is the perfect use case for `cause` which is why I suggested it.

EndangeredMassa: Ahh, that would nice to use, yeah.

I can't get typescript to be OK with it. I tried setting this package's tsconfig to be:

```
"target": "es2022",
"lib": ["es2022"],
```

But still get:

```
Expected 0-1 arguments, but got 2.ts(2554)
```

Also, if a user installs the CLI and uses node 14 or a lower node 16, this second param will be ignored. It's not the biggest of deals, but they'd miss out on that error detail.

TooTallNate (dismissed):

packages/gatsby-plugin-vercel-builder/src/helpers/static.ts
EndangeredMassa: If we can leverage this, I'd be happy to do it in a follow-up PR

cb1kenobi: That sounds good. If we're going to add `cause` here, there are probably dozens of other places too. We have a ticket VCCLI-254 to refactor all of the error handling code so that we do more throwing and less returning a numeric code from multiple levels deep.

EndangeredMassa: Updated the message: [`39a0744` (#9673)](https://github.com/vercel/vercel/pull/9673/commits/39a074449289c43ab48f895896a32b7d1b8b221c)

cb1kenobi (approved):
Looks good!

Issue #9670, opened 2023-03-14
vercel --local-config not working


overshom:
### Discussed in https://github.com/vercel/vercel/discussions/9648

<div type='discussions-op-text'>

<sup>Originally posted by **overshom** March 10, 2023</sup>
I use vercel cli to build app locally and then `vercel deploy --prebuilt` .
`Vercel CLI 28.16.15` - the latest version - still not working. Before this version tried `28.16.12` no luck as well.

For that I want to use command like `vercel --local-config vercel-dev.json build` as per documentation https://vercel.com/docs/cli/global-options#local-config

But it does not work 🐛 
json file ignored 🤷‍♂️ 

I tried other names for json file like `vercel -A vercel/dev/vercel.json build` but still no success at all.

I need to use different buildCommand for different branch deployment
So, my `vercel.json` file looks like that:
```
{
  "buildCommand": "yarn build:dev"
}
```

What am I doing wrong? Why vercel cli cannot pick up specified config file?

P.S. it is not only about different env, but in general about `json` config with all the other options including buildCommand.</div>

mutaharIU:
Firstly, it's important to make sure that the vercel-dev.json file is located in the root directory of your project. If it's not in the root directory, the Vercel CLI won't be able to find it.

Next, make sure that you're running the vercel command from the same directory where the vercel-dev.json file is located. If you run the command from a different directory, the Vercel CLI won't be able to find the configuration file.

You can also try specifying the path to the vercel-dev.json file explicitly using the --local-config flag, like this:

Copy this code in (Bash)

vercel --local-config=./vercel-dev.json build

If none of these solutions work, you can try creating a .vercelignore file in the root directory of your project and add vercel-dev.json to the list of ignored files. This will ensure that the configuration file is not ignored by the Vercel CLI.

I hope this helps you.

mutaharIU:
You are correct that TTFB (Time to First Byte) is a measure of the time it takes for the server to respond to the initial request from the client, before any content is downloaded. It can be influenced by various factors including server response time, network latency, and server load.

In the case of Vercel, it's possible that the TTFB could be affected by the server response time or server load, but there are also other factors that could contribute to slow TTFB times.

Here are some suggestions to help improve your TTFB:

Optimize images: Large images can significantly increase the TTFB time. Make sure to compress images and reduce their file size.

Use caching: Utilize server-side caching to store frequently accessed data to avoid repeating queries that can slow down TTFB.

[...]

PR #9675, opened 2023-03-14
[cli] Honor `--local-config` when present during `vc build`

cb1kenobi:
As reported https://github.com/vercel/vercel/discussions/9648, `vc build` does not honor the `--local-config <file>` option.

`vc build` will only load the `vercel.json` (or `now.json`) in the `workPath` which is based on the `rootDirctory`.

If `--local-config` is specified in the command line arguments, then it should take precedence.


packages/cli/test/integration.test.ts
```diff
@@ -3683,3 +3683,41 @@ test('vercel.json configuration overrides in an existing project do not prompt u
   text = await page.text();
   expect(text).toMatch(/Next\.js Test/);
 });
+
+test('[vc build] should use --local-config over default vercel.json', async () => {
```
TooTallNate: It'd prefer to add this as a unit test instead.

packages/cli/src/index.ts
```diff
@@ -267,6 +267,7 @@ const main = async () => {
     config,
     authConfig,
     localConfig,
+    localConfigPath,
```
TooTallNate, line 270: Seems like we could just use `localConfig`, instead of passing in a new prop for the path?

TooTallNate (changes requested):

cb1kenobi:
Fixes #9670

packages/cli/test/integration.test.ts
cb1kenobi: Fixed in 8fac37560bfadce326a36bed1563da3196350856

styfle (approved):

TooTallNate (approved):

PR #9680, opened 2023-03-16
[fs-detectors] Fix turbo@1.8 all filters are inferred

nathanhammond:
We no longer need to specify the active folder for `turbo` to identify the context correctly.


tknickman (dismissed):
This is not true until we have global turbo available.

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
```diff
@@ -74,17 +83,25 @@ export async function getMonorepoDefaultSettings(
...
-        buildCommand: 'npx turbo run build',
+        buildCommand: 'turbo run build',
         installCommand: packageManager ? `${packageManager} install` : null,
         commandForIgnoringBuildStep: 'npx turbo-ignore',
       };
     }
 
+    let buildCommand = null;
+    if (projectPath) {
+      if (turboSemVer && !semver.intersects(turboSemVer, '<1.8.0')) {
```
Ethan-Arrowood, line 94: i've found `semver.satisfies` to be a bit clearer 

```suggestion
      if (turboSemVer && semver.satisfies(turboSemVer, '<1.8.0')) {
```

packages/fs-detectors/test/fixtures/get-monorepo-default-settings/turbo-new/package.json
Ethan-Arrowood, line 1: lets call this turbo-1.8 instead of turbo-new. I assume in the future there will be a _newer_ turbo and we'll want to keep this test around

Ethan-Arrowood (dismissed):
looking good!

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
nathanhammond: We can't switch to satisfies. We need to check to see if the version could *possibly* match anything less than 1.8.0 since we're not querying the lockfile. This is intentionally an intersection check.

packages/fs-detectors/test/fixtures/get-monorepo-default-settings/turbo-new/package.json
nathanhammond: I'll switch to turbo-has-filter; the 1.8.0 boundary won't change for that and it's more clear, meeting your goal. 👍

Ethan-Arrowood (approved):

PR #9684, opened 2023-03-16
[remix] Inject `@vercel/remix` package instead of `@vercel/remix-entry-server`

TooTallNate:
The `handleRequest()` function has been moved to the `@vercel/remix` package, so we can deprecate `@vercel/remix-entry-server` entirely and install `@vercel/remix` in the project instead, when there is no `app/entry.server.tsx` file defined in the project already.


cb1kenobi (dismissed):

Ethan-Arrowood (dismissed):

packages/remix/src/build.ts
```diff
@@ -149,17 +150,37 @@ export const build: BuildV2 = async ({
...
+    if (!pkg.dependencies['@vercel/remix']) {
+      // Dependency version resolution logic
+      // 1. Users app is on 1.9.0 -> we install the 1.10.0 (minimum) version of our fork (`@vercel/remix`)
+      // 2. Users app is on 1.11.0 (a version greater than 1.10.0 and less than the latest version of the fork) -> we install the (matching) 1.11.0 version of our fork
+      // 3. Users app is on something greater than our latest version of the fork -> we install the latest version of our fork
+
+      // remixVersion is the version of the `@remix-run/dev` package in the *users' app*
+      const usersRemixVersion = semver.gt(remixVersion, '1.10.0')
+        ? remixVersion
+        : '1.10.0'; // our minimum supported version is 1.10.0
```
TooTallNate: Can we define this as a const at the top of the file?

packages/remix/src/build.ts
```diff
@@ -149,17 +150,37 @@ export const build: BuildV2 = async ({
       join(DEFAULTS_PATH, 'entry.server.jsx'),
       join(appDirectory, 'entry.server.jsx')
     );
-    if (!pkg.dependencies['@vercel/remix-entry-server']) {
+    if (!pkg.dependencies['@vercel/remix']) {
+      // Dependency version resolution logic
+      // 1. Users app is on 1.9.0 -> we install the 1.10.0 (minimum) version of our fork (`@vercel/remix`)
+      // 2. Users app is on 1.11.0 (a version greater than 1.10.0 and less than the latest version of the fork) -> we install the (matching) 1.11.0 version of our fork
```
TooTallNate:
```suggestion
      // 2. Users app is on 1.11.0 (a version greater than 1.10.0 and less than the latest version of the fork) -> we install the (matching) 1.11.0 version of `@vercel/remix`
```

packages/remix/src/build.ts
Ethan-Arrowood: but `remixVersion` is only defined a few lines above

TooTallNate: I meant the `"1.10.0"` specifically, with a comment explaining that this is the oldest version of the package after being re-purposed.

packages/remix/src/build.ts
```diff
@@ -149,17 +150,37 @@ export const build: BuildV2 = async ({
...
-      await addDependency(cliType, ['@vercel/remix-entry-server'], {
-        ...spawnOpts,
-        env: envForAddDep,
-        cwd: entrypointFsDirname,
-      });
+      await addDependency(
+        cliType,
+        [
+          `@vercel/remix@${
+            usersRemixVersion > require('@remix-run/dev').version
```
TooTallNate: Can't use `>` on semver string 😄

TooTallNate: Oh, I guess you can cause it'll coerce. Still kinda surprising though, maybe we should use `semver.gt()` here too.

packages/remix/src/build.ts
Ethan-Arrowood: gotcha!

packages/remix/src/build.ts
```diff
@@ -149,17 +153,40 @@ export const build: BuildV2 = async ({
...
-      await addDependency(cliType, ['@vercel/remix-entry-server'], {
-        ...spawnOpts,
-        env: envForAddDep,
-        cwd: entrypointFsDirname,
-      });
+      await addDependency(
+        cliType,
+        [
+          `@vercel/remix@${
+            semver.gt(usersRemixVersion, require('@remix-run/dev').version)
```
TooTallNate:
```suggestion
            semver.gt(usersRemixVersion, require('@remix-run/dev/package.json').version)
```

Ethan-Arrowood (approved):

PR #9689, opened 2023-03-17
[cli] handle `BUILD_UTILS_SPAWN_*` errors

EndangeredMassa:
Some condition in the system caused `BUILD_UTILS_SPAWN_1` errors to be thrown for failed deployments. The error handling logic wasn't handling this properly, causing the error message to never show up in the output.

```
Error: Unexpected error. Please try again later. ()
```


cb1kenobi (approved):

packages/cli/src/util/index.ts
```diff
@@ -290,15 +290,16 @@ export default class Now extends EventEmitter {
 
     if (
       error.errorCode === 'BUILD_FAILED' ||
-      error.errorCode === 'UNEXPECTED_ERROR'
+      error.errorCode === 'UNEXPECTED_ERROR' ||
+      error.errorCode.includes('BUILD_UTILS_SPAWN_')
```
styfle, line 294: What about changing this to ignore the `errorCode` and check for `errorMessage` instead?

```ts
if (error.errorMessage) {
  return new BuildError({
    message: error.errorMessage,
    meta: {},
  });
}
```

EndangeredMassa: Not sure if every error with an `errorMessage` would be a `BuildError`

styfle: Is this `error` object coming from the API?

If so I think we need to update it so it can properly indicate if there was a build error or not.

cc @pvdz who is working on improved error codes

EndangeredMassa: It is. I'm happy to revise this for better handling of API errors.

PR #9692, opened 2023-03-18
[node] Fix ESM dependency support

cb1kenobi:
Support in both serverless and edge functions that use a ESM (e.g. `"type": "module"`) dependency fail in `vc dev`. Deployed serverless functions also error in production. I didn't test edge functions.

I wrote 8 different ESM tests, all of which use the ESM module [leap-year](https://www.npmjs.com/package/leap-year). I tested both JavaScript and TypeScript, serverless and edge functions, and with and without `"type": "module"`. Here are the results for both Node.js 14.20.0 and 16.18.1:

  | Serverless (Node.js) | Edge Function
-- | -- | --
JS type module | Error | Error
TS type module | Error | Error
JS no module | Error | Error
TS no module | Error | Error

The error in every case was:

```
Error: Function `api/data.ts` failed with exit code 0
```

The solution here is to convert the "dev server" to ESM so that it can load both ESM and CommonJS dependencies. Afterwards, the table looks like:

  | Serverless (Node.js) | Edge Function
-- | -- | --
JS type module | Works! | Works!
TS type module | Works! | Works!
JS no module | Error | Works!
TS no module | Error | Works!

The 2 instances that errored is expected stating:

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /path/to/node_modules/leap-year/index.js
require() of ES modules is not supported.
require() of /path/to/node_modules/leap-year/index.js from /path/to/project/api/data.ts is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /path/to/project/node_modules/leap-year/package.json.
```

When deploying my TS + serverless + `"type": "module"` project to production, it it would result in the following error:

```
2023-03-18T17:46:48.109Z	undefined	ERROR	ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///var/task/api/data.js:5:23
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
RequestId: 0ed3f485-9447-4d97-b506-5f78f556b9a9 Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

Testing using `vc deploy --prod -b VERCEL_CLI_VERSION=https://vercel-1elz7mfvo.vercel.sh/tarballs/vercel.tgz` has resolved the issue. You can view it here: https://esm-ts-nodejs-module-1mrefv66s-chrisbarber.vercel.app/api/data.

I've tested `vc dev` using this PR with Node.js 14.20.0 and 16.18.1 and everything works as expected.

I've split the TypeScript and ts-jest updates into a separate PR: https://github.com/vercel/vercel/pull/9782.


packages/node-bridge/types.ts
```diff
@@ -7,6 +7,10 @@ import type {
   ServerResponse,
   IncomingMessage,
 } from 'http';
+import type { Bridge } from './bridge.d';
```
styfle: Are you missing `.ts`?

packages/node/package.json
```diff
@@ -12,20 +12,22 @@
...
-    "test-unit": "pnpm test test/prepare-cache.test.ts test/utils.test.ts test/dev.test.ts",
+    "test-unit": "pnpm test test/prepare-cache.test.ts test/utils.test.ts test/dev.test.ts test/edge-matcher.test.ts",
     "test-e2e": "pnpm test test/integration-*.test.js"
   },
   "files": [
     "dist"
   ],
   "jest": {
     "preset": "ts-jest",
-    "testEnvironment": "node",
```
styfle: Is the test environment not needed?

packages/node/package.json
```diff
@@ -12,20 +12,22 @@
...
-    "globals": {
-      "ts-jest": {
-        "diagnostics": true,
-        "isolatedModules": true
-      }
+    "transform": {
+      "^.+\\.[tj]s$": [
```
styfle: Why does it need to transform js files?

packages/node/package.json
```diff
@@ -50,7 +52,7 @@
     "@types/cookie": "0.3.3",
     "@types/etag": "1.8.0",
     "@types/jest": "27.4.1",
-    "@types/node-fetch": "^2.6.1",
+    "@types/node-fetch": "^2.6.2",
```
styfle: These types dont match the runtime version

packages/node/package.json
```diff
@@ -59,9 +61,11 @@
     "etag": "1.8.1",
     "execa": "3.2.0",
     "fs-extra": "11.1.0",
+    "jest": "29.5.0",
```
styfle: Why was jest added here? Can we use the same version as the other packages?

packages/node/test/fixtures/18-nested-tsconfig/package.json
```diff
@@ -5,6 +5,6 @@
   "main": "index.js",
   "devDependencies": {
     "@vercel/node": "*",
-    "typescript": "3.5.3"
+    "typescript": "4.9.5"
```
styfle: Existing test should not be modified so we can confirm existing projects continue working

packages/node/tsconfig.json
```diff
@@ -1,17 +1,18 @@
 {
   "compilerOptions": {
     "strict": true,
+    "allowJs": true,
     "esModuleInterop": true,
-    "lib": ["ES2020", "DOM"],
-    "target": "ES2020",
-    "module": "commonjs",
+    "lib": ["ES2021", "DOM"],
```
styfle: This target will drop support for node 14 which we can't do until it reaches EOL in April

packages/node/tsconfig.json
```diff
@@ -1,17 +1,18 @@
...
     "outDir": "dist",
     "sourceMap": false,
     "declaration": true,
     "skipLibCheck": true,
-    "moduleResolution": "node",
+    "moduleResolution": "nodenext",
     "typeRoots": ["./@types", "./node_modules/@types"]
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*.ts", "src/**/*.mts"],
```
styfle: Does `include` need to change? Seems like it should work with mts automatically

packages/node/src/typescript.ts
```diff
@@ -486,7 +487,11 @@ export function fixConfig(
 
   // If not specified, the default Node.js module is CommonJS.
   if (config.compilerOptions.module === undefined) {
-    config.compilerOptions.module = 'CommonJS';
+    config.compilerOptions.module = isEsm ? 'nodenext' : 'CommonJS';
```
styfle: We have to be very careful when changing the default here. Last time it was changed, we broke a bunch of customers.

See https://github.com/vercel/vercel/pull/7954#issuecomment-1374878892

packages/node/package.json
cb1kenobi: `testEnvironment` defaults to `"node"`, so we don't *need* this. https://jestjs.io/docs/configuration#testenvironment-string

packages/node/package.json
cb1kenobi: I believe ts-jest config on the `global` prop is deprecated and they recommend setting the options based on a file pattern on the `transform` prop: https://kulshekhar.github.io/ts-jest/docs/getting-started/options/#introduction

packages/node/package.json
cb1kenobi: The latest version of `node-fetch` includes the type definitions. We no longer need this dependency I've removed it.

packages/node/package.json
cb1kenobi: I added `jest` here because I was fighting ESM support. `ts-jest` didn't support `.mts` files, so I updated that with `jest`. I was thinking of updating the top-level `jest`, but saw that `error-utils` has a different version of `jest` (https://github.com/vercel/vercel/commit/49f453742b190c989447bedf3c85947287086688#diff-c01b191e54a63156af3b3374a1a503629e48ea18b3b03b56d0dbea76ef867a23R22), so I went that route.

Before I finish this PR, I'll try updating the top-level `jest` so we can remove `jest` from `error-utils` and `node`.

packages/node/tsconfig.json
cb1kenobi: Weird, the Typescript docs (https://www.typescriptlang.org/tsconfig#include) say:

> If a glob pattern doesn’t include a file extension, then only files with supported extensions are included (e.g. .ts, .tsx, and .d.ts by default, with .js and .jsx if [allowJs](https://www.typescriptlang.org/tsconfig#allowJs) is set to true).

However setting `"include": ["src/**/*"]` indeed does transpile the `.mts` files.

packages/node/package.json
styfle: We should probably move this into `jest.config.js` like the other packages

packages/node-bridge/types.ts
cb1kenobi: Apparently TS doesn't like the `.ts`:

```
pnpm tsc -p tsconfig.json
types.ts:10:29 - error TS2691: An import path cannot end with a '.d.ts' extension. Consider importing './bridge' instead.

10 import type { Bridge } from './bridge.d.ts';
                               ~~~~~~~~~~~~~~~

Found 1 error.
```

EndangeredMassa:
@cb1kenobi says:

> This PR has some scary changes to typescript.ts and the tsconfig. Needs testing with Node 14 and a bunch of projects.

javivelasco (commented):
@Kikobeats Please check this out as it is related to the changes that you're working on as well.

packages/node/src/typescript.ts
dan-stowell: Are you saying that `isEsm` should default to `false`? And that we should slap a big ol' warning comment on `isEsm`?

packages/node/test/unit/dev.test.ts
```diff
@@ -18,7 +18,56 @@ function testForkDevServer(entrypoint: string) {
...
+  headers: http.IncomingHttpHeaders;
+  json: () => Promise<any>;
+  status: number | undefined;
+  text: string;
+}
+
+/**
+ * This is a minimal replacement for `node-fetch`. We cannot use the real
+ * `node-fetch` because it is a ESM package and Jest doesn't support
+ * statically or dynamically importing ESM packages.
```
dan-stowell: Does Jest not support importing ESM packages? Or just the version of Jest we're using?

packages/node/tsconfig.json
dan-stowell: @cb1kenobi can you land this change with `ES2020` or does this change have to wait on Node 14 EOL from Amazon?

packages/node/src/typescript.ts
cb1kenobi: `isEsm` is only set when `fixConfig()` is called by `startDevServer()` (e.g. `vc dev`). The intention was to not touch `vc deploy` or `vc build` code paths. `isEsm` will be true only if `vc dev` is executing an ESM function based https://github.com/vercel/vercel/blob/46348201b4eadb03b403f99dc5574c31d2b3926e/packages/node/src/index.ts#L554-L557

In other words, I don't feel a warning is necessary.

packages/node/test/unit/dev.test.ts
cb1kenobi: The latest version of Jest has "experimental" support for ESM source files. Our code must be ESM to import ESM packages.

Jest supports TypeScript via either Babel or `ts-jest`. We use `ts-jest`. We specify the transpilation settings in the `tsconfig.json`, that alone isn't enough. We need to make sure we're using a newer version of `ts-jest` which supports `.mts` file extensions.

We default all `.js` and `.ts` files to use CommonJS and thus those file extensiosn can never load ESM packages such as `node-fetch`. We can only load ESM packages if we set the `"type": "module"` in the `package.json`, which we can't do yet, or we use `.mjs` and `.mts` file extensions.

Given this test file is a `.ts` file, it cannot load the ESM package `node-fetch`. As a quick workaround, I mock `fetch()` using Node.js's built-in `http` module. Once we drop support for Node.js 16 and older, we can use Node 18's built-in `fetch()`.

packages/node/tsconfig.json
cb1kenobi: Today I tested using Node.js 14 with both `lib` and `target` set to `ES2021` and `vc dev` worked. In theory it should have not worked. Tomorrow I'll set them to `ES2020` and test again with Node 14 and 16.

packages/node/src/typescript.ts
styfle: @cb1kenobi If this is only used for development, I think it should happen in `startDevServer()` only since `fixConfig()` is used for both dev and prod. I dont want someone to see `isEsm` in the future and think they need to pass that variable for prod too.

packages/node/test/unit/dev.test.ts
styfle: One solution here is to set the tsconfig.json for jest to use `module: nodenext` and then you can import ESM packages just fine. I think you can do that here:

https://github.com/vercel/vercel/blob/66fe3db888ff1d736023ce7cb70e8cdd7ff69ed3/packages/node/test/tsconfig.json#L1-L4

cb1kenobi: I just tested `module: 'nodenext'` and it had no effect. I tried renaming `dev.test.ts` to `.mts` as well as set the target and libs to ES2020 and ts-jest to `useESM`. Nothing works. I like Jest a little less now.

cb1kenobi: Oh, I also tested `NODE_OPTIONS=--experimental-vm-modules` with no luck.

styfle: How about reverting `"node-fetch": "3.3.1"` back to the previous version? Why was it updated to begin with?

cb1kenobi: Reverted. Good question. I don't recall the exact reason for the update, but I recall running into issues at one point with importing the "default" for certain packages.

cb1kenobi: This -> https://github.com/vercel/vercel/actions/runs/4681143008/jobs/8293359158?pr=9692#step:6:354

packages/cli/test/dev/integration-5.test.ts
```diff
@@ -447,3 +447,129 @@ test(
...
+describe('[vercel dev] ESM edge functions', () => {
+  test(
+    '[vercel dev] ESM .js type=module',
+    testFixtureStdio(
+      'esm-js-edge-module',
+      async (_testPath: any, port: any) => {
+        let res = await fetch(`http://localhost:${port}/api/data`);
+        validateResponseHeaders(res);
+        const body = await res.text();
+        console.log(body);
```
EndangeredMassa: **suggestion (soft-blocking):** remove the `console.log` statements from these tests‏

packages/cli/test/dev/integration-5.test.ts
```diff
@@ -447,3 +447,129 @@ test(
...
+        const json = JSON.parse(body);
+        expect(json).toHaveProperty('isLeapYear');
+      },
+      { skipDeploy: true }
```
EndangeredMassa, line 462: **question (soft-blocking):** ‏Is the plan to remove `skipDeploy` after the build container is updated?

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -82,6 +82,9 @@ describe('build', () => {
     const output = join(cwd, '.vercel/output');
     try {
       process.chdir(cwd);
+      client.stdout.pipe(process.stdout);
+      client.stderr.pipe(process.stderr);
```
EndangeredMassa: **suggestion (soft-blocking):** ‏remove logs

Can we remove these (and the ones below in this file) so they don't show up every run?

Ideally, we'd capture the logs and show them if this test fails.

packages/next/src/edge-function-source/get-edge-function-source.ts
```diff
@@ -8,7 +8,7 @@ import zlib from 'zlib';
 import { promisify } from 'util';
 import { prettyBytes } from '../utils';
 
-// @ts-expect-error this is a prebuilt file, based on `../../scripts/build-edge-function-template.js`
```
EndangeredMassa, line 11: **suggestion (soft-blocking):** ‏revert

We should prefer `@ts-expect-error` so that it will itself be an error if the next line does not have a linting error.

Does `@ts-expect-error` not work here?

packages/node/src/dev-server.mts
```diff
@@ -6,18 +6,18 @@ if (!entrypoint) {
...
-import { createServerlessEventHandler } from './serverless-functions/serverless-handler';
-import { EdgeRuntimes, isEdgeRuntime, logError } from './utils';
+import { createServerlessEventHandler } from './serverless-functions/serverless-handler.mjs';
+import { EdgeRuntimes, isEdgeRuntime, logError } from './utils.js';
 import { getConfig } from '@vercel/static-config';
 import { Project } from 'ts-morph';
-import listen from 'async-listen';
+import asyncListen from 'async-listen';
+
+const { default: listen } = asyncListen;
```
EndangeredMassa, line 20: **question (non-blocking):** ‏In the old world, `import listen from 'async-listen'` would assign `listen` to the commonjs export's `default` property. In the new world, we have to reference the `default` property directly. Is this because the old world was more/less pretending to be JavaScript Modules and the new world is actually JavaScript Modules that has to do this to import CommonJS?

Reference: It looks like `async-listen` is published as CommonJS, but transpiled from TypeScript / Modules where it does `export default listen`.

EndangeredMassa (dismissed):
I have some notes, but nothing that would block this. Nice work!

packages/cli/test/dev/integration-5.test.ts
cb1kenobi: Fixed!

packages/cli/test/dev/integration-5.test.ts
cb1kenobi: Since this is a `vc dev` test, we don't need or want to deploy the project.

packages/cli/test/unit/commands/build/index.test.ts
cb1kenobi: Done!

packages/next/src/edge-function-source/get-edge-function-source.ts
cb1kenobi: This is an interesting one. I agree that we should prefer `@ts-expect-error`, however VS Code will complain
```
Unused '@ts-expect-errorr` directive. ts(2578)
```

Interestingly, if I remove it, VS Code is happy, however TypeScript will complain

```
Error: @vercel/next:build: src/edge-function-source/get-edge-function-source.ts(11,22): error TS2307: Cannot find module '../../dist/___get-nextjs-edge-function.js' or its corresponding type declarations.
```

So, if adding or removing `@ts-expect-error` causes issues, that leaves us with `@ts-ignore` which makes everything happy.

packages/node/src/dev-server.mts
cb1kenobi: I showed this in the virtual coffee 4/21. Since this file is a `.mts` file (e.g. a real ESM module), it doesn't resolve static default export in CommonJS files. You can't destructure `import { default as listen } ...` either.

Since this authoring of that code, `async-listen` has been "fixed" to export `listen()` as a named export: https://github.com/vercel/async-listen/releases/tag/2.1.0. I could update the `async-listen` dependency and in theory the code would be destructurable like `import { listen } from 'async-listen'`, but not 100% certain.

EndangeredMassa: Ohh, right. This was that weird case.

EndangeredMassa (approved):

packages/next/src/edge-function-source/get-edge-function-source.ts
EndangeredMassa: Ah, that's a TypeScript version mismatch between the project and VS Code. I think we can solve this: https://github.com/vercel/vercel/pull/9858

cb1kenobi: Ah, VS Code is using 5.0.4.

Kikobeats (approved):

packages/static-build/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "allowJs": true,
```
styfle, line 3: Why did this setting change?

cb1kenobi: I changed it in `next` and `node` builders also. It was added because `ts-jest` logs a warning:

```
ts-jest[ts-jest-transformer] (WARN) Got a `.js` file to compile while `allowJS` option is not set to `true`
```

The warning says to enable `allowJs: true` to fix it. In hindsight, this probably should have been added to the `test/tsconfig.json` file instead of the package-level, but I didn't see any side effects of it being in the package-level `tsconfig.json`.

Do you think I should move them?

cb1kenobi: @styfle Just in case :) https://github.com/vercel/vercel/pull/9862

PR #9694, opened 2023-03-19
[node] support node:[lib] in vc dev

Schniz:
- [node] add edge-node-compat-plugin
- [node] implement edge-handler


packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -2,7 +2,7 @@
 /* global addEventListener */
 
 function buildUrl(requestDetails) {
-  let proto = requestDetails.headers['x-forwarded-proto'].split(/\b/).shift(); // handling multi-protocol like https,http://...
+  let proto = requestDetails.headers['x-forwarded-proto']?.split(/\b/).shift(); // handling multi-protocol like https,http://...
```
EndangeredMassa: **suggestion (non-blocking):** revert this change

Just in case this happens, I think we should throw a more explicit error message. I opened a PR with some changes around error handling that covers this case: https://github.com/vercel/vercel/pull/9697

EndangeredMassa (dismissed):

packages/node/src/edge-functions/edge-handler-template.js
cb1kenobi: I not sure we should revert. Now that I think about it, the current code seems bugged as it's possible for the current code to resolve `undefined` which would cause the result to be `'undefined://hostname/path'`, which is certainly not what we want.

Instead of throwing an error, we could just default `proto` to the same protocol as parsed from `requestDetails.url`. That's basically what happens if `x-forwarded-proto` isn't specified.

Schniz: I added this because the test I added failed for this reason and it seemed better to change this instead of failing for a missing header. We can revert this change tomorrow and I’ll add the header in my test.

EndangeredMassa: The code as-is on main will at least fail close to the root of the problem (missing `x-forwarded-proto` header). I think it's worth reverting for this PR.

Let's get a better fix in https://github.com/vercel/vercel/pull/9697

EndangeredMassa: We ended up implementing fallbacks for these values: https://github.com/vercel/vercel/pull/9697

cb1kenobi (commented):
Reviewed and tested. Looks good once the merge conflicts are resolved.

javivelasco (approved):

cb1kenobi (approved):

EndangeredMassa (approved):

PR #9697, opened 2023-03-20
[node] Improve edge-handler-template error handling

EndangeredMassa:
When passing the `Request` into the user's edge handler, the url is constructed from proxy headers from the edge runtime. There are cases where these headers can be overridden (like using cloudflare tunnel). To protect for those cases, this PR updates the url construction to fall back to likely good values.

* host: `127.0.0.1`
* path: `/`
* proto: `http`

### DIFF
--- a/packages/node/src/edge-functions/edge-handler-template.js
+++ b/packages/node/src/edge-functions/edge-handler-template.js
@@ -2,21 +2,18 @@
 /* global addEventListener */
 
 function buildUrl(requestDetails) {
+  const host = requestDetails.headers['x-forwarded-host'] || '127.0.0.1';
+  const path = requestDetails.url || '/';
+
   const allProtocols = requestDetails.headers['x-forwarded-proto'];
-  const host = requestDetails.headers['x-forwarded-host'];
-  const path = requestDetails.url;
-
-  if (!allProtocols) {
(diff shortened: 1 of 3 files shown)


cb1kenobi (dismissed):

packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -2,9 +2,21 @@
...
-  let proto = requestDetails.headers['x-forwarded-proto'].split(/\b/).shift(); // handling multi-protocol like https,http://...
-  let host = requestDetails.headers['x-forwarded-host'];
-  let path = requestDetails.url;
+  const allProtocols = requestDetails.headers['x-forwarded-proto'];
+  const host = requestDetails.headers['x-forwarded-host'];
+  const path = requestDetails.url;
+
+  if (!allProtocols) {
+    throw new Error('missing header "x-forwarded-proto"');
+  }
```
EndangeredMassa: Instead, let's fall back to the protocol of `requestDetails.url`.

EndangeredMassa: Turns out that the `url` here is just a path

cb1kenobi (approved):

PR #9700, opened 2023-03-21
[node][remix] Update nft conditions to include "edge-light" for Edge Functions

TooTallNate:


Ethan-Arrowood (approved):

packages/node/src/index.ts
```diff
@@ -175,7 +175,9 @@ async function compile(
     return source;
   }
 
-  const conditions = isEdgeFunction ? ['worker', 'browser'] : undefined;
+  const conditions = isEdgeFunction
+    ? ['edge-light', 'browser', 'module', 'import', 'require']
```
styfle, line 179: Why `require()`? Isn't that going to include files that don't work on the Edge Runtime?

TooTallNate: CJS does still work in Edge runtime.

PR #9702, opened 2023-03-22
[node-bridge] remove AWS Launcher

Kikobeats:
It is not used anymore


styfle (commented):
I can see that `@vercel/node` uses the aws launcher here:

https://github.com/vercel/vercel/blob/20bd71ce70aee9e8aa969ea19ef9e9d2a70aaa3d/packages/node/src/index.ts#L352-L365

PR #9704, opened 2023-03-22
[next] Remove extra env variables

ijjk:
These env variables were temporarily added to bust turbo/nx cache although this should no longer be needed as a proper fix has been landed in Next.js itself. 

x-ref: [slack thread](https://vercel.slack.com/archives/C04Q0GWSB8W/p1679518040946589)


styfle (commented):
Will this impact older versions of Next.js?

ijjk:
Those versions are already affected as it was only a partial fix

styfle (approved):

PR #9705, opened 2023-03-23
Add vercel/otel package

jankaifer:
We are working on OpenTelemetry support for Next.js, and because open telemetry API is convoluted and complex, we wanted to make it easier for users to get started.

Usually, when using OTEL, you need to set up how you want to export your traces (`register` function), and you need a bunch of dependencies here. You need a bunch of dependencies here, and this is the most complex part.

Also without this package dependencies need to be conditionally required based on the environment which is also needlessly complex when don't know that concept.

The instrumentation is relatively easy since you need only `@opentelemetry/api` (and sometimes `@opentelemetry/resources`, but that's only for advanced instrumentation).
We reexport those two packages so that `@vercel/otel` is the only hotel package that you need.

### DIFF
--- a/packages/otel/package.json
+++ b/packages/otel/package.json
@@ -1,0 +1,61 @@
+{
+  "name": "@vercel/otel",
+  "version": "0.0.1",
+  "description": "Vercel wrapper around OpenTelemetry APIs",
+  "exports": {
+    ".": {
+      "edge": {
+        "types": "./dist/index.edge.d.ts",
+        "default": "./dist/index.edge.js"
+      },
+      "edge-light": {
+        "types": "./dist/index.edge.d.ts",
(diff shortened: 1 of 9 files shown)


styfle (commented):
I don't this new package belongs in this repo. 

Some reasons this repo isn't a good idea:
- This repo is for Vercel CLI and build/deploy features. This seems like a package an end user would install so it probably belongs in a separate repo.
- Packages in this monorepo cannot ship independently. They all ship under Vercel CLI as seen [here](https://github.com/vercel/vercel/releases).
- Issues are not monitored. The assumption is that any issue should go through Customer Support since it would involve personal information like deployment url, team id, etc.

Also it needs tests 🙂

PR #9708, opened 2023-03-23
[go] Support 'go.work' file and resolve shared deps relative to work path

cb1kenobi:
This PR fixes a few issues related to `vc dev`.

1. Create a default `go.work` file in the cache dir when building the `vercel-dev-server-go` executable
2. Copy the existing `go.mod` file into the cache dir and update any "replace" relative paths
3. Split the "build" logic into separate functions for the legacy "main" package build and the `go.mod` build

Additionally, it fixes:

1. `vc build`: pass in `build.env` from `vercel.json`
2. Fix several tests to work with `vc dev` and `vc build`

Linear: https://linear.app/vercel/issue/VCCLI-638/vc-dev-go-builder-cant-resolve-workspace-dependencies

The user that reported the issue has tested this build and seems to fix their use case: https://github.com/vercel/vercel/issues/9393#issuecomment-1490726785.


packages/go/test/fixtures/26-go-work-with-shared/vercel.json
```diff
@@ -0,0 +1,6 @@
+{
+  "version": 2,
+  "builds": [
+    { "src": "api/index.go", "use": "@vercel/go" }
+  ]
+}
```
styfle: This vercel.json file can be removed

styfle (commented):
> This PR fixes a few issues related to `vc dev`

Can you add a vc dev test?

For example: https://github.com/vercel/vercel/blob/ad107ecf79a15082dff683855e066d0b251e9127/packages/cli/test/dev/integration-2.test.ts#L172-L183

vedantmgoyal2009:
Please give the URL for this PR which I can put in `VERCEL_CLI_VERSION` environment variable and test it.

packages/go/test/fixtures/26-go-work-with-shared/vercel.json
cb1kenobi: Done!

vedantmgoyal9 (dismissed):
Fixes #9393. Thanks a lot for #9576 and this :shipit:

EndangeredMassa (changes requested):
This go project works with latest `vercel`, but not with this branch: https://github.com/EndangeredMassa/rdhawladar-golang-vercel

> api/entrypoint.go:8:2: package golang-vercel/docs is not in GOROOT (/Users/smassa/source/demo/golang-vercel/.vercel/cache/golang/src/golang-vercel/docs)
> Error: Command failed: go build -ldflags -s -w -o ./vercel-dev-server-go ./...

EndangeredMassa (approved):
I tested some Go projects with this and everything seems to work. Thanks!

vedantmgoyal2009:
@cb1kenobi #10016. Also, I tried prefixing files with "_" (underscore) as mentioned in https://vercel.com/docs/concepts/functions/serverless-functions#adding-utility-files-to-the-/api-directory, but it seems that it also gets ignored.

cb1kenobi:
@vedantmgoyal2009 Yeah, I don't think the underscore convention works with Go serverless functions. Go support is still considered beta. I think we would need to do a bit of refactoring to support this and it would be a good thing to do. :)

vedantmgoyal2009:
Since we don't have to do any debugging this time, I don't think it should take much longer. Actually, my project is still in development and I'm trying to implement the best ways to do a thing that leads to these types of feature requests. Hope I helped in polishing up your lib...

cb1kenobi:
@vedantmgoyal2009 sorry, I haven't had time to look into it. I created an issue in our internal ticket system, but we haven't prioritized it yet. Go support it beta and so there will be functionality gaps like the underscore prefix.

PR #9709, opened 2023-03-24
[build-utils] Fix system env var detection for prefixed env vars

styfle:
This PR fixes a bug that was causing too many env vars to be exposed to the frontend by adding the framework `envPrefix`.

Some frameworks, such as CRA, will include all of these in the frontend even if not explicitly used so we must only prefix known [System Environment Variables](https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables).

This PR adds an allowlist to ensure this is handle properly.

- Fixes https://linear.app/vercel/issue/VCCLI-639


cb1kenobi (approved):
Good catch!

PR #9711, opened 2023-03-24
[node-bridge]: remove API Gateway normalization

Kikobeats:
We no longer use since a long time ago

### DIFF
--- a/packages/node-bridge/bridge.js
+++ b/packages/node-bridge/bridge.js
@@ -84,17 +84,12 @@ function normalizeProxyEvent(event) {
 }
 
 /**
- * @param {import('./types').VercelProxyEvent | import('aws-lambda').APIGatewayProxyEvent} event
- * @return {import('./types').VercelProxyRequest | undefined }
+ * @param {import('./types').VercelProxyEvent } event
+ * @return {import('./types').VercelProxyRequest }
  */
 function normalizeEvent(event) {
-  if ('Action' in event) {
-    if (event.Action === 'Invoke') {
-      return normalizeProxyEvent(event);
-    } else {
-      throw new Error(`Unexpected event.Action: ${event.Action}`);
-    }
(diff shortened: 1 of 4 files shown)


packages/node-bridge/bridge.js
```diff
@@ -83,40 +83,9 @@ function normalizeProxyEvent(event) {
...
-    responseCallbackCipherKey: undefined,
-    responseCallbackStream: undefined,
-    responseCallbackUrl: undefined,
-  };
-}
-
 /**
  * @param {import('./types').VercelProxyEvent | import('aws-lambda').APIGatewayProxyEvent} event
- * @return {import('./types').VercelProxyRequest}
+ * @return {import('./types').VercelProxyRequest | undefined }
```
Kikobeats: maybe should the method throw an error if `event.Action` is `undefined`?

TooTallNate: Ya I think so

packages/node-bridge/bridge.js
```diff
@@ -83,40 +83,9 @@ function normalizeProxyEvent(event) {
...
-    responseCallbackCipher: undefined,
-    responseCallbackCipherIV: undefined,
-    responseCallbackCipherKey: undefined,
-    responseCallbackStream: undefined,
-    responseCallbackUrl: undefined,
-  };
-}
-
 /**
  * @param {import('./types').VercelProxyEvent | import('aws-lambda').APIGatewayProxyEvent} event
```
TooTallNate:
```suggestion
 * @param {import('./types').VercelProxyEvent} event
```

styfle (approved):
I think this is no longer used, but you might want to try it on staging to be sure

styfle:
Technically this should be `semver: major`

ijjk (approved):

PR #9717, opened 2023-03-27
[cli] support `build -y` shorthand

EndangeredMassa:
The `build` command wasn't supporting the `-y` shorthand for `--yes`. Now it does.

I looked at the other uses and they all seem fine.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -157,6 +157,7 @@ export default async function main(client: Client): Promise<number> {
     '--output': String,
     '--prod': Boolean,
     '--yes': Boolean,
+    '-y': '--yes',
   });
 
   if (argv['--help']) {


cb1kenobi (approved):

packages/cli/src/commands/build.ts
```diff
@@ -157,6 +157,7 @@ export default async function main(client: Client): Promise<number> {
     '--output': String,
     '--prod': Boolean,
     '--yes': Boolean,
+    '-y': '--yes',
```
styfle, line 160: Let's also make sure all of the other commands support this too

styfle (approved):

packages/cli/src/commands/build.ts
EndangeredMassa: I did check all the other commands to make sure they shorthand was supported

PR #9719, opened 2023-03-27
[cli] refactor pkg.ts into `@vercel-internals/get-package-json`

Ethan-Arrowood:
Extracts the logic from `pkg.ts` into a new utility function `getPackageJSON` from `@vercel-internals/utils`.


internals/get-package-json/src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  let rootDir = path.dirname(filePath);
+
+  let packageJSONPath = getPackageJSONPath(rootDir);
+  while (!fs.existsSync(packageJSONPath)) {
+    rootDir = path.join(rootDir, '..');
+    packageJSONPath = getPackageJSONPath(rootDir);
+  }
+
+  Error.prepareStackTrace = _prepareStackTrace;
```
TooTallNate: Let's restore this immediately after `new Error()` on line 14.

internals/get-package-json/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  const stack = (new Error().stack as unknown as Array<NodeJS.CallSite>).slice(
+    1
+  ); // remove self from stack
+
+  Error.prepareStackTrace = _prepareStackTrace;
+
+  // TypeScript rule noUncheckedIndexedAccess makes `stack[0]` potentially undefined;
+  // however, we can safely assert it is not undefined as the callstack will always
+  // have atleast 2 entries (we slice the first above), and then rely on the second here.
+  const callsite = stack[0]!;
```
styfle: Why do you need to `stack.slice(1)` and then `stack[0]`.

Could you instead use `stack[1]` and then you don't need to slice?

pnpm-lock.yaml
```diff
@@ -5355,7 +5524,7 @@ packages:
   /@types/graceful-fs/4.1.6:
     resolution: {integrity: sha512-Sig0SNORX9fdW+bQuTEovKj3uHcUL6LQKbCrrqb1X7J6/ReAbhCXRAhc+SMejhLELFj2QcyuxmUooZ4bt5ReSw==}
     dependencies:
-      '@types/node': 18.14.6
+      '@types/node': 16.18.11
```
styfle: Is it expected for this `@types/node` version to revert to a previous version?

pnpm-lock.yaml
```diff
@@ -18262,7 +18894,7 @@ packages:
     dev: true
 
   /utils-merge/1.0.1:
-    resolution: {integrity: sha512-pMZTvIkT1d+TFGvDOqodOclx0QWkkgi6Tdoa8gC8ffGAAqz9pzPTZWAybbsHHoED/ztMtkv/VoYTYyShUn81hA==}
+    resolution: {integrity: sha1-n5VxD1CiZ5R7LMwSR0HBAoQn5xM=}
```
styfle: Any idea why this was changed from sha512 to sha1?

Ethan-Arrowood: No, and I can't figure out why it happened.

internals/get-package-json/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "private": true,
+  "name": "@vercel-internals/get-package-json",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
```
TooTallNate:
```suggestion
    "build": "tsc",
```

TooTallNate:
Looks like `eslintignore` needs to be updated to ignore some test fixtures.

internals/utils/package.json
```diff
@@ -1,17 +0,0 @@
-{
-  "private": true,
-  "name": "@vercel-internals/utils",
```
styfle, line 3: Did you intend to delete this?

Ethan-Arrowood: yeah cause we renamed it. I can retain this empty utils package if we want

styfle: Oh I didn't realize it was empty. I must have confused that with the other internal packages

Ethan-Arrowood: Yeah I added it as empty previously just so that i had something to work off of. But I think we've come around to decoupling things as the best path forward.

internals/get-package-json/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "private": true,
+  "name": "@vercel-internals/get-package-json",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "test-unit": "jest tests/unit"
```
styfle: Is this running in CI?

I think you need to add the generic `test` like the other packages so it gets picked up by CI

internals/get-package-json/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "private": true,
+  "name": "@vercel-internals/get-package-json",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "test": "pnpm test-unit",
```
styfle:
```suggestion
    "test": "jest --env node --verbose --runInBand --bail"
```

All the other packages use this. 

I'm still not seeing this run in CI either 🤔

Ethan-Arrowood: oh i see i got it backwards

internals/get-package-json/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  // The `stack` reference above is of type `Array<NodeJS.CallSite>`
+  const stack = new Error().stack as unknown as Array<NodeJS.CallSite>;
+
+  Error.prepareStackTrace = _prepareStackTrace;
+
+  // TypeScript rule noUncheckedIndexedAccess makes `stack[1]` potentially undefined;
+  // however, we can safely assert it is not undefined as the callstack will always
+  // have at least 2 entries, the first being this function, and the second wherever
+  // its being called from.
+  const callsite = stack[1]!;
```
EndangeredMassa: **suggestion (soft-blocking):** use `try-finally`‏

```suggestion
  const _prepareStackTrace = Error.prepareStackTrace;
  let callSite:
  try {
    Error.prepareStackTrace = (_, stack) => stack;

    // The `stack` reference above is of type `Array<NodeJS.CallSite>`
    const callSites = new Error().stack as unknown as Array<NodeJS.CallSite>;
    
    // TypeScript rule noUncheckedIndexedAccess makes `stack[1]` potentially undefined;
	  // however, we can safely assert it is not undefined as the callstack will always
	  // have at least 2 entries, the first being this function, and the second wherever
	  // its being called from.
	  callsite = stack[1]!;
  } catch {
    // TODO: what should we do if this fails? Throw a better error?
  } finally {
    Error.prepareStackTrace = _prepareStackTrace;
  }
```

I renamed `stack` to `callSites` here. I think it's useful to make it clear that this is not the normal `Error#stack`.

I also suggest extracting this bit into a function that makes it clear it does one specific thing to make it easier to understand. Maybe `captureCallerCallSite`.

internals/get-package-json/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  let packageJSONPath = getPackageJSONPath(rootDir);
+  while (!fs.existsSync(packageJSONPath)) {
+    rootDir = path.join(rootDir, '..');
+    packageJSONPath = getPackageJSONPath(rootDir);
+  }
+
+  const packageJSON = JSON.parse(fs.readFileSync(packageJSONPath, 'utf-8'));
+
+  return packageJSON;
```
EndangeredMassa, line 57: **suggestion (soft-blocking):** ‏cast to `PackageJson` type

If we assert the type here (as we do in other places), we won't need to type it later to `as typeof import('../../pacakge.json)` in "packages/cli/src/util/pkg.ts". We could then more easily get rid of "packages/cli/src/util/pkg.ts" and depend on this internal package directly.

Or am I missing some TypeScript complexity?

Ethan-Arrowood: IMO having this function do the casting is dangerous. Unless we add JSON validation, the best return type for this function is `any` or `unknown`. Because authors may _think_ we are validating when we really aren't and it could introduce hard to find bugs. Leaving the casting up to the user of the function is best because then it is clear in their code where that type comes from (and thus if there are issues they can quickly see "oh i'm type casting when I should be type validating").

Ethan-Arrowood: Also, replacing `pkg.ts` with this function call everywhere is going to be a perf hit, as every code path that uses it would have to re-execute this entire function every time they need a value from package.json

It would be cool to add some kind of auto caching/memoization so that this function could be reused in a code base, but not have to find and read the package.json file every time.

internals/get-package-json/src/index.ts
Ethan-Arrowood: Agree with all of this. I'll make some adjustments now

internals/get-package-json/src/index.ts
Ethan-Arrowood: Added caching 😄

styfle (dismissed):

internals/get-package-json/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    packageJSONPath = getPackageJSONPath(rootDir);
+  }
+
+  if (cache.has(packageJSONPath)) {
+    return cache.get(packageJSONPath);
+  }
+
+  const packageJSON = JSON.parse(fs.readFileSync(packageJSONPath, 'utf-8'));
+
+  cache.set(packageJSONPath, packageJSON);
```
TooTallNate:
```suggestion
  let packageJSON = cache.get(packageJSONPath);

  if (!packageJSON) {
    packageJSON = JSON.parse(fs.readFileSync(packageJSONPath, 'utf-8'));
    cache.set(packageJSONPath, packageJSON);
  }
```

TooTallNate (changes requested):
Small nit

TooTallNate (approved):

internals/get-package-json/src/index.ts
EndangeredMassa: > IMO having this function do the casting is dangerous.

It is a weaker typing, that's true.

> replacing pkg.ts with this function call everywhere is going to be a perf hit

Oh, you're doing a one-time execution in `pkg.ts` and returning the result. I didn't notice that. Yeah, it's worth keeping in that form.

> Added caching 😄

Sounds good!

EndangeredMassa (approved):

styfle (approved):

PR #9726, opened 2023-03-28
[build-utils][next] move `operationType` to virtual lambdas to fix ISR detection

EndangeredMassa:
When `operationType` was added to `Lambda` instances, there was an assumption that a single lambda contains a single operation type. This turned out to not be true.

In Next.js, for example, multiple lambdas of different operation types (such as SSG and ISR) can be bundled into the same Lambda instance.

To be able to differentiate these in the deployment metadata that's shown in the build/deploy UI, operation type has to be kept track of in the virtual lambda that's inside the physical `Lambda` instance.

---

### `build-utils`

This PR deprecates `Lambdas#operationType` in favor of `Lambda#virtualLambdas`, which is an object that maps a path to a `VirtualLambda`. For now, `VirtualLambda` only contains `path` and `operationType`, but we could easily expand it to include other details in the future.

The default for `Prerender`s was also changed from `ISR` to `SSG`, which better matches how these terms are used by frameworks.

### `next` builder

This PR changes the determination of `operationType` such that all `Prerenders` are considered "SSG" unless there's a `revalidate` property, which means it's "ISR".

---

* depends on: https://github.com/vercel/vercel/pull/9739


packages/next/test/integration/index.test.js
```diff
@@ -169,14 +175,13 @@ it('should build using server build', async () => {
...
-      .operationType
-  ).toBe('ISR');
+    getOperationType(output, '_next/data/testing-build-id/fallback/[slug].json')
+  ).toBe(undefined);
 
   expect(output['fallback/first'].type).toBe('Prerender');
   expect(output['fallback/first'].allowQuery).toEqual([]);
-  expect(output['fallback/first'].lambda.operationType).toBe('ISR');
   expect(output['fallback/first'].sourcePath).toBe('/fallback/[slug]');
+  expect(getOperationType(output, 'fallback/first')).toBe(undefined);
```
EndangeredMassa, line 184: I think this is supposed to be `SSG` as well. Because of the fallback to the deprecated `Lambda#operationType` and how that defaults to `SSG` for Prerenders, this will work as expected, but it's not great to rely on the deprecated property.

I'm not sure where this path gets added to the outputs in a way that misses the virtual lambda assignment. If someone can help me find that, I'd be happy to update it there.

cb1kenobi (dismissed):

packages/build-utils/src/prerender.ts
```diff
@@ -45,8 +45,8 @@ export class Prerender {
 
     this.lambda = lambda;
     if (this.lambda) {
-      // "ISR" is the platform default lambda label for prerender functions
-      this.lambda.operationType = this.lambda.operationType || 'ISR';
+      // "SSG" is the platform default lambda label for prerender functions
+      this.lambda.operationType = this.lambda.operationType || 'SSG';
```
ijjk, line 49: I don't think we should be labeling this as SSG as mentioned in slack since a Prerender implies ISR especially with On-Demand Revalidate now.

EndangeredMassa:
We don't need to differentiate these, yet

PR #9727, opened 2023-03-28
[build-utils] Allow corepack resolution without lockfiles

Ethan-Arrowood:
Adds support for determining package manager using corepack without a lockfile existing.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -392,7 +392,9 @@ export async function runNpmInstall(
   try {
     await runNpmInstallSema.acquire();
     const { cliType, packageJsonPath, lockfileVersion } = await scanParentDirs(
-      destPath
+      destPath,
+      true,
+      spawnOpts?.env?.ENABLE_EXPERIMENTAL_COREPACK === '1'
     );
 
     // Only allow `runNpmInstall()` to run once per `package.json`
@@ -568,7 +570,11 @@ export async function runCustomInstallCommand({
   spawnOpts?: SpawnOptions;
 }) {
   console.log(`Running "install" command: \`${installCommand}\`...`);
-  const { cliType, lockfileVersion } = await scanParentDirs(destPath);
+  const { cliType, lockfileVersion } = await scanParentDirs(
+    destPath,
+    true,
+    spawnOpts?.env?.ENABLE_EXPERIMENTAL_COREPACK === '1'
+  );
   const env = getEnvForPackageManager({
     cliType,
     lockfileVersion,
@@ -592,7 +598,8 @@ export async function runPackageJsonScript(
 
   const { packageJson, cliType, lockfileVersion } = await scanParentDirs(
     destPath,
-    true
+    true,
+    spawnOpts?.env?.ENABLE_EXPERIMENTAL_COREPACK === '1'
   );
   const scriptName = getScriptName(
     packageJson,
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -234,7 +234,8 @@ export async function getNodeVersion(
 
 export async function scanParentDirs(
   destPath: string,
-  readPackageJson = false
+  readPackageJson = false,
+  corepackEnabled = false
 ): Promise<ScanParentDirsResult> {
   assert(path.isAbsolute(destPath));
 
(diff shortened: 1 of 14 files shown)


packages/build-utils/test/fixtures/29-corepack-without-lockfile/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "private": true,
+  "scripts": {
+    "build": "mkdir -p public && (printf \"pnpm version: \" && pnpm -v) > public/index.txt"
+  },
+  "dependencies": {
+    "once": "^1.4.0"
+  },
```
styfle: No need for dependencies anymore

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -234,7 +234,8 @@ export async function getNodeVersion(
 
 export async function scanParentDirs(
   destPath: string,
-  readPackageJson = false
+  readPackageJson = false,
+  corepackEnabled = false
```
styfle: Lets use `options` here instead because this could turn into a lot more props in the future

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -371,7 +392,9 @@ export async function runNpmInstall(
   try {
     await runNpmInstallSema.acquire();
     const { cliType, packageJsonPath, lockfileVersion } = await scanParentDirs(
-      destPath
+      destPath,
+      true,
+      spawnOpts?.env?.ENABLE_EXPERIMENTAL_COREPACK === '1'
```
styfle: Maybe we should pass the whole `spawnOpts` instead?

Ethan-Arrowood: sure

styfle (changes requested):
Every package calling `scanParentDirs()` also needs to be updated

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -247,42 +248,62 @@ export async function scanParentDirs(
...
-    cliType = 'pnpm';
-    lockfilePath = pnpmLockPath;
-    lockfileVersion = Number(pnpmLockYaml.lockfileVersion);
-  } else if (packageLockJson) {
-    cliType = 'npm';
-    lockfilePath = npmLockPath;
-    lockfileVersion = packageLockJson.lockfileVersion;
+  let cliType: CliType | undefined;
+
+  if (corepackEnabled && readPackageJson) {
```
styfle, line 280: Not sure about this `readPackageJson` boolean.

I'm starting to think it always needs to read or else corepack won't be detected, right?

Ethan-Arrowood: this function is doing too much IMO. 
it seems to be called in some places purely to read the package.json, and in other places to do the cliType and lockfile resolution. maybe splitting it up would be better

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -234,7 +236,13 @@ export async function getNodeVersion(
...
 export async function scanParentDirs(
   destPath: string,
-  readPackageJson = false
+  {
+    readPackageJson = false,
+    spawnOpts,
+  }: {
+    readPackageJson?: boolean;
+    spawnOpts?: SpawnOptions;
+  } = {}
```
styfle, line 251: We can't introduce breaking changes to build-utils because it gets injected for 3rd party builders.

So this will need to support the old api signature too

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -247,42 +255,62 @@ export async function scanParentDirs(
...
-    lockfilePath = pnpmLockPath;
-    lockfileVersion = Number(pnpmLockYaml.lockfileVersion);
-  } else if (packageLockJson) {
-    cliType = 'npm';
-    lockfilePath = npmLockPath;
-    lockfileVersion = packageLockJson.lockfileVersion;
+  let cliType: CliType | undefined;
+
+  if (spawnOpts?.env?.ENABLE_EXPERIMENTAL_COREPACK === '1' && readPackageJson) {
+    const packageManager = packageJson?.packageManager?.split('@')[0]; // { "packageManager": "pnpm@8.0.0" } -> "pnpm"
```
styfle, line 281: Another problem: the `packageJson` here is from `walkParentDirs()` but finding the nearest package.json is not enough. It has to be the nearest one that contains `packageManager` prop. So we'll likely need to add a predicate to `walkParentDirs()`

packages/build-utils/src/fs/run-user-scripts.ts
styfle: Yeah I think the cases that only care about package.json could call `walkParentDirs()` directly

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -234,55 +242,86 @@ export async function getNodeVersion(
...
   assert(path.isAbsolute(destPath));
 
   const pkgJsonPath = await walkParentDirs({
     base: '/',
     start: destPath,
     filename: 'package.json',
+    predicate: fullpath => {
+      const data = fs.readFileSync(fullpath, 'utf-8');
+      const json = JSON.parse(data);
+      return json.packageManager != undefined;
```
styfle:
```suggestion
      return type json.packageManager === 'string';
```

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -294,6 +333,7 @@ export async function walkParentDirs({
   base,
   start,
   filename,
+  predicate = () => true,
```
styfle, line 342:
```suggestion
  predicate = (fullPath: string) => true,
```

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -234,55 +242,86 @@ export async function getNodeVersion(
...
   assert(path.isAbsolute(destPath));
 
   const pkgJsonPath = await walkParentDirs({
     base: '/',
     start: destPath,
     filename: 'package.json',
+    predicate: fullpath => {
+      const data = fs.readFileSync(fullpath, 'utf-8');
+      const json = JSON.parse(data);
+      return type json.packageManager === 'string';
```
styfle: I think this is wrong depending on the caller because only the case of corepack would care about the `packageManager` prop.

The code as written would always return false for every package.json if they don't have `packageManager` prop.

Ethan-Arrowood: true

PR #9739, opened 2023-03-31
[tests] update typescript version used in next fixtures

EndangeredMassa:
Updates the typescript version in next fixtures to accommodate newer syntax used in types for `next@canary`.


packages/next/test/fixtures/00-legacy-mono-repo-server-mode/packages/nextjs/package.json
```diff
@@ -10,7 +10,7 @@
     "next": "canary",
     "react": "latest",
     "react-dom": "latest",
-    "typescript": "4.2.3"
+    "typescript": "4.9.5"
```
ijjk, line 13:
```suggestion
    "typescript": "latest"
```
I think we're good to just use latest here so we don't need to keep bumping

packages/next/test/fixtures/06-lambda-with-memory/package.json
```diff
@@ -6,6 +6,6 @@
     "next": "canary",
     "react": "latest",
     "react-dom": "latest",
-    "typescript": "^4.0.2"
+    "typescript": "4.9.5"
```
ijjk, line 9:
```suggestion
    "typescript": "latest"
```

cb1kenobi (dismissed):

packages/next/test/fixtures/00-legacy-mono-repo-server-mode/packages/nextjs/package.json
EndangeredMassa: I don't like using latest because random PRs can fail because of a TS change. I think we should also change the fixtures to not use next canary or latest for the same reason. If we want to keep track of updated code, we can leverage cron tasks to update the versions and submit new PRs.

I think we should merge this PR as-is to fix this specific problem. We can always discuss other approaches to this general problem outside of this blocking issue.

styfle: Agreed, we should use cron to update tests the same way we use cron to update examples

https://github.com/vercel/vercel/blob/main/.github/workflows/cron-update-next.yml

That would make it much more clear that a failure is due to a next/react/typescript update rather than a regression introduced in `@vercel/next`

styfle (dismissed):

cb1kenobi (approved):

PR #9502, opened 2023-02-21
[node] Handle multi-protocol in Edge Functions

PonomareVlad:
This is proposal to fix #9501

PR #9745, opened 2023-04-03
[node] Add streaming support for `vc dev`

Kikobeats:
Until now, the user code response it's buffered and serialized. This is mismatching how Vercel works these days.

This PR enables streaming response in `vc dev` for Edge/Serverless.

As part of the implementation, the `node-bridge` which spawns a process to consume the user code is not necessary anymore.

Some necessary files (like HTTP server helpers) have been moved to live in node builder package instead.


packages/node-bridge/launcher.js
```diff
@@ -1,199 +0,0 @@
...
-      process.env.NODE_ENV = region === 'dev1' ? 'development' : 'production';
-    }
-
-    /**
-     * @param {string} p - entrypointPath
-     */
-    async function getListener(p) {
-      let listener = useRequire
-        ? require(p)
-        : await import(isAbsolute(p) ? pathToFileURL(p).href : p);
```
Kikobeats, line 71: I'm not totally sure, but it could be possible we have to migrate this helper

Kikobeats:
This logic
https://github.com/vercel/vercel/blob/d8f5052d9a93646b88770e5336e2b497a1a44623/packages/node/src/fork-dev-server.ts#L30

was added at that level to be passed to node-bridge process using Node.js flags.

It's now implemented as part of `serverless-handler` logic:
https://github.com/vercel/vercel/pull/9745/files#diff-080055c718bfdb1c0351a2c0421846f8e01a84f628f1c52c2a90d8bb5edb41eaR45

I think it's safe to drop, need double-check by @Schniz there 👁️

packages/node/src/utils.ts
```diff
@@ -1,6 +1,8 @@
 import { extname } from 'path';
 import { pathToRegexp } from 'path-to-regexp';
 import { debug } from '@vercel/build-utils';
+import { IncomingMessage } from 'http';
+import { streamToBuffer } from '@vercel/build-utils';
```
styfle: Move this into line 3

packages/node/package.json
```diff
@@ -29,17 +29,18 @@
...
     "@vercel/static-config": "2.0.14",
+    "async-listen": "1.2.0",
     "edge-runtime": "2.0.0",
     "esbuild": "0.14.47",
     "exit-hook": "2.2.1",
-    "node-fetch": "2.6.7",
     "ts-node": "10.9.1",
-    "typescript": "4.3.4"
+    "typescript": "4.3.4",
+    "undici": "5.21.0"
```
styfle: I think all new `dependencies` need to be `devDependencies` unless you explicitly add to `--external` in the ncc build script

packages/node/src/serverless-functions/serverless-handler.ts
```diff
@@ -1,58 +1,74 @@
...
+      ? require(entrypointPath)
+      : await dynamicImport(entrypointPath);
+  }
+
+  const server = await createServerlessServer(userCode, options);
 
-    return bridge.launcher(event, {
-      callbackWaitsForEmptyEventLoop: false,
+  return async function (request: IncomingMessage) {
+    const response = await undici.fetch(server.url, {
```
EndangeredMassa: **question (non-blocking):** Why did we switch from `node-fetch` to `undici.fetch` here? 

@Ethan-Arrowood: What's our overall plan with `fetch`?‏

packages/node/src/serverless-functions/dynamic-import.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  dynamicImport: filepath => import(filepath).then(mod => mod.default),
```
styfle: Is `.default` guaranteed to exist? What about the case with `.default.default`?

packages/node/src/serverless-functions/helpers.ts
```diff
@@ -91,23 +63,43 @@ function redirect(
   return res;
 }
 
-function setCharset(type: string, charset: string) {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const { parse, format } = require('content-type');
-  const parsed = parse(type);
-  parsed.parameters.charset = charset;
-  return format(parsed);
```
styfle, line 99: Why was this function moved up a couple lines? Can we avoid moving functions to keep the diff smaller?

packages/node/src/serverless-functions/serverless-handler.ts
Kikobeats: Initially, I switched from `node-fetch` to `undici` for catching some invalid requests with non-allowed headers, since undici implementation is much stricter. After that, I thought it will be a good replacement to keep.

packages/node/src/serverless-functions/dynamic-import.js
Kikobeats: Now you mention that, is it possible this code is taking care of that behavior?
https://github.com/vercel/vercel/pull/9745/files#diff-0b5ebdcaf8e4e9cd4e257835ef35598d820da612d518fac10122e7ab6722c9cdL75

Ideally, I want to add a some tests there. What could be an input user code to reproduce that behavior?

BTW, I added this helper just because the TypeScript configuration is overriding import to be equivalent to require

packages/node-bridge/launcher.js
styfle: We need this to support Windows 

- https://github.com/vercel/vercel/pull/6953

packages/node/src/serverless-functions/serverless-handler.ts
styfle: > catching some invalid requests

Which requests? Did you add tests for this?

packages/node/src/serverless-functions/dynamic-import.js
styfle: Yes the `.default` loop in `getListener()` is designed to handle those cases

packages/node/src/serverless-functions/serverless-handler.ts
Ethan-Arrowood: undici fetch is the fetch in Node.js

we move over to it sooner than later, easier our transition will be when we get to Node 18 LTS (while away since we are only moving off of Node 14 now)

packages/node/src/dev-server.ts
```diff
@@ -124,14 +119,14 @@ export async function onDevRequest(
   }
 
   try {
-    const result = await handleEvent(req);
-    res.statusCode = result.statusCode;
-    for (const [key, value] of Object.entries(result.headers)) {
-      if (typeof value !== 'undefined') {
+    const { headers, body, status } = await handleEvent(req);
+    res.statusCode = status;
+    for (const [key, value] of headers as Headers) {
```
Ethan-Arrowood: why is type cast necessary here? Can we update the return type of `handleEvent` to be more correct?

packages/node/src/dev-server.ts
```diff
@@ -124,14 +119,14 @@ export async function onDevRequest(
...
+      if (value !== undefined) {
         res.setHeader(key, value);
       }
     }
-    res.end(Buffer.from(result.body, result.encoding));
+    toToReadable(body).pipe(res);
```
Ethan-Arrowood: Why are we not using: https://nodejs.org/api/stream.html#streamreadablefromwebreadablestream-options here? Is the `@edge-runtime/node-utils` `toToReadable` doing something else?

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -111,7 +99,7 @@ async function compileUserCode(
       wasmAssets,
       nodeCompatBindings: nodeCompatPlugin.bindings,
     };
-  } catch (error) {
+  } catch (error: any) {
```
Ethan-Arrowood: Lets use `@vercel/error-utils` here

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -162,7 +150,7 @@ async function createEdgeRuntime(params?: {
     exitHook(server.close);
 
     return server;
-  } catch (error) {
+  } catch (error: any) {
```
Ethan-Arrowood, line 164: Lets use `@vercel/error-utils` here

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -196,20 +184,23 @@ export async function createEdgeEventHandler(
       redirect: 'manual',
       method: 'post',
       body: await serializeRequest(request),
+      //@ts-expect-error
+      headers: request.headers as HeadersInit,
```
Ethan-Arrowood: Are we sure we can pass Node's `IncomingMessage.Headers` type to Fetch's headers parameter without any kind of transform? I feel like the type casting shouldn't be necessary then

packages/node/src/index.ts
```diff
@@ -588,10 +588,10 @@ export const startDevServer: StartDevServer = async opts => {
     if (pathToTsConfig) {
       try {
         tsConfig = ts.readConfigFile(pathToTsConfig, ts.sys.readFile).config;
-      } catch (err) {
-        if (err.code !== 'ENOENT') {
+      } catch (error: any) {
+        if (error.code !== 'ENOENT') {
```
Ethan-Arrowood: Use `@vercel/error-utils` here

packages/node/src/index.ts
```diff
@@ -674,10 +674,10 @@ async function doTypeCheck(
     const json = JSON.stringify(tsconfig, null, '\t');
     await fsp.mkdir(entrypointCacheDir, { recursive: true });
     await fsp.writeFile(tsconfigPath, json, { flag: 'wx' });
-  } catch (err) {
+  } catch (error: any) {
     // Don't throw if the file already exists
-    if (err.code !== 'EEXIST') {
-      throw err;
+    if (error.code !== 'EEXIST') {
+      throw error;
```
Ethan-Arrowood: Use `@vercel/error-utils` here

packages/node/src/serverless-functions/serverless-handler.ts
```diff
@@ -1,58 +1,74 @@
+import type { VercelProxyResponse } from '../types';
+import type { VercelRequest, VercelResponse } from './helpers';
+import { addHelpers } from './helpers';
+import { create as createTsCompiler } from 'ts-node';
+import { createServer, ServerResponse } from 'http';
+// @ts-expect-error
+import { dynamicImport } from './dynamic-import.js';
```
Ethan-Arrowood, line 4: Can you add a comment what error is expected here? Do we just need to add `allowJs` to the tsconfig?

packages/node/src/serverless-functions/serverless-handler.ts
```diff
@@ -1,58 +1,74 @@
...
-        encoding: 'base64',
-        body: body.toString('base64'),
-      }),
-    };
+  let userCode;
+  if (extension.startsWith('.ts')) {
+    const { compile } = createTsCompiler();
+    const content = fs.readFileSync(entrypointPath, 'utf8');
+    const filename = path.basename(entrypointPath);
+    userCode = eval(compile(content, filename)) as ServerlessFunctionSignature;
```
Ethan-Arrowood: `eval` always worries me. Thought we had eslint rules prevent it too. Can you add a comment here explaining what is being passed to `eval`, and maybe include why it _isnt_ a security concern?

Ethan-Arrowood (changes requested):
looking good!

packages/node/src/serverless-functions/serverless-handler.ts
Kikobeats: for your curiosity: https://github.com/nodejs/undici/blob/eae68072260fd5080985c1b94d1f54e38a3fb789/lib/core/request.js#L291

packages/node/src/dev-server.ts
Kikobeats: It's the same but supporting Node.js v14
https://github.com/vercel/edge-runtime/blob/main/packages/node-utils/src/edge-to-node/stream.ts#L11

packages/node/src/serverless-functions/serverless-handler.ts
Kikobeats: The `eval` is necessary there for turning the user code in a real javascript function

packages/node/src/serverless-functions/serverless-handler.ts
Kikobeats: The error is expected since the imported file is a JavaScript file, so there are no types there. Maybe there is a way for adding types for this case, but I don't know; The `dynamicImport` should be the `typeof import` at this point.

Also, import a JavaScript file is necessary to avoid the TypeScript mutation over `import`

packages/node/src/edge-functions/edge-handler.ts
Kikobeats: Ideally, I want to remove that `@ts-expect-error` assertion. This how the error looks after removing it:

CleanShot 2023-04-04 at 16 21 21@2x

packages/node/src/dev-server.ts
Kikobeats: It's messing with the global types that they don't look compatible

CleanShot 2023-04-04 at 16 23 58@2x

packages/node/package.json
styfle: Nevermind, I see this was changed to a dependency in https://github.com/vercel/vercel/pull/7969

packages/node/src/serverless-functions/helpers.ts
styfle: Lets keep the diff as small as possible

packages/node/src/dev-server.ts
Ethan-Arrowood: It messing with global types because you are using the global type `Headers` in the `./types` file for `VercelProxyResponse`. There is a better solution here. I recommend looking for a better type to use in `VercelProxyResponse`.

packages/node/src/dev-server.ts
Ethan-Arrowood: Okay cool. Good thin 14 is EOL and we can start deprecating some of this stuff soon.

packages/node/src/serverless-functions/serverless-handler.ts
styfle: > Also, import a JavaScript file is necessary to avoid the TypeScript mutation over import

This is because `module: node` instead of `module: nodenext`. 

There is a PR from @cb1kenobi that will likely fix this:

- #9692

packages/node/src/edge-functions/edge-handler.ts
Ethan-Arrowood: Hmm so its what I thought. TypeScript is being correct here. The fetch `headers` option is expecting type `HeadersInit`, and `request.headers` is only `Headers`. You can't just add `as HeadersInit` to get TypeScript to stop complaining. You would need to do `as unknown as HeadersInit` to properly type cast. 

But that doesn't answer my original question of, are you _certain_ that these can be casted like this? The answer can be yes, but you should add a detailed comment why (and then use the correct type cast i shared above).

Ethan-Arrowood: Fixed

packages/node/test/fixtures/37-vercel-og/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "private": true,
-  "type": "module",
```
TooTallNate, line 3: Why did this need to be removed?

Kikobeats: The test fails:

> Rejected to value: [Error: Must use import to load ES Module: /home/runner/work/vercel/vercel/packages/node/test/fixtures/37-vercel-og/probe.js]

I think its legit since the code is not actually using `import`

packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -1,49 +1,22 @@
...
-  const request = new Request(buildUrl(requestDetails), {
-    headers: requestDetails.headers,
-    method: requestDetails.method,
-    body: body,
-  });
-
-  event.request = request;
-
-  let response = await userEdgeHandler(event.request, event);
+  const host = event.request.headers.get('x-forwarded-host');
```
EndangeredMassa: **question (non-blocking):** ‏Is it possible for `host` to be `undefined` in this flow?

This was defaulted to `127.0.0.1` previously: https://github.com/vercel/vercel/pull/9697

packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -1,49 +1,22 @@
...
+  event.request.headers.set('host', host);
+  let response = await userEdgeHandler(
+    new Request(getUrl(event.request.url, host), event.request),
```
EndangeredMassa: **suggestion (blocking):** ‏`getUrl` doesn't appear to handle `x-forwarded-proto`, especially the case where multiple values exist

This was defaulted to `http` previously: https://github.com/vercel/vercel/pull/9697

This was updated to handle multiple values previously: https://github.com/vercel/vercel/pull/9502

packages/node/src/serverless-functions/helpers.ts
```diff
@@ -91,23 +102,43 @@ function redirect(
...
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function json(
+  req: VercelRequest,
+  res: VercelResponse,
+  jsonBody: any
+): VercelResponse {
+  const body = JSON.stringify(jsonBody);
+  // content-type
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
```

EndangeredMassa (changes requested):

packages/node/src/edge-functions/edge-handler-template.js
Kikobeats: I don't think it could be undefined since the value is sets here https://github.com/vercel/vercel/blob/eb736c62b831a6a43c0300bc414ccb793598ea39/packages/node/src/dev-server.ts#L82

packages/node/src/edge-functions/edge-handler-template.js
Kikobeats: I removed that code because I didn't see any test using it more than the unit test around the template. 

What could be an e2e that demonstrates that logic is still necessary?

packages/node/src/serverless-functions/helpers.ts
```diff
@@ -1,21 +1,36 @@
...
+};
+
+export type VercelResponse = ServerResponse & {
+  send: (body: any) => VercelResponse;
+  json: (jsonBody: any) => VercelResponse;
+  status: (statusCode: number) => VercelResponse;
+  redirect: (statusOrUrl: string | number, url?: string) => VercelResponse;
+};
+
+class ApiError extends Error {
```
Ethan-Arrowood, line 22: _Ugh_ we have an APIError already in `packages/cli`. I think this is fine as is, but makes me think about how we'd want to hoist some of these error interfaces.

Ethan-Arrowood (dismissed):
This is a big and complex PR but I read over the test cases closely and feel confident in it. Good work! Thanks!

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -192,24 +191,31 @@ export async function createEdgeEventHandler(
       process.exit(1);
     }
 
-    const response = await fetch(server.url, {
+    const headers = new Headers(request.headers as HeadersInit);
```
cb1kenobi, line 194: Got this error testing an edge function that uses a WASM module using Node 14. Node 16 works.

```
import init, { greet } from "../wasm/pkg";
//@ts-ignore
import wasmModule from "../wasm/pkg/wasm_bg.wasm?module";

export const config = {
	runtime: "experimental-edge",
};

export default async function handler() {
	await init(wasmModule);
	const helloWorld = greet();
	return new Response(helloWorld);
}
```

```
TypeError: undici_1.Headers is not a constructor
    at /Users/chris/code/vercel/packages/node/dist/edge-functions/edge-handler.js:146:25
    at Server.onDevRequest (/Users/chris/code/vercel/packages/node/dist/dev-server.js:85:49)
    at Server.emit (events.js:400:28)
    at Server.emit (domain.js:475:12)
    at parserOnIncoming (_http_server.js:900:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:127:17)%
```

Ethan-Arrowood: I think there was a breaking change for fetch in Undici between those node versions. We may need to keep node-fetch around unless we drop Node 14 support.

packages/node/src/serverless-functions/helpers.ts
Kikobeats: you're right, I can see `APIError` and `responseError` at `packages/cli`, but I'm how I can import one of them here.

Note this class was ported from https://github.com/vercel/vercel/blob/main/packages/node-bridge/helpers.ts#L228

packages/node/src/edge-functions/edge-handler.ts
Kikobeats: That's a good catch; let me bring back node-fetch for now until we dropped node14

cb1kenobi: Confirmed reverting to `node-fetch` has resolved the error.

packages/node/src/edge-functions/edge-handler-template.js
EndangeredMassa: I was having trouble writing an e2e test for it, but it does solve the problem the user had. I think we should restore the functionality and unit tests, at least.

EndangeredMassa: I still think this is a regression

EndangeredMassa (changes requested):

packages/node/test/unit/edge-handler/edge-handler-template.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        )
+      ).toBe('https://somewhere.com/');
+    });
+
+    test('respect `x-forwarded-proto` with no `x-forwarded-host`', async () => {
+      expect(
+        getUrl(
+          'https://127.0.0.1:51126/api/add',
+          new Headers({
+            'x-forwarded-host': '127.0.0.1',
```
EndangeredMassa: **question (non-blocking):** ‏Is the intent of this test that the `x-forwarded-host` matches the url or did you mean to remove this header in this test entirely?

packages/node/test/unit/edge-handler/edge-handler-template.test.ts
```diff
@@ -0,0 +1,122 @@
...
+          entrypointLabel,
+        },
+        {
+          Request,
+          Response,
+        }
+      );
+
+      const text = await response.text();
+      console.log('text', text);
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
```

EndangeredMassa (dismissed):

EndangeredMassa (approved):

EndangeredMassa:
@cb1kenobi is still reviewing this PR

MichaelDeBoey:
@Kikobeats We use `@vercel/node-bridge`'s `createServerWithHelpers` function to test our Remix adapter (see https://github.com/remix-run/remix/blob/main/packages/remix-vercel/__tests__/server-test.ts)

Since the `createServerWithHelpers` function is completely removed now, can you please tell me what I should change to these tests to make them work without it?

TooTallNate:
I don't believe `@remix-run/vercel` actually uses the helper functions, so you should be able to use Node's `http.createServer()` instead.

Aside from that, `@remix-run/vercel` should probably be deprecated. We're recommending users to use `@vercel/remix` instead at this point.

MichaelDeBoey:
@TooTallNate The team is currently still deciding what to do with the Vercel adapter, but deprecation of `@remix-run/vercel` is coming once https://github.com/remix-run/remix/pull/5964 is merged

Would that mean that https://github.com/remix-run/remix/pull/5537 isn't necessary anymore? 🤔

PR #9747, opened 2023-04-04
[next] support `functions` property in appDir

smaeda-ks:
Currently, `functions` property (i.e., `memory` and `maxDuration`) are not supported in appDir routes. This PR attempts to add support.


packages/fs-detectors/src/detect-builders.ts
```diff
@@ -667,7 +667,12 @@ function checkUnusedFunctions(
   // Next.js can use functions only for `src/pages` or `pages`
   if (frontendBuilder && isOfficialRuntime('next', frontendBuilder.use)) {
     for (const fnKey of unusedFunctions.values()) {
-      if (fnKey.startsWith('pages/') || fnKey.startsWith('src/pages')) {
+      if (
+        fnKey.startsWith('pages/') ||
+        fnKey.startsWith('src/pages') ||
+        fnKey.startsWith('app/') ||
+        fnKey.startsWith('src/app/')
```
styfle, line 674: This will need to check for the manifest file first to ensure the app directory is enabled. 

cc @ijjk @timneutkens for which file should be used here

ijjk: It can use this manifest https://github.com/vercel/vercel/blob/main/packages/next/src/server-build.ts#L157-L158

smaeda-ks: Thank you! But I'm not sure how fs-detectors can interact with them

ijjk: The functions config handling should remain in the Next.js builder itself I think https://github.com/vercel/vercel/blob/434c794713608a03fa9e05b489da9eb0d9e2604d/packages/next/src/utils.ts#L1382-L1391

smaeda-ks: What was your main concern? @styfle 
The change here is to avoid throwing the `unused_function` error block right after this.

In my very rough understanding, all things happen in the Next.js build still, as @ijjk shared above, so we don't accidentally produce functions out of `app` dir when `appDir` is disabled, even if we pass through them with these conditions in fs-detectors.

smaeda-ks: Looks like I should resubmit this PR from a direct branch and not from my fork repo to run tests. I may close this one and resubmit if you don't mind. Thanks!

smaeda-ks:
Moving this to https://github.com/vercel/vercel/pull/9811

PR #9749, opened 2023-04-04
[tests] move test files to subdirectory

Kikobeats:
Some unit tests didn't run since Aug 2022 because they are not included as part of the testing command.

This PR introduce `unit` and `integration` folders for testing, so anything that it's placed there will be run as part of the testing command.


packages/node/test/unit/edge-functions/edge-matcher.test.ts
```diff
@@ -63,6 +63,12 @@ describe('middleware matchers', () => {
     expect(buildResult.routes).toEqual([
       {
         src: regExps.join('|'),
+        middlewareRawSrc:
+          matcher === undefined
+            ? []
+            : Array.isArray(matcher)
+            ? matcher
+            : [matcher],
```
EndangeredMassa, line 71: **suggestion (non-blocking):** extract or refactor

I find ternary expressions to be confusing, especially when nested. I suggest we extract this or even write a small helper here like:

```
function toArray(obj) {
  if (!obj) {
    return [];
  }

  if (Array.isArray(obj)) {
    return obj;
  }

  return [obj];
```

then:

```
middlewareRawSrc: toArray(matcher)
```

EndangeredMassa (approved):
Thanks!

styfle (approved):
Perfect, thanks! 🥇

PR #9754, opened 2023-04-05
[all] Use correct "license" field in `package.json`

TooTallNate:
Follow-up to #9696. Ensures that all packages in this repository have the correct "license" field in their respective `package.json` files set to "Apache-2.0", to match the `LICENSE` file at the root of the repository.

The `LICENSE` file the authoritative source, so the "MIT" value was incorrect. This change corrects the previous values and is not considered a relicensing.


packages/node/bench/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "bench",
   "version": "1.0.0",
   "main": "index.js",
-  "license": "MIT",
+  "license": "Apache-2.0",
```
Ethan-Arrowood: Do we need to update this one? Should this be `"private": true` instead?

Ethan-Arrowood (dismissed):

packages/node/bench/package.json
TooTallNate: 6f27e651c15315dc129c95c979757cc4e6cd6c40

Ethan-Arrowood (approved):
Do we need to add `UNLICENSED` to the `@vercel-internals` packages?

EndangeredMassa (approved):

PR #9756, opened 2023-04-06
[next] Skip broken Next.js tests

Ethan-Arrowood:
This PR skips some tests that are breaking because they are running Next.js canary.


styfle (changes requested):
Lets use https://github.com/vercel/vercel/pull/9758 instead

PR #9772, opened 2023-04-07
[tests] combine examples e2e into one chunk

EndangeredMassa:
We don't need 5 chunks of the very fast examples tests. We spend more time in setup for each one. We are also more likely to hit the [usage limit](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) with these.

This PR combines them into one test suite.


TooTallNate (approved):

styfle (commented):
@EndangeredMassa This PR made the tests take over an hour.

https://github.com/vercel/vercel/actions/runs/4653481139/jobs/8238269899

utils/chunk-tests.js
```diff
@@ -16,6 +16,28 @@ const runnersMap = new Map([
...
+    runnerOptions = Object.assign(
+      {},
+      runnerOptions,
+      optionsOverrides[packageName]
+    );
+  }
+  return (
+    runnerOptions || {
+      min: 1,
+      max: 1,
```
styfle, line 35: I suggest using `max: 2` or perhaps `max: 3` here

EndangeredMassa: Ack. I thought I was testing this uncached, but I must have been looking at cached runs. Will update.

EndangeredMassa: PR: https://github.com/vercel/vercel/pull/9792

PR #9774, opened 2023-04-08
[tests] split next-local tests into chunks

EndangeredMassa:
The next-local e2e tests where split between middleware (fast) and all other (slow) tests. This meant there was not a good balance of amount of time spent across these test files, making overall test run times as slow as "all other tests" running.

This PR rebalances the split of tests across 2 files. The new run time is ~9-10m.


packages/next/test/integration/integration-1.test.js
```diff
@@ -2,15 +2,54 @@ process.env.NEXT_TELEMETRY_DISABLED = '1';
 
 const path = require('path');
 const fs = require('fs-extra');
-const builder = require('../../');
+const builder = require('../../dist');
```
styfle: Was this intentional?

packages/next/test/integration/integration-1.test.js
styfle, line 1: Lets make sure this file also sets `process.env.NEXT_TELEMETRY_DISABLED = '1';`

EndangeredMassa: It does, but it was tucked into a single line of github line folding

packages/next/test/integration/integration-1.test.js
EndangeredMassa: It was not. Updated. Thanks!
[`3d0e1ce` (#9774)](https://github.com/vercel/vercel/pull/9774/commits/3d0e1cea2e7fd171aa865411ed1ffb8f10b43d98)

cb1kenobi (approved):
LGTM

PR #9777, opened 2023-04-09
Add 'PUBLIC_' prefix to SvelteKit 1.0 framework

devunt:
With the SvelteKit 1.0 release, public environment variable prefix has been changed to `PUBLIC_` from `VITE_` [1].
Current Vercel build pipeline for SvelteKit exposes environment variable with no prefix [2], making it impossible to access from client-side code while build time.

This PR fixes it.

[1] https://kit.svelte.dev/docs/configuration#env
[2]
image

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1020,6 +1020,7 @@ export const frameworks = [
       'SvelteKit is a framework for building web applications of all sizes.',
     description: 'A SvelteKit app optimized Edge-first.',
     website: 'https://kit.svelte.dev',
+    envPrefix: 'PUBLIC_',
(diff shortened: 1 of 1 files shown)


styfle (changes requested):
I think this was rejected before

https://github.com/vercel/vercel/pull/6821#issuecomment-1501116251

elliott-with-the-longest-name-on-github:
We opted not to do this because any client-side variables have to be shipped with each request, regardless of whether or not they're being used (because of `$env/dynamic`). This unnecessarily bloats bundle size. We do cover this in the docs, along with how to expose the variables you need to the client:

https://kit.svelte.dev/docs/adapter-vercel#environment-variables

PR #9782, opened 2023-04-10
[tests] Update Typescript, Jest, and ts-jest

cb1kenobi:
Updates TypeScript, Jest, and ts-jest to newer versions which have improved support for ESM.

This PR contains a subset of changes from https://github.com/vercel/vercel/pull/9692.


styfle (dismissed):

packages/edge/package.json
```diff
@@ -33,11 +33,14 @@
   "jest": {
```
styfle: In a future PR, I would like to move these all out of package.json and use the `jest.config.js` like we are using in the other packages in this monorepo

cb1kenobi: Done!

styfle (dismissed):
Thanks!

TooTallNate (approved):

packages/cli/test/unit/commands/pull.test.ts
```diff
@@ -98,10 +98,10 @@ describe('pull', () => {
 
       const config = await fs.readJSON(path.join(cwd, '.vercel/project.json'));
       expect(config).toMatchInlineSnapshot(`
-        Object {
```
EndangeredMassa, line 101: **question (non-blocking):** ‏was this a different between versions of jest?

packages/static-build/test/gatsby.test.ts
```diff
@@ -281,7 +281,7 @@ describe('gatsby utilities', () => {
     `);
     expect(gatsbyConfig).toMatchInlineSnapshot(`
       "// This file was generated by @vercel/static-build
-      import userConfig from \\"./gatsby-config.mjs.__vercel_builder_backup__.mjs\\";
```
EndangeredMassa, line 284: **question (non-blocking):** ‏was this also a difference between versions of jest?

EndangeredMassa (approved):

packages/static-build/test/gatsby.test.ts
cb1kenobi: I'm not 100% certain, but I'm guessing it was a breaking change in v29.0.0: https://github.com/facebook/jest/blob/main/CHANGELOG.md#2900

```
[jest-config] [BREAKING] Make snapshotFormat default to escapeString: false and printBasicPrototype: false (https://github.com/facebook/jest/pull/13036)
```

packages/cli/test/unit/commands/pull.test.ts
styfle: I might actually start using snapshots now that the syntax is fixed 😌

styfle (approved):

packages/cli/test/unit/commands/pull.test.ts
cb1kenobi: No kidding!

PR #9784, opened 2023-04-10
[remix] Install `@vercel/remix-run-dev` at build-time instead of using symlink

### LINKED ISSUE #10027: deprecated @npmcli/move-file@1.1.2: This functionality has been moved to @npmcli/fs
Whenever I use vercel plugin in the project. The below warning is emitted. 

`WARN deprecated @npmcli/move-file@1.1.2: This functionality has been moved to @npmcli/fs`

### LINKED ISSUE #10222: deprecated vm2@3.9.19 - The library contains critical security issues
After installing with `pnpm i -g vercel` we get:

`WARN  deprecated vm2@3.9.19: The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to isolated-vm.`

I know it's just a Warning, but it also mentions "Critical Security Issues"

Would be nice to see this fixed.

And also https://github.com/vercel/vercel/issues/10027

TooTallNate:
Instead of including the fork `@remix-run/dev` package as a regular dependency of `@vercel/remix-builder`, install it at build-time by modifying the project's `package.json` file. The reasons for this are:

* Avoids deprecation warnings from a few packages that currently exist on the `@remix-run/dev` package when installing Vercel CLI (those warnings already show up in the build logs anyways, so nothing new there).
* Allows us to install a version as close as possible to the version specified in the user's `package.json` (similar to how we do when auto-injecting the `@vercel/remix` package). This will be especially important once Remix v2 is released, which will have breaking changes compared to v1.

**Note:** `@vercel/remix-run-dev` is still a _dev_ dependency, so that we can use TypeScript types from it, as well as, at runtime, we use the version in the Builder's `package.json` to determine the maximum versions of `@vercel/remix-run-dev` and/or `@vercel/remix` which can safely be installed.

Fixes #10027.
Fixes #10222.


packages/remix/src/prepare-cache.ts
```diff
@@ -1,6 +1,6 @@
 import { glob } from '@vercel/build-utils';
 import { dirname, join, relative } from 'path';
-import { chdirAndReadConfig } from './utils';
+import { _require, chdirAndReadConfig } from './utils';
```
cb1kenobi, line 3: **question (non-blocking):** Why not do `const _require = eval('require');` in each file? I know technically it works, but `require()` is bound to the module for which it was defined. In theory you wouldn't be able to use `_require()` to require a relative path from a source file in a subdirectory since everything would resolve to the file the `eval('require')` happened. Not everyone might know what's actually happening under the hood.‏

packages/remix/src/build.ts
```diff
@@ -63,8 +68,17 @@ const nodeServerSrcPromise = fs.readFile(
   'utf-8'
 );
 
-// This value is the minimum supported version for our fork of Remix
-const minimumSupportRemixVersion = '1.10.0';
+// Minimum supported version of the `@vercel/remix` package
+const VECEL_REMIX_MIN_VERSION = '1.10.0';
```
cb1kenobi:
```suggestion
const VERCEL_REMIX_MIN_VERSION = '1.10.0';
```

packages/remix/src/build.ts
```diff
@@ -155,44 +183,27 @@ export const build: BuildV2 = async ({
...
-          env: envForAddDep,
-          cwd: entrypointFsDirname,
-        }
+      // 1. Users app is on 1.9.0 -> we install the 1.10.0 (minimum) version of `@vercel/remix`.
+      // 2. Users app is on 1.11.0 (a version greater than 1.10.0 and less than the known max
+      //    published version) -> we install the (matching) 1.11.0 version of `@vercel/remix`.
+      // 3. Users app is on something greater than our latest version of the fork -> we install
+      //    the latest known published version of `@vercel/remix`.
+      const vercelRemixVersion = resolveSemverMinMax(
+        VECEL_REMIX_MIN_VERSION,
```
cb1kenobi:
```suggestion
        VERCEL_REMIX_MIN_VERSION,
```

cb1kenobi (changes requested):
Tested both `vc dev` and `vc deploy` and both of my test apps worked. Found a typo.

cb1kenobi (changes requested):
Tested both `vc dev` and `vc deploy` and both of my test apps worked. Found a typo.

packages/remix/src/prepare-cache.ts
TooTallNate: It's only intended to be used with absolute paths (or in the case of this file, with `require.resolve()`). I moved one of the references into a helper module though (`isESM()`).

packages/remix/src/utils.ts
```diff
@@ -369,3 +390,14 @@ async function ensureSymlink(
   await fs.symlink(relativeTarget, symlinkPath);
   debug(`Created symlink for "${pkgName}"`);
 }
+
+export function isESM(path: string): boolean {
```
cb1kenobi, line 394: How do I test this? If I make my `remix.config.js` file ESM and add `"type": "module"` in the `package.json`, `vc dev` blows up with the error:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/chris/code/sites/chrisbarber-remix/build/index.js from /Users/chris/code/sites/chrisbarber-remix/node_modules/.pnpm/@remix-run+serve@1.19.1/node_modules/@remix-run/serve/dist/index.js not supported.
/Users/chris/code/sites/chrisbarber-remix/build/index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename /Users/chris/code/sites/chrisbarber-remix/build/index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/chris/code/sites/chrisbarber-remix/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at /Users/chris/code/sites/chrisbarber-remix/node_modules/.pnpm/@remix-run+serve@1.19.1/node_modules/@remix-run/serve/dist/index.js:43:17
<snip>
```

When I `vc build --prod && bc deploy --prebuilt --prod`, it deploys, but blows up at runtime with a 500:

```
file:///var/task/build/server-build-nodejs-eyJydW50aW1lIjoibm9kZWpzIn0.mjs:12
import build from './build-nodejs-eyJydW50aW1lIjoibm9kZWpzIn0.js';
       ^^^^^
SyntaxError: The requested module './build-nodejs-eyJydW50aW1lIjoibm9kZWpzIn0.js' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
RequestId: ac168cec-2088-4662-9595-755adef1b306 Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

TooTallNate: Remix doesn't support `"type": "module"`. This bit of code is specifically for the `remix.config` file, in which `remix.config.mjs` is supported.

cb1kenobi (approved):

PR #9787, opened 2023-04-11
[gatsby-plugin-vercel-builder] Ignore `assetPrefix` completely

TooTallNate:
Before, we were stripping off only the hostname portion of `assetPrefix`, so any pathname was still being "handled" by the plugin.

Instead, we should be ignoring the `assetPrefix` entirely, so that these two configurations are treated differently:

 * `{ assetPrefix: 'http://example.com', pathPrefix: '/foo' }` - Gatsby project is served from `/foo`
 * `{ assetPrefix: 'http://example.com/foo', pathPrefix: undefined }` - Gatsby project is served from the root `/`

Currently, both configuration behave as the first version listed above. This change makes it so they can behave differently, specifically for some proxy configurations.

In both cases, the generated HTML contains the full URL `http://example.com/foo/*` for JS/CSS/etc. assets


Ethan-Arrowood (approved):

cb1kenobi (approved):
LGTM

PR #9789, opened 2023-04-11
[internals] Refactor `@vercel-internals/types` enums into constants

Ethan-Arrowood:
- Creates new internals package, `@vercel-internals/constants`
- Refactors the `enum`s from `@vercel-internals/types` into `as const` objects, and moves them to `@vercel-intenrals/constants`
- Updates all relevant code within `packages/cli`, including `@vercel-internals/types` imports to be `import type`


internals/types/index.d.ts
```diff
@@ -1,6 +1,12 @@
 import type { BuilderFunctions } from '@vercel/build-utils';
 import type { Readable, Writable } from 'stream';
 import type { Route } from '@vercel/routing-utils';
+import { PROJECT_ENV_TARGET, PROJECT_ENV_TYPE } from '@vercel-internals/constants';
+
+export type ValueOf<T> = T[keyof T];
+
+export type PROJECT_ENV_TARGET_VALUES = ValueOf<PROJECT_ENV_TARGET>;
+export type PROJECT_ENV_TYPE_VALUES = ValueOf<PROJECT_ENV_TYPE>;
```
styfle: Lets use PascalCase for these two types since that is the convention

internals/constants/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  Preview: 'preview',
+  Development: 'development'
+} as const;
+
+export const PROJECT_ENV_TYPE = {
+  Plaintext: 'plain',
+  Secret: 'secret',
+  Encrypted: 'encrypted',
+  System: 'system'
+} as const;
```
styfle: I wonder if we could eliminate enums and instead use types like:

```ts
export type ProjectEnvType = 'plain' | 'secret' | 'encrypted' | 'system'
```

Ethan-Arrowood: Nope because these strings are used as literals throughout the code base

internals/types/index.d.ts
Ethan-Arrowood: Not for constant values is it? I've always thought constants were upper-underscore cased

styfle: Its not a constant. That would be `export const`. I'm talking about the `export type`

Ethan-Arrowood: Ahh okay I think i get it. My brain still sees this as _constant types_ since they are based off of the constants, but I also see your point that they are _types_ and not _constants_. Changing!

internals/constants/index.ts
EndangeredMassa: I don't understand. It also seems to me that we could use the string literals for this.

Ethan-Arrowood: I think this comes down to preference. If we only have the type, then throughout the app wherever you'd previously get to do `ProjectEnvType.Encrypted`, you now have to know to write `'encrypted'`. Luckily, type preview in VSCode does help with that. And typescript will enforce we are getting it write.

I for one enjoy the pattern of `as const` objects to emulate `enum`. I don't know where else we might have this pattern in the code base. IMO more of our _constant_ string values should be defined this way. Making updates is simpler because you just have to modify the singular string rather than everywhere it might be copied too (think tests, similar code, etc.).

Happy to bikeshed this some more. Lots of these refactor PRs are intended to expose some of the codebase's _smell_. More than happy to clean it up as I go instead of carrying it along lol

styfle: > Happy to bikeshed this some more

Lets try the `export type ProjectEnvType = 'plain' | 'secret' | 'encrypted' | 'system'` solution and see if it can be refactored in 30 min or less (timebox it)

styfle: And you can move it from `internal/constants` to `internal/types` 👍

EndangeredMassa: > you just have to modify the singular string rather than everywhere it might be copied too

In theory, yes, but in practice, I've always updated the property name as well because it would be confusing otherwise, which meant I was updated it everywhere regardless.

The string literal version is (in my opinion) easier to write, easier to read, and just as type safe.

I'm fine with anything that's not a typescript enum, though.

Ethan-Arrowood: Refactored to just use types.

packages/cli/src/util/credit-cards.js
styfle, line 1: Is this unused?

Ethan-Arrowood: oh yeah but i didn't mean to delete that in this branch. let me restore since its unrelated

packages/cli/src/util/env/env-target.ts
```diff
@@ -1,22 +1,30 @@
...
+  },
+  {
+    name: 'Preview',
+    value: 'preview',
+  },
+  {
+    name: 'Development',
+    value: 'development',
+  },
+] as const;
```
styfle: Can we avoid duplicating code by using the pattern here:

https://twitter.com/styfle/status/1208787362613989381

I would expect a single source of truth.

Ethan-Arrowood: So we do wanna use constants for these? Sorry I misunderstood what you were asking in the slack thread

packages/cli/src/util/env/env-target.ts
```diff
@@ -1,22 +1,30 @@
...
-
-export function getEnvTargetChoices() {
-  return Object.entries(ProjectEnvTarget).map(([key, value]) => ({
-    name: key,
-    value: value,
-  }));
-}
+export const envTargetChoices = [
+  {
+    name: 'Production',
```
styfle: Is `name` used anywhere? Could we instead use `.toUpperCase()` when needed?

Ethan-Arrowood: Yes, its used in `src/commands/env/add.ts` in a prompt to select which one the user wants to use. If we turn these into an array of strings then we can generate this choices array using map and toUpperCase i guess.

internals/types/index.d.ts
```diff
@@ -1,6 +1,11 @@
 import type { BuilderFunctions } from '@vercel/build-utils';
 import type { Readable, Writable } from 'stream';
 import type { Route } from '@vercel/routing-utils';
+import { PROJECT_ENV_TARGET } from '@vercel-internals/constants';
+
+export type ProjectEnvTargetTuple = typeof PROJECT_ENV_TARGET;
```
TooTallNate: I don't really think we need the Tuple export

Ethan-Arrowood: True I wound up not using it anywhere

packages/cli/package.json
```diff
@@ -84,6 +84,7 @@
     "@types/which": "1.3.2",
     "@types/write-json-file": "2.2.1",
     "@types/yauzl-promise": "2.1.0",
+    "@vercel-internals/constants": "workspace:*",
```
TooTallNate: Drop `workspace:` to be consistent with the others?

```suggestion
    "@vercel-internals/constants": "*",
```

packages/cli/src/util/env/env-target.ts
```diff
@@ -1,22 +1,22 @@
...
+  // Specify `map` returns strings, instead of string constants so `.includes` works
+  return (
+    typeof target === 'undefined' ||
+    envTargetChoices.map<string>(c => c.value).includes(target)
+  );
 }
 
 export function getEnvTargetPlaceholder() {
-  return `<${envTargets().join(' | ')}>`;
+  return `<${envTargetChoices.map<string>(c => c.value).join(' | ')}>`;
```
TooTallNate: **non-blocking:** Is `<string>` necessary here?

TooTallNate (dismissed):
LGTM

EndangeredMassa (dismissed):

packages/cli/src/util/env/env-target.ts
Ethan-Arrowood: Oh good catch. not required here, but needed in the includes above.

internals/constants/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "private": true,
+  "name": "@vercel-internals/constants",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc -p tsconfig.json"
```
cb1kenobi, line 7: **nitpick (non-blocking):** `tsc` will use the closest `tsconfig.json` by default, so the project file is optional.

```suggestion
    "build": "tsc"
```

packages/cli/test/mocks/project.ts
```diff
@@ -1,8 +1,7 @@
 import { client } from './client';
-import {
-  Project,
+import type {
   ProjectEnvTarget,
-  ProjectEnvType,
+  Project,
```
cb1kenobi, line 4: **nitpick (non-blocking):** Sort imports alphabetically?

Ethan-Arrowood: If we want to enforce that we need to enable it using eslint or something.

TooTallNate (approved):

PR #9790, opened 2023-04-11
[fs-detectors] check turbo versions to allow for tags like "latest"

EndangeredMassa:
When a project depends on `turbo` at version `latest`, this code was breaking because `latest` is not a valid range for `semver.intersects`.

### DIFF
--- a/packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
+++ b/packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
@@ -22,6 +22,18 @@ export class MissingBuildTarget extends Error {
   }
 }
 
+function supportsRootCommand(turboSemVer: string | undefined) {
+  if (!turboSemVer) {
+    return false;
+  }
+
+  if (!semver.validRange(turboSemVer)) {
+    return false;
+  }
+
+  return !semver.intersects(turboSemVer, '<1.8.0');
+}
+
 type MonorepoDefaultSettings = {
   buildCommand?: string | null;
   installCommand?: string | null;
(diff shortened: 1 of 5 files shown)


TooTallNate (dismissed):

cb1kenobi (dismissed):

Ethan-Arrowood (approved):

cb1kenobi (approved):

TooTallNate (approved):

packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts
```diff
@@ -22,6 +22,18 @@ export class MissingBuildTarget extends Error {
...
 }
 
+function supportsRootCommand(turboSemVer: string | undefined) {
+  if (!turboSemVer) {
+    return false;
+  }
+
+  if (!semver.validRange(turboSemVer)) {
+    return false;
+  }
```
EndangeredMassa, line 32: @nathanhammond We decided to fix this such that any tags (which are invalid ranges) will NOT use the new command for now to resolve the issue.

We could say tags like `latest` and `canary` should use the new command, but to do that completely, we'd have to check the lockfile. Would you like to follow-up with that fix or leave this as-is?

nathanhammond: The `fs-detectors` setup is completely unaware of lockfiles and that would require significant rearchitecting to support N different versions of lockfiles for M different package managers, as well as possible non-existence.

This was originally opting people in for a conservative, known-safe setup and your change continues in that vein.

In general we want a world where people don't install `turbo` in their project—especially if they're just using `latest` or `canary`.

PR #9792, opened 2023-04-11
[tests] re-chunk example tests

EndangeredMassa:
It turns out that these do take a long time to run when not cached. We do want these chunks.

### DIFF
--- a/utils/chunk-tests.js
+++ b/utils/chunk-tests.js
@@ -16,17 +16,17 @@ const runnersMap = new Map([
   ['test-dev', { min: 1, max: 5, runners: ['ubuntu-latest', 'macos-latest'] }],
 ]);
 
-const optionsOverrides = {
-  examples: { min: 1, max: 1 },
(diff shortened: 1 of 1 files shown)


utils/chunk-tests.js
```diff
@@ -16,17 +16,17 @@ const runnersMap = new Map([
   ['test-dev', { min: 1, max: 5, runners: ['ubuntu-latest', 'macos-latest'] }],
 ]);
 
-const optionsOverrides = {
-  examples: { min: 1, max: 1 },
+const packageOptionsOverrides = {
+  // 'some-package': { min: 1, max: 1 },
```
EndangeredMassa, line 20: I left this in for easy use later. In production code, I wouldn't leave this, but for a testing script, I think it's fine.

I could remove it entirely, though.

styfle (approved):

TooTallNate (approved):

cb1kenobi (approved):

PR #9796, opened 2023-04-12
[cli] Simplify `parseRepoUrl()` and fix edge case

TooTallNate:
Fixes an edge case in `parseRepoUrl()` when there is a `.com` in the repo name. The code was hard to refactor in its previous form so I refactored it to be simpler as well.

### DIFF
--- a/packages/cli/src/util/git/connect-git-provider.ts
+++ b/packages/cli/src/util/git/connect-git-provider.ts
@@ -108,14 +108,15 @@ function getURL(input: string) {
 export function parseRepoUrl(originUrl: string): RepoInfo | null {
   const url = getURL(originUrl);
   if (!url) return null;
+
   const hostParts = url.hostname.split('.');
   if (hostParts.length < 2) return null;
   const provider = hostParts[hostParts.length - 2];
 
-  const pathParts = url.pathname.split('/');
-  if (pathParts.length !== 3) return null;
-  const org = pathParts[1];
-  const repo = pathParts[2].replace(/\.git$/, '');
+  const pathParts = url.pathname.split('/').filter(Boolean);
+  if (pathParts.length !== 2) return null;
+  const org = pathParts[0];
+  const repo = pathParts[1].replace(/\.git$/, '');
   return { url: originUrl, provider, org, repo };
 }
 
--- a/packages/cli/src/util/git/connect-git-provider.ts
+++ b/packages/cli/src/util/git/connect-git-provider.ts
@@ -1,10 +1,11 @@
+import { URL } from 'url';
 import Client from '../client';
-import { Org } from '@vercel-internals/types';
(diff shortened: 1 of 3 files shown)


cb1kenobi:
I ran a bunch of examples through it and all of them worked with the exception of the example git URLs that end with a trailing slash as found in the `git clone` docs: https://git-scm.com/docs/git-clone#_git_urls. Do we need to support the trailling slash? If so, then something like `.replace(/\/+$/, '')` and it should parse fine.

TooTallNate:
Interesting, I've never seen a Git URL like that haha. Added support for it in 1bb21ee30866e1d6ca96467a6bdaa93987470b5f.

packages/cli/src/util/git/connect-git-provider.ts
```diff
@@ -86,40 +87,39 @@ export function formatProvider(type: string): string {
...
-  };
+export function parseRepoUrl(originUrl: string): RepoInfo | null {
+  const url = getURL(originUrl);
+  if (!url) return null;
+
+  const hostParts = url.hostname.split('.');
+  if (hostParts.length < 2) return null;
+  const provider = hostParts[hostParts.length - 2];
+
+  const pathParts = url.pathname.split('/').filter(Boolean);
```
cb1kenobi, line 116: That'll work! Now we can handle crazy stuff like
```
https://username:password@github.com/foo//bar.git/
```

cb1kenobi (approved):

packages/cli/test/unit/util/deploy/create-git-meta.test.ts
```diff
@@ -67,44 +67,59 @@ describe('parseRepoUrl', () => {
...
+    const repoInfo = parseRepoUrl('https://github.com/vercel/vercel.com.git');
+    expect(repoInfo).toBeTruthy();
+    expect(repoInfo?.provider).toEqual('github');
+    expect(repoInfo?.org).toEqual('vercel');
+    expect(repoInfo?.repo).toEqual('vercel.com');
+  });
   it('should parse url with a period in the repo name', () => {
     const repoInfo = parseRepoUrl('https://github.com/vercel/next.js');
-    expect(repoInfo).toBeDefined();
+    expect(repoInfo).toBeTruthy();
```
styfle, line 79: I'm curious, why change from `toBeDefined()` to `toBeTruthy()`?

cb1kenobi: It seems to me that it should have been `toBeTruthy()` from the start since `parseRepoUrl()` previously returned `null` or an object, ya?

TooTallNate: Ya, turns out `toBeDefined()` returns true for `null`, so it was giving a false positive.

PR #9800, opened 2023-04-13
[cli] Add support for Vercel CLI Extensions

TooTallNate:
# Vercel CLI Extensions

Adds a new mechanism to add additional sub-commands to Vercel CLI, inspired by how Git handles sub-commands:

* Extensions are standalone executables that Vercel CLI will spawn as a child process.
* The name of the executable must begin with `vercel-`. For example, to add a sub-command called `vercel example`, there should exist an executable called `vercel-example`.
* The executable can either be a npm package with a `"bin"` entry installed into the local project's workspace, or be globally available in the `$PATH`.
* Extensions can access the [Vercel REST API](https://vercel.com/docs/rest-api), pre-authenticated, by utilizing the `VERCEL_API` env var. Vercel CLI spawns a local HTTP server that adds the `Authorization` header and then proxies to the Vercel REST API.

## Environment Variables

A few environment variables which provide features and context to the extension:

| Name      | Description |
| ----------- | ----------- |
| `VERCEL_API`      | HTTP URL to access the pre-authenticated Vercel API.       |
| `VERCEL_TEAM_ID`  | Provided when the currently selected scope is a Team. |
| `VERCEL_DEBUG` | Provided when the `--debug` flag is used. The command _may_ output additional debug logging. |
| `VERCEL_HELP` | Provided when the `--help` flag is used. The command _should_ print the help output and then end with exit code **2**. |

## Example

```bash
#!/usr/bin/env bash
set -euo pipefail

echo Hi, from a Vercel CLI Extension!
user="$(curl -s "$VERCEL_API/v2/user" | jq -r .user.username)"
echo "Logged in as: $user"
```

Usage:

```
$ vc example
Vercel CLI 28.18.5
Hi, from a Vercel CLI Extension!
Logged in as: tootallnate
```


packages/cli/src/util/extension/exec.ts
```diff
@@ -0,0 +1,91 @@
...
+export async function execExtension(
+  client: Client,
+  name: string,
+  args: string[],
+  cwd: string,
+  apiUrl: string,
+  token?: string
+): Promise<number> {
+  const { debug } = client.output;
+  const extensionCommand = `vercel-${name}`;
```
mrmckeb, line 25: Is it OK if we use `vercel-cli-`? I think that will reduce the risk of collisions in future.

For now I'll use that in our package, but we can remove the `cli-` part if you feel the name-spacing isn't worth the effort. We could also use `vercel-ext-` or `vc-*-`.

mrmckeb (dismissed):
I've tested this with a few commands against our CLI when testing locally, and it works well - including user input - thanks @TooTallNate!

Note that I've only tested this against pnpm.

packages/cli/src/util/extension/proxy.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  const server = createServer((req, res) => {
+    if (token && !req.headers.authorization) {
+      req.headers.authorization = `Bearer ${token}`;
+    }
+    proxy.web(req, res);
+  });
+
+  return server;
+}
```
mrmckeb: I should mention that I didn't test this part sorry, as we're not _yet_ authenticating.

packages/cli/src/util/extension/exec.ts
TooTallNate: Is it really necessary though? I don't really see any risk of a collision. If there's a `vercel-foo` command that exists out in the wild today, it would still work just fine as an extension (just wouldn't be relying on any extension features) or as standalone.

mrmckeb: It would work fine, but may not have been intended to be an extension - which may be an issue in future if we're doing more with extensions, like providing more context, etc.

I don't think it hurts to be safe here, but I understand that it may feel unneeded. I'll defer to you and the CLI team to make the call on this.

mrmckeb (dismissed):
Happy for this to go ahead as discussed!

packages/cli/src/util/extension/exec.ts
```diff
@@ -0,0 +1,90 @@
...
+    extensionPath = await walkParentDirs({
+      base: dirname(baseFile),
+      start: cwd,
+      filename: `node_modules/.bin/${extensionCommand}`,
+    });
+  }
+
+  if (!extensionPath) {
+    // Attempt global `$PATH` lookup
+    extensionPath = which.sync(extensionCommand, { nothrow: true });
```
styfle, line 43: Why sync?

packages/cli/src/index.ts
```diff
@@ -484,96 +488,121 @@ const main = async () => {
...
+          // Fall back to `vc deploy <dir>`
+          targetCommand = subcommand = 'deploy';
+        } else {
+          throw err;
+        }
+      }
     }
 
-    exitCode = await func(client);
+    if (subcommand) {
```
EndangeredMassa, line 516: **suggestion (non-blocking):** ‏can we update the conditions here?

It looks like `targetCommand` can only be set to a known command. If it's undefined, we'll get into the condition above. If no error is thrown, logic will continue and check if `subcommand` is defined here.

But `subcommand` can never be a known command at this point if `targetCommand` has no value.

To make this logic clearer, should this be an `else` case of `if (!targetCommand) {` above?

Or maybe I'm misreading this logic.

EndangeredMassa (dismissed):

packages/cli/src/util/extension/exec.ts
TooTallNate: I didn't see any benefit to using async in this case. There's no other async working happening at this point that this sync call could block on.

Sync should technically be faster anyways since it avoids the thread pool.

packages/cli/src/index.ts
TooTallNate: IIRC it had to be done this way because of the `vc <dir>` (deprecated) syntax. When using that syntax, a CLI extension will be attempted, but if it doesn't exist then the logic needs to fall back to invoking the `deploy` command assuming a dir was provided.

I agree this logic is confusing though. We can add a comment if you think it would help.

TooTallNate: e06f959b7880ab63fb5db675ea4dd49fe49ca3fb

EndangeredMassa (approved):

packages/cli/src/index.ts
```diff
@@ -484,96 +488,123 @@ const main = async () => {
...
+          break;
+        case 'whoami':
+          func = require('./commands/whoami').default;
+          break;
+        default:
+          func = null;
+          break;
+      }
+
+      if (!func || !targetCommand) {
```
cb1kenobi, line 596: **dead code? (non-blocking):** I can't come up with a single practical case where this will be true.

packages/cli/src/util/client.ts
```diff
@@ -65,7 +65,7 @@ export default class Client extends EventEmitter implements Stdio {
   localConfig?: VercelConfig;
   localConfigPath?: string;
   prompt!: inquirer.PromptModule;
-  private requestIdCounter: number;
+  requestIdCounter: number;
```
cb1kenobi, line 68: **question (non-blocking):** Is this change random cleanup?

packages/cli/src/util/extension/exec.ts
cb1kenobi: I'm fine with it being sync, but your statement that "Sync should technically be faster anyways since it avoids the thread pool" has piqued my curiosity. Do you think other areas that we know the code path is guaranteed to be synchronous that we should use the sync versions of functions?

packages/cli/src/index.ts
TooTallNate: I think it was just to make TS happy.

packages/cli/src/util/extension/exec.ts
TooTallNate: For a CLI command, yes.

PR #9802, opened 2023-04-13
[cli] Add `vc deploy --no-wait` and `vc inspect --wait`

cb1kenobi:
There's a need to allow `vc deploy` to not block and not wait for the deployment to finish, yet return the URL.

```
$ vc deploy --no-wait
Vercel CLI 28.18.5
🔍  Inspect: https://vercel.com/chrisbarber/next13fun/L8X4oxp5LGcmy51yVptXwMK7n4wt [1s]
📝  Note: Deployment is still processing...
✅  Production: https://next13fun-kecpx6za2-chrisbarber.vercel.app [1s]
```

Normally this deployment takes around 30 seconds, but the `--no-wait` flag causes the command to exit in less than 2 seconds.

The next step is to add the ability for `vc inspect` to wait as well as specify the maximum time to wait (defaults to 3 minutes, same as `vc rollback`).

```
$ vc inspect --wait --timeout 1m https://next13fun-kecpx6za2-chrisbarber.vercel.app
Vercel CLI 28.18.5
> Fetched deployment "next13fun-ov2r4pvdz-chrisbarber.vercel.app" in chrisbarber [23s]

  General

    id		dpl_9VUuV23EGeoqWf7akEeL8rP1c8cb
    name	next13fun
    status	● Ready
    url		https://next13fun-ov2r4pvdz-chrisbarber.vercel.app
    created	Thu Apr 13 2023 12:25:07 GMT-0500 (Central Daylight Time) [24s ago]
<snip>
```

Also added the ability to pipe the URL into `vc inspect`:

```
echo https://next13fun-ov2r4pvdz-chrisbarber.vercel.app | vc inspect
```

Combined, it allows us to support cool things like:

```
$ vc inspect $(vc deploy --prod --no-wait) --wait
```


packages/cli/src/commands/deploy/index.ts
```diff
@@ -848,12 +853,30 @@ const printDeploymentStatus = async (
...
   indications = indications || [];
   const isProdDeployment = target === 'production';
 
-  if (readyState !== 'READY') {
+  let stillBuilding = false;
+  if (noWait) {
+    if (
+      readyState !== 'READY' &&
+      readyState !== 'CANCELED' &&
+      readyState !== 'ERROR'
```
EndangeredMassa: **suggestion (non-blocking):** ‏group states into named collections

We use this group of terminal states here and a different group of in-progress states [below](https://github.com/vercel/vercel/pull/9802/files#diff-87f688bc2f474b97f3a8a520f3522827bfe7b7981671ba7f965b78ff045971beR144). I like to name these collections of states for clarity and ease of updating in the future.

I suggest putting this near the top of the file:

```
import type { Build } from '@vercel-internals/types';

// ...

const deploymentProgressStates: Build['readyState'][] = ['QUEUED', 'BUILDING', 'INITIALIZING'];
const deploymentTerminalStates: Build['readyState'][] = ['READY', 'CANCELED', 'ERROR'];
```

and using them in both places, like:

```
if (!deploymentTerminalStates.includes(readyState)) { ... }
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -868,8 +891,8 @@ const printDeploymentStatus = async (
     );
   } else {
     // print preview/production url
-    let previewUrl: string;
-    if (Array.isArray(aliasList) && aliasList.length > 0) {
+    let previewUrl: string = deploymentUrl;
```
EndangeredMassa: **suggestion (non-blocking):** revert?‏

```suggestion
    let previewUrl: string;
```

The `deploymentUrl value is always overwritten before `previewUrl` is read, isn't it?

packages/cli/src/commands/inspect.ts
```diff
@@ -34,6 +37,10 @@ const help = () => {
     -d, --debug                    Debug mode [off]
     --no-color                     No color mode [off]
     -S, --scope                    Set a custom scope
+    --timeout=${chalk.bold.underline(
+      'TIME'
+    )}                 Time to wait for deployment completion [3m]
+    --wait                         Blocks until site has deployed
```
EndangeredMassa: **suggestion (soft-blocking):** reword‏

```suggestion
    --wait                         Blocks until deployment completes
```

I think the wording here should more clearly allow for deployments that suceed or fail.

packages/cli/src/commands/inspect.ts
```diff
@@ -67,12 +87,28 @@ export default async function main(client: Client) {
   // extract the first parameter
   let [, deploymentIdOrHost] = argv._;
 
-  if (argv._.length !== 2) {
+  if (!deploymentIdOrHost) {
+    // if the URL is not passed in, check stdin
+    // allows cool stuff like `vc inspect $(vc deploy --prod --no-wait) --wait`
```
EndangeredMassa: **suggestion (non-blocking):** ‏change example

I like allowing for the url to be piped in, but this specific example doesn't seem useful. Is there a reason to do:

```
vc inspect $(vc deploy --prod --no-wait) --wait
```

over

```
vc deploy --prod
```

?

```suggestion
    // allows cool stuff like `echo my-deployment.vercel.app | vc inspect --wait`
```

packages/cli/src/commands/inspect.ts
```diff
@@ -98,12 +134,26 @@ export default async function main(client: Client) {
     `Fetching deployment "${deploymentIdOrHost}" in ${chalk.bold(contextName)}`
   );
 
+  const until = Date.now() + ms(timeout);
```
EndangeredMassa: **suggestion (non-blocking):** ‏`ms` only once

This isn't a performance issue, I'm sure. It just seems a bit weird that we invoke `ms(timeout)` twice here.

I notice that we do a similar thing in `vc rollback`. I'd rather parse the args closer to where we pluck them from the global context and pass along parsed args through the system.

EndangeredMassa (changes requested):
**requirement (blocking):** add tests

This looks great! Can we add a test that shows the `--no-wait` and `--wait` behaviors?‏

packages/cli/src/commands/inspect.ts
cb1kenobi: Fixed!

packages/cli/src/commands/inspect.ts
cb1kenobi: Fixed. I guess I was just excited that it worked. :)

packages/cli/src/commands/inspect.ts
cb1kenobi: Fixed.

packages/cli/src/commands/deploy/index.ts
cb1kenobi: Since the list of states are being referenced in both the `deploy` and `inspect` commands, I thought it made sense to move the state business into a common file. I named it `is-deploying` and it exports a `isDeploying(readyState): boolean` function. I'm definitely open to another solution or different name. What do you think?

packages/cli/src/commands/deploy/index.ts
```diff
@@ -505,6 +508,7 @@ export default async (client: Client): Promise<number> => {
   });
   let deployStamp = stamp();
   let deployment = null;
+  const noWait = !!argv['--no-wait'];
```
TooTallNate, line 511: I think `!!` is unnecessary, no? It should already be a boolean.

cb1kenobi: No problem, I'll remove in the next commit.

Side note, the `vc list` and `vc link` commands do `!!argv['--yes']`, perhaps those should also be updated?

cb1kenobi: Apparently we need the `!!` because the value could be `undefined`:

```
TS2345: Argument of type 'boolean | undefined' is not assignable to parameter of type 'boolean'
```

packages/cli/src/commands/deploy/index.ts
EndangeredMassa: That's reasonable as well

EndangeredMassa (dismissed):

EndangeredMassa (commented):
Updates look good! As soon as we have tests, I think this is good to go.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -86,6 +88,7 @@ export default async (client: Client): Promise<number> => {
       '--meta': [String],
       // This is not an array in favor of matching
       // the config property name.
+      '--no-wait': Boolean,
```
EndangeredMassa: **suggestion (soft-blocking):** ‏move in list

I think the comment above this line was referring to `--regions`. Can we move `--no-wait`?

cb1kenobi: Yeah, I wasn't sure when I added it. No worries, I moved it in the upcoming commit.

packages/cli/test/unit/commands/inspect.test.ts
```diff
@@ -34,4 +48,42 @@ describe('inspect', () => {
...
+
+    const runInspect = async () => {
+      exitCode = await inspect(client);
+      await expect(client.stderr).toOutput(
+        `> Fetched deployment "${deployment.url}" in ${user.username}`
+      );
+    };
+
+    const slowlyDeploy = async () => {
+      await sleep(1234);
```
EndangeredMassa, line 77: **question (non-blocking):** ‏Is there a way to reduce this sleep?

The best sleep is no sleep, but second best is a very small one. Is there a smaller value that will still test that inspect is actually waiting here?

packages/cli/src/commands/inspect.ts
```diff
@@ -98,12 +135,22 @@ export default async function main(client: Client) {
...
-    client,
-    contextName,
-    deploymentIdOrHost
-  );
+  let deployment = await getDeployment(client, contextName, deploymentIdOrHost);
+
+  while (Date.now() < until) {
+    if (!wait || !isDeploying(deployment.readyState)) {
+      break;
+    }
```
EndangeredMassa, line 147: **suggestion (non-blocking):** ‏check `wait` option first

The logic here does what it's supposed to do, but it was confusing to read at first. I was worried that this was always waiting, but a careful readthrough and testing showed that's not true.

I think this would make more sense if it started with a check on the flag that enables this entire behavior, like:

```
if (wait) {
  // ...
}
```

packages/cli/src/commands/inspect.ts
```diff
@@ -98,12 +135,22 @@ export default async function main(client: Client) {
     `Fetching deployment "${deploymentIdOrHost}" in ${chalk.bold(contextName)}`
   );
 
+  const until = Date.now() + timeout;
+  const wait = argv['--wait'];
```
EndangeredMassa, line 139: **question (non-blocking):** ‏Do we want to add another flag to remove the typical verbose output of `vc inspect` that can be paired with `--wait`?

Even on a small project, if I want to use `vc inspect --wait`, I'll get all this output:

```
$  vc inspect --wait https://nextjs-19cgjb33n-uncurated-tests.vercel.app
Vercel CLI 28.18.5-1b347fb
> Fetched deployment "nextjs-19cgjb33n-uncurated-tests.vercel.app" in endangeredmassa [902ms]

  General

    id		dpl_EwDbkYnm35e6ZC2wsPDtGwMNAKBf
    name	nextjs
    status	● Ready
    url		https://nextjs-19cgjb33n-uncurated-tests.vercel.app
    created	Mon Apr 17 2023 10:30:43 GMT-0500 (Central Daylight Time) [32s ago]


  Aliases

    ╶ https://nextjs-endangeredmassa-uncurated-tests.vercel.app


  Builds

    ┌ .        [0ms]
    ├── λ api/[pid] (2.86MB) [sfo1]
    ├── λ api/demo (2.86MB) [sfo1]
    ├── λ _next/data/ZYi_oIKyufen_B3Pcv_4v/blog-fallback-blocking/[id].json (2.87MB) [sfo1]
    ├── λ blog-fallback-blocking/[id] (2.87MB) [sfo1]
    ├── λ _next/data/ZYi_oIKyufen_B3Pcv_4v/blog-fallback-false/[id].json (2.87MB) [sfo1]
    └── 5 output items hidden


  Routes

    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/blog\-fallback\-blocking/(?<id>[^/]+?)(?:/)?.json$              ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/blog-fallback-blocking/[id].json?id=$id
    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/blog\-fallback\-false/(?<id>[^/]+?)(?:/)?.json$                 ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/blog-fallback-false/[id].json?id=$id
    ╶ ^/$                                                                                                ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/index.json
    ╶ ^/$                                                                                                ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/index.json
    ╶ ^/((?!_next/)(?:.*[^/]|.*))/?$                                                                     ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/$1.json
    ╶ ^/((?!_next/)(?:.*[^/]|.*))/?$                                                                     ->      /_next/data/ZYi_oIKyufen_B3Pcv_4v/$1.json
    ╶ ^/blog\-fallback\-blocking/(?<id>[^/]+?)(?:/)?$                                                    ->      /blog-fallback-blocking/[id]?id=$id
    ╶ ^/blog\-fallback\-false/(?<id>[^/]+?)(?:/)?$                                                       ->      /blog-fallback-false/[id]?id=$id
    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/(.*).json$                                                      ->      __next_data_catchall
    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/api/(?<pid>[^/]+?)(?:/)?.json$                                  ->      /api/[pid]?pid=$pid
    ╶ ^/api/(?<pid>[^/]+?)(?:/)?$                                                                        ->      /api/[pid]?pid=$pid
    ╶ ^/_next/__private/trace$                                                                           ->      /404      [404]
    ╶ ^/.*$                                                                                              ->      /404      [404]
    ╶ ^/.*$                                                                                              ->      /500      [500]
    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/(.*).json$                                                      ->      /$1
    ╶ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/(.*).json$                                                      ->      /$1
    ╶ ^/_next/static/(?:[^/]+/pages|pages|chunks|runtime|css|image|media)/.+$                            ->      $0      [404]
    ╶ ^/index(?:/)?$                                                                                     ->      /
    ╶ ^/index(?:/)?$                                                                                     ->      /
    ┌ ^/_next/static/(?:[^/]+/pages|pages|chunks|runtime|css|image|media|ZYi_oIKyufen_B3Pcv_4v)/.+$
    └── cache-control: public,max-age=31536000,immutable
    ╶ ^(?:/(_next/data/[^/]{1,}))?/templates(.*)(.json)?[/#\?]?$
    ╶ ^(?:/(_next/data/[^/]{1,}))?/dashboard(.json)?[/#\?]?$
    ┌ ^/_next/data/ZYi_oIKyufen_B3Pcv_4v/(.*).json$
    └── x-nextjs-matched-path: /$1
    ┌ ^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))/$                                                                                                                                                      [308]
    └── Location: /$1
    ┌ ^/((?!index$).*)$
    └── x-matched-path: /$1
    ╶ ^/404/?$                                                                                                                                                                                  [404]
    ┌ ^/index$
    └── x-matched-path: /
    ╶ ^/500$                                                                                                                                                                                    [500]
    ╶ ^/.*$                                                                                                                                                                                     [404]
    ╶ filesystem╶ resource╶ miss╶ rewrite╶ hit╶ error
```

When maybe we could have `--basic` (or a much better name):

```
$  vc inspect --wait --basic https://nextjs-19cgjb33n-uncurated-tests.vercel.app/
Vercel CLI 28.18.5-1b347fb
> READY
```

which could pipe just the status output "READY", when successful.

EndangeredMassa (approved):
I tested this and it works great. Thanks!

packages/cli/test/unit/commands/inspect.test.ts
cb1kenobi: I chose 1.234s to simulate a deployment and allow the mechanism to poll 4 times, but not 5 times.

packages/cli/src/commands/inspect.ts
cb1kenobi: I like the idea of reducing the output, though it's outside the scope of this PR. I like the idea something like `npm view` where you can specify a column from the package.json to return. Then we could do:

```
$ vc inspect --wait https://nextjs-19cgjb33n-uncurated-tests.vercel.app/ readyState
```

cb1kenobi: Created ticket VCCLI-667.

PR #9804, opened 2023-04-13
[tests] run @vercel/node tests every 5 minutes

dan-stowell:
This [test run](https://github.com/vercel/vercel/actions/runs/4691613947/jobs/8316357810?pr=9745#step:8:199) produced 7 `certificate has expired` FetchErrors. Let's run this test suite continuously to see if we can provoke more errors!


styfle (changes requested):
This takes more than 5 minutes to run. I think it was about 45 minutes or so last I checked.

PR #9805, opened 2023-04-13
[cli] refactor to use create deployment param type

EndangeredMassa:
Small refactor to provide type safety to this invocation of `createDeploy`.

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -280,7 +280,7 @@ export default async (client: Client): Promise<number> => {
   let { org, project, status } = link;
 
   let newProjectName = null;
-  let rootDirectory = project?.rootDirectory || undefined;
+  let rootDirectory = project?.rootDirectory;
   let sourceFilesOutsideRootDirectory: boolean | undefined = true;
 
   if (status === 'not_linked') {
@@ -335,11 +335,10 @@ export default async (client: Client): Promise<number> => {
 
     if (typeof projectOrNewProjectName === 'string') {
       newProjectName = projectOrNewProjectName;
-      rootDirectory =
-        (await inputRootDirectory(client, path, autoConfirm)) ?? undefined;
+      rootDirectory = await inputRootDirectory(client, path, autoConfirm);
     } else {
       project = projectOrNewProjectName;
-      rootDirectory = project.rootDirectory ?? undefined;
+      rootDirectory = project.rootDirectory;
       sourceFilesOutsideRootDirectory = project.sourceFilesOutsideRootDirectory;
 
       // we can already link the project
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -529,7 +529,7 @@ export default async (client: Client): Promise<number> => {
       prebuilt: argv['--prebuilt'],
       rootDirectory,
       quiet,
-      wantsPublic: (argv['--public'] || localConfig.public) ?? false,
+      wantsPublic: Boolean(argv['--public'] || localConfig.public),
       nowConfig: {
         ...localConfig,
         // `images` is allowed in "vercel.json" and processed
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -511,7 +511,9 @@ export default async (client: Client): Promise<number> => {
 
   const name = project ? project.name : newProjectName;
   if (!name) {
-    throw new Error();
+    throw new Error(
+      '`name` not found on project or provided by existing project'
+    );
   }
 
   try {
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -75,7 +75,8 @@ function compactObject<T>(dict: Record<string, T>) {
   const newDict: Record<string, NonNullable<T>> = {};
   for (const [key, value] of Object.entries(dict)) {
     if (typeof value !== 'undefined' && value !== null) {
-      newDict[key] = value;
+      // TS should know this is already NonNullable<T>, but oh well
+      newDict[key] = value as NonNullable<T>;
     }
   }
   return newDict;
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/deploy/index.ts
```diff
@@ -71,6 +71,17 @@ import { parseEnv } from '../../util/parse-env';
 import { errorToString, isErrnoException, isError } from '@vercel/error-utils';
 import { pickOverrides } from '../../util/projects/project-settings';
 
+function compactObject<T>(dict: Record<string, T>) {
+  const newDict: Record<string, NonNullable<T>> = {};
+  for (const [key, value] of Object.entries(dict)) {
+    if (typeof value !== 'undefined' && value !== null) {
+      // TS should know this is already NonNullable<T>, but oh well
+      newDict[key] = value as NonNullable<T>;
```
EndangeredMassa: Does anyone know what I'm missing with TS here?

packages/cli/src/commands/deploy/index.ts
```diff
@@ -508,18 +521,22 @@ export default async (client: Client): Promise<number> => {
...
+      name,
       env: deploymentEnv,
       build: { env: deploymentBuildEnv },
       forceNew: argv['--force'],
       withCache: argv['--with-cache'],
       prebuilt: argv['--prebuilt'],
       rootDirectory,
       quiet,
-      wantsPublic: argv['--public'] || localConfig.public,
-      type: null,
```
EndangeredMassa, line 526: I didn't see `type` being used anywhere.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -469,17 +482,17 @@ export default async (client: Client): Promise<number> => {
...
+  const deploymentEnv = compactObject(
+    Object.assign({}, parseEnv(localConfig.env), parseEnv(argv['--env']))
   );
 
   // Merge build env out of  `build.env` from vercel.json, and `--build-env` args
-  const deploymentBuildEnv = Object.assign(
-    {},
-    parseEnv(localConfig.build && localConfig.build.env),
-    parseEnv(argv['--build-env'])
+  const deploymentBuildEnv = compactObject(
```
cb1kenobi: Instead of using `compactObject()`, what about "fixing" `parseEnv()`? That would avoid the `NonNullable<T>` concern above.

cb1kenobi (changes requested):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -280,7 +292,7 @@ export default async (client: Client): Promise<number> => {
   let { org, project, status } = link;
 
   let newProjectName = null;
-  let rootDirectory = project ? project.rootDirectory : null;
+  let rootDirectory = project?.rootDirectory || undefined;
```
styfle: Are you sure `null` and `undefined` mean the same thing here?

packages/cli/src/commands/deploy/index.ts
```diff
@@ -508,18 +521,22 @@ export default async (client: Client): Promise<number> => {
 
   const localConfigurationOverrides = pickOverrides(localConfig);
 
+  const name = project ? project.name : newProjectName;
+  if (!name) {
+    throw new Error();
```
styfle: This error should have a message

packages/cli/src/commands/deploy/index.ts
EndangeredMassa: It's possible we'd want to differentiate between an env var not existing and it having no value, but not likely. I think it's reasonable to update `parseEnv` instead.

Done: [`2ba676c` (#9805)](https://github.com/vercel/vercel/pull/9805/commits/2ba676ce4cc7875cce2ef4269fd565ea7e00929e)

packages/cli/src/commands/deploy/index.ts
EndangeredMassa: Oh my. That's a miss. Added! [`bfaa098` (#9805)](https://github.com/vercel/vercel/pull/9805/commits/bfaa09879c7ccb421f16f11785105f70776a272b)

packages/cli/src/commands/deploy/index.ts
EndangeredMassa: I'm pretty sure, but to be careful, I switched this back. I updated the types here to allow `null`, which matches the types in `vercel/api`.

[`bc66fa4` (#9805)](https://github.com/vercel/vercel/pull/9805/commits/bc66fa46bcaf600c0c0e1067dc3c987cd67cb3c5) + [`40d8fd6` (#9805)](https://github.com/vercel/vercel/pull/9805/commits/40d8fd6750e630b008e76e346a7f718f4e0c4429)

packages/cli/src/commands/deploy/index.ts
```diff
@@ -512,18 +512,24 @@ export default async (client: Client): Promise<number> => {
...
       env: deploymentEnv,
       build: { env: deploymentBuildEnv },
       forceNew: argv['--force'],
       withCache: argv['--with-cache'],
       prebuilt: argv['--prebuilt'],
       rootDirectory,
       quiet,
-      wantsPublic: argv['--public'] || localConfig.public,
-      type: null,
+      wantsPublic: (argv['--public'] || localConfig.public) ?? false,
```
```suggestion
      wantsPublic: Boolean(argv['--public'] || localConfig.public),
```

Nit pick: this reads better to me since its less logic and more about the type

cb1kenobi: I thought our style guide prohibited `Boolean()`, so I've been doing `!!`.

EndangeredMassa: It looks like the style guide prefers `Boolean()`: https://github.com/vercel/style-guide/blob/main/eslint/rules/best-practice.js#L93-L98

EndangeredMassa:
Fixing type error:

> vercel:build:       TS2322: Type 'string | null | undefined' is not assignable to type 'string | undefined'.

packages/cli/src/commands/deploy/index.ts
cb1kenobi: Ha!

cb1kenobi (approved):
LGTM

styfle (approved):

cb1kenobi (approved):
Bonus points if you want to add a test :)

EndangeredMassa:
> Bonus points if you want to add a test :)

I'll be adding/updating tests in the follow up

PR #9806, opened 2023-04-14
[cli] Remove a bunch of `isCanary` checks

TooTallNate:
We no longer publish versions that include `-canary` in the package.json `version` field, so these are dead code. Just doing a little bit of cleanup.


cb1kenobi (approved):
Wow, looks good! I wondered about this a long while ago.

packages/cli/src/index.ts
```diff
@@ -70,7 +68,7 @@ sourceMap.install();
 Sentry.init({
   dsn: SENTRY_DSN,
   release: `vercel-cli@${pkg.version}`,
-  environment: isCanary ? 'canary' : 'stable',
+  environment: 'stable',
```
styfle, line 71: Perhaps we should have a similar concept for versions auto published from PRs. Basically detect if the version was installed from npm or from a tarball.

TooTallNate: We already get that information from `release` via `pkg.version`. Perhaps we should just remove this tag entirely?

styfle: I think the tag is still useful because I would want to filter out any errors that are not stable versions.

In fact, a lot of these `isCanary` checks could probably by changed to stable checks instead.

packages/cli/src/index.ts
```diff
@@ -165,13 +163,7 @@ const main = async () => {
       )}`
     );
   } else {
-    output.print(
-      `${chalk.grey(
-        `${getTitleName()} CLI ${pkg.version}${
-          isCanary ? ' — https://vercel.com/feedback' : ''
-        }`
```
styfle, line 172: Similarly here, we probably shouldn't print feedback for non-stable versions?

packages/cli/src/index.ts
TooTallNate: SGTM. Let's do it in a follow-up though to keep this one "clean" from behavior changes.

styfle: I actually think its better to do it in this PR because most of the same lines would be changed

TooTallNate: The only place that makes sense to me to do it would be this line and perhaps the feedback one below.

PR #9808, opened 2023-04-14
[cli] Remove `--target` global common arg

TooTallNate:
It doesn't make sense to have `--target` as a global argument, because not all subcommands utilize this flag (actually only `vc deploy` does, and it's even deprecated).

This is technically a breaking change (though risk of someone using this flag on non-`vc deploy` is very unlikely), so we may want to hold off on merging this until we're ready to do a major version bump.


styfle (commented):
I think we discussed adding `--environment` to `vc deploy` to match `vc pull`.

Can we do that change in this PR too?

EndangeredMassa:
I didn't think we came to an agreement on what the flags should be, but maybe I am misremembering.

If we add `--environment` to deploy, deploy would support `--env` and `--environment`, which sounds confusing.

EndangeredMassa (approved):

styfle:
> If we add --environment to deploy, deploy would support --env and --environment, which sounds confusing.

@EndangeredMassa Perhaps we should keep `--target` so it can be used for `vc pull` too?

cb1kenobi (approved):

styfle (approved):
Approved since I just realized that `vc deploy --target` will still work 👍 

https://github.com/vercel/vercel/blob/2081f10b221fee2c7172ebf1e9f75c3a58de8a2e/packages/cli/src/commands/deploy/index.ts#L108

PR #9811, opened 2023-04-14
[next] support functions property in appDir

smaeda-ks:
Currently, `functions` property (i.e., `memory` and `maxDuration`) are not supported in appDir routes. This PR attempts to add support.


packages/next/src/utils.ts
```diff
@@ -1242,30 +1242,36 @@ async function getSourceFilePathFromPage({
...
+      fsPath = path.join(workPath, 'src', pageType, page);
+    }
 
-  for (const ext of extensionsToTry) {
-    fsPath = `${extensionless}.${ext}`;
     if (fs.existsSync(fsPath)) {
       return path.relative(workPath, fsPath);
     }
-  }
+    const extensionless = fsPath.slice(0, -3); // remove ".js"
```
cb1kenobi: `slice()` scares me. I know you just moved this line, but I'd feel more comfortable if this replacement was more explicit:

```suggestion
    const extensionless = fsPath.replace(path.extname(fsPath), '');
```

packages/next/test/fixtures/00-app-dir/vercel.json
```diff
@@ -2,7 +2,19 @@
   "builds": [
     {
       "src": "package.json",
-      "use": "@vercel/next"
+      "use": "@vercel/next",
+      "config": {
+        "functions": {
+          "app/api/**/*": {
```
cb1kenobi, line 8: **question (non-blocking):** I'm just getting up to speed with the app directory and I know this is just a test, but I have a question about using this in my project. With `route.[tj]s`, you can have the API endpoint anywhere right? So I would want to do `"app/**/route.{ts,js}"` (or whatever the glob syntax is), right?

cb1kenobi (dismissed):

packages/next/test/fixtures/00-app-dir/vercel.json
smaeda-ks: @cb1kenobi yes, that should work too! As you said, wanted to make it a bit more explicit because of the test readability 👍

packages/next/src/utils.ts
smaeda-ks: 100%! Thanks for catching this!

ijjk (approved):
Looks good, thanks for the PR!

TooTallNate (approved):

PR #9813, opened 2023-04-14
[cli] Support pnpm in `getUpdateCommand()`

TooTallNate:
We're currently showing bad instructions when CLI was installed via pnpm:

image

This adds support for it by leveraging our existing `scanParentDirs()` logic from build-utils.

### DIFF
--- a/packages/cli/src/util/get-update-command.ts
+++ b/packages/cli/src/util/get-update-command.ts
@@ -84,7 +84,14 @@ export default async function getUpdateCommand(): Promise<string> {
   const tag = isCanary() ? 'canary' : 'latest';
   const pkgAndVersion = `${getPkgName()}@${tag}`;
 
-  const { cliType } = await scanParentDirs(dirname(dirname(process.argv[1])));
+  const entrypoint = await realpath(process.argv[1]);
+  let { cliType, lockfilePath } = await scanParentDirs(
+    dirname(dirname(entrypoint))
+  );
+  if (!lockfilePath) {
+    // Global installs for npm do not have a lockfile
+    cliType = 'npm';
+  }
   const yarn = cliType === 'yarn';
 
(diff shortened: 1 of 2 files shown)


cb1kenobi:
FWIW, this may solve the one scenario, but it doesn't solve mine local machine where `process.argv[1]` equals `"/usr/loca/bin/vercel"`.
image
Interestingly, I'm not sure why it's selecting `yarn` over `npm`. The code before suggested `npm`:
image

TooTallNate:
@cb1kenobi Try again with latest.

cb1kenobi:
Still no global flag, but now it's suggesting `pnpm`!
image

cb1kenobi (dismissed):

cb1kenobi:
Looks good! Tested npm, pnpm, yarn v1 + nvm on macOS as well as npm and yarn v1 + nvm on Windows.

packages/cli/src/util/get-update-command.ts
```diff
@@ -100,13 +82,24 @@ async function isGlobal() {
...
+  let { cliType, lockfilePath } = await scanParentDirs(
+    dirname(dirname(entrypoint))
+  );
+  if (!lockfilePath) {
+    // Global installs for npm do not have a lockfile
+    cliType = 'npm';
+  }
+  const yarn = cliType === 'yarn';
+
+  let install = yarn ? 'add' : 'i';
```
styfle, line 95: I think `pnpm add` should be used too, right?

https://pnpm.io/cli/add

cb1kenobi (approved):

packages/cli/src/util/get-update-command.ts
TooTallNate: Both work, but ya from the docs it looks like `add` would be more appropriate for this use-case.

PR #9823, opened 2023-04-17
[node] Move `path-to-regexp` and `ts-morph` from dev deps to deps

### LINKED ISSUE #9735: Error: Cannot find module 'path-to-regexp'
Looks like `@vercel/node` depends on `path-to-regexp` at runtime, but that dependency is [specified as a dev dependency](https://github.com/vercel/vercel/blob/main/packages/node/package.json#L62) for `@vercel/node`.

```
Require stack:
- C:\app\node_modules\@vercel\node\dist\utils.js     
- C:\app\node_modules\@vercel\node\dist\dev-server.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1075:15)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (C:\app\node_modules\@cspotcode\source-map-support\source-map-support.js:811:30)
    at Function.Module._load (node:internal/modules/cjs/loader:920:27)
    at Module.require (node:internal/modules/cjs/loader:1141:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at Object.<anonymous> (C:\app\node_modules\@vercel\node\dist\utils.js:5:26)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Object.require.extensions.<computed> [as .js] (C:\app\node_modules\ts-node\src\index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:1117:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'C:\\app\\node_modules\\@vercel\\node\\dist\\utils.js',
    'C:\\app\\node_modules\\@vercel\\node\\dist\\dev-server.js'
  ]
}
```

cb1kenobi:
`@vercel/node` uses `ncc` to bundle the `index.js` spawns the dev server entrypoint which does not bundle the required dependencies, thus `path-to-regexp` and `ts-morph` must be moved from dev dependencies to dependencies.

Fixes #9735.


cb1kenobi:
We have verification that this fixes the original issue: https://github.com/vercel/vercel/issues/9735#issuecomment-1512282827.

styfle (commented):
You also need to update the ncc script to mark these as `--external`

Also, can you add a test?

Ethan-Arrowood (dismissed):

cb1kenobi:
> You also need to update the ncc script to mark these as `--external`
> 
> Also, can you add a test?

I've excluded `path-to-regexp` and `ts-morph` from the bundle.

I don't think we can reasonably test this as the `path-to-regexp` dependency is hoisted by `@vercel/remix-builder` and `@vercel/routing-utils`. I tried to reproduce with a project that depends on `vercel` (e.g. not global), but I couldn't replicate the problematic `node_modules` tree. I tried with npm 8.

packages/node/build.js
```diff
@@ -34,6 +34,10 @@ async function main() {
       '@vercel/build-utils',
       '-e',
       'typescript',
+      '-e',
+      'ts-morph',
+      '-e',
+      'path-to-regexp',
       '-o',
```
styfle: Now that I look at it, this needs to excluded all `dependencies`.

Lets do the same loop we use in the cli package here:

https://github.com/vercel/vercel/blob/b52df7a533160f605f5584e2e5401e3701297ade/packages/cli/scripts/build.ts#L46-L54

cb1kenobi: Done!

styfle (approved):

PR #9836, opened 2023-04-20
[cli] implement `vc deploy --prod --skip-domain`

EndangeredMassa:
Adds the `--skip-domain` option to be used like `vc deploy --prod --skip-domain`. This passes along the `autoAliasCustomDomains: false` flag to the API so that the deployment will not auto alias domains for this deployment.

This PR also moves the `commands/deploy/archive` fixture to `commands/deploy/static` to better describe what the fixture is, not how it is sometimes used in a test.

---

Depends on: https://github.com/vercel/api/pull/18730
Card: https://linear.app/vercel/issue/VCCLI-276/cli-allow-creation-of-production-deployment-without-auto-alias


packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -200,6 +200,7 @@ export default async function setupAndLink(
           ...localConfigurationOverrides,
           sourceFilesOutsideRootDirectory,
         },
+        autoPromote: true,
```
cb1kenobi: `autoPromote` defaults to `true`. I'm fine being explicit, it's just not required.

cb1kenobi (dismissed):

packages/cli/src/util/link/setup-and-link.ts
EndangeredMassa: I made it required in the `CreateOptions` interface because I wanted to be explicit, yeah

packages/cli/src/commands/deploy/index.ts
```diff
@@ -93,6 +93,7 @@ export default async (client: Client): Promise<number> => {
       '--prod': Boolean,
       '--archive': String,
       '--no-wait': Boolean,
+      '--no-alias': Boolean,
```
styfle: Is "alias" the term we want to use here? I think it should match the name we call it in the dashboard and I dont think we have used "alias" in many years.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -520,6 +521,8 @@ export default async (client: Client): Promise<number> => {
   }
 
   try {
+    const noAlias = argv['--no-alias'] ?? false;
```
```suggestion
    const autoPromote = !argv['--no-alias']
```

How about this instead?

packages/cli/src/commands/deploy/index.ts
EndangeredMassa: I'm wondering the same. I am working on a discussion topic in Slack about this.

EndangeredMassa:
Going to test this against an API change in staging next

cb1kenobi (approved):

EndangeredMassa:
The plan is to merge this without help output for now. We'll add that in after it's ready for GA with corresponding changes to the dashboard.

PR #3864, opened 2020-03-03
[examples] Remove examples without detectors

AndyBitz:
This removes all examples that do not have detectors.

[PRODUCT-1307]

[PRODUCT-1307]: https://zeit.atlassian.net/browse/PRODUCT-1307

PR #9839, opened 2023-04-21
[cli][fs-detectors][frameworks] Add Storybook support

cb1kenobi:
Adds support for Storybook projects. Normally, the fs detector will see the project as a Next.js (or something else) and this can cause issues. For example, if the project has a `middleware.ts` file, a corresponding function will be generated and interfere with the Storybook site.

The general idea is a user would create two projects associated to the same repo. The first one would be for the main site. The second one is for the storybook site. In the second site's project settings, the user would manually set the framework to "Storybook" as well as override the build command (e.g. `npm run storybook-build`). The user can also manually set the framework in the project settings (or `.vercel/project.json`) to `"storybook"`.

The Storybook framework is configured to treat the project as a static build which will exclude middleware, functions, etc.

Ref: https://github.com/orgs/vercel/discussions/1321


packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,33 @@ export const frameworks = [
...
+  {
+    name: 'Storybook',
+    slug: 'storybook',
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/storybook.svg',
+    tagline: 'Frontend workshop for UI development',
+    description:
+      'Storybook is a frontend workshop for building UI components and pages in isolation.',
+    website: 'https://storybook.js.org/',
+    ignoreRuntimes: ['@vercel/next', '@vercel/node'],
+    useRuntime: { src: 'package.json', use: '@vercel/static-build' },
```
styfle: You can remove `useRuntime` since that is implied

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,33 @@ export const frameworks = [
...
     ],
   },
+  {
+    name: 'Storybook',
+    slug: 'storybook',
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/storybook.svg',
+    tagline: 'Frontend workshop for UI development',
+    description:
+      'Storybook is a frontend workshop for building UI components and pages in isolation.',
+    website: 'https://storybook.js.org/',
```
```suggestion
    website: 'https://storybook.js.org',
```

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,33 @@ export const frameworks = [
...
+    ignoreRuntimes: ['@vercel/next', '@vercel/node'],
+    useRuntime: { src: 'package.json', use: '@vercel/static-build' },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install`, `pnpm install`, or `npm install`',
+      },
+      buildCommand: {
+        placeholder: '`npm run build-storybook`',
```
styfle: placeholder should have `or` (see others)

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,33 @@ export const frameworks = [
...
+        value: 'npm run build-storybook',
```
styfle: value should not use `npm run` (see others)

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,33 @@ export const frameworks = [
...
+      },
+      devCommand: {
+        placeholder: '`npm run storybook`',
+        value: `npm run storybook`,
+      },
+      outputDirectory: {
+        value: null,
```
styfle: I believe the output dir is required

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -531,6 +517,20 @@ function detectFrontBuilder(
...
+  ) {
+    const deps: PackageJson['dependencies'] = {
+      ...pkg.dependencies,
+      ...pkg.devDependencies,
+    };
+
+    if (deps['next']) {
+      framework = 'nextjs';
+    }
+  }
```
styfle: Why was this moved?

packages/frameworks/src/frameworks.ts
cb1kenobi: Done!

packages/frameworks/src/frameworks.ts
cb1kenobi: Done!

packages/frameworks/src/frameworks.ts
cb1kenobi: Done!

packages/frameworks/src/frameworks.ts
cb1kenobi: Done!

packages/fs-detectors/src/detect-builders.ts
cb1kenobi: Reverted the change, but I think the code may be flawed. `createdAt` defaults to `0`, so if `framework` is valid, but `createdAt` is `undefined`, it'll force the framework to be `nextjs`.

packages/frameworks/src/frameworks.ts
styfle: @cb1kenobi I still see `npm run build-storybook`

It should be `storybook build`

Also now that I think about it, the placeholder would be `npm run build` since thats the default.

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,32 @@ export const frameworks = [
...
+      installCommand: {
+        placeholder: '`yarn install`, `pnpm install`, or `npm install`',
+      },
+      buildCommand: {
+        placeholder:
+          '`npm run build-storybook`, `yarn build-storybook`, or `pnpm build-storybook`',
+        value: 'npm run build-storybook',
+      },
+      devCommand: {
+        value: `start-storybook -p 6006`,
```
styfle: This should use a PORT env var (see other frameworks)

cb1kenobi: Done!

packages/frameworks/src/frameworks.ts
cb1kenobi: Whoops, done!

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,30 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'Storybook',
```
EndangeredMassa, line 1936: To be clear: This means that Storybook is never auto-detected, because there are no detectors. I think this is what we want for now, but it's worth calling out.

EndangeredMassa (dismissed):

packages/frameworks/src/frameworks.ts
cb1kenobi: Correct.

Interestingly enough, we used to have Storybook in the list of frameworks, but it was removed (https://github.com/vercel/vercel/pull/3864) because it didn't have a detector.

styfle: @cb1kenobi Lets add a comment here to explain that is intentional so it doesn't get removed in the future

styfle: I'm also curious - why not have any detectors?

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,30 @@ export const frameworks = [
...
+    slug: 'storybook',
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/storybook.svg',
+    tagline: 'Frontend workshop for UI development',
```
styfle, line 1939: Are you missing the `demo` prop?

packages/frameworks/src/frameworks.ts
cb1kenobi: I added the detector. Originally I didn't think it mattered since according to https://storybook.js.org/docs/react/get-started/install:

> Storybook needs to be installed into a project that is already set up with a framework.

In theory, we will detect the site as Next.js or whatever before we would detect it as Storybook.

However, maybe the user is doing something special like the UI components are in an external package or something which would cause the project to be detected as Storybook.

packages/frameworks/src/frameworks.ts
cb1kenobi: `demo` is not required and I didn't prepare a demo. Can that be added in a future PR?

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,37 @@ export const frameworks = [
...
+    description:
+      'Storybook is a frontend workshop for building UI components and pages in isolation.',
+    website: 'https://storybook.js.org',
+    ignoreRuntimes: ['@vercel/next', '@vercel/node'],
+    detectors: {
+      every: [
+        {
+          matchPackage: 'storybook/.+',
```
```suggestion
          matchPackage: '@storybook/cli',
```

cb1kenobi: This is not correct. The `storybook` package is a wrapper for `@storybook/cli`. Users don't install the `storybook` package as a dependency. Instead they run `pnpm dlx storybook@latest init` which added project specific packages like react vs preact vs vue.

styfle: @cb1kenobi So in that case, it should check for `@storybook/cli`, right?

cb1kenobi: OK, not sure what's up with my example, but I don't have `storybook` in the `package.json`, but it's in `node_modules`. I created a new app and now `storybook` is in the `package.json`. Weirdness. PR updated to detect `storybook`. Good catch.

EndangeredMassa (dismissed):

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,37 @@ export const frameworks = [
...
+    },
+    settings: {
+      installCommand: {
+        placeholder: '`yarn install`, `pnpm install`, or `npm install`',
+      },
+      buildCommand: {
+        value: 'storybook build',
+      },
+      devCommand: {
+        value: `start-storybook -p $PORT`,
```
```suggestion
        value: `storybook dev -p $PORT`,
```

Where did you get `start-storybook` from? 

The `examples/storybook/package.json` seems to indicate this is `storybook dev`

packages/frameworks/src/frameworks.ts
```diff
@@ -1932,6 +1932,37 @@ export const frameworks = [
...
+      },
+      outputDirectory: {
+        value: 'storybook-static',
```
styfle, line 1962: Is the output dir configurable? We might need to read the config file to handle all cases

packages/frameworks/src/frameworks.ts
cb1kenobi: Good question. Copy and paste error.

packages/frameworks/src/frameworks.ts
cb1kenobi: There is a `.storybook/main.ts` where you tell Storybook which framework the project is using, where to find your stories, etc. It doesn't appear that the output directory is one of the configurable settings.

The `storybook build` command supports a `--output-dir [path]` option. The default value is `"storybook-static"`: https://github.com/storybookjs/storybook/blob/e7817e82757618992f930f6157ddb68fc7541260/code/lib/cli/src/build.ts#L9.

EndangeredMassa (approved):

bencao:
Got the error "The provided framework must match one of the supported options." when I was trying to save a project's framework setting to "Storybook".

cb1kenobi:
~~@bencao try `"storybook"` all lowercase.~~

ovflowd:
Hey there 👋 is this released by now?

ovflowd:
Oddly getting this here 👀 should I open an issue? Payload seems to be with lowercase "s"

image

image

cb1kenobi:
@bencao, @ovflowd I just checked and the Storybook support added via this pull request has not released yet. I'm not quite sure why it's showing up in the dropdown. In any case, this will ship in the next CLI release.

ovflowd:
Thanks @cb1kenobi any eta for that release? (When it usually happens 👀)

PR #9846, opened 2023-04-23
[examples] Update `remix` template

MichaelDeBoey:
Bring in line with latest template


TooTallNate (approved):

examples/remix/app/root.tsx
```diff
@@ -1,4 +1,5 @@
-import type { MetaFunction } from "@vercel/remix";
+import { cssBundleHref } from "@remix-run/css-bundle";
+import type { LinksFunction } from "@remix-run/node";
```
TooTallNate: All instances of `@remix-run/node` should be replaced with `@vercel/remix`.

examples/remix/app/root.tsx
```diff
@@ -1,4 +1,5 @@
-import type { MetaFunction } from "@vercel/remix";
+import { cssBundleHref } from "@remix-run/css-bundle";
```
TooTallNate, line 1: I tried to build this and got an error:

```
✘ [ERROR] Could not resolve "@remix-run/css-bundle"

    app/root.tsx:1:30:
      1 │ import { cssBundleHref } from "@remix-run/css-bundle";
        ╵                               ~~~~~~~~~~~~~~~~~~~~~~~
```

Is there a future flag we need to enable or something?

TooTallNate (changes requested):

examples/remix/app/root.tsx
MichaelDeBoey: I forgot to include the dependency 🙈

examples/remix/app/root.tsx
MichaelDeBoey: Is there a specific reason for this?

Just trying to understand the reasoning behind it

MichaelDeBoey:
@TooTallNate Is there anything I still need to do in order to get this one merged?

MichaelDeBoey:
@leerob @TooTallNate Is there anything I still need to do in order to get this one merged?

examples/remix/package.json
```diff
@@ -3,24 +3,28 @@
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
-    "dev": "remix dev"
+    "dev": "remix dev",
```
leerob, line 6: Should we use the new dev server?

leerob (approved):
LGTM!

TooTallNate (approved):

examples/remix/app/root.tsx
TooTallNate: This package ensures compatibility with both Vercel Serverless and Edge functions. You can read the docs [here](https://vercel.com/docs/frameworks/remix#@vercel/remix).

examples/remix/package.json
MichaelDeBoey: @leerob As you can see in https://github.com/remix-run/remix/pull/6650, the only thing that changed for `v2_dev` is enabling the future flag

MichaelDeBoey:
@leerob @TooTallNate It seems like the deploy isn't authorized and the `Enforce Changeset` lint is also failing
Anything I can do to make them work?

I'd love to get this one merged, so we can start pointing to the official template instead in the CLI

TooTallNate:
No changeset necessary since this only updates an example. I'm just doing some final verification and will get this merged ASAP.

TooTallNate:
Thank you @MichaelDeBoey!

PR #9851, opened 2023-04-24
[node] validate function runtime and throw when "nodejs"

EndangeredMassa:
We want to keep the currently unused `runtime: "nodejs"` value clear for future use. This PR effectively makes any unrecognized value fail the build. If the value is "nodejs" specifically, a direct error is thrown suggesting they remove the value.

Only a handful of projects set this value currently. We're reaching out to them to let them know that this value is not currently supported and that they should remove it. Then, we should deploy this PR so that anyone else using this value will know to remove it.


packages/node/src/utils.ts
```diff
@@ -81,6 +81,31 @@ export function isEdgeRuntime(runtime?: string): runtime is EdgeRuntimes {
   );
 }
 
+const ALLOWED_RUNTIMES: string[] = [...Object.values(EdgeRuntimes)];
```
```suggestion
const ALLOWED_RUNTIMES: string[] = Object.values(EdgeRuntimes);
```

Is the spread necessary?

packages/node/src/index.ts
```diff
@@ -409,20 +407,12 @@ export const build: BuildV3 = async ({
...
-        `Detected unused static config runtime "nodejs" in "${entrypointPath}"`
-      );
-    }
-    isEdgeFunction = isEdgeRuntime(staticConfig.runtime);
+
+  const runtime = staticConfig?.runtime;
+  validateConfiguredRuntime(runtime, entrypoint);
+
+  if (runtime) {
+    isEdgeFunction = isEdgeRuntime(runtime);
```
styfle, line 415: Seems like the `validateConfiguredRuntime()` should be inside the `if(runtime) {  }` statement and that way you dont need to validate an undefined runtime.

packages/cli/test/dev/integration-1.test.ts
```diff
@@ -207,8 +207,8 @@ test('[vercel dev] should handle config errors thrown in edge functions', async
     expect(await res.text()).toMatch(
       /<strong>500<\/strong>: INTERNAL_SERVER_ERROR/g
     );
-    expect(stderr).toMatch(
-      /Invalid function runtime "invalid-runtime-value" for "api\/edge-error-config.js". Valid runtimes are: \["edge","experimental-edge"\]/g
+    expect(stderr).toContain(
+      'unsupported "runtime" value in `config`: "invalid-runtime-value" (must be one of: ["edge","experimental-edge"]). Learn more: https://vercel.link/creating-edge-functions'
```
styfle: Looks like the error message no longer tells you which file is problematic

styfle: Oh the file name is at the start of the error message so maybe check for that too?

packages/node/src/index.ts
EndangeredMassa: It can go either way. I liked the clarity of the code as written because it deals with validating `runtime` next to creating it, then moves on. `undefined` is a valid `runtime`.

packages/cli/test/dev/integration-1.test.ts
EndangeredMassa: Done! [`f240ba5` (#9851)](https://github.com/vercel/vercel/pull/9851/commits/f240ba56d080f340bfd2ef7d8d6d902656a0a02f)

cb1kenobi (dismissed):
LGTM

styfle (dismissed):
Looks good once we confirm 0% (or very low percentage) of users rely on this

EndangeredMassa:
There have been 2 external users affected by this in the past 60 days. We're reaching out to them before shipping this PR.

Kikobeats (dismissed):

styfle (approved):

cb1kenobi (approved):

PR #9858, opened 2023-04-25
[tests] install root typescript version `4.9.5` and standardize on that version

EndangeredMassa:
This add a root-level `typescript` version that matches the one used throughout for VS Code (and other IDEs) to use when browsing the code. After this PR merges, you will be able to set VS Code's TypeScript version to match the project's version.

This will remove issues where the IDE says TypeScript compilation is good to go, but `pnpm build` disagrees.

If there's a better way to allow this, please suggest it!

---

To enable:

Screenshot 2023-04-25 at 4 28 22 PM

Then:

Screenshot 2023-04-25 at 4 29 20 PM


styfle (dismissed):

cb1kenobi (dismissed):

cb1kenobi (dismissed):

EndangeredMassa:
I removed `typescript` dependencies where possible and updated the rest to `4.9.5`, which is the version coming in the https://github.com/vercel/vercel/pull/9692/ PR.

I made some small changes to make the code pass TypeScript transpilation at that version.

packages/node/test/fixtures/18-nested-tsconfig/functions/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "main": "index.js",
   "devDependencies": {
-    "@types/node": "12.6.1"
+    "@types/node": "12.6.1",
+    "typescript": "3.7.7"
```
EndangeredMassa, line 7: This was necessary to get this fixture to use version `3.7.7`. Not sure how this was working before, exactly.

cb1kenobi: Maybe the test runner is using `npm` and it's resolving the `typescript` package installed from the parent directory `package.json`?

Also note that this test in particular I noticed that the ES5 transpilation for string concatenation changed and I wasn't entirely sure why, but it prompted me to update the probe: https://github.com/vercel/vercel/pull/9692/files#diff-08299ba5c1fe93eaa22b12bbb372e3ddf2e181a524bec1d94811a8f9c1225a5bL10.

EndangeredMassa:
I had to revert the TS version standardization for `packages/go`, specifically. The newer version was changing execution directory to be `./dist`. We can look at upgrading that separately.

TooTallNate (approved):

TooTallNate:
Nice work!

packages/node/test/fixtures/18-nested-tsconfig/probes.json
```diff
@@ -7,7 +7,7 @@
     },
     {
       "path": "/functions/es5.ts",
-      "mustContain": "concat"
+      "mustContain": "+"
```
EndangeredMassa, line 10: @cb1kenobi I was able to turn this back with the change here: https://github.com/vercel/vercel/pull/9858/files#r1178091742

packages/python/package.json
```diff
@@ -25,7 +25,6 @@
     "@types/node": "14.18.33",
     "@vercel/build-utils": "6.7.1",
     "@vercel/ncc": "0.24.0",
-    "execa": "^1.0.0",
-    "typescript": "4.3.4"
+    "execa": "^1.0.0"
```
cb1kenobi, line 28: **non-blocking:** We should consider updating to `execa@7.x` (currently 7.1.1). v1.x is 5 years old. Plus we should pin the version.

cb1kenobi (approved):

Kikobeats (approved):

PR #9860, opened 2023-04-26
[node] expose `WebSocket` constructor in Edge functions

Kikobeats:
This PR exposes `WebSocket` to be used for Edge Runtime for `vc dev`.

Ideally, that's a thing we should to handle inside `edge-runtime` dependency.

Unfortunately, that is not as easy as it sounds, since Edge Runtime is doing a heavy use of vm module and `undici` is using some Node.js APIs that are not available, or they are hard to make them working inside vm.

Related PR: https://github.com/vercel/edge-runtime/pull/309

While we're finishing that work, in order to don't delay users to start using `WebSocket`, we can simply expose it directly from `undici` as dependency.

### DIFF
--- a/packages/node/src/fork-dev-server.ts
+++ b/packages/node/src/fork-dev-server.ts
@@ -21,7 +21,7 @@ export function forkDevServer(options: {
    */
   devServerPath?: string;
 }) {
-  let nodeOptions = process.env.NODE_OPTIONS;
+  let nodeOptions = process.env.NODE_OPTIONS ?? '--no-warnings';
   const tsNodePath = options.require_.resolve('ts-node');
   const esmLoader = pathToFileURL(join(tsNodePath, '..', '..', 'esm.mjs'));
   const cjsLoader = join(tsNodePath, '..', '..', 'register', 'index.js');
--- a/packages/node/src/fork-dev-server.ts
+++ b/packages/node/src/fork-dev-server.ts
@@ -21,7 +21,11 @@ export function forkDevServer(options: {
    */
   devServerPath?: string;
 }) {
-  let nodeOptions = process.env.NODE_OPTIONS ?? '--no-warnings';
+  let nodeOptions = process.env.NODE_OPTIONS || '';
+  
+  if (!nodeOptions.includes('--no-warnings')) {
+    nodeOptions  += ' --no-warnings';
+  }
   const tsNodePath = options.require_.resolve('ts-node');
   const esmLoader = pathToFileURL(join(tsNodePath, '..', '..', 'esm.mjs'));
   const cjsLoader = join(tsNodePath, '..', '..', 'register', 'index.js');
--- a/packages/node/src/fork-dev-server.ts
+++ b/packages/node/src/fork-dev-server.ts
@@ -22,9 +22,9 @@ export function forkDevServer(options: {
   devServerPath?: string;
 }) {
   let nodeOptions = process.env.NODE_OPTIONS || '';
-  
+
   if (!nodeOptions.includes('--no-warnings')) {
-    nodeOptions  += ' --no-warnings';
+    nodeOptions += ' --no-warnings';
   }
   const tsNodePath = options.require_.resolve('ts-node');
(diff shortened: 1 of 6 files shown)


packages/node/src/fork-dev-server.ts
```diff
@@ -20,7 +20,7 @@ export function forkDevServer(options: {
    */
   devServerPath?: string;
 }) {
-  let nodeOptions = process.env.NODE_OPTIONS;
+  let nodeOptions = process.env.NODE_OPTIONS ?? '--no-warnings';
```
Kikobeats: If this is not added, the terminal that runs `vc dev` shows the following message:

```
node:54367) [UNDICI-WS] Warning: WebSockets are experimental, expect them to change at any time.
(Use `node --trace-warnings ...` to show where the warning was created)
```

packages/node/src/edge-functions/edge-handler.ts
```diff
@@ -140,6 +140,8 @@ async function createEdgeRuntimeServer(params?: {
           // This is required for esbuild wrapping logic to resolve
           module: {},
 
+          WebSocket: require('undici').WebSocket,
```
EndangeredMassa: Can you add a test that uses `WebSocket` in some way?

packages/node/src/fork-dev-server.ts
cb1kenobi: This will only add `--no-warnings` if `NODE_OPTIONS` is not defined. It might be better to append `--no-warnings`:

```suggestion
  let nodeOptions = process.env.NODE_OPTIONS || '';
  
  if (!nodeOptions.includes('--no-warnings')) {
    nodeOptions  += ' --no-warnings';
  }
```

Secondly, `nodeOptions` is overwritten by lines 33, 35, and 41. Those should be appended too.

Thirdly, I'd like to land the `vc dev` ESM PR before making more changes to the dev server related bits. That PR touches `nodeOptions`: https://github.com/vercel/vercel/pull/9692/files#diff-5f9eca40d7958b4ff2cd6b5ebcd853b745fe4c76681941846d05a87d1f686c92L32.

cb1kenobi (changes requested):

packages/node/src/edge-functions/edge-handler.ts
Kikobeats: I tested manually with the following code:

```js
import ms from 'ms'

export const config = {
  runtime: 'edge'
}

const createWebSocket = url =>
  new Promise((resolve, reject) => {
    const ws = new WebSocket(url)
    ws.onopen = () => resolve(ws)
    ws.onerror = reject
  })

export default async function websocket () {
  const ws = await createWebSocket('wss://ws.postman-echo.com/raw')
  const start = Date.now()
  const fromStart = () => ms(Date.now() - start)

  const readable = new ReadableStream({
    start (controller) {
      ws.onmessage = ({ data }) => controller.enqueue(`${data}\n`)
      ws.send(fromStart())
      setInterval(() => ws.send(fromStart()), 1000)
    }
  }).pipeThrough(new TextEncoderStream())

  return new Response(readable)
}
```

I cant turn it into an e2e test using a local websocket server

packages/node/test/unit/dev.test.ts
```diff
@@ -24,6 +26,34 @@ function testForkDevServer(entrypoint: string) {
...
+    const child = testForkDevServer('./edge-websocket.js');
+    try {
+      const result = await readMessage(child);
+      if (result.state !== 'message') {
+        throw new Error('Exited. error: ' + JSON.stringify(result.value));
+      }
+
+      const { address, port } = result.value;
+      const response = await fetch(
+        `http://${address}:${port}/api/edge-websocket`
```
Kikobeats, line 41: there is a bug in node-fetch that requires to don't use localhost
https://github.com/node-fetch/node-fetch/issues/1624

cb1kenobi: Ha, their recommendation is don't use node-fetch if you're using Node 18.

cb1kenobi (approved):
LGTM

EndangeredMassa (approved):

PR #9866, opened 2023-04-27
[node] Add log message about compiling ESM to CJS

styfle:
When `@vercel/node` was created 5 or so years ago, Node.js didn't support ESM out of the box. So to create a zero configuration experience, babel was introduce to compile the authored ESM into the runtime CJS.

Today, Node.js supports ESM but you have to opt in using `type: module` in package.json or the `.mjs` file extension.

Indeed `@vercel/node` detects ESM and skips babel compilation. However, there is a case where the source code is ESM and a dependency supports both ESM and CJS via `imports` in package.json. This case will break because `@vercel/nft` traces the source file as ESM and assumes the dependency will be resolved as ESM too. But after `@vercel/nft` traces the source file, `@vercel/node` will compile to CJS and thus the trace will be incorrect and fail at runtime.

So this PR adds a warning when babel compiles ESM to CJS because it is no longer expected in 2023.

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -337,7 +337,7 @@ async function compile(
       if (!meta.compiledToCommonJS) {
         meta.compiledToCommonJS = true;
         console.warn(
-          'Warning: Node.js functions will be compiled from ESM to CommonJS. If this was not intended, add "type": "module" to your package.json file.'
+          'Warning: Node.js functions are compiled from ESM to CommonJS. If this is not intended, add "type": "module" to your package.json file.'
         );
       }
       console.log(`Compiling "${filename}" from ESM to CommonJS...`);
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -119,6 +119,7 @@ async function compile(
   baseDir: string,
   entrypointPath: string,
   config: Config,
+  meta: Meta,
   nodeVersion: NodeVersion,
   isEdgeFunction: boolean
 ): Promise<{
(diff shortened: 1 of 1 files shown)


packages/node/src/index.ts
```diff
@@ -333,6 +334,13 @@ async function compile(
         stream: preparedFiles[path].toStream(),
       });
 
+      if (!meta.compiledToCommonJS) {
+        meta.compiledToCommonJS = true;
+        console.log(
```
cb1kenobi: Do we want this to be rendered to stdout or stderr? If stderr, then consider using `console.warn()` instead.

styfle:
Putting this warning message on hold because I just discovered that we are also compiling ESM => CJS for Edge Functions which I did not expect. 

@Schniz @feugy @javivelasco Is that expected? For example, [this test fixture](https://github.com/vercel/vercel/tree/main/packages/node/test/fixtures/61-edge-function).

styfle:
After further investigation, this could be turned off for edge functions. See PR #9868

packages/node/src/index.ts
```suggestion
        console.warn(
```

styfle: Done in [ad8c880](https://github.com/vercel/vercel/pull/9866/commits/ad8c880023fa06e154a0cb012dc2c869e30b4944)

packages/node/src/index.ts
```diff
@@ -333,6 +334,13 @@ async function compile(
         stream: preparedFiles[path].toStream(),
       });
 
+      if (!meta.compiledToCommonJS) {
+        meta.compiledToCommonJS = true;
+        console.warn(
+          'Warning: Node.js functions will be compiled from ESM to CommonJS. If this was not intended, add "type": "module" to your package.json file.'
```
```suggestion
          'Warning: Node.js functions are compiled from ESM to CommonJS. If this is not intended, add "type": "module" to your package.json file.'
```

TooTallNate (approved):

cb1kenobi (approved):

PR #9868, opened 2023-04-28
[node] Don't compile "edge" functions with Babel

TooTallNate:


packages/node/src/index.ts
```diff
@@ -308,7 +308,7 @@ async function compile(
       !file.endsWith('.mjs') &&
       !file.match(libPathRegEx)
   );
-  if (esmPaths.length) {
+  if (!isEdgeFunction && esmPaths.length) {
```
styfle: Lets try this behind a flag?

```suggestion
  const shouldCompile = !isEdgeFunction || (isEdgeFunction && process.env.VERCEL_SKIP_COMPILE_EDGE_FUNC === '1')
  if (shouldCompile&& esmPaths.length) {
```

styfle (approved):

PR #9869, opened 2023-04-28
[next] Add support for `home` and `global` as edge function regions

timneutkens:
Adds `home` and `global` as supported values for `regions` when coming from Next.js. We'll port these to the build output API at a later time.

Related Next.js changes: https://github.com/vercel/next.js/pull/48959

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2393,6 +2393,10 @@ if (vercelFunctionRegionsVar) {
  * - `global` refers to all regions.
  */
 function normalizeRegions(regions: Regions): string[] {
+  if (typeof regions === 'string') {
+    regions = [regions];
+  }
+
   const newRegions: string[] = [];
   for (const region of regions) {
     // Explicitly mentioned as `home` is one of the explicit values for preferredRegion in Next.js.
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2352,6 +2352,8 @@ interface MiddlewareManifestV2 {
   functions?: { [page: string]: EdgeFunctionInfoV2 };
 }
 
+type Regions = 'home' | 'edge' | 'auto' | string[] | 'all' | 'default';
+
 interface BaseEdgeFunctionInfo {
   env: string[];
   files: string[];
@@ -2359,7 +2361,7 @@ interface BaseEdgeFunctionInfo {
   page: string;
   wasm?: { filePath: string; name: string }[];
   assets?: { filePath: string; name: string }[];
-  regions?: 'auto' | string[] | 'all' | 'default';
+  regions?: Regions;
 }
 
 interface EdgeFunctionInfoV1 extends BaseEdgeFunctionInfo {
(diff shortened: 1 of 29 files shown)


packages/next/src/utils.ts
```diff
@@ -2377,6 +2379,45 @@ interface EdgeFunctionMatcher {
...
+let vercelFunctionRegions: string[] | undefined;
+if (vercelFunctionRegionsVar) {
+  vercelFunctionRegions = vercelFunctionRegionsVar.split(',');
+}
+
+/**
+ * Normalizes the regions config that comes from the Next.js edge functions manifest.
+ * Ensures that config like `home` and `global` are converted to the corresponding Vercel region config.
+ * In the future we'll want to make `home` and `global` part of the Build Output API.
+ * - `home` refers to the regions set in vercel.json or on the Vercel dashboard project config.
```
cramforce, line 2392: https://www.youtube.com/watch?v=a_B80AIQegE

packages/next/src/utils.ts
```diff
@@ -2352,14 +2352,16 @@ interface MiddlewareManifestV2 {
   functions?: { [page: string]: EdgeFunctionInfoV2 };
 }
 
+type Regions = 'home' | 'edge' | 'auto' | string[] | 'all' | 'default';
```
EndangeredMassa, line 2355: **question (non-blocking):** ‏

packages/next/src/utils.ts
```diff
@@ -2506,7 +2547,7 @@ export async function getMiddlewareBundle({
                   ...wasmFiles,
                   ...assetFiles,
                 },
-                regions: edgeFunction.regions,
+                regions: normalizeRegions(edgeFunction.regions ?? []),
```
ijjk: Does an empty array behave the same as `undefined`? If so seems good

packages/next/src/utils.ts
EndangeredMassa: **question (non-blocking):** ‏Can `'global'` be a value here too or does it only show up in the `string[]`?

packages/next/src/utils.ts
```diff
@@ -2377,6 +2379,45 @@ interface EdgeFunctionMatcher {
...
+}
+
+/**
+ * Normalizes the regions config that comes from the Next.js edge functions manifest.
+ * Ensures that config like `home` and `global` are converted to the corresponding Vercel region config.
+ * In the future we'll want to make `home` and `global` part of the Build Output API.
+ * - `home` refers to the regions set in vercel.json or on the Vercel dashboard project config.
+ * - `global` refers to all regions.
+ */
+function normalizeRegions(regions: Regions): string[] {
```
EndangeredMassa, line 2395: **suggestion (soft-blocking):** ‏add tests

In this or a follow-up PR: Can we add tests?

EndangeredMassa (dismissed):
I don't have context in the `home`/`global` requirements, but from the code and comments, this makes sense to me.

packages/next/src/utils.ts
timneutkens: Changed it to be `undefined` when there are no regions

packages/next/src/utils.ts
```diff
@@ -2377,6 +2379,45 @@ interface EdgeFunctionMatcher {
...
+  const newRegions: string[] = [];
+  for (const region of regions) {
```
cb1kenobi, line 2401: It appears as if `regions` could be a `string` or `string[]`, so in the event it's a string (e.g. `"home"`), consider arrayifying it:

```suggestion
  if (!Array.isArray(regions)) {
    regions = [regions];
  }

  for (const region of regions) {
```

cb1kenobi (dismissed):

packages/next/src/utils.ts
timneutkens: Thank you, super good catch! Got bit by the "string can be iterated" with TypeScript once again 🤦

packages/next/src/utils.ts
```diff
@@ -2377,6 +2379,49 @@ interface EdgeFunctionMatcher {
...
+  }
+
+  const newRegions: string[] = [];
+  for (const region of regions) {
+    // Explicitly mentioned as `home` is one of the explicit values for preferredRegion in Next.js.
+    if (region === 'home' && vercelFunctionRegions) {
+      // Includes the regions from the VERCEL_FUNCTION_REGIONS env var.
+      newRegions.push(...vercelFunctionRegions);
+      continue;
+    }
```
cb1kenobi: If `region === 'home'`, but `vercelFunctionRegions` is falsey, then `"home"` will be pushed into `newRegions` below. Is that to be expected?

If not, then an inner `if` would prevent that:

```suggestion
    if (region === 'home') {
      if (vercelFunctionRegions) {
        // Includes the regions from the VERCEL_FUNCTION_REGIONS env var.
        newRegions.push(...vercelFunctionRegions);
      }
      continue;
    }
```

ijjk (dismissed):

packages/next/src/utils.ts
ijjk: We can look at tests in a follow-up yeah

ijjk (dismissed):

TooTallNate (dismissed):

packages/next/src/utils.ts
t3dotgg: omfg

ijjk (approved):

packages/next/src/utils.ts
codercatdev: @cramforce nice!

PR #9875, opened 2023-04-30
Revert "[next] Add support for `home` and `global` as edge function regions"

styfle:
Reverts vercel/vercel#9869 since it caused an error:

```
Invalid request: `regions` should NOT have fewer than 1 items
```

image

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2352,8 +2352,6 @@ interface MiddlewareManifestV2 {
   functions?: { [page: string]: EdgeFunctionInfoV2 };
 }
 
-type Regions = 'home' | 'global' | 'auto' | string[] | 'all' | 'default';
-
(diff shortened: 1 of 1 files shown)


cramforce (approved):

packages/next/src/utils.ts
```diff
@@ -2379,53 +2377,6 @@ interface EdgeFunctionMatcher {
...
-
-    // Explicitly mentioned as `auto` is one of the explicit values for preferredRegion in Next.js.
-    if (region === 'auto') {
-      newRegions.push('auto');
-      continue;
-    }
-
-    newRegions.push(region);
-  }
-  return [];
```
styfle, line 2426: @timneutkens I believe the issue here is that `[]` is invalid. It should be `undefined`.

timneutkens:
Fixed in #9876

Issue #255, opened 2017-01-27
Unable to deploy because of proxy


vladejs:
Hello, I am behind a Proxy, I can access npm registry by putting `http_proxy` and `https_proxy` environment variables, but I can't deploy using `now`. It throws me the following error:
```
> now --token 123567778998876767

> Deploying ~\Desktop\scrapaor
> Error! Unexpected error. Please try later. (request to https://api.zeit.co/now/create failed, reason: getaddrinfo EAI_AGAIN api.zeit.co:443)
```
Is there a solution for this?

OlliV:
It looks like a DNS issue on your side. Can you try `nslookup api.zeit.co`?

jamo:
This indeed sounds like normal issue of an "enterprise" proxied environment...

leo:
👍 ^

alexei87:
but why can I download packages from `npm` and even install `meteor` packages just by exporting `http_proxy`? I think proxy support *is* on your side. You can implement it right?

alexei87:
any progress on this @jamo , @leo ?

vladejs:
@leo , any progress???

leo:
No progress, folks. But we will definitely tackle this once we have time!

vladejs:
@leo , I saw it was  deployed v5.3.2 of `now-cli`. Still without luck trying to deploy behind a proxy. I have a Premium account and this feature is important to me. Do you have any plans to support proxy envs soon?

vladejs:
@leo , I saw it was deployed v6.1.1 of now-cli. Still without luck trying to deploy behind a proxy. I have a Premium account and this feature is important to me. Do you have any plans to support proxy envs soon?

leo:
Hmm, how can we fix this? @OlliV  @jamo

thiagoarrais:
I think I'm being bit by this too.

Here's what I'm getting when trying to install 7.2.4:

    > now@7.2.4 postinstall /usr/lib/node_modules/now
    > node download/install.js

    > For the source code, check out: https://github.com/zeit/now-cli

    > Downloading now CLI 7.2.4 [                    ] 0%{ FetchError: request to https://cdn.zeit.co/releases/now-cli/7.2.4/now-linux failed, reason: socket hang up
        at ClientRequest.<anonymous> (/usr/lib/node_modules/now/download/dist/download.js:6869:11)
        at emitOne (events.js:96:13)
        at ClientRequest.emit (events.js:188:7)
[...]

PR #9880, opened 2023-04-30
[cli] Add support for `HTTPS_PROXY` env var

TooTallNate:
By leveraging the [`proxy-agent`](https://www.npmjs.com/package/proxy-agent) npm module, enable CLI to support making HTTP requests over a proxy, by respecting the `HTTPS_PROXY` or `ALL_PROXY` env var (same convention as `curl` uses).

Finally closes this ancient issue: https://github.com/vercel/vercel/issues/255


packages/cli/src/index.ts
```diff
@@ -251,6 +252,7 @@ const main = async () => {
 
   // Shared API `Client` instance for all sub-commands to utilize
   client = new Client({
+    agent: new ProxyAgent({ keepAlive: true }),
```
styfle, line 255: Is there a way to add a test for this?

I suspect when we switch from `node-fetch` to native `fetch` in the future, that this agent won't work.

TooTallNate: > Is there a way to add a test for this?

Added a unit test in e9476c8352f2faf48a76bcac19883070c7758223.

> I suspect when we switch from `node-fetch` to native `fetch` in the future, that this agent won't work.

Looks like undici uses a different Agent format, so ya. We can cross that bridge when the time comes.

styfle (approved):

PR #9883, opened 2023-05-01
[fs-detectors] Add `disableRootMiddleware`, bypass `ignoreRuntimes` for root-level "middleware" files

TooTallNate:
`ignoreRuntimes: ['@vercel/node']` will also disable root-level middleware, which is not intended. So ensure that middleware still work, even for frameworks that have this `ignoreRuntimes` configuration enabled (Remix, RedwoodJS).

### DIFF
--- a/packages/fs-detectors/src/detect-builders.ts
+++ b/packages/fs-detectors/src/detect-builders.ts
@@ -130,12 +130,17 @@ export async function detectBuilders(
 
   const { projectSettings = {} } = options;
   const { buildCommand, outputDirectory, framework } = projectSettings;
-  const ignoreRuntimes = new Set(
-    slugToFramework.get(framework || '')?.ignoreRuntimes
-  );
+  const frameworkConfig = slugToFramework.get(framework || '');
+  const ignoreRuntimes = new Set(frameworkConfig?.ignoreRuntimes);
   const withTag = options.tag ? `@${options.tag}` : '';
   const apiMatches = getApiMatches()
-    .filter(b => b.config?.middleware || !ignoreRuntimes.has(b.use))
+    .filter(
+      b =>
+        // Root-level middleware is enabled, unless `ignoreMiddleware: true`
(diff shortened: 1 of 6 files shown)


chloetedder (dismissed):

styfle (commented):
> ignoreRuntimes: ['@vercel/node'] will also disable root-level middleware, which is not intended

I thought it was intended which is why @cb1kenobi added #9839

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -135,7 +135,7 @@ export async function detectBuilders(
   );
   const withTag = options.tag ? `@${options.tag}` : '';
   const apiMatches = getApiMatches()
-    .filter(b => !ignoreRuntimes.has(b.use))
+    .filter(b => b.config?.middleware || !ignoreRuntimes.has(b.use))
```
EndangeredMassa: **requirement (blocking):** ‏find a different solution

As Styfle mentions, Storybook specifically uses `ignoreRuntimes` to prevent middleware from being used.

Ref: https://github.com/vercel/vercel/blob/main/packages/frameworks/src/frameworks.ts#L1943

Can we find a solution that works for both?

EndangeredMassa (changes requested):

cb1kenobi:
> > ignoreRuntimes: ['@vercel/node'] will also disable root-level middleware, which is not intended
> 
> I thought it was intended which is why @cb1kenobi added #9839

This is a good point. If a project is explicitly set to to use the `"storybook"` framework, then we want to ignore `@vercel/node` so that the middleware is ignored.

TooTallNate:
Ok so we need another configuration option then.

cb1kenobi (dismissed):
LGTM. Don't forget to remove the `console.log()`.

EndangeredMassa (approved):

cb1kenobi (approved):

PR #9888, opened 2023-05-02
[remix] Suppress warnings when reading config

TooTallNate:
Currently, any warnings logged when calling `readConfig()` are emitted in build logs _three times_:

1. When initially reading the config before executing the Build Command
2. When the Build Command is executed (`remix build` reads the config)
3. After the build, in the `prepareCache()` function

For the callsites that we are in control of (`1.` and `3.`), we should suppress those warnings and only allow the standard warnings when invoking `remix build` to be emitted to the build logs.


packages/remix/src/utils.ts
```diff
@@ -225,11 +225,14 @@ export async function chdirAndReadConfig(dir: string, packageJsonPath: string) {
     modifiedPackageJson = true;
   }
 
+  const warn = console.warn;
   let remixConfig: RemixConfig;
   try {
     process.chdir(dir);
+    console.warn = () => {};
```
EndangeredMassa, line 232: **suggestion (soft-blocking):** ‏add comment and maybe debug

In this or a follow-up PR: Can we add a comment here about why we're doing this? Reading code and coming across this line is a big flag that some funky stuff is happening.

Should we also still add these to the debug log just in case this ends up silencing something else important later?

EndangeredMassa (approved):

PR #9889, opened 2023-05-02
[next] Fix `functions` config with App Router

smaeda-ks:
We added appDir support in https://github.com/vercel/vercel/pull/9811 so that users can customize `memory`/`maxDuration` for routes in appDir.

But since RSC is enabled by default in Next.js 13, `vercel build` still reports some warning messages, such as:

```json
"functions": {
  "app/**/*": {
    "maxDuration": 5,
    "memory": 512
  }
}
```
```
WARNING: Unable to find source file for page hello.js with extensions: ts, tsx, js, jsx, md, mdx, this can cause functions config from `vercel.json` to not be applied
WARNING: Unable to find source file for page index.js with extensions: ts, tsx, js, jsx, md, mdx, this can cause functions config from `vercel.json` to not be applied
```

To suppress these errors and properly apply `functions` setting to those routes, updating the current detection logic to also search `page.${ext}` files.


packages/next/test/integration/integration-1.test.js
```diff
@@ -65,7 +65,13 @@ if (parseInt(process.versions.node.split('.')[0], 10) >= 16) {
       }
     }
 
-    expect(lambdas.size).toBe(3);
+    expect(lambdas.size).toBe(2);
```
styfle: Why are there fewer lambdas?

smaeda-ks: @styfle I think that's because the current rule `app/api/**/*` is only catching `app/api/hello-again/route.js` route, and applying the custom `memory`/`maxDuration` setting to this route only, so it was producing one for `app/api/hello-again/route.js`, one for `app/**/*` (e.g., `page.js`), and another for `pages/api/**/*` === 3

but we're changing the rule to `app/**/*` in this PR so this reduces the extra one and makes it 2

ijjk (dismissed):

packages/next/src/utils.ts
```diff
@@ -1254,21 +1254,38 @@ async function getSourceFilePathFromPage({
     const extensionless = fsPath.replace(path.extname(fsPath), '');
 
     for (const ext of extensionsToTry) {
-      fsPath = `${extensionless}.${ext}`;
+      // for appDir, we need to treat "index.js" as root-level "page.js"
```
styfle, line 1291: Should this special case be wrapped in `if (pageType === 'app')`?

packages/next/src/utils.ts
```diff
@@ -1254,21 +1254,38 @@ async function getSourceFilePathFromPage({
...
+      fsPath =
```
styfle: Should this use a helper method like `normalizeIndexOutput()`?

packages/next/src/utils.ts
```diff
@@ -1254,21 +1254,38 @@ async function getSourceFilePathFromPage({
...
+        extensionless === path.join(workPath, 'app/index') ||
+        extensionless === path.join(workPath, 'src/app/index')
```
styfle: Is the check for `src` here different than line 1247?

packages/next/src/utils.ts
smaeda-ks: applied the suggestion 👍

packages/next/src/utils.ts
smaeda-ks: The check requires multiple vars to be passed, so I think I would leave it as-is

packages/next/src/utils.ts
smaeda-ks: It is for a slightly different purpose. Though, actually, we might need to update `usesSrcDirectory` to also check `src/app` path. Let me include that in this PR.

smaeda-ks: @styfle made some tiny optimizations 👍

packages/next/src/utils.ts
```diff
@@ -1215,10 +1215,14 @@ let _usesSrcCache: boolean | undefined;
...
   if (!_usesSrcCache) {
-    const source = path.join(workPath, 'src', 'pages');
+    const sourcePages = path.join(workPath, 'src', 'pages');
+    const sourceAppdir = path.join(workPath, 'src', 'app');
 
     try {
-      if ((await fs.stat(source)).isDirectory()) {
+      if (
+        (await fs.stat(sourcePages)).isDirectory() ||
+        (await fs.stat(sourceAppdir)).isDirectory()
```
styfle: You'll need two different try/catch because usesSrcDirectory() would need to return true if either one exists. Right now, this would throw when `sourcePages` doesn't exist but  `sourceAppdir` does exist.

smaeda-ks: ah, thanks! fixed it

styfle (dismissed):

styfle (dismissed):

smaeda-ks:
@styfle I had to resolve merge conflicts, which dismissed your previous review approval. Could you approve this PR again? Thanks!

styfle (dismissed):
Thanks!

styfle (approved):

PR #9892, opened 2023-05-02
[cli] Change `vc env pull` default output file to `.env.local`

dan-stowell:
`vc deploy` ignores `.env.local`. To make sure we don't inadvertently push people's secrets to source control, have all environment pulls default to writing to `.env.local`.

[Linear ticket](https://linear.app/vercel/issue/VCCLI-461/change-pull-from-env-to-envlocal-by-default)


styfle (commented):
~~Does `vc dev` need to be updated to read `.env.local`?~~

Looks like we decided that `vc dev` should be changed to stop reading `.env` entirely.

styfle (commented):
Don't forget to update `--help` as well as vercel.com/docs

dan-stowell:
@styfle does `MAJOR` mean we need to bump the major version for this change?

styfle:
@dan-stowell Yes this is semver-major because its a breaking change (needs the label too)

cb1kenobi (dismissed):
Tested and works as advertised.

EndangeredMassa (dismissed):
Should we update these references to `.env` as well?

- errors/missing-env-file.md - should refer to `.env.local`?
- examples/sanity/.env.template - should be `.env.local.template`?
- packages/cli/src/util/dev/server.ts - line 674 and 675 should use `.env.local` and maybe fall back to `.env`?
- packages/cli/src/util/dev/types.ts - comments in `EnvConfigs` should refer to `.env.local`

I'm not positive how best to update `dev/server.ts`, but I think we should take a stab at them in this PR.

dan-stowell:
> Should we update these references to `.env` as well?
> 
> * errors/missing-env-file.md - should refer to `.env.local`?

Done & pushed.

> * examples/sanity/.env.template - should be `.env.local.template`?

Done & pushed.

> * packages/cli/src/util/dev/server.ts - line 674 and 675 should use `.env.local` and maybe fall back to `.env`?
> * packages/cli/src/util/dev/types.ts - comments in `EnvConfigs` should refer to `.env.local`

Based on discussion in [Linear](https://linear.app/vercel/issue/VCCLI-65/update-vc-dev-to-use-vercelenvdevelopmentlocal), it seems that we do not want `vc dev` to read from `.env` at all. If that's the case, I'd like to make that change in a separate pull request.

styfle (dismissed):

EndangeredMassa (dismissed):

cb1kenobi (dismissed):
LGTM

styfle (approved):

cb1kenobi (approved):

PR #9899, opened 2023-05-03
[node] Explicitly set 'set-cookie' header

cb1kenobi:
When looping over the response headers returned by `node-fetch`, it will join `set-cookie` header value into a single string. The whatwg calls for a `headers.getSetCookie()` function to handle the `set-cookie` special case, but `node-fetch@2.x` doesn't support it.

As a workaround, we need to grab the raw `set-cookie` header value.

### DIFF
--- a/packages/node/src/dev-server.mts
+++ b/packages/node/src/dev-server.mts
@@ -109,10 +109,8 @@ async function onDevRequest(
     for (const [key, value] of headers as unknown as Headers) {
       // node-fetch does not support headers.getSetCookie(), so we need to
       // manually set the raw value which can be an array of strings
-      if (key === 'set-cookie') {
-        res.setHeader(key, headers.raw()[key]); // jam it in there
-      } else if (value !== undefined) {
-        res.setHeader(key, value);
+      if (value !== undefined) {
+        res.setHeader(key, key === 'set-cookie' ? headers.raw()[key] : value);
       }
     }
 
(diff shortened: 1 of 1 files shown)


packages/node/src/dev-server.mts
```diff
@@ -105,8 +105,13 @@ async function onDevRequest(
...
     res.statusCode = status;
+
     for (const [key, value] of headers as unknown as Headers) {
-      if (value !== undefined) {
+      // node-fetch does not support headers.getSetCookie(), so we need to
+      // manually set the raw value which can be an array of strings
+      if (key === 'set-cookie') {
+        res.setHeader(key, headers.raw()[key]); // jam it in there
+      } else if (value !== undefined) {
         res.setHeader(key, value);
```
Ethan-Arrowood, line 115:
```suggestion
        res.setHeader(key, key === 'set-cookie' ? headers.raw()[key] : value);
```

Ethan-Arrowood:
```suggestion
      // node-fetch does not support headers.getSetCookie(), so we need to
      // manually set the raw value which can be an array of strings
      if (value !== undefined) {
        res.setHeader(key, key === 'set-cookie' ? headers.raw()[key] : value);
```

Ethan-Arrowood (approved):

packages/node/src/dev-server.mts
```diff
@@ -105,9 +105,12 @@ async function onDevRequest(
   try {
     const { headers, body, status } = await handleEvent(req);
     res.statusCode = status;
+
     for (const [key, value] of headers as unknown as Headers) {
+      // node-fetch does not support headers.getSetCookie(), so we need to
```
styfle, line 110: Perhaps we should switch from node-fetch to undici fetch

EndangeredMassa (commented):
Can we add a test for this case?

PR #9900, opened 2023-05-03
[frameworks] validate getOutputDirName and other missing values on framework entries

EndangeredMassa:
Covers the validation in Framework entries over function properties and other missing holes.

### DIFF
--- a/packages/frameworks/test/frameworks.unit.test.ts
+++ b/packages/frameworks/test/frameworks.unit.test.ts
@@ -297,7 +297,7 @@ function getValidator() {
 
   ajv.addKeyword('isFunction', {
     compile: shouldMatch => data => {
-      const matches = data instanceof Function;
+      const matches = typeof data === 'function';
       return (shouldMatch && matches) || (!shouldMatch && !matches);
     },
   });
--- a/packages/frameworks/test/frameworks.unit.test.ts
+++ b/packages/frameworks/test/frameworks.unit.test.ts
@@ -62,10 +62,25 @@ const SchemaSettings = {
   ],
 };
 
+const RouteSchema = {
+  type: 'array',
+  items: {
+    properties: {
+      src: { type: 'string' },
+      dest: { type: 'string' },
+      status: { type: 'number' },
+      handle: { type: 'string' },
+      headers: { type: 'object' },
+      continue: { type: 'boolean' },
+    },
+  },
+};
+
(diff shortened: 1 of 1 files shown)


packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -62,16 +62,40 @@ const SchemaSettings = {
...
+    },
+  },
+};
+
 const Schema = {
   type: 'array',
   items: {
     type: 'object',
-    required: ['name', 'slug', 'logo', 'description', 'settings'],
+    additionalProperties: false,
```
EndangeredMassa, line 83: Adding a new property to a framework entry will now cause this test to fail, forcing the author to add a type to the schema for it.

packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -62,16 +62,40 @@ const SchemaSettings = {
...
+    required: [
+      'name',
+      'slug',
+      'logo',
+      'description',
+      'settings',
+      'getOutputDirName',
```
EndangeredMassa, line 90: `getOutputDirName` is now required because it's definitely used during deploy

styfle (dismissed):
Very nice, thanks! 🎉

packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -246,3 +291,16 @@ describe('frameworks', () => {
     );
   });
 });
+
+function getValidator() {
+  const ajv = new Ajv();
+
+  ajv.addKeyword('isFunction', {
+    compile: shouldMatch => data => {
+      const matches = data instanceof Function;
```
TooTallNate: I'd prefer `typeof data === 'function'`, but not a huge deal.

TooTallNate (dismissed):

packages/frameworks/test/frameworks.unit.test.ts
EndangeredMassa: I like that better as well. [`bed27ff` (#9900)](https://github.com/vercel/vercel/pull/9900/commits/bed27ffd21c2cd7de6783452f9491b7241f34dc0)

TooTallNate (approved):

cb1kenobi (approved):

styfle (approved):

PR #9901, opened 2023-05-04
[node] prefer fetch over node-fetch

Kikobeats:
I replaced `node-fetch` to use `undici`in the past; However, that changed was finally reverted since undici doesn't support node14:

https://github.com/vercel/vercel/pull/9745/commits/bd3bc0c2cc63a83d1a23e94e36d8c3053fceef80

Now this PR brings back `undici`, but it's using the bundled version available in https://edge-runtime.vercel.app/ that is being patched to support old node versions.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -19,9 +19,10 @@
     "dist"
   ],
   "dependencies": {
+    "@edge-runtime/node-utils": "2.0.3",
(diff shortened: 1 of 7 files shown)


Kikobeats:
closing in favour of https://github.com/vercel/vercel/pull/9906

PR #9910, opened 2023-05-05
[tests] Pin Next.js version in test fixtures

cb1kenobi:
This PR extends the existing `update-next.js` script, which currently only updates the `examples/nextjs` project, and updates the `"next"` dependency in test fixtures.

I've introduced a new `"ignoreNextjsUpdates"` property to the `package.json`. If that property is set to `true`, it will NOT update Next.js in that test fixture.

Currently, I only update Next.js for the `storybook-with-middleware` test fixture of which I authored.

Related, I needed to add `"name"` and `"version"` to a few of the test fixture `package.json` files so that I could update the lock file without issue.

Lastly, there's one test that uses Yarn 2 berry and I can't for the life of me figure out how to update just the `yarn.lock` without installing the dependencies which is slow and creates a bunch of new files. So if it detects a `.yarnrc.yml`, it assumes the test fixture uses Yarn 2/3 and skips the lock file updating.

To test, simply run the `node` REPL and execute:

```
require('./utils/update-next')()
```


utils/update-next.js
```diff
@@ -1,55 +1,133 @@
...
+          deps.next = newVersion;
+          await writeJson(file, pkgJson, { spaces: 2 });
+
+          // update lock file
+          const cwd = dirname(file);
+          try {
+            if (await pathExists(join(cwd, 'yarn.lock'))) {
+              if (await pathExists(join(cwd, '.yarnrc.yml'))) {
+                console.warn(
+                  'Found .yarnrc.yml and this package probably uses Yarn v2/v3 which is not supported'
```
EndangeredMassa: **suggestion (non-blocking):** make message clear about where this file is‏

```suggestion
                  `Found ${join(cwd, '.yarnrc.yml')} and this package probably uses Yarn v2/v3 which is not supported`
```

utils/update-next.js
```diff
@@ -1,55 +1,133 @@
...
+    'packages/cli/test/fixtures/unit/**/package.json',
+    'packages/fs-detectors/test/fixtures/**/package.json',
+    'packages/next/test/fixtures/*/package.json',
+    'packages/next/test/integration/*/package.json',
+    'packages/static-build/test/fixtures/*/package.json',
+  ];
+  for (const pattern of testGlobs) {
+    const files = await glob(pattern);
+    for (const file of files) {
+      const pkgJson = await readJson(file);
```
EndangeredMassa: **suggestion (blocking):** add ‏guard

I get an error here when I try this. It looks like the globs go too deep if there are `node_modules` in a fixture. I added this, then it worked:

```
if (file.includes('node_modules')) {
  // don't go into node_modules of fixtures
  return;
}
```

EndangeredMassa (changes requested):

utils/update-next.js
cb1kenobi: Good catch! I've been continually resetting my repo when testing this PR!

cb1kenobi: Fixed!

utils/update-next.js
```diff
@@ -1,55 +1,141 @@
...
     base: 'main',
-    title: `[examples] Upgrade Next.js to version ${newVersion}`,
-    body: `This auto-generated PR updates Next.js to version ${newVersion}`,
+    title: `[examples][tests] Upgrade Next.js to version ${newVersion}`,
+    body: `This auto-generated PR updates ${updatedCount} package${
+      updatedCount === 1 ? '' : 's'
+    } to Next.js version ${newVersion}`,
   });
 
   await github.rest.pulls.requestReviewers({
```
EndangeredMassa: **suggestion (non-blocking):** ‏try-catch assigning reviewers

[Last I checked](https://github.com/vercel/api/pull/18913/files#diff-3635151cf672ae57a6155f2cb97dbf8232762bd59f6a52a1c04aa3f4886818a6R34-R45), there were still token issues with requesting reviewers, which will make this script fail at this point.

EndangeredMassa (dismissed):

utils/update-next.js
cb1kenobi: Fixed!

EndangeredMassa (dismissed):

cb1kenobi:
@styfle do you have any opinions on this PR? The storybook fixture I recently added is the only fixture where Next.js will be updated. Regardless what happens with the Next builder, I think we'll still want this mechanism to pin Next.js versions. Are you ok with merging this?

utils/update-next.js
```diff
@@ -1,68 +1,160 @@
+/* eslint-disable no-console */
+
+const { dirname, join } = require('path');
 const { execFileSync } = require('child_process');
+const { glob } = require('glob');
+const { pathExists, readJson, writeJson } = require('fs-extra');
```
styfle: I dont think this will work without installing the dependencies first.

utils/update-next.js
```diff
@@ -1,68 +1,160 @@
...
+
+  // update tests
+  const testGlobs = [
+    'packages/cli/test/dev/fixtures/*/package.json',
+    'packages/cli/test/fixtures/unit/**/package.json',
+    'packages/fs-detectors/test/fixtures/**/package.json',
+    'packages/next/test/fixtures/*/package.json',
+    'packages/next/test/integration/*/package.json',
+    'packages/static-build/test/fixtures/*/package.json',
+  ];
```
styfle, line 64: I wonder if we could use the new recursive readdir that @Ethan-Arrowood shipped in Node.js core instead of `glob`

utils/update-next.js
```diff
@@ -1,68 +1,160 @@
...
-  ) {
-    // eslint-disable-next-line no-console
-    console.log(`Branch ${branch} already exists, skipping update.`);
+  console.log(
+    `Updated ${updatedCount} package${
+      updatedCount === 1 ? '' : 's'
+    } to Next.js version ${newVersion}`
+  );
+
+  if (!github || !context) {
```
styfle, line 98: I think we should throw if these are not defined

utils/update-next.js
cb1kenobi: That's easily solved. Installing the deps should only add maybe 20-30 seconds to the CI job.

utils/update-next.js
cb1kenobi: This is by design so it can be tested locally by running `require('./utils/update-next')()`.

utils/update-next.js
cb1kenobi: OK, forget installing deps... no need, I removed them!

EndangeredMassa (approved):

ijjk (approved):
This seems alright to do for very specific test fixtures while leaving the rest unaffected and still testing against latest canary

utils/update-next.js
```diff
@@ -1,68 +1,181 @@
...
   const output = execFileSync(cmd, args, opts).toString().trim();
   console.log({ output });
   console.log();
   return output;
 }
 
-module.exports = async ({ github, context }) => {
-  const oldVersion = require('../examples/nextjs/package.json').dependencies
-    .next;
+module.exports = async ({ github, context } = {}) => {
```
Ethan-Arrowood, line 21: Why was this change necessary? These files should only be invoked by the action and so those properties are definitely assigned. You can add a JSDoc comment like I did in the Gatsby updater for better types: https://github.com/vercel/vercel/blob/91b7f6dcd90ecec469b537f5fc358d579262fe52/utils/update-gatsby-fixtures.js#LL7C1-L11C4

utils/update-next.js
Ethan-Arrowood: Not sure when or if thats getting backported so we'd have to be running node 20 to get access to it

utils/update-next.js
cb1kenobi: I wouldn't say "necessary", but I think it's harmless and even helpful so you can test it in the Node.js repl without having to comment out all the github/context stuff run it.

utils/update-next.js
cb1kenobi: I removed the dependency on `glob` in favor of a iterative, non-recursive solution. This should be more efficient and faster than a recursive solution.

PR #9912, opened 2023-05-05
[cli] refactor help output around default and deploy commands

EndangeredMassa:
When getting help output for the default command `vc --help`, you get a list of commands. When you specify the `deploy` command with `vc deploy --help`, you get the same list of commands.

This PR makes a distinction between an explicit `deploy` command and a default one for the purposes of help output.

Should show CLI help:

- `vc -h`
- `vc --help`
- `vc help`

Should show `deploy` command help:

- `vc deploy -h`
- `vc deploy --help`
- `vc help deploy`

### DIFF
--- a/packages/cli/src/args.ts
+++ b/packages/cli/src/args.ts
@@ -53,7 +53,6 @@ export const help = () => `
 )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
     --no-color                     No color mode [off]
-    -f, --force                    Force a new deployment even if nothing has changed
     -S, --scope                    Set a custom scope
     -t ${chalk.underline('TOKEN')}, --token=${chalk.underline(
   'TOKEN'
--- a/packages/cli/src/args.ts
+++ b/packages/cli/src/args.ts
@@ -1,0 +1,81 @@
+import chalk from 'chalk';
+import logo from './util/output/logo';
+import { getPkgName } from './util/pkg-name';
+
+export const help = () => `
+  ${chalk.bold(`${logo} ${getPkgName()}`)} [options] <command | path>
+
+  ${chalk.dim('Commands:')}
+
+    ${chalk.dim('Basic')}
+
+      deploy               [path]      Performs a deployment ${chalk.bold(
+        '(default)'
+      )}
+      dev                              Start a local development server
+      env                              Manages the Environment Variables for your current Project
+      git                              Manage Git provider repository for your current Project
+      help                 [cmd]       Displays complete help for [cmd]
+      init                 [example]   Initialize an example project
+      inspect              [id]        Displays information related to a deployment
+      link                 [path]      Link local directory to a Vercel Project
+      ls | list            [app]       Lists deployments
+      login                [email]     Logs into your account or creates a new one
+      logout                           Logs out of your account
+      pull                 [path]      Pull your Project Settings from the cloud
+      rollback             [url|id]    Quickly revert back to a previous deployment [beta]
+      switch               [scope]     Switches between teams and your personal account
+
+    ${chalk.dim('Advanced')}
+
+      alias                [cmd]       Manages your domain aliases
+      bisect                           Use binary search to find the deployment that introduced a bug
+      certs                [cmd]       Manages your SSL certificates
+      dns                  [name]      Manages your DNS records
+      domains              [name]      Manages your domain names
+      logs                 [url]       Displays the logs for a deployment
+      projects                         Manages your Projects
+      rm | remove          [id]        Removes a deployment
+      secrets              [name]      Manages your global Secrets, for use in Environment Variables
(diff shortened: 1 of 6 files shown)


packages/cli/src/args.ts
```diff
@@ -0,0 +1,81 @@
+import chalk from 'chalk';
```
EndangeredMassa, line 1: This is mostly the original `args.ts` file from the `commands/deploy` directory, with the deploy-specific stuff left in that original place.

packages/cli/src/util/arg-common.ts
```diff
@@ -28,6 +25,10 @@ const ARG_COMMON = {
   '--api': String,
 
   '--cwd': String,
+
+  // Deprecated
+  '--platform-version': Number,
+  '-V': '--platform-version',
```
EndangeredMassa, line 31: Removed these from help outputs. This isn't used anywhere that I can find. I left the reference here so that existing scripts wouldn't complain.

packages/cli/src/args.ts
```diff
@@ -0,0 +1,81 @@
...
+    --cwd                          Current working directory
+    -A ${chalk.bold.underline('FILE')}, --local-config=${chalk.bold.underline(
+  'FILE'
+)}   Path to the local ${'`vercel.json`'} file
+    -Q ${chalk.bold.underline('DIR')}, --global-config=${chalk.bold.underline(
+  'DIR'
+)}    Path to the global ${'`.vercel`'} directory
+    -d, --debug                    Debug mode [off]
+    --no-color                     No color mode [off]
+    -f, --force                    Force a new deployment even if nothing has changed
```
EndangeredMassa: Should `--force` be removed from the global options? It looks like it's only used in the `deploy`, `domains`, and `init` commands.

TooTallNate: IMO yes.

packages/cli/src/commands/deploy/args.ts
```diff
@@ -89,14 +36,15 @@ export const help = () => `
...
+  ${chalk.gray('–')} Deploy with prebuilt outputs
 
-  ${chalk.gray('–')} Show the usage information for the sub command ${chalk.dim(
-  '`list`'
-)}
+    ${chalk.cyan(`$ ${getPkgName()} build`)}
+    ${chalk.cyan(`$ ${getPkgName()} deploy --prebuilt`)}
 
-    ${chalk.cyan(`$ ${getPkgName()} help list`)}
+  ${chalk.gray('–')} Save reference to Deployment URL
```
TooTallNate: "Reference" is kinda vague. How about:

```suggestion
  ${chalk.gray('–')} Write Deployment URL to a file
```

packages/cli/src/util/arg-common.ts
TooTallNate: Also see https://github.com/vercel/vercel/pull/9807.

packages/cli/src/args.ts
EndangeredMassa: Oh, it's not even a common arg now. Yeah, let's remove it. [`c2e0ca5` (#9912)](https://github.com/vercel/vercel/pull/9912/commits/c2e0ca562b1a3d3b877508bd3f91872bf5e99668)

TooTallNate (approved):

packages/cli/src/args.ts
EndangeredMassa: Removed from docs, too: [vercel/front@`95b71be` (#22110)](https://github.com/vercel/front/pull/22110/commits/95b71be4e41b901f7860e5128c73416ac493e963)

EndangeredMassa:
I also noticed that `--prebuilt` wasn't documented in help output, but I have a separate PR around that: https://github.com/vercel/vercel/pull/9913

packages/cli/src/args.ts
```diff
@@ -0,0 +1,80 @@
...
+      dns                  [name]      Manages your DNS records
+      domains              [name]      Manages your domain names
+      logs                 [url]       Displays the logs for a deployment
+      projects                         Manages your Projects
+      rm | remove          [id]        Removes a deployment
+      secrets              [name]      Manages your global Secrets, for use in Environment Variables
+      teams                            Manages your teams
+      whoami                           Shows the username of the currently logged in user
+
+  ${chalk.dim('Options:')}
```
cb1kenobi, line 43: This is still confusing to me. These are technically global options which is what our docs call them: https://vercel.com/docs/cli/global-options.

I think if this section was renamed to "Global Options:", then add above it a "Deploy Options:" (or simply "Options:") with the `vc deploy` options, that would provide a better user experience.

EndangeredMassa: How about this? https://github.com/vercel/vercel/pull/9917

cb1kenobi: Love it!

PR #9913, opened 2023-05-05
[cli] rename `--prebuilt` to `--skip-build`

EndangeredMassa:
As part of moving towards a more flexible deployment process, we want to make it clearer to users that some of those steps can be skipped. You could consider prebuilt deployments as skipping the build step. In that mental model, the option `--skip-build` makes more sense than `--prebuilt`.

This PR creates the new `--skip-build` option, but leaves the `--prebuilt` option in place. Both will work just fine.

This will eventually be in line with:

- `--skip-build`: skip building from source for this deployment
- `--skip-check`: skip any checks for this deployment
- `--skip-domain`: skip assigning custom domains for this deployment

---

Depends on: 
- https://github.com/vercel/vercel/pull/9912

### DIFF
--- a/packages/cli/src/commands/deploy/args.ts
+++ b/packages/cli/src/commands/deploy/args.ts
@@ -2,80 +2,27 @@ import chalk from 'chalk';
 import logo from '../../util/output/logo';
 import { getPkgName } from '../../util/pkg-name';
 
-export const help = () => `
-  ${chalk.bold(`${logo} ${getPkgName()}`)} [options] <command | path>
+export const help = () => {
+  return `
+  ${chalk.bold(`${logo} ${getPkgName()} [deploy]`)} [path-to-project] [options]
 
-  ${chalk.dim('Commands:')}
-
-    ${chalk.dim('Basic')}
-
-      deploy               [path]      Performs a deployment ${chalk.bold(
-        '(default)'
-      )}
-      dev                              Start a local development server
-      env                              Manages the Environment Variables for your current Project
-      git                              Manage Git provider repository for your current Project
-      help                 [cmd]       Displays complete help for [cmd]
-      init                 [example]   Initialize an example project
-      inspect              [id]        Displays information related to a deployment
-      link                 [path]      Link local directory to a Vercel Project
-      ls | list            [app]       Lists deployments
-      login                [email]     Logs into your account or creates a new one
-      logout                           Logs out of your account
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/deploy/args.ts
```diff
@@ -7,6 +7,7 @@ export const help = () => {
   ${chalk.bold(`${logo} ${getPkgName()} [deploy]`)} [path-to-project] [options]
 
   --prod                         Create a production deployment
+  --skip-build, --prebuilt       Use prebuilt output from ".vercel/output"
```
EndangeredMassa, line 10: We could alternatively add the alias, but leave `--prebuilt` as the documented flag for now.

packages/cli/src/commands/deploy/args.ts
```diff
@@ -41,7 +42,7 @@ export const help = () => {
   ${chalk.gray('–')} Deploy with prebuilt outputs
 
     ${chalk.cyan(`$ ${getPkgName()} build`)}
-    ${chalk.cyan(`$ ${getPkgName()} deploy --prebuilt`)}
+    ${chalk.cyan(`$ ${getPkgName()} deploy --skip-build`)}
```
styfle, line 45: I think `vercel deploy --skip-build` is confusing.

You're not telling it to skip the build, you're telling it that you are deploying files you already built which is why I think `--prebuilt` makes more sense.

When I read `--skip-build`, it sounds like a static deployment with `index.html`, etc that doesn't require a build to be deployed. But that's not the same as `--prebuilt` which requires you put any build outputs into the `.vercel/output` directory. 

Also consider using something in the future like `vercel deploy --prebuilt ./path/to/output`.

EndangeredMassa: I think it depends on how we frame it for the user. The option `--skip-build` makes sense when you are thinking about deployments steps: "skip the build because I already did it".

If we did allow a different path argument for where to find the BOA, then yes, the `--skip-build` makes less sense. But we could rearrange that:

```
vercel deploy ./path/to/output --skip-build
```

In this framing, we're deploying the built output. You could say we don't even need the `--skip-build` option in this case, but I think being explicit here is helpful. Technically you could want to deploy something that looks like BOA, but as static assets.

EndangeredMassa:
After some discussion, we decided not to do this for now.

PR #9914, opened 2023-05-05
Use `changesets` instead of `lerna`

TooTallNate:
`changesets` will improve our release workflow, since we will no longer need to manually curate and publish the changelog and GitHub Release. Instead, you simply merge the publish PR that the changesets GH action maintains as we push commits to `main`.


EndangeredMassa:
Do you have a draft of what the new release process looks like?
https://github.com/vercel/vercel/wiki/Creating-a-Release

TooTallNate:
The new release process is:

> Merge the "Version Packages" PR.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8.3.1
+
+      - name: Setup Node.js 16.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+          cache: 'pnpm'
```
styfle: The github cache is usually slower. Lets try to use similar steps we had for the existing publish

TooTallNate: 6e4b5f77376e5f5f07217f520e1d3dafc1f87f4f

.changeset/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
```
EndangeredMassa, line 3: **suggestion (non-blocking):** use custom release line format‏

I would like to change the release line format, which it looks like [we can do here](https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md), to reference the PR instead of the commit and to include the author.

TooTallNate: Sure, perhaps [Svelte's](https://www.npmjs.com/package/@svitejs/changesets-changelog-github-compact) would be satisfactory already.

Example: https://github.com/vercel/edge-runtime/releases/tag/%40edge-runtime%2Fvm%402.1.2

Ethan-Arrowood (approved):
<picture data-single-emoji=":ship-on-skis:" title=":ship-on-skis:">:ship-on-skis:</picture>

PR #9933, opened 2023-05-10
Use Svelte's changelog generator and don't delete `CHANGELOG.md`

TooTallNate:
Partial revert of #9932, since apparently [that didn't work](https://github.com/vercel/vercel/actions/runs/4941266637/jobs/8833725828).

We can investigate that more later, but this gets it working again properly at least.

Also use Svelte's changelog generator, since it includes the PR numbers which Sean wanted.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,6 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "changelog": [
+    "@svitejs/changesets-changelog-github-compact",
+    { "repo": "vercel/vercel" }
+  ],
   "commit": false,
   "fixed": [],
   "linked": [],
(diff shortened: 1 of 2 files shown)


.changeset/config.json
```diff
@@ -1,6 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "changelog": [
+    "@svitejs/changesets-changelog-github-compact",
+    { "repo": "vercel/vercel" }
+  ],
```
styfle, line 6: Is this the correct syntax to have an array with first element a string and second element an object?

TooTallNate: Ya, it's similar to eslint plugins. First is the package name, and next is the options. See [here](https://github.com/vercel/edge-runtime/blob/main/.changeset/config.json#L3-L6).

cb1kenobi (approved):

styfle (approved):

PR #9944, opened 2023-05-12
[vc dev] Skip 50MB zip size limit for Python

lucleray:
Do not apply the 50MB zip size limit for Python serverless function during `vc dev`.

Related to internal PR: https://github.com/vercel/api/pull/19279


styfle (approved):

cb1kenobi (approved):

packages/cli/src/util/dev/builder.ts
```diff
@@ -337,7 +337,11 @@ export async function executeBuild(
   // Enforce the lambda zip size soft watermark
   const maxLambdaBytes = bytes('50mb');
   for (const asset of Object.values(result.output)) {
-    if (asset.type === 'Lambda') {
+    if (
+      asset.type === 'Lambda' &&
+      typeof asset.runtime === 'string' &&
+      asset.runtime.startsWith('python')
```
pvdz, line 343: @lucleray does this not do the opposite?

Doesn't this effectively _only_ apply the size limit to lambdas whose runtime start with "python"?

PR #9947, opened 2023-05-12
[node] Add isomorphic functions

Kikobeats:
This PR allow using Web APIs in Serverless functions

```js
// api/serverless.js
export const GET = () => {
  return new Response(`new Response('👋 Hello from Serverless Web!)`)
}
```

More about that https://nextjs.org/docs/app/building-your-application/routing/router-handlers

### DIFF
--- a/packages/cli/src/util/extension/proxy.ts
+++ b/packages/cli/src/util/extension/proxy.ts
@@ -8,10 +8,7 @@ import {
 import type { Server } from 'http';
 import type Client from '../client';
 
-const toHeaders = buildToHeaders({
-  // @ts-expect-error - `node-fetch` Headers is missing `getAll()`
-  Headers,
-});
+const toHeaders = buildToHeaders({ Headers });
 
 export function createProxy(client: Client): Server {
   return createServer(async (req, res) => {
--- a/packages/node/src/edge-functions/edge-handler-template.js
+++ b/packages/node/src/edge-functions/edge-handler-template.js
@@ -10,14 +10,14 @@ function getUrl(url, headers) {
   return urlObj.toString();
 }
 
-async function respond(userEdgeHandler, event, options, dependencies) {
+async function respond(handler, event, options, dependencies) {
   const { Request, Response } = dependencies;
   const { isMiddleware } = options;
   event.request.headers.set(
     'host',
     event.request.headers.get('x-forwarded-host')
   );
-  let response = await userEdgeHandler(
+  let response = await handler(
     new Request(
       getUrl(event.request.url, event.request.headers),
       event.request
@@ -62,15 +62,31 @@ async function parseRequestEvent(event) {
 
 // This will be invoked by logic using this template
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-function registerFetchListener(userEdgeHandler, options, dependencies) {
+function registerFetchListener(module, options, dependencies) {
   addEventListener('fetch', async event => {
+    let handler;
+    if (typeof module.default === 'function') {
+      handler = module.default;
+    } else {
+      if (
+        ['GET', 'HEAD', 'OPTIONS', 'POST', 'PUT', 'DELETE', 'PATCH'].some(
+          method => typeof module[method] === 'function'
+        )
+      ) {
+        const method = event.request.method ?? 'GET';
+        handler =
+          typeof module[method] === 'function'
+            ? module[method]
+            : () => new dependencies.Response(null, { status: 405 });
+      } else {
+        throw new Error(
+          'No default export was found. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware'
+        );
+      }
+    }
+
     try {
-      const response = await respond(
-        userEdgeHandler,
-        event,
-        options,
-        dependencies
-      );
+      const response = await respond(handler, event, options, dependencies);
       return event.respondWith(response);
     } catch (error) {
       event.respondWith(toResponseError(error, dependencies.Response));
--- a/packages/node/src/edge-functions/edge-handler-template.js
+++ b/packages/node/src/edge-functions/edge-handler-template.js
@@ -80,7 +80,7 @@ function registerFetchListener(module, options, dependencies) {
             : () => new dependencies.Response(null, { status: 405 });
       } else {
         throw new Error(
-          'No default export was found. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware'
+          `No default export was found at ${event.request.url}. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware`
         );
       }
     }
--- a/packages/node/src/edge-functions/edge-handler-template.js
+++ b/packages/node/src/edge-functions/edge-handler-template.js
@@ -82,14 +82,12 @@ function registerFetchListener(module, options, dependencies) {
               : () => new dependencies.Response(null, { status: 405 });
         }
       }
-
       if (!handler) {
-        const error = new Error(
-          `No default export was found at ${event.request.url}. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware`
+        const url = getUrl(event.request.url, event.request.headers);
+        throw new Error(
+          `No default export was found at ${url}. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware`
         );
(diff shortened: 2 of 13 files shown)


packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -62,15 +62,31 @@ async function parseRequestEvent(event) {
...
+        )
+      ) {
+        const method = event.request.method ?? 'GET';
+        handler =
+          typeof module[method] === 'function'
+            ? module[method]
+            : () => new dependencies.Response(null, { status: 405 });
+      } else {
+        throw new Error(
+          'No default export was found. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware'
```
EndangeredMassa: **suggestion (non-blocking):** ‏add url to error

Can we add the url to the error message to make it easier to debug where this error comes from?

```suggestion
          `No default export was found at ${event.request.url}. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware`
```

packages/node/src/edge-functions/edge-handler.mts
```diff
@@ -104,7 +99,7 @@ async function compileUserCode(
       ${edgeHandlerTemplate};
       const dependencies = { Request, Response };
       const options = { isMiddleware, entrypointLabel };
-      registerFetchListener(userEdgeHandler, options, dependencies);
+      registerFetchListener(handler, options, dependencies);
```
EndangeredMassa: **suggestion (non-blocking):** ‏rename `handler` to something like `userModule`

Calling this `handler` here, but `module` inside the function is confusing to follow. Using `module` as a variable name is also a bit confusing because it shadows the node module global.

Can we rename `handler` outside the function and `module` inside the function both to something like `userModule` to make this clearer?

packages/node/src/serverless-functions/serverless-handler.mts
```diff
@@ -21,40 +21,59 @@ type ServerlessFunctionSignature = (
...
-  userCode: ServerlessFunctionSignature,
-  options: ServerlessServerOptions
-) {
-  const server = createServer(async (req, res) => {
-    if (options.shouldAddHelpers) await addHelpers(req, res);
-    return userCode(req, res);
-  });
+const [NODE_MAJOR] = process.versions.node.split('.').map(v => Number(v));
+
+/* https://nextjs.org/docs/app/building-your-application/routing/router-handlers#supported-http-methods */
```
EndangeredMassa, line 26: **question (non-blocking):** ‏Is there a plan to produce a Vercel page for supported methods?

packages/node/src/serverless-functions/serverless-handler.mts
```diff
@@ -21,40 +21,59 @@ type ServerlessFunctionSignature = (
...
   /**
    * In some cases we might have nested default props due to TS => JS
    */
   for (let i = 0; i < 5; i++) {
-    if (fn.default) fn = fn.default;
+    if (listener.default) listener = listener.default;
+  }
+
+  if (HTTP_METHODS.some(method => typeof listener[method] === 'function')) {
+    if (NODE_MAJOR < 18) throw new Error('Node.js v18 or above is needed');
```
EndangeredMassa: **suggestion (soft-blocking):** add detail to error

In this or a follow-up PR: Can we expand this error message to make it clearer what's going on?‏

```suggestion
    if (NODE_MAJOR < 18) {
      throw new Error('Node.js v18 or above is required to use HTTP method exports in your functions. Learn more: ${SOME_LINK}');
    }
```

EndangeredMassa (commented):
I didn't manually test this yet, but it's looking awesome

packages/node/src/edge-functions/edge-handler.mts
Kikobeats: `userModule` sounds perfect to me!

packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -62,19 +62,35 @@ async function parseRequestEvent(event) {
 
 // This will be invoked by logic using this template
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-function registerFetchListener(userEdgeHandler, options, dependencies) {
+function registerFetchListener(module, options, dependencies) {
+  let handler;
+
   addEventListener('fetch', async event => {
-    try {
```
EndangeredMassa, line 67: **suggestion (blocking):** ‏add back try-catch

The try-catch was here to catch errors in `addEventListener` callback, which includes user code, and render out the error message. I think we still want that protection.

EndangeredMassa (changes requested):

packages/node/src/serverless-functions/serverless-handler.mts
Kikobeats: Yes, it will be updated in a follow-up PR 🙂

packages/node/src/serverless-functions/serverless-handler.mts
Kikobeats: For now, we are going to just throw the error. When docs are ready, we can link it properly.

packages/node/src/edge-functions/edge-handler-template.js
```diff
@@ -62,16 +62,34 @@ async function parseRequestEvent(event) {
...
+          handler =
+            typeof module[method] === 'function'
+              ? module[method]
+              : () => new dependencies.Response(null, { status: 405 });
+        }
+      }
+      if (!handler) {
+        const url = getUrl(event.request.url, event.request.headers);
+        throw new Error(
+          `No default export was found at ${url}. Add a default export to handle requests. Learn more: https://vercel.link/creating-edge-middleware`
```
EndangeredMassa: **suggestion (non-blocking):** ‏mention named exports too

Should we also mention named exports in this message?

packages/node/src/serverless-functions/serverless-handler.mts
EndangeredMassa: I think we should still add a bit more detail:

```suggestion
    if (NODE_MAJOR < 18) {
      throw new Error('Node.js v18 or above is required to use HTTP method exports in your functions.');
    }
```

cb1kenobi: This message is not rendered correctly as an error:

image

cb1kenobi (changes requested):

packages/node/test/unit/dev.test.ts
```diff
@@ -24,6 +26,41 @@ function testForkDevServer(entrypoint: string) {
   });
 }
 
+(NODE_MAJOR < 18 ? test.skip : test)(
```
EndangeredMassa, line 29: Chris and I ran this test manually on node 18 and it passed. 🎉

packages/node/src/utils.ts
```diff
@@ -59,7 +59,12 @@ export function entrypointToOutputPath(
 }
 
 export function logError(error: Error) {
-  console.error(error.message);
+  let message = error.message;
+  if (!message.startsWith('Error:')) {
+    message = `Error: ${message}`;
+  }
+  console.error(message);
```
EndangeredMassa, line 66: This is an improvement, but I realized we show errors differently between startup/runtime and serverless-node/edge. We should standardize those differences, but not in this PR.

EndangeredMassa (approved):

cb1kenobi (approved):
Code looks good. I tested a handful of scenarios and didn't find any issues.

packages/cli/src/util/extension/proxy.ts
```diff
@@ -8,10 +8,7 @@ import {
 import type { Server } from 'http';
 import type Client from '../client';
 
-const toHeaders = buildToHeaders({
-  // @ts-expect-error - `node-fetch` Headers is missing `getAll()`
-  Headers,
-});
+const toHeaders = buildToHeaders({ Headers });
```
TooTallNate, line 11: I don't understand how this TS error was fixed <picture data-single-emoji=":thinkies:" title=":thinkies:">:thinkies:</picture>

TooTallNate (approved):

packages/cli/src/util/extension/proxy.ts
EndangeredMassa: Upgrading `@edge-runtime/node-utils` fixed the issue I think because of this PR: https://github.com/vercel/edge-runtime/pull/309

PR #9956, opened 2023-05-15
[cli] New `vc redeploy` command

cb1kenobi:
This adds a new `vc redeploy <url-or-id>` command. It fetches the requested deployment, then performs a redeploy with similar output to `vc deploy` including the ability to pipe the deployment URL into a file or program.

### Redeploy Example:

image

### Bad deployment URL:

image

### No args:

image

Linear: https://linear.app/vercel/issue/VCCLI-558/cli-new-command-to-redeploy


packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  ${chalk.dim('Examples:')}
+
+  ${chalk.gray('–')} Rebuild and deploy an existing deployment using id or url
+
+    ${chalk.cyan(`$ ${getPkgName()} redeploy <deployment id/url>`)}
+
+  ${chalk.gray('–')} Redeploy a deployment using id or url
+
+    ${chalk.cyan(`$ ${getPkgName()} redeploy <deployment id/url>`)}
```
EndangeredMassa: **suggestion (soft-blocking):** remove or update example

‏‏Should this be a different example? Right now, it looks like a duplicate of the example before it

packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+ * `vc redeploy` command
+ * @param {Client} client
+ * @returns {Promise<number>} Resolves an exit code; 0 on success
+ */
+export default async (client: Client): Promise<number> => {
+  let argv;
+  try {
+    argv = getArgs(client.argv.slice(2), {
+      '--debug': Boolean,
+      '-d': '--debug',
```
EndangeredMassa: **suggestion (soft-blocking):** remove

In this or a follow-up PR: The debug option is a global/common one. I don't think we need this here. Can we remove it?‏

packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+    const fromDeployment = await getDeploymentByIdOrURL({
+      client,
+      contextName,
+      deployId,
+    });
+
+    const deployStamp = stamp();
+    output.spinner(`Redeploying project ${fromDeployment.id}`, 0);
+
+    let deployment = await client.fetch<any>(`/v13/deployments?forceNew=1`, {
```
TooTallNate, line 104: Can we get rid of `any`?

packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+    output.stopSpinner();
+    output.print(
+      `${prependEmoji(
+        `Inspect: ${chalk.bold(deployment.inspectorUrl)} ${deployStamp()}`,
+        emoji('inspect')
+      )}\n`
+    );
+
+    if (!client.stdout.isTTY) {
+      process.stdout.write(`https://${deployment.url}`);
```
TooTallNate:
```suggestion
      client.stdout.write(`https://${deployment.url}`);
```

packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+              const error = await now.handleDeploymentError(event.payload, {
+                env: {},
+              });
+              throw error;
+            }
+          }
+        } catch (err: unknown) {
+          if (err instanceof Error) {
+            output.error(err.toString());
+          }
```
TooTallNate:
```suggestion
          output.prettyError(err);
```

packages/cli/src/commands/redeploy.ts
```diff
@@ -0,0 +1,214 @@
...
+          process.exit(1);
+        }
+      }
+    }
+
+    return printDeploymentStatus(client, deployment, deployStamp, noWait);
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      output.error(err.message);
+    }
```
TooTallNate:
```suggestion
    output.prettyError(err);
```

packages/cli/src/util/rollback/request-rollback.ts
```diff
@@ -28,97 +27,54 @@ export default async function requestRollback({
...
+      contextName,
+      deployId,
+    });
 
-    const [teamResult, deploymentResult] = await Promise.allSettled([
-      config.currentTeam ? getTeamById(client, config.currentTeam) : undefined,
-      getDeployment(client, contextName, deployId),
-    ]);
+    // create the rollback
+    await client.fetch<any>(
```
TooTallNate: If we're not reading the response, I think we can drop `any`:

```suggestion
    await client.fetch(
```

packages/cli/src/util/rollback/request-rollback.ts
```diff
@@ -28,97 +27,54 @@ export default async function requestRollback({
...
+      client,
+      contextName,
+      deployment,
+      project,
+      timeout,
+    });
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      output.error(err.message);
     }
```
TooTallNate:
```suggestion
    output.prettyError(err);
```

TooTallNate (changes requested):

EndangeredMassa (commented):
I tried a redeploy, but I get an error that the deployment is not part of my current team (uncurated-tests), but it is!

Screenshot 2023-05-18 at 9 35 34 AM

EndangeredMassa (changes requested):
@cb1kenobi and I paired on it and figured out my issue: I was using the dashboard url for a deployment instead of a deploy url. Using the deploy url does work, but using the dashboard url should show a clearer message (or ideally it would work just the same).

EndangeredMassa:
Can we allow deployment ids without the `dpl_` prefix?

packages/cli/src/commands/redeploy.ts
cb1kenobi: In short, yes, but it's outside the scope of this PR. This line was copied from `deploy.ts`. Additionally, `deployment` here is passed around into other functions such as `checkDeploymentStatus()` in `client` which uses a different `Deployment` type than the one used in `cli` and `@vercel-internals/types`. There should not be multiple versions of the `Deployment` type, but fixing it is going to touch a lot of code, so best to do it in a different PR.

packages/cli/src/commands/redeploy.ts
cb1kenobi: Ha, yeah, I think I was going to do a `--no-wait` example, but completely forgot. I don't see a `--no-wait` example for `vc deploy`, so I just removed it in 05ce35864ca8281cbd1a03e3fbfe806a9b074ddb.

packages/cli/src/commands/redeploy.ts
cb1kenobi: Fixed in 25eb87f97e624c77bfa34197494652df5b3df1ff.

cb1kenobi:
> Can we allow deployment ids without the `dpl_` prefix?

@EndangeredMassa Added support for deployment ids without the `dpl_` prefix in chrisbarber/vccli-558-cli-new-command-to-redeploy.

packages/cli/src/util/get-deployment.ts
```diff
@@ -27,12 +29,17 @@ export default async function getDeployment(
...
 
   try {
-    return await client.fetch<Deployment>(
+    return client.fetch<Deployment>(
       `/v13/deployments/${encodeURIComponent(hostOrId)}`
     );
   } catch (err: unknown) {
     if (isAPIError(err)) {
       if (err.status === 404) {
+        if (/^[A-Za-z0-9]+$/.test(hostOrId)) {
```
EndangeredMassa: **suggestion (non-blocking):** ‏use strict check

This check makes me nervous that it will hit something we didn't intend. Can we check for the value starting with `dpl_` explicitly?

packages/cli/src/util/deploy/get-deployment-by-id-or-url.ts
```diff
@@ -0,0 +1,89 @@
...
+}: {
+  client: Client;
+  contextName: string;
+  deployIdOrUrl: string;
+}): Promise<Deployment> {
+  const { config, output } = client;
+
+  if (!isValidName(deployIdOrUrl)) {
+    throw new Error(
+      `The provided argument "${deployIdOrUrl}" is not a valid deployment or project`
```
EndangeredMassa: **suggestion (soft-blocking):** don't mention project

```suggestion
      `The provided argument "${deployIdOrUrl}" is not a valid deployment ID or URL`
```

EndangeredMassa (dismissed):

packages/cli/src/util/get-deployment.ts
cb1kenobi: `hostOrId` is either a URL or a deployment id. Both a URL and a valid deployment id will fail that regex because they contain `.:/_-`. If a `hostOrId` is only alpha-numeric, then the code simply retries with the `dpl_` prefix. We specifically do not want this conditional to check for `dpl_`.

EndangeredMassa (dismissed):

EndangeredMassa (dismissed):

packages/cli/src/util/get-deployment.ts
TooTallNate: Is this because we strip `dpl_` in the deployment inspector URL? So the thinking is that a user could have copy/pasted that value to this command?

cb1kenobi: @TooTallNate yes. Good idea? Bad idea?

TooTallNate: I'm undecided 😅 But I do think that if this were to be supported it should be done at the API level.

cb1kenobi: OK, reverted auto `dpl_` prefixing.

packages/cli/src/util/get-deployment.ts
```diff
@@ -10,7 +10,9 @@ import mapCertError from './certs/map-cert-error';
 import toHost from './to-host';
 
 /**
- * Retrieves a v13 deployment.
+ * Retrieves a deployment by id or URL. If the deployment is not found and the
+ * deployment `hostOrId` contains alpha-numeric characters only, it will retry
+ * with the `dpl_` prefix.
```
TooTallNate:
```suggestion
 * Retrieves a deployment by id or URL.
```

cb1kenobi: Thank you!

TooTallNate (approved):

PR #9959, opened 2023-05-16
fix release script

EndangeredMassa:
The [previous PR](https://github.com/vercel/vercel/pull/9942) used `github.repos`, but I think this needs to be `github.rest.repos`.

- [Docs](https://octokit.github.io/rest.js/v19#repos)
- [Failed GH Action](https://github.com/vercel/vercel/actions/runs/4994578940/jobs/8945329301)

### DIFF
--- a/utils/update-latest-release.js
+++ b/utils/update-latest-release.js
@@ -1,6 +1,6 @@
 module.exports = async ({ github, context }) => {
   const { owner, repo } = context.repo;
-  const response = await github.repos.listReleases({ owner, repo });
+  const response = await github.rest.repos.listReleases({ owner, repo });
 
   function isVercelCliRelease(release) {
     return release.tag_name.startsWith('vercel@');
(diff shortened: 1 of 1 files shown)


utils/update-latest-release.js
```diff
@@ -1,6 +1,6 @@
 module.exports = async ({ github, context }) => {
   const { owner, repo } = context.repo;
-  const response = await github.repos.listReleases({ owner, repo });
+  const response = await github.rest.repos.listReleases({ owner, repo });
```
cb1kenobi, line 3: Is `github` an Octokit instance? If so, this should work.

Additionally, I was told that I should use the pagination API like this:

```
const iter = await github.paginate.iterator(
  github.rest.repos.listReleases,
  { owner, repo, per_page: 100 }
);

for await (const { releases } of iterator) {
  console.log(`Processing ${releases.length} releases`);
  for (const release of releases) {
    console.log(release);
  }
}
```

EndangeredMassa: > Is `github` an Octokit instance? If so, this should work.

Yup: https://github.com/actions/github-script

cb1kenobi (approved):

PR #9960, opened 2023-05-16
Revert "Version Packages"

dan-stowell:
Reverts vercel/vercel#9950


TooTallNate:
I don't think we should revert the release PR. Just revert the individual commit instead.

EndangeredMassa:
Turns out this was not related.

PR #9967, opened 2023-05-17
[cli] Fix `vercel git connect` command when passing a URL parameter

bodinsamuel:
Hey team,

I had an error this morning when trying to link my repo using the CLI 
Screenshot 2023-05-17 at 11 38 01

Seems like a combination of `: any` and a recent refacto introduced an issue here.

~- Removed all `any` that were hiding the error~
- Correctly pass the URL already parsed

~- Removed the second useless parsing~
~- Added missing `--help` flag~
  ~I noticed you never specify them and Typescript screams because of that, not sure if on purpose and how you make the compilation pass with this error. Don't hesitate to tell me.~

The fix could be improved by using `arg.Result<THE_ACTUAL_TYPE>` but that would require to store the dictionnary of flags somewhere else and also create an external type. This is already better so...

--- 

NB: 
I had multiple issues while cloning this repo
- pnpm bootstrap does not work
- missing `constants.ts` file that is not explained in the Contributing guidelines
- maybe due to bootstrap not working, almost nothing else worked and the tests too.


zezo2019 (approved):

packages/cli/src/commands/git/connect.ts
```diff
@@ -116,7 +117,7 @@ export default async function connect(
       confirm,
       org,
       project,
-      repoInfo: repoArg,
+      repoInfo: parsedUrlArg,
```
cb1kenobi, line 119: I have reproduced the issue. After doing some testing, I believe this line (`repoInfo: parsedUrlArg,`) is the only line that needed to be changed. I suggest reverting all other changes.

Also, please add a test for this scenario to `packages/cli/test/unit/commands/git.test.ts`. The test needs to use a unlinked fixture, so there's a `unlinked` fixture you can use. The test should succeed if `vc git connect https://github.com/test_org/test_repo` correctly errors out failing to connect to `test_org/test_repo`.

cb1kenobi (dismissed):

packages/cli/src/commands/git/connect.ts
bodinsamuel: I have reverted the changes as requested but I still believe there is an issue with the types that could occur again if not fixed. And the double parsing is unnecessary.

--- 

For the test I tried my best but still can't run the codebase properly so it'll need to be double checked

cb1kenobi: I agree the types can be improved, but I think that's beyond the scope of this issue. The double parsing isn't ideal, but it's ok in this moment.

As for getting the code base to work, be in the root of the repo, make sure you have Node 16 and pnpm 8, run `pnpm i && pnpm build`. That should complete without error. To run the test, `cd packages/cli`, then `pnpm test test/unit/commands/git.test.ts`.

bodinsamuel: Unfortunately, I still can't build, I get this error when building `@vercel/edge`, which I assume comes from `esbuild`  😩 It might be because I have a fresh install and dependencies are not fixed in tsup but I don't know much about this library.


```
✘ [ERROR] Cannot start service: Host version "0.14.54" does not match binary version "0.14.47"
```

packages/cli/test/unit/commands/git.test.ts
```diff
@@ -65,6 +65,58 @@ describe('git', () => {
...
+          'Which scope should contain your project?'
+        );
+        client.stdin.write('\r');
+
+        await expect(client.stderr).toOutput('Found project');
+        client.stdin.write('y\n');
+
+        await expect(client.stderr).toOutput(
+          `Connecting Git remote: https://github.com/user2/repo2.git`
+        );
```
TooTallNate:
```suggestion
        await expect(client.stderr).toOutput(
          `Do you still want to connect https://github.com/user2/repo2?`
        );
        client.stdin.write('y\n');

        await expect(client.stderr).toOutput(
          `Connecting Git remote: https://github.com/user2/repo2`
        );
```

TooTallNate (changes requested):
This test wasn't passing for me. There's an additional confirmation prompt that needs to be approved.

TooTallNate (approved):

TooTallNate:
Thank you @bodinsamuel!

bodinsamuel:
Thanks a lot for taking over @TooTallNate 
I wish I could have finished it but the tooling was not on my side :/

PR #9975, opened 2023-05-17
ensure cleanup after the gatsby plugin runs

EndangeredMassa:
There is an issue where you can interrupt a gatsby project build and the temporary files are not cleaned up. If you try to build again, it will get stuck in an infinite loop.

This PR adds an exit hook to make sure the cleanup happens.

### DIFF
--- a/packages/static-build/src/utils/gatsby.ts
+++ b/packages/static-build/src/utils/gatsby.ts
@@ -23,6 +23,8 @@ const PLUGIN_PATHS: Record<PluginName, string> = {
   ),
 };
 
+let GLOBAL_EXIT_HANDLER: undefined | (() => void);
+
 export async function injectPlugins(
   detectedVersion: string | null,
   dir: string
@@ -34,9 +36,12 @@ export async function injectPlugins(
     if (version && semver.satisfies(version, '>=4.0.0')) {
       plugins.add('@vercel/gatsby-plugin-vercel-builder');
 
-      process.on('exit', async () => {
-        cleanupGatsbyFiles(dir);
-      });
+      if (!GLOBAL_EXIT_HANDLER) {
+        GLOBAL_EXIT_HANDLER = () => {
+          cleanupGatsbyFiles(dir);
+        };
+        process.on('exit', GLOBAL_EXIT_HANDLER);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/static-build/src/utils/gatsby.ts
```diff
@@ -311,32 +321,39 @@ module.exports = gatsbyNode;
...
-        fs.readFile(backupFilePath, 'utf-8').catch(() => null),
-      ]);
-
-      if (
-        baseFileContent &&
-        baseFileContent.startsWith(GENERATED_FILE_COMMENT)
-      ) {
-        await fs.rm(baseFilePath);
+// must remain sync because it's called in an exit handler
+export function cleanupGatsbyFiles(dir: string) {
```
EndangeredMassa, line 325: This had to be made sync to be executable in a process exit hook.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -33,6 +36,13 @@ export async function injectPlugins(
     const version = semver.coerce(detectedVersion);
     if (version && semver.satisfies(version, '>=4.0.0')) {
       plugins.add('@vercel/gatsby-plugin-vercel-builder');
+
+      if (!GLOBAL_EXIT_HANDLER) {
+        GLOBAL_EXIT_HANDLER = () => {
+          cleanupGatsbyFiles(dir);
+        };
+        process.on('exit', GLOBAL_EXIT_HANDLER);
+      }
```
EndangeredMassa, line 45: I don't like adding exit hooks deep inside some logic, but we do try to make sure it's cleaned up later. If anyone has a suggestion on a better way to do this, please let me know.

packages/static-build/src/utils/gatsby.ts
```diff
@@ -353,3 +370,19 @@ export async function createPluginSymlinks(dir: string) {
...
+  try {
+    return fs.readFileSync(filePath, 'utf-8');
+  } catch (error: unknown) {
+    if (!isErrnoException(error)) {
+      console.error(error);
+    } else {
+      if (error.code !== 'ENOENT') {
+        console.error(error);
+      }
+    }
```
TooTallNate:
```suggestion
    if (!isErrnoException(error) || error.code !== 'ENOENT') {
      console.error(error);
    }
```

TooTallNate (dismissed):

Ethan-Arrowood (dismissed):

packages/static-build/src/utils/gatsby.ts
cb1kenobi: Maybe the files could be cleaned up before building?

packages/static-build/src/utils/gatsby.ts
EndangeredMassa: I think this is clearer: [`f6d2c12` (#9975)](https://github.com/vercel/vercel/pull/9975/commits/f6d2c121fb0e5b4d0645cf74e24078c30e674b49)

Ethan-Arrowood (approved):

packages/static-build/src/utils/gatsby.ts
EndangeredMassa: We really don't want to leave the user's project code in a temporary state. I think cleaning them up in process is best.

PR #9976, opened 2023-05-18
Deprecate Node.js 14.x and 16.x with warning

EndangeredMassa:
Node.js 14 and 16 are being deprecated. This PR will cause a warning when they are used today and an error after August 15.


styfle (dismissed):
Thanks!

cb1kenobi (approved):

Ethan-Arrowood (approved):

PR #9979, opened 2023-05-18
clarify next.js dupe api directory warning

EndangeredMassa:
In a Next.js app, if you have Node.js Functions in both `api` and `pages/api`, you'll get a warning about moving those `api` functions into `pages/api`. This does specify Node.js Functions already, but a user was confused and tried to move their non-Node.js functions into `pages/api` as well.

This PR tries to make it clearer that you can do both. Note that this still requires a non-ideal local dev setup where you:

- update next config to fall back to the `vc dev` port for `api` handlers
- run `next dev` for the Next.js app on one port
- run `vc dev` for the top-level `api` Functions on another port

---

Example repo: https://github.com/rexfordessilfie/vercel-api-languages


packages/fs-detectors/src/detect-builders.ts
```diff
@@ -339,7 +339,7 @@ export async function detectBuilders(
       warnings.push({
         code: 'conflicting_files',
         message:
-          'When using Next.js, it is recommended to place Node.js Serverless Functions inside of the `pages/api` (provided by Next.js) directory instead of `api` (provided by Vercel).',
+          'When using Next.js, it is recommended to place Node.js Functions inside of the `pages/api` (provided by Next.js) directory instead of `api` (provided by Vercel). Other (non-Node.js) Functions should still go in the `api` directory.',
```
styfle:
```suggestion
          'When using Next.js, it is recommended to place JavaScript Functions inside of the `pages/api` (provided by Next.js) directory instead of `api` (provided by Vercel). Other languages (Python, Go, etc) should still go in the `api` directory.',
```

Maybe we use JavaScript since both Node.js vs Edge Runtime should both be handled by Next.js

This message is starting to get really confusing now that we have `app` dir which can have Route Handlers at any path.

EndangeredMassa:
```suggestion
          'When using Next.js, it is recommended to place JavaScript/TypeScript Functions inside of the `pages/api` (provided by Next.js) directory instead of `api` (provided by Vercel). Other languages (Python, Go, etc) should still go in the `api` directory.',
```

Makes sense to me! Should we go ahead and mention TypeScript here as well?

TooTallNate (approved):

styfle (approved):
Let's ship this and then think about how we can improve the error message when using `app` with Route Handlers

unicornware:
@EndangeredMassa @styfle 

does this warning respect the `framework` option in `vercel.json`? i have a `pages` directory, but i'm **not using next.js**.

```sh
WARN! When using Next.js, it is recommended to place JavaScript Functions inside of the `pages/api` (provided by Next.js) directory instead of `api` (provided by Vercel). Other languages (Python, Go, etc) should still go in the `api` directory.
Learn More: https://nextjs.org/docs/api-routes/introduction
```

my terminal is also cluttered because the warning is printed on every request when running `vc dev`.

PR #9984, opened 2023-05-19
[cli] New `vc promote` command

cb1kenobi:
~This PR is blocked by https://github.com/vercel/api/pull/19508.~

Linear: https://linear.app/vercel/issue/VCCLI-262/cli-new-command-to-promote-deployment


internals/types/index.d.ts
```diff
@@ -353,7 +353,11 @@ export interface Project extends ProjectSettings {
   link?: ProjectLinkData;
   alias?: ProjectAliasTarget[];
   latestDeployments?: Partial<Deployment>[];
-  lastRollbackTarget: RollbackTarget | null;
+  lastAliasRequest?: LastAliasRequest;
+  /**
+   * @deprecated - Use `lastAliasRequest` instead
+   */
+  lastRollbackTarget?: RollbackTarget;
```
EndangeredMassa: **question (non-blocking):** ‏can `lastRollbackTarget` still be `null` sometimes?

internals/types/index.d.ts
```diff
@@ -388,20 +392,35 @@ export type ProjectLinkResult =
...
   requestedAt: number;
   toDeploymentId: string;
 }
 
+export interface LastAliasRequest {
+  fromDeploymentId: string;
+  jobStatus: 'pending' | 'in-progress' | 'succeeded' | 'failed' | 'skipped';
+  requestedAt: number;
+  toDeploymentId: string;
+  type: string;
```
EndangeredMassa: **suggestion (soft-blocking):** use literal types

In this or a follow-up PR: Can we change this tothe literal types to be more specific?‏

```suggestion
  type: 'rollback' | 'promote';
```

packages/cli/src/commands/promote.ts
```diff
@@ -0,0 +1,123 @@
...
+import handleError from '../util/handle-error';
+import { isErrnoException } from '@vercel/error-utils';
+import logo from '../util/output/logo';
+import ms from 'ms';
+import requestPromote from '../util/promote/request-promote';
+import promoteStatus from '../util/promote/status';
+
+const help = () => {
+  console.log(`
+  ${chalk.bold(`${logo} ${getPkgName()} promote`)} [deploymentId|deploymentName]
```
EndangeredMassa: **question (soft-blocking):** What's a `deploymentName`?‏

packages/cli/src/commands/promote.ts
```diff
@@ -0,0 +1,123 @@
...
+ * `vc promote` command
+ * @param {Client} client
+ * @returns {Promise<number>} Resolves an exit code; 0 on success
+ */
+export default async (client: Client): Promise<number> => {
+  let argv;
+  try {
+    argv = getArgs(client.argv.slice(2), {
+      '--debug': Boolean,
+      '-d': '--debug',
```
EndangeredMassa: **suggestion (soft-blocking):** remove common args

In this or a follow-up PR: The `--debug` option is a common arg. It shouldn't be necessary to list it here. Can we remove it?
‏
```suggestion
```

packages/cli/src/commands/promote.ts
```diff
@@ -0,0 +1,123 @@
...
+import type Client from '../util/client';
+import getArgs from '../util/get-args';
+import getProjectByCwdOrLink from '../util/projects/get-project-by-cwd-or-link';
+import { getPkgName } from '../util/pkg-name';
+import handleError from '../util/handle-error';
+import { isErrnoException } from '@vercel/error-utils';
+import logo from '../util/output/logo';
+import ms from 'ms';
+import requestPromote from '../util/promote/request-promote';
+import promoteStatus from '../util/promote/status';
```
EndangeredMassa: **suggestion (soft-blocking):** move utils to command directory

In this or a follow-up PR: ‏Can we colocate these files with the only place that uses them? They should be in `packages/cli/src/commands/promote/*` and the current `promote.ts` could be the `indext.ts` in that directory.

packages/cli/src/util/promote/status.ts
```diff
@@ -0,0 +1,265 @@
...
+ * Continuously checks a deployment status until it has succeeded, failed, or
+ * taken longer than the timeout (default 3 minutes).
+ *
+ * @param {Client} client - The Vercel client instance
+ * @param {string} [contextName] - The scope name; if not specified, it will be
+ * extracted from the `client`
+ * @param {Deployment} [deployment] - Info about the deployment which is used
+ * to display different output following a promotion request
+ * @param {Project} project - Project info instance
+ * @param {string} [timeout] - Milliseconds to poll for succeeded/failed state
```
EndangeredMassa: **suggestion (soft-blocking):** clarify value

In this or a follow-up PR: Can we say something more specific than "Milliseconds" here? That's not strictly true. It can be any format that `ms` accepts.
‏
```suggestion
 * @param {string} [timeout] - Milliseconds (or `ms` shorthand) to poll for succeeded/failed state
```

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+import { Request, Response } from 'express';
+import promote from '../../../src/commands/promote';
+import { LastAliasRequest } from '@vercel-internals/types';
+import { setupUnitFixture } from '../../helpers/setup-unit-fixture';
+import { useDeployment } from '../../mocks/deployment';
+import { useTeams } from '../../mocks/team';
+import { useUser } from '../../mocks/user';
+import sleep from '../../../src/util/sleep';
+
+jest.setTimeout(60000);
```
EndangeredMassa, line 13: **suggestion (non-blocking):** ‏comment about why

When increasing the test timeout, I think we should leave a comment about why we need the longer timeout.

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+    await expect(exitCodePromise).resolves.toEqual(1);
+  });
+
+  it('should error if invalid deployment name', async () => {
+    const { cwd } = initPromoteTest();
+    client.setArgv('promote', '????', '--yes', '--cwd', cwd);
+    const exitCodePromise = promote(client);
+
+    await expect(client.stderr).toOutput(
+      'Error: The provided argument "????" is not a valid deployment'
```
EndangeredMassa, line 42: **suggestion (non-blocking):** ‏change error messages to say "deployment ID"

I think we should change all instances of this error message to say "not a valid deployment ID" instead. That makes more sense.

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+  it('should error if timeout is invalid', async () => {
+    const { cwd } = initPromoteTest();
+    client.setArgv('promote', '--yes', '--cwd', cwd, '--timeout', 'foo');
+    const exitCodePromise = promote(client);
+
+    await expect(client.stderr).toOutput('Error: Invalid timeout "foo"');
+    await expect(exitCodePromise).resolves.toEqual(1);
+  });
+
+  it('should error if invalid deployment name', async () => {
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
  it('should error if invalid deployment ID', async () => {
```

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    await expect(client.stderr).toOutput('Fetching deployment "foo" in ');
+    await expect(client.stderr).toOutput(
+      'Error: Error: Can\'t find the deployment "foo" under the context'
+    );
+
+    await expect(exitCodePromise).resolves.toEqual(1);
+  });
+
+  it('should show status when not rolling back', async () => {
```
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
  it('should show status when not promoting', async () => {
```

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+    const { cwd, previousDeployment, project } = initPromoteTest({
+      promotePollCount: 10,
+    });
+
+    // start the promote
+    client.setArgv('promote', previousDeployment.id, '--yes', '--cwd', cwd);
+    promote(client);
+
+    // need to wait for the promote request to be accepted
+    await sleep(500);
```
EndangeredMassa: **suggestion (non-blocking):** ‏poll for the change

Instead of sleeping, can we poll (with a timeout) for the change we expect to see?

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+      `Fetching deployment "${previousDeployment.id}" in ${previousDeployment.creator?.username}`
+    );
+
+    // we need to wait a super long time because fetch will return on 500
+    await expect(client.stderr).toOutput('Response Error (500)', 20000);
+
+    await expect(exitCodePromise).resolves.toEqual(1);
+  });
+
+  it('should error if rollback fails (no aliases)', async () => {
```
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
  it('should error if promote fails (no aliases)', async () => {
```

packages/cli/test/unit/commands/promote.test.ts
```diff
@@ -0,0 +1,380 @@
...
+      promotePollCount: 10,
+    });
+    client.setArgv(
+      'promote',
+      previousDeployment.id,
+      '--yes',
+      '--cwd',
+      cwd,
+      '--timeout',
+      '2s'
```
EndangeredMassa: **suggestion (non-blocking):** ‏make timeout smaller

Can we reduce the timeout to something trivially small to reduce the amount of fixed waiting time the tests have?

EndangeredMassa (dismissed):

internals/types/index.d.ts
cb1kenobi: Yeah, I think `api` could return `null`. Better safe than sorry. Fixed!

packages/cli/src/commands/promote.ts
cb1kenobi: Good question! I copied this from rollback and I fixed the examples, but missed the usage. I've also fixed the help usage in `vc rollback`.

packages/cli/src/commands/promote.ts
cb1kenobi: Done!

packages/cli/test/unit/commands/promote.test.ts
cb1kenobi: We can't poll here in the test. It's testing the internal polling of the promote status. The promote command will poll every 250ms. Setting any lower things get flakey. I set it to 300ms.

packages/cli/test/unit/commands/promote.test.ts
cb1kenobi: I lowered it from 2s to 1s and then down to 1ms.

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #9988, opened 2023-05-19
docs(sveltekit): re-introduce speed insights

feugy:
### What's in there?

A customer recently leverage an outdated sveltekit example of ours to enable speed insights on a SvelteKit application, and had some issue with the `VERCEL_ANALYTICS_ID` env variable.

While migrating from pre-SvelteKit 1 to latest SvelteKit it seems that we removed the webvitals handling.
This PR brings back the web vitals collection, which allows displaying speed insights once deployed and enabled on Vercel.

### How to test?

Run the full example on Vercel with Speed insights enabled:

1. Extract the content of `vercel/examples/sveltekit-1` into its own folder
2. Install dependencies `pnpm i`
3. Deploy to vercel `vc deploy`
4. On Vercel dashboard, browse to the "Speed Insights" tab, enable and hit the redeploy button
5. Visit your application and browse back-and-forth the about page
6. Back to Vercel dashboard, insights should appear eventually 

image

### Notes to reviewers

The previous implementation was leveraging vite to read `VERCEL_ANALYTICS_ID` from the environment, and use a replacement plugin to inject into the bundled code, at build time.

This time we're leveraging Vercel-adapter, which makes [Vercel env variables available to the sveltekit application](https://kit.svelte.dev/docs/adapter-vercel#environment-variables), through Svelte native module.

Note: if you wonder why we're using `$env/dynamic/private` and not  `$env/dynamic/public`, this is because we can't  change the variable name to be prefixed with `PUBLIC_` (or anything user would configure with `config.kit.env.publicPrefix`). When kit would support a list of variable prefix, then we'll revisit this example.


examples/sveltekit-1/package.json
```diff
@@ -16,19 +16,21 @@
 		"@fontsource/fira-mono": "^4.5.10",
 		"@neoconfetti/svelte": "^1.0.0",
 		"@playwright/test": "^1.28.1",
-		"@sveltejs/adapter-vercel": "^1.0.0",
-		"@sveltejs/kit": "^1.0.0",
+		"@sveltejs/adapter-auto": "^2.0.0",
```
dummdidumm: We install adapter-vercel already, we don't need adapter-auto then
```suggestion
```

dummdidumm (dismissed):

examples/sveltekit-1/vite.config.js
```diff
@@ -1,11 +1,9 @@
 import { sveltekit } from '@sveltejs/kit/vite';
+import { defineConfig } from 'vitest/config';
```
dominiksipowicz, line 2: shouldn't it be in `vitest.config.ts` and not main `vite.config.js`?

dominiksipowicz (commented):
Is it possible to add Vercel deploy button?

ref: https://vercel.com/docs/integrations/deploy-button

examples/sveltekit-1/vite.config.js
feugy: This comes from the scafolding tool when creating a new app with `pnpm create svelte@latest` and enabling unit test.
[Source](https://github.com/sveltejs/kit/blob/master/packages/create-svelte/shared/%2Bvitest/vite.config.ts#L2)

dominiksipowicz (dismissed):

examples/sveltekit-1/src/routes/+layout.svelte
```diff
@@ -1,6 +1,20 @@
...
+	import { browser } from '$app/environment';
+	import { page } from '$app/stores';
+	import { webVitals } from '$lib/vitals';
 	import Header from './Header.svelte';
 	import './styles.css';
+
+	/** @type {import('./$types').LayoutServerData} */
+	export let data;
+
+	if (browser && data?.analyticsId) {
```
dummdidumm: I just realized that you probably want to rerun this on every page change (in the client), not just on first page load - in that case the statement needs to be reactive to fire on changes. [the old example](https://github.com/vercel/vercel/blob/main/examples/sveltekit/src/routes/%2Blayout.svelte#LL10C3-L10C6) also does it this way
```suggestion
	$: if (browser && data?.analyticsId) {
```

examples/sveltekit-1/package.json
dummdidumm: We install adapter-vercel already, we don't need adapter-auto then
```suggestion
```

StephDietz (approved):

dummdidumm (approved):

PR #9989, opened 2023-05-19
[next] Ensure app functions are detected/separated properly

ijjk:
Follow-up to https://github.com/vercel/vercel/pull/9974 this uses the `lambdaAppPaths` as the source of truth instead of the manifest which is more accurate for separating/detecting. Also adds additional test cases for new root and index app functions. 

x-ref: https://github.com/vercel/next.js/issues/49169

### DIFF
--- a/.changeset/tiny-melons-bow.md
+++ b/.changeset/tiny-melons-bow.md
@@ -1,0 +1,5 @@
+---
+"@vercel/next": patch
+---
+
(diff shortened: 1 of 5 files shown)


shuding (approved):

EndangeredMassa (approved):
I lack context, but if you manually tested this and it fixes the issue, this looks good to me.

PR #10001, opened 2023-05-22
[examples] Update "vue" template to use Node 18

swarnava:


examples/vue/package.json
```diff
@@ -6,20 +6,20 @@
     "lint": "vue-cli-service lint"
   },
   "engines": {
-    "node": "16.x"
+    "node": "18.x"
```
swarnava, line 9: should we pin it?

TooTallNate (approved):

PR #10004, opened 2023-05-22
Create release PR as "github-actions" bot

TooTallNate:
There's no need to run the tests for the release PR, so create the PR as "github-bot" instead of "vercel-release-bot", which will prevent Actions from being executed on the release PR. The consequence of this is that we will have to force-merge since the required tests will not be passing.

This reverts commit 5825e3070046f090cbca998b82a8e01261559e79 (#9932).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
           version: pnpm version:prepare
           publish: pnpm release
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_CONFIG_PROVENANCE: 'true'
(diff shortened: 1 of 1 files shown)


EndangeredMassa:
This means that only admins will be able to release, correct?

styfle:
> The consequence of this is that we will have to force-merge since the required tests will not be passing.

Lets also update tests to skip in this case?

Add a `if` statement like this: https://github.com/vercel/vercel/blob/7ff321310f5d8d2224947a87ddfe2125b7abf5f3/.github/workflows/test.yml#L55

TooTallNate:
@styfle I don't think that would do anything. The tests already wouldn't be run at all since Actions are not executed for github-actions bot PRs ([example](https://github.com/vercel/vercel/pull/9931)).

Overall I'm kind of ±0 on this change. The tests running doesn't hurt anything (we can still force merge if needed ourselves), and if this would make it harder/impossible for non-admins to release then that seems like a loss to me.

PR #10011, opened 2023-05-23
[examples] Update preact template to use Node 18

swarnava:


.changeset/thick-colts-jam.md
swarnava, line 1: I have no idea how this was added, is this intentional?

swarnava:
Opened #10011

PR #10013, opened 2023-05-23
[cli] Add repo linking support for `deploy` command

TooTallNate:
Adds support for `vercel deploy` command when the repository has been linked via `vercel link --repo`.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -357,6 +345,12 @@ export default async (client: Client): Promise<number> => {
     }
   }
 
+  // For repo-style linking, reset the path to the root of the repository
+  //console.log(link);
+  if (link.status === 'linked' && link.repoRoot) {
+    cwd = join(link.repoRoot);
```
cb1kenobi: I don't think the `join()` is doing anything. Any truthy value for `repoRoot` will be just end up being returned by `join()`. Did you mean `resolve()`?

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -72,10 +70,9 @@ export default async function processDeployment({
     rootDirectory,
   } = args;
 
-  const { debug } = output;
-
+  const client = now._client;
```
cb1kenobi, line 73: Can remove the `_` and make `client` a public property?

packages/cli/src/util/link/repo.ts
```diff
@@ -270,3 +282,13 @@ export function findProjectFromPath(
       );
     });
 }
+
+/**
+ * TODO: remove
```
cb1kenobi, line 287: Out of curiosity, remove when and why?

packages/cli/src/util/projects/get-project-by-cwd-or-link.ts
```diff
@@ -26,27 +25,10 @@ export default async function getProjectByCwdOrLink({
...
-  }
-
   // ensure the current directory is a linked project
-  const linkedProject = await ensureLink(
-    commandName,
-    client,
-    pathValidation.path,
-    { autoConfirm }
-  );
+  const linkedProject = await ensureLink(commandName, client, cwd, {
```
cb1kenobi, line 29: **nitpick (non-blocking):** ‏I think the old formatting with params on separate lines is easier to read that the changed code.

packages/cli/src/util/validate-paths.ts
```diff
@@ -1,14 +1,11 @@
-import { lstat as lstatRaw } from 'fs';
-import { promisify } from 'util';
+import { lstat } from 'fs-extra';
```
cb1kenobi, line 1: **convention (non-blocking):** ‏What do you think about using the built-in Node.js version?

```suggestion
import { lstat } from 'fs/promises';
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -357,6 +345,12 @@ export default async (client: Client): Promise<number> => {
     }
   }
 
+  // For repo-style linking, reset the path to the root of the repository
+  //console.log(link);
```
EndangeredMassa: **suggestion (soft-blocking):** 
‏
```suggestion
```

EndangeredMassa (commented):
Looking good!

packages/cli/src/commands/deploy/index.ts
TooTallNate: You're right, good catch.

```suggestion
    cwd = link.repoRoot;
```

packages/cli/src/util/deploy/process-deployment.ts
TooTallNate: Fine by me. Let's do in a separate PR though (would love to refactor out the entire legacy `Now` client).

packages/cli/src/util/link/repo.ts
TooTallNate: I plan on addressing the use-case where there's multiple projects with the same root directory (i.e. Storybook) in the next PR. Thus, the singular project match function will go away.

packages/cli/src/util/projects/get-project-by-cwd-or-link.ts
TooTallNate: Prettier <picture data-single-emoji=":blob_shrug:" title=":blob_shrug:">:blob_shrug:</picture>

packages/cli/src/util/validate-paths.ts
TooTallNate: I don't think it makes much difference since we're already using `fs-extra` throughout the code base.

cb1kenobi (approved):
LGTM. Nice work, especially getting all the path stuff cleaned up!

PR #10015, opened 2023-05-24
[next] Pass `pageExtensions` data to `apiLambdaGroups`

smaeda-ks:
Right now, we can't detect API routes correctly if `pageExtensions` is set:

> WARNING: Unable to find source file for page xxxx with extensions: js, jsx, ts, tsx, this can cause functions config from `vercel.json` to not be applied


ijjk (dismissed):
Good catch!

smaeda-ks:
@ijjk Thanks for approving! I had to fix the merge conflict with upstream, which dismissed your review. Could you approve again and merge this PR? Thanks!

styfle (dismissed):

styfle (approved):

PR #10019, opened 2023-05-24
[cli] Clean up 'vc rollback'

cb1kenobi:
This PR removes dependency on the deprecated `lastRollbackTarget` project property and adopts many of the code conventions used in `vc promote`.

Important! Please merge #9984 first!


packages/cli/src/commands/rollback/index.ts
```diff
@@ -74,48 +73,49 @@ export default async (client: Client): Promise<number> => {
...
       timeout,
     });
-  }
+  } catch (err) {
+    if (isErrnoException(err)) {
+      if (err.code === 'ERR_CANCELED') {
+        return 0;
+      }
+      if (err.code === 'ERR_INVALID_CWD' || err.code === 'ERR_LINK_PROJECT') {
+        // do not show the message
```
EndangeredMassa, line 113: **suggestion (non-blocking):** ‏Can we debug log the error here?

EndangeredMassa (dismissed):

packages/cli/src/commands/rollback/index.ts
cb1kenobi: No need, `validatePaths()` already prints an error: https://github.com/vercel/vercel/blob/main/packages/cli/src/util/validate-paths.ts#L72. These error codes are in theory temporary until we refactor the code to throw instead of returning the exit code from functions.

EndangeredMassa (approved):

PR #10021, opened 2023-05-24
[cli] Add `expect` dev dependency to fix type error in `toOutput`

TooTallNate:
Also make the timeout print what was buffered to help with debugging failed tests.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -119,6 +119,7 @@
     "escape-html": "1.0.3",
     "esm": "3.1.4",
     "execa": "3.2.0",
+    "expect": "29.5.0",
     "express": "4.17.1",
     "fast-deep-equal": "3.1.3",
     "find-up": "4.1.0",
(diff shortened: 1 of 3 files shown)


styfle (commented):
Bummer its 5MB just for a type https://packagephobia.com/result?p=expect@29.5.0

cb1kenobi:
Where is the type error being displayed? When tests are being run?

TooTallNate:
@styfle It's not a new dependency though (it's a transient dependency of `jest`). In fact, if you look at the lockfile, this actually _removes_ an older version of `expect` and a few others. The red squiggly was probably a side-effect of switching over to pnpm.

TooTallNate:
@cb1kenobi You normally wouldn't see it, since we don't validate TS errors for test files. But it's a red squiggly in the `toOutput.ts` file.

cb1kenobi:
@TooTallNate I see it now! It's right there!

cb1kenobi (approved):

EndangeredMassa:
This version doesn't affect the build container.

PR #10028, opened 2023-05-26
[node] add `maxDuration` config support for vc node deployments

ecklf:
This PR enables specifying `maxDuration` in `config` for node vc deployments.


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -371,6 +371,13 @@ async function writeLambda(
   }
 
   const memory = functionConfiguration?.memory ?? lambda.memory;
+
+  if (functionConfiguration?.maxDuration && lambda.maxDuration) {
+    throw new Error(
+      `Cannot specify both \`maxDuration\` in \`vercel.json\` and \`${path}\``
```
styfle: Why does `maxDuration` throw but `memory` on line 373 doesn't?

We should be consistent for all config options.

packages/node/src/index.ts
```diff
@@ -506,6 +506,9 @@ export const build: BuildV3 = async ({
       shouldAddSourcemapSupport,
       awsLambdaHandler,
       supportsResponseStreaming,
+      ...(staticConfig?.maxDuration
+        ? { maxDuration: staticConfig.maxDuration }
+        : {}),
```
styfle: Is it necessary to spread? Can you do this instead:

```suggestion
      maxDuration: staticConfig?.maxDuration
```

packages/cli/src/util/build/write-build-result.ts
ecklf: Defining in both the `vercel.json` and config seems to not throw an error but silently have the value in `vercel.json` take precedent. Should we retain this behavior or diff between those settings and throw an error?

packages/node/src/index.ts
ecklf: Yeah we can also do that 👍🏻

packages/cli/src/util/build/write-build-result.ts
styfle: I'm not sure if we want to break existing deployments though.

This feature was added a year ago in https://github.com/vercel/vercel/pull/8243

@TooTallNate @EndangeredMassa Thoughts here?

cb1kenobi: I don't think we should have this throw. It's a breaking change. We should document that the `vercel.json` overrides the function's config.

I advise against logging a warning in the event both are defined as it could make the output noisy. Documentation should clearly describe the behavior.

If we were to introduce any breaking change, I would have the function's config override the `vercel.json` and document this change.

TooTallNate (changes requested):
Needs a changeset

packages/cli/src/util/build/write-build-result.ts
TooTallNate: I think we should swap the order actually. The `lambda` value would take precedence. Reason being that the `lambda` value presumably comes from the `export const config` which is more specific/explicit (individual file level) than the `vercel.json` which is a glob pattern. Plus we are trying to encourage config-in-code more anyways.

This would technically be a breaking change, but since there is currently no way to define those configs on `lambda` anyways aside from `runtime`, it would be net-new behavior and thus non-breaking.

javivelasco: Agree with @TooTallNate

ecklf: Retaining the existing behavior in this PR, let's ship another PR with that breaking change of having the lambda config take priority?

ecklf:
> Needs a changeset

Added a patch one, LMK if it should be minor.

cb1kenobi (dismissed):

.changeset/tricky-mangos-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/node': patch
```
TooTallNate:
```suggestion
'@vercel/node': minor
```

styfle (approved):

.changeset/tricky-mangos-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/node': minor
+---
+
+Add maxDuration config support for vc node deployments
```
nuta, line 5: Super nitpick: Can we replace `[edge-compute]` in the PR title with `[node]`? 🙏

nuta (approved):
LGTM 👍 Left a nitpick.

ecklf:
Merging because the requested change from Nate was applied.

PR #10031, opened 2023-05-26
[cli] Add `client.cwd` to unify all working directory related logic

TooTallNate:
A few commands were still checking on `--cwd` explicitly, which is incorrect since the entrypoint file already handles the directory change.

The new `client.cwd` property is a helper to make writing tests easier. Tests no longer need to `chdir()` explicitly and then revert afterwards.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -144,12 +144,6 @@ const main = async () => {
     return 1;
   }
 
-  let cwd = argv['--cwd'];
-  if (cwd) {
-    process.chdir(cwd);
-  }
-  cwd = process.cwd();
-
   // The second argument to the command can be:
   //
   //  * a path to deploy (as in: `vercel path/`)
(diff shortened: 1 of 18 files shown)


packages/cli/src/index.ts
```diff
@@ -144,12 +144,6 @@ const main = async () => {
     return 1;
   }
 
-  let cwd = argv['--cwd'];
-  if (cwd) {
-    process.chdir(cwd);
-  }
-  cwd = process.cwd();
```
styfle, line 151: Don't we need to keep this here so that `await mkdirp(VERCEL_DIR)` uses the correct cwd?

packages/cli/src/util/client.ts
```diff
@@ -207,4 +207,12 @@ export default class Client extends EventEmitter implements Stdio {
       output: this.stderr as NodeJS.WriteStream,
     });
   }
+
+  get cwd(): string {
+    return process.cwd();
+  }
+
+  set cwd(v: string) {
+    process.chdir(v);
```
cb1kenobi, line 216: This is brilliant, love it!

packages/cli/src/index.ts
TooTallNate: No, `VERCEL_DIR` is the global dir.

styfle (approved):

PR #10039, opened 2023-05-29
[next] Add id env handling

ijjk:
x-ref: [slack thread](https://vercel.slack.com/archives/C04TNLL0LSK/p1684879925060159)

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -200,6 +200,8 @@ export const build: BuildV2 = async ({
   entrypoint,
   config = {},
   meta = {},
+}: Parameters<BuildV2>[0] & {
+  config: Parameters<BuildV2>[0]['config'] & { deploymentId?: 'string' };
 }) => {
   validateEntrypoint(entrypoint);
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):

packages/next/src/index.ts
```diff
@@ -245,6 +247,13 @@ export const build: BuildV2 = async ({
     process.env[key] = value;
   }
 
+  // expose deploymentId if available, this isn't NEXT_PUBLIC
+  // prefixed as it shouldn't cause a nx/turbo cache miss if it'd
+  // otherwise be a cache hit
+  if (config.deploymentId) {
+    process.env.NEXT_DEPLOYMENT_ID = config.deploymentId;
+  }
```
styfle, line 255: I'm thinking we should do this in the build container instead of the builder since we already know the targeted framework and can prefix it appropriately.

ijjk: Yeah sounds good to keep it there

PR #10043, opened 2023-05-30
[frameworks] Update dev command for Sanity v3

unitof:
Sanity v2 is now deprecated, and [Sanity v3](https://www.sanity.io/help/studio-v2-vs-v3) prefers a new `sanity dev` command. `sanity start` results in following error/warning:

```
▶ vc dev
Vercel CLI 30.0.0
> Running Dev Command “sanity start --port $PORT”
╭───────────────────────────────────────────────────────────╮
│                                                           │
│  You're running Sanity Studio v3. In this version the     │
│  [start] command is used to preview static builds.        |
│                                                           │
│  To run a development server, use the [npm run dev] or    |
│  [npx sanity dev] command instead. For more information,  │
│  see https://www.sanity.io/help/studio-v2-vs-v3           │
│                                                           │
╰───────────────────────────────────────────────────────────╯

Could not find a production build in the '[Sanity project root]/dist' directory.
Try building your studio app with 'sanity build' before starting the preview server.
```

**But before anyone merges: would changing this retroactively change the dev command for all existing Sanity-detected projects on Vercel?** That would break Sanity v2 projects and I'd need to split into Sanity v2 and Sanity v3 frameworks.

Or does framework detection only run once on new project creation?

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1914,7 +1914,7 @@ export const frameworks = [
         value: 'sanity build',
       },
       devCommand: {
-        value: 'sanity start --port $PORT',
+        value: 'sanity dev --port $PORT',
       },
       outputDirectory: {
         value: 'dist',
--- a/.changeset/nervous-icons-wait.md
+++ b/.changeset/nervous-icons-wait.md
@@ -1,0 +1,5 @@
+---
+'@vercel/frameworks': patch
+---
+
+Update dev command for Sanity v3


packages/frameworks/src/frameworks.ts
```diff
@@ -1914,7 +1914,7 @@ export const frameworks = [
         value: 'sanity build',
       },
       devCommand: {
-        value: 'sanity start --port $PORT',
```
EndangeredMassa, line 1917: > But before anyone merges: would changing this retroactively change the dev command for all existing Sanity-detected projects on Vercel? That would break Sanity v2 projects and I'd need to split into Sanity v2 and Sanity v3 frameworks.

We do need to keep these separate. We should make another `sanity-3` entry that can configure v2 (`sanity`) and v3 (`sanity-3`) separately. We just have to decide how the `detectors` can accurately determine the version.

We can detect the `sanity` package version. Is that always going to be in a sanity project?

EndangeredMassa (changes requested):

packages/frameworks/src/frameworks.ts
unitof: Package version might be more accurate, but just throwing out that you could still use file detection. Sanity 2 uses `sanity.json`, while v3 migrated to `sanity.config.[js|ts|jsx|tsx]`.

There's a way to have the framework detector do conditional commands without creating multiple frameworks?

EndangeredMassa: Ahh, I saw that those files were already listed, but that's because [you added them earlier to support v3](https://github.com/vercel/vercel/pull/8540). Yeah, that sounds like a good way to detect the versions.

There is not currently a way to differentiate a command based on version. We'll have to create a new entry here.

unitof: Yep! That was me, turns out they did change commands after all.

So should the slug `sanity` become `sanity@latest` (currently Sanity 3), and create a new `sanity-2`?

Or keep `sanity` as Sanity v2, and make a new `sanity-3`? That seems to be what Docusaurus did.

EndangeredMassa: The original slug should stay the same. It seems that v3 is in error, let's make `sanity-3` for that.

trek:
Ah shoot @unitof I wish I'd seen this PR. We received another report and did https://github.com/vercel/vercel/pull/11991. Does this also resolve the issue for you?

unitof:
Yep, that seems to do the trick! Thanks for reaching out @trek, no worries.

PR #10044, opened 2023-05-30
Use `pnpm publish -r` and `changeset tag` to publish packages

TooTallNate:
This is a re-application of #10022 (which was reverted in #10032), but with the addition of `changeset tag` after the `pnpm publish -r` command. This ensures the proper git tags are also created, allowing for the GitHub releases to be published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,8 @@
     "prettier-check": "prettier --check .",
     "prepare": "husky install",
     "pack": "cd utils && node -r ts-eager/register ./pack.ts",
-    "version:prepare": "changeset version && pnpm install --no-frozen-lockfile",
-    "release": "changeset publish"
+    "ci:version": "changeset version && pnpm install --no-frozen-lockfile",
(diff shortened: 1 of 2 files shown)


styfle (approved):

package.json
```diff
@@ -48,8 +48,8 @@
     "prettier-check": "prettier --check .",
     "prepare": "husky install",
     "pack": "cd utils && node -r ts-eager/register ./pack.ts",
-    "version:prepare": "changeset version && pnpm install --no-frozen-lockfile",
-    "release": "changeset publish"
+    "ci:version": "changeset version && pnpm install --no-frozen-lockfile",
+    "ci:publish": "pnpm publish -r && changeset tag"
```
EndangeredMassa, line 52: Will this push the tags too? The docs are not clear.
https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#tag

TooTallNate: The command itself does not, but `changeset/action` pushes the tags after invoking this script.

PR #10048, opened 2023-05-31
[tests] Fix summary workflow to always run

styfle:
### Description
This ensures the Summary job always runs to check if any test failed and cause itself to fail as well.

This allows us to dynamically add as many concurrent jobs as we want but only have one (Summary) marked as required.

### Testing
I verified this worked by making a change to `@vercel/static-config` which is not required and the Summary check failed:

https://github.com/vercel/vercel/actions/runs/5136560652/jobs/9243924596

Then I reverted it so that the test was passing and the Summary check passed:

https://github.com/vercel/vercel/actions/runs/5137401343/jobs/9245554297


TooTallNate (dismissed):

.github/workflows/test.yml
```diff
@@ -97,11 +97,20 @@ jobs:
...
+    timeout-minutes: 5
+    if: always()
     needs:
       - test
-    runs-on: ubuntu-latest
-    name: E2E
     steps:
-      - name: Done
-        run: echo "Done."
+      - run: |-
```
TooTallNate: Should we add a name? Currently it shows as "Run for status in failure" or "Run for status in success"

styfle: Fixed in [2ff02a8](https://github.com/vercel/vercel/pull/10048/commits/2ff02a83a9200842e9fd81025469d5c054a56d4f)

PR #10049, opened 2023-05-31
[cli] Ensure .npmrc does not contain use-node-version

cb1kenobi:
When a project has a `.npmrc` containing `use-node-version`, package managers (notably `pnpm`) will download the specified Node.js version. This is not the correct way as it can lead to `pnpm` downloading Node.js 18 or newer which depends on a version of GLIBC that is not present in the current AWS image. The proper way is to set the `"engines"` in the `package.json`.

image

Discussion: https://github.com/orgs/vercel/discussions/2436


packages/cli/src/commands/deploy/index.ts
```diff
@@ -174,6 +175,13 @@ export default async (client: Client): Promise<number> => {
   let { path: cwd } = pathValidation;
   const autoConfirm = argv['--yes'];
 
+  try {
+    await validateNpmrc(cwd);
+  } catch (err) {
+    prettyError(err);
+    return 1;
+  }
```
styfle: This validation should happen in `vc build` (not `vc deploy`).

We might even want to move it all the way to `@vercel/build-utils` which where we invoke `npm install`

pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
styfle: Looks like you are using the wrong pnpm version.

Try `corepack enable pnpm` to select the correct version.

Also see https://styfle.dev/blog/corepack

packages/cli/src/commands/deploy/index.ts
cb1kenobi: Happy to add this validation to `vc build`, but I think we should keep it in `vc deploy` just to fail fast.

styfle: But vc deploy doesn’t run npm install. 

Imagine the user has overridden the Install Command - we don’t need to fail in that case.

styfle: But vc deploy doesn’t run npm install. 

Imagine the user has overridden the Install Command - we don’t need to fail in that case. 

Also this PR won’t solve the most common deployment case: git push

cb1kenobi: Agreed. OK, then I can move the validate function to `build-utils`, then call it wherever we install dependencies. This won't fail fast, but it should cover all the bases and improve the error message.

styfle (approved):
Thanks!

Jrwolf99:
Not sure if this is the place to ask, but after the release of 30.1.1, my build failed due to an unknown function. I think this was modified in this PR? My 30.1.0 build just fine. Do I need to update anything else?


`Run vercel build --token=*** Vercel CLI 30.1.1 Error: (0 , build_utils_1.validateNpmrc) is not a function Error: Process completed with exit code 1.`

cb1kenobi:
@Jrwolf99 Thank you very much for reporting this! A fix, and new CLI release, is underway.

cb1kenobi:
@Jrwolf99 [vercel@30.1.2](https://github.com/vercel/vercel/releases/tag/vercel%4030.1.2) has been released. Give it a go!

Jrwolf99:
It works! Thanks for the fast turnaround 😄

PR #10054, opened 2023-06-01
[cli] Add `client.fetchPaginated()` async iterator

TooTallNate:
Adds async iterator helper function for interacting with endpoints that have pagination.

Implemented for use with `vc bisect` (fetching deployments) and `vc link --repo` (fetching projects).


internals/types/index.d.ts
```diff
@@ -383,7 +383,7 @@ export interface ProjectLink {
 export interface PaginationOptions {
   prev: number;
   count: number;
-  next?: number;
+  next?: number | null;
```
EndangeredMassa: **question (non-blocking):** ‏Is there some type shuffling we can do to make this not both `null` and `undefined`?

EndangeredMassa (dismissed):

internals/types/index.d.ts
TooTallNate: From what I've seen so far I think the API will always return `null`. Will double check on that though.

EndangeredMassa: We could wrap it to get rid of one of the types. It's not a huge deal, but I do like simplifying these when they come up. I don't want to read the type later and wonder what the difference between `undefined` and `null` means for this property.

cb1kenobi (dismissed):

internals/types/index.d.ts
TooTallNate: Updated to use the same type as defined in API repo

EndangeredMassa (approved):

Issue #7009, opened 2021-11-16
vercel-cli failed to populate framework prefixed system envs for `vc dev` and `vc env pull`


wiiiimm:
_From Slack: https://vercel.slack.com/archives/CLDDX2Y0G/p1634630636165900_

According to https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables:

> Frameworks typically use a prefix in order to expose Environment Variables to the browser.
> The following prefixed Environment Variables will be available during the Build Step, based on the Project's selected Framework Preset.

> **_System Environment Variables will be populated with an empty string if the value is not applicable, for example when running vercel dev during Development or if the Deployment is not triggered by a Git Commit._**

Expected behaviour: `*_VERCEL_ENV` should be populated for `vc dev` and `vc env pull`.

### Next.js

Repo: https://github.com/vercel-support/48677-nextjs-env-cli-test/
Project: https://vercel.com/vercel-support/48677-nextjs-env-cli-test
Deployment: https://48677-nextjs-env-cli-test.vercel-support.app/

When doing `vc dev`, `NEXT_PUBLIC_VERCEL_ENV` is empty.
When running `vc env pull`, the following .env is created:

```
# Created by Vercel CLI
VERCEL="1"
VERCEL_ENV="development"
VERCEL_URL=""
VERCEL_GIT_PROVIDER=""
VERCEL_GIT_REPO_SLUG=""
VERCEL_GIT_REPO_OWNER=""
VERCEL_GIT_REPO_ID=""
VERCEL_GIT_COMMIT_REF=""
VERCEL_GIT_COMMIT_SHA=""
VERCEL_GIT_COMMIT_MESSAGE=""
VERCEL_GIT_COMMIT_AUTHOR_LOGIN=""
VERCEL_GIT_COMMIT_AUTHOR_NAME=""
NEXT_PUBLIC_ENV_TEST_1="Electricity-Anyway-Atmosphere-Growth-5"
```

### React

Repo: https://github.com/vercel-support/48677-react-env-cli-test/
Project: https://vercel.com/vercel-support/48677-react-env-cli-test
Deployment: https://48677-react-env-cli-test.vercel-support.app/

When doing `vc dev`, `REACT_APP_VERCEL_ENV` is empty.
When running `vc env pull`, the following .env is created:

```
# Created by Vercel CLI
VERCEL="1"
VERCEL_ENV="development"
VERCEL_URL=""
VERCEL_GIT_PROVIDER=""
[...]

PR #10064, opened 2023-06-05
[vc dev] Automatically expose framework system env vars

smaeda-ks:
Exposing `VERCEL_ENV` and `VERCEL_URL` to `vc dev`, alnong with their framework aliases.

- Related to #7009

### DIFF
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -544,6 +544,29 @@ export default class DevServer {
     return undefined;
   }
 
+  exposePrefixedEnvVars({
+    envPrefix,
+    envs,
+  }: {
+    envPrefix: string | undefined;
+    envs: Env;
+  }): Env {
+    // TODO: expose VERCEL_GIT_ env vars as well once we add support (VCCLI-303)
+    const allowed = ['VERCEL_URL', 'VERCEL_ENV'];
+    const newEnvs: Env = {};
+    if (envPrefix) {
+      Object.keys(envs).forEach(key => {
+        if (allowed.includes(key)) {
+          newEnvs[`${envPrefix}${key}`] = envs[key];
+        }
+        newEnvs[key] = envs[key];
+      });
+      return newEnvs;
+    } else {
+      return envs;
+    }
+  }
+
   async _getVercelConfig(): Promise<VercelConfig> {
     const configPath = getVercelConfigPath(this.cwd);
 
@@ -693,11 +716,33 @@ export default class DevServer {
     buildEnv['NOW_REGION'] = 'dev1';
     allEnv['NOW_REGION'] = 'dev1';
 
-    // mirror how VERCEL_REGION is injected in prod/preview
-    // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
-    // and should not contain VERCEL_REGION
     if (this.projectSettings?.autoExposeSystemEnvs) {
+      // mirror how VERCEL_REGION is injected in prod/preview
+      // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
+      // and should not contain VERCEL_REGION
       runEnv['VERCEL_REGION'] = 'dev1';
+
+      // expose VERCEL_ENV and VERCEL_URL to dev command
+      allEnv['VERCEL_ENV'] = 'development';
+      if (this.address.host) {
+        allEnv['VERCEL_URL'] = this.address.host;
+      }
+
+      // automaticaly expose VERCEL_URL, VERCEL_ENV
+      // as prefixed (framework) env vars
+      if (this.projectSettings?.framework) {
+        const frameworkSlug = this.projectSettings.framework;
+        const framework = frameworkList.find(f => f.slug === frameworkSlug);
+        const envPrefix = framework?.envPrefix;
+
+        if (envPrefix) {
+          [buildEnv, runEnv, allEnv] = await Promise.all([
+            this.exposePrefixedEnvVars({ envPrefix, envs: buildEnv }),
+            this.exposePrefixedEnvVars({ envPrefix, envs: runEnv }),
+            this.exposePrefixedEnvVars({ envPrefix, envs: allEnv }),
+          ]);
+        }
+      }
     }
 
     this.envConfigs = { buildEnv, runEnv, allEnv };
@@ -818,6 +863,8 @@ export default class DevServer {
         env['VERCEL_URL'] = this.address.host;
       } else if (name === 'VERCEL_REGION') {
         env['VERCEL_REGION'] = 'dev1';
+      } else if (name === 'VERCEL_ENV') {
+        env['VERCEL_ENV'] = 'development';
       }
     }
 
--- a/packages/cli/src/util/dev/server.ts
+++ b/packages/cli/src/util/dev/server.ts
@@ -37,6 +37,7 @@ import {
   FileFsRef,
   PackageJson,
   spawnCommand,
+  getPrefixedEnvVars,
 } from '@vercel/build-utils';
 import {
   detectBuilders,
@@ -489,7 +490,6 @@ export default class DevServer {
       const dotenv = await fs.readFile(filePath, 'utf8');
       this.output.debug(`Using local env: ${filePath}`);
       env = parseDotenv(dotenv);
-      env = this.injectSystemValuesInDotenv(env);
     } catch (err: unknown) {
       if (!isErrnoException(err) || err.code !== 'ENOENT') {
         throw err;
@@ -542,29 +542,6 @@ export default class DevServer {
       }
     }
     return undefined;
-  }
-
-  exposePrefixedEnvVars({
-    envPrefix,
-    envs,
-  }: {
-    envPrefix: string | undefined;
-    envs: Env;
-  }): Env {
-    // TODO: expose VERCEL_GIT_ env vars as well once we add support (VCCLI-303)
-    const allowed = ['VERCEL_URL', 'VERCEL_ENV'];
-    const newEnvs: Env = {};
-    if (envPrefix) {
-      Object.keys(envs).forEach(key => {
-        if (allowed.includes(key)) {
-          newEnvs[`${envPrefix}${key}`] = envs[key];
-        }
-        newEnvs[key] = envs[key];
-      });
-      return newEnvs;
-    } else {
-      return envs;
-    }
   }
 
   async _getVercelConfig(): Promise<VercelConfig> {
@@ -698,19 +675,14 @@ export default class DevServer {
       this.getLocalEnv('.env.build', vercelConfig.build?.env),
     ]);
 
-    let allEnv = { ...buildEnv, ...runEnv };
-
     // If no .env/.build.env is present, use cloud environment variables
-    if (Object.keys(allEnv).length === 0) {
+    if (Object.keys(runEnv).length + Object.keys(buildEnv).length === 0) {
       const envValues = { ...this.envValues };
-      if (this.address.host) {
-        envValues['VERCEL_URL'] = this.address.host;
-      }
-      allEnv = { ...envValues };
       runEnv = { ...envValues };
       buildEnv = { ...envValues };
     }
 
+    let allEnv = { ...buildEnv, ...runEnv };
     // legacy NOW_REGION env variable
     runEnv['NOW_REGION'] = 'dev1';
     buildEnv['NOW_REGION'] = 'dev1';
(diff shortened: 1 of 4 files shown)


packages/cli/src/util/dev/server.ts
```diff
@@ -544,6 +544,29 @@ export default class DevServer {
...
+  exposePrefixedEnvVars({
+    envPrefix,
+    envs,
+  }: {
+    envPrefix: string | undefined;
+    envs: Env;
+  }): Env {
+    // TODO: expose VERCEL_GIT_ env vars as well once we add support (VCCLI-303)
+    const allowed = ['VERCEL_URL', 'VERCEL_ENV'];
+    const newEnvs: Env = {};
```
styfle: This looks copy/pasta from `@vercel/build-utils`. Can we reuse the existing function?

smaeda-ks: Thanks! updated

packages/cli/src/util/dev/server.ts
```diff
@@ -693,11 +694,42 @@ export default class DevServer {
...
+              ...buildEnv,
+              ...getPrefixedEnvVars({ envPrefix: envPrefix, envs: buildEnv }),
+            },
+            {
+              ...runEnv,
+              ...getPrefixedEnvVars({ envPrefix: envPrefix, envs: runEnv }),
+            },
+            {
+              ...allEnv,
+              ...getPrefixedEnvVars({ envPrefix: envPrefix, envs: allEnv }),
```
styfle: I'm pretty sure you only need to inject this in `buildEnv` since the prefixed env vars are only provided at build time

smaeda-ks: @styfle setting these system env vars to `buildEnv` and doing this only for `buildEnv` didn't make them available to `vc dev` unless I assign them to `allEnv` this way (referencing them in user code always output `undefined`).

styfle: You'll need to assign to `buildEnv` prior to `allEnv` being created

styfle: In fact, that looks like a bug with the existing `VERCEL_REGION` too

packages/cli/test/dev/fixtures/10a-nextjs-routes/pages/api/env.js
```diff
@@ -0,0 +1,3 @@
+export default (_req, res) => {
+  res.send(process.env);
```
styfle, line 2: This will only test runtime env, not build time env, right?

packages/cli/src/util/dev/server.ts
```diff
@@ -689,17 +690,39 @@ export default class DevServer {
     }
 
     // legacy NOW_REGION env variable
-    runEnv['NOW_REGION'] = 'dev1';
     buildEnv['NOW_REGION'] = 'dev1';
```
smaeda-ks, line 688: @styfle should I just remove this entirely?

packages/cli/src/util/dev/server.ts
```diff
@@ -689,17 +690,39 @@ export default class DevServer {
...
+          buildEnv = {
+            ...buildEnv,
+            ...getPrefixedEnvVars({ envPrefix: envPrefix, envs: buildEnv }),
+          };
+        }
+      }
     }
 
+    // re-compute allEnv with altered envs above
+    allEnv = { ...buildEnv, ...runEnv };
```
smaeda-ks: @styfle  from the old code comment above, it sounds like I should not include `runEnv` here. What do you think?
I'm referring to this comment:

>       // mirror how VERCEL_REGION is injected in prod/preview
>       // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
>       // and should not contain VERCEL_REGION

packages/cli/test/dev/fixtures/10a-nextjs-routes/pages/api/env.js
smaeda-ks: @styfle What's the best way to test build time env?

styfle: Use the env var on a page

packages/cli/src/util/dev/server.ts
styfle: No, I think we need to keep it for backwards compat

styfle: You can move it to the end after constructing `allEnv` though

packages/cli/src/util/dev/server.ts
styfle: You shouldn't need to recompute. Instead, move the code in the `if` statement up before the `allEnv` is created.

packages/cli/src/util/dev/server.ts
```diff
@@ -689,17 +690,39 @@ export default class DevServer {
...
+      // mirror how VERCEL_REGION is injected in prod/preview
+      // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
+      // and should not contain VERCEL_REGION
       runEnv['VERCEL_REGION'] = 'dev1';
+
+      // expose VERCEL_ENV and VERCEL_URL to dev command
+      buildEnv['VERCEL_ENV'] = 'development';
+      if (this.address.host) {
+        buildEnv['VERCEL_URL'] = this.address.host;
+      }
```
styfle, line 703: ~~This `VERCEL_URL` should be handled on line 685 already. No need to add it again.~~

Nevermind, this was only when there are no env vars.

I think we can actually delete the one on 685 instead and unconditionally set `VERCEL_URL`

packages/cli/src/util/dev/server.ts
```diff
@@ -689,17 +690,39 @@ export default class DevServer {
...
-    // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
-    // and should not contain VERCEL_REGION
     if (this.projectSettings?.autoExposeSystemEnvs) {
+      // mirror how VERCEL_REGION is injected in prod/preview
+      // only inject in `runEnvs`, because `allEnvs` is exposed to dev command
+      // and should not contain VERCEL_REGION
       runEnv['VERCEL_REGION'] = 'dev1';
+
+      // expose VERCEL_ENV and VERCEL_URL to dev command
+      buildEnv['VERCEL_ENV'] = 'development';
```
styfle, line 699: `VERCEL_ENV` should work at both build time and runtime

packages/cli/src/util/dev/server.ts
```diff
@@ -818,6 +841,8 @@ export default class DevServer {
         env['VERCEL_URL'] = this.address.host;
       } else if (name === 'VERCEL_REGION') {
         env['VERCEL_REGION'] = 'dev1';
+      } else if (name === 'VERCEL_ENV') {
+        env['VERCEL_ENV'] = 'development';
```
styfle: Is this `injectSystemValuesInDotenv()` function still necessary? Seems like it will get overridden by the code you added above.

smaeda-ks: deleted the use of `injectSystemValuesInDotenv()` entirely

packages/cli/src/util/dev/server.ts
```diff
@@ -675,31 +675,48 @@ export default class DevServer {
       this.getLocalEnv('.env.build', vercelConfig.build?.env),
     ]);
 
-    let allEnv = { ...buildEnv, ...runEnv };
-
     // If no .env/.build.env is present, use cloud environment variables
-    if (Object.keys(allEnv).length === 0) {
+    if (Object.keys({ ...runEnv, ...buildEnv }).length === 0) {
```
```suggestion
    const keyLength = Object.keys(runEnv).length + Object.keys(buildEnv).length
    if (keyLength === 0) {
```

Small change to avoid iterating more times than necessary

packages/cli/src/util/dev/server.ts
```diff
@@ -675,31 +675,48 @@ export default class DevServer {
...
+      if (this.projectSettings?.framework) {
+        const frameworkSlug = this.projectSettings.framework;
+        const framework = frameworkList.find(f => f.slug === frameworkSlug);
+        const envPrefix = framework?.envPrefix;
+
+        if (envPrefix) {
+          buildEnv = {
+            ...buildEnv,
+            ...getPrefixedEnvVars({ envPrefix: envPrefix, envs: buildEnv }),
+          };
```
```suggestion
           const prefixedEnvs = getPrefixedEnvVars({
             envPrefix,
             envs: buildEnv
           });

           for (const [key, value] of Object.entries(prefixedEnvs)) {
             buildEnv[key] = value;
           }
```

Lets avoid iterating over buildEnv when we can mutate instead.

This also follows the same pattern for usage of `getPrefixedEnvVars()` in other places.

packages/cli/test/dev/fixtures/10a-nextjs-routes/pages/api/env.js
styfle: Maybe it won't be a problem for `./pages/api/*` because Next.js wont know the difference.

I bet if you used `./api/*` instead then it might make a difference between build/runtime.

packages/cli/src/util/dev/server.ts
```diff
@@ -675,29 +675,52 @@ export default class DevServer {
...
+            envPrefix,
+            envs: buildEnv,
+          });
+          for (const [key, value] of Object.entries(prefixedEnvs)) {
+            buildEnv[key] = value;
+          }
+        }
+      }
+
+      allEnv = { ...allEnv, ...buildEnv };
```
smaeda-ks, line 723: @styfle I tried setting `allEnv` variable after all these env manipulations (i.e., `let allEnv = { ...buildEnv, ...runEnv };`), but it breaks some tests and potentially adds backward-compatibility issues. For example, `VERCEL_REGION` should not be included in `allEnv`. Otherwise, this line of test case fails:
https://github.com/vercel/vercel/blob/42c0b32a8d4ad1b1c88a318cbb247cd31c6e9521/packages/cli/test/integration-1.test.ts#L943

so I'm doing this reassignment by excluding `runEnv`.

jamiter:
Double checking: Will this also set `VERCEL_GIT_COMMIT_SHA` and related ENV variables correctly on `vercel pull`, or only `VERCEL_URL`? Or is this only related to `vc dev` and not to `vercel pull` at all?

I'm missing `VERCEL_GIT_COMMIT_SHA` on `vercel pull` and was hoping this PR would fix it 😉  Maybe have to look futher.

packages/cli/src/util/dev/server.ts
styfle, line 723:
```suggestion
      allEnv = { ...buildEnv, ...runEnv };
```

Should this be the same as line 658?

packages/cli/test/dev/fixtures/10a-nextjs-routes/pages/env.js
```diff
@@ -0,0 +1,14 @@
+function Env() {
+  return (
+    <main>
+      <p>
+        NEXT_PUBLIC_VERCEL_ENV: {process.env.NEXT_PUBLIC_VERCEL_ENV}
+      </p>
+      <p>
+        NEXT_PUBLIC_VERCEL_URL: {process.env.NEXT_PUBLIC_VERCEL_URL}
+      </p>
```
styfle, line 9: If you want to make this test less dependent on react internals `<!-- -->`, you can use `<em>` instead.

And even add `<ul>` and `<li>`

```suggestion
      <ul>
      <li>
        NEXT_PUBLIC_VERCEL_ENV: <em>{process.env.NEXT_PUBLIC_VERCEL_ENV}</em>
      </p>
      <li>
        NEXT_PUBLIC_VERCEL_URL: <em>{process.env.NEXT_PUBLIC_VERCEL_URL}</em>
      </li>
      </ul>
```

packages/cli/test/dev/integration-3.test.ts
```diff
@@ -136,6 +136,33 @@ test(
...
+      expect(apiJson.NEXT_PUBLIC_VERCEL_ENV).toBe('development');
+      expect(apiJson.VERCEL_URL).toBe(`localhost:${port}`);
+      expect(apiJson.NEXT_PUBLIC_VERCEL_URL).toBe(`localhost:${port}`);
+
+      const homeRes = await fetch(`http://localhost:${port}/env`);
+      const homeText = await homeRes.text();
+      expect(homeText).toContain('NEXT_PUBLIC_VERCEL_ENV: <!-- -->development');
+      expect(homeText).toContain(
+        `NEXT_PUBLIC_VERCEL_URL: <!-- -->localhost:${port}`
+      );
```
styfle, line 159: Should this test also check for the non-prefixed vars to ensure they are undefined?

styfle (commented):
Please reply to outstanding comments

packages/cli/src/util/dev/server.ts
smaeda-ks: I was unable to do so because of: https://github.com/vercel/vercel/pull/10064#discussion_r1229144628

EndangeredMassa:
Closing this old PR and tracking this as a backlog item for `vc dev` internally.

PR #10065, opened 2023-06-05
Enforce a changeset file to be present

TooTallNate:
Based off of these setup instructions:
https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#blocking

If a commit should really not publish a release, you can run `changeset --empty`.

In a follow-up PR we can add logic to enforce that the changeset updates specific package(s) based off of the diff.

### DIFF
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -33,3 +33,6 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run prettier-check
+
+      # Enforce a changeset file to be present
+      - run: changeset status --since=main
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -40,11 +40,12 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
 
       - run: git fetch origin main
+      - run: git branch
+      # Enforce a changeset file to be present
+      - run: pnpm exec changeset status --since=main
+
       - name: install pnpm@8.3.1
         run: npm i -g pnpm@8.3.1
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run prettier-check
-
-      # Enforce a changeset file to be present
-      - run: pnpm exec changeset status --since=main
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -29,7 +29,6 @@ jobs:
           fetch-depth: 0
           ref: main
       - run: git checkout ${{ github.event.pull_request.head.ref }}
-      - run: git branch
       - name: install pnpm@8.3.1
         run: npm i -g pnpm@8.3.1
       - run: pnpm install
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -19,31 +19,32 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
+  enforce-changeset:
+    name: Enforce Changeset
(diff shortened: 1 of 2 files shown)


.github/workflows/test-lint.yml
```diff
@@ -33,3 +33,6 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run prettier-check
+
+      # Enforce a changeset file to be present
+      - run: changeset status --since=main
```
EndangeredMassa: Note that this will only exit with non-zero if there's a code change in a package. Code changes outside of any package will still exit with `0`.

EndangeredMassa (dismissed):

styfle (dismissed):
Thanks!

.github/workflows/test-lint.yml
```diff
@@ -28,8 +28,12 @@ jobs:
...
           node-version: ${{ env.NODE_VERSION }}
+      - run: git fetch origin main
       - name: install pnpm@8.3.1
         run: npm i -g pnpm@8.3.1
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run prettier-check
+
+      # Enforce a changeset file to be present
+      - run: pnpm exec changeset status --since=main
```
styfle: This is failing with

```
Error: Failed to find where HEAD diverged from main. Does main exist?
```

.github/workflows/test-lint.yml
TooTallNate: <picture data-single-emoji=":ack:" title=":ack:">:ack:</picture> Probably ok I think?

.github/workflows/test-lint.yml
TooTallNate: Fixed.

.github/workflows/test-lint.yml
```diff
@@ -19,6 +19,23 @@ concurrency:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: main
+      - run: git checkout ${{ github.event.pull_request.head.ref }}
+      - run: git branch
+      - name: install pnpm@8.3.1
+        run: npm i -g pnpm@8.3.1
+      - run: pnpm install
```
styfle, line 34: Hopefully the two `pnpm install` will use the same filesystem and thus the second one will be cached.

If not, we might want to move this into the job below since you can use `if` conditions on `run`

styfle: Hmm, seems like its not cached. Are they run concurrently?

https://github.com/vercel/vercel/actions/runs/5181960729/jobs/9338166740?pr=10065#step:5:12

https://github.com/vercel/vercel/actions/runs/5181960729/jobs/9338166887?pr=10065#step:6:12

styfle (dismissed):

styfle:
@TooTallNate Is it expected that any changeset would satisfy this command, even if it wasn't added on the branch?

For example, `main` currently has https://github.com/vercel/vercel/blob/main/.changeset/ninety-pugs-melt.md so when I make a new branch and modify `./packages/cli` without adding a changeset, this command is successful.

.github/workflows/test-lint.yml
TooTallNate: Ya they're concurrent. And would be different filesystems. I don't think it matters much to optimize here since other workflows will take longer anyways.

TooTallNate:
@styfle Take a look at the previous two commits on this branch. I added a change in CLI to confirm the behavior that @EndangeredMassa was describing (the check failed, as expected), and then reverted again to make the check pass (since there's no packages being modified).

styfle (approved):

PR #10067, opened 2023-06-05
Add vercel.json setting overrides to storybook example for zero-config experience

Ethan-Arrowood:
This PR adds necessary configuration overrides to the storybook example so that it can have a zero-config experience.


examples/storybook/vercel.json
```diff
@@ -0,0 +1,4 @@
+{
+  "framework":"storybook",
```
leerob:
```suggestion
  "framework": "storybook",
```

leerob (dismissed):

leerob (approved):

PR #10069, opened 2023-06-06
[next] support maxDuration in Next.js deployments

ecklf:
Follow up PR to https://github.com/vercel/next.js/commit/8703c55f9f3d96df18ce192612141e094b3dc4fd which reads the newly created function config manifest and patches in the options for resource creation.


packages/build-utils/src/lambda.ts
```diff
@@ -267,5 +269,8 @@ export async function getLambdaOptionsFromFunction({
     }
   }
 
-  return {};
+  const project = new Project();
+  const staticConfig = getConfig(project, sourceFile);
+
+  return staticConfig ? staticConfig : {};
```
nuta: I guess this function is for extracting lambda parameters from framework builders, not parsing the script again. Perhaps we parse the script somewhere else?

packages/static-config/test/index.test.ts
```diff
@@ -16,6 +16,30 @@ describe('getConfig()', () => {
...
+    `);
+  });
+
+  it('should prioritize config export over named export from Node.js file', () => {
+    const project = new Project();
+    const sourcePath = join(__dirname, 'fixtures/node-duplicate-config.js');
+    const config = getConfig(project, sourcePath);
+    expect(config).toMatchInlineSnapshot(`
+      {
+        "maxDuration": 60,
```
nuta: Bikesheddy question: Is this a design decision that `const config` take precedence over `const maxDuration`? The latter looks more explicit and is intuitive to me. I don't have strong opinion here btw 😄

ecklf: > Bikesheddy question: Is this a design decision that `const config` take precedence over `const maxDuration`? The latter looks more explicit and is intuitive to me. I don't have strong opinion here btw 😄

My strong opinion is that we should error if people try to do it twice, but that would not be in sync with what we have with `export const config` and `vercel.json`. In those cases the `vercel.json` values have higher priority. I think we can revisit this at a later stage when reworking this entire behavior?

packages/static-config/src/index.ts
```diff
@@ -68,6 +88,29 @@ function getConfigNode(sourceFile: SourceFile) {
     });
 }
 
+function getNamedConfigNodes(sourceFile: SourceFile) {
+  const namedExports = ['maxDuration'];
```
ecklf: TODO memory

packages/static-config/test/index.test.ts
TooTallNate: I think it should error. `vercel.json` isn't exactly the same case, since it specifies a glob pattern, whereas the in-code config is specific to that specific function, so declaring configuration twice would seem like an error to me.

We should make a decision on this now since it would be a breaking change if we wanted to change it in the future.

ecklf: Thanks for your input Nate. I’ll chat about this with Javi tomorrow, but most likely changing this to error then.

nuta (changes requested):
Let's add tests 🙏

packages/next/src/utils.ts
```diff
@@ -2731,6 +2760,37 @@ export async function getMiddlewareBundle({
...
+  const functionConfigManifestPath = path.join(
+    entryPath,
+    outputDirectory,
+    './server/functions-config-manifest.json'
+  );
+
+  const hasManifest = await fs
+    .access(functionConfigManifestPath)
+    .then(() => true)
+    .catch(() => false);
```
Ethan-Arrowood, line 2774: I think you could use fs-extra's `fs.ensureFile` here: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureFile.md

packages/next/src/utils.ts
```diff
@@ -2731,6 +2760,37 @@ export async function getMiddlewareBundle({
...
+
+  if (!hasManifest) {
+    return;
+  }
+
+  const manifest = (await fs.readJSON(
+    functionConfigManifestPath
+  )) as FunctionsConfigManifest;
+
+  return manifest;
```
Ethan-Arrowood: Any concern for lack of JSON validation here?

Ethan-Arrowood (dismissed):
Test looks good, as long as its passing I'm content with this PR. 

See my other comments though for some ways this could be improved.

packages/next/src/utils.ts
```diff
@@ -1436,16 +1438,36 @@ export async function getPageLambdaGroups({
...
-      opts = await getLambdaOptionsFromFunction({
+
+      const configOpts = await getLambdaOptionsFromFunction({
         sourceFile,
         config,
       });
+
+      // If a user tries to customize in both handler and vercel.json we throw
+      const duplicateKeys = Object.keys(opts).filter(key =>
+        Object.keys(configOpts).includes(key)
```
styfle: You should move `Object.keys(configOpts)` out of the loop so it only runs once.

Also if there are a lot of keys, might be best to use `new Set(Object.keys(configOpts))`

packages/next/src/utils.ts
```diff
@@ -1436,16 +1438,36 @@ export async function getPageLambdaGroups({
...
+      );
+
+      if (duplicateKeys.length > 0) {
+        throw new Error(
+          `The route ${routeName} has duplicate configuration in both handler and vercel.json for: ${duplicateKeys.join(
```
styfle: Please add a test for this case

packages/next/src/utils.ts
TooTallNate: That would create the file if it doesn't exist. Really this should just be a try/catch around the `fs.readJSON()` call and return on `err.code === 'ENOENT'`.

ecklf: I'm using the same approach here as in the other `util` functions that read metadata files. We could do a follow-up PR refactoring this?

packages/next/src/utils.ts
ecklf: We could validate for the right types, though I wouldn't do strict validation as upstream changes in Next.js may break builds then.

packages/next/src/utils.ts
ecklf: Good catch, actually removing this as currently the default behavior is that `vercel.json` takes precedent.

packages/next/src/utils.ts
```diff
@@ -1410,6 +1411,7 @@ export async function getPageLambdaGroups({
 }: {
   entryPath: string;
   config: Config;
+  functionsConfigManifest?: FunctionsConfigManifest;
```
styfle: Lets make this required so its not missed on any calling function

```suggestion
  functionsConfigManifest: FunctionsConfigManifest;
```
We do this for `config` already because you can see some callers use `config = {}` if they don't have it.

packages/next/src/utils.ts
styfle: @ecklf The manifest file should have a `version` prop.

For example, see the routes manifest here: https://github.com/vercel/vercel/blob/d7fddfb07e089043144b886c77d45f8631d0c630/packages/next/src/utils.ts#L218

https://github.com/vercel/vercel/blob/d7fddfb07e089043144b886c77d45f8631d0c630/packages/next/src/utils.ts#L311

ecklf: @styfle I've asked @feedthejim about this too, but not all manifest files use a `version` prop. Is this absolutely required (we don't expect the signature of this to change).

ecklf: Next.js PR here https://github.com/vercel/next.js/pull/52507

Ethan-Arrowood (dismissed):

Ethan-Arrowood:
Tests make sense to me; approved 👍

styfle (changes requested):
The tests are failing

https://github.com/vercel/vercel/actions/runs/5521676682/jobs/10070225763?pr=10069#step:9:1477

https://vercel.com/vtest314-31nwfqbb1jo0kruyymbx/00-app-dir-edge/Hj6H6V4s6tLWWHKP6q92JLxvwRUr?filter=all

```
Error: ENOENT: no such file or directory, open '/vercel/path1/.next/server/app/index/index/page_client-reference-manifest.js'
```

ecklf:
> The tests are failing
> 
> https://github.com/vercel/vercel/actions/runs/5521676682/jobs/10070225763?pr=10069#step:9:1477
> 
> https://vercel.com/vtest314-31nwfqbb1jo0kruyymbx/00-app-dir-edge/Hj6H6V4s6tLWWHKP6q92JLxvwRUr?filter=all
> 
> ```
> Error: ENOENT: no such file or directory, open '/vercel/path1/.next/server/app/index/index/page_client-reference-manifest.js'
> ```

This is unrelated to my changes? 🤔

shuding (dismissed):
The failing test is caused by Next.js, but it should be fixed in the latest Next.js canary already. Which version is this test using?

ecklf:
> The failing test is caused by Next.js, but it should be fixed in the latest Next.js canary already. Which version is this test using?

It's pulling canary as far as I can see, I'll retrigger and see if it pulls the latest canary released an hour ago.

packages/next/src/utils.ts
```diff
@@ -2388,6 +2397,13 @@ export {
   getSourceFilePathFromPage,
 };
 
+export type FunctionsConfig = Record<
+  string,
+  {
+    maxDuration?: number;
+  }
+>;
```
```suggestion
export type FunctionsConfigManifestV1 = Record<
  string,
  {
    version: 1,
    maxDuration?: number;
  }
>;
```

packages/next/test/fixtures/00-app-dir-segment-options/next.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  experimental: {
+    appDir: true,
```
styfle: You can remove `appDir: true` since its no longer needed https://nextjs.org/blog/next-13-4

I'm not sure but I think `runtime` can be removed here too

packages/next/test/fixtures/00-app-dir-segment-options/vercel.json
styfle, line 1: I think you can remove vercel.json here since its the default behavior

packages/next/test/fixtures/00-app-dir-segment-options/next.config.js
ecklf: I'll check, thank you

packages/next/test/fixtures/00-app-dir-segment-options/vercel.json
ecklf: Seems the `run-build-lambda` util requires the file to be present:
```
    Received: undefined

      11 |
      12 |     if (typeof expect !== 'undefined') {
    > 13 |       expect(nowJsonRef).toBeDefined();
         |                          ^
      14 |     }
      15 |     const nowJson = json5.parse(await fs.readFile(nowJsonRef.fsPath, 'utf8'));
      16 |     const build = nowJson.builds[0];

      at ../../test/lib/run-build-lambda.js:13:26
      at Object.<anonymous> (test/integration/integration-1.test.js:126:29)
```

packages/next/test/fixtures/00-app-dir-segment-options/data-context.js
```diff
@@ -0,0 +1,3 @@
+import react from 'react'
+
+export const DataContext = react.createContext(null)
```
nuta: Maybe newbie question: who uses this context?

packages/next/test/integration/integration-1.test.js
```diff
@@ -122,6 +122,40 @@ if (parseInt(process.versions.node.split('.')[0], 10) >= 16) {
...
+    expect(buildResult.output['api/hello']).toBeDefined();
+    expect(buildResult.output['api/hello'].type).toBe('Lambda');
+    expect(buildResult.output['api/hello'].maxDuration).toBe(7);
+
+    expect(buildResult.output['api/hello-again']).toBeDefined();
+    expect(buildResult.output['api/hello-again'].type).toBe('Lambda');
+    expect(buildResult.output['api/hello-again'].maxDuration).toBe(7);
+    expect(
+      buildResult.output['api/hello-again'].supportsResponseStreaming
+    ).toBe(true);
```
nuta, line 156: suggestion: Is it intentional not to add this assertion for `api/hello`?

packages/next/src/utils.ts
```diff
@@ -2731,6 +2753,37 @@ export async function getMiddlewareBundle({
...
+
+  if (!hasManifest) {
+    return;
+  }
+
+  const manifest: FunctionsConfigManifestV1 = await fs.readJSON(
+    functionConfigManifestPath
+  );
+
+  return manifest.version === 1 ? manifest : undefined;
```
nuta, line 2784: suggestion: Should we throw an error instead of ignoring config silently?

nuta (dismissed):
Nice work! Left some nitpicks and suggestions.

packages/next/src/utils.ts
ecklf: @nuta issue being this would loudly break every Next.js build if there are upstream changes?

packages/next/test/fixtures/00-app-dir-segment-options/data-context.js
ecklf: Good catch this can go, copied the entire fixture and overlooked that

packages/next/test/integration/integration-1.test.js
ecklf: As far as I know only app dir supports streaming?

nuta: I thought `api/hello` is also in app dir. Looks good as is! 👍

packages/next/test/fixtures/00-app-dir-segment-options/next.config.js
ecklf: Removed

packages/next/src/utils.ts
ecklf: Fixed

styfle (approved):
Thanks!

PR #10071, opened 2023-06-06
[cli] Add support for `vc pull` command with repo link

TooTallNate:
When the repo is linked to Vercel with `vc link --repo`, the `.vercel` directory will be created at the _project root_ instead of the _repo root_, and will still contain the `project.json`.

The difference is that the `orgId` and `projectId` props will not be present in the `project.json` file when repo linked, since that information is available at the root level `repo.json` file.


EndangeredMassa (approved):
I tested this in a couple situations. It seems to work fine after the repo is linked.

I found an issue linking the repo in the first place, though: https://github.com/vercel/vercel/pull/10073

PR #10075, opened 2023-06-06
[cli] Add support for `vc build` command with repo link

TooTallNate:
When the repo is linked to Vercel with `vc link --repo`, the `vc build` command should be invoked from the project subdirectory (otherwise the project selector is displayed). The output directory is at `<projectRoot>/.vercel/output` instead of at the repo root.


cb1kenobi (approved):
LGTM

PR #10077, opened 2023-06-07
handle undefined content type in `vc dev`

EndangeredMassa:
When no content type header is sent to an API request during `vc dev`, the request fails:

```
/Users/smassa/source/vercel/vercel-2/node_modules/.pnpm/content-type@1.0.4/node_modules/content-type/index.js:108
    throw new TypeError('argument string is required')
          ^
TypeError: argument string is required
...
```

This comes from some runtime type validation happening in the `content-type` package, for which we do not have types, which is why typescript didn't catch this.

### DIFF
--- a/.changeset/sour-lies-pay.md
+++ b/.changeset/sour-lies-pay.md
@@ -1,0 +1,5 @@
+---
+"@vercel/node": patch
+---
+
+handle undefined content type in `vc dev`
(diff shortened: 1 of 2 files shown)


TooTallNate (approved):

styfle (commented):
Can you add a test?

EndangeredMassa:
Added tests: https://github.com/vercel/vercel/pull/10109

PR #10084, opened 2023-06-07
[cli] Move readme copy logic to a helper function for `vc link`

TooTallNate:
ncc has an issue with detecting + rewriting this file path for some reason. Moving to a helper function should work around that issue.

Fixing:

```
Error: ENOENT: no such file or directory, open '/node_modules/vercel/projects/VERCEL_DIR_README.txt'
```

### DIFF
--- a/packages/cli/src/util/projects/link.ts
+++ b/packages/cli/src/util/projects/link.ts
@@ -285,6 +285,13 @@ export async function getLinkedProject(
   return { status: 'linked', org, project, repoRoot: link.repoRoot };
 }
 
+export async function writeReadme(path: string) {
+  await writeFile(
+    join(path, VERCEL_DIR, VERCEL_DIR_README),
+    await readFile(join(__dirname, 'VERCEL_DIR_README.txt'), 'utf8')
+  );
+}
+
 export async function linkFolderToProject(
   client: Client,
   path: string,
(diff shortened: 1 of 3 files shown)


styfle (approved):

packages/cli/src/util/projects/link.ts
```diff
@@ -285,6 +285,13 @@ export async function getLinkedProject(
   return { status: 'linked', org, project, repoRoot: link.repoRoot };
 }
 
+export async function writeReadme(path: string) {
+  await writeFile(
+    join(path, VERCEL_DIR, VERCEL_DIR_README),
+    await readFile(join(__dirname, 'VERCEL_DIR_README.txt'), 'utf8')
+  );
```
cb1kenobi, line 292: **suggestion (non-blocking):** Node 14 introduce `import { copyFile } from 'fs/promises';`: https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fspromisescopyfilesrc-dest-mode

```suggestion
  await copyFile(
    join(__dirname, 'VERCEL_DIR_README.txt'),
    join(path, VERCEL_DIR, VERCEL_DIR_README)
  );
```

EndangeredMassa (approved):
I can no longer reproduce the issue with the tarball from this PR.

PR #10085, opened 2023-06-07
[cli] vc env pull should add `.env*.local` to `.gitignore`

styfle:
This is a follow up to PR #9892 which changed the default to `.env.local`.

Now that we know local files should never be committed to git, we can automatically add `.env*.local` to `.gitignore`. Note that this is the same ignore pattern that ships with create-next-app [as seen here](https://github.com/vercel/next.js/blob/06abd634899095b6cc28e6e8315b1e8b9c8df939/packages/create-next-app/templates/app/js/gitignore#L28), so most Next.js users won't see anything change.

See the related [Linear ticket](https://linear.app/vercel/issue/VCCLI-461/)

### DIFF
--- a/packages/cli/src/commands/env/pull.ts
+++ b/packages/cli/src/commands/env/pull.ts
@@ -19,6 +19,8 @@ import {
   createEnvObject,
 } from '../../util/env/diff-env-files';
 import { isErrnoException } from '@vercel/error-utils';
+import { addToGitIgnore } from '../../util/link/add-to-gitignore';
+import { getRepoLink } from '../../util/link/repo';
 
 const CONTENTS_PREFIX = '# Created by Vercel CLI\n';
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/env/pull.ts
```diff
@@ -136,11 +138,23 @@ export default async function pull(
...
   }
 
+  let isGitIgnoreUpdated = false;
+  if (filename === '.env.local') {
+    // When the file is `.env.local`, we also add it to `.gitignore`
+    // to avoid accidentally committing it to git.
+    // We use '.env*.local' to match the default .gitignore from
+    // create-next-app template. See:
+    // https://github.com/vercel/next.js/blob/06abd634899095b6cc28e6e8315b1e8b9c8df939/packages/create-next-app/templates/app/js/gitignore#L28
+    const repoLink = await getRepoLink(cwd);
```
TooTallNate: Let's pass the `link` in from `env/index.ts` as a parameter instead?

styfle: @TooTallNate The `link` in `env/index.ts` is not the same. Its a `ProjectLinkResult` type.

https://github.com/vercel/vercel/blob/ef30a46c037adfdf31cdd1e10642130ad41ae8ea/packages/cli/src/commands/env/index.ts#L147

However we need the `RepoLink` type.

TooTallNate: It's narrowed to `ProjectLinked` by the time `pull()` is called, which contains the `repoRoot` prop:

https://github.com/vercel/vercel/blob/94d5612dcef3a2a40558a8a571daf7471f8a0dac/internals/types/index.d.ts#L410

styfle: Oh nice I didn't know that `repoRoot` was there too 👍 

Updated in [fee8c02](https://github.com/vercel/vercel/pull/10085/commits/fee8c027a7be1094ec47a2f72d1bb19286598370)

TooTallNate (approved):

PR #10090, opened 2023-06-09
Improve `help` output UX and DX

Ethan-Arrowood:
This PR adds a new utility for generating `help` output for CLI commands. Backed by a JSON data structure that represents the commands, this utility uses deterministic functions for generating the different output sections (such as `options` and `examples`). This change will let contributor easily add/modify command details. The data structure is also used by the CLI to generate the args parsing structure for the `deploy` command. Optimistically, this same data structure can be used to generate documentation output for `front` such that the online docs will match the `help` output.


Ethan-Arrowood:
The algorithm now intelligently wraps the description for each option based on terminal width. Thanks @cb1kenobi for the idea! 

Screenshot 2023-06-09 at 23 12 26


Screenshot 2023-06-09 at 23 16 13

cb1kenobi:
Instead of moving the deploy command options into an internal package, I think we should use a framework, such as Commander.js, that keeps the command arguments and options in each command's source file. We don't have to use the framework's built-in help generation.

EndangeredMassa:
I agree with moving the `deploy` command options back to be next to the `deploy` command itself.

We don't have to switch to Commander right now, but the pattern is useful to follow.

Ethan-Arrowood:
The only reason I don't want to do that, is then anything else that wants to read that info will have to depend on the CLI package. I could instead do a new internal package `deploy-command` that for now is just the static command information. Hopefully we could eventually pull out the entirety of the `deploy` command too

EndangeredMassa:
Another process depending on the whole CLI would be chunky, but would it cause any problems? 

I suspect this could happen in a Github Action where we generate a docs PR based on updates. In that case, I don't think we need to optimize for speed of generating that PR.

packages/cli/package.json
```diff
@@ -86,6 +86,7 @@
     "@types/write-json-file": "2.2.1",
     "@types/yauzl-promise": "2.1.0",
     "@vercel-internals/constants": "1.0.2",
+    "@vercel-internals/get-help-output": "1.0.0",
```
EndangeredMassa: **suggestion (blocking):** ‏remove

```suggestion
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+import { parseEnv } from '../util/parse-env';
 import { errorToString, isErrnoException, isError } from '@vercel/error-utils';
-import { pickOverrides } from '../../util/projects/project-settings';
-import { printDeploymentStatus } from '../../util/deploy/print-deployment-status';
+import { pickOverrides } from '../util/projects/project-settings';
+import { printDeploymentStatus } from '../util/deploy/print-deployment-status';
+import { help } from './help';
+import type { Command } from './help';
+
+const deployCommand: Command = {
```
EndangeredMassa: **suggestion (non-blocking):** ‏I think it would still make sense to have the arg definition in the `args.ts` file. Either works, though.

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      deprecated: false,
+      description: 'Create a production deployment',
+      multi: false,
+    },
+    {
+      name: 'archive',
+      shorthand: null,
+      type: 'string',
+      deprecated: false,
+      description: 'TODO: Description for --archive',
```
EndangeredMassa: **suggestion (blocking):** ‏

```suggestion
      description: 'Compress the deployment code into a file before uploading it',
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      deprecated: false,
+      description: 'TODO: Description for --skip-domain',
+      multi: false,
+    },
+    {
+      name: 'yes',
+      shorthand: 'y',
+      type: 'boolean',
+      deprecated: false,
+      description: 'TODO: Description for --yes',
```
EndangeredMassa: **suggestion (blocking):** ‏

```suggestion
      description: 'Use default options to skip all prompts',
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      multi: false,
+    },
+    {
+      name: 'name',
+      shorthand: 'n',
+      type: 'string',
+      deprecated: true,
+      description: 'TODO: Description for --name',
```
EndangeredMassa: **suggestion (blocking):** ‏

```suggestion
      description: 'provide a Vercel Project name',
```

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+  if (command.options.length > 0) {
+    line.push('[options]');
+  }
+  return lineToString(line);
+}
+
+function buildCommandOptionLines(command: Command) {
+  // Filter out deprecated options.
+  command.options = command.options.filter(option => !option.deprecated);
```
EndangeredMassa: **suggestion (soft-blocking):** can we allow for intentionally undocumented as well?

```suggestion
  // Filter out deprecated and intentionally undocumented options
  command.options = command.options.filter(option => !option.deprecated && !option.description);
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      deprecated: false,
+      description: "Don't wait for the deployment to finish",
+      multi: false,
+    },
+    {
+      name: 'skip-domain',
+      shorthand: null,
+      type: 'boolean',
+      deprecated: false,
+      description: 'TODO: Description for --skip-domain',
```
EndangeredMassa: **suggestion (blocking):** ‏leave undocumented

If you accept my other suggestion to filter out undocumented (empty description) options, can we clear out the description here? This is intentionally undocumented until we can provide guidance on the full updated workflow.

```suggestion
      description: undefined,
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      deprecated: true,
+      description: 'TODO: Description for --name',
+      multi: false,
+    },
+    {
+      name: 'no-clipboard',
+      shorthand: null,
+      type: 'boolean',
+      deprecated: true,
+      description: 'TODO: Description for --no-clipboard',
```
EndangeredMassa: **suggestion (soft-blocking):** 

‏
```suggestion
      description: 'Do not copy deployment URL to clipboard',
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      multi: false,
+    },
+    {
+      name: 'target',
+      shorthand: null,
+      type: 'string',
+      deprecated: true,
+      description: 'TODO: Description for --target',
```
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
      description: 'Specify the target deployment environment',
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+      multi: false,
+    },
+    {
+      name: 'confirm',
+      shorthand: 'c',
+      type: 'boolean',
+      deprecated: true,
+      description: 'TODO: Description for --confirm',
```
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
      description: 'Use default options to skip all prompts',
```

packages/cli/src/commands/deploy.ts
```diff
@@ -40,73 +40,239 @@ import {
...
+  for (const option of deployCommand.options) {
+    argOptions[`--${option.name}`] =
+      option.type === 'boolean' ? Boolean : String;
+    if (option.shorthand) {
+      argOptions[`-${option.shorthand}`] = `--${option.name}`;
+    }
+    if (
+      option.name === 'env' ||
+      option.name === 'build-env' ||
+      option.name === 'meta'
```
EndangeredMassa: **question (non-blocking):** ‏Can we encode this in the Command definition instead of special casing it here?

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,212 @@
...
+const NEWLINE = '\n';
+const MAX_LINE_LENGTH = process.stdout.columns;
+
+export interface CommandOption {
+  name: string;
+  shorthand: string | null;
+  type: 'boolean' | 'string';
+  argument?: string;
+  deprecated: boolean;
+  description: string;
```
EndangeredMassa: **suggestion (non-blocking):** allow intentionally undocumented‏

```suggestion
  description?: string;
```

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,212 @@
...
+      startLine.push(`-${option.shorthand},`);
+    }
+    startLine.push(`--${option.name}`);
+    if (option.argument) {
+      startLine.push(`<${option.argument}>`);
+    }
+    // the length includes the INDENT
+    const lineLength = calcLineLength(startLine);
+    maxLineStartLength =
+      lineLength > maxLineStartLength ? lineLength : maxLineStartLength;
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
    maxLineStartLength = Math.max(lineLength, maxLineStartLength);
```

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,212 @@
...
+    '',
+    buildCommandExampleLines(command),
+  ];
+
+  return outputArrayToString(outputArray);
+}
+
+export function help(command: Command) {
+  return buildHelpOutput(command);
+}
```
EndangeredMassa, line 229: **suggestion (blocking):** ‏add tests

This logic looks like it's doing what was intended, but it would be really easy to be off by one somewhere. Can we get some tests around this logic?

EndangeredMassa (changes requested):

packages/cli/src/commands/deploy.ts
Ethan-Arrowood: I didn't see a clear way to do so in the definition. Our arg option handling could be improved whole-sale too at a later point. But I want the new stuff to work as is with the existing args handling.

packages/cli/src/commands/deploy.ts
Ethan-Arrowood: done

packages/cli/src/commands/help.ts
Ethan-Arrowood: added!

packages/cli/src/util/get-project-name.ts
```diff
@@ -2,7 +2,7 @@ import { basename } from 'path';
 import { VercelConfig } from '@vercel/client';
 
 export interface GetProjectNameOptions {
-  argv: { '--name'?: string };
+  argv: { [k in string]?: string };
```
TooTallNate: this doesn't seem right. does it need to be changed? did you mean this?

```suggestion
  argv: { [k: string]: string };
```

Ethan-Arrowood: yeah i think so. typescript didn't seem to mind though 😆

Ethan-Arrowood: Screenshot 2023-06-28 at 15 00 51
This is what im getting locally.

TooTallNate: did we need to change this line though? what's the purpose?

Ethan-Arrowood: Now that the input to `args()` is generated from the data structure, it messed with some of the types. I don't really understand it, but this was the only spot typescript was failing. This change fixes it

Ethan-Arrowood: The `argOptions` on line 102 i think trickles down to this type change. Idk why. Theres some TypeScript magic at work in the `args` package

TooTallNate: `{ [k: string]: string | undefined }`

Ethan-Arrowood: fixed

packages/cli/src/commands/deploy/index.ts
```diff
@@ -69,44 +68,38 @@ import { parseEnv } from '../../util/parse-env';
...
+import { help } from '../help';
+import { deployCommand } from './command';
 
 export default async (client: Client): Promise<number> => {
   const { output } = client;
 
   let argv = null;
 
+  const argOptions: {
+    [k in string]:
```
EndangeredMassa: **suggestion (non-blocking):** use alternate syntax‏

```suggestion
    [k: string]:
```

Can we use the simpler syntax here too? I tried it locally and it had no TS errors.

packages/cli/test/unit/commands/help.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  describe('outputArrayToString', () => {
+    test('should join a list of strings using newlines', () => {
+      expect(outputArrayToString(['line 1', 'line 2', 'line 3'])).toBe(
+        'line 1\nline 2\nline 3'
+      );
+    });
+  });
+
+  test('help', () => {
```
EndangeredMassa: **suggestion (non-blocking):** ‏add a small column test

Can we add another test with a smaller column size to exercise word wrapping?

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/help.test.ts
Ethan-Arrowood: I can't figure out how to get jest to respect changes to `process.stdout.columns` :(

Ethan-Arrowood: aye i figured it out. fixes pushing now

packages/cli/test/unit/commands/help.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+  });
+
+  describe('help output snapshots', () => {
+    const origCol = process.stdout.columns;
+    afterAll(() => {
+      process.stdout.columns = origCol;
+    });
+    test.each([40, 80, 120])('column width %i', width => {
+      process.stdout.columns = width;
```
TooTallNate: These should be using `client.stderr` which is meant for mocking (and because help is printed to stderr).

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+  }
+  if (command.options.length > 0) {
+    line.push('[options]');
+  }
+  return lineToString(line);
+}
+
+export function buildCommandOptionLines(command: Command) {
+  const MAX_LINE_LENGTH = process.stdout.columns;
```
TooTallNate: Should be `client.stderr`

packages/cli/test/unit/commands/help.test.ts
TooTallNate: Also needs to test the `undefined` case (when not a TTY).

packages/cli/src/commands/help.ts
Ethan-Arrowood: Good callout. Noticed that the client types don't use `tty.WriteStream`, so I change that. Not sure if you did that intentionally or if you just didn't know that type existed. This is the type that `process.stderr` uses.

TooTallNate: Ya we should probably use that type then.

TooTallNate (approved):

PR #10093, opened 2023-06-09
[cli] Remove `findProjectFromPath()`

TooTallNate:
Superceded by the `findProjectsFromPath()` function which may return multiple matches, and it was only being used in tests.

The relevant tests have been updated to use the multiple matches version instead, and updated to include the case where multiple matches are returned.

__Note:__ No changeset for this one since it's an internal function being removed, and doesn't need to be referenced in the public changelog.


.changeset/yellow-eels-stare.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
styfle, line 2: The contents are missing here

TooTallNate: Yep. See note in PR description (the file is the output of the `changeset --empty` command).

packages/cli/src/util/link/repo.ts
```diff
@@ -288,14 +288,9 @@ export function findProjectsFromPath(
...
- */
-export function findProjectFromPath(
-  projects: RepoProjectConfig[],
-  path: string
-): RepoProjectConfig | undefined {
-  return findProjectsFromPath(projects, path)[0];
+  // If there are multiple matches, we only want the most relevant
+  // selections (with the deepest directory depth), so pick the first
+  // one and filter on those matches.
+  const firstMatch = matches[0];
```
styfle, line 294: Does this guarantee that `matches.length > 0`?

TooTallNate: No, but it's ok because then the `filter()` below would not be executed in the case of an empty array.

.changeset/yellow-eels-stare.md
styfle: I wonder if Changesets can follow Conventional Commits which has the convention of using `chore:` or `refactor:` prefix for this case.

https://www.conventionalcommits.org/en/v1.0.0/#:~:text=types%20other%20than%20fix%3A%20and%20feat%3A%20are%20allowed

styfle (approved):

PR #10100, opened 2023-06-12
[fs-detectors] Use `LocalFileSystemDetector` instead of `FixtureFilesystem`

TooTallNate:
The code for these two are almost identical, so consolidate into one codebase.

Also adjusts the `pnpm test` script to allow for specifying a file name to be executed, instead of running all tests.


.changeset/lemon-shoes-hug.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
styfle: This doesn't seem right 🤔 

How can you ship a new version of `@vercel/fs-detectors` without a patch?

styfle (changes requested):
Please add proper changeset file

.changeset/lemon-shoes-hug.md
TooTallNate: There's only (non-breaking) changes to the tests, which doesn't warrant a new version bump.

styfle (dismissed):

styfle (dismissed):

packages/fs-detectors/src/detectors/local-file-system-detector.ts
```diff
@@ -44,14 +49,22 @@ export class LocalFileSystemDetector extends DetectorFilesystem {
...
     }));
   }
+
   _chdir(name: string): DetectorFilesystem {
     return new LocalFileSystemDetector(this.getFilePath(name));
   }
+
+  private getRelativeFilePath(name: string) {
+    return name.startsWith(this.rootPath)
+      ? name.substring(this.rootPath.length + 1)
```
styfle: What is the `+ 1` offset here? 

Seems like it should use `path.relative()` instead?

TooTallNate: `+1` is to strip off the leading `/`. `path.relative()` should be essentially the same thing, but ya sure, switched over in c2cf1e8d017d92034bd86f9e87307fce0c1e6782.

.changeset/lemon-shoes-hug.md
TooTallNate: Nvm. Seems that `LocalFileSystemDetector` had a bug, so this will be a semver major change.

.changeset/calm-apes-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/fs-detectors': major
+---
+
+`LocalFileSystemDetector#readdir()` now returns paths relative to the root dir, instead of absolute paths. This is to align with the usage of the detectors that are using the `DetectorFilesystem` interface.
```
styfle, line 5: Does "root dir" here mean Root Directory setting or repository root (or something else)?

TooTallNate: Root of the "virtual filesystem"

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -5,7 +5,7 @@ import { LocalFileSystemDetector, DetectorFilesystem } from '../src';
 
 const tmpdir = path.join(os.tmpdir(), 'local-file-system-test');
 
-const dirs = ['', 'a', 'a/b']; // root, single-nested, double-nested
+const dirs = ['', 'a', `a${path.sep}b`]; // root, single-nested, double-nested
```
styfle, line 8: I wonder if this was intentionally `/` on all systems?

Should we use `path.posix.join()` instead of `path.join()`?

Ethan-Arrowood (approved):

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
TooTallNate: tbh I have no idea. `DetectorFilesystem` is not very well defined IMO. Maybe @Ethan-Arrowood @agadzik @chloetedder can clarify.

TooTallNate: In a perfect world I would vote to redesign the API for `DetectorFilesystem#readdir()` such that it would _only_ return the filenames, and not have the directory context with it. That would match POSIX behavior. If we wanted to do that then we would need to coordinate with the API side as well since this interface is used there.

Ethan-Arrowood: I don't remember. Didn't know about that expectation of readdir either so it could just be a mistake. I'm +1 for your proposed change if the API side isn't going to be a ton of work

agadzik: I'm pretty sure that we don't use any directory info from readdir for our detection utils, just file info

PR #10110, opened 2023-06-15
[cli] do not force auto-assign value on deployments

EndangeredMassa:
When `vc deploy` happens, it was forcing a `true`/`false` value for `autoAssignCustomDomains`. This overwrites the project setting in some cases, which is not what we want.

If a CLI command or API call explicitly wants to force `autoAssignCustomDomains` on or off, it can. Otherwise, the project setting value should be used.

This PR only sets `autoAssignCustomDomains` to `false` when the `--skip-domain` flag is passed. Otherwise, the value is `undefined` so that the project setting can take effect.

### DIFF
--- a/.changeset/fluffy-socks-refuse.md
+++ b/.changeset/fluffy-socks-refuse.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] do not force auto-assign value on deployments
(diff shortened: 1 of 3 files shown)


cb1kenobi (approved):

EndangeredMassa:
Manually tested this tarball and it does fix the problem.

TooTallNate (approved):

PR #10112, opened 2023-06-16
[cli] Add debug logging for repo link root lookup

TooTallNate:
Adds some logging during the root directory resolving logic to help debug any cases where that lookup isn't behaving correctly.


cb1kenobi (approved):
LGTM

PR #10114, opened 2023-06-16
[cli] Only show relevant Project matches in Project selector

TooTallNate:
When showing the Project selector and there's more than one Project matched given the current cwd, only show the matched Projects instead of all Projects listed in the `repo.json` file.


adityadafe (approved):

adityadafe:
sorry didnt meant to approve

packages/cli/src/util/projects/link.ts
```diff
@@ -95,11 +95,13 @@ async function getProjectLinkFromRepoLink(
   if (projects.length === 1) {
     project = projects[0];
   } else {
+    const selectableProjects =
+      projects.length > 0 ? projects : repoLink.repoConfig.projects;
     const { p } = await client.prompt({
       name: 'p',
       type: 'list',
       message: `Please select a Project:`,
-      choices: repoLink.repoConfig.projects.map(p => ({
```
EndangeredMassa, line 102: **question (non-blocking):** When is there more than one matched project that hits this condition? I tried this in a monorepo and couldn't tell what was supposed to be different.

TooTallNate: It would happen if there were multiple Projects with the same `rootDirectory` setting applied. This could happen for example with Storybook, where there's a Next.js app and the Storybook project with the same root directory. In this case, it's ambiguous which project is intended to be interacted with, so we show the picker.

EndangeredMassa (approved):
Tested this by setting up a monorepo where one of the packages was Next.js + Storybook and both were created as Vercel Projects, linking the repo `vc link --repo`, going into the Next.js + Storybook directory, and running `vc pull`.

In latest CLI, I see a picker with all of the linked projects. In this PR's version of CLI, I see only the storybook and next.js projects.

PR #10126, opened 2023-06-22
[fs-detectors] Resolve symlinks in `LocalFileSystemDetector#readdir()`

TooTallNate:
`LocalFileSystemDetector#readdir()` was throwing an error when a symlink was encountered, due to `fs.readdir()` `withFileTypes: true` option performing an lstat instead of a stat operation.

So re-implement the `readdir()` logic to use `fs.stat()` so that the symlink is resolved (only "dir" and "file" types are expected in the result).

### DIFF
--- a/packages/fs-detectors/test/unit.get-workspaces.test.ts
+++ b/packages/fs-detectors/test/unit.get-workspaces.test.ts
@@ -35,7 +35,6 @@ describe.each<[string, Workspace[]]>([
 
   it(testName, async () => {
     const fixture = path.join(__dirname, 'fixtures', fixturePath);
-    console.log(fixture);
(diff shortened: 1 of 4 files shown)


packages/fs-detectors/test/unit.get-workspaces.test.ts
```diff
@@ -34,6 +35,7 @@ describe.each<[string, Workspace[]]>([
 
   it(testName, async () => {
     const fixture = path.join(__dirname, 'fixtures', fixturePath);
+    console.log(fixture);
```
EndangeredMassa:
```suggestion
```

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #10135, opened 2023-06-23
require `--yes` flag to promote preview deployment

EndangeredMassa:
In order to promote a preview deployment, we want to make sure the user knows that this is not typical. Now the `--yes` flag is required to make this work.

### DIFF
--- a/.changeset/green-badgers-reflect.md
+++ b/.changeset/green-badgers-reflect.md
@@ -2,4 +2,4 @@
 "vercel": patch
 ---
 
-require force to promote preview deployment
+require `--yes` to promote preview deployment
--- a/.changeset/green-badgers-reflect.md
+++ b/.changeset/green-badgers-reflect.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+require force to promote preview deployment
--- a/packages/cli/src/commands/promote/request-promote.ts
+++ b/packages/cli/src/commands/promote/request-promote.ts
@@ -4,6 +4,7 @@ import { getCommandName } from '../../util/pkg-name';
 import getProjectByDeployment from '../../util/projects/get-project-by-deployment';
 import ms from 'ms';
 import promoteStatus from './status';
+import confirm from '../../util/input/confirm';
 
 /**
  * Requests a promotion and waits for it complete.
@@ -16,12 +17,12 @@ export default async function requestPromote({
   client,
   deployId,
   timeout,
-  force,
+  yes,
 }: {
   client: Client;
   deployId: string;
   timeout?: string;
-  force: boolean;
+  yes: boolean;
 }): Promise<number> {
   const { output } = client;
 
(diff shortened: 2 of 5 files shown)


cb1kenobi (dismissed):
LGTM!

cb1kenobi (dismissed):

styfle (changes requested):
Let's use `--yes` for consistency with other commands

cb1kenobi:
> Let's use `--yes` for consistency with other commands

Personally, as much as I love consistency, I think `--force` makes more sense than `--yes` in this case.

packages/cli/src/commands/promote/request-promote.ts
```diff
@@ -29,6 +31,13 @@ export default async function requestPromote({
     output: client.output,
   });
 
+  if (deployment.target !== 'production' && !force) {
+    output.error(
+      'Cannot promote deployment that does not target "production". If you are sure you want to do this, add `--force`.'
+    );
```
styfle: It seems contradictory to say “cannot” and then say that you can with a flag. 

I think we can improve this message and include reasons why you wouldn’t want to do this. 

Something like:

> Warning: This deployment does not target production, therefore promotion will not apply production environment variables. Are you sure you want to continue? Y/n

EndangeredMassa: I like this better. In this framing, it does make more sense to use `--yes`.

EndangeredMassa: Switched! https://github.com/vercel/vercel/pull/10135/commits/de9914719cc5fca88e6977447550862eafa16c74

packages/cli/src/commands/promote/request-promote.ts
```diff
@@ -29,6 +32,18 @@ export default async function requestPromote({
...
 
+  if (deployment.target !== 'production' && !yes) {
+    const question =
+      'This deployment does not target production, therefore promotion will not apply production environment variables. Are you sure you want to continue?';
+    const answer = await confirm(client, question, true);
+    if (!answer) {
+      output.error(
+        'User declined to promote a deployment that did not target production.'
+      );
+      return 1;
```
styfle:
```suggestion
    const answer = await confirm(client, question, false);
    if (!answer) {
      output.error('Canceled');
      return 0;
```

Looks like the pattern here is to default to `false` and then false will print Canceled with 0 exit code.

See other commands here:

https://github.com/vercel/vercel/blob/36db62a491aa3155ebb286f0e533bd8edfc93670/packages/cli/src/commands/env/rm.ts#L104-L116

https://github.com/vercel/vercel/blob/36db62a491aa3155ebb286f0e533bd8edfc93670/packages/cli/src/commands/env/pull.ts#L88-L97

.changeset/green-badgers-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
+---
+
+require force to promote preview deployment
```
styfle:
```suggestion
require `--yes` to promote preview deployment
```

styfle (approved):
Thanks!

cb1kenobi (approved):

PR #10144, opened 2023-06-26
Delete `.changeset/beige-carrots-begin.md`

TooTallNate:
We don't want changesets to try and create a release / changelog for the `examples` dir.


cb1kenobi (approved):

styfle (approved):
Could we automate this so it doesn't accidentally happen again?

PR #10150, opened 2023-06-27
[build-utils] Add `getNodeBinPaths()` function

TooTallNate:
The `getNodeBinPath()` function is problematic because it assumes that commands are installed in the `node_modules` directory alongside the detected lockfile. This works fine the majority of the time, but ends up not being the case when using a monorepo that uses a package manager in "linked" mode (i.e. pnpm by default).

Consider the following:

```
.
├── pnpm-lock.yaml
├── node_modules
├── blog
│   ├── node_modules
│   │   ├── hexo -> .pnpm/hexo@3.9.0/node_modules/hexo
```

In this setup, adding the root-level `node_modules/.bin` would not make the `hexo` command be visible in the `$PATH`.

To solve this issue, the new `getNodeBinPaths()` function returns an array of all directories up to the specified `root`, which can then be placed into the `$PATH`. It's also more efficient (synchronous) since it does not need to scan for a lockfile anymore (the `root` needs to be specified explicitly).

The new function is being used in `@vercel/next` and `@vercel/static-build`.

The `traverseUpDirectories()` function from CLI was moved to `build-utils` to implement this function. Consequently, that makes the implementations of `walkParentDirs()` and `walkParentDirsMulti()` simpler, since it's using this generator now.


.changeset/tough-singers-live.md
styfle, line 1: Is it expected that there are two files in the `.changeset` dir?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -131,6 +131,21 @@ export async function execCommand(command: string, options: SpawnOptions = {}) {
   return true;
 }
 
+export function* traverseUpDirectories(start: string, root?: string) {
```
styfle: Let's change this from two args to a single obj arg

```suggestion
export function* traverseUpDirectories({ start, root }: {start: string, root?: string}) {
```

This is much safer long term if we make changes to arguments

packages/build-utils/test/unit.traverse-up-directories.test.ts
```diff
@@ -0,0 +1,48 @@
+import { traverseUpDirectories } from '../src/fs/run-user-scripts';
```
styfle, line 1: Do we want to change this to `walk` instead of `traverse`?

We already have a similar `walkParentDirs()` function

styfle: Actually, could this be an option to the existing `walkParentDirs()` instead of a new function?

Perhaps `{ all: true }`?

.changeset/tough-singers-live.md
TooTallNate: Ya. I wanted to use different text in the respective changelogs

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -327,9 +346,8 @@ async function walkParentDirsMulti({
   start: string;
   filenames: string[];
 }): Promise<(string | undefined)[]> {
-  let parent = '';
-  for (let current = start; base.length <= current.length; current = parent) {
-    const fullPaths = filenames.map(f => path.join(current, f));
+  for (const dir of traverseUpDirectories(start, base)) {
+    const fullPaths = filenames.map(f => path.join(dir, f));
```
styfle, line 356: Very nice! Glad that could be reused 👍

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -141,6 +156,18 @@ export async function getNodeBinPath({
   return path.join(dir, 'node_modules', '.bin');
 }
 
+export function getNodeBinPaths({
+  start,
+  root,
```
styfle: Can we change `root` to `base` in these new functions to match the existing ones?

It also matches the option from `@vercel/nft` https://github.com/vercel/nft#options

packages/build-utils/test/unit.traverse-up-directories.test.ts
TooTallNate: Maybe, not sure. It might be confusing with `walkParentDirs()` (I think that function has a bad name for what it does FWIW). `traverse` might be a good distinction in this case since it's a generator.

packages/build-utils/src/fs/run-user-scripts.ts
TooTallNate: 29c0e0adb542b2ae793dee598a1aec9e6a1627db

packages/build-utils/src/fs/run-user-scripts.ts
TooTallNate: 29c0e0adb542b2ae793dee598a1aec9e6a1627db

styfle (dismissed):
🥇

styfle (dismissed):

styfle (approved):

PR #10154, opened 2023-06-28
[cli] Optimize write build result for vc build

ijjk:
As noticed this is copying all fallback files from output even if they are already on the disk in the build container which should be able to be considered a "pure" environment. This copying can cause disk space/perf issues when there are massive amounts of fallback files generated e.g. 8000 outputs at 200KB/output is already 1.6GB of copying needing to be done that could just be a move operation. 

x-ref: [slack thread](https://vercel.slack.com/archives/C03F2CMNGKG/p1687969734598509?thread_ts=1687963253.178369&cid=C03F2CMNGKG)
x-ref: [slack thread](https://vercel.slack.com/archives/C02SM2K2UM9/p1687970267370669)
x-ref: [slack thread](https://vercel.slack.com/archives/C03DQ3QFV7C/p1687338157228459)

### DIFF
--- a/packages/cli/src/util/build/write-build-result.ts
+++ b/packages/cli/src/util/build/write-build-result.ts
@@ -131,6 +131,9 @@ async function writeBuildResultV2(
 
   const lambdas = new Map<Lambda, string>();
   const overrides: Record<string, PathOverride> = {};
+
+  const isBuildContainer = !!process.env.VERCEL;
+
   for (const [path, output] of Object.entries(buildResult.output)) {
     const normalizedPath = stripDuplicateSlashes(path);
     if (isLambda(output)) {
@@ -159,8 +162,8 @@ async function writeBuildResultV2(
 
         // if we're in the build container and fallback is already
         // on the disk just move it to avoid duplicating output
-        if (process.env.VERCEL && 'fsPath' in fallback) {
-          await fs.move(fallback.fsPath, fallbackPath);
+        if (isBuildContainer && 'fsPath' in fallback) {
+          await fs.link(fallback.fsPath, fallbackPath);
         } else {
           const stream = fallback.toStream();
           await pipe(
--- a/packages/cli/src/util/build/write-build-result.ts
+++ b/packages/cli/src/util/build/write-build-result.ts
@@ -156,11 +156,18 @@ async function writeBuildResultV2(
         const ext = getFileExtension(fallback);
         const fallbackName = `${normalizedPath}.prerender-fallback${ext}`;
         const fallbackPath = join(outputDir, 'functions', fallbackName);
-        const stream = fallback.toStream();
-        await pipe(
-          stream,
-          fs.createWriteStream(fallbackPath, { mode: fallback.mode })
-        );
+
+        // if we're in the build container and fallback is already
+        // on the disk just move it to avoid duplicating output
+        if (process.env.VERCEL && 'fsPath' in fallback) {
+          await fs.move(fallback.fsPath, fallbackPath);
+        } else {
+          const stream = fallback.toStream();
+          await pipe(
+            stream,
+            fs.createWriteStream(fallbackPath, { mode: fallback.mode })
+          );
+        }
         fallback = new FileFsRef({
           ...output.fallback,
           fsPath: basename(fallbackName),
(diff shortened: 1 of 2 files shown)


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -156,11 +156,18 @@ async function writeBuildResultV2(
...
         const fallbackPath = join(outputDir, 'functions', fallbackName);
-        const stream = fallback.toStream();
-        await pipe(
-          stream,
-          fs.createWriteStream(fallbackPath, { mode: fallback.mode })
-        );
+
+        // if we're in the build container and fallback is already
+        // on the disk just move it to avoid duplicating output
+        if (process.env.VERCEL && 'fsPath' in fallback) {
```
styfle: Are you sure `process.env.VERCEL` will work as you expected?

Seems like `vc pull && vc build` will assign that env var locally when the user has System Env Vars checkbox checked.

packages/cli/src/util/build/write-build-result.ts
```diff
@@ -156,11 +156,18 @@ async function writeBuildResultV2(
...
+          await fs.move(fallback.fsPath, fallbackPath);
```
styfle: Should this be `fs.rename()`?

https://nodejs.org/api/fs.html#fsrenameoldpath-newpath-callback

cb1kenobi: FWIW, `fs.move()` will create the destination parent directories if needed.

packages/cli/src/util/build/write-build-result.ts
ijjk: Is there a better env variable you are aware of we could rely on instead?

packages/cli/src/util/build/write-build-result.ts
ijjk: Updated to use `fs.link` instead as it's more reliable and has same perf/disk space wins

packages/cli/src/util/build/write-build-result.ts
TooTallNate: Checking for `VERCEL` is indeed not going to be reliable. It's set when running `vc build` locally as well. IMO we should introduce an env var in build-container. We could also check the execution path, i.e. build-container has a root-level `/vercel` directory.

TooTallNate: In general though, I've tried to avoid logic like this in `vc build`. Instead, we should add a CLI flag (or env var) which alters the behavior of `vc build`. For example, build-container sets the `--output` flag to specify where the BOA files should go, rather then having that logic baked directly in to `vc build`.

packages/cli/src/util/build/write-build-result.ts
```diff
@@ -156,11 +159,18 @@ async function writeBuildResultV2(
...
         const fallbackName = `${normalizedPath}.prerender-fallback${ext}`;
         const fallbackPath = join(outputDir, 'functions', fallbackName);
-        const stream = fallback.toStream();
-        await pipe(
-          stream,
-          fs.createWriteStream(fallbackPath, { mode: fallback.mode })
-        );
+
+        // if we're in the build container and fallback is already
+        // on the disk just move it to avoid duplicating output
```
styfle: The comment is no longer true since its not moving, right?

Also, can we do this every time (not just build container)?

ijjk: I think hardlink should be safe always actually, only was branching since move could have potentially caused issues if the build files were used separate of `vc build` locally

TooTallNate: Perhaps default to link in a try / catch and fall back to copy.

ijjk: Updated to fallback to copy on link failure

packages/cli/src/util/build/write-build-result.ts
```diff
@@ -156,11 +157,25 @@ async function writeBuildResultV2(
...
-          stream,
-          fs.createWriteStream(fallbackPath, { mode: fallback.mode })
-        );
+
+        // if file is already on the disk we can hard link
+        // instead of creating a new copy
+        let usedHardLink = false;
+        if ('fsPath' in fallback) {
+          try {
+            await fs.link(fallback.fsPath, fallbackPath);
```
TooTallNate, line 166:
```suggestion
            await fs.link(fallback.fsPath, fallbackPath);
            usedHardLink = true;
```

styfle (approved):
Great, thanks!

TooTallNate (approved):

PR #10157, opened 2023-06-28
[cli] Remove `preinstall` script

styfle:
This will solve two of the warnings from [Socket.dev](https://socket.dev/npm/package/vercel):

image

This preinstall script isn't necessary to run Vercel CLI, its only used to improve the error message in two cases:

- The system has `node` version prior to 14.x (3 years ago)
- The system still has `now` installed but should be using `vercel` (last shipped 3 years ago)

The usage of `preinstall` scripts have gone out of style due to security audits so we should remove this script now.


TooTallNate (approved):
Absolutely <picture data-single-emoji=":+1-vote:" title=":+1-vote:">:+1-vote:</picture>

EndangeredMassa (approved):

cb1kenobi (approved):

PR #10169, opened 2023-07-03
[node] add named export support for `maxDuration`

ecklf:
This PR enables named export parsing to configure route behavior for `maxDuration`.
Uniform API to Next.js https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config


packages/node/test/integration/integration-build-output.test.js
```diff
@@ -0,0 +1,17 @@
...
+const runBuildLambda = createRunBuildLambda(builder);
+
+it('should have configured `maxDuration` in build output', async () => {
+  const { buildResult } = await runBuildLambda(
+    path.join(__dirname, '../fixtures/63-routes-config')
+  );
+
+  expect(buildResult.output).toBeDefined();
+  expect(typeof buildResult.output.maxDuration).toBe('number');
+  expect(buildResult.output.maxDuration).toMatchInlineSnapshot(`5`);
```
TooTallNate:
```suggestion
  expect(buildResult.output.maxDuration).toEqual(5);
```

packages/node/src/index.ts
```diff
@@ -416,7 +416,29 @@ export const build: BuildV3 = async ({
...
+  const segmentConfig = getSegmentConfig(project, entrypointPath);
+
+  if (objectConfig && segmentConfig) {
+    // If a user tries to customize in both handler and vercel.json we throw
+    const segmentConfigKeys = new Set(Object.keys(segmentConfig));
+    const duplicateKeys = Object.keys(objectConfig).filter(key =>
+      segmentConfigKeys.has(key)
+    );
+
+    if (duplicateKeys.length > 0) {
```
styfle, line 429: I'm not sure this is the right solution.

Doesn't `vercel.json` let you use a glob pattern to match many files?

packages/node/test/fixtures/63-routes-config/vercel.json
```diff
@@ -0,0 +1,3 @@
+{
+  "builds": [{ "src": "api/handler.js", "use": "@vercel/node" }]
+}
```
styfle, line 3: You can remove this `vercel.json` file since it isn't needed

packages/node/src/index.ts
ecklf: Correct me if I'm wrong but I don't think `maxDuration` in `vercel.json` is respected in our Node runtime at all currently. The `vercel.json` config should be passed to the `build` function as `build` prop, but in the current implementation (https://github.com/vercel/vercel/blob/main/packages/node/src/index.ts) I don't see where this value would be consumed. In regards to `getConfig` and `getSegmentConfig` — they do static code analysis on the handler file directly. This line of code basically prevents doing something like this:

```
export const config = { memory: 128, maxDuration: 1 };
export const maxDuration = 5;
```

styfle: It has been supported in `vercel.json` for 4 years

https://vercel.com/blog/customizing-serverless-functions

PR #10172, opened 2023-07-03
[cli] Allow additional project settings in `createProject()`

TooTallNate:
Previously, `createProject()` only allowed a string `name` to be provided, and another API call would need to be done after creation with `updateProject()`. This should not be necessary since the `POST` endpoint can accept additional properties already.


packages/cli/src/util/projects/update-project.ts
```diff
@@ -1,26 +1,16 @@
 import Client from '../client';
-import type { JSONObject, ProjectSettings } from '@vercel-internals/types';
-
-interface ProjectSettingsResponse extends ProjectSettings {
-  id: string;
-  name: string;
-  updatedAt: number;
-  createdAt: number;
-}
```
styfle, line 9: Seems like removing this will result in a return type of `any`.

Is that intentional?

TooTallNate: Not the case. Return type is `Promise<Project>`:

Screenshot 2023-07-03 at 2 35 26 PM

styfle (approved):

PR #10178, opened 2023-07-05
[rbac]s show forbidden error in case of missing permission

balazs4:
fixes [iam-941](https://linear.app/vercel/issue/IAM-941/cli-improve-handling-of-insufficient-permissions-no-access)


anatrajkovska (commented):
Some of the tests are failing. Is it because of the changes in this PR, or unrelated?

EndangeredMassa:
The test failures appear to be related

balazs4:
@anatrajkovska They seem to be related. I am fixing/updating them.

balazs4:
Test are passing ✅

cb1kenobi (dismissed):
LGTM

anatrajkovska (dismissed):

packages/cli/src/util/client.ts
```diff
@@ -146,7 +146,8 @@ export default class Client extends EventEmitter implements Stdio {
       if (!res.ok) {
         const error = await responseError(res);
 
-        if (isSAMLError(error)) {
+        // we should force reauth only if error has a teamId
+        if (isSAMLError(error) && error.teamId !== null) {
```
styfle: What about when `teamId` is undefined?

Is there a difference between `teamId: null` and `teamId: undefined`?

balazs4: @styfle 
I found this once I tried accessing deployment logs with `vc logs <deployment-url>` where I don't have sufficient permission.
This is the [api](https://github.com/vercel/api/blob/540c0cac334ef33f33837445925613fc10abfaff/services/api-deployments/src/endpoints/get-deployment.ts?plain=1#L134-L155) code.

In case of `SAMLError` looks like `teamId` is always defined.

But I think I got your point: `client.ts` is generic, so it might be just `undefined`. However according [typedef](https://github.com/vercel/vercel/blob/4bf2ca55ff64b5994ee386b99d0e491f1a70c00a/packages/cli/src/util/login/types.ts?plain=1#L16) should not be the case. Although we are talking about error cases coming from `api` calls, so type check would not catch it, and runtime does not know about types.

I'll go with `&& error.teamId` instead. Good catch 👍

balazs4: see https://github.com/vercel/vercel/pull/10178/commits/fbccee9a6666caef435601ce5673846969ba2d05

cb1kenobi (approved):

styfle (approved):

packages/cli/src/commands/project/rm.ts
```diff
@@ -38,6 +38,10 @@ export default async function rm(client: Client, args: string[]) {
       client.output.error('No such project exists');
       return 1;
     }
+    if (isAPIError(err) && err.status === 403) {
+      client.output.error(err.message);
+      return 1;
+    }
```
styfle, line 44: This is for `vc rm` but I think we need a similar check in `vc link` and other commands

balazs4: `vc link` looks fine to me:

image

PR #10182, opened 2023-07-06
[WIP][node] don't decompress when body is buffered

Kikobeats:
This PR disabled `node-fetch` default compression handling.

The default behavior in node-fetch is to handle the compression. That's great if you interact with node-fetch as user, but bad if you use it as intermediate proxy server, since it's creating a mismatching related with the body format and length expectations.

In this way, our proxy server doesn't decompress the body, just pass it to the final client who can decompress properly.


packages/node/package.json
```diff
@@ -23,15 +23,15 @@
...
+    "@types/node-fetch": "2.6.4",
     "@vercel/build-utils": "6.8.0",
     "@vercel/error-utils": "1.0.10",
     "@vercel/static-config": "2.0.17",
     "async-listen": "3.0.0",
     "edge-runtime": "2.4.3",
     "esbuild": "0.14.47",
     "exit-hook": "2.2.1",
-    "node-fetch": "2.6.9",
+    "node-fetch": "2.6.12",
```
nuta, line 34: Looks an unrelated change. Let's upgrade in a separate PR if possible.

PR #10184, opened 2023-07-07
[node] fix decompress mismatching

Kikobeats:
This PR disabled `node-fetch` default compression handling when the response is not streamed.

The default behavior in node-fetch is to handle the compression. That's great if you interact with node-fetch as user, but bad if you use it as intermediate proxy server, since it's creating a mismatching related with the body format and length expectations.

Instead, we disable compression, get the buffered response and compress it again if needed.

### DIFF
--- a/packages/node/test/unit/dev.test.ts
+++ b/packages/node/test/unit/dev.test.ts
@@ -64,7 +64,7 @@ function testForkDevServer(entrypoint: string) {
   }
 );
 
-(NODE_MAJOR < 18 ? test.skip : test.only)(
+(NODE_MAJOR < 18 ? test.skip : test)(
   'buffer fetch response correctly',
   async () => {
     const child = testForkDevServer('./serverless-fetch.js');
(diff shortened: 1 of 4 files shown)


packages/node/test/unit/dev.test.ts
```diff
@@ -61,6 +64,90 @@ function testForkDevServer(entrypoint: string) {
   }
 );
 
+(NODE_MAJOR < 18 ? test.skip : test.only)(
```
EndangeredMassa: **suggestion (soft-blocking):** 

In this or a follow-up PR: ‏

```suggestion
(NODE_MAJOR < 18 ? test.skip : test)(
```

EndangeredMassa (dismissed):
Looks good to me, but I'm less familiar with the streaming/compression details

Ethan-Arrowood (dismissed):

Ethan-Arrowood (approved):

PR #10189, opened 2023-07-07
Revert "Add Changesets "ignore" configuration"

EndangeredMassa:
Reverts vercel/vercel#10181

This seems to be breaking releases of packages that depend on ignored packages. Example: https://github.com/vercel/vercel/actions/runs/5489491313/jobs/10003713249


EndangeredMassa:
The changeset bot is confused because I reverted a PR with an empty changeset and added a new empty changeset. Github says this is a file move, which I think is why the bot doesn't believe a new changeset file was created. I think we can merge overriding this check for this PR.

TooTallNate (approved):
I was afraid that was gonna happen <picture data-single-emoji=":doh:" title=":doh:">:doh:</picture>

PR #10201, opened 2023-07-12
[cli] Fix redeploy target to be `undefined` when `null`

cb1kenobi:
When redeploying an existing deployment, the target can be `null` and `null` will get sent, however the API only allows a string or `undefined`.


TooTallNate (dismissed):

Ethan-Arrowood (dismissed):

styfle (dismissed):
How about a test?

styfle (approved):
Thanks!

PR #10220, opened 2023-07-17
[node] fix: runs edge user code inside IIFE

Kikobeats:
In this way, `self` is isolated and modify it doesn't break some Edge Runtime internals

Originally reported by @jawj at https://github.com/jawj/neon-vercel-zapatos-minimal-crash

### DIFF
--- a/packages/node/test/unit/dev.test.ts
+++ b/packages/node/test/unit/dev.test.ts
@@ -148,6 +148,27 @@ function testForkDevServer(entrypoint: string) {
   }
 );
 
+test("user code doesn't interfer with runtime", async () => {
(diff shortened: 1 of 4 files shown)


EndangeredMassa (approved):

packages/node/test/unit/dev.test.ts
```diff
@@ -148,6 +148,27 @@ function testForkDevServer(entrypoint: string) {
   }
 );
 
+test("user code doesn't interfer with runtime", async () => {
```
styfle, line 151:
```suggestion
test("user code doesn't interfere with runtime", async () => {
```

PR #10224, opened 2023-07-17
[remix] Create ensured dependency symlink at the `start` directory, instead of root of repo

TooTallNate:
When the `ensureResolvable()` function runs in a monorepo, the symlink should be created in the `node_modules` directory within the app subdirectory, rather than the root of the repository. Fixes the following issue:

image


packages/remix/src/utils.ts
```diff
@@ -273,3 +276,96 @@ export function addDependency(
...
+    start,
+    filename: 'node_modules/.pnpm',
+  });
+  if (pnpmDir) {
+    const prefix = `${pkgName.replace('/', '+')}@`;
+    const packages = await fs.readdir(pnpmDir);
+    const match = packages.find(p => p.startsWith(prefix));
+    if (match) {
+      const pkgDir = join(pnpmDir, match, 'node_modules', pkgName);
+      await ensureSymlink(pkgDir, join(start, 'node_modules'), pkgName);
```
TooTallNate, line 316:
```diff
-await ensureSymlink(pkgDir, join(pnpmDir, '..'), pkgName);
+await ensureSymlink(pkgDir, join(start, 'node_modules'), pkgName);
```

This is the main fix in this PR. The rest of the diff was moving the function to the `utils` directory and adding a unit test for the updated behavior.

packages/remix/src/utils.ts
```diff
@@ -273,3 +276,96 @@ export function addDependency(
...
+    filename: 'node_modules/.store',
+  });
+  if (npmDir) {
+    const prefix = `${basename(pkgName)}@`;
+    const prefixDir = join(npmDir, dirname(pkgName));
+    const packages = await fs.readdir(prefixDir);
+    const match = packages.find(p => p.startsWith(prefix));
+    if (match) {
+      const pkgDir = join(prefixDir, match, 'node_modules', pkgName);
+      await ensureSymlink(pkgDir, join(start, 'node_modules'), pkgName);
```
TooTallNate, line 334: Here as well (see above comment).

EndangeredMassa (approved):
I couldn't reproduce the issue, but I tested this and it doesn't break remix deploys. So, looks good to me.

PR #10225, opened 2023-07-17
[cli] Use `getNodeBinPaths()` in `vc dev`

TooTallNate:
This allows for the "dev command" of a Project to work better in monorepos, where the dev server might live up the node_modules hierarchy within the repo.


styfle (commented):
Looks good, just needs tests 👍

cb1kenobi (approved):
CR'd and looks good! Did not test.

Ethan-Arrowood (approved):
I don't think tests need updated here since this seems like you are just replacing a previous api. LGTM 🚀

PR #10226, opened 2023-07-18
[cli] Update `proxy-agent` to v6.3.0

TooTallNate:
This version includes a refactor for proxies specified via PAC files such that it no longer uses the deprecated `vm2` module.

See https://github.com/TooTallNate/proxy-agents/issues/218.


pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
```
styfle, line 5: This is surprising to me. I wouldn't expect these here unless they were also defined in `.npmrc`.

styfle (approved):

pnpm-lock.yaml
cb1kenobi: I'm not sure we want this enabled. This setting may have been set by pnpm based on the `.npmrc`. I swear I saw something like this come up in the past couple weeks.

PR #10228, opened 2023-07-19
[frameworks][static-build] Add `ignorePackageJsonScript` configuration for Framework command settings

TooTallNate:
When this property is set to `true`, then the corresponding `package.json` script will not be invoked, allowing for the default setting value will be executed.

This is enabled for Storybook's `buildCommand`, since we do not want the "build" script to be invoked, since that would belong to the frontend application's build instead of Storybook's.

### DIFF
--- a/packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/.vercel/project.json
+++ b/packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/.vercel/project.json
@@ -2,8 +2,6 @@
   "orgId": ".",
   "projectId": ".",
   "settings": {
-    "framework": "storybook",
-    "buildCommand": "npm run build-storybook",
-    "outputDirectory": "storybook-static"
+    "framework": "storybook"
   }
 }
--- a/packages/frameworks/src/types.ts
+++ b/packages/frameworks/src/types.ts
@@ -46,7 +46,8 @@ export interface SettingValue {
    * When set to `true`, then the builder will not
    * invoke the equivalent script in `package.json`,
    * and instead will invoke the command specified in
-   * configuration setting directly.
+   * configuration setting directly. When this
+   * configuration is enabled, `value` must be a string.
    */
   ignorePackageJsonScript?: boolean;
 }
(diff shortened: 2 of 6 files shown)


packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/.vercel/project.json
```diff
@@ -2,8 +2,6 @@
   "orgId": ".",
   "projectId": ".",
   "settings": {
-    "framework": "storybook",
-    "buildCommand": "npm run build-storybook",
```
EndangeredMassa, line 6: **suggestion (soft-blocking):** test failing build command

In this or a follow-up PR: ‏Should we leave a bad build command in here to test that the ignore is happening properly?

EndangeredMassa (dismissed):

packages/frameworks/src/types.ts
```diff
@@ -32,11 +32,23 @@ export interface SettingPlaceholder {
...
+   * the user is configuring this setting value.
+   * @example "`npm run build` or `next build`"
+   */
   placeholder?: string;
+  /**
+   * When set to `true`, then the builder will not
+   * invoke the equivalent script in `package.json`,
+   * and instead will invoke the command specified in
+   * configuration setting directly.
+   */
```
styfle, line 51: Should we mention that `value` cannot be `null` when `ignorePackageJsonScript` is `true`?

Maybe even add a test to ensure that never happens in `@vercel/frameworks`

packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/.vercel/project.json
TooTallNate: If `buildCommand` is set here (when the user explicitly places an override on the build command in the UI), it takes highest precedence. What I wanted to test was that the default command (lowest precedence) is what gets executed here.

packages/frameworks/src/types.ts
TooTallNate: Sounds good. I updated the test and added a comment.

styfle (approved):
Great, thanks! 🎉

PR #10232, opened 2023-07-20
#10187: Respect .gitignore file when deploying

### LINKED ISSUE #10187: Respect .gitignore file when deploying
Vercel's cli to deploy is great BUT it should respect the gitignore file.

If I'm using Vercel via a GitHub repo connection it respects that - since pushing to GitHub respects it.

But if I'm using Vercel solely from a local repo via the cli, it does not respect that. Instead it requires me to add and maintain a `.vercelignore` file.

Respecting the `.gitignore` file would follow the principle of least surprise and keep users safer.

additional info: https://vercel.com/docs/concepts/deployments/vercel-ignore

AbdulRehmanMehar:
Added support for `.gitignore` along with `.vercelignore` and `.nowignore`. Here's the link to related Issue: https://github.com/vercel/vercel/issues/10187

### DIFF
--- a/packages/build-utils/src/get-ignore-filter.ts
+++ b/packages/build-utils/src/get-ignore-filter.ts
@@ -47,12 +47,23 @@ export default async function (
     rootDirectory || '',
     '.nowignore'
   );
+
+  const gitIgnorePath = path.join(
+    downloadPath,
+    rootDirectory || '',
+    '.gitignore'
+  );
+
   const ignoreContents = [];
 
   try {
     ignoreContents.push(
       ...(
-        await Promise.all([readFile(vercelIgnorePath), readFile(nowIgnorePath)])
+        await Promise.all([
+          readFile(gitIgnorePath),
+          readFile(vercelIgnorePath),
+          readFile(nowIgnorePath),
+        ])
       ).filter(Boolean)
     );
   } catch (error) {
@@ -63,9 +74,9 @@ export default async function (
     }
   }
 
-  if (ignoreContents.length === 2) {
+  if (ignoreContents.length >= 2) {
     throw new Error(
-      'Cannot use both a `.vercelignore` and `.nowignore` file. Please delete the `.nowignore` file.'
+      'Cannot use `.gitignore`, `.vercelignore` and `.nowignore` together. Please make sure that only one of these files is present.'
     );
   }
 
--- a/packages/client/src/utils/index.ts
+++ b/packages/client/src/utils/index.ts
@@ -152,7 +152,6 @@ export async function getVercelIgnore(
       '.vercel',
       '.npmignore',
       '.dockerignore',
-      '.gitignore',
       '.*.swp',
       '.DS_Store',
       '.wafpicke-*',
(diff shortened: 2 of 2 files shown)


packages/build-utils/src/get-ignore-filter.ts
```diff
@@ -47,12 +47,23 @@ export default async function (
     rootDirectory || '',
     '.nowignore'
   );
+
+  const gitIgnorePath = path.join(
+    downloadPath,
+    rootDirectory || '',
+    '.gitignore'
+  );
```
styfle, line 55: Do we need to handle recursive `.gitignore` files, for example with monorepos?

packages/build-utils/src/get-ignore-filter.ts
```diff
@@ -63,9 +74,9 @@ export default async function (
     }
   }
 
-  if (ignoreContents.length === 2) {
+  if (ignoreContents.length >= 2) {
     throw new Error(
-      'Cannot use both a `.vercelignore` and `.nowignore` file. Please delete the `.nowignore` file.'
+      'Cannot use `.gitignore`, `.vercelignore` and `.nowignore` together. Please make sure that only one of these files is present.'
```
styfle, line 79: This looks like a breaking change so it needs to be marked semver major

packages/client/src/utils/index.ts
```diff
@@ -172,19 +171,24 @@ export async function getVercelIgnore(
...
           maybeRead(join(cwd, '.vercelignore'), ''),
           maybeRead(join(cwd, '.nowignore'), ''),
+          maybeRead(join(cwd, '.gitignore'), ''),
         ]);
-        if (vercelignore && nowignore) {
+        if (
+          (vercelignore && nowignore) ||
+          (gitignore && vercelignore) ||
+          (gitignore && nowignore)
+        ) {
```
styfle, line 183: Let's use the same logic in the other file with `length >= 2`

packages/build-utils/src/get-ignore-filter.ts
cb1kenobi: This IS a breaking change.

cb1kenobi:
I'm -1 on this PR. `.gitignore` file is intended for git. I'm not convinced that the Vercel CLI should piggyback on another specific program's file, especially when A) the Vercel CLI already has a `.vercelignore` file and B) it introduces a major breaking behavior change.

There are users out there that are explicitly ignoring `vercel.json` in their `.gitignore` because they don't want it to be committed. If we start honoring `.gitignore`, then all the sudden all these users' projects are going to fail or worse be misconfigured.

Ethan-Arrowood:
FWIW - Prettier and ESLint also do not look at `.gitignore`. They require usage of `.prettierignore` and `.eslintignore` respectively.

AbdulRehmanMehar:
> .vercelignore

Thanks for providing valuable feedback. What are your thoughts about adding support for extending `.gitignore` in `.vercelignore`?

example `.vercelignore`:

```
[extends .gitignore]

node_modules
another_directory_to_ignore
```

EndangeredMassa:
Thanks for the PR, but we're not going to accept this feature at this time.

It's not clear that the Vercel CLI should do this from a product or user perspective, but even if we decided that it should, it's not clear how we should handle the complexity of the different formats, nesting, and user expectations.

If a user wants to make these match, they can copy the git ignore to vercel ignore and tweak the file to fit.

motdotla:
Thank you for considering. I was the original reporter. 

I agree with the complexity and especially the built-in history of product usage by current Vercel devs. Makes total sense.

But I'll still add additional context here if this is ever reconsidered:

As a new Vercel user, this product experience was a surprise. More significantly, I unknowingly pushed a sensitive credential file to your platform. I'm a long time Heroku user and their platform has always respected gitingore files rather than maintain their own ignore files. The solution with Vercel was easy though once I became aware of the vercelignore file. I just wasn't initially aware of it as a new user of the cli.

Thanks for a great service!

PR #10236, opened 2023-07-21
[cli] Do not remove in-progress deployments in safe mode

igordanchenko:
The `remove` command currently removes in-progress deployments when invoked in a safe mode (with `--safe` parameter). It can be totally unexpected that "safe" removal process deletes an in-progress deployment that is about to become production one. This PR addresses this use case by filtering out in-progress deployments.

### DIFF
--- a/.changeset/improve-safe-remove.md
+++ b/.changeset/improve-safe-remove.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Do not remove in-progress deployments in safe mode
--- a/packages/cli/src/commands/remove/index.ts
+++ b/packages/cli/src/commands/remove/index.ts
@@ -155,7 +155,11 @@ export default async function remove(client: Client) {
   }
 
   deployments = deployments.filter((match, i) => {
-    if (safe && aliases[i].length > 0) {
+    if (
+      safe &&
+      (aliases[i].length > 0 ||
+        ['QUEUED', 'INITIALIZING', 'BUILDING'].includes(match.status))
+    ) {
       return false;
     }
 
--- a/packages/cli/test/unit/commands/remove.test.ts
+++ b/packages/cli/test/unit/commands/remove.test.ts
@@ -56,4 +56,60 @@ describe('remove', () => {
 
     expect(deleteAPIWasCalled);
   });
+
+  it('does not remove in-progress deployments', async () => {
+    const user = useUser();
+
+    const project = useProject({
+      ...defaultProject,
+      id: '123',
+    });
+
+    useUnknownProject();
+
+    client.scenario.get('/v3/now/aliases', (_req, res) => {
+      res.json({ aliases: [] });
+    });
+
+    const testCases = [
+      ['READY', true],
+      ['QUEUED', false],
+      ['BUILDING', false],
+      ['INITIALIZING', false],
+    ] as const;
+
+    const deployments = testCases
+      .map(([state]) =>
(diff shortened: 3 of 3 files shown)


TooTallNate (commented):
The change itself looks good, but would be best to add a test to prevent this from regressing in the future.

igordanchenko:
I'm more than happy to add a unit test for this specific scenario, but I'm unable to locate any unit tests for the `remove` command at all... Am I looking in the wrong place?

cb1kenobi:
@igordanchenko you are correct, we currently do not have any unit tests for the `remove` command. :(

Maybe we/Vercel can just manually test this?

igordanchenko:
Is there anything I can do to move the needle?

trek:
Hey @igordanchenko I'm going to try to make some unit tests for existing `remove` behavior today.

trek:
@igordanchenko I've added a test that still needs review but if you wanted to work off it, I think it's in a good place:

https://github.com/vercel/vercel/pull/10304/files#diff-94eee5d8843e8276fdcc839a10a4e4ee2e68b86c302b9c9c129b3b8d59623a1bR33

In your case we'd want to assert that `deleteAPIWasCalled` was _not_ called for a deployment in the specific states you're looking at.

igordanchenko:
@trek thank you for sharing this draft. 

I'm afraid there is an issue with mocked deployment status in unit tests. As I can see, mocked deployment status is originally set in the `status` field [here](https://github.com/vercel/vercel/blob/877f09ff5c99fdf8710485dd9e70b0119567bd8a/packages/cli/test/mocks/deployment.ts#L67), but then [this](https://github.com/vercel/vercel/blob/877f09ff5c99fdf8710485dd9e70b0119567bd8a/packages/cli/src/util/deploy/get-deployments-by-project-id.ts#L74-L101) conversion from legacy format wipes it out as it expects it to be in the `state` field. As a result, all mocked deployments have their `status` as `undefined`. Can you please take a look?

trek:
@igordanchenko can you push up your failing test? I altered my test to created a `BUILDING` mock deployment and am not seeing how the endpoint you linked to comes into play.

igordanchenko:
@trek I've pushed failing test case. I guess invocation of the `remove` command with project name as an argument is key here.

```js
client.setArgv('remove', `${project.project.name}`, '--safe', '--yes')
```

trek:
@igordanchenko just to clarify the behavior you're interested in is using `vercel remove projectname --safe` to remove all deployments for a project except those in the specific states you've mentioned?

igordanchenko:
The use case is to remove old production and preview deployments that are no longer in use.

trek:
@igordanchenko followup question: what's the need for deleting older deploys en masse?

igordanchenko:
> what's the need for deleting older deploys en masse?

General housekeeping, data retention policy compliance.

igordanchenko:
@trek I've updated the unit test with a temporary workaround for #10316. The unit test is passing, and the workaround can be removed if/when #10316 is resolved. Please see if this PR can be reviewed and merged.

trek:
@igordanchenko I doubt https://github.com/vercel/vercel/pull/10316 will get resolved. More likely we'll update our mocks to be more structured but I don't want to block this PR. That said, I'm too new on the team to know if this is a change we want, so I'll leave that to @TooTallNate @Ethan-Arrowood @EndangeredMassa @styfle and @cb1kenobi

PR #10239, opened 2023-07-21
[cli] Add "Environment" column to `vc list`

cb1kenobi:
Also adds a new `--environment=[preview|production]` filter and ability to pipe deployment URLs to `stdout`!

image

```
vc list --environment preview
```

```
vc list --environment production
```

```
vc list --environment preview > preview_deployments.txt
```


packages/cli/src/commands/pull.ts
```diff
@@ -35,6 +35,7 @@ const help = () => {
     'DIR'
   )}    Path to the global ${'`.vercel`'} directory
     -d, --debug                    Debug mode [off]
+    --git-branch                   Specify the Git branch to pull specific Environment Variables for
```
TooTallNate, line 38: This seems unrelated to the PR?

cb1kenobi: Um, what the

TooTallNate (approved):

packages/cli/src/commands/pull.ts
cb1kenobi: Yes, unrelated. I found that option was missing when I documented `--git-branch` last week. Want me to move this to a separate PR?

trek (commented):
Tested locally and all looks good.

PR #10244, opened 2023-07-24
support for detecting bun as a package manager

james-elicx:
This PR does the following:
- Introduces support for detecting/using Bun as a package manager in projects.
- Adds relevant unit tests.

Note that I added unit tests where I saw similar ones for other package managers, but when it comes to e2e tests, the fixture for Bun is currently added to the `skipFixtures` array as Bun doesn't appear to be part of Vercel's build image yet.

I would understand if you wish to hold off from reviewing this PR until Bun is part of Vercel's build image. I opened it as it's a useful nice-to-have, and I intend to support using Bun in Next.js deployments made with `@cloudflare/next-on-pages`, and we use the Vercel CLI when building Next.js apps so this would be beneficial for us.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -284,26 +284,34 @@ export async function scanParentDirs(
...
+  const [hasYarnLock, packageLockJson, pnpmLockYaml, hasBunLock] =
+    await Promise.all([
+      Boolean(yarnLockPath),
+      npmLockPath
+        ? readConfigFile<{ lockfileVersion: number }>(npmLockPath)
+        : null,
+      pnpmLockPath
+        ? readConfigFile<{ lockfileVersion: number }>(pnpmLockPath)
+        : null,
+      Boolean(bunLockPath),
```
styfle, line 311: Does bun have a lock file version corresponding to its CLI version?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -451,6 +462,9 @@ export async function runNpmInstall(
       commandArgs = args
         .filter(a => a !== '--prefer-offline')
         .concat(['install', '--unsafe-perm']);
+    } else if (cliType === 'bun') {
+      opts.prettyCommand = 'bun install';
+      commandArgs = ['install', ...args];
```
styfle, line 467: This will not work since the `bun` binary is not installed on Vercel yet.

styfle (changes requested):
Have you tried this yet? 

I wouldn't expect it to work since bun isn't installed in the build container

packages/build-utils/src/fs/run-user-scripts.ts
james-elicx: There is a hardcoded value `bun-lockfile-format-v0` at the start of the lockfile. I couldn't tell if this is supposed to relate to the CLI version or not.

https://github.com/oven-sh/bun/blob/0297cb/src/install/lockfile.zig#L4000

james-elicx:
> Have you tried this yet?
> 
> I wouldn't expect it to work since bun isn't installed in the build container

Hello @styfle, that is correct. I touched on this a little further above, but until Vercel's build image has support for Bun, it won't work there (which is why I added the e2e fixture to the `skipFixtures` array). That's not to say this isn't useful though - Bun can still be used with the CLI locally, in other projects, or in other environments, just not as a package manager on Vercel.

A possible compromise could be checking if an environment variable like `VERCEL` is set (I assume you have one for deployments?) and, if so, throwing an error instead of continuing with the build. At least, until your build image supports Bun. What do you think about this?

My use case isn't strictly related to Vercel's build image - we use the Vercel CLI to build apps in `@cloudflare/next-on-pages`, so this would be beneficial for us, however, as I wrote above, I would understand if there is hesitation to add this before Bun is part of your build image.

styfle:
I think its best to configure the Install Command and Build Command for your project, because we can't add zero config support like this without also adding bun to the build container.

https://vercel.com/docs/concepts/deployments/configure-a-build

Furthermore, bun is not just a package manager but an entire runtime. So this begs the question - does the user expect bun to work with Serverless Functions as a new runtime?

https://vercel.com/docs/concepts/functions/serverless-functions/runtimes

james-elicx:
> I think its best to configure the Install Command and Build Command for your project, because we can't add zero config support like this without also adding bun to the build container.
> 
> [vercel.com/docs/concepts/deployments/configure-a-build](https://vercel.com/docs/concepts/deployments/configure-a-build)

Yeah, I'm currently creating a temporary config file in my PR over in next-on-pages to leverage those options. It just would be nice to have out-of-the-box support in the Vercel CLI as well instead of using a temporary config for our users.

I can understand preferring not to proceed with adding it to the Vercel CLI at this time though while its not in the build image 👍🏻

> Furthermore, bun is not just a package manager but an entire runtime. So this begs the question - does the user expect bun to work with Serverless Functions as a new runtime?
> 
> [vercel.com/docs/concepts/functions/serverless-functions/runtimes](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes)

Bun is many different things - one of its features is a Node.js-compatible package manager that is incredibly impressive when compared to other package managers, and can work as a drop-in replacement in many cases. This would solely be looking at supporting using it as a package manager, not whether Vercel is adding first-class support for a new runtime.

I would add that Bun is also being looked at to support as a package manager in the Cloudflare Pages build image fwiw.

EndangeredMassa:
We don't currently plan to support `bun` as a package manager or a runtime. If we do, we'll want to do a lot of testing around that support because every package manager has many quirks that we have to accommodate.

If `bun` was supported by `corepack`, we'd be more interested in supporting it. Feel free to open an issue about it in the [corepack](https://github.com/nodejs/corepack) repo.

PR #10248, opened 2023-07-25
enable localePrefixed404 when i18n is available

LuizPelegrini:
## Context
Upgrading to [vercel@31.1.0](https://github.com/vercel/vercel/releases/tag/vercel%4031.1.0) in a non-app router Next project that uses i18n gives the following error, when using `vercel build` command
```
Error: ENOENT: no such file or directory, open '/vercel/path1/.next/server/pages/404'
``` 

## Expected Outcome
This PR aims to fix the above build error by simply enabling back the flag `localePrefixed404` after if i18n is present in project

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -217,10 +217,10 @@ export async function serverBuild({
       : undefined;
 
   if (!static404Page && i18n) {
+    localePrefixed404 = true;
     if (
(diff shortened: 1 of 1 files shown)


ijjk (changes requested):
Hi, this doesn't appear correct as this flag should be enabled if a locale variant is present. Can you share a repro where you are seeing this not work properly?

PR #10250, opened 2023-07-25
[cli] Add a "Global Options" section to help output

trek:
Followup PR to #10090. Some of our commands duplicate global options into their `args` structure e.g. https://github.com/vercel/vercel/blob/2661f56347d5c32de369a6cea74bc9fb46d565a2/packages/cli/src/commands/logs.ts#L25-L26

 Others commands omit it entirely.

This updates the `--help` output for commands migrated to our new structure (so far, only `deploy`) will show a "Global Options" section:

```
▲ vercel deploy [project-path] [options]

Deploy your project to Vercel. The `deploy` command is the default command for the Vercel CLI, and can be omitted (`vc deploy my-app` equals `vc my-app`).

Options

  --archive                    Compress the deployment code into a file before uploading it
  -b, --build-env <key=value>  Specify environment variables during build-time (e.g. `-b KEY1=value1 -b KEY2=value2`)
  -e, --env <key=value>        Specify environment variables during run-time (e.g. `-e KEY1=value1 -e KEY2=value2`)
  -f, --force                  Force a new deployment even if nothing has changed
  -m, --meta <key=value>       Specify metadata for the deployment (e.g. `-m KEY1=value1 -m KEY2=value2`)
  --no-wait                    Don't wait for the deployment to finish
  --prebuilt                   Use in combination with `vc build`. Deploy an existing build
  --prod                       Create a production deployment
  -p, --public                 Deployment is public (`/_src`) is exposed)
  --regions                    Set default regions to enable the deployment on
  --with-cache                 Retain build cache when using "--force"
  -y, --yes                    Use default options to skip all prompts

Global Options

  --cwd <DIR>                Sets the current working directory for a single run of a command
  -d, --debug                Debug mode (default off)
  -Q, --global-config <DIR>  Path to the global ${'`.vercel`'} directory
  -h, --help                 Output usage information
  -A, --local-config <FILE>  Path to the local `vercel.json` file
  --no-color                 No color mode (default off)
  -S, --scope                Set a custom scope
  -t, --token <TOKEN>        Login token
  -v, --version              Output the version number
```

As commands are migrated to this new structure, they'll gain this output automatically.


packages/cli/src/commands/help.ts
```diff
@@ -209,7 +292,9 @@ export function buildHelpOutput(
     '',
     command.description,
     '',
-    buildCommandOptionLines(command, options),
+    buildCommandOptionLines(command.options, options, 'Options'),
+    '',
+    buildCommandOptionLines(globalCommandOptions, options, 'Global Options'),
```
cb1kenobi, line 297:
```suggestion
    buildCommandOptionLines(command.options, options, 'Options:'),
    '',
    buildCommandOptionLines(globalCommandOptions, options, 'Global Options:'),
```

packages/cli/src/commands/help.ts
```diff
@@ -75,27 +155,32 @@ export function buildCommandSynopsisLine(command: Command) {
...
     option => !option.deprecated && option.description !== undefined
   );
 
+  if (commandOptions.length === 0) {
+    return '';
+  }
+
   // Initialize output array with header and empty line
-  const outputArray: string[] = [chalk.dim(`Options:`), ''];
+  const outputArray: string[] = [chalk.dim(sectionTitle), ''];
```
cb1kenobi: Please preserve the semicolon :)

```suggestion
  const outputArray: string[] = [`${chalk.dim(sectionTitle)}:`, ''];
```

cb1kenobi (changes requested):
Looks good, just 1 small change.

packages/cli/src/commands/help.ts
```diff
@@ -30,6 +30,86 @@ export interface Command {
...
+    description: 'Path to the local `vercel.json` file',
+    deprecated: false,
+    multi: false,
+  },
+  {
+    name: 'global-config',
+    shorthand: 'Q',
+    type: 'string',
+    argument: 'DIR',
+    description: "Path to the global ${'`.vercel`'} directory",
```
cb1kenobi:
```suggestion
    description: "Path to the global `.vercel` directory",
```

cb1kenobi (approved):

PR #10254, opened 2023-07-26
Fix matched path regex with basePath

ijjk:
Seems the `x-matched-path` header isn't being set as expected for the index of the `basePath` which is breaking this test so this makes the regex match as expected. 

Fixes: https://github.com/vercel/vercel/actions/runs/5660473696/job/15342856245?pr=10249#step:9:563


ijjk:
Closing in favor of fixing on the Next.js side

PR #10255, opened 2023-07-26
[node] bump edge-runtime to 2.4.4

hassanbazzi:
# Summary

Edge-runtime 2.4.4 has many bug fixes, but most importantly it adds support for stream cancellation to the edge runtime. This is extremely important since a lot of projects are using `streams` related to `ai`. They currently have no way of handling a cancellation coming from the client.

This was introduced to `next` with as described by this comment: https://github.com/vercel-labs/ai/issues/90#issuecomment-1618915409

You can find the PR for that here: https://github.com/vercel/next.js/pull/51727

It also has a good description for what we're trying to do here, but for people not using `next`

# Problem

When a client sends an abort signal, it is currently not being handled by edge functions. This was fixed in edge-runtime@2.4.4

# Solution

Update the package


cb1kenobi (approved):
Tests successful ran! https://github.com/vercel/vercel/pull/10326

TooTallNate (approved):
Thanks!

.changeset/famous-timers-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/node': minor
```
styfle, line 2: Any reason you chose minor? Is this adding a new feature?

.changeset/famous-timers-mix.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Update 'edge/runtime' to 2.4.4
```
styfle, line 5:
```suggestion
Update 'edge-runtime' to 2.4.4
```

.changeset/famous-timers-mix.md
TooTallNate: https://github.com/vercel/vercel/pull/10328

PR #10256, opened 2023-07-26
[fs-detectors] fix: Next.js builds not always excluding Node.js middleware builder

james-elicx:
This PR does the following:
- Fixes the null equality check for Next.js builds when middleware is present and the project settings `framework` option is `undefined`.
- Adds a test for the fix.

### DIFF
--- a/packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
+++ b/packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
@@ -2359,6 +2359,26 @@ describe('Test `detectBuilders` with `featHandleMiss=true`', () => {
     ]);
   });
 
+  it('should not add middleware builder when building "nextjs" + undefined project settings', async () => {
+    const files = ['package.json', 'pages/index.ts', 'middleware.ts'];
+    const pkg = {
+      scripts: { build: 'next build' },
+      dependencies: { next: '12.2.0' },
+    };
+    const { builders } = await detectBuilders(files, pkg, {
+      featHandleMiss,
+    });
+    expect(builders).toEqual([
+      {
+        use: '@vercel/next',
+        src: 'package.json',
+        config: {
+          zeroConfig: true,
+        },
+      },
+    ]);
+  });
+
   it('should add middleware builder with "remix" framework preset', async () => {
     const files = ['package.json', 'app/routes/index.ts', 'middleware.ts'];
     const projectSettings = {
--- a/.changeset/lazy-parrots-guess.md
+++ b/.changeset/lazy-parrots-guess.md
@@ -1,0 +1,5 @@
+---
+'@vercel/fs-detectors': patch
+---
+
+Fix Next.js apps not excluding the Node.js builder for middleware when `framework` is `undefined` in the project settings.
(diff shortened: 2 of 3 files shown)


packages/fs-detectors/test/unit.builds-and-routes-detector.test.ts
```diff
@@ -2359,6 +2359,26 @@ describe('Test `detectBuilders` with `featHandleMiss=true`', () => {
...
+    };
+    const { builders } = await detectBuilders(files, pkg, {
+      featHandleMiss,
+    });
+    expect(builders).toEqual([
+      {
+        use: '@vercel/next',
+        src: 'package.json',
+        config: {
+          zeroConfig: true,
```
styfle, line 2376: This should never happen. This should have `framework: nextjs` here if it really is Next.js

This might be related to my previous comment here: https://github.com/vercel/vercel/pull/8239#discussion_r929081740

cc @nuta

james-elicx: Hello, I copied the test directly above this one and changed the arguments that `detectBuilders()` is called with so that it will have `undefined` for `projectSettings` and, by proxy, the `framework` field in the project settings (which is what happens with the field there is no Vercel config file in a project). The above test was testing when the `framework` option had the value `null`. This is testing when it is `undefined` as this is fixing a long-time bug with that case.

The outputted builder is the exact same that the other Next.js tests uses.

styfle: I understand the test you created, but I dont understand the conditions for which it can happen.

Do you have a deployment url where this happened?

james-elicx: It happens when you build a Next.js project with no Vercel config file. You can clone https://github.com/james-elicx/app-playground and run `npx vercel build`.

You will see that it builds an extra `middleware.js` file for the middleware function from the nodejs builder because the `framework` value was `undefined`, and therefore it didn't exclude the nodejs builder. It's pretty simple really; no Vercel config leads to the value being undefined, but the equality check is only checking it is strictly equal to null and not accounting for circumstances where it's undefined, like when there's no config.

I don't have a deployment URL because I don't deploy to Vercel, I'm referring to building an app locally or in other environments.

styfle: The issue is that you have a fake `project.json` file here: https://github.com/james-elicx/app-playground/blob/main/.vercel/project.json

If you run through the steps to create a project, then it will work correctly:

```
vc link
vc pull
vc build
```

Here are the settings I see from a new project:

```json
{
  "projectId": "prj_...",
  "orgId": "team_...",
  "settings": {
    "createdAt": 1690568775159,
    "framework": "nextjs",
    "devCommand": null,
    "installCommand": null,
    "buildCommand": null,
    "outputDirectory": null,
    "rootDirectory": null,
    "directoryListing": false,
    "nodeVersion": "18.x"
  }
}
```

styfle (changes requested):
Closing since `zeroConfig: true` will never have `framework: undefined`

PR #10266, opened 2023-07-28
[cli] Migrate `whoami` to new command structure

trek:
Migrating `whoami` to the command structure created in #10090. Unsure about tests. This is technically a refactor so existing tests should suffice.


packages/cli/src/commands/whoami/command.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const whoamiCommand: Command = {
+  name: 'whoami',
+  description: 'Shows the username of the currently logged in user',
```
cb1kenobi: Missing period:

```suggestion
  description: 'Shows the username of the currently logged in user.',
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #10268, opened 2023-07-28
[cli] Migrate `remove` command to new structure

trek:
Migrating `remove` to the command structure created in #10090. Unsure about tests. This is technically a refactor so existing tests should suffice.


cb1kenobi (approved):

packages/cli/src/commands/remove/command.ts
```diff
@@ -0,0 +1,45 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const removeCommand: Command = {
+  name: 'remove',
+  description: 'Remove a deployment by name or id',
```
cb1kenobi: This should probably end with a period.

```suggestion
  description: 'Remove a deployment by name or id.',
```

cb1kenobi (changes requested):

trek:
Had to force push. Github got into some weird state where newer commits did not appear, but attempts to bring the branch up to date via the Github UI errored.

cb1kenobi (approved):

packages/cli/src/commands/remove/command.ts
EndangeredMassa: @cb1kenobi I don't think this was a blocking comment

cb1kenobi: Correct, however consistency is king.

PR #10271, opened 2023-07-30
[node] Fix dependencies

LitoMore:
### Problem

I got an error `Error: Cannot find module 'content-type'` after I upgrade my `vercel` to the latest version `31.2.0`.

### How to reproduce

You can use my Vercel app to reproduce this error:

1. Clone https://github.com/LitoMore/simple-icons-cdn
2. Install the latest version of `vercel`
3. Run `vercel dev`
4. Visit `https://localhost:3000/vercel` on your local machine

### Changes

- Move `content-type` from `devDependencies` to `dependencies`
- Add `querystring` to `@vercel/node`

### Root cause

The usage of these two packages was added in #10109.

We should add these dependencies to the `dependencies` field.

### Additonal comments

The `settings` field in `pnpm-lock.yaml` was updated automatically with my `pnpm` script. Feel free to revert this change if you don't want it.


cb1kenobi:
Huge thanks for the PR! We just merged #10274 which addresses the same issue. We'll get the fix released asap. Thanks!

cb1kenobi:
Vercel CLI 31.2.1 has been release with the fix. Thanks!

PR #10275, opened 2023-07-31
[next] Preserve sourceMappingURL comments in template literals

nuta:
We noticed some edge function builds are failing unexpectedly due to `SyntaxError`. It's triggered by valid code in [Webpack's style-loader](https://github.com/webpack-contrib/style-loader/blob/16e401b17a39544d5c8ca47c9032f02e2b60d8f5/src/runtime/styleDomAPI.js#L35C1-L40C1):

```
  if (sourceMap && typeof btoa !== "undefined") {
    css += `\n/*# sourceMappingURL=data:application/json;base64,${btoa(
      unescape(encodeURIComponent(JSON.stringify(sourceMap)))
    )} */`;
  }
```

After minification this script looks like:

```
sourceMap&&typeof btoa!="undefined"&&(css+=`
/*# sourceMappingURL=data:application/json;base64,${btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap))))} */`);
```

The key point here is `\n` is replaced with newline character (`0x0A`) and the second line starts with `/*# sourceMappingURL=`. This confuses `removeComments` API in [convert-source-map](https://github.com/thlorenz/convert-source-map/tree/master) package and it removes the second line:

```
sourceMap&&typeof btoa!="undefined"&&(css+=`

```

In this PR, we implement source map stripping based on that package's regex but add some heuristic checks to prevent false positives like above.


javivelasco (dismissed):
Incredible dig deep 👑

craigandrews (dismissed):

ecklf:
LFG

ijjk (dismissed):

trek (dismissed):

packages/next/src/sourcemapped.ts
```diff
@@ -99,3 +99,42 @@ export function stringifySourceMap(
   delete obj.sourcesContent;
   return JSON.stringify(obj);
 }
+
+// Based on https://github.com/thlorenz/convert-source-map/blob/f1ed815b4edacfa9c3c5552dd342e71a3cffbb0a/index.js#L4 (MIT license)
+// Groups: 1: media type, 2: MIME type, 3: charset, 4: encoding, 5: data.
+const SOURCE_MAP_COMMENT_REGEX =
+  // eslint-disable-next-line no-useless-escape
```
cb1kenobi: Instead of disabling the eslint rule, can we remove the useless escapes?

packages/next/src/sourcemapped.ts
```diff
@@ -99,3 +99,42 @@ export function stringifySourceMap(
...
+  /^\s*?\/[\/\*][@#]\s+?sourceMappingURL=data:(((?:application|text)\/json)(?:;charset=([^;,]+?)?)?)?(?:;(base64))?,(.*?)$/gm;
+
+function isValidSourceMapData(encoding: string, data: string): boolean {
+  // Remove any spaces and the trailing `*/`.
+  data = data.replace(/\s/g, '').replace('*/', '');
```
cb1kenobi: Can we move this `data` assignment after the `encoding` check? There's no need to do two replacements when the `encoding` is not `base64`.

cb1kenobi (changes requested):

packages/next/src/sourcemapped.ts
```diff
@@ -99,3 +99,42 @@ export function stringifySourceMap(
...
 }
+
+// Based on https://github.com/thlorenz/convert-source-map/blob/f1ed815b4edacfa9c3c5552dd342e71a3cffbb0a/index.js#L4 (MIT license)
+// Groups: 1: media type, 2: MIME type, 3: charset, 4: encoding, 5: data.
+const SOURCE_MAP_COMMENT_REGEX =
+  // eslint-disable-next-line no-useless-escape
+  /^\s*?\/[\/\*][@#]\s+?sourceMappingURL=data:(((?:application|text)\/json)(?:;charset=([^;,]+?)?)?)?(?:;(base64))?,(.*?)$/gm;
+
+function isValidSourceMapData(encoding: string, data: string): boolean {
+  // Remove any spaces and the trailing `*/`.
```
styfle: The comment says "trailing" but `.replace()` will only replace the first one.

For example, try this:

```js
'/*first*/ and /*second*/'.replace('*/', '')
```

It doesn't replace the trailing.

nuta: Good point, thanks!

styfle: @nuta It might be good to add a test with a comment like that to see how it behaves now

cb1kenobi (dismissed):
LGTM

packages/next/src/sourcemapped.ts
nuta: Good idea. Added!

cb1kenobi (approved):

styfle (approved):
Great, thanks! 🎉

ijjk (approved):

nuta:
Thank you all for helping me to make this fix more robust! For the record, I'm toggling `pr:automerge` label as CI got stuck.

PR #10276, opened 2023-07-31
[cli] Migrate `bisect` command to new structure

trek:
Migrating `bisect` to the command structure created in https://github.com/vercel/vercel/pull/10090. Unsure about tests. This is technically a refactor so existing tests should suffice.

Before:
CleanShot 2023-07-31 at 08 14 29@2x

After:
CleanShot 2023-07-31 at 08 15 02@2x


packages/cli/src/commands/bisect/command.ts
```diff
@@ -0,0 +1,69 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const bisectCommand: Command = {
+  name: 'bisect',
+  description: 'Bisect the current project interactively',
```
cb1kenobi: End the desc with a period.

```suggestion
  description: 'Bisect the current project interactively.',
```

cb1kenobi (changes requested):

packages/cli/src/commands/bisect/command.ts
trek: It's not currently. Is this an update you'd like to see for all descriptions?

cb1kenobi (approved):

PR #10281, opened 2023-07-31
[cli] Migrate `logs` command to new structure

trek:
Before:
CleanShot 2023-07-31 at 09 36 25@2x

After:
CleanShot 2023-07-31 at 09 36 46@2x


packages/cli/src/commands/logs/command.ts
```diff
@@ -0,0 +1,66 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const logsCommand: Command = {
+  name: 'logs',
+  description: 'Print vercel logs for a deployment.',
```
cb1kenobi: I don't think we need to say "vercel" and it's nice to say "specific" as it implies there's an argument required.

```suggestion
  description: 'Display logs for a specific deployment.',
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #10282, opened 2023-07-31
[cli] Migrate `logout` command to new structure

trek:
Before:
CleanShot 2023-07-31 at 09 40 48@2x

After:
CleanShot 2023-07-31 at 09 41 04@2x

### DIFF
--- a/packages/cli/src/commands/logout/command.ts
+++ b/packages/cli/src/commands/logout/command.ts
@@ -3,7 +3,7 @@ import { getPkgName } from '../../util/pkg-name';
 
 export const logoutCommand: Command = {
   name: 'logout',
-  description: 'Logout of the currently authenticated user or team.',
+  description: 'Logout the current authenticated user or team.',
   arguments: [],
   options: [],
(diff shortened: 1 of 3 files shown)


cb1kenobi (changes requested):
There seems to be a bit of excess whitespace when the command has no options: 
image

trek:
@cb1kenobi can we put that into the work for fixing spacing?

packages/cli/src/commands/logout/command.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const logoutCommand: Command = {
+  name: 'logout',
+  description: 'Logout of the currently authenticated user or team.',
```
cb1kenobi: Something about this reads funny. How do you feel about slimming it down?

```suggestion
  description: 'Logout the current authenticated user or team.',
```

cb1kenobi (approved):

PR #10283, opened 2023-07-31
[cli] Migrate `login` command to new structure

trek:
We currently incorrectly specify the argument is required with `<>`, but the examples show you can do `vercel login` without an argument, so I've corrected that as part of this migration.

Before:
CleanShot 2023-07-31 at 09 44 28@2x

After:
CleanShot 2023-07-31 at 09 44 43@2x

### DIFF
--- a/packages/cli/src/commands/login/command.ts
+++ b/packages/cli/src/commands/login/command.ts
@@ -1,0 +1,49 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const loginCommand: Command = {
+  name: 'login',
+  description: 'Authenticate a email or team.',
+  arguments: [
+    {
+      name: 'email or team',
+      required: false,
+    },
+  ],
+  options: [
+    {
+      name: 'github',
+      description: 'Log in with github',
+      shorthand: null,
+      type: 'string',
+      deprecated: false,
+      multi: false,
+    },
+    {
+      name: 'oob',
+      description: 'Log in with "out of band" authentication',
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/login/command.ts
```diff
@@ -0,0 +1,49 @@
...
+  arguments: [
+    {
+      name: 'email or team',
+      required: false,
+    },
+  ],
+  options: [
+    {
+      name: 'github',
+      description: 'Log in with github',
```
cb1kenobi:
```suggestion
      description: 'Log in with GitHub',
```

packages/cli/src/commands/login/command.ts
```diff
@@ -0,0 +1,49 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const loginCommand: Command = {
+  name: 'login',
+  description: 'Authenticate a email or team.',
```
cb1kenobi: This doesn't sound accurate.

```suggestion
  description: 'Authenticate using your email or team id.',
```

packages/cli/src/commands/login/command.ts
```diff
@@ -0,0 +1,49 @@
...
+  arguments: [
+    {
+      name: 'email or team',
```
cb1kenobi:
```suggestion
      name: 'email or team id',
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #10285, opened 2023-07-31
[cli] Migrate `link` command to new structure

trek:
Before:
CleanShot 2023-07-31 at 09 55 03@2x

After:
CleanShot 2023-07-31 at 09 55 03@2x


packages/cli/src/commands/link/command.ts
```diff
@@ -0,0 +1,54 @@
+import { Command } from '../help';
+import { getPkgName } from '../../util/pkg-name';
+
+export const linkCommand: Command = {
+  name: 'inspect',
```
cb1kenobi:
```suggestion
  name: 'link',
```

packages/cli/src/commands/link/command.ts
```diff
@@ -0,0 +1,54 @@
...
+  description: 'Show information about a deployment.',
```
cb1kenobi:
```suggestion
  description: 'Link a local directory to a Vercel Project.',
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #10287, opened 2023-07-31
[cli] Update spacing on `--help` output

trek:
Indent more items rightwards to tidy up spacing.

Example of unmigrated `bisect` on current `main`:
CleanShot 2023-07-31 at 14 47 38@2x

Migrated example (https://github.com/vercel/vercel/pull/10276) before changes in this PR:
CleanShot 2023-07-31 at 15 07 26@2x

Migrated example after changes in this PR:
CleanShot 2023-07-31 at 14 46 39@2x


packages/cli/src/commands/help.ts
```diff
@@ -273,12 +278,15 @@ export function buildCommandExampleLines(command: Command) {
...
     outputArray.push('');
   }
-  // delete the last newline added after examples iteration
-  outputArray.splice(-1);
 
   return outputArrayToString(outputArray);
 }
 
+function buildDescriptionLine(command: Command) {
+  const line: string[] = [INDENT, command.description];
```
cb1kenobi: Let's skip indenting the description. When the description is long, it wraps and looks bad.

packages/cli/src/commands/help.ts
```diff
@@ -142,7 +142,12 @@ export function outputArrayToString(outputArray: string[]) {
  * @returns
  */
 export function buildCommandSynopsisLine(command: Command) {
-  const line: string[] = [LOGO, chalk.bold(NAME), chalk.bold(command.name)];
+  const line: string[] = [
+    INDENT,
```
cb1kenobi, line 146: Because I don't think we should indent the description (see below), I don't think we should indent the command.

cb1kenobi (changes requested):

packages/cli/src/commands/help.ts
TooTallNate: I like the spacing personally. Perhaps for a long description we should automatically add newlines at the terminal width so that we can add spacing on subsequent lines to keep things aligned.

TooTallNate (approved):

packages/cli/src/commands/help.ts
cb1kenobi: It's tough. I like the indention until the line wraps. @trek and I discussed word wrapping, but I had concerns with conflicts if we ever choose to manually format the description (e.g. paragraphs, embedded example, links). Perhaps we get all these PRs merged, then when we have some data, we can revisit the indentation?

trek: I'm currently playing with https://www.npmjs.com/package/word-wrap and https://www.npmjs.com/package/cli-table3 to see if I can get more consistent indenting with less custom code. But it might be a fool's errand.

cb1kenobi (approved):

PR #10288, opened 2023-07-31
fix: compress condition

Kikobeats:
I'm experiencing an issue due the condition is wrongly set.

When the response is buffered, it should be `compress: true` since we're acting as proxy, and we're going to recompress again.

When the response is streaming, then it should be `compress: false` since the response is going to be piped.


styfle (dismissed):
Looks good! Can you add a test?

cb1kenobi (dismissed):
YES! This is awesome!!! Huge thank you. I ran into this issue last week and created a ticket https://linear.app/vercel/issue/VCCLI-814/vc-dev-doesnt-respect-content-encoding-response-header and now we can resolve it. :)

trek (commented):
Just double checking so we don't ping pong between the value of `compress` if we find out that there was a reason for it being set `false`

The change from
https://github.com/vercel/vercel/commit/5e5332fbc9317a8f3cc4ed0b72ec1a2c76020891#diff-e6873fb4f9050c6292e6593621207fc74ab591b55b4cbdc2551bb194bda9c953R110

was in error?

packages/cli/test/dev/fixtures/43-compress-encoding/vercel.json
```diff
@@ -0,0 +1,3 @@
+{
+  "builds": [{ "src": "**/*.js", "use": "@vercel/node" }]
```
styfle: Lets move the `index.js` to `api/index.js` and then we can remove `vercel.json`

packages/cli/test/dev/fixtures/43-compress-encoding/index.js
```diff
@@ -0,0 +1,10 @@
+const { createGzip } = require('zlib');
+// const { createServer } = require('http');
```
styfle: Can we remove the comment?

cb1kenobi: Done!

packages/cli/test/dev/fixtures/43-compress-encoding/vercel.json
cb1kenobi: Done!

styfle (approved):
Great, thanks!

PR #10293, opened 2023-08-02
Update `ts-morph` to v19.0.0

TooTallNate:
Will be using `ts-morph` for some upcoming Hydrogen v2 transformations, so figured we should update to latest version.


TooTallNate:
ncc failing to build strikes again <picture data-single-emoji=":cry-a-lot:" title=":cry-a-lot:">:cry-a-lot:</picture>

PR #10297, opened 2023-08-03
Update discussion templates

amyegan:
The current note on discussion templates is not bold enough and easily overlooked.

In an effort to reduce off-topic discussion clutter in this repo and direct folks to the general support community which I monitor more closely, this update makes the note more prominent.


cb1kenobi (dismissed):
LGTM

styfle (approved):

PR #10303, opened 2023-08-04
Handle calls for deployment aliases when mocking deployment API

trek:
Attempting to add tests to unblock a contributor https://github.com/vercel/vercel/pull/10236#issuecomment-1664183891 and finding a deployment also checks its aliases. This PR updates the mock deployment code to also create a handler for its aliases.


dan-stowell (dismissed):

packages/cli/test/mocks/deployment.ts
```diff
@@ -170,6 +170,16 @@ function setupDeploymentEndpoints(): void {
     res.json({ builds });
   });
 
+  client.scenario.get('/now/deployments/:id/aliases', (req, res) => {
```
cb1kenobi: We should probably treat the first path segment as a version as described in the docs: https://vercel.com/docs/rest-api/endpoints#list-deployment-aliases

```suggestion
  client.scenario.get('/:version/deployments/:id/aliases', (req, res) => {
```

packages/cli/test/mocks/deployment.ts
```diff
@@ -170,6 +170,16 @@ function setupDeploymentEndpoints(): void {
...
+    const limit = parseInt(req.query.limit);
+    const { id } = req.params;
+    const deployment = deployments.get(id);
+    res.json({
+      aliases: deployment?.alias ? [deployment.alias] : [],
```
cb1kenobi: I believe `deployment.alias` is a string. The `aliases` return value is an array of objects:

```
{
  uid: string;
  alias: string;
  created: string;
  redirect?: string | null;
  protectionBypass?: {
    [key: string]:
      | { createdAt: number; createdBy: string; scope: 'sharable-link' | 'automation-bypass'; }
      | { createdAt: number; lastUpdatedAt: number; lastUpdatedBy: string; access: 'requested' | 'granted'; scope: 'user'; }
  }
}[]
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #10304, opened 2023-08-04
Add basic tests for remove command

trek:
Adds some minimal unit testing for the `remove` command. There is currently none.


.changeset/tough-forks-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Handle requests for deployment aliases when mocking deployments
```
styfle:
```suggestion
---
---

```

Since this is changing tests only, it doesn't require any version change to semver. 

You can do this with an empty changeset.

cb1kenobi (approved):

PR #10305, opened 2023-08-05
[remix] Add initial support for Hydrogen v2

TooTallNate:
Enables support for Hydrogen v2 apps using the "Remix" preset. This initial support works with the Hydrogen demo store template unmodified, and all pages will use Edge functions.

Node.js runtime, and also any other configuration via `export const config`, are not supported at this time, due to some pending blockers.


packages/remix/src/hydrogen.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ *  2. Track usages of the `env` parameter which (which gets removed),
+ *     so that we can create that object based on `process.env`.
+ */
+export function patchHydrogenServer(
+  project: Project,
+  serverEntryPoint: string
+) {
+  const sourceFile = project.addSourceFileAtPath(serverEntryPoint);
+  const defaultExportSymbol = sourceFile.getDefaultExportSymbol();
```
cb1kenobi: In my tests, `defaultExportSymbol` is `undefined` but my `server.js` has:

```
export default {
  async fetch(request, env, executionContext) {
```

Is that to be expected?

TooTallNate: Nice catch. Should be fixed by https://github.com/vercel/vercel/pull/10305/commits/a68723dd49d8b16a5fefb95353bfdfeb377c8bb9. I'll follow-up and add an E2E test as well.

cb1kenobi (dismissed):
Once I figured out that you need to set the `SESSION_SECRET` and `PUBLIC_STORE_DOMAIN` env vars, things worked great!

cb1kenobi (approved):

packages/remix/src/build.ts
```diff
@@ -349,11 +414,21 @@ module.exports = config;`;
...
     ),
-    ensureResolvable(entrypointFsDirname, repoRootPath, '@remix-run/node'),
+    !isHydrogen2
+      ? ensureResolvable(entrypointFsDirname, repoRootPath, '@remix-run/node')
+      : null,
   ]);
 
+  const staticDir = join(
+    remixConfig.assetsBuildDirectory,
+    ...remixConfig.publicPath
```
joelvh, line 424: @TooTallNate It looks like this change you made breaks Vercel builds using `publicPath`.

I've been troubleshooting why a deployment on Vercel with the `publicPath` setting is deploying files outside of my project root, and I think this is the reason -- assuming this influences the build process of my Remix app on Vercel.

The logic you introduced assumes `publicPath` matches the source path during the build, but it's meant to be a URL or path prepended to asset URLs in the browser. This can be the URL to a CDN or any other arbitrary path.

I have a Shopify app built with Remix that loads behind their "app proxy" (reverse proxy), which is a path on the storefront of a shop (e.g. `/apps/my-app`). I use `publicPath` to specify the prefix I need prepended to asset URLs to load them through that path.

This logic converts the path to `../..` and grabs unrelated files 2 folders up and puts them in the build output.

I'd be happy to contribute a PR, but it's not clear what the use case is you're addressing with this change. Please let me know if you have a workaround I can use or if there's a change I can help contribute.

Thanks you!

cb1kenobi: Hi @joelvh!

Just so I understand this correctly, `publicPath` is a "URL prefix" that applies to resolving assets from the browser at runtime. It is not meant to be a filesystem related path. By using the `publicPath` to resolve a filesystem path, we are using it to resolve files outside the project directory that most certainly should not be deployed. Long story short, we should not use the `publicPath`, right?

For the use case, the `publicPath` is being used to find all static files so they can be uploaded. I'm not super close to the Remix builder code, but perhaps there's another parameter we can use to resolve where static files are located?

joelvh: Hi @cb1kenobi!

That's correct -- it's a browser feature at runtime, unrelated to filesystem. In our case, the Vercel project root was in `packages/my-app` and it put all package source files in the output (client and server code).

I put together a PR (#10666) to aim at solving both use cases, but I don't think it's bullet proof. I also recognize the challenge to make this a zero-config feature.

Is the `assetsBuildDirectory` not the path you want? Are you assuming that points to `public/build` and you're trying to get to the root of `public` where you assume the rest of the static assets are?

joelvh: I should clarify: if you traverse up the path specified in `assetsBuildDirectory` you might get closer to what you want. But let me know what you think of the solution I suggested as a potential stopgap.

PR #10311, opened 2023-08-07
[cli] Sanitize argv for `vc build`

cb1kenobi:
`vc build` writes a file `.vercel/output/build.json` that contains the `argv` contents which could include sensitive information that needs to be sanitized.


packages/cli/src/util/build/scrub-argv.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Redacts `--env`, `--build-env`, and `--token` in `argv`.
+ */
+export function scrubArgv(argv: string[]) {
```
TooTallNate, line 4: Let's re-use this function in `report-error.ts`?

packages/cli/src/util/build/scrub-argv.ts
```diff
@@ -0,0 +1,18 @@
...
+  const clonedArgv = [...argv];
+  const tokenRE = /^(-[A-Za-z]*[bet]|--(?:build-env|env|token))(=[^\w]*)?/;
```
styfle: I'm confused why this is using a regex here. I thought we already parsed arvg using [`arg`](https://www.npmjs.com/package/arg) package. Could we reuse the parsed result instead?

packages/cli/src/util/build/scrub-argv.ts
cb1kenobi: That code mutates `argv` post-parsing where the `build.json` contains `process.argv`. In other words, `report-error` code won't work without modifications at which point we're probably back to where we started.

packages/cli/src/util/build/scrub-argv.ts
cb1kenobi: The original code was logging `process.argv`. My intent was to preserve this and not output the parsed `args`.

packages/cli/test/unit/util/build/scrub-argv.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    it('should handle -t without a value', () => {
+      const result = scrubArgv(['foo', '-t']);
+      expect(result).toEqual(['foo', '-t']);
+    });
+
+    it('should scrub -t from grouped short flags', () => {
+      const result = scrubArgv(['foo', '-xyt', 'bar', 'baz']);
+      expect(result).toEqual(['foo', '-xyt', 'REDACTED', 'baz']);
+    });
```
styfle, line 112: Lets add a test with more combinations of options such as

```
vc build --cwd /path/to/dir --env "NODE_ENV=production"  --token "$TOKEN" --prod --yes
```

packages/cli/src/util/build/scrub-argv.ts
trek: Although if we moved to using `client.args` and stringified it, it would only print any parsed args. `--token` is a no-op for build, so would just get dropped.

packages/cli/src/util/build/scrub-argv.ts
TooTallNate: I think we could just do `args(scrubArgv(process.env))` and it would be the same thing.

cb1kenobi: @TooTallNate does the build container do anything with `argv` from the `build.json` or is it just for debugging?

TooTallNate: Build container doesn't do anything with it.

packages/cli/test/unit/util/build/scrub-argv.test.ts
cb1kenobi: Done!

trek (approved):

PR #10313, opened 2023-08-07
`edge-light` condition interoperability with `vercel dev`

PonomareVlad:
1. [Ref](https://github.com/vercel/vercel/blob/2995781a583a9c6df15f6447667ff7adaa0d3bfd/packages/node/src/index.ts#L179)
2. [Spec](https://runtime-keys.proposal.wintercg.org/#edge-light)
3. [Test](https://github.com/PonomareVlad/VercelEdgeExportsTest)


EndangeredMassa:
@PonomareVlad can you explain why this is needed and how it works?

I did find https://esbuild.github.io/api/#conditions, which was helpful.

PonomareVlad:
@EndangeredMassa You can test [this repo](https://github.com/PonomareVlad/VercelEdgeExportsTest).

If we deploy it to Vercel, then we see that `edge-light` target is used for modules import.

But, if we run `vercel dev` then we see using `browser` target.

This is inconsistently, and PR should fix that.

Also `development` target improve stack traces readability for dependencies source without minification (if they have that). 🌚

styfle (dismissed):
Looks correct since prod deploys check for `edge-light` since https://github.com/vercel/vercel/pull/9700

We might need to add more conditions here for dev.

cc @TooTallNate

cb1kenobi:
Tests pass: https://github.com/vercel/vercel/pull/10332

cb1kenobi (approved):

cb1kenobi:
Thank you @PonomareVlad!

PR #10316, opened 2023-08-08
Update types to match mocks/reality

trek:
Not 100% convinced this is the right path forward, but also want to avoid a situation where all of a sudden I have the car's entire engine disassembled in the shop.

A contributor is having trouble writing tests for https://github.com/vercel/vercel/pull/10236 because on `main` our mocks for deployments, the cli's internal `type Deployment`, the data returned from the API, and the data we've _mocked_ being returned from the API have all drifted slightly from each other.

In this particular case `vercel remove my-project-name --safe` does a bit of chatter back and forth with the API to correctly identify what the user would like to remove. For `remove` uses where a user provides a project name there is:

1. [Check the API for deployments or projects](https://github.com/vercel/vercel/blob/2995781a583a9c6df15f6447667ff7adaa0d3bfd/packages/cli/src/commands/remove/index.ts#L99-L112) by the provided argument (`my-project-name`)
2. [If a project is found, get all its deployments](https://github.com/vercel/vercel/blob/2995781a583a9c6df15f6447667ff7adaa0d3bfd/packages/cli/src/commands/remove/index.ts#L120-L127)
3. Finally, call the API to delete.

IRL the response format for step `2` is:

```
    {
      uid: 'dpl_somedeployid',
      name: 'my-project-name',
      url: 'a-url.vercel.app',
      created: 1690810409511,
      source: 'cli',
      state: 'READY',
      readySubstate: 'PROMOTED',
      type: 'LAMBDAS',
      creator: {
        uid: 'some-random-uid',
        email: 'trek.glowacki@example.com',
        username: 'trek'
      },
      inspectorUrl: 'https://vercel.com/username/projectname/somedeployid',
      meta: {},
      target: 'production',
      aliasError: null,
      aliasAssigned: 1690810413263,
      isRollbackCandidate: true,
      createdAt: 1690810409511,
      buildingAt: 1690810410569,
      ready: 1690810412871,
      projectSettings: { commandForIgnoringBuildStep: null }
    }
```

Under tests, step `2` is set to return mocked POJOs of `type Deployment` which _doesn't_ currently have `uid` (you set `id`) or `state` (you set `status`).

The result is that, under test, you get a "reply" with a structure like

```
{
  id: 'dpl_somedeployid',
  ... snip ...
  status: 'READY',
}
```

And when the URL to delete a deployment is generated it reaches for `deployment.uid` which works fine IRL but will results in `'/now/deployments/undefined'` in tests.

This prevents the contributor in #10236 from being able to provide a test case.

I'm not familiar enough with the codebase to know if there's a better, more wholistic solution or not, so putting this up for conversation.


styfle (changes requested):
I don't think this is the right solution.

Instead, we need different types for different versions of the API.

For example, `/v6/deployments` returns `uid` but `/v13/deployments` API returns `id`.

I think you are combining two API responses into a single type which is going to lead to more confusion.

https://vercel.com/docs/rest-api/endpoints#list-deployments

https://vercel.com/docs/rest-api/endpoints#get-a-deployment-by-id-or-url

trek:
@styfle yes, that's the "car's entire engine disassembled in the shop" situation I was hoping to avoid.

I don't think the underlying issue is the shape of the mock itself, I think it's:

1. We use a single type for both the object internally _and_ mocks.
2. We [return objects of this shape regardless of API version](https://github.com/vercel/vercel/blame/main/packages/cli/test/mocks/deployment.ts#L197)
3. Mocks don't pass through a serialization step to match actual API endpoint reply formats.

A solution like "create a mock specific to the API that will get called" feels brittle and complicated for TDD since you'd need to know the code that will be called before even starting a test. We've kind of already been bitten by this when the tests I added for `remove deployment.url` just don't work for the related sibling test of `remove project.name`.

I'd propose 

1. Splitting the types for our internal use of `Deployment` from `MockDeployment` (which are supposed to represent a remote resource anyway, not our local objects) and doing this for all mocks.
2. Making test URLs more explicit (which ironically countermands the advice I got around #10300) 
3. Adding a specific serialization step to each mocked endpoint to convert the mock object into an accurate representation of the response. This might be as simple as `Object.assign`

But that's a fair bit of refactoring and I have no idea if that's valuable to the team or not.

cb1kenobi:
I also don't think this is correct. For example, API will not return a `Deployment` with a `uid`. We should close this, audit our API calls, ensure we're calling the latest versions, ensure our types match the response, and ensure our mocks match the response.

EndangeredMassa:
Ideally, I don't think we'd use API response types throughout. The response from an API does not have to (and often should not) match the concept that we use internally to the program. This pattern protects the operations of the program from the changes of API responses by reducing the transition across concepts to a single point.

Pragmatically, shifting the entire codebase to that model is a large effort and not likely worth doing all at once. A migration towards that seems hard to plan as well.

I'd be fine with the current model. And audit at least is a good starting point, but migrating to latest versions of API calls may take some time.

I think we should only do one codebase migration at a time. We're currently migrating command definitions. Maybe next we can migrate API calls and corresponding types.

cb1kenobi:
The API responses are apart of the public contract. It can't change without a version rev. The CLI's types should match the shape of the data returned by the API server.

+1 on migrating one API at a time.

EndangeredMassa:
The current mock is wrong and tests cannot be written against it that test the wrong parts. I think we can discuss how to move the codebase to one thing or another separately from fixing the current issue.

The PR as written is a fine step towards a better solution, in my opinion.

PR #10319, opened 2023-08-09
[examples] Add "hydrogen-2" template

TooTallNate:
Adds a Hydrogen v2 template which is the output of the `npm create @shopify/hydrogen@latest` command.

Note that a `vercel.json` file is being used to define the environment variables that are required at runtime. This is required for the template to deploy with zero configuration, however the user should update these values (including replacing the session secret) and migrate them to the Project settings in the Vercel dashboard.

[Live example](https://hydrogen-v2-template.vercel.app)


examples/hydrogen-2/vercel.json
```diff
@@ -0,0 +1,6 @@
+{
+  "env": {
+    "SESSION_SECRET": "foobar",
```
leerob, line 3: Should these be in the `vercel.json`?

TooTallNate: See my note in the PR description. Ideally we'd prompt for the user to enter these when importing the template on vercel.com/new, but I don't think we currently have a mechanism to do so from within these framework templates.

leerob (dismissed):

leerob (approved):
🎉

PR #10320, opened 2023-08-09
[frameworks] Add "(v1)" suffix to "hydrogen" preset

TooTallNate:
Now that the "remix" preset should be used for Hydrogen v2, it's not very clear that the "hydrogen" preset is only meant to be used for Hydrogen v1.

So add clarification in the "name" of the "hydrogen" preset to help remedy that.


leerob (approved):

cb1kenobi (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -1691,7 +1691,7 @@ export const frameworks = [
     defaultVersion: '0.13.0', // Must match the build image
   },
   {
-    name: 'Hydrogen',
+    name: 'Hydrogen (v1)',
```
styfle, line 1694: @TooTallNate This doesn't follow the naming convention of others.

For example, we have `Docusaurus 1` / `Docusaurus 2`

PR #10323, opened 2023-08-10
[cli] tidy up use of package name and logo

trek:
I erred on the side of preferring the more recently added versions of this behavior. I assume at some point in the distant past the removed function correctly toggled between some now/zeit/vercel transitioning.


TooTallNate (approved):

cb1kenobi (approved):
Awesome

styfle:
> I assume at some point in the distant past the removed function correctly toggled between some now/zeit/vercel transitioning

Indeed. And it looks like your PR will still work since its still using package.json as the source of truth 👍

PR #10329, opened 2023-08-11
allow Vercel to be run while offline

hassanbazzi:
# Summary

Right now, you need to be connected to the internet in order to use the Vercel dev command, although everything is spun locally. This change will simply skip the constant linking and allow users to work offline as long as they have setup the project before.

### DIFF
--- a/.changeset/chilly-clouds-relate.md
+++ b/.changeset/chilly-clouds-relate.md
@@ -1,0 +1,6 @@
+---
+'@vercel-internals/types': patch
+'vercel': patch
+---
+
+Allow dev command to be ran while offline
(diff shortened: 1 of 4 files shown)


EndangeredMassa:
@hassanbazzi thanks for the contribution! We've wanted to do this for some time, but there are other impacts to the overall local development workflows that we need to consider.

We'll get back to you about this after next week.

EndangeredMassa:
@hassanbazzi we discussed this and decided that we need to solve this in a more complete way. We do plan to do this, but have no ETA.

Thanks for the contribution, but we're going to close this for now.

PR #10333, opened 2023-08-11
[cli] Swap custom table code with `cli-table3`

trek:
Replaces use of hand-rolled table code with `cli-table3` in help output. Looking over the `cli` package we have several bits of custom terminal table-like code being used. Some are completely custom (like the two column implementation this PR replaces) and some rely on other existing terminal table packages (`text-table`) which lack word wrap/line wrap abilities of our custom solutions.

In migrating `--help` sections specifically I need a three column implementation. Rather than update our custom table function to support `n` columns and wrapping or update our `text-table` use to support wrapping, we discussed adopting a terminal table library that supports all these features.

The main two under consideration were  [`table`](https://www.npmjs.com/package/table) and [`cli-table3`](https://www.npmjs.com/package/cli-table3). @cb1kenobi has prior experience with `cli-table3` so we opted to defer to that.

As an added fun coincidence, @rauchg was an original developer.

Followup PRs will:

1. Use three column tables for outputting subcommands as part of `--help`
2. Update our use of `text-table` to uses of `cli-table3` and remove the `text-table` dependency.


packages/cli/package.json
```diff
@@ -40,7 +40,8 @@
     "@vercel/redwood": "1.1.15",
     "@vercel/remix-builder": "1.10.0",
     "@vercel/ruby": "1.3.76",
-    "@vercel/static-build": "1.3.45"
+    "@vercel/static-build": "1.3.45",
+    "cli-table3": "0.6.3"
```
styfle: Please move this to `devDependencies` so that `ncc` will bundle it.

packages/cli/src/commands/help.ts
```diff
@@ -166,97 +168,89 @@ export function buildCommandOptionLines(
...
   // Sort command options alphabetically
   commandOptions.sort((a, b) =>
     a.name < b.name ? -1 : a.name > b.name ? 1 : 0
   );
-  // Keep track of longest "start" of an option line to determine description spacing
-  let maxLineStartLength = 0;
-  // Iterate over options and create the "start" of each option (e.g. `  -b, --build-env <key=value>`)
+
+  // word wrapping requires the wrapped cell to have a fixed width.
+  // We need to track cell sizes to make the final column of cells is
```
EndangeredMassa, line 186:
```suggestion
  // We need to track cell sizes to make sure the final column of cells is
```

packages/cli/src/commands/help.ts
```diff
@@ -166,97 +168,89 @@ export function buildCommandOptionLines(
...
+  // equal to  the remaindner of unused horizontal space.
```
EndangeredMassa, line 187:
```suggestion
  // equal to the remainder of unused horizontal space.
```

packages/cli/src/commands/help.ts
```diff
@@ -166,97 +168,89 @@ export function buildCommandOptionLines(
...
+      'bottom-left': '',
+      'bottom-right': '',
+      left: '',
+      'left-mid': '',
+      mid: '',
+      'mid-mid': '',
+      right: '',
+      'right-mid': '',
+      middle: '',
+    },
```
EndangeredMassa, line 237: **question (non-blocking):** ‏we have to specify all of these to remove the lines?

EndangeredMassa (approved):
I tested this out locally. The output looks good from about 50+ columns, which sounds reasonable to me.

packages/cli/src/commands/help.ts
trek: Yes. Not a fan but there are defaults applied. This is how to unset them :-/

cb1kenobi (approved):
Looks great! Option descriptions wrap better than the command description.
image

packages/cli/src/commands/help.ts
cb1kenobi: Sadly, yeah. The defaults add padding to every cell.

trek:
@cb1kenobi we could put it in a one column table 😛 . I kid. I did pull in `word-wrap` in a demo branch but wanted to do one dep at a time. We could also use the word wrapping util exported here https://github.com/cli-table/cli-table3/blob/149205c6184fc8946c5fcc25d037b511b7d665ce/src/utils.js#L333 but I'm guessing that's not intended to be public API and just exported so they can import in tests.

PR #10334, opened 2023-08-11
Connecting Datadog / GH Actions / Turbo / Jest to detect flakey tests

trek:
Update our Github Actions test workflow to report on flakey tests. This should helps us identify and hopefully eliminate or mitigate test variance. Not the biggest fan of updating every `test` run to include the `--reporter` args, but I couldn't find a way to reliably hide this away somewhere else. If anyone knows a way that would be great.

There were two general data gathering techniques. Tracing (which nextjs is doing) and junit.xml generation (used elsewhere at Vercel). I opted for the latter because it was a bit simpler.

Additional comment inline.


.github/workflows/test.yml
```diff
@@ -88,6 +86,7 @@ jobs:
         run: node utils/gen.js && node_modules/.bin/turbo run test --cache-dir=".turbo" --log-order=stream --scope=${{matrix.packageName}} --no-deps -- ${{ join(matrix.testPaths, ' ') }}
         shell: bash
         env:
+          JEST_JUNIT_OUTPUT_FILE: ${{github.workspace}}/.junit-reports/${{matrix.scriptName}}-${{matrix.packageName}}-${{matrix.chunkNumber}}-${{ matrix.runner }}.xml
```
trek, line 89: Under our current test sharding strategy we technically would only generate a single `junit.xml` file per run/container. But, I wanted to bake in a bit of safety in case that changes in the future and we end up with multiple files per run.

.github/workflows/test.yml
```diff
@@ -96,6 +95,15 @@ jobs:
       - name: fetch ssl certificate after tests (linux, os x)
         if: matrix.runner != 'windows-latest'
         run: echo | openssl s_client -showcerts -servername 'api.vercel.com' -connect 76.76.21.21:443
+      - name: install datadog-ci
```
trek: Datadog also has a [GitHub Action](https://github.com/DataDog/junit-upload-github-action) for this but it's just basically just taking the DD cli interface and exposing that through specific yaml keys. IMO, not worth adding.

.github/workflows/test.yml
```diff
@@ -96,6 +95,15 @@ jobs:
...
+        if: always()
+        run: npm i -g @datadog/datadog-ci@2.18.1
+      - name: 'Upload Test Report to Datadog'
+        if: always()
+        run: 'datadog-ci junit upload --service vercel-cli .junit-reports'
```
styfle:
```suggestion
        run: 'npx @datadog/datadog-ci@2.18.1 junit upload --service vercel-cli .junit-reports'
```

I think you can simplify this to one line

.github/workflows/test.yml
trek: Technically we could probably lump in the install as part of global installs. This might be marginally faster but didn't appear to make to much of a difference so I kept it in its own section.

https://github.com/vercel/vercel/blob/9db3794735481881289c560baef9702c1710c70c/.github/workflows/test.yml#L74

EndangeredMassa (approved):

.github/workflows/test.yml
```diff
@@ -97,6 +96,13 @@ jobs:
         if: matrix.runner != 'windows-latest'
         run: echo | openssl s_client -showcerts -servername 'api.vercel.com' -connect 76.76.21.21:443
 
+      - name: 'Upload Test Report to Datadog'
+        if: always()
+        run: 'npx @datadog/datadog-ci@2.18.1 junit upload --service vercel-cli .junit-reports'
```
trek, line 101: I picked `vercel-cli` as a "service" name in Datadog because "vercel" seems too broad and I didn't have a good term for what all this repo encompass. Suggestions welcome.

cb1kenobi (approved):
Looks good! Thanks!

examples/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "test-unit": "pnpm test __tests__/unit/",
     "test-e2e": "pnpm test __tests__/integration/",
-    "test": "jest --env node --verbose --runInBand --bail"
+    "test": "jest --reporters=default --reporters=jest-junit --env node --verbose --runInBand --bail"
```
styfle, line 8: Can we move these to the jest config file instead of command line flags?

trek: This was what I [tried initially](https://github.com/vercel/vercel/pull/10334/commits/c9a6dfec3328e675a9a0ed256d2766367ec72666#diff-ac4313523217deffcd0ebf639a98c806b787a42b7d6440853ce9149a06a43cc5) but these files weren't being sourced in CI. Probably why all the args are inlined.

PR #10337, opened 2023-08-14
[cli] Respect `--yes` flag for all prompts during `vc link --repo`

TooTallNate:
A couple of the prompts during `vc link --repo` were not honoring the `--yes` flag.


trek (approved):

packages/cli/src/util/link/repo.ts
```diff
@@ -124,22 +131,29 @@ export async function ensureRepoLink(
     if (!remoteUrls) {
       throw new Error('Could not determine Git remote URLs');
     }
-    const remoteNames = Object.keys(remoteUrls);
+    const remoteNames = Object.keys(remoteUrls).sort();
```
styfle, line 134: Why sort? Is this handling the case when you have `origin` and `original` or something like that?

It would be great to add a test.

cb1kenobi: I tinkered with this and I think it's fine. If there's a `'origin'` remote, then it forces the remote to be `'origin'`. The sort order doesn't matter.

Interestingly enough, apparently the default `.sort()` will sort capitalized words before lowercase words:

```
> ['a', 'B'].sort()
[ 'B', 'a' ]
```

cb1kenobi (approved):
LGTM

packages/cli/src/util/link/repo.ts
styfle: The default sort is basically comparing the unicode values of the chars.

You could handle cases by using `localeCompare()`.

```js
['a', 'B'].sort((a,b) => a.localeCompare(b)) // ['a', 'B']
```

cb1kenobi: I guess I've been using `localeCompare()` for so many years, I rarely do a `.sort()` without it.

PR #10345, opened 2023-08-15
[frameworks][fs-detectors] Add "supersedes" prop to Framework interface

TooTallNate:
The `supersedes` property on the `Framework` interface may be specified by a framework in which the superseded framework should _not_ be included in the resulting framework matches during detection.

For example, a Hydrogen v2 project matches both the "remix" and "hydrogen" framework detectors, but we really only want "remix" to be selected.


.changeset/odd-kangaroos-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/frameworks': minor
+'@vercel/fs-detectors': patch
+---
+
+Add "supercedes" prop to Framework interface
```
trek: Is the use of the archaic spelling form intended? I love it, if so.

trek (dismissed):

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -135,6 +135,31 @@ async function matches(
...
+function removeSupercededFrameworks(
+  matches: (Pick<Framework, 'supercedes' | 'slug'> | null)[]
+) {
+  for (const match of matches.slice()) {
+    if (!match) continue;
+    if (match.supercedes) {
+      removeSupercededFramework(matches, match.supercedes);
+    }
+  }
+}
```
EndangeredMassa, line 160: **suggestion (soft-blocking):** add unit tests

In this or a follow-up PR: Can we add some unit tests around this behavior?‏

It took me some careful reading to understand exactly what was happening here. This logic also supports chained superseding, which isn't currently tested.

EndangeredMassa (dismissed):

cb1kenobi (dismissed):
Tested, looks good!

.changeset/odd-kangaroos-fix.md
TooTallNate: Actually I'm just dumb. Will get it "modernized".

packages/fs-detectors/src/detect-framework.ts
TooTallNate: There's updated unit tests already. I'll add one that tests the recursiveness though.

EndangeredMassa: The tests updated in this PR are testing full framework detection. I meant to suggest specific unit tests for these functions directly. This would make it easier to enumerate different cases.

.changeset/odd-kangaroos-fix.md
trek: Reject modernity, embrace tradition!

styfle: Idk, I think `supercedes` has been superseded `supersedes` so lets go with that. Especially since this ends up on the public API so we can't change it later.

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -135,6 +135,31 @@ async function matches(
   };
 }
 
+function removeSupercededFramework(
+  matches: (Pick<Framework, 'supercedes' | 'slug'> | null)[],
+  slug: string
+) {
+  const index = matches.findIndex(f => f?.slug === slug);
+  if (index !== -1) {
+    const framework = matches[index]!;
```
styfle: If you hoist this out of the `if`, you might be able to remove the `!` by checking for truthyness in the `if` statement

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -135,6 +135,31 @@ async function matches(
...
+      removeSupercededFramework(matches, framework.supercedes);
+    }
+    matches.splice(index, 1);
+  }
+}
+
+function removeSupercededFrameworks(
+  matches: (Pick<Framework, 'supercedes' | 'slug'> | null)[]
+) {
+  for (const match of matches.slice()) {
```
styfle, line 155: Can you explain the `matches.slice()`. It seems like its meant to clone the array but the clone isn't used at all.

styfle: Oh I think I'm confused because I thought `removeSupercededFrameworks()` was calling itself but the function call below is missing the `s` and therefore a different function <picture data-single-emoji=":doh:" title=":doh:">:doh:</picture>

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -135,6 +135,31 @@ async function matches(
...
+    if (!match) continue;
+    if (match.supercedes) {
```
styfle: Combine these into a single line

```suggestion
    if (match?.supercedes) {
```

packages/fs-detectors/src/detect-framework.ts
```diff
@@ -135,6 +135,30 @@ async function matches(
   };
 }
 
+function removeSupercededFramework(
```
cb1kenobi: Spelling?

```suggestion
function removeSupersededFramework(
```

cb1kenobi (changes requested):

cb1kenobi (approved):

packages/fs-detectors/src/detect-framework.ts
EndangeredMassa: @cb1kenobi I don't think spelling is worth blocking the PR. I do think it should have been fixed, but could have been a follow-up PR.

PR #10346, opened 2023-08-15
[tests] Fix failures due to Node 14.x being discontinued

TooTallNate:
This PR upgrades test fixtures from Node.js 14 to 16 and also removes the `35-puppeteer` test fixture since it doesn't support Node.js 16 unfortunately. 

- See https://github.com/alixaxel/chrome-aws-lambda/pull/274


styfle (dismissed):

styfle:
I think `35-puppeteer` has to update dependencies

styfle (approved):
Looks like we'll have to delete it since its no longer supported 🫡

- https://github.com/alixaxel/chrome-aws-lambda/pull/274

cb1kenobi (approved):

PR #10354, opened 2023-08-16
[frameworks] Updated default Hugo command to not publish draft pages

cb1kenobi:
We should not publish Hugo draft pages during deployments.

Linear: https://linear.app/vercel/issue/VCCLI-319/create-hugo-migration-to-set-default-build-command


TooTallNate:
Dup of https://github.com/vercel/vercel/pull/7326.

PR #10363, opened 2023-08-17
prepare spawnAsync and BuildResultV2Typical for saving buildRuntimeVersion on deployment doc

EndangeredMassa:


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -90,19 +90,30 @@ export function spawnAsync(
...
+    const stdoutLogs: Buffer[] = [];
+
     opts = { stdio: 'inherit', ...opts };
     const child = spawn(command, args, opts);
 
     if (opts.stdio === 'pipe' && child.stderr) {
       child.stderr.on('data', data => stderrLogs.push(data));
     }
 
+    // TODO: Do we need to check for "pipe" here and for stderr?
```
EndangeredMassa, line 104: Do we need to guard on `stdio` being "pipe" here? It seems like a weird restriction, but I'll leave it if we need it.

PR #10369, opened 2023-08-18
BREAKING CHANGE: Drop Node.js 14, bump minimum to Node.js 16

styfle:
https://vercel.com/changelog/node-js-14-and-16-are-being-deprecated

https://github.com/nodejs/Release/blob/main/README.md#end-of-life-releases


cb1kenobi (dismissed):
LGTM

trek (approved):

EndangeredMassa (approved):

packages/cli/src/util/dev/router.ts
```diff
@@ -58,7 +58,7 @@ export async function devRouter(
 ): Promise<RouteResult> {
   let result: RouteResult | undefined;
   let { pathname: reqPathname, search: reqSearch } = url.parse(reqUrl);
-  reqPathname ??= '/';
+  reqPathname = reqPathname || '/';
```
Ethan-Arrowood, line 61: I don't understand this change. How did we regress in the syntax we support?

styfle: @Ethan-Arrowood  Previously, TS was downleveling and then ncc would bundle it.

Since TS is no longer downleveling, ncc is choking because its very outdated https://github.com/vercel/vercel/blob/dde2b9b34f262addc89eea599b3bab047e28f4d7/packages/cli/package.json#L95

In a future PR, we should update ncc or switch to something else for bundling.

Kikobeats (approved):

PR #10370, opened 2023-08-18
[cli] Wrap description in help output

trek:
If we wanted to avoid pulling in another dependency (`word-wrap`), this is a too-clever-by-half way to get word wrapping on the `--help` description text.

Before:
CleanShot 2023-08-18 at 10 11 14@2x

After:
CleanShot 2023-08-18 at 10 08 52@2x


packages/cli/src/commands/help.ts
```diff
@@ -277,9 +282,25 @@ export function buildCommandExampleLines(command: Command) {
...
+      'padding-left': INDENT.length,
+    },
+  });
+  const table = new Table(_tableOptions);
+  table.push([
+    {
+      content: command.description,
+      wordWrap: true,
+    } as _CellOptions,
+  ]);
```
EndangeredMassa: **suggestion (soft-blocking):** extract function

Can we extract this into a named function to make it super clear that we're using this table module purely for word wrapping?‏

cb1kenobi (dismissed):
LGTM

packages/cli/src/commands/help.ts
EndangeredMassa: **alternate suggestion (non-blocking):** use `word-wrap`

I don't think the module [word-wrap](https://github.com/jonschlinkert/word-wrap) is [too big](https://packagephobia.com/result?p=word-wrap) to bring in. I'd be fine with that.

EndangeredMassa (dismissed):
I don't love using a table module for this, but if we don't want another dependency, it does work.

packages/cli/src/commands/help.ts
cb1kenobi: If we use `word-wrap` will the wrapped description continue to be indented on subsequent lines?

trek: @cb1kenobi yessir.

Ethan-Arrowood (dismissed):

packages/cli/src/commands/help.ts
trek: Tested out `word-wrap` and hilariously it does a poorer job unless you manually add a bit of right padding. It seems to struggle with word boundaries:

CleanShot 2023-08-18 at 13 34 56@2x

I do like Sean's suggestion of extracting into a function.

cb1kenobi (approved):
Tested and looks good!

PR #10372, opened 2023-08-18
[cli] Update new `--help` structure to support subcommands

trek:
Before:
CleanShot 2023-08-21 at 11 07 25@2x

After:
CleanShot 2023-08-21 at 11 13 21@2x

Some notes re: "Before"

* We've dropped wrapping required args in `<>` some time ago.
* The command name for alias is optional (default is `set`) but was incorrectly still listed as required.


EndangeredMassa (approved):
As a migration of how it currently works, this is good. The actual help output for `vc help alias set` (for example) could use some updates. We can do that another time, though.

PR #10377, opened 2023-08-22
Migrate `certs` to new command structure

trek:
Before:
CleanShot 2023-08-22 at 09 31 19@2x

After:
CleanShot 2023-08-22 at 09 32 42@2x

### DIFF
--- a/packages/cli/src/commands/certs/command.ts
+++ b/packages/cli/src/commands/certs/command.ts
@@ -1,0 +1,116 @@
+import { Command } from '../help';
+import { packageName } from '../../util/pkg-name';
+
+export const certsCommand: Command = {
+  name: 'certs',
+  description:
+    'Interact with SSL certificates. This command is intended for advanced use only. By default, Vercel manages your certificates automatically.',
+  arguments: [
+    {
+      name: 'command',
+      required: false,
+    },
+  ],
+  subcommands: [
+    {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/certs/command.ts
```diff
@@ -0,0 +1,116 @@
...
+    {
+      name: 'ls',
+      description: 'Show all available certificates',
+      arguments: [],
+      options: [],
+      examples: [],
+    },
+    {
+      name: 'issue',
+      description: ' Issue a new certificate for a domain',
```
EndangeredMassa, line 24: **suggestion (soft-blocking):** 

```suggestion
      description: 'Issue a new certificate for a domain',
```

packages/cli/src/commands/certs/command.ts
```diff
@@ -0,0 +1,116 @@
...
+      arguments: [
+        {
+          name: 'cn',
```
EndangeredMassa, line 27: **question (soft-blocking):** ‏do we want to still show this accepts any number of domain arguments?

The example shows this. So does the old help output:

```
issue     <cn> [<cn>]
```

Side point: I don't know what `cn` is supposed to stand for. These are domains.

EndangeredMassa (approved):

PR #10378, opened 2023-08-22
[error-utils] fix `files` in package.json to use `dist`

styfle:
These packages were publishing unnecessary files to npm so we can reduce to only the `dist` directory.

- Example: https://unpkg.com/browse/@vercel/error-utils@2.0.0/
- Docs: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

### DIFF
--- a/internals/constants/package.json
+++ b/internals/constants/package.json
@@ -4,6 +4,9 @@
   "version": "1.0.4",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "tsc -p tsconfig.json"
   },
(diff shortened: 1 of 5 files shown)


trek (approved):

internals/constants/package.json
```diff
@@ -4,6 +4,9 @@
   "version": "1.0.4",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
+  "files": [
```
cb1kenobi, line 7: This is a private package. Why do we need to specify the `files`?

styfle: Its good practice. Depending on the package manager, it might change the behavior of workspaces so you don't have access to files outside of the allowlist. I haven't confirmed this though.

Also if we ever decide to publish it, then it will already be configured 😄

PR #10379, opened 2023-08-22
[cli] Update `dns` commands to new structure

trek:
Before:
CleanShot 2023-08-22 at 10 46 34@2x

After:
CleanShot 2023-08-22 at 10 48 04@2x


packages/cli/src/commands/dns/command.ts
```diff
@@ -0,0 +1,136 @@
...
+    },
+    {
+      name: 'Import a Zone file',
+      value: [
+        `${packageName} dns import <DOMAIN> <FILE>`,
+        `${packageName} dns import zeit.rocks ./zonefile.txt`,
+      ],
+    },
+    {
+      name: 'Paginate results, where `1584722256178` is the time in milliseconds since the UNIX epoch.',
```
cb1kenobi, line 129: knit: it's cool if the example name ends with a period, but we should be consistent.

cb1kenobi (approved):
LGTM

packages/cli/src/commands/dns/command.ts
trek: Agreed, but I'm just migrating over what exists currently. Not sure why some have periods and others don't.

cb1kenobi (approved):

PR #10380, opened 2023-08-22
[frameworks] move some `devDependencies` to `dependencies`

EndangeredMassa:
Because `frameworks` isn't bundled with `ncc`, we need the runtime dependencies to exist in the `dependencies` section of `package.json`.

### DIFF
--- a/packages/frameworks/package.json
+++ b/packages/frameworks/package.json
@@ -19,15 +19,15 @@
   },
   "dependencies": {
     "@iarna/toml": "2.2.3",
-    "js-yaml": "3.13.1"
+    "js-yaml": "3.13.1",
+    "@vercel/error-utils": "2.0.0",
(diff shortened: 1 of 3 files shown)


packages/frameworks/package.json
```diff
@@ -19,15 +19,15 @@
   },
   "dependencies": {
     "@iarna/toml": "2.2.3",
-    "js-yaml": "3.13.1"
+    "js-yaml": "3.13.1",
+    "@vercel/error-utils": "2.0.0",
+    "@vercel/routing-utils": "3.0.0"
```
styfle: I'm not sure if `@vercel/routing-utils` is necessary since its just used for types

trek (approved):

styfle (approved):
Thanks!

cb1kenobi (approved):

PR #10387, opened 2023-08-23
[node] use `undici` instead of `node-fetch`

Kikobeats:


packages/node/src/dev-server.mts
```diff
@@ -6,13 +6,14 @@ if (!entrypoint) {
...
 import { join } from 'path';
-import type { Headers } from 'node-fetch';
+import type { Headers } from 'undici';
 import type { VercelProxyResponse } from './types.js';
 import { Config } from '@vercel/build-utils';
 import { createEdgeEventHandler } from './edge-functions/edge-handler.mjs';
 import { createServer, IncomingMessage, ServerResponse } from 'http';
 import { createServerlessEventHandler } from './serverless-functions/serverless-handler.mjs';
 import { isEdgeRuntime, logError, validateConfiguredRuntime } from './utils.js';
+import { toToReadable } from '@edge-runtime/node-utils';
```
styfle, line 16: Why is this called `toToReadable`?

Shouldn't it be called `toReadable` instead?

Kikobeats: the good new is... it can be dropped too!
https://github.com/vercel/edge-runtime/blob/bbce361d6f771415996ca15bb46790f34b5b0d74/packages/node-utils/src/edge-to-node/stream.ts#L11

Kikobeats: `stream.Readable.fromWeb` was added in v17 so it's missing in v16 <picture data-single-emoji=":oh_no:" title=":oh_no:">:oh_no:</picture>

packages/node/src/serverless-functions/serverless-handler.mts
```diff
@@ -116,16 +115,25 @@ export async function createServerlessEventHandler(
...
         body = compress(body, contentEncoding);
-        response.headers.set('content-length', Buffer.byteLength(body));
+        const clonedHeaders = Object.fromEntries(response.headers);
+        clonedHeaders['content-length'] = String(Buffer.byteLength(body));
+        /**
+         * `transfer-encoding` is related to streaming chunks.
+         * Since we are buffering the response.body, it should be stripped.
+         */
+        delete clonedHeaders['transfer-encoding'];
+        headers = new Headers(clonedHeaders);
```
styfle:
```suggestion
        const clonedHeaders = []
        for (const [key, value] of response.headers.entries()) {
          if (key !== 'transfer-encoding') {
            // transfer-encoding is only for streaming response
            clonedHeaders.push([key, value])
          }
        }
        clonedHeaders.push(['content-length', String(Buffer.byteLength(body))]);
        headers = new Headers(clonedHeaders);
```

packages/node/src/dev-server.mts
```diff
@@ -104,18 +105,20 @@ async function onDevRequest(
...
+          key,
+          key === 'set-cookie' ? headers.getSetCookie() : value
+        );
     }
 
-    if (body instanceof Buffer) {
-      res.end(body);
-    } else {
-      body.pipe(res);
+    if (body !== null) {
```
styfle: If body is null, don't we still need to end the request?

Perhaps something like:

```js
if (body === null) {
  res.end();
}
```

styfle (approved):
Looks great, thanks! 🎉

packages/node/src/serverless-functions/serverless-handler.mts
```diff
@@ -116,28 +115,33 @@ export async function createServerlessEventHandler(
...
     } else {
-      body = await streamToBuffer(response.body);
+      body = Buffer.from(await response.arrayBuffer());
 
       const contentEncoding = response.headers.get('content-encoding');
       if (contentEncoding) {
         body = compress(body, contentEncoding);
-        response.headers.set('content-length', Buffer.byteLength(body));
+        const clonedHeaders = [];
+        console.log(response.headers.entries());
```
ronag, line 130: Accidentally left here?

Kikobeats: yes; https://github.com/vercel/vercel/pull/10417 🙏

PR #10388, opened 2023-08-23
[next] fix RSC matching behavior & 404 status code on `fallback: false`

### LINKED ISSUE #53776: router.push() issue when navigating from app router to pages
### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

```bash
Operating System:
      Platform: darwin
      Arch: arm64
      Version: Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000
    Binaries:
      Node: 16.19.0
      npm: 8.19.3
      Yarn: 1.22.19
      pnpm: 8.6.11
    Relevant Packages:
      next: 13.4.14-canary.1
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0
      typescript: 5.1.3
    Next.js Config:
      output: N/A
```

### Which area(s) of Next.js are affected? (leave empty if unsure)

App Router, Routing (next/router, next/navigation, next/link)

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/nicomatsmart/next-router-push-issue

### To Reproduce

1. Visit https://next-router-push-issue.vercel.app/
2. Click on the button
3. You end up on the `CatchAllPage`, url says `/about`
4. Refreshing should show the proper about page

Bug does not reproduce locally.

### Describe the Bug

Using `router.push()` to navigate from app router to a `pages` page seems to get caught in the `[...all]` page in app router.

Bug might be related to both the middleware setup and the folder structure. Removing the `[...all]` page "solves" the issue. Worth repeating that the bug does not reproduce locally, only on Vercel.

### Expected Behavior

Using a `router.push()` from the app router should show the page you wanted to navigate to (in this reproduction, the `/about` page).

### Which browser are you using? (if relevant)

_No response_

### How are you deploying your application? (if relevant)

Vercel

<sub>[NEXT-1537](https://linear.app/vercel/issue/NEXT-1537/routerpush-issue-when-navigating-from-app-router-to-pages)</sub>

ztanner:
Fixes two separate issues for the Next builder:

- `pages` routes unexpectedly matching to RSC routes when prefetching from `app`. This update will attempt to match the route with the corresponding `pages` entry rather than falling back to a catch-all RSC
  - Fixes https://github.com/vercel/next.js/issues/53776
- `fallback: false` returning a successful status code when underlying page as a param (e.g. `/blog/[slug]` would 200 but `/blog/non-existent` would 404)
  - [slack x-ref](https://vercel.slack.com/archives/C03S8ED1DKM/p1692817762403579)


packages/next/src/server-build.ts
```diff
@@ -1543,6 +1543,39 @@ export async function serverBuild({
       // with that routing section
       ...afterFilesRewrites,
 
+      ...(appDir
```
ijjk, line 1542: I think we want to keep these above `afterFilesRewrites` so that those can match properly without the extension.

ijjk (approved):

PR #10392, opened 2023-08-24
[build-utils] Support serverless function architecture

AndrewBarba:
# Problem

Framework authors often produce build outputs from platforms like Github Actions or M1 Macs where the arm64 architecture is being used. They then deploy these outputs to Vercel using `vercel deploy`, bypassing Vercel's build system. Today they must cross compile to x86_64 in order to deploy compatible Serverless functions to Vercel.

# Solution

Allow Framework authors to detect the current architecture and specify either x86_64 or arm64 when deploying a Serverless function to Vercel.

# Related PRs

https://github.com/vercel/api/pull/21559

https://github.com/vercel/proxy/pull/6901

https://github.com/vercel/front/pull/24924


packages/build-utils/src/lambda.ts
```diff
@@ -102,6 +107,17 @@ export class Lambda {
     assert(typeof runtime === 'string', '"runtime" is not a string');
     assert(typeof environment === 'object', '"environment" is not an object');
 
+    if (architecture !== undefined) {
+      assert(
```
styfle, line 111: You can remove the assert for string if you're going to check the actual values below

TooTallNate (dismissed):
<picture data-single-emoji=":lgtm:" title=":lgtm:">:lgtm:</picture> We should probably update all of our internal builders which return Lambda instances to set this value properly, but that can happen in a separate PR.

Don't forget to add a changeset please.

TooTallNate (approved):

PR #10393, opened 2023-08-24
[build-utils] Add getOutputPath and isDirectory

pvdz:
Ported from vc build. Need this logic in the build container too.
The logic was ported from vc-build/build.ts

### DIFF
--- a/packages/build-utils/CHANGELOG.md
+++ b/packages/build-utils/CHANGELOG.md
@@ -1,5 +1,10 @@
 # @vercel/build-utils
 
+## 7.0.1
+
+- Added `getOutputPath` to abstract the logic of determining the output folder. It will return a string which ought to be the path to the output folder.
+- Added `isDirectoryPath` which is the helper function used to determine whether a path is a directory, or not when it is a file or doesn't exist at all. It will return a boolean.
+
 ## 7.0.0
 
 ### Major Changes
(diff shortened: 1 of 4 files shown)


cb1kenobi (dismissed):
Looks good, just needs a changeset.

styfle (changes requested):
This shouldn't be in `@vercel/build-utils` since that is designed for builders.

If it can go in another package, then it should since `@vercel/build-utils` is an implicit dep of all builders.

packages/build-utils/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # @vercel/build-utils
 
+## 7.0.1
+
+- Added `getOutputPath` to abstract the logic of determining the output folder. It will return a string which ought to be the path to the output folder.
+- Added `isDirectoryPath` which is the helper function used to determine whether a path is a directory, or not when it is a file or doesn't exist at all. It will return a boolean.
```
styfle, line 6: You shouldn't edit this file directly, its automatically generated by a bot

pvdz: ack

pvdz:
@styfle I Was explicitly asked to put it here. Have no pref myself. Please see slack.

pvdz:
Ok. going to add this in https://github.com/vercel/api/pull/21531 and let somebody else decide where to share this.

PR #10395, opened 2023-08-24
[examples] fix redwood template and update examples to use at least node@16

EndangeredMassa:
The redwood template was broken because it would use node@18, which is not supported for the version of redwood used in the template. This PR updates that version to be node@16, which does work.

While we're at it, I also updated other examples to be at least node@16. I tested deployments of each of these and the all work.


examples/redwoodjs/package.json
```diff
@@ -14,7 +14,7 @@
     "extends": "@redwoodjs/eslint-config"
   },
   "engines": {
-    "node": ">=12",
+    "node": ">=16",
```
TooTallNate: This would still use Node 18, no?

EndangeredMassa: ack, forgot to push up that change; thanks!

https://github.com/vercel/vercel/pull/10395/commits/6ba9b2d7a3d429a8199222233d6eb1e3e5f39c07

TooTallNate (dismissed):

leerob (dismissed):

examples/umijs/package.json
```diff
@@ -46,6 +46,6 @@
     ]
   },
   "engines": {
-    "node": ">=8.0.0 <17"
+    "node": ">=16.0.0 <17"
```
cb1kenobi: **suggestion (non-blocking):** shorthand range?‏

```suggestion
    "node": "16.x"
```

cb1kenobi (dismissed):

trek (approved):

cb1kenobi (approved):
LGTM

leerob (approved):

TooTallNate (approved):

PR #10396, opened 2023-08-24
[cli] Improve error messages for JSON parse issues

trek:
Replaces `JSON.parse` with [`json-parse-better-errors`](https://www.npmjs.com/package/json-parse-better-errors) in locations where we've read the JSON from a file which should give the person encountering an error a bit more info to correct what went wrong.

Before:
CleanShot 2023-08-24 at 13 33 31@2x

After:
CleanShot 2023-08-24 at 13 33 50@2x

I skipped updating `commands/secrets.js` since that commands appears to basically be dead.


EndangeredMassa (approved):

EndangeredMassa:
@trek Can we add some tests around the error messages?

Examples:
- https://github.com/vercel/vercel/blob/db27f409b6d6d0ebcf4344d2e25a1636b264faac/packages/cli/test/integration-3.test.ts#L717-L723
- https://github.com/vercel/vercel/blob/667af829c42b97988c161810c8c7404567b23ddb/packages/cli/test/unit/util/dev/validate.test.ts#L40-L45

PR #10397, opened 2023-08-24
[examples] Update Astro starter

leerob:
Deployed https://astro.vercel.app/image.


examples/astro/src/pages/image.astro
```diff
@@ -0,0 +1,6 @@
+---
+import { Image } from 'astro:assets';
+import astroLogo from '../assets/logo.png';
+---
+
+<Image src={astroLogo} alt="Astro Logo" width={50} />
```
styfle: I checked the deployed image and it looks like the default quality is `100` which is different that Next.js

That's probably okay if Astro wants a different default, but maybe we should add quality here to make sure users don't end up with huge images?

```suggestion
<Image src={astroLogo} alt="Astro Logo" width={50} quality={75} />
```

styfle (dismissed):
Looks like lint is failing.

Also, don't forget to deploy again.

leerob:
Ready to go again 👍

styfle (approved):
Looks good!

In the future, we might want to add a lock file.

cb1kenobi (approved):

PR #10403, opened 2023-08-26
[build-utils] add descriptions to NodeVersion properties

EndangeredMassa:
It wasn't clear to me what these values could look like. After figuring it out, I wanted to save others the trouble.

### DIFF
--- a/.changeset/brown-wasps-exercise.md
+++ b/.changeset/brown-wasps-exercise.md
@@ -1,0 +1,5 @@
+---
+"@vercel/build-utils": patch
+---
+
+add descriptions to NodeVersion properties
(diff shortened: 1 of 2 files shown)


.changeset/brown-wasps-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/build-utils": patch
```
EndangeredMassa, line 2: I think it's debatable if this is a versioned change, but I think it counts because the interface (to a dev) changes and there could always be a typo that breaks syntax.

styfle (approved):

PR #10410, opened 2023-08-28
Keep `vercel@canary` up to date with `vercel@latest`

TooTallNate:


styfle (changes requested):
Let's delete the canary tag instead

styfle (dismissed):
I just checked, and it looks like 10% of downloads are canary so we do need it 

https://www.npmjs.com/package/vercel?activeTab=versions

image

ijjk (approved):
As discussed in office big +1 so that the 10% of users on canary get the latest fixes

PR #10415, opened 2023-08-30
[next]: Fix RSC rewrite behavior

### LINKED ISSUE #54698: non-pre-generated dynamic pages will cause unexpected hard refresh when routing on Vercel via `generateStaticParams`
### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

```bash
Operating System:
      Platform: darwin
      Arch: arm64
      Version: Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:44 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T8103
    Binaries:
      Node: 16.20.0
      npm: 8.19.4
      Yarn: 1.22.19
      pnpm: 8.6.7
    Relevant Packages:
      next: 13.4.20-canary.11
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0
      typescript: 5.1.3
    Next.js Config:
      output: N/A
```

### Which area(s) of Next.js are affected? (leave empty if unsure)

App Router, Data fetching (gS(S)P, getInitialProps), Routing (next/router, next/navigation, next/link)

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/Howard86/vercel-nextjs-dynamic-path-reload/tree/main

### To Reproduce

This issue only happens when deploying on Vercel, running it locally will not have issues.

1. (Optional) clone this [repo](https://github.com/Howard86/vercel-nextjs-dynamic-path-reload) and deploy to a Vercel project
2. Access pathname [/dynamic/two](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two) on a browser 
3. Open browser network tab on the side
4. Click `toggle` button (this will call `router.push` on the same `pathname` appending new search params)
5. It will unexpectedly cause the whole page to hard refresh and land on [/dynamic/two?modal=1](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two?modal=1) with a black modal open

Following the same step above on [/dynamic/one](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/one) will not cause the hard refresh.

### Describe the Bug

Browser unexpectedly hard refreshes when using `router.push` or `router.replace` to the same `pathname` with a search param on a non-pre-generated dynamic path via `generateStaticParams`

e.g. 
- on `/app/dynamic/[text]/page.tsx`, we are adding `generateStaticParams` to only pre-generate `/dynamic/one` without `/dynamic/two`. Also adding a `toggle` button so it uses `router.push` to the same `pathname` appending some `searchParams` (e.g. [/dynamic/two](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two) to [/dynamic/two?modal=1](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two?modal=1))
- Clicking `toggle` button on path [/dynamic/one](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/one) will work fine
- Clicking `toggle` button on path [/dynamic/two](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two) will cause hard refresh

However, when cling the `toggle` button again on path [/dynamic/two?modal=1](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two?modal=1) to navigate to [/dynamic/two](https://vercel-nextjs-dynamic-path-reload.vercel.app/dynamic/two)) to close the modal, it will work smoothly without hard refresh.

Note: 
- This only happens when deploying on Vercel, running it locally with `npm run build && npm run start` will work fine
- Such modal behaviour is controlled by a client-side component [ModalWrapper](https://github.com/Howard86/vercel-nextjs-dynamic-path-reload/blob/main/app/ModalWrapper.tsx) using `useSearchParams`, which is used in [root layout](https://github.com/Howard86/vercel-nextjs-dynamic-path-reload/blob/main/app/layout.tsx#L18) together with `Suspense`
- By completely removing `generateStaticParams`, both dynamic pages will work fine (e.g. [/dynamic/one](https://vercel-nextjs-dynamic-path-reload-git-bugfix-re-e46c12-howard86.vercel.app/dynamic/one) &  [/dynamic/two](https://vercel-nextjs-dynamic-path-reload-git-bugfix-re-e46c12-howard86.vercel.app/dynamic/two)  deployed via a preview on a new [branch](https://github.com/Howard86/vercel-nextjs-dynamic-path-reload/tree/bugfix/remove-generate-static-params) 

### Expected Behavior

Browser should not hard refresh when using `router.push` or `router.replace` to the same `pathname` with a search param on a non-pre-generated dynamic path via `generateStaticParams`

More specifically,
- Developers should be able to use `router.push` and `router.replace` on dynamic path, and controls UI behaviours with `useSearchParams` regardless usage of  `generateStaticParams`
- Users should be able to navigate on dynamic paths without hard refreshes

### Which browser are you using? (if relevant)

Should not be relevant but tested with Safari 16.5, Chrome 116.0.5845.110, Firefox 109.0.1

### How are you deploying your application? (if relevant)

Vercel

NEXT-1585

ztanner:
- Removes some of the hacks from #10388 that were attempting to resolve an issue with RSC prefetches to `pages` routes in favor of adding rsc rewrites for all dynamic paths, and letting it fall through to a 404 if there's no match
- Fixes an issue where RSC requests were matching the wrong path (filesystem rather than RSC variant) introduced in above mentioned change
  - Closes https://github.com/vercel/next.js/issues/54698


packages/next/test/fixtures/00-app-dir-catchall-pages/probes.json
```diff
@@ -12,7 +12,7 @@
     },
     {
       "path": "/en/about",
-      "status": 200,
+      "status": 404,
```
styfle: Is this expected to be 404 now? <picture data-single-emoji=":thonking2:" title=":thonking2:">:thonking2:</picture>

ztanner: Yeah I'm torn on this -- before it would most likely have been a 200 response with error HTML in the response body, for example:
CleanShot 2023-09-01 at 08 58 02@2x

it's a trade-off with this approach but semantically it seems correct since there's no RSC data on pages routes.

ztanner: Trying an approach in [e8e5e04](https://github.com/vercel/vercel/pull/10415/commits/e8e5e049bc1036574152c07c26b6f1167ce99220) to avoid the 404s 🤞

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1089,36 @@ export async function serverBuild({
...
+    const pagesData = await fs.readJSON(pagesManifest);
+    const pagesEntries = Object.keys(pagesData);
+
+    if (pagesEntries.length > 0) {
+      const dummyFilePath = path.join('./', 'dummyFile');
+      await fs.writeFile(dummyFilePath, 'RSC Placeholder');
+
+      pagesPlaceholderRscEntries = pagesEntries.reduce((acc, page) => {
+        acc[`${page.slice(1)}.rsc`] = new FileFsRef({
+          fsPath: dummyFilePath,
```
styfle: I assumed this would fail since I thought `fsPath` needed to be an absolute path and this looks like a relative path

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1089,36 @@ export async function serverBuild({
...
+      entryPath,
+      outputDirectory,
+      `server/pages-manifest.json`
+    );
+
+    const pagesData = await fs.readJSON(pagesManifest);
+    const pagesEntries = Object.keys(pagesData);
+
+    if (pagesEntries.length > 0) {
+      const dummyFilePath = path.join('./', 'dummyFile');
```
styfle: Is there a difference between `path.join('./', 'dummyFile')` and the shorter `./dummyFile`?

Was the latter not working on windows?

ztanner: I lifted this from https://github.com/vercel/vercel/blob/main/packages/next/src/server-build.ts#L1785 assuming there was a OS-specific reason, but I think I'm going to update this to write to `tmpdir()` since that feels more appropriate (unless there's an easier way to do this without writing anything to disk? 🤔 )

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1090,36 @@ export async function serverBuild({
     )
   );
 
+  let pagesPlaceholderRscEntries;
```
styfle:
```suggestion
  let pagesPlaceholderRscEntries: Record<string, FileFsRef> = {}
```

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1090,36 @@ export async function serverBuild({
...
+    if (pagesEntries.length > 0) {
+      const placeholderFilePath = path.join(tmpdir(), 'rsc-placeholder');
+      await fs.writeFile(placeholderFilePath, 'RSC Placeholder');
+
+      pagesPlaceholderRscEntries = pagesEntries.reduce((acc, page) => {
+        acc[`${page.slice(1)}.rsc`] = new FileFsRef({
+          fsPath: placeholderFilePath,
+        });
+        return acc;
+      }, {} as Record<string, FileFsRef>);
```
styfle:
```suggestion
      for (const page of pagesEntries) {
        pagesPlaceholderRscEntries[`${page.slice(1)}.rsc`] = new FileFsRef({
          fsPath: placeholderFilePath,
        });
      }
```

styfle (dismissed):

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1090,36 @@ export async function serverBuild({
...
+      outputDirectory,
+      `server/pages-manifest.json`
+    );
+
+    const pagesData = await fs.readJSON(pagesManifest);
+    const pagesEntries = Object.keys(pagesData);
+
+    if (pagesEntries.length > 0) {
+      const placeholderFilePath = path.join(tmpdir(), 'rsc-placeholder');
+      await fs.writeFile(placeholderFilePath, 'RSC Placeholder');
```
ijjk: Should we write to the output directory instead of the tmpdir the same as we do for the `__next_data_catchall.json`?

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1090,36 @@ export async function serverBuild({
...
+
+      for (const page of pagesEntries) {
+        const pathName = page.startsWith('/') ? page.slice(1) : page;
+        pagesPlaceholderRscEntries[`${pathName}.rsc`] = new FileFsRef({
```
ijjk: Do we also need the dummy `.prefetch.rsc` outputs as that is being requested when the prefetch header is present?

packages/next/src/server-build.ts
ztanner: Switched to using `FileBlob` instead

packages/next/src/server-build.ts
```diff
@@ -1090,6 +1089,34 @@ export async function serverBuild({
...
+    const pagesManifest = path.join(
+      entryPath,
+      outputDirectory,
+      `server/pages-manifest.json`
+    );
+
+    const pagesData = await fs.readJSON(pagesManifest);
+    const pagesEntries = Object.keys(pagesData);
+
+    if (pagesEntries.length > 0) {
```
styfle: You can remove this `if` statement since the for loop will skip iteration when `pagesEntries` length is zero

ijjk (approved):

styfle (approved):
Thanks!

hiretimsf:
The issue still persists. It hasn't solved the issue.

PR #10430, opened 2023-08-31
Use `esbuild` for non-ncc'd packages

TooTallNate:
The intention is for this to be a drop-in replacement for compiling TypeScript to JavaScript, but using `esbuild` instead of `tsc`. `tsc` is still needed, but only for the cases where we want to generate type definitions.


cb1kenobi (dismissed):
LGTM

internals/get-package-json/package.json
```diff
@@ -8,7 +8,9 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "run-p build:*",
+    "build:js": "node ../../scripts/esbuild.mjs",
+    "build:types": "tsc --declaration --emitDeclarationOnly",
```
styfle: Should these flags be moved into `tsconfig.json` instead?

packages/build-utils/package.json
```diff
@@ -11,7 +11,11 @@
     "directory": "packages/now-build-utils"
   },
   "scripts": {
-    "build": "node build",
+    "build": "run-p build:js build:types",
+    "build:js": "run-s build:single build:bundle",
+    "build:single": "node ../../scripts/esbuild.mjs",
+    "build:bundle": "node ../../scripts/esbuild.mjs --bundle",
```
styfle: Since these both have the same output, is it going to cause conflicts depending on which finishes first?

scripts/esbuild.mjs
```diff
@@ -0,0 +1,18 @@
+import { posix } from 'path';
+import { globSync } from 'glob';
+import { build } from 'esbuild';
+
+const cwd = process.cwd();
+const dist = posix.join(cwd, 'dist');
+const pattern = posix.join(cwd, 'src/**/*.{js,ts}');
```
styfle: Should this instead read the include/excludes from `tsconfig.json`?

scripts/esbuild.mjs
```diff
@@ -0,0 +1,18 @@
...
+const files = globSync(pattern);
+
+await build({
+  entryPoints: files,
+  bundle: process.argv.includes('--bundle'),
+  format: 'cjs',
+  outdir: dist,
+  platform: 'node',
+  target: 'es2021',
```
styfle: Also, should this use the target from `tsconfig.json` too?

internals/get-package-json/package.json
TooTallNate: We could do that. I don't have a strong preference.

packages/build-utils/package.json
TooTallNate: No, because `build:js` uses `run-s` (serial invocation), so the single-file output happens first (`tsc` equivalent), and then the bundle output happens second (`ncc` equivalent).

scripts/esbuild.mjs
TooTallNate: 3f5f368b197226466f38ba3a237ad6745edf0156

scripts/esbuild.mjs
TooTallNate: 3f5f368b197226466f38ba3a237ad6745edf0156

package.json
```diff
@@ -28,6 +29,7 @@
     "lint-staged": "9.2.5",
     "node-fetch": "2.6.7",
     "npm-package-arg": "6.1.0",
+    "npm-run-all": "4.1.5",
```
styfle: This package hasn't been updated in 5 years.

https://www.npmjs.com/package/npm-run-all

https://github.com/mysticatea/npm-run-all/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+vulnerability

Are you sure we need this?

packages/build-utils/package.json
styfle: Did you do a diff on the `dist` directory before and after to see if anything changed? For example, do we need to move any asset files?

TooTallNate: I did, and there's no external files used in build-utils. The file counts are identical:

```
$ ls dist.old
clone-env.d.ts      errors.js         fs/                         hard-link-dir.d.ts  nodejs-lambda.js   types.d.ts
clone-env.js        file-blob.d.ts    get-ignore-filter.d.ts      hard-link-dir.js    prerender.d.ts     types.js
debug.d.ts          file-blob.js      get-ignore-filter.js        index.d.ts          prerender.js       validate-npmrc.d.ts
debug.js            file-fs-ref.d.ts  get-platform-env.d.ts       index.js            schemas.d.ts       validate-npmrc.js
edge-function.d.ts  file-fs-ref.js    get-platform-env.js         lambda.d.ts         schemas.js
edge-function.js    file-ref.d.ts     get-prefixed-env-vars.d.ts  lambda.js           should-serve.d.ts
errors.d.ts         file-ref.js       get-prefixed-env-vars.js    nodejs-lambda.d.ts  should-serve.js

$ ls dist.new
clone-env.d.ts      errors.js         fs/                         hard-link-dir.d.ts  nodejs-lambda.js   types.d.ts
clone-env.js        file-blob.d.ts    get-ignore-filter.d.ts      hard-link-dir.js    prerender.d.ts     types.js
debug.d.ts          file-blob.js      get-ignore-filter.js        index.d.ts          prerender.js       validate-npmrc.d.ts
debug.js            file-fs-ref.d.ts  get-platform-env.d.ts       index.js            schemas.d.ts       validate-npmrc.js
edge-function.d.ts  file-fs-ref.js    get-platform-env.js         lambda.d.ts         schemas.js
edge-function.js    file-ref.d.ts     get-prefixed-env-vars.d.ts  lambda.js           should-serve.d.ts
errors.d.ts         file-ref.js       get-prefixed-env-vars.js    nodejs-lambda.d.ts  should-serve.js

$ ls dist.old/fs
download.d.ts                get-writable-directory.js  node-version.d.ts    normalize-path.js      rename.d.ts            run-user-scripts.js
download.js                  glob.d.ts                  node-version.js      read-config-file.d.ts  rename.js              stream-to-buffer.d.ts
get-writable-directory.d.ts  glob.js                    normalize-path.d.ts  read-config-file.js    run-user-scripts.d.ts  stream-to-buffer.js

$ ls dist.new/fs
download.d.ts                get-writable-directory.js  node-version.d.ts    normalize-path.js      rename.d.ts            run-user-scripts.js
download.js                  glob.d.ts                  node-version.js      read-config-file.d.ts  rename.js              stream-to-buffer.d.ts
get-writable-directory.d.ts  glob.js                    normalize-path.d.ts  read-config-file.js    run-user-scripts.d.ts  stream-to-buffer.js
```

package.json
TooTallNate: There's some other similar tools available. The repo has a thread about alternatives here: https://github.com/mysticatea/npm-run-all/issues/10

I wanted one that was npm-script-centric (i.e. don't need to specify "pnpm run …" in the child commands), so this one seemed to fit that.

Agreed about the unmaintained aspect. I didn't notice that. Open to suggestions for alternatives.

TooTallNate: Another possibility - Rust-based script invoker `dum`. Though it doesn't currently support serial/parallel execution, there is a tracking issue for it: https://github.com/egoist/dum/issues/18

Maybe one of the Rust aficionados on the team would want to implement support for that 😄

styfle: I think we should drop the concurrent runner so we can merge this PR, then we can look for solutions and create a new PR in the future.

cb1kenobi: I'm cool with dropping the concurrent runner, but do you remember why we choose not to use https://www.npmjs.com/package/concurrently?

scripts/esbuild.mjs
```diff
@@ -0,0 +1,52 @@
+import ts from 'typescript';
+import * as path from 'path';
```
cb1kenobi: **suggestion (non-blocking):** I don't think we need the `* as`.‏

```suggestion
import path from 'node:path';
```

cb1kenobi (commented):
I have tested this a bunch and everything seems to work. I agree we should probably not use `npm-run-all` and I'll green light once a resolution has been reached.

package.json
TooTallNate: I took a different approach and removed `npm-run-all` here: f08e46af883b261e6fb63bcfbc9bc3cb2d3e1be1

cb1kenobi (approved):
This is tight, love it

styfle (approved):
Very nice, thanks! 🎉

PR #10432, opened 2023-08-31
[cli] Update `project` command to new structure

trek:
Before:
CleanShot 2023-08-31 at 13 29 29@2x

After:
CleanShot 2023-08-31 at 13 31 50@2x

PR #10434, opened 2023-08-31
[cli] migrate teams command

Ethan-Arrowood:
Before:
Screenshot 2023-08-31 at 13 59 35

After:
Screenshot 2023-08-31 at 14 00 23


trek (dismissed):

packages/cli/src/commands/teams/index.ts
```diff
@@ -1,63 +1,12 @@
...
-      ${chalk.cyan(`$ ${packageName} switch <slug>`)}
-
-      ${chalk.gray(
-        '–'
-      )} If your team's url is 'vercel.com/teams/name', 'name' is the slug
-      ${chalk.gray('–')} If the slug is omitted, you can choose interactively
-
-      ${chalk.yellow(
-        'NOTE:'
-      )} When you switch, everything you add, list or remove will be scoped that team!
```
EndangeredMassa, line 48: This note didn't make it over, but I think it's OK.

packages/cli/src/commands/teams/command.ts
```diff
@@ -0,0 +1,77 @@
...
+        },
+      ],
+      options: [],
+      examples: [],
+    },
+  ],
+  options: [
+    {
+      name: 'next',
+      shorthand: 'n',
```
EndangeredMassa: **suggestion (blocking):** ‏use capital

```suggestion
      shorthand: 'N',
```

The shorthand for this option is capital. https://github.com/vercel/vercel/blob/e43191b1866da70a3dab3815a3f2176942240ef3/packages/cli/src/commands/list/index.ts#L34

EndangeredMassa (changes requested):

packages/cli/src/commands/teams/command.ts
Ethan-Arrowood: Hm, other `--next` commands use `-n` though. Maybe we keep the `-N`, but i add another shorthand line for `-n` so that we are consistent across commands?

EndangeredMassa: It looks like `-N` everywhere in practice, but it was documented as `-n` in some: https://github.com/vercel/vercel/pull/10432#pullrequestreview-1610392313

In use, I can only find `-N`:

```
8 results - 8 files

packages/cli/src/commands/secrets.js:
  120      '-y': '--yes',
  121:     '-N': '--next',
  122    });

packages/cli/src/commands/alias/index.ts:
  27        '-y': '--yes',
  28:       '-N': '--next',
  29      });

packages/cli/src/commands/certs/index.ts:
  34        '--next': Number,
  35:       '-N': '--next',
  36        '--limit': Number,

packages/cli/src/commands/dns/index.ts:
  25        '--next': Number,
  26:       '-N': '--next',
  27        '--limit': Number,

packages/cli/src/commands/domains/index.ts:
  96        '--next': Number,
  97:       '-N': '--next',
  98        '-y': '--yes',

packages/cli/src/commands/list/index.ts:
  33        '--next': Number,
  34:       '-N': '--next',
  35        '--prod': Boolean, // this can be deprecated someday

packages/cli/src/commands/project/index.ts:
  24        '--next': Number,
  25:       '-N': '--next',
  26      });

packages/cli/src/commands/teams/list.ts:
  19      '-C': '--count',
  20:     '-N': '--next',
  21    });
```

I think this should be documented as `-N` everywhere to match existing behavior.

EndangeredMassa (approved):

PR #10437, opened 2023-08-31
[cli] Rename runtime package directories to runtime-* to group them

trek:
Renames various runtime packages to have a `runtime-` prefix so they're grouped.


trek:
Closing for now. There's too much active development in these packages to do a rename at the moment.

PR #10443, opened 2023-09-02
feat: add flag to geolocation

PaulieScanlon:
This PR adds `flag` to the geolocation export in the edge package.

E.g

```
import { geolocation } from '@vercel/edge';

export default async function handler(req) {
  const { flag } = geolocation(req);

  console.log(flag)
}

export const config = {
  runtime: 'edge',
};
```

```
$  🇬🇧 
```


TooTallNate (dismissed):

packages/edge/src/edge-headers.ts
```diff
@@ -68,6 +71,16 @@ function getHeader(request: Request, key: string): string | undefined {
   return request.headers.get(key) ?? undefined;
 }
 
+function getFlag(countryCode: string | undefined): string | undefined {
```
EndangeredMassa, line 83: **suggestion (non-blocking):** ‏add description

This function returns a string of unicode characters that represent the letters passed in, right? It took me a while to figure out what was happening here. Can you add a jsdoc description to this function?

For example:

```
'AA' -> '🇦🇦'
```

packages/edge/src/edge-headers.ts
```diff
@@ -68,6 +71,16 @@ function getHeader(request: Request, key: string): string | undefined {
...
+  if (!countryCode) return undefined;
+  return String.fromCodePoint(
+    ...countryCode
+      .toUpperCase()
+      .split('')
+      .map(char => 127397 + char.charCodeAt(0))
```
EndangeredMassa: **suggestion (non-blocking):** ‏name magic number

Can we pull `127397` into a named variable to make it clearer what it's used for? Is this the unicode offset before we get to the desired blocky letters?

packages/edge/src/edge-headers.ts
```diff
@@ -68,6 +71,16 @@ function getHeader(request: Request, key: string): string | undefined {
   return request.headers.get(key) ?? undefined;
 }
 
+function getFlag(countryCode: string | undefined): string | undefined {
+  if (!countryCode) return undefined;
```
EndangeredMassa: **question (non-blocking):** ‏Should there be a validation that the string is not above a certain length or contains certain allowed characters?

EndangeredMassa (dismissed):

packages/edge/src/edge-headers.ts
PaulieScanlon: Good point. This function will work providing the return value of `getHeader(request, COUNTRY_HEADER_NAME)` is a two-digit country code. There's nothing currently in place to check the return values of `getHeader`, I assume because it'll depend on what's passed in?

I'll add a check and test to ensure `countryCode` is only two-digits.

packages/edge/src/edge-headers.ts
```diff
@@ -68,6 +74,23 @@ function getHeader(request: Request, key: string): string | undefined {
...
+ *
+ * @param countryCode The country code returned by: `getHeader(request, COUNTRY_HEADER_NAME)`.
+ * @returns A flag emoji or undefined.
+ */
+function getFlag(countryCode: string | undefined): string | undefined {
+  const regex = new RegExp('^[A-Z]{2}$').test(countryCode!);
+  if (!countryCode || !regex) return undefined;
+  return String.fromCodePoint(
+    ...countryCode
+      .toUpperCase()
```
PaulieScanlon: `.toUpperCase()` might not be required now since the regex will catch any codes that aren't uppercase and return undefined.

Would you prefer to leave it in, or remove it?

PaulieScanlon: I'm also wondering if the `countryCode` param should be `country`, since that's how it's referred to else where.

packages/edge/docs/README.md
```diff
@@ -37,7 +38,7 @@ City of the original client IP as calculated by Vercel Proxy.
 
 #### Defined in
 
-[packages/edge/src/edge-headers.ts:4](https://github.com/vercel/vercel/blob/main/packages/edge/src/edge-headers.ts#L4)
+[packages/edge/src/edge-headers.ts:4](https://github.com/PaulieScanlon/vercel/blob/main/packages/edge/src/edge-headers.ts#L4)
```
EndangeredMassa: **question (blocking):** should these link changes be reverted?

packages/edge/src/edge-headers.ts
EndangeredMassa: `.toUpperCase()`: seems reasonable to remove

`countryCode`: seems reasonable to leave as is

EndangeredMassa (changes requested):

packages/edge/docs/README.md
PaulieScanlon: Oops, yes. I don't know how that happened, but good catch!

packages/edge/docs/interfaces/Geo.md
```diff
@@ -23,7 +24,7 @@ The city that the request originated from.
 
 #### Defined in
 
-[packages/edge/src/edge-headers.ts:47](https://github.com/vercel/vercel/blob/main/packages/edge/src/edge-headers.ts#L47)
+[packages/edge/src/edge-headers.ts:51](https://github.com/PaulieScanlon/vercel/blob/main/packages/edge/src/edge-headers.ts#L51)
```
EndangeredMassa: Can you fix these URLs as well?

EndangeredMassa (changes requested):

packages/edge/docs/interfaces/Geo.md
PaulieScanlon: Ah man, I don't know how this happened. I think when I added the const declaration it auto-generated this stuff?

Schniz: yeah it's probably our script that did that. we might need to make `github.com/vercel` as a fixed url in the typedoc config! <picture data-single-emoji=":meow_sorry:" title=":meow_sorry:">:meow_sorry:</picture>

packages/edge/src/edge-headers.ts
```diff
@@ -68,6 +74,22 @@ function getHeader(request: Request, key: string): string | undefined {
...
+ */
+function getFlag(countryCode: string | undefined): string | undefined {
+  const regex = new RegExp('^[A-Z]{2}$').test(countryCode!);
+  if (!countryCode || !regex) return undefined;
+  return String.fromCodePoint(
+    ...countryCode
+      .split('')
+      .map(char => EMOJI_FLAG_UNICODE_STARTING_POSITION + char.charCodeAt(0))
+  );
+}
```
Schniz: I saw this code on a blog post recently (cc @Kikobeats who sent it), do we need to provide attribution?

PaulieScanlon: Yes, it's this one: https://dev.to/jorik/country-code-to-flag-emoji-a21

I can add attribution, no problem.

cb1kenobi:
I've branched this PR and running the tests here: https://github.com/vercel/vercel/pull/10496

cb1kenobi (changes requested):
Code looks good, manually tested and works great. However, I branched this PR to run the tests (https://github.com/vercel/vercel/pull/10496) and the "docs are up to date" test is failing: https://github.com/vercel/vercel/actions/runs/6164248432/job/16729621798?pr=10496. I can also cause the tests to fail when running `pnpm test-unit` from the `packages/edge` directory. Give this a look and I can re-run the tests.

PaulieScanlon:
Hi @cb1kenobi, thanks for looking at this, with regard to failing tests. I've just pulled this branch and the following tests are failing, they are all outside of the edge package. 

- test/utils/changelog/parse.test.js
- test/utils/changelog/filter.test.js
- test/utils/changelog/group.test.js


I don't believe any of the changes I've made touch these tests, could it be they are failing from a previous commit?

I have update the docs by running `pnpm build:docs`... is that what I needed to do? sorry, totally confused by this repo!

packages/edge/docs/README.md
cb1kenobi: These links are breaking the tests: https://github.com/vercel/vercel/actions/runs/6235767980/job/16925730405?pr=10496#step:9:68

cb1kenobi (changes requested):

PaulieScanlon:
When i run `pnpm test` from the edge package, it changes all the references to the line numbers to use my username.

All the tests in the edge package now pass but what am I doing wrong? Why does running tests change stuff in a .md file?

cb1kenobi:
@PaulieScanlon I honestly don't know why it's changing the URLs. We don't have to deal with that because we don't use forks. In any case, I just pulled your branch into my PR so the tests can run. Fingers crossed! :) https://github.com/vercel/vercel/pull/10496

cb1kenobi:
@PaulieScanlon Tests pass! https://github.com/vercel/vercel/pull/10496

Huge thanks!

PaulieScanlon:
@cb1kenobi Super stuff! Do you know when this will likely be published to npm?

cb1kenobi:
@PaulieScanlon should be soon, like, this week. Hopefully. :)

PR #10452, opened 2023-09-05
[cli] Remove `mri` workaround

trek:
As of https://github.com/vercel/vercel/pull/10389 the `mri` package is no longer used and this workaround can be removed.

### DIFF
--- a/packages/cli/src/commands/secrets.js
+++ b/packages/cli/src/commands/secrets.js
@@ -334,7 +334,6 @@ async function run({ output, contextName, currentTeam, client }) {
     }
 
     const [name, value] = args;
-    let value = parsedValue;
 
     if (typeof value === 'boolean') {
       const example = chalk.cyan(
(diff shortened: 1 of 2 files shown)


styfle (dismissed):

packages/cli/src/commands/secrets.js
```diff
@@ -334,18 +334,7 @@ async function run({ output, contextName, currentTeam, client }) {
     }
 
     const [name, parsedValue] = args;
```
styfle:
```suggestion
    const [name, value] = args;
```

packages/cli/src/commands/secrets.js
```diff
@@ -334,18 +334,7 @@ async function run({ output, contextName, currentTeam, client }) {
...
-    const [originalName, originalValue] = client.argv.slice(-2);
-
     let value = parsedValue;
```
styfle:
```suggestion
```

cb1kenobi (dismissed):

styfle (approved):

cb1kenobi (approved):

PR #10458, opened 2023-09-06
[CLI] show instant preview url on `vc deploy` and `vc redeploy`

EndangeredMassa:
Show Instant Preview URLs immediately during `vc deploy`.

This does mean that we'll no longer show the nicer aliased URLs, but that's probably fine.

https://github.com/vercel/vercel/assets/41545/5d6d5695-60c3-49ca-a54d-a16828583070

---

- [Card](https://linear.app/vercel/issue/VCCLI-897/show-instant-preview-url-on-vc-deploy)


packages/cli/src/util/deploy/get-preferred-preview-url.ts
```diff
@@ -1,48 +0,0 @@
-import isWildcardAlias from '../alias/is-wildcard-alias';
-import { getCertsForCn } from '../certs/get-certs-for-cn';
-import Client from '../client';
-
-/**
- * Tries to find the "best" alias url.
- * @param aliasList
- */
-export async function getPreferredPreviewURL(
```
EndangeredMassa, line 9: This function was showing the best alias we could find, but for instant preview URLs, we want the unique deployment URL. Otherwise, we'll end up showing the old deployment.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -243,6 +265,7 @@ export default async function processDeployment({
 
       // Handle alias-assigned event
       if (event.type === 'alias-assigned') {
+        output.stopSpinner();
```
EndangeredMassa, line 264: This looks like a bug on `main`. During testing, I'd sometimes see the spinner run forever. This change fixed it.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -162,33 +166,47 @@ export default async function processDeployment({
...
       if (event.type === 'created') {
+        const deployment: Deployment = event.payload;
+
         await linkFolderToProject(
           client,
           cwd,
           {
             orgId: org.id,
-            projectId: event.payload.projectId,
+            projectId: deployment.projectId!,
```
EndangeredMassa, line 176: I had a check in here that would error if `deployment.projectId` was missing, but wasn't sure if that was save given current usage or possibly old deployments. The tests didn't like it either.

I don't love using `deployment.projectId!` here, but the rest of this bit of code is more typesafe than the rampant `any` usage before.

trek (commented):
🥁 tests tests tests 🥁

trek (dismissed):
I can't comment on the removal of the nicer urls, but otherwise 👍

TooTallNate (dismissed):

cb1kenobi (dismissed):
Tested and looks good!

EndangeredMassa:
> 🥁 tests tests tests 🥁

@trek there are many tests that cover this area of the code, which guided me through this change. The timing wasn't specifically tested before, but the output was. I thought that was enough, but asserting the timing would be good too.

Added an assertion for the timing. I tested that it fails without the code changes. https://github.com/vercel/vercel/pull/10458/commits/dfd0ede2a197aa1b259aae45d19f3d418432a30a

Ethan-Arrowood (approved):

packages/cli/src/util/deploy/process-deployment.ts
styfle: This seems like code smell. 

I think we've discussed this before with other PRs where `Deployment` type was incorrect based on the API response (I think @trek perhaps)

Did you check which API endpoint version is used here to make sure that projectId is always available?

EndangeredMassa: It uses `/v10/deployments` and `projectId` appears to always be available.

PR #10461, opened 2023-09-06
[cli] Update secrets.js to more current styles

trek:
Updates the `secrets` command to be a bit more modern. Initially I was going to covert this to typescript but that change was quite large, so going stepwise for easier review.

1. Moves the command implementation into a directory like other commands
2. Shifts the command data structure into its own file
3. Adds a test.

Other relevant comments inline.


packages/cli/src/commands/secrets/index.js
```diff
@@ -1,111 +1,17 @@
...
 import table from 'text-table';
 import ms from 'ms';
-import strlen from '../util/strlen.ts';
-import { handleError, error } from '../util/error';
-import NowSecrets from '../util/secrets';
-import exit from '../util/exit';
-import getScope from '../util/get-scope.ts';
-import confirm from '../util/input/confirm';
-import getCommandFlags from '../util/get-command-flags';
-import { packageName, getCommandName } from '../util/pkg-name.ts';
```
trek, line 11: extensions in imports no longer appear to be required.

packages/cli/test/mocks/secrets.ts
```diff
@@ -0,0 +1,29 @@
+import chance from 'chance';
+import { client } from './client';
+
+export function useSecrets({
```
trek, line 4: I don't love our current mock data/request setup but I did want to follow our existing pattern.

packages/cli/src/commands/secrets/index.js
```diff
@@ -130,7 +36,7 @@ const main = async client => {
     client.output.print(
       help(secretsCommand, { columns: client.stderr.columns })
     );
-    await exit(2);
```
trek, line 133: All our other commands use return values instead of `exit`, so migrated this to match. This is our final use of the `exit` library so that is removed in this PR as well.

packages/cli/src/commands/secrets/index.js
```diff
@@ -142,9 +48,11 @@ const main = async client => {
   try {
     ({ contextName } = await getScope(client));
   } catch (err) {
-    if (err.code === 'NOT_AUTHORIZED' || err.code === 'TEAM_DELETED') {
```
trek, line 145: Not strictly required here but TS will complain in the future and this is how other commands wrap API errors

packages/cli/src/commands/secrets/index.js
```diff
@@ -155,7 +63,7 @@ const main = async client => {
 
 export default async client => {
   try {
-    await main(client);
```
trek, line 158: need to return to test.

packages/cli/src/commands/secrets/index.js
```diff
@@ -218,7 +126,7 @@ async function run({ output, contextName, currentTeam, client }) {
       );
 
       if (out) {
-        console.log(`\n${out}\n`);
```
trek, line 221: need `client.output` to tests.

packages/cli/test/unit/commands/secrets.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const timeAgo = `${ms(
+      new Date().getTime() - new Date(created).getTime()
+    )} ago`;
+    await expect(client.stderr).toOutput(
+      '> NOTE: The `vercel env ls` command is recommended instead of `vercel secret ls`'
+    );
+
+    const output = client.stderr.read();
+    await expect(output).toMatch(name);
+    await expect(output).toMatch(timeAgo);
```
trek, line 35: This _may_ flake occasionally if you hit a `1000ms` boundary by chance. Sadly

```
jest.useFakeTimers();
jest.setSystemTime(new Date('2020-02-19'));
```

causes the test to hang and timeout due to other timer use in the code. A bit more refactoring will be required to mock time it seems.

packages/cli/src/commands/secrets/index.js
```diff
@@ -218,7 +126,7 @@ async function run({ output, contextName, currentTeam, client }) {
...
+        client.output.print(`\n${out}\n`);
```
trek, line 129: need `client.output` to tests.

.changeset/popular-impalas-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': minor
```
styfle: Why is it semver minor? Is this a new feature?

packages/cli/src/commands/secrets/index.js
Ethan-Arrowood: niceeeee

Ethan-Arrowood (commented):
lgtm - but i agree with Styfle that this isn't a minor change. I think `patch` is sufficient as it does _fix_ some things but doesn't add any new features for users.

Ethan-Arrowood (approved):

PR #10462, opened 2023-09-06
[static-build] Use esbuild

TooTallNate:
Switch to using esbuild to compile + bundle `@vercel/static-build`.


cb1kenobi (approved):
Works great! Let's go!

PR #10468, opened 2023-09-07
[go] Update to esbuild script

cb1kenobi:
Updated the `go` build script to use the new esbuilt script. Moved the `.ts` code into a `src` directory.


packages/go/tsconfig.json
```diff
@@ -11,8 +11,11 @@
...
     "noUnusedParameters": true,
     "noImplicitThis": false,
+    "outDir": "./dist",
     "types": ["node", "jest"],
     "strict": true,
     "target": "ES2021"
-  }
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules"]
```
TooTallNate: `node_modules` is excluded by default

```suggestion
  "include": ["src/**/*"]
```

.changeset/dry-baboons-happen.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
styfle, line 3: Should be patch so it gets published

packages/go/build.mjs
```diff
@@ -0,0 +1,3 @@
+import { esbuild } from '../../utils/build.mjs';
+
+await esbuild({ bundle: true, external: ['@vercel/build-utils'] });
```
styfle: Should also mark `dependencies` as external

TooTallNate (approved):

.changeset/dry-baboons-happen.md
cb1kenobi: Fixed

packages/go/build.mjs
cb1kenobi: Non issue now

styfle (approved):

PR #10469, opened 2023-09-07
Add optional experimentalStreamingLambda field for prerender

nabsul:
This adds a new `experimentalStreamingLambda` field to Prerender outputs, allowing references to an optional streaming lambda path.

### DIFF
--- a/packages/build-utils/src/prerender.ts
+++ b/packages/build-utils/src/prerender.ts
@@ -12,7 +12,7 @@ interface PrerenderOptions {
   initialStatus?: number;
   passQuery?: boolean;
   sourcePath?: string;
-  experimentalSteramingLambda?: string;
+  experimentalStreamingLambda?: string;
 }
 
 export class Prerender {
--- a/packages/build-utils/src/prerender.ts
+++ b/packages/build-utils/src/prerender.ts
@@ -12,7 +12,7 @@ interface PrerenderOptions {
   initialStatus?: number;
   passQuery?: boolean;
   sourcePath?: string;
-  experimentalStreamingLambda?: string;
+  experimentalStreamingLambdaPath?: string;
 }
 
 export class Prerender {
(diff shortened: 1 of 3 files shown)


cb1kenobi:
Do we have any tests that exercise this `experimentalStreamingLambda` property?

packages/build-utils/src/prerender.ts
```diff
@@ -130,5 +133,7 @@ export class Prerender {
       }
       this.allowQuery = allowQuery;
     }
+
+    this.experimentalStreamingLambda = experimentalStreamingLambda;
```
ijjk: We should validate this field the same way we do above for other fields.

ijjk:
@cb1kenobi no tests as this is being validated still and is experimental/might be removed.

packages/build-utils/src/prerender.ts
```diff
@@ -12,6 +12,7 @@ interface PrerenderOptions {
   initialStatus?: number;
   passQuery?: boolean;
   sourcePath?: string;
+  experimentalStreamingLambda?: string;
```
ijjk:
```suggestion
  experimentalStreamingLambdaPath?: string;
```
Maybe we call this the path to stress that's what it is and not a Lambda output itself like the `lambda` field.

packages/build-utils/src/prerender.ts
nabsul: Ah yeah. I forgot that TypeScript doesn't save you from that :-)

packages/build-utils/src/prerender.ts
nabsul: That is a better name!

nabsul:
Thanks for the feedback @cb1kenobi, I went ahead and added a unit test.

nabsul:
Moving to a non-forked PR: https://github.com/vercel/vercel/pull/10476

PR #10470, opened 2023-09-07
[python] Update to esbuild script

cb1kenobi:


packages/python/build.mjs
```diff
@@ -0,0 +1,3 @@
+import { esbuild } from '../../utils/build.mjs';
+
+await esbuild({ bundle: true, external: ['@vercel/build-utils'] });
```
styfle: Don't you need to mark all the `dependencies` in package.json as external too?

.changeset/slow-tips-jam.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
styfle, line 3: This should be patch so that it gets published

cb1kenobi: Fixed

.changeset/slow-tips-jam.md
```diff
@@ -0,0 +1,5 @@
...
+'@vercel/go': patch
```
styfle:
```suggestion
'@vercel/python': patch
```

styfle (approved):

packages/python/build.mjs
cb1kenobi: Non-issue... using the new builder-build.mjs script

PR #10473, opened 2023-09-07
[next] fix ENOENT on /404.html when `fallback: false` w/ `basePath`

ztanner:
The following error occurs during build when `basePath` is present in conjunction with `fallback: false` in `getStaticPaths`:

> Error: ENOENT: no such file or directory, open '/vercel/path0/.next/server/pages/404.html'

`localePrefixed404` was incorrectly being set to `false` because it was looking for `/<basePath>/<locale>/404.html` (when it's actually `/<locale>/404.html`)

This meant that inside `onPrerenderRoute`, `htmlFsRef` was pointing to `/404.html` rather than `/en/404.html`.

### DIFF
--- a/.changeset/nice-lies-sip.md
+++ b/.changeset/nice-lies-sip.md
@@ -1,0 +1,6 @@
+---
+"@vercel/next": patch
+"api": patch
+---
+
+fix ENOENT on /404.html when `fallback: false` w/ `basePath`
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -1321,11 +1321,7 @@ export const build: BuildV2 = async ({
       const localePrefixed404 = !!(
         routesManifest.i18n &&
         originalStaticPages[
-          path.posix.join(
-            entryDirectory,
-            routesManifest.i18n.defaultLocale,
-            '404.html'
-          )
+          path.posix.join('.', routesManifest.i18n.defaultLocale, '404.html')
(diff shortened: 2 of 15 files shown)


ijjk (approved):
Thanks!

.changeset/nice-lies-sip.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/next": patch
+"api": patch
```
styfle, line 3: Any idea why `api` is changing here? 

I think we need to modify that since `api` is private so its not published to npm and therefore isn't versioned.

https://github.com/vercel/vercel/pull/10474

ztanner: hmm, not sure, I just clicked the usual button from the bot reply and committed it 🤔

ztanner: Can actually see it encoded in the URL params:

```
value=---%0A%22%40vercel%2Fnext%22%3A%20patch%0A%22api%22%3A%20patch%0A---%0A%0A%5Bnext%5D%20fix%20ENOENT%20on%20%2F404.html%20when%20%60fallback%3A%20false%60%20w%2F%20%60basePath%60%0A```

styfle: @TooTallNate Any idea why the changesets bot thought that `api` changed?

PR #10476, opened 2023-09-08
Add optional experimentalStreamingLambda field for prerender

nabsul:
This adds a new `experimentalStreamingLambda` field to Prerender outputs, allowing references to an optional streaming lambda path.


cb1kenobi (approved):
LGTM

PR #10479, opened 2023-09-08
[remix] Bundle, remove types and source maps

TooTallNate:
For consistency with other Builders.


packages/remix/package.json
```diff
@@ -20,18 +20,18 @@
     "defaults"
   ],
   "dependencies": {
-    "@vercel/build-utils": "7.1.1",
     "@vercel/nft": "0.22.5",
```
TooTallNate, line 23: Note that `@vercel/nft` (has trouble bundling with esbuild) and `ts-morph` (it's huge) are intentionally left as regular deps, so that they will be excluded from the bundle, and shared among the other Builders which use these packages.

styfle: I wonder if `@vercel/static-config` should also stay in dependencies since it depends on `ts-morph`

TooTallNate: Oh interesting, I thought it was in `devDependencies`, but you are correct.

However, it's only used for types in `static-config`, and the bundle size does drop about 9.5mb after moving only the `ts-morph` here as expected, so I don't think it's necessary for `static-config` to be external. We may want to consider doing so to de-dup some other deps of `static-config` though, like ajv (about 1mb).

styfle: > However, it's only used for types in static-config,

I don't think so. I see `Node` from `ts-morph` is used at runtime here:

https://github.com/vercel/vercel/blob/667af829c42b97988c161810c8c7404567b23ddb/packages/static-config/src/index.ts#L72

TooTallNate: Good call. Ok let's move `static-config` too then.

styfle (dismissed):

styfle (approved):

PR #10480, opened 2023-09-08
[hydrogen] Bundle, remove types and source maps

TooTallNate:
Similar to #10479, but for `@vercel/hydrogen`.


cb1kenobi (approved):

packages/hydrogen/package.json
```diff
@@ -10,22 +10,24 @@
...
+    "build": "node ../../utils/build-builder.mjs",
     "test-e2e": "pnpm test test/test.js",
     "test": "jest --reporters=default --reporters=jest-junit --env node --verbose --bail --runInBand"
   },
   "files": [
     "dist",
     "edge-entry.js"
   ],
+  "dependencies": {
+    "ts-morph": "12.0.0"
```
styfle, line 22: I think `@vercel/static-config` should also be here since it depends on `ts-morph`

TooTallNate: Let's continue discussion [here](https://github.com/vercel/vercel/pull/10479#discussion_r1320382820).

PR #10485, opened 2023-09-11
next.js: add option to use bundled runtime

feedthejim:
This PR adds an environment variable that should allow us to test the bundled version for Next.js on Vercel, see https://github.com/vercel/next.js/pull/52997 for reference.

The changes include:
- a new environment variable `VERCEL_NEXT_BUNDLED_SERVER`
- some logic changes that will put app route handlers into their own lambda groups
- extra logic to require early the rendering runtimes (see PR above for details)


.changeset/old-teachers-dress.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/next": patch
+"api": patch
```
styfle:
```suggestion
```

Remove `api` here since its not a public package

.changeset/poor-baboons-report.md
styfle, line 1: This whole file should be removed. Each PR should have a single changeset

packages/next/src/server-build.ts
```diff
@@ -428,7 +451,11 @@ export async function serverBuild({
       }
 
       if (lambdaAppPaths[page]) {
-        appRouterPages.push(page);
+        if (lambdaAppPaths[page].fsPath.endsWith('route.js')) {
+          appRouteHandlers.push(page);
+        } else {
+          appRouterPages.push(page);
+        }
```
styfle, line 458: This change is not behind a flag. Is that intentional?

feedthejim: yes, we should have app route handlers into their own lambdas!

packages/next/test/fixtures/37-bundled-server/next.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  experimental: {
+    
+  },
+};
```
styfle: You can probably remove this file

packages/next/test/fixtures/37-bundled-server/probes.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      "path": "/api/hello",
+      "status": 200,
+      "mustContain": "hello world"
+    },
+    {
+      "path": "/",
+      "status": 200,
+      "mustContain": "hello world"
```
styfle: All of these tests are looking for the same text. Can we update each so they return something different that that will ensure it works as expected?

styfle (approved):
Thanks!

Might be good to have @ijjk take a look too

ijjk (approved):

PR #10490, opened 2023-09-12
[examples] Use placeholder for API Key

AndyBitz:
The Ionic example has an actual Google Maps API key by default and we'd like to not have it displayed, so we'll replace it with a placeholder instead. Considering it's commented out anyways, this will be a no-op.


styfle (dismissed):

styfle (approved):

aaronbrown-vercel (approved):
lgtm! 🚢

PR #10492, opened 2023-09-12
[frameworks] Add `bun install` placeholder

styfle:
image


trek (dismissed):

EndangeredMassa (dismissed):
Should we bring Astro up to date with the full list as well? "Other" is kind of a separate thing, but should we add the other package managers there too?

styfle:
@EndangeredMassa Nice catch, updated 👍

trek (approved):

EndangeredMassa (approved):

cb1kenobi (approved):

Ethan-Arrowood (approved):
Lol i think we need a "default placeholder". There is no point in having this string duplicated like this so many times

cb1kenobi (approved):

PR #10495, opened 2023-09-12
[routing-utils] Adds support for `statusCode` property on `rewrites`

ofhouse:
Allows to define a `statusCode` property on `rewrites` in the same way as we already have it for `redirects`.
This is still blocked internally as it requires a feature flag to build with that setting.

So it would allow to set the status in `vercel.json` like this:
```json
{
  "rewrites": [
    {
      "source": "/*",
      "destination": "not-found.html",
      "statusCode": 404
    }
  ]
}

```

Once released, I'll update the documentation on that as well.


packages/cli/test/dev/fixtures/test-rewrites/vercel.json
```diff
@@ -9,5 +9,15 @@
...
-  "rewrites": [{ "source": "/hello", "destination": "index.html" }]
+  "rewrites": [
+    {
+      "source": "/hello",
+      "destination": "index.html"
+    },
+    {
+      "source": "/status-rewrite-425",
+      "destination": "index.html",
+      "statusCode": 425
```
styfle, line 20: I'm not sure I understand the use case for this feature. Why would you need to change the status code for a rewrite to a static file? And if its not static, seems like you could return the correct status code from the function instead.

ofhouse: Originally I had an example with `404` which is a more practical use case for that.
I decided to use the `425` status to make sure the tests don't pass if mistakenly it hits a non-existing route which would produce a `404` automatically.

packages/cli/test/dev/integration-2.test.ts
```diff
@@ -81,6 +81,13 @@ test(
   })
 );
 
+test(
+  '[vercel dev] test rewrites can change status code',
+  testFixtureStdio('test-rewrites', async (testPath: any) => {
+    await testPath(425, '/status-rewrite-425', 'Hello World');
```
styfle, line 81: Lets move this into the previous test testFixtureStio on line 80 since this is going to create a deployment and both rewrites are in the same deployment.

styfle (approved):
Looks good, thanks!

Might want to get one more review. Perhaps @ijjk or @EndangeredMassa

EndangeredMassa (approved):

PR #10500, opened 2023-09-13
[speed insights] Prepare for migration to new speed insights package

tobiaslins:
During `vc build` do following when `@vercel/speed-insights` package is in dependencies:
- Show a warning when `VERCEL_ANALYTICS_ID` is set in environment variables
- Unset it in process.env to prevent auto-injecting old speed insights in Next.js

Durning `vc env pull` prevent pulling internal environment variables `VERCEL_ANALYTICS_ID`, `VERCEL_SPEED_INSIGHTS_ID` & `VERCEL_WEB_ANALYTICS_ID`. They are never required in the frontend


packages/cli/src/commands/build/index.ts
```diff
@@ -426,6 +426,18 @@ async function doBuild(
 
   const ops: Promise<Error | void>[] = [];
 
+  const isUsingSpeedInsights = Boolean(
+    pkg?.dependencies?.['@vercel/speed-insights']
```
feugy: **thought:** what if user includes it in their `devDependencies`?

It seems that next [makes no distinction](https://github.com/vercel/next.js/issues/43066#issuecomment-1319969015) since every required lib will be bundled together.

As a vite user, I personally puts everything in `devDependencies` because eventually, there are no node modules for a bundle app.

feugy (commented):
For other reviewers: this PR comes as part of a [larger plan](https://www.notion.so/vercel/Dedicated-Speed-Insights-package-fc8969cde78340428e48a2c9122d967c?pvs=4#d731c3ec2db94275bf61761d3d4be44d) to sunset auto-injecting speed insights when deploying an application, in favor of an explicit package import and usage.

@EndangeredMassa could you point any automated test the could inspire from please?

EndangeredMassa (dismissed):

feugy (dismissed):
Looks good to me, let's wait for zero-config team to share their feedback.

packages/cli/src/commands/build/index.ts
```diff
@@ -426,6 +426,23 @@ async function doBuild(
...
+    d => d === '@vercel/speed-insights'
+  );
+
+  if (isUsingSpeedInsights && process.env.VERCEL_ANALYTICS_ID) {
+    output.warn(
+      `The \`VERCEL_ANALYTICS_ID\` environment variable is deprecated and will be removed in a future release. Please remove it from your environment variables`
+    );
+
+    delete process.env.VERCEL_ANALYTICS_ID;
+  }
```
EndangeredMassa, line 446: I wrote up an example test here. It's only asserting the warning, but I think that's good enough for this logic. https://github.com/vercel/vercel/pull/10541

cb1kenobi (dismissed):
LGTM

EndangeredMassa (approved):

Ethan-Arrowood (approved):

PR #10504, opened 2023-09-13
[next] missed a prerender for experimentalBypassFor

ztanner:
Missed this in https://github.com/vercel/vercel/pull/10497


trek (dismissed):

styfle (dismissed):

cb1kenobi (dismissed):

packages/next/src/utils.ts
```diff
@@ -2213,6 +2213,7 @@ export const onPrerenderRoute =
           fallback: jsonFsRef,
           group: prerenderGroup,
           bypassToken: prerenderManifest.bypassToken,
+          experimentalBypassFor,
```
ijjk, line 2232: Looks like we also need to parse this value for `fallbackRoutes` so it's available once we have it in the `dynamicRoutes` field as well

x-ref: https://github.com/vercel/vercel/blob/1abda9ca871ce4fad98b260c9467fd5a7f294a40/packages/next/src/utils.ts#L1207-L1227

cb1kenobi (dismissed):

cb1kenobi:
PR looks good to me, but I don't have all the context. @ijjk what do you think?

ijjk (approved):
Changes look good, failure appears to be unrelated https://github.com/vercel/next.js/pull/55345

PR #10510, opened 2023-09-14
[node] Run installCommand when zero config

cb1kenobi:
When building a project that does not have any `"builds"` in the `vercel.json`, it is considered a zero config build and should try to run the `installCommand` and right now it doesn't.


packages/node/src/index.ts
```diff
@@ -99,7 +100,21 @@ async function downloadInstallAndBundle({
     meta
   );
   const spawnOpts = getSpawnOptions(meta, nodeVersion);
-  await runNpmInstall(entrypointFsDirname, [], spawnOpts, meta, nodeVersion);
+
+  if (config.zeroConfig) {
```
```suggestion
  if (config.zeroConfig && config.installCommand === 'string') {
```

This might fix it.

Also I think we should turn this into a helper function so all builders can use it. Perhaps pass config to `runNpmInstall()`

cb1kenobi: Wow, that worked. I did a little cleanup. Dunno about passing `config` to `runNpmInstall()`. I would prefer to not touch public APIs in this PR.

styfle (commented):
Looks good, but needs a test.

Also what about other builders?

TooTallNate (changes requested):
This change is not correct. The `installCommand` is specific to the _frontend Builders_ (i.e. for your framework), of which `@vercel/node` is not.

When in zero-config mode, then `@vercel/static-build` (at least) will be executed, which already handles running the `installCommand`, so I'm not sure which configuration this change is attempting to solve.

Ethan-Arrowood (approved):

Ethan-Arrowood (changes requested):
actually - needs a test

styfle:
Closing per Nate's comment above: https://github.com/vercel/vercel/pull/10510#pullrequestreview-1631344104

PR #10527, opened 2023-09-18
Restore datadog checking with logs not exit codes

trek:
Restores DataDog flakey test detection. Briefly disabled because the old method of using exit codes would bail the entire workflow.

Taking the Turbo team's suggestion of looking at `runData.execution` values to avoid an extra loop. Now with unit tests!


.github/workflows/test.yml
```diff
@@ -91,11 +91,19 @@ jobs:
           VERCEL_TEST_TOKEN: ${{ secrets.VERCEL_TEST_TOKEN }}
           VERCEL_TEST_REGISTRATION_URL: ${{ secrets.VERCEL_TEST_REGISTRATION_URL }}
           FORCE_COLOR: '1'
+      - name: 'Determing Turbo HIT or MISS'
+        id: turbo-summary
+        shell: bash
+        run: |
+          TURBO_MISS_COUNT=`node utils/determine-turbo-hit-or-miss.js`
+          echo "MISS COUNT"
+          echo "$TURBO_MISS_COUNT"
```
styfle:
```suggestion
          echo "MISS COUNT: $TURBO_MISS_COUNT"
```

utils/determine-turbo-hit-or-miss.js
```diff
@@ -1,30 +1,35 @@
...
+
+  // Return so we can unit test.
+  return missCount;
 }
 
-main().catch(err => {
+const turboRunDirectory = '.turbo/runs';
+const turboRunDirectoryParent = path.join(__dirname, '..');
+
+main(turboRunDirectory, turboRunDirectoryParent).catch(err => {
```
styfle, line 30: Why two variables? Can you pass in the one var?

For example:

```ts
const turboRunDir = path.join(__dirname, '..', '.turbo', 'runs')
```

trek: To support having two named variables in the script body for readability without needing to split a single value.

styfle (approved):

.github/workflows/test.yml
```diff
@@ -91,11 +91,18 @@ jobs:
...
+        run: |
+          TURBO_MISS_COUNT=`node utils/determine-turbo-hit-or-miss.js`
+          echo "MISS COUNT: $TURBO_MISS_COUNT"
+          echo "misses=$TURBO_MISS_COUNT" >> $GITHUB_OUTPUT
       - name: fetch ssl certificate after tests (linux, os x)
         if: matrix.runner != 'windows-latest'
         run: echo | openssl s_client -showcerts -servername 'api.vercel.com' -connect 76.76.21.21:443
       - name: 'Upload Test Report to Datadog'
-        if: always()
+        if: ${{ steps['turbo-summary'].outputs.misses != '0' }}
```
EndangeredMassa, line 105: @trek This seems to say "if there's at least one miss, upload the reports to datadog". Will there not almost always be at least one miss in a given PR?

cb1kenobi: I suppose if the PR only touches a README or something, it wouldn't have any misses and thus we wouldn't want to upload the report.

trek: On PRs that have code changes for a given package, yes. For packages without code changes we were still uploading the cached JUnit output files. For some reason installing `datadog-ci` is super slow on Windows so this was adding several minutes to those runs even when the package has no changes from the previous runs.

trek: Just to be super clear, this step runs once per each item in our test matrix.

trek: I changed a test file in this PR to demonstrate, although it's a bit hard to find. But compare https://github.com/vercel/vercel/actions/runs/6224030003/job/16892150477?pr=10527 to https://github.com/vercel/vercel/actions/runs/6224030003/job/16892124841?pr=10527

CleanShot 2023-09-18 at 10 17 32@2x
CleanShot 2023-09-18 at 10 17 43@2x

utils/determine-turbo-hit-or-miss.js
```diff
@@ -1,30 +1,35 @@
...
-    const runData = await fs.readJson(runFile);
-    const tasksReports = runData.tasks || [];
+  let missCount = 0;
 
-    const missCount = tasksReports.reduce((total, taskData) => {
-      if (taskData.cache.status === 'MISS') {
-        return total + 1;
-      }
+  await Promise.all(
+    turboRunFiles.map(async fileName => {
```
cb1kenobi, line 11: Do you know how many files are in the `.turbo/runs` directory? I've it's dozens, then cool, but if it's hundreds or more, I have see issues with degraded performance and used something like https://www.npmjs.com/package/p-limit to limit the "concurrency".

trek: @cb1kenobi currently only ever 1, but there was a suggestion somewhere in Slack that we might have multiple if we ever changed our test matrix to run `turbo` several times in one run. Then it would have 1 file per invocation of `turbo --summarize`

cb1kenobi (approved):

utils/determine-turbo-hit-or-miss.js
cb1kenobi: Ha, yeah, nevermind!

Ethan-Arrowood (approved):

.github/workflows/test.yml
EndangeredMassa: ohh, that makes more sense

PR #10530, opened 2023-09-18
[remix] Fix ESM mode for Edge runtime, add Remix v2 E2E test

TooTallNate:
Similar to the default import fix from #10525 which was for Node runtime, but this one fixes Edge runtime.

Also adds an E2E test for Remix v2, including both a Node route and Edge route.


Ethan-Arrowood (approved):
easy fix - love it

cb1kenobi (approved):

PR #10531, opened 2023-09-18
[next] fix feature flag detection

styfle:
For feature flags, we use the string `1` for enabled and the string `0` for disabled.

So we need to check the value of the env var, not the presence of the env var.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -313,7 +313,7 @@ export async function serverBuild({
 
     const useBundledServer =
       semver.gte(nextVersion, BUNDLED_SERVER_NEXT_VERSION) &&
-      process.env.VERCEL_NEXT_BUNDLED_SERVER;
+      process.env.VERCEL_NEXT_BUNDLED_SERVER === '1';
 
     if (useBundledServer) {
       debug('Using bundled Next.js server');
--- a/.changeset/late-impalas-nail.md
+++ b/.changeset/late-impalas-nail.md
@@ -1,0 +1,5 @@
+---
+"@vercel/next": patch
+---
+
+Fix feature flag detection


feedthejim (dismissed):

ijjk (dismissed):

TooTallNate (approved):

trek (approved):

packages/next/src/server-build.ts
```diff
@@ -313,7 +313,7 @@ export async function serverBuild({
 
     const useBundledServer =
       semver.gte(nextVersion, BUNDLED_SERVER_NEXT_VERSION) &&
-      process.env.VERCEL_NEXT_BUNDLED_SERVER;
+      process.env.VERCEL_NEXT_BUNDLED_SERVER === '1';
```
cb1kenobi, line 316: Or you can coerce it into a boolean :)

```suggestion
      ~~process.env.VERCEL_NEXT_BUNDLED_SERVER;
```

cb1kenobi (approved):

packages/next/src/server-build.ts
styfle: @cb1kenobi No we can't, see the PR description

cb1kenobi: That's why I said `~~`:

```
> ~~null
0
> ~~undefined
0
> ~~''
0
> ~~false
0
> ~~0
0
> ~~'0'
0
> ~~true
1
> ~~1
1
> ~~'1'
1
```

TooTallNate: Too clever / unreadable IMO. I have no idea what that code does just by looking at it.

styfle: Very strange, thats not even coercing to a boolean, its coercing to a number.

Almost the same as `Number(x)`. But I think we should be explicit here here so I'm going to leave it.

I agree with Nate

Ethan-Arrowood: I didn't even know `~` is an operator lol

Ethan-Arrowood (approved):

PR #10533, opened 2023-09-18
[cli] Implement python dev server

trek:
Implements the `startDevServer` function for python uses of `vc dev`. Having a little trouble connecting the node bits to the python bits at the moment.


trek:
Closing for now. The draft work is stored on https://github.com/vercel/vercel/tree/trek/zero-1012-create-python-dev-server

PR #10534, opened 2023-09-18
Update TypeScript to v5.2.2

TooTallNate:


devjiwonchoi:
Assuming #11516 is caused by typescript version, do you mind if I bump ts version to v5?

EndangeredMassa:
@devjiwonchoi last time we tried to update to 5+, there were a lot of typecheck failures to work through. We'd like to update, but haven't prioritized that just yet.

We will get to it, though.

devjiwonchoi:
Got it. Thank you!

PR #10541, opened 2023-09-19
[tests] add test to #10500

EndangeredMassa:
Adding a test to https://github.com/vercel/vercel/pull/10500

### DIFF
--- a/packages/cli/test/helpers/prepare.js
+++ b/packages/cli/test/helpers/prepare.js
@@ -421,6 +421,7 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
         projectId: '.',
         settings: {
           framework: null,
+          installCommand: 'echo "skipping install"',
         },
       }),
       'package.json': JSON.stringify({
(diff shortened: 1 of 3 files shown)


trek (approved):

packages/cli/test/helpers/prepare.js
```diff
@@ -415,6 +415,23 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
...
+        settings: {
+          framework: null,
+        },
+      }),
+      'package.json': JSON.stringify({
+        scripts: {
+          build: 'mkdir -p public && echo hi > public/index.txt',
+        },
+        dependencies: {
+          '@vercel/speed-insights': 'latest',
```
styfle: Does this get installed? If so, lets use a specific version here

cb1kenobi (approved):
LGTM

packages/cli/test/helpers/prepare.js
EndangeredMassa: yeah, good call; updated

styfle (approved):

EndangeredMassa:
huh, I had this test passing locally; will look into it

tobiaslins:
@EndangeredMassa thanks!
The issue is that our package is private and you most probably have the NPM_TOKEN on your local machine. Lets just map it to some random library `"@vercel/speed-insights": "npm:server-only"` or something. This should be enough. 

We can change this later on, once we published the package!

EndangeredMassa:
Yeah, makes sense. I just updated it to skip the install step for this test.

PR #10549, opened 2023-09-20
[cli][next][build-utils] Variants

AndyBitz:
Add variants. Updates CLI, `build-utils`, and `next`.


dferber90 (dismissed):

packages/cli/src/commands/build/index.ts
```diff
@@ -774,3 +778,17 @@ function mergeWildcard(
...
+function mergeFlags(
+  buildResults: Iterable<BuildResult | BuildOutputConfig>
+): BuildResultV2Typical['flags'] {
+  let flags: Flag[] = [];
+
+  for (const result of buildResults) {
+    if ('flags' in result) {
+      flags.push(...(result.flags ?? []));
+    }
+  }
```
Ethan-Arrowood: I think `Array.prototype.flatMap` is more appropriate here. I find this code not very readable

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

packages/next/src/utils.ts
```diff
@@ -3025,3 +3025,35 @@ export function isApiPage(page: string | undefined) {
...
+  const pathVariantsManifest = path.join(
+    entryPath,
+    outputDirectory,
+    'variants-manifest.json'
+  );
+
+  const hasVariantsManifest = await fs
+    .access(pathVariantsManifest)
+    .then(() => true)
+    .catch(() => false);
```
Ethan-Arrowood, line 3050: Can you use the sync [`fs.existsSync`](https://nodejs.org/api/fs.html#fsexistssyncpath) instead?

packages/next/src/server-build.ts
```diff
@@ -1345,6 +1349,17 @@ export async function serverBuild({
...
+  const flags: Flag[] = [];
+  if (variantsManifest) {
+    Object.entries(variantsManifest).forEach(([key, value]) => {
+      flags.push({
+        key,
+        ...value,
+        metadata: value.metadata ?? {},
+      });
+    });
+  }
```
Ethan-Arrowood: Use map instead

```suggestion
  let flags: Flag[] = variantsManifest ? Object.entries(variantsManifest).map(([key, value]) => ({
        key,
        ...value,
        metadata: value.metadata ?? {},
    })) : [];
```

Ethan-Arrowood (changes requested):

packages/next/src/utils.ts
AndyBitz: I've copied this pattern from the other manifest files, I'm not sure if there's a specific reason for that, but for now I'd like to keep it consistent if possible.

Ethan-Arrowood: okay - this will be slower but I'm okay with consistency.

dferber90 (dismissed):

dferber90 (approved):

Ethan-Arrowood (approved):
Are there tests you can include with this PR or will that be a follow up?

AndyBitz:
> Are there tests you can include with this PR or will that be a follow up?

A follow up, eventually.

PR #10551, opened 2023-09-20
[vc cert ls] remove error shallowing

balazs4:
During [project-level-rbac bug bash](https://linear.app/vercel/issue/IAM-1267/vc-certs-ls-fails-with-typeerror-for-contributors) we were facing a runtime error 

```
Error: An unexpected error occurred in certs: TypeError: Cannot read properties of undefined (reading 'length')
```

The reason for that is because the API responded with `403`, which was expected in our test. However the CLI should not exit with a generic runtime error in this case.

This PR removes the `.catch(err => err)` error shallowing instruction.


.changeset/pink-pigs-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': patch
+---
+
+let bail APIErrors from getCerts call
```
TooTallNate: This goes in the changelog, so let's use a message more user-facing:

```suggestion
Better rendering upon authentication error in `vc cert ls`
```

TooTallNate (dismissed):

packages/cli/src/commands/certs/ls.ts
```diff
@@ -43,10 +43,7 @@ async function ls(
   }
 
   // Get the list of certificates
-  const { certs, pagination } = await getCerts(
-    client,
-    ...paginationOptions
-  ).catch(err => err);
+  const { certs, pagination } = await getCerts(client, ...paginationOptions);
```
EndangeredMassa, line 46: **suggestion (soft-blocking):** Can we add a test for the failure case here?

EndangeredMassa (dismissed):

packages/cli/src/commands/certs/ls.ts
balazs4: good point; I'll add a test for it. 🙌

balazs4: @EndangeredMassa 

error message if the "fix" would not be in place:

image


https://github.com/vercel/vercel/pull/10551/commits/02327c95d17311b03c94722f6339f99545c663c7

EndangeredMassa (approved):

PR #10552, opened 2023-09-20
[go] add support for Go 1.21.0

Ethan-Arrowood:
as titled

updates a test to check for runtime version

### DIFF
--- a/packages/go/src/go-helpers.ts
+++ b/packages/go/src/go-helpers.ts
@@ -22,6 +22,7 @@ import type { Env } from '@vercel/build-utils';
 const streamPipeline = promisify(pipeline);
 
 const versionMap = new Map([
+  ['1.21', '1.21.0'],
   ['1.20', '1.20.2'],
   ['1.19', '1.19.6'],
   ['1.18', '1.18.10'],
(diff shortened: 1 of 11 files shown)


packages/go/src/go-helpers.ts
```diff
@@ -22,6 +22,7 @@ import type { Env } from '@vercel/build-utils';
 const streamPipeline = promisify(pipeline);
 
 const versionMap = new Map([
+  ['1.21', '1.21.0'],
```
styfle:
```suggestion
  ['1.21', '1.21.1'],
```

Let's use the latest version https://go.dev/dl/

packages/go/test/fixtures/10-go-mod-newer/go.mod
```diff
@@ -1,3 +1,3 @@
 module go-mod
 
-go 1.19
+go 1.21
```
styfle: Can you keep this test and instead update the tests without the go mod file? (those will likely fail since they get auto upgraded)

Ethan-Arrowood: okay

packages/go/src/go-helpers.ts
Ethan-Arrowood: Hmm I thought I committed this but I guess I just "resolved it". fixing now

TooTallNate (approved):

PR #10553, opened 2023-09-20
[node] Use "esbuild" to build package

TooTallNate:


Ethan-Arrowood (approved):

packages/node/package.json
```diff
@@ -47,16 +48,13 @@
     "@types/cookie": "0.3.3",
     "@types/etag": "1.8.0",
     "@types/jest": "29.5.0",
-    "@types/test-listen": "1.1.0",
     "@vercel/ncc": "0.24.0",
```
styfle, line 51: Can we remove `@vercel/ncc` too?

packages/node/src/index.ts
```diff
@@ -376,9 +377,6 @@ function getAWSLambdaHandler(entrypoint: string, config: Config) {
   return '';
 }
 
-// Ensures that everything from `types.ts` is exported in the final `index.d.ts` file.
-export * from './types';
```
styfle, line 380: I think this part is required so that users can

```ts
import type { VercelRequest } from '@vercel/node'
```

packages/node/package.json
TooTallNate: We can. I'll do it in a separate PR (https://linear.app/vercel/issue/ZERO-1003/remove-vercelncc-dep-from-all-packages-packagejson-files)

packages/node/tsconfig.json
```diff
@@ -7,11 +7,9 @@
     "module": "nodenext",
     "outDir": "dist",
     "sourceMap": false,
-    "declaration": true,
+    "declaration": false,
```
styfle, line 10: Similarly here, I think this will cause problems for 

```ts
import type { VercelRequest } from '@vercel/node'
```

packages/node/src/index.ts
TooTallNate: This is still supported. See in the `build.mjs` script that `src/types.d.ts` is copied to `dist/index.d.ts` which makes those types work for end-users.

styfle (approved):

packages/node/tsconfig.json
TooTallNate: I changed `types.ts` to be `types.d.ts`, so it doesn't need to be compiled via `tsc` (there was basically nothing to compile anyways, since it's already all types).

PR #10583, opened 2023-09-25
[cli] Update bun detection and add tests for projects with both `bunlock.b` and `yarn.lock` files

trek:
Updates package manager detection to account for two lock files. All other managers will only have the one lock file. Bun, however, may have both a `bun.lockb` _and_ a `yarn.lock` file. To ensure bun is properly detected, the presence of `bun.lockb` with `yarn.lock` must occur before `yarn.lock` so we don't mistake the presence of a `yarn.lock` to mean "Yarn".

This PR also adds a test for this situation in `fs-detectors`. The behavior is currently correct there, but was not tested initially. It is now to avoid future regressions.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -312,7 +312,12 @@ export async function scanParentDirs(
     ]);
 
   // Priority order is Yarn > pnpm > npm > bun
```
styfle, line 314: This comment looks like its no longer true

trek: True, but I'm not touching this PR and fighting flakey tests for another two days! Happy to fixup in a followup PR.

trek:
For reference, 
* according to `git diff`, https://github.com/vercel/vercel/pull/10596 is identical code, but with failing tests.
* https://github.com/vercel/vercel/pull/10600 is only this code change and changeset extracted, but with failing tests.

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
trek: https://github.com/vercel/vercel/pull/10603

PR #10586, opened 2023-09-25
[ci] use `--dry-run` to only run tests in packages that changed

styfle:
- Follow up to #10584

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -4,6 +4,7 @@
   "preferGlobal": true,
   "license": "Apache-2.0",
   "description": "The command-line interface for Vercel",
+  "//": "no op change to test turbo",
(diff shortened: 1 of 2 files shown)


styfle:
Closing since this didn't work as expected.

> There is a task in the graph which is a no-op, and it has a hash. Dry run simply queries the file system and the network to identify if that cache item is loadable. That is distinct from "will this task be effective?"
>
> Aside, "existence in cache" is a side-effect check. You're wanting to ask, "What was the outcome the last time I ran the task associated with this hash? If I've run it before and it exited 0, don't run it."
But we don't have a nice way to do that.

More context in this [slack thread](https://vercel.slack.com/archives/C03LMQZL205/p1695667273523879) discussion with the turbo team.

PR #10592, opened 2023-09-26

trek:
Just like #10591 but no code changes, no tests, just a new fixture.

🚫  Code Change
🚫 Test Change
✅  Fixture

Fails:
* `@vercel/next` 00-server-build › should update content for prerendered path correctly
* `@vercel/build-utils` Should build "22-pnpm"


trek:

PR #10593, opened 2023-09-26

trek:
Just like #10583 but no tests, not fixures:

✅ Code Change
🚫 Test Change
🚫  Fixture

Fails:
* `@vercel/next` 00-app-dir-middleware › should deploy and pass probe checks


trek:

PR #10596, opened 2023-09-26

trek:
Just like https://github.com/vercel/vercel/pull/10583/files but with a yarn.lock removed.


trek:

PR #10598, opened 2023-09-26
Only the fakest of fake codes. Literally nothing else.

trek:
Just like #10583 but no nothing. Just a fake code path with a non-existent  ENV var to trigger cache invalidation I guess.

🚫 Code Change
🚫 Test Change
🚫  Fixture

Fails:

* `@vercel/next` 00-middleware › should revalidate content correctly for optional catch-all route


trek:

PR #10600, opened 2023-09-26
Real code, real changelog

trek:
Just like #10583 but no tests, no fixtures, but real code:

✅ Code Change
🚫 Test Change
🚫 Fixture

Fails:

* `@vercel/next`: 00-app-dir-base-path › should deploy and pass probe checks


trek:

PR #10606, opened 2023-09-26
[tests] update contributing.md test commands

EndangeredMassa:
The instructions in the `.github/CONTRIBUTING.md` around testing are out of date. This PR clarifies the commands and tokens required.


.github/CONTRIBUTING.md
```diff
@@ -16,11 +16,16 @@ cd vercel
 corepack enable
 pnpm install
 pnpm build
-pnpm lint
-pnpm test-unit
 ```
 
-Make sure all the tests pass before making changes.
+To run the tests, we'll need a `VERCEL_TOKEN`. To get that token:
```
styfle:
```suggestion
To run the e2e tests, you'll need a `VERCEL_TOKEN`. To get that token:
```

Pretty sure `test-unit` works without a token

.github/CONTRIBUTING.md
```diff
@@ -16,11 +16,16 @@ cd vercel
...
+
+- Option 1: Login to https://vercel.com, go to Settings, Tokens, then create a "Full Account" token to use for this purpose.
+- Option 2: If you've logged in with the `vercel` CLI before, you can find the token value in `~/Library/Application Support/com.vercel.cli/auth.json`.
```
styfle: This will be in a different location for windows. Probably best to omit option 2 here since its not advised to share the same token you use for development and testing

.github/CONTRIBUTING.md
```diff
@@ -16,11 +16,16 @@ cd vercel
...
+
+Save that token or export it in your terminal profile.
+
+Then run the tests with `pnpm turbo run test`.
```
styfle: This will only run a couple tests, I dont think this is what you want

https://github.com/vercel/vercel/blob/873327e9bdb56be5dbfcf3ca45308d9f8308debc/package.json#L42

styfle: Also you need to run the `gen.js` script to set the proper cache key

.github/CONTRIBUTING.md
cb1kenobi: On Windows the path is `%APPDATA%\com.vercel.cli\Data\auth.json`. `%APPDATA%` typically resolves to `C:\Users\<username>\AppData\Roaming`.

.github/CONTRIBUTING.md
cb1kenobi: I wasn't aware of `gen.js`. I've never run it. Is it something that needs to be manually run?

styfle: Every script runs it before turbo

https://github.com/vercel/vercel/blob/873327e9bdb56be5dbfcf3ca45308d9f8308debc/package.json#L45

styfle: - See https://github.com/vercel/vercel/pull/8830

.github/CONTRIBUTING.md
```diff
@@ -35,14 +41,14 @@ See [CLI Local Development](../packages/cli#local-development) for more details.
 
 ## Verifying your change
 
-Once you are done with your changes (we even suggest doing it along the way), make sure all the tests still pass by running:
+Once you are done with your changes (we even suggest doing it along the way), make sure the relevant tests still pass by running the relevant tests. Example:
```
styfle:
```suggestion
Once you are done with your changes (we even suggest doing it along the way), make sure the relevant tests still pass by running them. Example:
```

Nit pick: avoid "relevant tests" twice in the same sentence

.github/CONTRIBUTING.md
```diff
@@ -35,14 +41,14 @@ See [CLI Local Development](../packages/cli#local-development) for more details.
...
 
 ```
-pnpm test-unit
+# from root of this project
+cd packages/cli
+pnpm test test/integration-1.test.ts
```
styfle, line 52: Shouldn't turbo run the relevant tests automatically?

I think this should mention you are only running a specific test file.

And you can run from the root to run all relevant tests.

.github/CONTRIBUTING.md
```diff
@@ -16,11 +16,17 @@ cd vercel
 corepack enable
 pnpm install
 pnpm build
-pnpm lint
```
styfle, line 19: Why not lint locally?

.github/CONTRIBUTING.md
```diff
@@ -16,11 +16,17 @@ cd vercel
...
-Make sure all the tests pass before making changes.
+To run the e2e tests, you'll need a `VERCEL_TOKEN`. To get that token, login to https://vercel.com, go to Settings, Tokens, then create a "Full Account" token to use for this purpose.
+
+Save that token or export it in your terminal profile.
+
+```
+export VERCEL_TOKEN="my-copied-token"
+```
+
+Then run the tests with `pnpm turbo run test-e2e`.
```
styfle, line 30: We should also put a big warning here that this will create a ton of deployments and it might even get rated limited by hobby account limits if run in parallel.

styfle, line 30:
```suggestion
Then run the tests with `pnpm test-e2e`.
```

Also need to run the gen script

.github/CONTRIBUTING.md
EndangeredMassa: It takes a fair bit of time, should always succeed, and if it does fail it's not something the contributor should have to fix.

It does help the contributor be sure their local state is good before moving forward, that's true. I'll put it back--people often skip these kinds of steps anyway and we can leave it up to them.

PR #10611, opened 2023-09-27
Add experimental field for fallback prefix metadata

nabsul:
Adds an experimental field for marking the length of metadata prefixed on the fallback content.


packages/build-utils/src/prerender.ts
```diff
@@ -165,5 +168,15 @@ export class Prerender {
       }
       this.experimentalStreamingLambdaPath = experimentalStreamingLambdaPath;
     }
+
+    if (experimentalFallbackMetadataPrefixLength !== undefined) {
+      if (typeof experimentalFallbackMetadataPrefixLength !== 'number') {
+        throw new Error(
+          'The `experimentalFallbackMetadataPrefixLength` argument for `Prerender` must be a string.'
```
cb1kenobi: The error says the value must be a string, but the conditional above says it must be a number.

```suggestion
          'The `experimentalFallbackMetadataPrefixLength` argument for `Prerender` must be a number.'
```

cb1kenobi (changes requested):

nabsul:
@cb1kenobi Thanks for the review. I'm going to put this PR back into a draft because I'm starting to think we can live without this change... Sorry for the churn :-)

PR #10535, opened 2023-09-18
[cli] Support northstar users

chibicode:
Supports users who have `version === 'northstar'` and `defaultTeamId`.

PR #10612, opened 2023-09-27
[cli] Restore setting of `currentTeam`

trek:
As part of a refactor in #10535, we lost the – admittedly untested – behavior of 

1. setting the new `teamId` from a login `result`
2. removing this value if it was `undefined` ensuring we didn't write `"currentTeam": undefined` to `config.json`.

This resulted in subsequent calls to `vc login` keeping the last known defined `teamId`.

This should restore that behavior but also retain the behavior of calling `updateCurrentTeamAfterLogin` to check if the user has a new `defaultTeamId` to be stored.

Not a fan of just how much mutation exists on `client.config` in this flow, but that predates this change and I wanted to keep this bugfix as small as possible.


packages/cli/src/commands/login/index.ts
```diff
@@ -68,14 +71,24 @@ export default async function login(client: Client): Promise<number> {
...
-  const isNewLogin = !client.authConfig.token;
-
   // Save the user's authentication token to the configuration file.
   client.authConfig.token = result.token;
 
-  // If we have a new login, update `currentTeam`
-  if (isNewLogin) {
-    await updateCurrentTeamAfterLogin(client, output, result.teamId);
+  // Save the user's team (possibly `undefined`) to the global configuration file.
+  client.config.currentTeam = result.teamId;
```
trek: This handles both setting it to a new value _and_ unsetting the old value. We later `delete` this key entirely if it remains `undefined`

packages/cli/src/commands/login/index.ts
```diff
@@ -68,14 +71,24 @@ export default async function login(client: Client): Promise<number> {
...
+
+  // Delete undefined `currentTeam` key to avoid writing key with no value to JSON file.
+  if (!client.config.currentTeam === undefined) {
```
cb1kenobi: The `!` seems unnecessary?

```suggestion
  if (client.config.currentTeam === undefined) {
```

packages/cli/src/commands/login/index.ts
cb1kenobi: Maybe this could be more readable by using an if/else?

```
if (result.teamId) {
  client.config.currentTeam = result.teamId;
} else {
  delete client.config.currentTeam;
}
```

cb1kenobi (approved):
LGTM

PR #10619, opened 2023-09-28
[cli] Move from yarn to npm for fixtures and ensure lockfiles to speed up installs

trek:
Some of our tests take quite some time to run and occasionally timeout when accessing https://yarnpkg.com/. We decided to make sure there is a lockfile for this _and_ just move over to `npm`. Shifting the `packages/cli` fixtures over in this PR.

I possibly missed one but I changed the install commands in tests and ran

* pnpm test
* pnpm test-unit
* pnpm test-dev
* pnpm test-e2e


packages/cli/test/dev/utils.js
```diff
@@ -144,7 +144,13 @@ async function exec(directory, args = []) {
 
 async function runNpmInstall(fixturePath) {
   if (await fs.pathExists(join(fixturePath, 'package.json'))) {
-    await execa('yarn', ['install'], {
+    let command;
+    if (await fs.pathExists(join(fixturePath, 'package-lock.json'))) {
+      command = 'npm';
+    } else {
+      command = 'yarn';
```
trek, line 151: We have some packages that simply _will not_ install and pass tests with `npm`. Letting these be for now as we're updating our test matrix of frameworks currently.

trek: Specifically `23-docusaurus` fails to install via npm.

trek: I even tried a `yarn.lock`->`package-lock.json` converter but it fails. https://github.com/vercel/vercel/actions/runs/6406470416/job/17391232739?pr=10655

trek: Even pining it to 2.0 causes `npm install` timeouts. https://github.com/vercel/vercel/actions/runs/6406652552/job/17391833595?pr=10655

packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/package.json
```diff
@@ -28,7 +28,8 @@
     "@storybook/nextjs": "^7.0.6",
     "@storybook/react": "^7.0.6",
     "@storybook/testing-library": "^0.0.14-next.2",
-    "storybook": "^7.0.6"
+    "storybook": "^7.0.6",
+    "esbuild": "0.19.2"
```
cb1kenobi, line 32: Is this needed? I don't recall this fixture explicitly needing `esbuild`.

packages/cli/test/dev/utils.js
cb1kenobi: The `23-docusaurus` fixture is using `@docusaurus/core@2.0.0-alpha.48` which is 4 years old. Yikes.

I updated to `@docusaurus/core@2.4.3` and `@docusaurus/preset-classic@2.4.3` and was able to `npm i` and generate the `package-lock.json`. Could still timeout during tests. Who knows?

packages/cli/test/fixtures/unit/commands/build/storybook-with-middleware/package.json
trek: Nothing else would resolve peer dependencies correctly without a `--force`

packages/cli/test/dev/utils.js
trek: The PR from the aforementioned run does indeed have a lockfile. https://github.com/vercel/vercel/pull/10655/. It's the test timeouts at `npm install` that is the mystery.

EndangeredMassa (approved):

PR #10622, opened 2023-09-29
Move full GitHub Actions expression into curlies in test workflow

trek:
The preferred version of this syntax is to enclose everything in `${{}}` on GHA.

With no cache hits (just the workflow change):
CleanShot 2023-09-29 at 08 12 37@2x

With cache hit (tiny change to trigger):
CleanShot 2023-09-29 at 08 11 53@2x


cb1kenobi (approved):
Makes sense to me!

PR #10623, opened 2023-09-29
[cli] indicates whether @vercel/speed-insights or @vercel/analytics are used

feugy:
### 🧐 What's in there?

When deploying to Vercel, the build container needs to know whether the deployed apps is using `@vercel/speed-insights` and `@vercel/analytics` or not.

As discussed on slack, using `builds.json` is a nice way to convey this piece of information.


packages/cli/test/fixtures/unit/commands/git/connect/unlinked/git/HEAD
Ethan-Arrowood, line 1: why are these removed in this PR?

Ethan-Arrowood (commented):
lgtm other than than the couple fixture files that are deleted. Not sure how that's related

Ethan-Arrowood:
don't forget to add a Changeset with `npx changeset`

styfle (changes requested):
Please restore the deleted test fixture 

```
packages/cli/test/fixtures/unit/commands/git/connect/unlinked
```

packages/cli/test/integration-2.test.ts
```diff
@@ -1165,6 +1165,11 @@ test('[vc build] should build project with `@vercel/speed-insights`', async () =
     expect(output.stderr).toContain(
       'The `VERCEL_ANALYTICS_ID` environment variable is deprecated and will be removed in a future release. Please remove it from your environment variables'
     );
+    console.log(output.stderr, output.stdout);
```
EndangeredMassa:
```suggestion
```

packages/cli/test/fixtures/unit/commands/git/connect/unlinked/git/HEAD
feugy: Apologies, that was no intentional indeed

packages/cli/test/helpers/prepare.js
```diff
@@ -433,6 +433,24 @@ module.exports = async function prepare(session, binaryPath, tmpFixturesDir) {
...
+          framework: null,
+          installCommand: 'echo "skipping install"',
+        },
+      }),
+      'package.json': JSON.stringify({
+        scripts: {
+          build: 'mkdir -p public && echo hi > public/index.txt',
+        },
+        dependencies: {
+          '@vercel/web-analytics': '0.0.1',
```
tobiaslins: the dependency here is wrong. its `@vercel/analytics`

packages/cli/src/commands/build/index.ts
```diff
@@ -807,3 +811,21 @@ function mergeFlags(
...
+}
+
+function isUsingWebAnaytics(pkg: PackageJson | null): boolean {
+  const dependencies = [
+    ...Object.keys(pkg?.dependencies ?? {}),
+    ...Object.keys(pkg?.devDependencies ?? {}),
+  ];
+
+  return dependencies.some(d => d === '@vercel/analytics');
+}
```
tobiaslins: maybe we can abstract away checking if a package is installed
```suggestion
function isPackageInstalled(name: string; pkg: PackageJson | null) {
  const dependencies = [
    ...Object.keys(pkg?.dependencies ?? {}),
    ...Object.keys(pkg?.devDependencies ?? {}),
  ];

  return dependencies.some(d => d === name);
}

function isUsingSpeedInsights(pkg: PackageJson | null): boolean {
  return isPackageInstalled('@vercel/speed-insights', pkg);
}

function isUsingWebAnaytics(pkg: PackageJson | null): boolean {
   return isPackageInstalled('@vercel/analytics', pkg);
}
```

tobiaslins (dismissed):
Overall great! the small proposed refactor would be nice!

tobiaslins (dismissed):

Ethan-Arrowood (dismissed):

packages/cli/src/commands/build/index.ts
```diff
@@ -807,3 +811,20 @@ function mergeFlags(
     return [];
   });
 }
+
+function isPackageInstalled(name: string, pkg: PackageJson | null): boolean {
+  const dependencies = [
+    ...Object.keys(pkg?.dependencies ?? {}),
+    ...Object.keys(pkg?.devDependencies ?? {}),
+  ];
```
styfle: Right now these packages are iterated twice because `isPackageInstalled()` is always called twice. 

Let's avoid that by hoisting out of this function (thats basically how it was working before)

You could even use a Set to avoid `.some()` iteration again.

styfle: For example:

```js
const dependencies = new Set();
for (const key in pkg?.dependencies ?? {}) {
  dependencies.add(key);
}
for (const key in pkg?.devDependencies ?? {}) {
  dependencies.add(key);
}


if (dependencies.has('@vercel/speed-insights')) {
  // do stuff
}

if (dependencies.has('@vercel/analytics')) {
  // do stuff
}
```

feugy: Good call. @chriswdmr suggested that we convey the package versions instead of a boolean.
I made it a map, and it's built only once

feugy: In the unlikely event of the same package being in both `devDependencies` and `dependencies`, the later will take over.

styfle (approved):

packages/cli/src/commands/build/index.ts
```diff
@@ -106,6 +106,7 @@ export interface BuildsManifest {
   argv: string[];
   error?: any;
   builds?: SerializedBuilder[];
+  features?: { speedInsightsVersion: string; webAnalyticsVersion: string };
```
cb1kenobi, line 109: @feugy FYI, this line is failing type checking: https://github.com/vercel/vercel/actions/runs/6421634817/job/17436437058?pr=10667#step:9:57

Fixed in my PR: https://github.com/vercel/vercel/pull/10667/files#diff-452a115a951a23488dd3257a83934cc9439918fbd41143a0ded219164dd13c6f

PR #10630, opened 2023-10-02
[next] fix lambda creation for i18n edge pages

csabapalfi:
An issue similar to https://github.com/vercel/next.js/issues/44381 exists whereby vercel creates lambda functions for edge pages when i18n is configured.

Fixes https://github.com/vercel/next.js/issues/42854

PR #10631, opened 2023-10-02
[next] Correct output file tracing and limit calculation

ijjk:
While investigating build times noticed that our lambda creation times were increasing linearly with the number of pages which is unexpected since there are mostly shared dependencies. After further investigation it seems we were falling back to our legacy manual `nodeFileTrace` calls in the builder when we shouldn't have been.

Also noticed we were still doing the un-necessary reading/calculating for uncompressed lambda sizes which is as discussed previously isn't required and the only limit we need to keep enforcing is the uncompressed size which is a lot less expensive to compute. 

As a further optimization this adds proper usage of our lstat cache/sema where it wasn't previously and proper parallelizing where applicable. 

These changes reduce tracing/lambda creation times by 3-5x in larger applications and can fix edge cases where we weren't leveraging our more accurate traces from the build. 

Before: 

```sh
Traced Next.js server files in: 444.05ms
Created all serverless functions in: 1:36.311 (m:ss.mmm)
Collected static files (public/, static/, .next/static): 241.47ms
```

After: 

```sh
Traced Next.js server files in: 10.4ms
Created all serverless functions in: 43.684s
Collected static files (public/, static/, .next/static): 250.828ms
```


cb1kenobi:
This PR is awesome.

ztanner (dismissed):

styfle (dismissed):

packages/next/src/utils.ts
```diff
@@ -2567,10 +2512,13 @@ export function normalizeEdgeFunctionPath(
...
       shortPath.replace(/(^|\/)(page|route)$/, '')
     ).replace(/^\//, '');
+  }
 
-    if (!shortPath || shortPath === '/') {
-      shortPath = 'index';
-    }
+  if (shortPath.startsWith('pages/')) {
+    shortPath = shortPath.replace(/^pages\//, '');
+  }
```
styfle, line 2519: This looks similar to https://github.com/vercel/vercel/pull/10630 but I don't see a test for this new behavior

```suggestion
```

This can be removed now that its on line 2524

styfle (approved):

PR #10637, opened 2023-10-02
[static-build] Add fixture for svelte 4 apps

trek:
Adds Svelte 4 fixture for testing as we expand our test matrix. This is an updated version of `11-svelte`


cb1kenobi (approved):
LGTM

PR #10648, opened 2023-10-03
[python] Add support for pip3.10 and pip3.11 in `vc build`

cb1kenobi:
The Python builder explicitly lists `pip3.6` and `pip3.9`, however newer versions of macOS (tested Ventura 13.5.2) does not have either of those versions, but it does have `pip3.10` and `pip3.11`. Adding these newer versions so that Python functions work again.

This PR also validates that `python` and `pip` are actually installed. This is nice because if the pipfile doesn't specify a specific version, it'll find the latest supported one, or throw if no supported versions found. This way `vc build` will work for users with newer installs and existing environments (e.g. build container) can upgrade when convenient.

Note that this change will only affect `vc build` and `vc deploy --prebuilt`. It does not affect `vc deploy` or `git push` deployments. The build container only supports Python 3.9, which is the only version which will be selected, regardless the required version in the `Pipfile`.


trek (dismissed):

Ethan-Arrowood (dismissed):

styfle (commented):
Did you add these versions to the build container?

Also can you add a test to ensure this works as expected?

cb1kenobi:
I did not add these to the build container. I was testing `vc dev` and `vc build`. I will take a peek at how to add them to the build container soon.

As for tests, I'm not sure how to add tests for this. Do we have tests for `pip3.6` and `pip3.9`?

styfle:
> I did not add these to the build container.

This is critical to land first or else it will likely break deployments.

> Do we have tests for pip3.6 and pip3.9?

I believe the version is defined in the `Pipfile.lock` See this old PR for example test fixtures: 

- https://github.com/vercel/vercel/pull/8300

cb1kenobi:
@styfle I've updated the code to check if python/pip is actually installed. This will allow build container to continue to use 3.9 and can be updated whenever.

packages/python/test/unit.test.ts
```diff
@@ -13,12 +22,16 @@ beforeEach(() => {
 afterEach(() => {
   console.warn = originalConsoleWarn;
   global.Date.now = realDateNow;
+  process.env.PATH = origPath;
+  if (fs.existsSync(tmpPythonDir)) {
+    fs.removeSync(tmpPythonDir);
+  }
 });
 
 it('should only match supported versions, otherwise throw an error', async () => {
```
styfle: I think this test name needs to be updated

packages/python/test/unit.test.ts
```diff
@@ -35,23 +48,33 @@ it('should ignore minor version in vercel dev', async () => {
 });
 
 it('should select latest version when no Piplock detected', async () => {
```
styfle: This test name needs to be updated if its no longer selecting a specific version

packages/python/test/unit.test.ts
```diff
@@ -35,23 +48,33 @@ it('should ignore minor version in vercel dev', async () => {
...
-  expect(
-    getSupportedPythonVersion({ pipLockPythonVersion: undefined })
-  ).toHaveProperty('runtime', 'python3.9');
+  const result = getSupportedPythonVersion({ pipLockPythonVersion: undefined });
+  expect(result).toHaveProperty('runtime');
+  expect(result.runtime).toMatch(/^python3\.\d+$/);
   expect(warningMessages).toStrictEqual([]);
 });
 
 it('should select latest version and warn when invalid Piplock detected', async () => {
```
styfle: same here

packages/python/src/version.ts
```diff
@@ -90,3 +116,10 @@ function isDiscontinued({ discontinueDate }: PythonVersion): boolean {
   const today = Date.now();
   return discontinueDate !== undefined && discontinueDate.getTime() <= today;
 }
+
+function isInstalled({ pipPath, pythonPath }: PythonVersion): boolean {
+  return (
+    !!which.sync(pipPath, { nothrow: true }) &&
+    !!which.sync(pythonPath, { nothrow: true })
```
styfle:
```suggestion
    Boolean(which.sync(pipPath, { nothrow: true })) &&
    Boolean(which.sync(pythonPath, { nothrow: true }))
```

Nit pick: this will avoid the double negative. I'm also thinking this should be async parallel instead with `await Promise.all()`

styfle:
> This will allow build container to continue to use 3.9 and can be updated whenever.

@cb1kenobi Its not just build container. The `runtime` that is used must match the version of the build container. Please update the build container first before landing this PR.

packages/python/src/version.ts
```diff
@@ -42,7 +56,16 @@ export function getLatestPythonVersion({
   if (isDev) {
     return getDevPythonVersion();
   }
```
styfle, line 58: > This way local dev with vc dev will work for users with newer installs and existing environments 

The PR description says this but `vc dev` should be hitting the code below, it should hit the `isDev` code and call `getDevPythonVersion()`

packages/python/test/unit.test.ts
cb1kenobi: I think the name is fine, but the test itself was not great and I have fixed it.

packages/python/test/unit.test.ts
cb1kenobi: Fixed.

packages/python/test/unit.test.ts
cb1kenobi: Fixed.

packages/python/src/version.ts
cb1kenobi: I've replaced the `!!` with a `Boolean()`, but I don't really agree. Calling the Boolean constructor costs more than a simple double negation.

I don't think it's not worth parallelizing these two stats. Using `Promise.all()` would require `isInstalled()` to return the promise and since `isInstalled()` is being called from a `Array.prototype.find()`. I'd have to replace the `find()` with more code and a for loop. With all that overhead and synchronization, two stat calls doesn't seem half bad.

packages/python/src/version.ts
cb1kenobi: I stand corrected. `vc dev` is fine. `vc build` is the issue:

```
$ vc build
Vercel CLI 32.4.1
Installing required dependencies...
Failed to run "pip3.9 install --disable-pip-version-check --target . werkzeug==1.0.1"
Error: spawn pip3.9 ENOENT
```

cb1kenobi:
> > This will allow build container to continue to use 3.9 and can be updated whenever.
> 
> @cb1kenobi Its not just build container. The `runtime` that is used must match the version of the build container. Please update the build container first before landing this PR.

@styfle do you mean the Lambda runtime? AWS Lambda supports 3.11, 3.10, 3.9, 3.8, and 3.7 (varies by OS).

If build container is using 3.9 and AWS Lambda supports 3.9 and my changes look for the newest installed version and ultimately choose 3.9, then what's the problem? Why does build container need to be updated?

styfle:
> Why does build container need to be updated?

@cb1kenobi The python version used at build time must match the lambda runtime version.

> and my changes look for the newest installed version and ultimately choose 3.9

Does it correctly error if the user has 3.10 or 3.11 in their lockfile? Can you add an e2e test for that?

cb1kenobi:
@styfle it currently displays a warning and uses the latest supported version.

```
      console.warn(
        `Warning: Python version "${pipLockPythonVersion}" detected in Pipfile.lock is invalid and will be ignored. http://vercel.link/python-version`
      );
```

Personally, I think it should throw and exit. If you say use Python 3.10 and 3.10 is not installed, that dependency is not fulfilled and thus do not proceed with the deployment. I'd argue the current behavior is a bug. Changing this behavior may be a breaking change.

How do you feel about replacing the warning with a throw and possibly major rev the `python` builder?

/cc @trek

styfle:
> Changing this behavior may be a breaking change.

Agreed, it would be a breaking change. We probably don't want to do that.

packages/python/src/version.ts
styfle: Can you update the PR description?

styfle:
The PR title and description should probably mention this is only `vc build && vc deploy --prebuilt` and wont work for normal deployments using `vc deploy` or `git push`.

cb1kenobi:
Just tested this build and it works as expected. I have a `Pipfile` requesting Python 3.10, but since it's not found, it prints:

```
Warning: Python version "3.10" detected in Pipfile.lock is invalid and will be ignored. http://vercel.link/python-version
```

Then continues to use Python 3.9:

```
Running "pip3.9 install --disable-pip-version-check --target . pipfile-requirements==0.3.0 --no-warn-script-location"...
```

At runtime, it uses Python 3.9: https://pythonfun-mwqxabuue-chrisbarber.vercel.app/api

styfle (approved):

trek (approved):

PR #10656, opened 2023-10-04
[static-build] Add fixtures for latest versions of tested frameworks

trek:
Adds `preact@10`, `eleventy@2` and `jekyll@4` fixtures.


packages/static-build/test/fixtures/eleventy-2/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "homepage": "https://github.com/11ty/eleventy-base-blog#readme",
+  "devDependencies": {
+    "@11ty/eleventy": "^2.0.0",
+    "@11ty/eleventy-plugin-rss": "^1.2.0",
+    "@11ty/eleventy-plugin-syntaxhighlight": "^5.0.0",
+    "luxon": "^1.12.0",
+    "markdown-it": "^8.4.2",
+    "markdown-it-anchor": "^5.0.2"
+  }
+}
```
EndangeredMassa, line 32: **suggestion (soft-blocking):** add lockfile

In this or a follow-up PR: Can we add a lockfile for this fixture?‏

EndangeredMassa (dismissed):

packages/static-build/test/fixtures/eleventy-2/package.json
trek: I was mystified by why it didn't have one, but its own generator maybe wants to ignore them? https://github.com/vercel/vercel/pull/10656/commits/dc7d64a117dc87fb6d0c9eeae4c7c0b8ce8a2422#diff-1ea7fca44fbf9989e209c50074eab7e51e440fb5dc53beed3e5a5a26a17b9632

Added now

Ethan-Arrowood (changes requested):
In the future I'd prefer we do one fixture per PR. Or at least one framework per pr if there are multiple fixtures for a single framework. 

Please go through and remove anything unnecessary from the fixtures that might slow them down. Images are the biggest thing. They should all be deleted (including svgs). CSS is another but less of a big deal. Each fixture should be testing the most minimal JS possible for that given framework to render an `index` route.

trek:
@Ethan-Arrowood 

> In the future I'd prefer we do one fixture per PR.

Me too, but there was one card and I did PRs for each https://github.com/vercel/vercel/pull/10661, https://github.com/vercel/vercel/pull/10657, and https://github.com/vercel/vercel/pull/10658. Merged them here when each framework got split into its own card so this could get merged before starting in on those.

> Please go through and remove anything unnecessary from the fixtures that might slow them down. Images are the biggest thing. They should all be deleted (including svgs).

This isn't our current pattern in any other fixture in this package. I agree we should minimize but it isn't always as easy as "delete some files" depending on how each framework manages files. Some reference them as component imports and you have to also make changes to app code.

It seemed like our current objective was quickly fleshing out fixture coverage to be more confident in releases but we can take a slower approach to creating new minimal fixtures as a team decision. The full list of frameworks has ~40 items on it.

Ethan-Arrowood:
Slimming down these fixtures was what I did to improve their timing. It is the pattern for many existing fixtures too such as the gatsby ones. If we don't do this now, we'll most likely have to do it in the future when we start asking ourselves, "why is it taking so long to test all these different frameworks?"

Ethan-Arrowood (approved):
I'll approve so that I'm not blocking this from going through, but performance of these tests shouldn't be an after thought. That's how we got to 15 minutes per static-build test run in the first place.

PR #10660, opened 2023-10-04
Add vue-3

trek:
Adds a fixture test for vue@3


packages/static-build/test/fixtures/vue-v2/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "13-vue",
```
EndangeredMassa, line 2: Do you want to change the name here too? We could update this all later or just leave it.

EndangeredMassa (approved):

Ethan-Arrowood (approved):

PR #10666, opened 2023-10-05
Fix regression: traverse parent folders as long as the path matches

joelvh:
Addresses regression in #10305 related to my [comments](https://github.com/vercel/vercel/pull/10305/files#r1346756667).

This update will assume the asset build path and the public path are related, as long as the end of the paths match. This _should_ address the use case in #10305 as well as the original intent of `publicPath` in Remix, which is to specify a path in the browser to an asset -- unrelated to the filesystem.

An example might be:
- `assetsBuildDirectory: /project/public/build/`
- `publicPath: /apps/my-app/build/`

The end of those paths match `/build/` and would result in the Remix build script traversing to the parent `/project/public` folder on the filesystem and regard that as the root.

Given, this might be a contrived example, but I would presume this addresses the use case in #10305 as well as preserves the original intent of the `publicPath`.

### DIFF
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -428,14 +428,20 @@ module.exports = config;`;
       : null,
   ]);
 
-  const staticDir = join(
-    remixConfig.assetsBuildDirectory,
-    ...remixConfig.publicPath
-      .replace(/^\/|\/$/g, '')
-      .split('/')
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      .map(_ => '..')
-  );
+  const assetsBuildDirectory = remixConfig.assetsBuildDirectory
+    .replace(/^\/|\/$/g, '')
+    .split('/');
+  const publicPath = remixConfig.publicPath.replace(/^\/|\/$/g, '').split('/');
+
+  // Remove matching segments at end of paths only (e.g. `/project/build/` compared to `/public/build/`)
+  while (
+    assetsBuildDirectory.length &&
+    assetsBuildDirectory[assetsBuildDirectory.length - 1] === publicPath.pop()
+  ) {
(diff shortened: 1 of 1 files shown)


cb1kenobi:
Thanks @joelvh! I'm actively investigating this. Sit tight.

joelvh:
Thanks @cb1kenobi. Regarding identifying an alternative to `publicPath`, how is what you're looking for different from the Output Directory path in the Vercel project?

packages/remix/src/build.ts
```diff
@@ -428,14 +428,20 @@ module.exports = config;`;
...
-    ...remixConfig.publicPath
-      .replace(/^\/|\/$/g, '')
-      .split('/')
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      .map(_ => '..')
-  );
+  const assetsBuildDirectory = remixConfig.assetsBuildDirectory
+    .replace(/^\/|\/$/g, '')
+    .split('/');
+  const publicPath = remixConfig.publicPath.replace(/^\/|\/$/g, '').split('/');
```
cb1kenobi, line 434: I tested this last week and if there's no `publicPath`, the path does not start with a `/` (or `C:\`). This leads me to wonder if there's a better way to use `path.relative()` and `path.join()`. I'll be in touch.

cb1kenobi:
Here's what I've come up with: https://github.com/vercel/vercel/pull/10685. I need to add some tests with various `publicPath` values and see if my change fixes things, or rather doesn't break anything. :)

joelvh:
> Here's what I've come up with: #10685. I need to add some tests with various `publicPath` values and see if my change fixes things, or rather doesn't break anything. :)

Very elegant!

PR #10667, opened 2023-10-05
Add `type-check` and unify tsconfig

cb1kenobi:
This adds a new `pnpm type-check` that leverages `turbo` to validate the TypeScript code. This can be run at the top-level or for an individual package.

The `test-lint` workflow will run it after linting and doing the prettier check.

As apart of this effort, each package's `tsconfig.json` has been simplified. There's a new top-level `tsconfig.base.json` file that extends the Vercel Style Guide for TypeScript. Each package's `tsconfig.json` has been audited and previously suppressed rules that no longer apply have been removed. The result is each package's `tsconfig.json` is greatly simplified and we can control common settings in the base config while keeping the flexibility of package-level overrides.

Lastly, in `package/cli`, `pnpm build` calls `scripts/build.mjs` which calls `scripts/compile-templates.mjs`. The `compile-templates.mjs` file was generating invalid TypeScript code. I've fixed it and now it's happier than ever.

Note: In order to run `pnpm type-check`, you must first `pnpm build` because we need the `.d.ts` definition files.


utils/build.mjs
```diff
@@ -59,7 +59,7 @@ export async function tsc() {
   const rootNodeModulesBin = fileURLToPath(
     new URL('../node_modules/.bin', import.meta.url)
   );
-  await execa('tsc', {
+  await execa('tsc', ['--declaration', '--emitDeclarationOnly'], {
```
styfle, line 62: Can these be moved to `tsconfig.base.json` instead?

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
+{
+  "description": "The base configuration for all packages.",
+  "extends": "@vercel/style-guide/typescript",
+  "compilerOptions": {
+    "lib": ["ES2021"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noEmitOnError": true,
+    "noImplicitReturns": true,
```
styfle, line 9: How about adding `noFallthroughCasesInSwitch: true` since most of the others used that too?

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
+{
+  "description": "The base configuration for all packages.",
+  "extends": "@vercel/style-guide/typescript",
+  "compilerOptions": {
```
styfle, line 4: How about adding `outDir: "./dist"` since it looks like all of them use it?

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
...
+    "lib": ["ES2021"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noEmitOnError": true,
+    "noImplicitReturns": true,
+    "noUncheckedIndexedAccess": false,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "sourceMap": false,
+    "strictNullChecks": true,
```
styfle:
```suggestion
    "strict": true,
```

Lets enable all strict features like we had before

tsconfig.base.json
```diff
@@ -0,0 +1,17 @@
+{
+  "description": "The base configuration for all packages.",
+  "extends": "@vercel/style-guide/typescript",
```
styfle, line 3: Let's make sure @Ethan-Arrowood reviews this since there were changes coming to the style guide:

- https://github.com/vercel/style-guide/pull/60

utils/build.mjs
cb1kenobi: No because the `type-check` scripts all run `tsc --noEmit` which cannot be combined with `emitDeclarationOnly`. This means we need to explicitly pass in these flags given the context.

tsconfig.base.json
cb1kenobi: It's already in the style guide base config: https://github.com/vercel/style-guide/blob/canary/typescript.json#L6. :)

tsconfig.base.json
cb1kenobi: It's already in the style guide base config: https://github.com/vercel/style-guide/blob/canary/typescript.json#L9. :)

tsconfig.base.json
cb1kenobi: TypeScript is funny with relative paths. If I put `outDir` in the base config, it will output all `.js` files into a top-level `dist` directory and it doesn't take long before the build fails because a dependent package built files can't be found. We need to leave `outDir` in each package's `tsconfig.json`.

package.json
```diff
@@ -144,5 +145,8 @@
         }
       }
     ]
+  },
+  "dependencies": {
+    "@vercel/style-guide": "4.0.2"
   }
```
Ethan-Arrowood: This should probably be a dev dep right? I guess it doesn't matter as its the root package.json

packages/cli/tsconfig.json
```diff
@@ -1,20 +1,16 @@
...
-    "module": "commonjs",
-    "target": "ES2021",
-    "esModuleInterop": true,
     "allowJs": true,
-    "lib": ["ES2021"],
-    "resolveJsonModule": true,
-    "sourceMap": false,
+    "noImplicitReturns": false,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
Ethan-Arrowood, line 6: So these are set to `false` because our `cli` code isn't compatible with the strictness of `@vercel/style-guide/typescript`, right? 

We should track this in Linear as a follow up to improve our code quality to match the strictness of the style-guide. Sorta like API

packages/node/tsconfig.json
```diff
@@ -1,15 +1,12 @@
...
-    "esModuleInterop": true,
     "lib": ["ES2021", "DOM"],
-    "target": "ES2021",
     "module": "nodenext",
-    "outDir": "dist",
-    "sourceMap": false,
-    "declaration": false,
-    "skipLibCheck": true,
-    "moduleResolution": "nodenext"
+    "moduleResolution": "nodenext",
```
Ethan-Arrowood, line 5: I prefer `node16` over `nodenext` as its clearer that it won't magically change in the future

tsconfig.base.json
Ethan-Arrowood: Until that PR is merged and released (in probably a new major) we are fine to stick to the existing config. It is good enough afaik.

tsconfig.base.json
Ethan-Arrowood: I prefer this is omitted from the base config. 

We could have a `tsconfig.build.json` that assigns this property and others (like `declaration`). This way the `.build.json` can include _just_ `src/` files and then the default config can include _everything_ (i.e. `src/`, `test/`, etc.).

Ethan-Arrowood (dismissed):
Looking good. I don't think I have any blocking changes.

tsconfig.base.json
cb1kenobi: I didn't want to touch anything having to do with paths (e.g. `include`, `exclude`, `outDir`) because it never works in monorepos. All of those relative paths are being resolved relative to the `tsconfig` file and that's not what we want.

tsconfig.base.json
cb1kenobi: Removed `strictNullChecks`.

package.json
cb1kenobi: Good catch! Fixed in fb26f5b61c7ea6b0d2886652672a52f74c9b4cea.

packages/cli/tsconfig.json
cb1kenobi: Yep.

packages/node/tsconfig.json
cb1kenobi: I dunno what exactly the differences are between `node16` and `nodenext`. Given I didn't change the value, I'm leaning more towards not changing it in this PR. I'd really like to not break anything, though I doubt it would break anything. @styfle what do you think?

tsconfig.base.json
Ethan-Arrowood: Another great point

packages/node/tsconfig.json
Ethan-Arrowood: fair enough. we can make another pass on this in the future

utils/build.mjs
styfle: Ok make sense. And it seems like `--declaration` is here because we want it for builders but we don't want it for cli 👍

styfle (approved):
Looks good!

In a future PR, it would be nice to set `noUncheckedIndexedAccess: true` to follow the style guide.

Ethan-Arrowood (approved):

ecklf (approved):
Approving to unblock compute codeowners requirement.

PR #10681, opened 2023-10-10
Show Env Var Comments on vc ls/pull

JohnPhamous:
Updates `vc env pull`/`vc env ls` to show env var comments.

- Depends on https://github.com/vercel/api/pull/22693


.changeset/early-queens-boil.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel-internals/types': minor
```
styfle:
```suggestion
```

You can remove this since internal packages are not verisoned

PR #10684, opened 2023-10-10
[static-build] hexo-v6 fixture

Ethan-Arrowood:


packages/static-build/test/fixtures/hexo-v6/.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
cb1kenobi, line 1: **suggestion (non-blocking):** ‏Probably don't need this file.

cb1kenobi (approved):

PR #10685, opened 2023-10-10
[remix] Resolve static dir using non-parent `publicPath` segments

cb1kenobi:
The Remix `publicPath` config option is a "URL prefix" used to resolve assets at runtime in the browser. If the Remix app is hosted in a subdirectory, it's feasible that the `publicPath` will resolve to `../..` and accidentally publish private files.

The current method of resolving the static assets at build time involves resolving the `assetsBuildDirectory` using the `publicPath` and when `assetsBuildDirectory` is set to a one level less than the `publicPath` (e.g. `assetsBuildDirectory="public"` and `publicPath="/some/path"`), it thinks the static asset directory is one level up from the actual project directory resulting in files being published that shouldn't be published.

This PR resolves public assets the exact same way that Remix does as well as the built assets.

Linear: https://linear.app/vercel/issue/ZERO-1100/remix-use-of-publicpath-causing-assets-outside-project-dir-to-be

Initial report: https://github.com/vercel/vercel/pull/10305#discussion_r1346756667

Related community PR: https://github.com/vercel/vercel/pull/10666


Ethan-Arrowood (approved):

Ethan-Arrowood (dismissed):

cb1kenobi:
@TooTallNate When you get a minute, can you please review this. I want to make sure this is a sufficient fix and doesn't break anything else. Thanks!

packages/remix/test/fixtures/13-remix-v2-public-path/remix.config.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import('@remix-run/dev').AppConfig} */
+export default {
+  ignoredRouteFiles: ["**/.*"],
+  // appDirectory: "app",
+  assetsBuildDirectory: "public",
+  publicPath: "../..",
```
TooTallNate: In the original PR, the publicPath value was reported as `/apps/my-app/build/`, which the current code ends up resolving to `../..`. I don't think the intent was that `../../` was the value explicitly.

cb1kenobi: No prob, I'll fix.

TooTallNate (changes requested):
After pondering this a bit, and also checking the `remix-serve` source code, here's what I think we should do:

1. Remove any complex calculations for `staticDir`. Instead, it's just "public", [matching the hard-coded value in `remix-serve`](https://github.com/remix-run/remix/blob/8d682124ad96a50e2fe88448be3c3f68f794c054/packages/remix-serve/cli.ts#L119).
1. Glob `staticDir` as-is into the output files returned by the Builder (without the assumption that the remix client build files are in there, because they may not be).
1. Glob the `assetsBuildDirectory`, and use the build-utils [`rename()`](https://github.com/vercel/vercel/blob/main/packages/build-utils/src/fs/rename.ts#L11) function to prefix the `publicPath` value on the filenames returned by the glob, again [matching the logic in `remix-serve`](https://github.com/remix-run/remix/blob/8d682124ad96a50e2fe88448be3c3f68f794c054/packages/remix-serve/cli.ts#L112-L118).
1. Also, update the routes array to respect the `publicPath` for the route that adds the caching headers. Right now it's hard-coded to `/build`.
1. Update test fixture to have:
    ```
      "publicPath": /foo/dist",
      "assetsBuildDirectory": "outside-public"
    ```
1. Add probe for the test fixture to check for a static file in "public" returns 200 (i.e. `/favicon.ico`)
1. Add probe for the test fixture to check for a build asset file returns 200 and includes the cache-control header (i.e. `/foo/dist/manifest-1CCB369F.js`)

packages/remix/test/fixtures/13-remix-v2-public-path/vercel.json
```diff
@@ -0,0 +1,7 @@
+{
```
TooTallNate, line 1: Delete?

cb1kenobi: If this file is deleted, then will the build find the `rootDirectory`?

packages/remix/src/build.ts
```diff
@@ -467,10 +463,18 @@ module.exports = config;`;
     }),
   ]);
 
-  const output: BuildResultV2Typical['output'] = staticFiles;
+  const publicPath = remixConfig.publicPath.replace(/^\/|\/$/g, '');
+  const transformedBuildAssets = rename(buildAssets, name => {
+    return posix.join(publicPath, name);
```
TooTallNate: Could simplify this a bit.

```suggestion
  const transformedBuildAssets = rename(buildAssets, name => {
    return posix.join('./', remixConfig.publicPath, name);
```

cb1kenobi: Done, thanks!

Ethan-Arrowood (approved):
I don't understand the fix necessarily, but I can make sense of the test cases so if those pass then that is good for me.

TooTallNate (approved):
Nice work 👍 👏

PR #10686, opened 2023-10-10
[static-build] storybook-v7

Ethan-Arrowood:


cb1kenobi:
I don't think the `storybook-static` directory needs to be committed. It's the generated output.

Ethan-Arrowood:
you're right. it was missing from the gitignore

trek (approved):

PR #10688, opened 2023-10-11
[static-build] hydrogen-v2023

Ethan-Arrowood:


packages/static-build/test/fixtures/hydrogen-v2023/probes.json
```diff
@@ -0,0 +1,5 @@
+{
+  "probes": [
+    { "path": "/", "mustContain": "Mock.shop" }
```
EndangeredMassa, line 3: **question (non-blocking):** ‏Where is "Mock.shop" generated in the response? I don't see it explicitly.

EndangeredMassa (approved):

packages/static-build/test/fixtures/hydrogen-v2023/probes.json
Ethan-Arrowood: it was on the page when I deployed it for real. Hydrogen is weird. it has no static html file in its output. wasn't really sure to make of that but the probe still works.

PR #10703, opened 2023-10-13
[cli] autoconfirm in CI to domain buys

trek:
The `try to purchase a domain` CLI test starting failing on  11 October when the API became slightly slower to respond. This is because `vercel domains buy example.com` displays a prompt asking the user to `y` into agreeing to the purchase. We interacted with this in tests by spamming `y` on STDIN for a few seconds. The timing required is now a bit longer but I think we can skip that pattern entirely by passing the `--yes` option to the subcommand.

I don't _think_ we want this anywhere but in CI and let people unintentionally buy domains , so I've gated the use based on that, but happy to make it available always.

Ironically this issue went away on 13 October when the API response times improved and we could just unskip the test https://github.com/vercel/vercel/pull/10707


EndangeredMassa (dismissed):
As long as we're not documenting `--yes` for this command in help output or docs, I think this is a good solution.

Ethan-Arrowood (commented):
rather than adding an undocumented `--yes` option, I'd rather we add an internal environment variable that we use in the test. Otherwise, users may inadvertently pass `--yes` and get unexpected and undocumented experience.

trek:
@Ethan-Arrowood the behavior is already based on the presence of `CI`. We could just gate the behavior on that if you'd like.

Ethan-Arrowood:
Yes please

Ethan-Arrowood (approved):

PR #10705, opened 2023-10-13
Fix update Angular output path discovery

trek:
Angular CLI version 17 new esbuild and vite builder will output resources files to .

This change updates the  method to handle this.


EndangeredMassa:
@trek tests pass now, but on the source PR there are 2 changeset files

PR #10708, opened 2023-10-13
[next] fix `build` in appDir on windows

ztanner:
On Windows, `globPath` is a non-POSIX path (ie `'favicon.ico\\route.js'`) which won't match the paths generated by the `glob` util, since it normalizes Windows separators to Unix separators ([ref](https://github.com/vercel/vercel/blob/main/packages/build-utils/src/fs/normalize-path.ts#L6-L8))


styfle (commented):
Thanks!

Please add a unit test so we can ensure this works on windows 👍

cb1kenobi:
Manually tested this PR on my Windows machine and the build gets past the part that was failing, but fails symlinking dependencies. Fails regardless if deps are installed via npm or pnpm.

packages/next/src/index.ts
```diff
@@ -2690,13 +2690,14 @@ async function getServerlessPages(params: {
   outputDirectory: string;
   appPathRoutesManifest?: Record<string, string>;
 }) {
+  const appDir = path.join(params.pagesDir, '../app');
```
styfle:
```suggestion
  const appDir = path.posix.join(params.pagesDir, '../app');
```

Should this be posix? I'm pretty sure glob should be forward slash

> Note Glob patterns should always use / as a path separator, even on Windows systems, as \ is used to escape glob characters.

https://github.com/isaacs/node-glob

packages/next/src/index.ts
```diff
@@ -2712,7 +2713,7 @@ async function getServerlessPages(params: {
         '.',
         normalizedEntry === '/' ? '/index' : normalizedEntry
       )}.js`;
-      const globPath = `${path.join('.', entry)}.js`;
+      const globPath = `${path.posix.join('.', entry)}.js`;
```
styfle: I'm curious if the dot prefix is doing anything here? I tried this locally and it doesn't do anything:

image

cb1kenobi: Seems to just strip the leading slash.

Worth noting that on Windows `path.join('.', '\\file.js')` results in `file.js` while `path.posix.join('.', '\\file.js')` results in `\\file.js`, which makes sense.

If `entry` can start with a `/`, then either keep the `join()` or strip the leading slash.

ztanner: I don't know why this was here, but I think it's probably safe to err on the side of leaving it. It will have different behavior for something like `path.posix.join('.', 'my//file.js')` (I don't know if this could happen, but assuming this maybe was added for a reason)

packages/next/src/index.ts
ztanner: This change results in the following error on a Windows machine:

> Error: basePath/cwd must be an absolute path (app)

Worth mentioning that this is not a standard glob function...

ztanner:
> Manually tested this PR on my Windows machine and the build gets past the part that was failing, but fails symlinking dependencies. Fails regardless if deps are installed via npm or pnpm.

@cb1kenobi symlinking logic is coming from here: https://github.com/vercel/vercel/blob/main/packages/cli/src/util/build/write-build-result.ts#L380 and seems unrelated to the Next builder, but I might be missing something?

For context, the EPERM error (when not running as administrator) that I receive is throwing right at that mentioned line:
```
Created all serverless functions in: 6.005s
Collected static files (public/, static/, .next/static): 21.834ms
(debug) writing lambda _not-found.func C:\Users\test\dev\my-app\.vercel\output\functions\_not-found.rsc.func
Error: EPERM: operation not permitted, symlink '_not-found.func' -> 'C:\Users\test\dev\my-app\.vercel\output\functions\_not-found.rsc.func'
```

ijjk (approved):

styfle (approved):
Thanks!

PR #10723, opened 2023-10-16
[static-build] next-gen tests

Ethan-Arrowood:


Ethan-Arrowood:
Last update before switching focus:

The next-gen test script is making great progress. It does run significantly faster than the other integration tests. It has uncovered that many of our tests rely on certain defaults from the build container. Mainly the existence and permissions for tools like Hugo, go, Python, Ruby, and more. I think in order to land this and retain the test capabilities here, we have two options:
- Create a docker file anyone can use to run these tests on their machine
- Band-aid fix a GitHub Action workflow that loads in all these necessary tools and then runs these tests. + Document what tools certain tests need so developers can run locally if necessary.

Here is the current list of problematic tests:

- `33-hugo-zola-versions`, `34-hugo-old-version` require globally installed Hugo, Gutenberg, and Zola execs
- `40-pelican`, `41-mkdocs` requires python
- `53-native-gems` requires ruby
- `61-63`, `65`, and `66` all are special build-output-api v1 and may not need to exist in their current state going forward.

PR #10725, opened 2023-10-16
Enable source maps in tests

cb1kenobi:
The `tsconfig` files were unified with a common configuration in https://github.com/vercel/vercel/pull/10667 including disabling source maps. The source maps greatly increase the distribution size, but they are invaluable when the tests run, so this PR enables source maps for tests including adding a `tsconfig.json` where absent.

Linear: https://linear.app/vercel/issue/ZERO-364/enable-sourcemaps-in-all-testtsconfigjson-files


styfle (approved):
Looks great, thanks! :tada:

trek (approved):

PR #10727, opened 2023-10-17
[static-build] remove old gatsby tests and merge into new gatsby-v2

Ethan-Arrowood:
Removes some old Gatsby tests and merge's their relevant probes into the newer `gatsby-v2` fixture. 

No point having 3 separate Gatsby v2 tests when they can all be in one.


packages/static-build/test/fixtures/10-gatsby-without-build-script/package.json
```diff
@@ -1,45 +0,0 @@
...
-    "react-helmet": "^6.1.0"
-  },
-  "devDependencies": {
-    "prettier": "^2.2.1"
-  },
-  "keywords": [
-    "gatsby"
-  ],
-  "license": "MIT",
-  "scripts": {
```
EndangeredMassa, line 31: **question (non-blocking):** ‏Is it OK that we're no longer asserting this framework works without a "build" script?

I think so because the default build command is the same as what we test anyway. I just wanted to make sure this was done intentionally.

EndangeredMassa (dismissed):
Can you update the PR description with some context/reasons behind this change?

packages/static-build/test/fixtures/10-gatsby-without-build-script/package.json
Ethan-Arrowood: Yep exactly

Ethan-Arrowood:
Yes I can - thought title was descriptive enough.  will add a little more context though

EndangeredMassa (approved):

cb1kenobi (approved):

PR #10734, opened 2023-10-19
Fix RSC prefetch for index route with catch-all

ijjk:
When matching outputs and we have an `index.{ext}` file it will take priority over dynamic routes which we don't want to match prefetch RSC index outputs unless it's specifically a prefetch request so this remaps this specific file to avoid the overlap. 

x-ref: [slack thread](https://vercel.slack.com/archives/C061DMGDXQW/p1697664158904469?thread_ts=1697647520.783419&cid=C061DMGDXQW)


Ethan-Arrowood (approved):

ijjk (commented):
Looks like test failure is from CI still using Node.js v16 instead of v18 and Next.js canary bumped min required version

ztanner (approved):

PR #10738, opened 2023-10-19
fix(next): do not generate conflicting lambdas when using basePath with app router

brkalow:
When using `basePath` with statically generated routes, the builder currently generates lambdas for the root route that conflict with the prerender functions generated.

E.g. if we have `basePath: "test"` and a root route at: `app/page.js`, this is currently the output:

```
test
test.rsc
test/index
test/index.rsc
```

The result is that visiting the deployed app at `/test` renders an incomplete page. On the other hand visiting `/test/index` outputs the expected markup.

It seems like the intent is that we want to delete the lambdas that conflict with the prerender functions. I've updated the lambda name generation logic when doing the deleting to line up with how the lambda names are generated initially.


ztanner (approved):
Tested in https://github.com/vercel/vercel/pull/10795 and change looks good to me. Thanks!

ijjk (approved):

PR #10764, opened 2023-10-24
[error-utils] use Node.js `util.types.isNativeError` for `isError` method

Ethan-Arrowood:
I recently learned about https://nodejs.org/api/util.html#utiltypesisnativeerrorvalue which is pretty great. It will return `true` for all errors including subclasses (`class MyError extends Error {}`)

### DIFF
--- a/packages/error-utils/src/index.ts
+++ b/packages/error-utils/src/index.ts
@@ -1,3 +1,5 @@
+import util from 'node:util';
+
 export interface SpawnError extends NodeJS.ErrnoException {
   spawnargs: string[];
 }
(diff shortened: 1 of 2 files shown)


packages/error-utils/src/index.ts
```diff
@@ -13,22 +15,9 @@ export const isObject = (obj: unknown): obj is Record<string, unknown> =>
 /**
  * A type guard for `try...catch` errors.
  *
- * This function is based on:
- * https://github.com/stdlib-js/assert-is-error
  */
 export const isError = (error: unknown): error is Error => {
```
TooTallNate, line 19: Should we mark this function as deprecated?

TooTallNate (dismissed):

cb1kenobi (dismissed):
Neato

trek (dismissed):

TooTallNate (dismissed):

TooTallNate (approved):

PR #10772, opened 2023-10-25
Collect lockfileVersion for yarn

mehulkar:
This PR gets the lockfile version from a `yarn.lock` into the returned value from `scanParentDirs`. This function is used by the Vercel build container for DD tracing, and we are currently missing the lockfileVersion when the package manager is yarn.

### DIFF
--- a/packages/build-utils/src/fs/read-config-file.ts
+++ b/packages/build-utils/src/fs/read-config-file.ts
@@ -30,7 +30,9 @@ export async function readConfigFile<T>(
     if (data) {
       const str = data.toString('utf8');
       try {
-        if (name.endsWith('.json')) {
+        if (name === 'yarn.lock') {
+          return yaml.safeLoad(str, { filename: name }) as T;
+        } else if (name.endsWith('.json')) {
           return JSON.parse(str) as T;
         } else if (name.endsWith('.toml')) {
           return toml.parse(str) as unknown as T;
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -299,9 +299,13 @@ export async function scanParentDirs(
   let lockfileVersion: number | undefined;
   let cliType: CliType = 'yarn';
 
-  const [hasYarnLock, packageLockJson, pnpmLockYaml, bunLockBin] =
+  const hasYarnLock = Boolean(yarnLockPath);
+
+  const [yarnLockYaml, packageLockJson, pnpmLockYaml, bunLockBin] =
     await Promise.all([
-      Boolean(yarnLockPath),
+      yarnLockPath
+        ? readConfigFile<{ __metadata: { version: number } }>(yarnLockPath)
+        : null,
       npmLockPath
         ? readConfigFile<{ lockfileVersion: number }>(npmLockPath)
         : null,
@@ -320,6 +324,7 @@ export async function scanParentDirs(
   } else if (hasYarnLock) {
     cliType = 'yarn';
     lockfilePath = yarnLockPath;
+    lockfileVersion = Number(yarnLockYaml.__metadata.version);
   } else if (pnpmLockYaml) {
     cliType = 'pnpm';
     lockfilePath = pnpmLockPath;


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -320,6 +324,7 @@ export async function scanParentDirs(
   } else if (hasYarnLock) {
     cliType = 'yarn';
     lockfilePath = yarnLockPath;
+    lockfileVersion = Number(yarnLockYaml.__metadata.version);
```
mehulkar, line 327: I see this field in yarn3 and yarn4, but need to check yarn 1 and 2 also

EndangeredMassa (changes requested):
@mehulkar can you update the PR description with some details about what is being done here and why?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -299,9 +299,13 @@ export async function scanParentDirs(
   let lockfileVersion: number | undefined;
   let cliType: CliType = 'yarn';
 
-  const [hasYarnLock, packageLockJson, pnpmLockYaml, bunLockBin] =
+  const hasYarnLock = Boolean(yarnLockPath);
+
+  const [yarnLockYaml, packageLockJson, pnpmLockYaml, bunLockBin] =
     await Promise.all([
-      Boolean(yarnLockPath),
+      yarnLockPath
```
EndangeredMassa, line 306: **suggestion (non-blocking):** ‏definitely not a big deal, but we do already have a slighty more explicit variable

```suggestion
      hasYarnLock
```

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: **suggestion (blocking)** ‏

I think this needs to be careful when accessing `yarnLockYaml.__metadata`. I don't see that in a `yarn@1` lockfile, for example.

We could look for the comment:

```
# yarn lockfile v1
```

and parse it out, but I don't know if that version changes in `yarn@1` at all. Maybe we can assume lockfile v1?

```suggestion
    lockfileVersion = Number(yarnLockYaml.__metadata?.version) || 1;
```

---

**Yarn 1**

```
$ yarn -v
1.22.19

$ head yarn.lock
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


"@ampproject/remapping@^2.2.0":
  version "2.2.0"
  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
  dependencies:
    "@jridgewell/gen-mapping" "^0.1.0"
```

**Yarn 2**

```
$ yarn -v
2.4.3

$ head yarn.lock
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!

__metadata:
  version: 4
  cacheKey: 7

"@planetscale/database@npm:^1.0.1":
  version: 1.11.0
  resolution: "@planetscale/database@npm:1.11.0"
```

packages/build-utils/src/fs/read-config-file.ts
```diff
@@ -30,7 +30,9 @@ export async function readConfigFile<T>(
     if (data) {
       const str = data.toString('utf8');
       try {
-        if (name.endsWith('.json')) {
+        if (name === 'yarn.lock') {
+          return yaml.safeLoad(str, { filename: name }) as T;
```
EndangeredMassa, line 34: **suggestion (blocking):** ‏add tests

Can you add tests for this behavior?

EndangeredMassa (changes requested):

packages/build-utils/src/fs/run-user-scripts.ts
mehulkar: it's not the same thing. `hasYarnLock` was being used to flag through the existence of the lockfile. Now it's the actual lockfilePath, (similar to the other paths). I moved the `hasYarnLockfile` binding further down

mehulkar:
I won't be able to follow through on this right now, closing.

PR #10779, opened 2023-10-28
Rework prefetch route handling

ijjk:
Follow-up to https://github.com/vercel/vercel/pull/10750 this removes the underscore prefetching from all prefetch outputs and instead only applies it to the index route itself as this causes issues with PPR making these outputs prerenders and being able to interpolate the route param values. 

There is the edge case of a user returning the literal value `index` from `generateStaticParams` but we can tolerate that more than ppr not working as expected.

### DIFF
--- a/.changeset/tricky-rules-sin.md
+++ b/.changeset/tricky-rules-sin.md
@@ -1,0 +1,5 @@
+---
+"@vercel/next": patch
+---
+
+Rework prefetch route handling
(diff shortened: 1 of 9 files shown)


Ethan-Arrowood (approved):
This seems sorta like a breaking change is it not? If I was previously relying on / expecting `index` to route to my page and now it doesn't?

ijjk:
It hasn't been rolled yet/isn't officially supported by our proxy so this removes the broken behavior before we do roll it out

PR #10819, opened 2023-11-09
[remix] Fix issue where `npm install` was not properly injecting forked compiler

TooTallNate:
User reported a scenario in which npm was ignoring the second `npm install` command to replace the Remix compiler with our forked version. This seems like a bug in npm. Workaround that seems to work is to remove `@remix-run/dev` entirely and install our forked version as a "standard" dependency (instead of using the `npm:` syntax). The `bin` entry for `remix` should at that point be our forked compiler.

### DIFF
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -225,6 +225,7 @@ export const build: BuildV2 = async ({
     // Remove the env vars that will cause pnpm to consider
     // as CI, which enables `--frozen-lockfile` enforcement
     const nonCiEnv = { ...spawnOpts.env };
+    console.log(nonCiEnv);
     delete nonCiEnv.VERCEL;
     delete nonCiEnv.NOW_BUILDER;
 
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -222,13 +222,22 @@ export const build: BuildV2 = async ({
   if (depsModified) {
     await fs.writeFile(packageJsonPath, JSON.stringify(pkg, null, 2) + '\n');
 
+    // Remove the env vars that will cause pnpm to consider
+    // as CI, which enables `--frozen-lockfile` enforcement
+    const nonCiEnv = { ...spawnOpts.env };
+    delete nonCiEnv.VERCEL;
+    delete nonCiEnv.NOW_BUILDER;
+
     // Purposefully not passing `meta` here to avoid
     // the optimization that prevents `npm install`
     // from running a second time
     await runNpmInstall(
       entrypointFsDirname,
       [],
-      spawnOpts,
+      {
+        ...spawnOpts,
+        env: nonCiEnv,
+      },
       undefined,
       nodeVersion
     );
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -222,13 +222,12 @@ export const build: BuildV2 = async ({
   if (depsModified) {
     await fs.writeFile(packageJsonPath, JSON.stringify(pkg, null, 2) + '\n');
 
-    // Remove the env vars that will cause pnpm to consider
-    // as CI, which enables `--frozen-lockfile` enforcement
+    // Bypass `--frozen-lockfile` enforcement by removing
+    // env vars that are considered to be CI
     const nonCiEnv = { ...spawnOpts.env };
     delete nonCiEnv.CI;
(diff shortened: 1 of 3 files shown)


gustavoguichard:
Hey @TooTallNate are you gonna open this PR? Looking fwd get it merged =)

packages/remix/src/build.ts
```diff
@@ -209,15 +214,34 @@ export const build: BuildV2 = async ({
...
+      depsModified = true;
     }
   }
 
-  if (depsToAdd.length) {
-    await addDependencies(cliType, depsToAdd, {
-      ...spawnOpts,
-      cwd: entrypointFsDirname,
-    });
+  if (depsModified) {
```
cb1kenobi, line 222: **suggestion (non-blocking):** ‏Consider adding a `debug()` log mentioning the `package.json` is being updated.

cb1kenobi (approved):

EndangeredMassa (approved):
Is there a test we can add here or in the framework test matrix to exercise this case?

lsnape:
Hi there,

Just a heads up that this change has inadvertently broke our CI build step, for reasons that aren't entirely clear to me.

Our dev dependencies has `@remix-run/dev` pinned to 2.3.0 in the package.json. After this change, it is now replaced with `@vercel/remix-run-dev` v2.2.0. Once the dependency switch happens, we see compilation errors appear on the lines that import CSS files in `root.tsx`.

AFAICT, the `@vercel/remix-run-dev` source code isn't public, so I'm unable to investigate much further. I've managed to temporarily fix the build by pinning the vercel CLI version to before this change went in, but I suspect that may cause further issues down the line.

Any ideas what might be causing the CSS import errors?

EndangeredMassa:
@lsnape sorry to hear it. We'll take a look! If you could provide more details in a [Vercel Support Case](https://vercel.com/help), that could be really helpful.

maxfi:
Likely similar to @lsnape's issue, this has also broken my build on vercel. Deployments that worked on previous versions of vercel cli (eg, Vercel CLI 32.4.1) are now failing with:

```
Error: Cannot find module '@remix-run/dev/dist/config'
Require stack:
- /vercel/path1/node_modules/.pnpm/remix-routes@1.4.4/node_modules/remix-routes/lib/cli.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/vercel/path1/node_modules/.pnpm/remix-routes@1.4.4/node_modules/remix-routes/lib/cli.js:40:18)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/vercel/path1/node_modules/.pnpm/remix-routes@1.4.4/node_modules/remix-routes/lib/cli.js'
  ]
}
Node.js v18.18.0
Error: Command "pnpm install" exited with 1
```

It looks like the issue is coming from here: https://github.com/yesmeck/remix-routes/blob/master/packages/remix-routes/src/cli.ts.

I'm calling `remix-routes` in the package.json `postinstall` script:

```json
"scripts": {
  "postinstall": "remix-routes && pnpm run prisma:generate && pnpm run prisma:generate:winmagi && pnpm run install:semantic-ui-assets"
}
```

Setting the `VERCEL_CLI_VERSION` environment variable to `vercel@32.5.6` fixes the issue.

maxfi:
In case it helps, I'm still seeing the [same issue](https://github.com/vercel/vercel/pull/10819#issuecomment-1848327832) with `Vercel CLI 32.7.1`.

TooTallNate:
To anyone commenting in this PR after the merge, please open a new issue if you are still having a problem. But also, I strongly encourage you to migrate to Remix + Vite, because when you deploy a Remix + Vite application to Vercel, we do not inject a forked compiler at all, so this entire category of errors should be eliminated.

PR #10820, opened 2023-11-09
Add .node-version file

Ethan-Arrowood:
This a small QOL change. Essentially if you have a tool like `fnm` on your machine for using multiple versions of Node.js, they will look for files like this one and automatically switch your terminal session node version to the value in it. 

The `vercel/front` repo uses this too


.node_version
```diff
@@ -0,0 +1 @@
+v16.20.2
```
TooTallNate, line 1: Why 16?

Ethan-Arrowood: That's the version we use for the repo right? Or are we baseline v18 nowadays?

EndangeredMassa: It's [still 16](https://github.com/vercel/vercel/blob/main/.github/workflows/test.yml#L12), but [some next tests run in 18](https://github.com/vercel/vercel/blob/main/utils/chunk-tests.js#L17).

While we're here, can we set [`.nvmrc`](https://github.com/nvm-sh/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file) as well? That's what I use and [`vercel/api` already has one](https://github.com/vercel/api/blob/main/.nvmrc).

EndangeredMassa (dismissed):

EndangeredMassa (approved):

cb1kenobi (approved):

PR #10822, opened 2023-11-09
[build-utils] Select Node.js version based on what's available in build-container

TooTallNate:
Makes `getLatestNodeVersion()` and `getSupportedNodeVersion()` (and thus by extension - `getNodeVersion()`) be aware of the available Node.js versions when running inside the build-container. This is to address the situation with the new AL2023 build-container image which has different Node versions installed compared to the existing AL2 build image.

### Project Settings `20.x` with package.json `"engines": "18.x"`

If the Project Settings Node Version is set to `20.x`, but the package.json has `"engines": "18.x"`, then the build will fail like so, because Node 18 is not present on the AL2023 build image:

Screenshot 2023-11-09 at 1 25 41 PM

### Project Settings `18.x` with package.json `"engines": "20.x"`

If the Project Settings Node Version is set to `18.x`, but the package.json has `"engines": "20.x"`, then the build will fail like so, because Node 20 is not present on the AL2 build image:

Screenshot 2023-11-09 at 1 34 43 PM

### Project Settings `18.x` with no package.json `"engines"`

If Project Settings Node Version is set to `18.x`, but the package.json has no "engines" (and thus wants "latest"), then the latest available Node version in the build-container, which would be Node 18.

### DIFF
--- a/.changeset/cuddly-squids-cough.md
+++ b/.changeset/cuddly-squids-cough.md
@@ -1,0 +1,5 @@
+---
+'@vercel/build-utils': patch
+---
+
+Select Node.js version based on what's available in build-container
--- a/packages/build-utils/src/fs/node-version.ts
+++ b/packages/build-utils/src/fs/node-version.ts
@@ -4,6 +4,8 @@ import { NodeVersion } from '../types';
 import { NowBuildError } from '../errors';
 import debug from '../debug';
 
+export type NodeVersionMajor = ReturnType<typeof getOptions>[number]['major'];
+
 function getOptions() {
   const options = [
     { major: 18, range: '18.x', runtime: 'nodejs18.x' },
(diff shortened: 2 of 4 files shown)


cb1kenobi:
PR looks good, but I don't love the warning/error message when the `engines` doesn't match the project settings.

The warning implies `engines` take precedence, however the error contradicts that and says the project settings take precedence. Which one takes precedence? I believe project settings would as it sounds like that determines the version is used to select the build image.

I'd argue don't show the warning if there's going to be an error or possibly get rid of the warning altogether.

I also don't love that it says `"Found invalid Node.js Version..."`. In my mind, Node 18 is a valid version. I feel the whole error message could be reworked. Maybe something like `Error: Detected unsupported Node.js version combination: "engines" is set to Node.js "18.x", but the project settings are set to "20.x". Please set "engines": { "node": "20.x" } in your "package.json" file to use Node.js 20.`

EndangeredMassa (approved):
Let's merge first and iterate on messaging

TooTallNate:
@cb1kenobi The error messages are what was already present. I agree they could use some tweaking but I'd prefer to do that in a separate PR. This PR is to just get things working.

cb1kenobi (approved):

cb1kenobi:
> @cb1kenobi The error messages are what was already present. I agree they could use some tweaking but I'd prefer to do that in a separate PR. This PR is to just get things working.

Sounds good to me! Nice work!

PR #10830, opened 2023-11-13
chore: Update Remix example README

brookslybrand:
Not sure the best spot to put this command in the Readme, but I thought having a `create-remix` command might be useful for people trying to get going.

If there's a more preferred way to scaffold this project on your machine, I'm open to that, just using what I know


leerob (approved):
SGTM!

PR #10834, opened 2023-11-15
[build-utils] Remove Node v20 env var check

TooTallNate:
It's safe to remove this env var check, because #10822 makes it explicit on which versions of Node are actually present in the build environment.


trek (dismissed):

EndangeredMassa (approved):

pvdz (approved):
:+1:

cb1kenobi (approved):

PR #10848, opened 2023-11-17
[cli] forbids globally installed @vercel/speed-insights and @vercel/analytics

feugy:
### 🧐 What's in there?

The CLI a detection feature for applications using the new `@vercel/speed-insights` package to collect performance metrics. It also checks for `@vercel/analytics` for tracking visits.

These packages **must** be referenced in the application's `package.json`, but:
- current detection returns the version range, which could differ from the actual package version
- current detection does not prevent from deploying an app which does no reference a package, but has it globally available (monorepos: triggered a Customer issue)

This PR fixes the detected versions, and the build command will fail in case of globally-installed packages.

### 🧪 How to test?

I updated and added e2e tests: `pnpm -F vercel test test/integration-2.test.ts`
The existing test has a dependency on `@vercel/speed-insights` in a close descriptor file.
The new test installs `@vercel/analytics` as global, to mimic a monorepo.

### ❗ Notes to reviewers

Initially I considered an approach similar to the [detectorFileSystem](https://github.com/vercel/vercel/blob/6f3ae1a0edb89c5e263bcff17b8fb1d72660acd8/packages/fs-detectors/src/monorepos/get-monorepo-default-settings.ts#L64C7-L64C25). However, it wouldn't solve problem # 2.

I went for the require approach, which downsize is to need installing the dependencies first. Therefore, we have to alter the build.json file to record the results.


packages/cli/src/commands/build/index.ts
```diff
@@ -475,9 +453,7 @@ async function doBuild(
     })
   );
   buildsJson.builds = Array.from(buildsJsonBuilds.values());
-  await fs.writeJSON(join(outputDir, 'builds.json'), buildsJson, {
-    spaces: 2,
-  });
+  await writeBuildJson(buildsJson, outputDir);
```
tobiaslins, line 456: praise: nice new helper

tobiaslins (approved):
looks good!

Ethan-Arrowood (approved):

PR #10859, opened 2023-11-21
[ruby] Enable `ruby3.2` runtime

TooTallNate:
Allows for Ruby 3.2 Serverless Functions to be created.

**Note:** New test fixtures will be added after this is deployed to production.


packages/ruby/src/index.ts
```diff
@@ -59,16 +60,30 @@ async function bundleInstall(
       gemfileContent.replace('ruby "~> 2.7.x"', 'ruby "~> 2.7.0"')
```
Ethan-Arrowood, line 61: do we need to change this line for 3.2?

trek (dismissed):

packages/ruby/src/index.ts
EndangeredMassa: **question (blocking):** ‏do we need to adjust this?

EndangeredMassa (changes requested):

packages/ruby/src/index.ts
TooTallNate: 49e93d3c744225f0f66322aea360a6725687cd42

EndangeredMassa (approved):

Issue #6899, opened 2021-10-28
[cli] Enable signup with email address


TooTallNate:
Right now you cannot sign up for a new Vercel account via email:

```
$ vc login
Vercel CLI 23.1.3-canary.15 — https://vercel.com/feedback
> Log in to Vercel email
> Enter your email address: n+aadsffadsfsa@n8.io
Error! Please sign up: https://vercel.com/signup
```

This should be allowed, per @rauchg.

leo:
Since a single signup flow provides drastic UX, maintenance, and reliability advantages, we'll rely on the web flow, but ensure people aren't asked to enter an email address multiple times.

We'll land a patch today for ensuring that the command doesn't prompt for an email address, and then we can use this issue for tracking that we also want to keep the command pending and automatically log people in after they've completed the signup flow on the web – instead of asking them to run `vercel login` afterward.

sreekeshiyer:
This is an interesting problem. 
We can at least have the "Signing up with email" in the CLI in my opinion, and the user can then have the third party providers connected from the UX. 
I'll be glad to work on this.

Aamir-Ansari-almost:
> This is an interesting problem. We can at least have the "Signing up with email" in the CLI in my opinion, and the user can then have the third party providers connected from the UX. I'll be glad to work on this.

I for one also find this issue quite interesting, I will be happy to contribute to the solution with @sreekeshiyer.

sreekeshiyer:
> > This is an interesting problem. We can at least have the "Signing up with email" in the CLI in my opinion, and the user can then have the third party providers connected from the UX. I'll be glad to work on this.
> 
> I for one also find this issue quite interesting, I will be happy to contribute to the solution with @sreekeshiyer.

Sure! 👍

shaziyabandukia:
This is currently in the HIT papercuts backlog, but our team doesn't manage this side of the house. Happy to support as needed, so plz feel free to ping!

Nukealbert:
@TooTallNate ,I find this issue is still pending, can you assign this issue to me?

PR #10863, opened 2023-11-21
Fixed this isse: [cli] Enable signup with email address #6899

mrayushmehrotra:
I Have added a regex expression and pipe operator "| |"   to check if the input email passes any of the conditions, the error will not execute. this will remove this error:  Error! Please sign up: https://vercel.com/signup


EndangeredMassa:
Thanks for the PR! We want to be more flexible at this point of the code.

PR #10868, opened 2023-11-22
[cli] Passthrough original arguments to CLI extensions

cb1kenobi:
When spawning a CLI extension, passthrough the original argv as the parsed argv doesn't contain parsed options.


packages/cli/src/index.ts
```diff
@@ -501,7 +501,7 @@ const main = async () => {
         exitCode = await execExtension(
           client,
           targetCommand,
-          argv._.slice(3),
+          process.argv.slice(3),
```
TooTallNate: I don't think we should go this route. The [default parsed args](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/arg-common.ts#L1-L28) are specifically what will be missing from `argv._`, and some of them don't make sense to pass along to the extension (`--cwd`, `--token`, `--api`, possibly others).

I think we should be explicit about which common args are passed through. Some of them might make more sense to pass along as an env var, for example (`VERCEL_DEBUG`, perhaps?).

TooTallNate (changes requested):

packages/cli/src/index.ts
mrmckeb: I feel that `vc` should pass everything along to the extension. For example, `--cwd` might be an option we add in future.

We definitely want args like `--help` and `--version` to go through.

cb1kenobi: So, common args plus `--version`. We can whitelist the args allowed through, that's fine.

TooTallNate: * `--cwd` - Vercel CLI has already changed into that directory by the time the extension is invoked, so doesn't seem relevant to pass to the extension.
* `--token` - Shouldn't be passed to the extension (unless possibly opted-in / allowlisted). Extension should use the local HTTP proxy server instead for most cases.
* `--api` - Extension probably doesn't need to know about this because the local HTTP proxy server would abstract that.

For `--help` and `--version`, ya fasho, makes sense 👍

TooTallNate: `--debug` might make sense to pass along, however, I was imagining that we could instead add additional endpoints to the proxy server so that logging / prompting could be standardized. For example:

```
POST /cli/debug
Content-Type: application/json

{ "message": "this is a debug log" }
```

```
POST /cli/prompt
Content-Type: application/json

{ "type": "select", "options": [ "apple", "orange", "banana" ] }
```

Ethan-Arrowood (approved):
love it

packages/cli/src/index.ts
trek: I think I'm tentatively with @mrmckeb on this one. Other than specifically blacklisting args with values we consider sensitive (e.g `--token`), if we whitelist the allow arguments no extenson will have the ability to define options that make sense for their specific use cases.

If an extension wanted the (hypothetical fun feature) of `--make-me-laugh` that prints a joke once a command completes, they'd have to beg permission from us.

OTOH letting a subcommand use _whatever_ options names they want could result in a collision of us adding a new global option with the same name and causing confusion. 

What does `git` do? Maybe they're a good example of how this pattern has evolved over time?

Ethan-Arrowood (approved):
Nice! I really like the test case for this too. Well done

EndangeredMassa:
We'll revisit this when we touch extensions again

PR #10880, opened 2023-11-28
[go] Set Lambda runtime to "provided.al2" behind env var

TooTallNate:
This is a re-land of https://github.com/vercel/vercel/pull/10856, but with the addition of the env var check so that we can feature flag this change.


packages/go/src/index.ts
```diff
@@ -50,7 +50,7 @@ export { shouldServe };
 // we need our `main.go` to be called something else
 const MAIN_GO_FILENAME = 'main__vc__go__.go';
 
-const HANDLER_FILENAME = `handler${OUT_EXTENSION}`;
+const HANDLER_FILENAME = `bootstrap${OUT_EXTENSION}`;
```
EndangeredMassa, line 53: **question (blocking):** ‏Do we need to check the env var and preserve the old behavior here, too?

EndangeredMassa (changes requested):

cb1kenobi (approved):

packages/go/src/index.ts
TooTallNate: It shouldn't make much of a difference. We would need to do the check in the `build()` function and pass around the result to the couple places that this is used. We can do that if you prefer though.

Ethan-Arrowood (approved):

packages/go/src/index.ts
EndangeredMassa: I thought that NOT setting `VERCEL_USE_GO_PROVIDED_RUNTIME` would use `go1.x`, which is expecting `handler${OUT_EXTENSION}`.

If it works before/after as-is, then I'm good.

EndangeredMassa (approved):

packages/go/src/index.ts
TooTallNate: With `go1.x`, we can name it anything. The name of the executable is specified as the `handler` proper when creating the Lambda.

PR #10882, opened 2023-11-28
[tests] tidy and make select-org tests more realistic

trek:
Tracking down a bug with "northstar" users and have a little prefactor that 

a) isolates setup code so it's less repeated
b) makes the "northstar" tests more reflective of actual execution.

Northstar users will always have a `currentTeam` value on their `client.config`:

https://github.com/vercel/vercel/blob/aa0f3d712b6be4aa1af2f28acb0f4c7a6cdf9f16/packages/cli/src/util/get-scope.ts#L18-L20

And finally c) add tests for the case of a non-northstar users where a team scope is supplied.

The final hypothetical matching group (northstart users with team scope provided) is the bug that I'll fix with a refactor.


packages/cli/test/unit/util/input/select-org.test.ts
```diff
@@ -4,33 +4,75 @@ import { useTeams } from '../../../mocks/team';
...
+    beforeEach(() => {
+      user = useUser({
         version: 'northstar',
       });
-      const team = useTeams()[0];
+    });
+
+    beforeEach(() => {
+      client.config.currentTeam = team.id;
+    });
```
Ethan-Arrowood: Lets merge these two `beforeEach`

packages/cli/test/unit/util/input/select-org.test.ts
```diff
@@ -4,33 +4,75 @@ import { useTeams } from '../../../mocks/team';
...
+        await expect(selectOrgPromise).resolves.toHaveProperty('id', team.id);
+      });
+
+      it('should allow selecting team', async () => {
+        const selectOrgPromise = selectOrg(client, 'Select the scope');
+        await expect(client.stderr).toOutput('Select the scope');
+        client.stdin.write('\x1B[B'); // Down arrow
+        client.stdin.write('\r'); // Return key
+        await expect(selectOrgPromise).resolves.toHaveProperty('id', user.id);
+      });
```
Ethan-Arrowood, line 61: Are these the exact same tests as the ones above? You should use one of the test iterator methods instead so we don't have duplicate test code.

trek: I'm always -1 on test iterators because they make it harder to isolate a test when it fails and during development but happy to be outvoted.

packages/cli/test/unit/util/input/select-org.test.ts
```diff
@@ -4,37 +4,92 @@ import { useTeams } from '../../../mocks/team';
...
+
+    it('should not allow selecting user', async () => {
       const selectOrgPromise = selectOrg(client, 'Select the scope');
       await expect(client.stderr).not.toOutput(user.name);
       client.stdin.write('\r'); // Return key
       await expect(selectOrgPromise).resolves.toHaveProperty('id', team.id);
     });
+
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('automatically selects the correct scope when autoconfirm flag is passed', async () => {
```
trek, line 90: This exercises the current broken behavior on main

TooTallNate (approved):

Ethan-Arrowood (approved):

packages/cli/test/unit/util/input/select-org.test.ts
EndangeredMassa: I prefer duplication in test code to some degree. I think this is better without an iterator.

EndangeredMassa (approved):

PR #10884, opened 2023-11-28
[cli] Fix behavior for combination of northstar user + team scope provided to cli as an argument.

trek:
Fixes a neat little bug with the intersection of northstar users, teams, and the behavior of `Array#find_index`:

When selecting an org for various cli actions _non_ northstar users will see a select list with their user in the 0th position, and their teams in the 1..nth positions, like so:

```
// user: Logan
// teams: [Avengers, X-Men]
○ Logan
○ Avengers
○ X-Men
``` 

We'd like to preselect either a team referenced in `client.config.currentTeam` or if said team cannot be found in the collection of a user's teams, preselect the user. So if `X-Men` is `currentTeam`, we find that item (element `1` in the teams array) and increment that value by `1` to account for the user being element `0` in the list of choices:

```
// currentTeam: X-Men
// user: Logan
// teams: [Avengers, X-Men]
○ Logan
○ Avengers
● X-Men
``` 

If we _can't_ find the `currentTeam` in the list of teams (or one isn't provided), `Array#find_index` returns `-1`, which we increment to by `1` to get the 0th element in the list of choices:

```
// currentTeam: undefined
// user: Logan
// teams: [Avengers, X-Men]
● Logan
○ Avengers
○ X-Men
``` 

Neat trick!

However, Northstar users _don't_ use the user account as the 0th element in the list of choices. Northstar users will in fact have a team that represents a hidden default team with the same name as their `user.name` and this will be the 0th element in their teams:

```
// user: Logan
// teams: [Logan, Avengers, X-Men]
○ Logan
○ Avengers
○ X-Men
``` 

This of course means the trick of `+1` to index of the team selects the wrong item _and_ can result in `undefined` in cases where the `currentTeam` references the last team.

```
// currentTeam: X-Men
// user: Logan
// teams: [Logan, Avengers, X-Men]
○ Logan
○ Avengers
○ X-Men
``` 

● how'd we get out here?!?! 👀 

To address the issue this PR:
1. calls `findIndex` on the array of _choices_ checking for matches to `currentTeam`
2. wraps that in `Math.max` to set to `0` if `find_index` returns `-1` when a `currentTeam` can't be found.


packages/cli/src/util/input/select-org.ts
```diff
@@ -43,18 +43,21 @@ export default async function selectOrg(
     })),
   ];
 
-  const defaultOrgIndex = teams.findIndex(team => team.id === currentTeam) + 1;
+  let defaultChoiceIndex = Math.max(
```
TooTallNate: `const`?

cb1kenobi (dismissed):

Ethan-Arrowood (dismissed):

cb1kenobi (dismissed):

cb1kenobi (approved):

PR #10892, opened 2023-12-01
[cli] fix error when @vercel/analytics is a transitive dependency of the deployed application

feugy:
### 🧐 What's in there?

We had to [revert](https://github.com/vercel/vercel/pull/10895/files#diff-38021e12d391784126cf681d144685353301be4981b855538b2d4d823d1b5536) #10848 which was incorrectly preventing builds when they have a transitive dependency to either `@vercel/analytics` or `@vercel/speed-insights`.

This PR partially restores the dependency detection and removes a undesired warning.

New behavior:
- we don't remove `VERCEL_ANALYTICS_ID` (legacy trigger) when identifying a dependency. Mixing legacy and new collection mechanism is now handled on Vercel intake APIs.
- if your application has a transitive dependency to `@vercel/speed-insights` or `@vercel/analytics`**after building**, then we store the actual version numbers in `builds.json`. This will be used further down the build chain.
- any other case: we don't do anything special.

### 🧪 How to test?

1. build the CLI
2. run `pnpm -F vercel test test/integration-2.test.ts`

### ❗ Notes to reviewers

> Why do we have 2 mechanisms?

Speed Insights comes in 2 flavors: 
1. (legacy) based on `VERCEL_ANALYTICS_ID` env var (misleading, historical name), Next.js will inject the measurement script into the built application
2. (upcoming) the application imports `@vercel/speed-insights` package and uses a React/Svelte/Vue/JS component to collect measures.

> Why don't we need to remove the env var anymore?

Instead of relying on dependency detection at build time, Vercel intake API will handle cases where the same application reports metrics with both new and legacy mechanisms.

> Why do we need to detect after the build?

The build chain is interested in knowing the deployed versions of `@vercel/speed-insights` and `@vercel/analytics`. However we don't need to be conservative, and we can safely report dependencies even if they are indirect or even unused.

That's why we're requiring the files after installing all dependencies, and store the used versions in `builds.json`for.


packages/cli/test/integration-2.test.ts
```diff
@@ -1187,17 +1187,15 @@ test('[vc build] should build project with `@vercel/speed-insights`', async () =
   }
 });
 
-test('[vc build] should not build project with `@vercel/analytics globally installed`', async () => {
+test('[vc build] should build project with `@vercel/analytics globally installed`', async () => {
```
TooTallNate: Minor nit, but I don't think the phrasing "globally" is correct here. To me, globally means with the `-g` flag (`npm i -g ...`), whereas here you mean higher up in the workspace hierarchy. I'd maybe say "at the root level" or "in a parent workspace"

tobiaslins (dismissed):
<picture data-single-emoji=":lfgtm:" title=":lfgtm:">:lfgtm:</picture>

packages/build-utils/src/fs/read-installed-version.ts
```diff
@@ -0,0 +1,17 @@
...
+  pkgName: string
+): Promise<string | undefined> {
+  try {
+    const descriptorPath = require.resolve(`${pkgName}/package.json`, {
+      paths: [cwd()],
+    });
+    const descriptor = await readJSON(descriptorPath);
+    return descriptor?.version;
+  } catch {
+    // ignore errors: the package is simply not installed.
```
EndangeredMassa: **convention (soft-blocking):** log all errors

In this or a follow-up PR: Can we debug log this error? I know it shouldn't matter, but I've spent plenty of time tracking down an issue of an ignored error that shouldn't have mattered, but did, in my time.‏

EndangeredMassa (dismissed):

packages/build-utils/src/fs/read-installed-version.ts
feugy: Hum, but in this case, the error is meant to be silent.
If the package is install, there will be no error.
If the package is not installed, require.resolve will throw, and we don't want to log it.
It's like when using `fs.access()` to check that a file exist, and catching the error when it does not.

Can you show me an example of a log that wouldn't be too noisy: most build command will have these 2 logs (since most project don't have `@vercel/analytics` nor `@vercel/speed-insights`), and I'd like to reduce the pollution.

Ethan-Arrowood: I think using `debug` for this would be sufficient. So its still silent unless you have `--debug` enabled in order to give more insights.

Ethan-Arrowood (dismissed):

packages/build-utils/src/fs/read-installed-version.ts
EndangeredMassa: Exactly. If you could pass in `client` and use the `output.debug` to log the error, that'd be awesome.

feugy: Thanks for all the tips.
Since `Client` interface is defined in cli package, and to avoid a circular dependency, I'm moving this utility function back to cli instead of build-utils.

Ethan-Arrowood (approved):

PR #10915, opened 2023-12-07
[node] Await waitUntil promises to resolve before exiting

nuta:
Say you have a middleware (`middleware.js`) that looks like this:

```js
export async default function middleware(req, ctx) {
  ctx.waitUntil(tooLongFunction())
}

async function tooLongFunction() {
  console.log('tooLongFunction started')
  await new Promise((resolve) => setTimeout(resolve, 10000))
  console.log('tooLongFunction finished')
}
```

When you run this middleware locally with `vercel dev`, you won't see the `tooLongFunction finished` message because the server process is killed right after finishing the response. While Edge Runtime's `server.close()`, which awaits all `waitUntil` promises, is called, but `exit-hook`'s callback in CLI doesn't wait for the `close` promise to resolve.

This PR fixes this by allowing an optional graceful shutdown callback instead of killing the process immediately.


craigandrews (dismissed):

packages/node/package.json
```diff
@@ -32,7 +32,7 @@
     "edge-runtime": "2.5.7",
     "esbuild": "0.14.47",
     "etag": "1.8.1",
-    "exit-hook": "2.2.1",
+    "exit-hook": "4.0.0",
```
cb1kenobi: I'm not sure we want to start introducing ESM dependencies quite yet. The CLI is still being transpiled to CommonJS and while `esbuild` does help with ESM compatibility, it hasn't been fully tested and there are known issues with dependencies that ESM features like `import.meta`.

Does `exit-hook@2.2.1` not work?

cb1kenobi:
Can you please add a test?

packages/node/package.json
Ethan-Arrowood: agreed

TooTallNate: I think this should be evaluated on a case by case basis. Seems like the tests are passing, so I'm fine with this update, especially if it's necessary to get the relevant hook for this change.

nuta: > Does exit-hook@2.2.1 not work?

Unfortunately no as we need a newer version to use `asyncExitHook` API. I have another idea to fix this bug without upgrading this dependency. Will try it first 👍

nuta:
Converted to draft until I finish rewriting this PR.

nuta:
Rewrote the PR without using `exit-hook` package, with a test 👍 @cb1kenobi

Ethan-Arrowood (dismissed):
very cool

Ethan-Arrowood (dismissed):

EndangeredMassa (dismissed):

nuta:
^ the GitHub history got messed up but TL;DR: Fix merge conflicts in `pnpm-lock.yaml` and force-pushed my changes again.

EndangeredMassa (dismissed):

nuta:
@EndangeredMassa Can I get an approval again? 🙏 I've added a error handling (https://github.com/vercel/vercel/pull/10915/commits/35d839657f9a748871307f3a390fab7030994295) to handle uncaught exceptions in `waitUntil` promises.

EndangeredMassa (approved):

PR #10918, opened 2023-12-07
[build-utils] Add getPathForPackageManager 

ofhouse:
To use this outside of CLI we want a way to suppress the `console.log`s in `getEnvForPackageManager`.
For achieving this, we introduce a new helper `getPathForPackageManager()` which contains the core logic.

Best to review commit by commit + hide whitespace.

**X-Ref**
- [Internal Context](https://vercel.slack.com/archives/C03F2CMNGKG/p1701970097725689)


cb1kenobi:
Just curious, did you consider using https://www.npmjs.com/package/debug?

ofhouse:
@cb1kenobi Yes, but I think it's intended to do a `console.log` during the build, so we don't want to set logging levels just to see this message right?

([Better context](https://vercel.slack.com/archives/C03F2CMNGKG/p1701970097725689) why we wanna add that)

cb1kenobi:
`debug` does log to stderr. All you're setting is the `DEBUG=<filter>` env var.

Ethan-Arrowood (dismissed):

ofhouse:
> debug does log to stdout.

Indeed yes, but this not a debug message, it's part of the build output during a build right?
Screenshot 2023-12-07 at 18 33 38

Using `debug` here with `DEBUG=` would have the side effect that we would show all debug messages in the build-log I think?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -521,11 +521,16 @@ export function getEnvForPackageManager({
...
+  showLogs = true,
 }: {
   cliType: CliType;
   lockfileVersion: number | undefined;
   nodeVersion: NodeVersion | undefined;
   env: { [x: string]: string | undefined };
+  /**
+   * Whether to print out on console what lockfile was detected.
+   */
+  showLogs?: boolean | undefined;
```
EndangeredMassa: **suggestion (non-blocking):** change side effects and return type

I think this would work better if the `getEnvForPackageManager` returned the new `PATH` and `YARN_NODE_LINKER` values, then the consumer would slap those onto the current env vars and log whatever it wants.

If you don't want to change the signature and side effects of functions that cross the package boundary, I suggest extracting the logic that determines the `PATH` and `YARN_NODE_LINKER` values into a new function that is also exported, allowing both options for consumer code.‏

ofhouse:
@EndangeredMassa I picked up your suggestion and created a new helper called `getPathForPackageManager()` that contains the main logic.

I kept `getEnvForPackageManager()` which I refactored to use `getPathForPackageManager()` internally, since it's used in many places I want to leave it to the team wether to replace the usages later or not.

TooTallNate (approved):

PR #10919, opened 2023-12-07
[cli] add `--deprecated` option to `vc project ls` command

Ethan-Arrowood:
Adds a `--deprecated` option to the `vc project ls` command that will only show projects currently running on a soon-to-be-deprecated Node.js version.

It also adds additional output providing more information to the user about what versions and a link to our documentation so they can learn more.

Example:

Screenshot 2023-12-07 at 15 01 22


trek (approved):
Love it!

packages/cli/src/commands/project/list.ts
```diff
@@ -48,10 +51,38 @@ export default async function list(
...
+  if (deprecated) {
+    const upcomingDeprecationVersions = NODE_VERSIONS.filter(
+      nodeVersion =>
+        nodeVersion.discontinueDate &&
+        nodeVersion.discontinueDate.valueOf() > Date.now()
+    );
+    const upcomingDeprecationVersionsList = upcomingDeprecationVersions.map(
+      nodeVersion => nodeVersion.range
+    );
+    console.log(projectList);
```
cb1kenobi, line 63: Remove debug `console.log()`.

packages/cli/src/commands/project/list.ts
```diff
@@ -48,10 +51,38 @@ export default async function list(
...
+    projectList = projectList.filter(
+      project =>
+        project.nodeVersion &&
+        upcomingDeprecationVersionsList.includes(project.nodeVersion)
+    );
+
+    output.log(
+      `The following Node.js versions will be deprecated soon: ${upcomingDeprecationVersionsList.join(
```
cb1kenobi, line 71: I feel this needs to be called out better. Either an `!` or colored yellow or something.

packages/cli/test/unit/commands/project.test.ts
```diff
@@ -49,6 +45,57 @@ describe('project', () => {
...
+      jest.useFakeTimers().setSystemTime(new Date('2023-12-08'));
+
+      const user = useUser();
+      useTeams('team_dummy');
+      const project = useProject({
+        ...defaultProject,
+        nodeVersion: '16.x',
+      });
+
+      console.log(project);
```
EndangeredMassa, line 58: remove?

PR #10920, opened 2023-12-07
detect middleware.mjs and middleware.mts

EndangeredMassa:
WIP


EndangeredMassa:
Should we also update https://github.com/vercel/vercel/blob/b67b97023b884a5a37fad420945979e2683d1e6f/packages/node/src/index.ts#L307 ?

PR #10922, opened 2023-12-07
[remix] Reinstall dependencies during `prepareCache()`

TooTallNate:
https://github.com/vercel/vercel/pull/10819 introduced a bug causing the `prepareCache()` function to fail (due to `@remix-run/dev` no longer being require-able). Even if it were not failing, the deps installed are not a valid representation of the user's `package.json` / lockfile. So to have more truthful cache contents, run `npm install` once again during `prepareCache()` to fix both issues at the same time.


trek (approved):
Not blocking since this fixes a bug, but is there a potential test we could add later to avoid a regression?

packages/remix/src/prepare-cache.ts
```diff
@@ -7,10 +12,32 @@ export const prepareCache: PrepareCache = async ({
...
+    undefined,
+    config
+  );
+  const spawnOpts = getSpawnOptions({}, nodeVersion);
+  await runNpmInstall(
+    entrypointFsDirname,
+    [],
+    {
+      ...spawnOpts,
+      stdio: 'ignore',
```
EndangeredMassa, line 35: **question (non-blocking):** ‏can we still log this somewhere, maybe debug only?

cb1kenobi:
**question (non-blocking):** ‏Why is `@remix-run/dev` no longer require-able?

packages/remix/src/build.ts
```diff
@@ -104,15 +104,17 @@ export const build: BuildV2 = async ({
...
+  const { cliType, packageJsonPath, lockfileVersion, lockfilePath } =
+    await scanParentDirs(entrypointFsDirname);
 
   if (!packageJsonPath) {
     throw new Error('Failed to locate `package.json` file in your project');
   }
 
-  const pkgRaw = await fs.readFile(packageJsonPath, 'utf8');
+  const [lockfileRaw, pkgRaw] = await Promise.all([
+    lockfilePath ? fs.readFile(lockfilePath) : null,
```
cb1kenobi, line 115: Lock files can get pretty big. Maybe copy the file instead of keeping it in memory?

packages/remix/src/prepare-cache.ts
TooTallNate: The problem is that `spawnAsync` does not return the child project instance. Maybe we can tweak that though.

TooTallNate:
> **question (non-blocking):** ‏Why is `@remix-run/dev` no longer require-able?

@cb1kenobi See https://github.com/vercel/vercel/pull/10819.

packages/remix/src/build.ts
TooTallNate: Perhaps. I think we should wait and see if it would cause an issue for anyone. Machines tend to have a decent amount of memory these days.

cb1kenobi (approved):

PR #10925, opened 2023-12-08
[cli] respect install command

Kikobeats:
Although I have this setting in my project:

CleanShot 2023-12-08 at 13 27 52@2x

pnpm will be never used unless my project doesn't have a lockfile present. Although I can understand the decision there, it's totally fine to run pnpm without lockfile (in fact, my build is failing because it's using yarn, but working flawless installing via pnpm), but even better, it allows any user to run the desired package manager without being vendorlocking to yarn.

### DIFF
--- a/.changeset/mighty-pants-pull.md
+++ b/.changeset/mighty-pants-pull.md
@@ -1,0 +1,5 @@
+---
+"@vercel/build-utils": patch
+---
+
+[cli] respect install command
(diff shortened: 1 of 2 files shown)


cb1kenobi:
I'm not sure how I feel about this PR. I would prefer using `packageManager` in the `package.json` as a means to override the lockfile-based detection.

TooTallNate:
Can you explain a bit more how this situation is triggered? The install command takes precedence over the auto-detected package manager, _but only for the frontend builders_ (`next`, `static-build`, etc.). So I'm assuming your situation is without a frontend at all? Just API endpoints?

trek (approved):

PR #10932, opened 2023-12-11
[cli] double page limit for vc project ls --deprecated

Ethan-Arrowood:
As titled. Increases api page size for `vc project ls` when `--deprecated` is used. 

This is a temporary workaround since we do the filtering "client-side". We will eventually update the API itself to support this and then we can remove the client-based filter and reduce the page limit back to the ideal `20`.


cb1kenobi (dismissed):

packages/cli/src/commands/project/list.ts
```diff
@@ -28,15 +28,15 @@ export default async function list(
 
   output.spinner(`Fetching projects in ${chalk.bold(contextName)}`);
 
-  let projectsUrl = '/v4/projects/?limit=20';
+  const deprecated = argv['--deprecated'] || false;
+
+  let projectsUrl = `/v4/projects/?limit=${deprecated ? 40 : 20}`;
```
EndangeredMassa, line 35: **suggestion (non-blocking):** ‏add a comment about why

My future self would like to understand why this difference is here. Can we specify?

```suggestion
  // use a larger limit when `--deprecated`
  // because client-side filtering produces fewer results
  let projectsUrl = `/v4/projects/?limit=${deprecated ? 40 : 20}`;
```

EndangeredMassa (dismissed):

trek (approved):

PR #10937, opened 2023-12-13
[cli] log extension execution failures

mrmckeb:
We've had some reports of random extension failures with no output.

Our theory is that these are possibly _execution_ failures. This change should help us identify any such errors.

_Discussed with @TooTallNate._


TooTallNate (dismissed):
LGTM

TooTallNate (approved):

cb1kenobi (approved):

Ethan-Arrowood (approved):
Any way to test this? We can verify things log on failure. May be useful here.

PR #10938, opened 2023-12-13
[cli] Use new `deprecated` query param in projects api for `vc project ls --deprecated`

Ethan-Arrowood:
Uses the new `deprecated` query param for the GET /projects API instead of client-side filtering.

This enables a better UX so that results are properly paginated. 

Here are some example runs:

Pagination works as expected:
Screenshot 2023-12-12 at 20 59 57

Usage without the flag results in different output (as expected!)
Screenshot 2023-12-12 at 21 04 52

### DIFF
--- a/.changeset/clever-peaches-allow.md
+++ b/.changeset/clever-peaches-allow.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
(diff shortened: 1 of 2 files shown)


EndangeredMassa (approved):
I test this manually and it looks good. Thanks!

PR #10956, opened 2023-12-14
[cli] Fix `vc bisect` selecting too many deployments

TooTallNate:


EndangeredMassa (approved):
can we test this?

TooTallNate:
Technically, yes, but it would involve setting up a pagination system in the API mocks, which is a bit involved. Will skip this time.

PR #10965, opened 2023-12-18
[cli] replace `--deprecated` with `--update-required` in `vc project ls`

Ethan-Arrowood:
Replaces the new `--deprecated` option with `--update-required` in the `vc project ls` command.

### DIFF
--- a/.changeset/sweet-cougars-judge.md
+++ b/.changeset/sweet-cougars-judge.md
@@ -1,0 +1,5 @@
+---
+"vercel": major
+---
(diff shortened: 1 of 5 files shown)


trek (dismissed):

EndangeredMassa:
We should make this a new major because we're technically removing a flag.

EndangeredMassa (dismissed):
tested manually and it works!

TooTallNate (approved):

PR #10975, opened 2023-12-19
[deps] Change use of isbot next version

omrilotan:
Instead of using "latest", using a specific major version can prevent future breakages.

[isbot](https://github.com/omrilotan/isbot/blob/next/CHANGELOG.md#400) version 4 aims to be more performant at load time, better for tree-shaking and Typescript friendly. The default import was replaced with a named import.


omrilotan:
Fixed on https://github.com/remix-run/remix/pull/8415

PR #10978, opened 2023-12-19
[next] App Pages Prerender Manifest Update

wyattjoh:
Previously routes that did not have a `dataRoute` key in the `prerender-manifest.json` would be treated as an App Route. The logic has been updated (for partial prerendering support) to also consider the new `prefetchDataRoute`. Entries with either of these keys are treated as an App Page instead of an App Route.

This also addressed the scenerio where a app route (`route.ts`) with a dynamic segment (`/api/[slug]/route.ts`) which doesn't emit a `.body` during build doesn't cause the build to fail by checking for the file first.


wyattjoh:
Current dependencies on/for this PR:
* `main`
  * **PR #10978** <a href="https://app.graphite.dev/github/pr/vercel/vercel/10978?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/vercel/vercel/10978?utm_source=stack-comment).

ztanner (approved):

wyattjoh:
A test case for this will land after the following also lands: https://github.com/vercel/next.js/pull/59780

PR #10980, opened 2023-12-19
Remove unneeded global directives

trek:
An update to latest `turbo` somehow made these no longer required? 🤷


EndangeredMassa:
The lint failure said that `htmlFsRef` was referenced before it was declared, which doesn't appear to be true? Updating and checking.

trek:
@EndangeredMassa I'm still digging into it and does not appear to be true. I'm quite confused.

EndangeredMassa (approved):

ijjk (approved):

cb1kenobi (approved):

PR #10984, opened 2023-12-21
[examples] Fix `ERR_PNPM_OUTDATED_LOCKFILE` when deploying SvelteKit example

lucleray:
Currently, [the SvelteKit (v1)](https://vercel.com/new/clone?b=main&s=https%3A%2F%2Fgithub.com%2Fvercel%2Fvercel%2Ftree%2Fmain%2Fexamples%2Fsveltekit-1&showOptionalTeamCreation=false&template=sveltekit-1&teamCreateStatus=hidden) example fails with:
```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with package.json
```

This fixes the lockfile, and therefore fixes the example.

---

As a side-quest, I also fixed the "unmet peer" warnings:
```
└─┬ @vercel/style-guide 4.0.2
  ├── ✕ unmet peer eslint@^8.24.0: found 8.14.0
  ├── ✕ unmet peer prettier@^2.7.0: found 2.6.2
  └─┬ eslint-plugin-unicorn 43.0.2
    └── ✕ unmet peer eslint@>=8.18.0: found 8.14.0
```


ofhouse (dismissed):

trek (approved):

mglagola (approved):

leerob (approved):

matheuss (approved):

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
```
styfle, line 5: @lucleray Looks like you forgot to run `corepack enable pnpm` on your machine because this was likely using a different version of pnpm.

lucleray: I think I did 🤔 

I'll double check though.

lucleray: Ok, I think you were right. I've enabled corepack and I can "cleanly" install without altering the pnpm-lock.yaml file.

That said, I don't think these added lines caused any trouble for the examples.

lucleray: I also noticed something else: when cloning the examples, the pnpm version used currently is `8.11.0` (not `8.3.1` like defined in the root package.json).

(That works as expected, because only the subfolder is cloned)

We should probably upgrade the pnpm version of the repo 🤔

PR #10991, opened 2023-12-21
chore: update node_version and nvmrc to node 18

arnabsen:
Update to latest Node LTS supported by Vercel (non beta) - https://vercel.com/changelog/node-js-18-lts-is-now-available


cb1kenobi:
Thank you for the pull request, however we still support Node 16 and will need to wait until 16 has been EOL before we switch to Node 18 as the default.

PR #11010, opened 2024-01-02
[build-utils] Deprecate `EdgeFunction#name` property

TooTallNate:
The `name` property of the `EdgeFunction` class is no longer necessary on the infra side. Instead, its value is inferred based on the URL path that the function is representing. So deprecate the property on the class, and remove its usage throughout the codebase.


.changeset/seven-carrots-begin.md
```diff
@@ -0,0 +1,9 @@
+---
+'@vercel/build-utils': minor
+'@vercel/static-build': patch
+'@vercel/hydrogen': patch
+'@vercel/remix-builder': patch
+'@vercel/node': patch
+---
```
EndangeredMassa, line 7: **suggestion (soft-blocking):** remove usage from `packages/next/src/utils.ts`

https://github.com/vercel/vercel/blob/21c700aa93916fbb97c3d2778839b4b67f042b91/packages/next/src/utils.ts#L2901

EndangeredMassa (approved):

.changeset/seven-carrots-begin.md
TooTallNate: I started to, but it was a bit more complicated. Will ask @ijjk @ztanner to take a look in a follow-up PR.

Ethan-Arrowood (approved):

PR #11027, opened 2024-01-08
[cli] Serialize duplicate `EdgeFunction` references as symlinks in `vc build`

TooTallNate:
Enables the symlink optimization that currently exists for `Lambda` instances, but now for `EdgeFunction` instances as well. This will be particularly beneficial for Remix applications which use edge functions for many routes, since they will now all be represented by the same underling function in production.

### DIFF
--- a/.changeset/large-baboons-train.md
+++ b/.changeset/large-baboons-train.md
@@ -1,0 +1,5 @@
+---
+'vercel': minor
+---
+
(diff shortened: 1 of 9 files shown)


cb1kenobi (dismissed):
LGTM!

EndangeredMassa (approved):

cb1kenobi (approved):

PR #11033, opened 2024-01-10
[cli] Add support for sensitive env var

anatrajkovska:
This PR updates the `vercel env add` command to support a new option called `--sensitive` that allows an env var to be created of type `sensitive`.


marcgreenstock (commented):
Can you add some information on the option in the command definition here: https://github.com/vercel/vercel/blob/5ef02a2ed38e3713e37e58f8dbd61acdec431ad3/packages/cli/src/commands/env/command.ts#L37

Take a look at other commands e.g. https://github.com/vercel/vercel/blob/5ef02a2ed38e3713e37e58f8dbd61acdec431ad3/packages/cli/src/commands/login/command.ts#L13 to see how to structure it.

marcgreenstock (approved):

packages/cli/test/integration-2.test.ts
```diff
@@ -681,6 +681,53 @@ test('vercel env with unknown `VERCEL_ORG_ID` or `VERCEL_PROJECT_ID` should erro
...
 });
 
+test('add a sensitive env var', async () => {
+  const dir = await setupE2EFixture('project-sensitive-env-vars');
+  const projectName = `project-sensitive-env-vars-${
+    Math.random().toString(36).split('.')[1]
+  }`;
+
+  // remove previously linked project if it exists
+  await remove(path.join(dir, '.vercel'));
```
EndangeredMassa, line 691: **suggestion (non-blocking):** ‏move to cleanup step

Can we clean this up in an `after` or `try`-`finally` instead? I'd prefer to have the state clean in between runs to avoid issues where this kind of cleanup can blind us to misunderstood behavior in the current test if changes are made to it in the future.

packages/cli/test/integration-2.test.ts
```diff
@@ -681,6 +681,53 @@ test('vercel env with unknown `VERCEL_ORG_ID` or `VERCEL_PROJECT_ID` should erro
   expect(output.stderr).toContain('Project not found');
 });
 
+test('add a sensitive env var', async () => {
+  const dir = await setupE2EFixture('project-sensitive-env-vars');
+  const projectName = `project-sensitive-env-vars-${
+    Math.random().toString(36).split('.')[1]
+  }`;
```
EndangeredMassa, line 688: **suggestion (soft-blocking):** remove randomness in tests

In this or a follow-up PR: Can we remove the randomness here? I'm sure the chance of collision is astronomical, but it's possible, especially if some test pollution elsewhere mocks out `Math.random`.‏

EndangeredMassa (approved):

packages/cli/test/integration-2.test.ts
anatrajkovska: Thanks @EndangeredMassa, I will handle it in a follow up PR.

packages/cli/test/integration-2.test.ts
anatrajkovska: Same here, I will handle it in a follow up PR.

PR #11050, opened 2024-01-16
Use `worker.name` instead of edge function name to fix type error in `@vercel/next`

Ethan-Arrowood:
Uses another `name` property instead of the now deprecated one.

It is my understanding that the fixtures: https://github.com/vercel/vercel/tree/main/packages/next/test/fixtures/00-app-dir-edge (app dir) and https://github.com/vercel/vercel/tree/main/packages/next/test/fixtures/00-edge-runtime (pages) cover this code.

### DIFF
--- a/.changeset/rich-lemons-matter.md
+++ b/.changeset/rich-lemons-matter.md
@@ -1,0 +1,5 @@
+---
+"@vercel/next": patch
+---
+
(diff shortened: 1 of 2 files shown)


ijjk (approved):
Thanks!

PR #11060, opened 2024-01-18
[cli] Emit "filePathMap" in `vc-config.json` for `FileFsRef` instances

TooTallNate:
This is a performance / bandwidth / storage space optimization for the **Build Output API**.

When a Builder returns function which contains a file that is a `FileFsRef` instance, it _won't_ be copied into the corresponding `.func` directory, but instead will be added to a `"files"` mapping in the `.vc-config.json` file. This mapping represents keys which are destination file paths within the function filesystem, and values which are relative paths from the root of the project codebase. This allows for common files (i.e. `node_modules`) to be referenced (instead of physically copied on the disk).

This introduces an additional complexity to `vc deploy --prebuilt`, in such that now it needs to build up the list of `"files"` referenced from the `.vc-config.json` files, and _also_ upload those file paths along with the `.vercel/output` directory.

Depends on:

* https://github.com/vercel/vercel/pull/11077
* https://github.com/vercel/vercel/pull/11080


EndangeredMassa (dismissed):
This approach looks good to me. Can we test it?

trek (approved):

PR #11061, opened 2024-01-18
Add retry handling when hitting SSO

ijjk:
This aims to reduce flakiness when SSO disabling for a test project is still disabling/propagating. 

x-ref: https://github.com/vercel/vercel/actions/runs/7563953104/job/20597320056?pr=11059
x-ref: https://github.com/vercel/vercel/actions/runs/7563953104/job/20597322078?pr=11059

### DIFF
--- a/test/lib/deployment/fetch-retry.js
+++ b/test/lib/deployment/fetch-retry.js
@@ -23,7 +23,7 @@ async function fetchRetry(url, ...rest) {
               requestIds.push(res.headers.get('x-vercel-id'));
               if (i === 0) {
                 console.error(
-                  `request ids: `,
+                  `Failed request ids (because of 401s): `,
                   JSON.stringify(requestIds, null, 2)
                 );
                 throw new Error(
--- a/test/lib/deployment/fetch-retry.js
+++ b/test/lib/deployment/fetch-retry.js
@@ -14,7 +14,7 @@ async function fetchRetry(url, ...rest) {
       try {
         const res = await fetch(url, ...rest);
 
-        if (res.status === 401 && retryIndex < 2) {
+        if (res.status === 401 && retryIndex < 4) {
           const error = new Error('sso error');
           error.type = 'sso-error';
           throw error;
@@ -43,7 +43,7 @@ async function fetchRetry(url, ...rest) {
         throw error;
       }
     },
-    { factor: 2, retries: 3 }
+    { factor: 2, retries: 5 }
   );
 }
 
--- a/test/lib/deployment/fetch-retry.js
+++ b/test/lib/deployment/fetch-retry.js
@@ -1,5 +1,4 @@
 const fetch = require('node-fetch');
-const { disableSSO } = require('./now-deploy.js');
 const retryBailByDefault = require('./retry-bail-by-default.js');
 
 const ABSOLUTE_URL_PATTERN = /^https?:\/\//i;
@@ -27,11 +26,6 @@ async function fetchRetry(url, ...rest) {
                 `Failed to fetch ${url}, received 401 status for over 1 minute`
               );
             }
-
-            if (i % 10 === 0) {
-              await disableSSO(url);
-            }
-
             await new Promise(resolve => setTimeout(resolve, 1000));
           } else {
             return res;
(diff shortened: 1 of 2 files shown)


TooTallNate (dismissed):

test/lib/deployment/now-deploy.js
```diff
@@ -3,11 +3,13 @@ const assert = require('assert');
...
 const path = require('path');
 const _fetch = require('node-fetch');
-const fetch = require('./fetch-retry.js');
+const fetch = require('./fetch-retry');
 const fileModeSymbol = Symbol('fileMode');
 const { logWithinTest } = require('./log');
 const ms = require('ms');
 
+console.error({ fetch });
+
```
TooTallNate:
```suggestion
```

test/lib/deployment/now-deploy.js
```diff
@@ -149,10 +149,14 @@ async function disableSSO(deploymentId, useTeam = true) {
...
-      if (res.status !== 401) {
-        break;
+    for (let i = 0; i < 10; i++) {
+      try {
+        const res = await fetch(`https://${deploymentUrl}`);
+        if (res.status !== 401) {
+          break;
+        }
+      } catch (err) {
+        console.error(err);
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
        console.error(err);
        break;
```

The try-catch changes the control flow of this function in ways that I think will be confusing in the log outputs. If `fetch` throws, the error will be logged, but then it will wait 5s and print out "Disabled deployment  protection...", which isn't true.

Why was the try-catch added?

test/lib/deployment/fetch-retry.js
```diff
@@ -11,7 +11,33 @@ async function fetchRetry(url, ...rest) {
...
+
+          if (res.status === 401) {
+            const clonedRes = res.clone();
+            const body = await clonedRes.text();
+
+            if (body.includes('https://vercel.com/sso-api')) {
+              requestIds.push(res.headers.get('x-vercel-id'));
+              if (i === 0) {
+                console.error(
+                  `request ids: `,
```
EndangeredMassa: **suggestion (non-blocking):** ‏

```suggestion
                  `Failed request ids (because of 401s): `,
```

test/lib/deployment/fetch-retry.js
```diff
@@ -11,7 +11,33 @@ async function fetchRetry(url, ...rest) {
...
+                );
+              }
+            } else {
+              return res;
+            }
+            await new Promise(resolve => setTimeout(resolve, 1000));
+          } else {
+            return res;
+          }
+        }
```
EndangeredMassa, line 40: **Note (non-blocking):** We have two different retry mechanisms intermixed here. The desired effect seems to be that we can retry general requests up to 3 (now 5) times, but for 401 responses specifically, we retry (within a given retry of that larger limit) 60 times.

I don't think we can encode this behavior into the existing retry mechanism clearly, but I do think this bit of weirdness (mixing the strategies) is a flag for thinking about it some more. Is there a better way to accomplish this that's just as reliable, more flexible, and less confusing?

test/lib/deployment/fetch-retry.js
```diff
@@ -33,7 +59,7 @@ async function fetchRetry(url, ...rest) {
         throw error;
       }
     },
-    { factor: 2, retries: 3 }
+    { factor: 2, retries: 5 }
```
EndangeredMassa: **question (non-blocking):** ‏Why did we increase the general retry count? This won't impact the root problem, right?

EndangeredMassa (dismissed):

EndangeredMassa:
I took off automerge just to give a chance to review my comments. I did approve this. Feel free to merge if you are good with the current state.

Thanks @ijjk!

test/lib/deployment/fetch-retry.js
ijjk: Oh this was left-over from previous retry handling we can revert it now

ijjk:
```suggestion
    { factor: 2, retries: 3 }
```

EndangeredMassa (approved):

PR #11062, opened 2024-01-18
Use production deploys for testing

ijjk:
Alternative to https://github.com/vercel/vercel/pull/11061 this uses production deploys for testing by default which avoids the need for disabling deployment protection for every test project and hopefully reduces flakiness in tests. 

x-ref: https://github.com/vercel/vercel/actions/runs/7563953104/job/20597320056?pr=11059
x-ref: https://github.com/vercel/vercel/actions/runs/7563953104/job/20597322078?pr=11059

### DIFF
--- a/test/lib/deployment/now-deploy.js
+++ b/test/lib/deployment/now-deploy.js
@@ -85,7 +85,7 @@ async function nowDeploy(projectName, bodies, randomness, uploadNowJson, opts) {
 
   for (let i = 0; i < 750; i += 1) {
     const deployment = await deploymentGet(deploymentId);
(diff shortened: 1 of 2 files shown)


test/lib/deployment/now-deploy.js
```diff
@@ -69,10 +71,15 @@ async function nowDeploy(projectName, bodies, randomness, uploadNowJson, opts) {
 
   {
     const json = await deploymentPost(nowDeployPayload, opts);
+    console.log(JSON.stringify(json, null, 2));
```
ijjk:
```suggestion
```

PR #11078, opened 2024-01-22
[static-build][tests] add dependabot to check for dep updates in static build

trek:
Adding dependabot for our static-build test fixtures to help us keep these examples up-to-date. Only includes latest majors of fixtures that use registries that GitHub tracks.


trek:
Need to add https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups to avoid getting 1 PR _per_ dependency

PR #11090, opened 2024-01-24
[static-build] Add a new Astro@4 fixture

trek:
Add a new Astro@4 fixture. We were missing this one entirely.


packages/static-build/test/fixtures/astro-v4/probes.json
```diff
@@ -0,0 +1,5 @@
+{
+  "probes": [
+    { "path": "/", "mustContain": "Welcome to Astro" }
```
EndangeredMassa, line 3: **suggestion (soft-blocking):** Is there a reasonable way to test the "Incremental Content Caching" feature too?
https://astro.build/blog/astro-4/

EndangeredMassa (approved):

PR #11094, opened 2024-01-25
add ruby3 to path during build

EndangeredMassa:
Add ruby to the path plus some debug code. I'll remove the debug code after we verify this works.

### DIFF
--- a/packages/ruby/vc_init.rb
+++ b/packages/ruby/vc_init.rb
@@ -82,7 +82,10 @@ def vc__handler(event:, context:)
   path = payload['path']
   headers = payload['headers']
 
-  puts headers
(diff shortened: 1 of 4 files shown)


packages/ruby/vc_init.rb
```diff
@@ -81,6 +81,9 @@ def vc__handler(event:, context:)
   payload = JSON.parse(event['body'])
   path = payload['path']
   headers = payload['headers']
+
+  puts headers
```
trek: Wrap in some env var so we can trigger this just for us.

trek (dismissed):

TooTallNate (dismissed):

PR #11098, opened 2024-01-26
[cli] Add Variants to Build Output API

AndyBitz:
Makes adjustments to replace `flags` with `variants`.

Also marks the current `flags` implementation as deprecated, as it should get removed soon. Which I'll do in a follow up PR.

### DIFF
--- a/packages/build-utils/src/types.ts
+++ b/packages/build-utils/src/types.ts
@@ -440,7 +440,9 @@ export interface Cron {
   schedule: string;
 }
 
-// TODO: Proper description once complete
+/**
+ * @deprecated Replaced by Variants. Remove once fully replaced.
+ */
 export interface Flag {
   key: string;
   defaultValue?: unknown;
@@ -471,7 +473,9 @@ export interface BuildResultV2Typical {
   framework?: {
     version: string;
   };
+  /** @deprecated Replaced by Variants. Remove once fully replaced. */
   flags?: Flag[];
+  variants?: Record<string, VariantDefinition>;
 }
 
 export type BuildResultV2 = BuildResultV2Typical | BuildResultBuildOutput;
@@ -491,3 +495,28 @@ export type ShouldServe = (
 export type StartDevServer = (
   options: StartDevServerOptions
 ) => Promise<StartDevServerResult>;
+
+/**
+ * TODO: The following types will eventually be exported by a more
+ *       relevant package.
+ */
+type VariantJSONArray = ReadonlyArray<VariantJSONValue>;
+
+type VariantJSONValue =
+  | string
+  | boolean
+  | number
+  | null
+  | VariantJSONArray
+  | { [key: string]: VariantJSONValue };
+
+type VariantOption = {
+  value: VariantJSONValue;
+  label?: string;
+};
+
+export interface VariantDefinition {
+  options?: VariantOption[];
+  url?: string;
+  description?: string;
+}
--- a/packages/cli/src/commands/build/index.ts
+++ b/packages/cli/src/commands/build/index.ts
@@ -843,9 +843,12 @@ async function writeVariantsJson(
 ): Promise<void> {
   const variantsFilePath = join(outputDir, 'variants.json');
 
+  let hasVariants = true;
+
   const variants = (await fs.readJSON(variantsFilePath).catch(error => {
     if (error.code === 'ENOENT') {
-      return { defintions: {} };
+      hasVariants = false;
+      return { definitions: {} };
     }
 
     throw error;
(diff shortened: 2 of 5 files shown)


dferber90 (dismissed):

TooTallNate (dismissed):

packages/cli/src/commands/build/index.ts
```diff
@@ -826,6 +832,43 @@ function mergeFlags(
...
+
+    for (const [key, defintion] of Object.entries(result.variants)) {
+      if (result.variants[key]) {
+        output.warn(
+          `The variant "${key}" was found multiple times. Only its first occurrence will be considered.`
+        );
+        continue;
+      }
+
+      result.variants[key] = defintion;
```
EndangeredMassa: **suggestion (blocking):** should this be saving to `variants` for writing later?‏

```suggestion
      variants[key] = defintion;
```

packages/cli/src/commands/build/index.ts
```diff
@@ -826,6 +832,43 @@ function mergeFlags(
...
+): Promise<void> {
+  const variantsFilePath = join(outputDir, 'variants.json');
+
+  const variants = (await fs.readJSON(variantsFilePath).catch(error => {
+    if (error.code === 'ENOENT') {
+      return { defintions: {} };
+    }
+
+    throw error;
+  })) as { definitions: Record<string, VariantDefinition> };
```
EndangeredMassa: **suggestion (soft-blocking):** simply type usage

In this or a follow-up PR: Can we remove the type assertion? If not, can we simplify this complex expression into a couple separate statements?‏

packages/cli/src/commands/build/index.ts
```diff
@@ -826,6 +832,43 @@ function mergeFlags(
   });
 }
 
+/**
+ * Takes the build output and writes all the variants into the `variants.json`
+ * file. It'll skip variants that already exist.
+ */
+async function mergeVariants(
```
EndangeredMassa: **suggestion (non-blocking):** rename to something like `writeVariantsJson`

This would make it clearer that (1) this doesn't work like the other `merge*` functions here and (2) is writing out a separate file, not building up data for the main `config.json` file.‏

EndangeredMassa (changes requested):

packages/cli/src/commands/build/index.ts
AndyBitz: Oh, thanks. That was an oversight.
Will add tests for this in upcoming PRs.

packages/cli/src/commands/build/index.ts
AndyBitz: The types will come from a different package eventually, so yup, will make sure to update those once it's ready

packages/cli/src/commands/build/index.ts
AndyBitz: Sounds good, done!

EndangeredMassa (dismissed):

dferber90 (approved):

packages/cli/src/commands/build/index.ts
```diff
@@ -826,6 +832,51 @@ function mergeFlags(
...
+      }
+
+      hasVariants = true;
+      variants.definitions[key] = defintion;
+    }
+  }
+
+  // Only create the file when there are variants to write,
+  // or when the file already exists.
+  if (hasVariants) {
```
EndangeredMassa, line 875: **suggestion (non-blocking):** ‏remove need for `hasVariants`

```suggestion
  if (Object.keys(variants.definitions).length) {
```

Can we check if there are any variants in the data structure instead? As this code changes, it will be easy to forget to set `hasVariants` somewhere, but as long as `variants.definitions` is updated, this check will work.

EndangeredMassa (approved):

packages/cli/src/commands/build/index.ts
AndyBitz: It would be possible now, but we might add more properties next to `definitions` that we'd need to check then. That's why the initial check is `true`.

Though, I'll see if I can improve it in follow up PRs.

packages/build-utils/src/types.ts
```diff
@@ -491,3 +495,28 @@ export type ShouldServe = (
...
+  | { [key: string]: VariantJSONValue };
+
+type VariantOption = {
+  value: VariantJSONValue;
+  label?: string;
+};
+
+export interface VariantDefinition {
+  options?: VariantOption[];
+  url?: string;
```
dferber90, line 520: @AndyBitz let's rename this property to "origin" so we can in the future extend it to something like

`origin: string | { name: string, url: string }`, which would not make sense if it's called "url"

dferber90, line 520: @AndyBitz let's rename this property to "origin" so we can in the future extend it to something like

`origin: string | { name: string, url: string }`, which would not make sense if it's called "url"

dferber90 (changes requested):

packages/build-utils/src/types.ts
AndyBitz: @dferber90 I'll do that in a follow up PR so that we can get the initial changes out sooner.

The types must come from a different package eventually anyways, so they shouldn't be blocking here.

dferber90 (approved):

PR #11099, opened 2024-01-26
Fix index normalizing for app outputs

ijjk:
This ensures we don't apply the index output normalizing for app paths which was previously already gated for prerenders but not all routes.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1130,7 +1130,7 @@ export async function serverBuild({
         }
         let outputName = path.posix.join(entryDirectory, pageNoExt);
 
-        if (!(group.isAppRouter || group.isAppRouteHandler)) {
+        if (!group.isAppRouter && !group.isAppRouteHandler) {
           outputName = normalizeIndexOutput(outputName, true);
         }
 
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1128,11 +1128,11 @@ export async function serverBuild({
             );
           });
         }
+        let outputName = path.posix.join(entryDirectory, pageNoExt);
 
-        let outputName = normalizeIndexOutput(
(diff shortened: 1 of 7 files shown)


packages/next/src/server-build.ts
```diff
@@ -1128,11 +1128,11 @@ export async function serverBuild({
             );
           });
         }
+        let outputName = path.posix.join(entryDirectory, pageNoExt);
 
-        let outputName = normalizeIndexOutput(
-          path.posix.join(entryDirectory, pageNoExt),
-          true
-        );
+        if (!(group.isAppRouter || group.isAppRouteHandler)) {
```
ztanner: maybe a skill issue 😆  but I usually have to think a bit extra on these conditions

```suggestion
        if (!group.isAppRouter && !group.isAppRouteHandler) {
```

packages/next/src/utils.ts
```diff
@@ -2051,7 +2051,7 @@ export const onPrerenderRoute =
       } = pr);
     }
 
-    let isAppPathRoute = false;
+    let isAppPathRoute = !!experimentalPPR;
```
ztanner: I think leaving this as `false` initially and adding the `experimentalPPR` check to the line below is a bit clearer

packages/next/src/utils.ts
```diff
@@ -2255,15 +2254,20 @@ export const onPrerenderRoute =
...
           ? outputPathPageOrig
           : path.posix.join(
               entryDirectory,
               srcRoute === '/' ? '/index' : srcRoute
-            ),
-        isServerMode
-      );
+            );
+
+      if (!isAppPathRoute && !experimentalPPR) {
```
ztanner: if we're setting `isAppPathRoute` to true when `experimentalPPR` is true, could this check just be `!isAppPathRoute`?

ztanner (approved):

PR #11104, opened 2024-01-29
Default ruby to only currently supported version (3.2.0)

trek:
We're currently using the directory name at `./vendor/bundle/ruby/`[0] to determine ruby version. This directory is created as part of running `bundle install` during the build process. However, when multiple ruby versions are available, you can end up with multiple directories and the 0th entry at `./vendor/bundle/ruby/` isn't guaranteed to be the Ruby version specified in the project.

Other companies in this space seem to do some order of 
 - check for value in `.ruby-version` file
 - check for value passed into `ruby` directive in the Gemfile
 - check the value in the `*.gemspec` file if the Gemfile has a `gemspec` directive.

We'll do that work in a future PR but for _now_ our only supported Ruby version is Ruby 3.2.x [as of 2023-11-22](https://vercel.com/changelog/upgrading-ruby-v2-7-to-v3-2). So this can be hard-coded for the moment.

Skipping ruby tests now since these target _current_ prod where Ruby 3.2.x is not in `PATH` until we redeploy after this gets deployed.


TooTallNate (approved):

packages/static-build/test/integration-setup.js
```diff
@@ -38,8 +38,14 @@ module.exports = function setupTests(groupIndex) {
     console.log('testing group', groupIndex, fixtures);
   }
 
+  const fixturesToSkip = ['jekyll-v3', 'jekyll-v4', 'middleman-v4'];
+
   // eslint-disable-next-line no-restricted-syntax
   for (const fixture of fixtures) {
+    if (fixturesToSkip.includes(fixture)) {
```
EndangeredMassa, line 45: I think we should have included a comment here about why this are skipped and when we plan to restore them.

Issue #11097, opened 2024-01-26
vercel@33.3.0 broke preBuilds in monorepo


TriPSs:
We are using the `preBuild` option in our monorepo but after updating to vercel@33.3.0 it no longer works, when trying to do the deploy we get the following error:
```
[debug] [2024-01-26T09:52:00.561Z] Error: Error: ENOENT: no such file or directory, lstat '/home/runner/work/<org>/<project>/dist/apps/<app-name>/node_modules/styled-jsx/index.js'

Error: ENOENT: no such file or directory, lstat '/home/runner/work/<org>/<project>/dist/apps/<app-name>/node_modules/styled-jsx/index.js'
```

When comparing it with the previous version it looks like the functions don't have the node_modules anymore. How can we solve this?

georgegebbett:
We've seen this too - we get the below error in Vercel when trying to deploy a prebuilt app

We are using Turborepo + pnpm which seems to be a fairly tricky setup for this kind of thing - our downfall is definitely coming from installing `vercel@latest` in CI as this isn't the first time our pipeline has been broken by a bad CLI release...

![image](https://github.com/vercel/vercel/assets/35996500/50e68437-533f-478b-8e8f-574966526c9e)

wilson-webdev:
Having the same issue with a GitHub action that uses `npx vercel`

andyvanosdale:
It looks like [commit 2b71ee6b427996dc26547a8601cf8b65c74a2ccd]( https://github.com/vercel/vercel/commit/2b71ee6b427996dc26547a8601cf8b65c74a2ccd#diff-5b7436deccee2ee3d2737e02d670d2e27c3fa229b1fea3646e7de3d5d33bf831R113) is checking the file paths in the `.vc-config.json` files to see if they are present. However, this incorrectly assumes that the `.next` directory was built out instead of using all the pre-built files. In essence, prebuilt is completely broken as a stand-alone feature, as documented in the (vercel build)[https://vercel.com/docs/cli/build] docs:

> The vercel build command can be used to build a Vercel Project locally or in your own CI environment. Build artifacts are placed into the .vercel/output directory according to the [Build Output API](https://vercel.com/docs/build-output-api/v3).
> 
> When used in conjunction with the vercel deploy --prebuilt command, this allows a Vercel Deployment to be created without sharing the Vercel Project's source code with Vercel.

andyvanosdale:
CC @TooTallNate for the commit

[...]

PR #11105, opened 2024-01-30
[client] More helpful error message when `vc deploy --prebuilt` has missing files

TooTallNate:
Print a more helpful error message for the scenario the user encountered here: https://github.com/orgs/vercel/discussions/5612

What was happening in they were running `vc build` in one GH Action job, and then `vc deploy --prebuilt` in a different job. The later job was does not have `node_modules` populated, causing the deploy command to fail:

Screenshot 2024-01-29 at 6 31 19 PM

The updated error message hints to the user that the `node_modules` should be installed for the command to succeed:

Screenshot 2024-01-29 at 6 30 54 PM

### DIFF
--- a/packages/client/src/create-deployment.ts
+++ b/packages/client/src/create-deployment.ts
@@ -1,5 +1,5 @@
 import { lstatSync } from 'fs-extra';
-import { isAbsolute, join, relative } from 'path';
+import { isAbsolute, join, relative, sep } from 'path';
 import { hash, hashes, mapToObject } from './utils/hashes';
 import { upload } from './upload';
 import { buildFileTree, createDebug } from './utils';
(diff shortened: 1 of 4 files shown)


packages/client/src/create-deployment.ts
```diff
@@ -90,27 +90,38 @@ export default function buildCreateDeployment() {
...
+              names: [join(workPath, '.vercel/source.tgz')],
+              data: tarBuffer,
+              mode: 0o666,
+            },
+          ],
+        ]);
+      } else {
+        files = await hashes(fileList);
+      }
+    } catch (err: any) {
```
EndangeredMassa: **suggestion (soft-blocking):** use `unknown` for error and type check it

EndangeredMassa (dismissed):

packages/client/src/create-deployment.ts
TooTallNate: fbbab36453c76843972f601c26e07d29aa5630f0

EndangeredMassa (approved):

andyvanosdale:
This still doesn't resolve the issue that folks are facing. Please see issue #11097 for more details. This is still completely broken, as it is not just npm modules not installed.

PR #11117, opened 2024-02-02
[gatsby-plugin-vercel-builder] use --keep-names esbuild flag

smaeda-ks:
Some packages are very sensitive to minification.
For example: https://github.com/node-fetch/node-fetch/issues/784 which ends up in runtime error, and is hard to notice as they don't reproduce on local dev.

Setting `--keep-names` should prevent those edge cases: https://esbuild.github.io/api/#keep-names

### DIFF
--- a/packages/gatsby-plugin-vercel-builder/src/handlers/build.ts
+++ b/packages/gatsby-plugin-vercel-builder/src/handlers/build.ts
@@ -35,7 +35,9 @@ export const writeHandler = async ({
       platform: 'node',
       bundle: true,
       minify: true,
-      keepNames: true,
+      // prevents renaming edge cases from causing failures like:
+      // https://github.com/node-fetch/node-fetch/issues/784 
+      keepNames: true, 
       define: {
         'process.env.NODE_ENV': "'production'",
         vercel_pathPrefix: JSON.stringify(prefix),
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-vercel-builder/src/handlers/build.ts
```diff
@@ -35,6 +35,7 @@ export const writeHandler = async ({
       platform: 'node',
       bundle: true,
       minify: true,
+      keepNames: true,
```
EndangeredMassa, line 40: **suggestion (soft-blocking):** add comment about why

In this or a follow-up PR: Can you add a comment about why we need this for future reference? Something like the following would be useful.
‏
```suggestion
      // prevents renaming edge cases from causing failures like:
      // https://github.com/node-fetch/node-fetch/issues/784 
      keepNames: true, 
```

EndangeredMassa (dismissed):

packages/gatsby-plugin-vercel-builder/src/handlers/build.ts
smaeda-ks: Thanks! @EndangeredMassa 
if you can re-approve 🙇

EndangeredMassa (approved):

PR #11121, opened 2024-02-03
[cli][build-utils] Rename variants to flags

AndyBitz:
- Removes all the legacy `flags`
 - Renames the new `variants` to `flags`

Neither the legacy flags, nor the new variants were exposed to anyone, except for the type in build-utils, so there shouldn't be any issues removing/renaming them.


packages/build-utils/src/types.ts
```diff
@@ -473,9 +464,7 @@ export interface BuildResultV2Typical {
   framework?: {
     version: string;
   };
-  /** @deprecated Replaced by Variants. Remove once fully replaced. */
-  flags?: Flag[];
-  variants?: Record<string, VariantDefinition>;
+  flags?: Record<string, FlagDefinition>;
```
dferber90: Let's sync on the format as the toolbar is expecting the format below

```
export interface FlagsDeploymentData {
  flags?: { definitions: FlagDefinitions };
}
```

packages/cli/src/commands/build/index.ts
```diff
@@ -820,60 +813,51 @@ function mergeWildcard(
...
-
-    return [];
-  });
-}
-
 /**
- * Takes the build output and writes all the variants into the `variants.json`
- * file. It'll skip variants that already exist.
+ * Takes the build output and writes all the flags into the `flags.json`
+ * file. It'll skip flags that already exist.
```
dferber90, line 818: What does the "It'll skip flags that already exist." part refer to?  Trying to understand this

dferber90 (dismissed):

packages/cli/src/commands/build/index.ts
AndyBitz: @dferber90 There is a chance that multiple builds, e.g. `@vercel/next` and `@vercel/node` being run on the same project, emit flags with the same name.

This is very unlikely, since it's more or less expected that only the main build, and eventually only the Build Output API has flags, but this is mostly to satisfy legacy behaviour and types.

packages/cli/src/commands/build/index.ts
```diff
@@ -820,60 +813,51 @@ function mergeWildcard(
...
  */
-async function writeVariantsJson(
+async function writeFlagsJSON(
   { output }: Client,
   buildResults: Iterable<BuildResult | BuildOutputConfig>,
   outputDir: string
 ): Promise<void> {
-  const variantsFilePath = join(outputDir, 'variants.json');
+  const flagsFilePath = join(outputDir, 'flags.json');
```
EndangeredMassa, line 825: **suggestion (blocking):** Can we add a test that this file is written under the correct circumstances?‏

EndangeredMassa (changes requested):

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1250,6 +1250,24 @@ describe('build', () => {
...
   });
+
+  it('should write to flags.json', async () => {
+    const cwd = fixture('with-flags'); // TODO
+    const output = join(cwd, '.vercel', 'output');
+
+    const exitCode = await build(client);
+    expect(exitCode).toEqual(0);
+
+    // TODO: Check the content of `flags.json`
```
EndangeredMassa: **suggestion (non-blocking):** ‏remove?

This is inspected now, right?

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1250,6 +1250,24 @@ describe('build', () => {
       (await fs.readFile(join(output, 'static/index.txt'), 'utf8')).trim()
     ).toEqual('marketing');
   });
+
+  it('should write to flags.json', async () => {
+    const cwd = fixture('with-flags'); // TODO
+    const output = join(cwd, '.vercel', 'output');
+
+    const exitCode = await build(client);
+    expect(exitCode).toEqual(0);
```
EndangeredMassa, line 1262: **suggestion (blocking):** ‏fix test

It looks like the sourcemaps are broken or something, but I think this is the line that's failing in the tests right now.

EndangeredMassa (changes requested):

packages/cli/test/unit/commands/build/index.test.ts
AndyBitz: Failed because Next 14 requires Node >16, and the tests are using Node 16

AndyBitz: At least that made it work locally <picture data-single-emoji=":sigh:" title=":sigh:">:sigh:</picture>

dferber90 (dismissed):

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1250,6 +1250,31 @@ describe('build', () => {
       (await fs.readFile(join(output, 'static/index.txt'), 'utf8')).trim()
     ).toEqual('marketing');
   });
+
+  it('should write to flags.json', async () => {
+    if (process.platform === 'win32') {
+      console.log('Skipping test on Windows');
```
EndangeredMassa: **question (soft-blocking):** Why does this fail on Windows?

If we need to skip this test on Windows, can we add a comment about why?‏

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/build/index.test.ts
AndyBitz: Sure, added a comment! Can't run it on windows because the fixture runs a bash script.

EndangeredMassa: Can we change the fixture to not use a bash script? I'd like to keep windows coverage if we can.

AndyBitz: Fair, converted it to a node script.

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1250,43 +1250,63 @@ describe('build', () => {
...
+    expect(fs.readJSONSync(join(output, 'flags.json'))).toEqual({
+      definitions: {
+        'my-next-flag': {
+          options: [{ value: true }, { value: false }],
+        },
+      },
+    });
+  });
+
+  it('should create symlinks for duplicate references to Lambda / EdgeFunction instances', async () => {
```
AndyBitz, line 1274: I moved this test into the `describe` block, since I assume this was the reason it failed sometimes.
Probably because the `beforeEach` in the describe block resets the env variable, which caused this test to appear as if it was running within `vercel build`, causing a failure.

packages/next/src/utils.ts
```diff
@@ -3307,19 +3308,14 @@ export function isApiPage(page: string | undefined) {
...
 
-/** @deprecated */
-export type VariantsManifestLegacy = Record<
-  string,
-  {
-    defaultValue?: unknown;
-    metadata?: Record<string, unknown>;
-  }
->;
+export type VariantsManifest = {
```
dferber90, line 3311: Should this be called `FlagsManifest`

AndyBitz: @dferber90 that will depend on what Next.js will ultimately call it. Kept it like this to have a differnce between Next and the rest for now

EndangeredMassa (approved):

PR #11126, opened 2024-02-06
[next] Preload common chunks

javivelasco:
This PR adds requiring the modules that _we know_ will be required in the Next.js Lambda in the module scope in a attempt to make a better use of resources.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -70,6 +70,30 @@ const NEXT_DATA_MIDDLEWARE_RESOLVING_VERSION = 'v12.1.7-canary.33';
 const EMPTY_ALLOW_QUERY_FOR_PRERENDERED_VERSION = 'v12.2.0';
 const CORRECTED_MANIFESTS_VERSION = 'v12.2.0';
 
+// Ideally this should be in a Next.js manifest so we can change it in
+// the future but this also allows us to improve existing versions
+const PRELOAD_CHUNKS = {
+  APP_ROUTER_PAGES: [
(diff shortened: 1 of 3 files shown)


dvoytenko (dismissed):

packages/next/src/server-build.ts
```diff
@@ -1068,9 +1092,27 @@ export async function serverBuild({
         }
       }
 
+      let launcherData = group.isAppRouter ? appLauncher : launcher;
+      let preloadChunks: string[] = [];
```
ijjk, line 1096: The preload chunks are version specific, we should either version lock or validate the files exist on disk before adding to the launcher

packages/next/src/server-build.ts
```diff
@@ -70,6 +70,30 @@ const NEXT_DATA_MIDDLEWARE_RESOLVING_VERSION = 'v12.1.7-canary.33';
 const EMPTY_ALLOW_QUERY_FOR_PRERENDERED_VERSION = 'v12.2.0';
 const CORRECTED_MANIFESTS_VERSION = 'v12.2.0';
 
+// Ideally this should be in a Next.js manifest so we can change it in
+// the future but this also allows us to improve existing versions
```
styfle, line 74: Agreed, this should be in Next.js manifest that we can emit during `next build`. Otherwise it will be painful mapping each version of Next.js each time it needs to change.

ijjk (approved):
Looks good to me for testing

Ethan-Arrowood (commented):
Looks good to me. Thank you!

PR #11131, opened 2024-02-08
[build-utils] change default package manager when no lockfile detected from `yarn` to `npm` (gated behind feature flag)

styfle:
This PR changes the default package manager from `yarn` to `npm` in the case that no lockfile is present.

Many years ago when `yarn` was first released, it was much faster than `npm` so we used it by default. That is no longer the case today and `yarn@1` is no longer receiving new features.

For example, `sharp` and `esbuild` no longer works with `yarn@1`:

### Related

- https://github.com/lovell/sharp/issues/3750
- https://github.com/evanw/esbuild/issues/2949

Note that this change will not impact most projects because most used a lockfile.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -316,8 +316,9 @@ export async function scanParentDirs(
       bunLockPath ? fs.readFile(bunLockPath, 'utf8') : null,
     ]);
 
-  // Priority order is bun with yarn lock > yarn > pnpm > npm > bun
-  if (bunLockBin && hasYarnLock) {
+  // Priority order is bun > yarn > pnpm > npm
+  // If no lockfile is detected, then npm is used.
+  if (bunLockBin) {
```
EndangeredMassa: **question (non-blocking):** ‏should we change the special case precedence?

I think changing the precedence here of the special case (both a bun and yarn lockfile) is fine, but it's not strictly necessary for this PR. It only affects projects with multiple package manager lockfiles.

Why do you think we should change this precedence?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -333,11 +334,8 @@ export async function scanParentDirs(
     cliType = 'npm';
     lockfilePath = npmLockPath;
     lockfileVersion = packageLockJson.lockfileVersion;
-  } else if (bunLockBin) {
-    cliType = 'bun';
-    lockfilePath = bunLockPath;
-    // TODO: read "bun-lockfile-format-v0"
-    lockfileVersion = 0;
+  } else {
+    cliType = 'npm';
```
EndangeredMassa: **question (blocking):** ‏Can we wrap this in a flag we can flip?

packages/build-utils/test/unit.test.ts
```diff
@@ -576,6 +576,28 @@ it(
...
+it('should return cliType npm when no lockfile is present', async () => {
+  const originalRepoLockfilePath = path.join(
+    __dirname,
+    '..',
+    '..',
+    '..',
+    'pnpm-lock.yaml'
+  );
+  const originalRepoLockfileData = await fs.readFile(originalRepoLockfilePath);
+  await fs.remove(originalRepoLockfilePath);
```
EndangeredMassa, line 588: **question (blocking):** ‏why are we referring to a lockfile that doesn't exist?

Am I missing something about how this test works?

EndangeredMassa (changes requested):

packages/build-utils/src/fs/run-user-scripts.ts
styfle: Hmm, I think I misread this code (and I think the comment was confusing too). There are no early returns so the order is backwards (meaning last in the elseif takes precedence, not the first).

I'll just revert that part to avoid changing anything with the order but I do think there is a bug here we should revisit.

packages/build-utils/test/unit.test.ts
styfle: I think you misunderstand.

The function we're testing here is `scanParentDirs()` and it will walk up the dir tree until it finds a lockfile. Since the test fixture doesn't have a lockfile, it keeps walking up the tree until it finds the monorepo root lockfile. So it incorrectly thinks the test fixture has a lockfile.

We could update `scanParentDirs()` to accept a new parameter that would stop walking the tree at a certain point but instead I just delete and restore the monorepo's lockfile so it won't be detected.

packages/build-utils/src/fs/run-user-scripts.ts
styfle: Yep good point! I added a flag `VERCEL_ENABLE_NPM_DEFAULT` in the latest commit

EndangeredMassa (dismissed):

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -414,6 +420,14 @@ export async function runNpmInstall(
       destPath
     );
 
+    if (!packageJsonPath) {
+      debug(
+        `Skipping dependency installation because no package.json was found for ${destPath}`
+      );
+      runNpmInstallSema.release();
+      return false;
+    }
```
styfle, line 429: Note: this change was required because yarn somehow allowed us to run `yarn install`, even if there was no `package.json` but `npm install` will throw.

This might end up being a tiny perf improvement for the case when no package.json is found since we then won't need to invoke a package manager at all.

PR #11148, opened 2024-02-13
build: upgrade edge-runtime

Kikobeats:
Upgrade Edge Runtime and related dependencies 🙂


packages/node/src/dev-server.mts
```diff
@@ -46,12 +52,31 @@ async function createEventHandler(
...
   }
 
+  const content = await readFile(entrypointPath, 'utf8');
+
+  const isStreaming =
+    staticConfig?.supportsResponseStreaming ||
+    (await hasWebHandlers(async () => parseCjs(content).exports)) ||
+    (await hasWebHandlers(async () =>
+      init.then(() => parseEsm(content)[1].map(specifier => specifier.n))
+    ));
```
EndangeredMassa, line 62: **question (soft-blocking):** ‏Can we expand this very complex statement?

```suggestion
  function hasCjsWebHandlers(content) {
    return hasWebHandlers(async () => 
      parseCjs(content).exports
    );
  }
  
  function hasEsmWebHandlers(content, init) {
    return hasWebHandlers(async () =>
      init.then(() => {
        // TODO: what is this part of the result?
        // What name can we give it?
        const namedThing = parseEsm(content)[1];
        if (!namedThing) {
          // TODO: What's an appropriate error here?
          throw new Error('Some Error');
        }
        
        // TODO: why do we need to do this?
        // Can we add a comment about what this does?
        return namedThing.map(specifier => specifier.n)
      })
    );
  }

  const isStreaming = staticConfig?.supportsResponseStreaming
    || await hasCjsWebHandlers(content)
    || await hasEsmWebHandlers(content, init);
```

I think the next person to read through this would appreciate some more explicit logic, names, and comment or two.

packages/node/src/serverless-functions/helpers-web.ts
```diff
@@ -1,26 +1,7 @@
...
-
-  respondWith(response: Response) {
-    this.response = response;
-  }
-
-  waitUntil() {
-    throw new Error('waitUntil is not implemented yet for Node.js');
-  }
-}
+import Edge from '@edge-runtime/primitives';
```
EndangeredMassa, line 4: **suggestion (blocking):** ‏add tests

Can we have a couple tests that ensure the use of this updated `FetchEvent` is working correctly? I think at least having the following would be awesome.

- test a serverless function with web handler, like [the edge function test](https://github.com/vercel/vercel/blob/142a397d8eb9575d6ae17ec36b0d1a021a2dfc5a/packages/node/test/unit/dev.test.ts#L311)
- test the updated isStreaming logic properly detects those cases with web handlers

EndangeredMassa (changes requested):

packages/node/test/unit/dev.test.ts
```diff
@@ -43,21 +43,250 @@ function testForkDevServer(entrypoint: string) {
...
+      //     body: await response.text(),
+      //     transferEncoding: response.headers.get('transfer-encoding'),
+      //     'x-web-handler': response.headers.get('x-web-handler')
+      //   }).toEqual({
+      //     status: 200,
+      //     body: 'Web handler using DELETE',
+      //     transferEncoding: 'chunked',
+      //     'x-web-handler': "Web handler using DELETE",
+      //   });
+      // }
```
EndangeredMassa: **suggestion (soft-blocking):** uncomment?

Are these `OPTIONS` calls not working?

EndangeredMassa (dismissed):

packages/node/test/dev-fixtures/web-handlers-edge.js
```diff
@@ -0,0 +1,56 @@
...
+          await wait(DEFER_MS);
+        }
+        controller.close();
+      },
+    }).pipeThrough(new TextEncoderStream());
+
+    const defer = async () => {
+      // console.log('waitUntil', DATA)
+      await wait(DATA.length * DEFER_MS);
+      // console.log('waitUntil:done', DATA)
```
EndangeredMassa: Clear these out later?

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #11151, opened 2024-02-13
Restore ruby fixtures post deploy

trek:
These fixtures were disabled to allow us to do a deploy of some newer ruby stuff. Restoring them now that the changes are out.


packages/static-build/test/integration-setup.js
```diff
@@ -38,7 +38,7 @@ module.exports = function setupTests(groupIndex) {
     console.log('testing group', groupIndex, fixtures);
   }
 
-  const fixturesToSkip = ['jekyll-v3', 'jekyll-v4', 'middleman-v4'];
+  const fixturesToSkip = [];
```
trek, line 41: Could remove entirely, but I'm inclined to leave it for possible later use? 🤷

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #11152, opened 2024-02-13
[remix] Don't install Remix fork when not using split configuration

TooTallNate:
The fork of the Remix compiler is only necessary when the project is utilizing split configurations in their routes (i.e. mixing Node.js and Edge functions).

However, the injecting of the forked compiler has proven to be brittle and has many edge cases which cause the fork to not be present. Considering that the more common case is to _not_ use split configuration, skip the fork entirely when that is the case.

This is somewhat of a band-aid / hold over for the "old" Remix. With the Remix + Vite configuration, the fork will be entirely unnecessary.

### DIFF
--- a/packages/remix/src/build.ts
+++ b/packages/remix/src/build.ts
@@ -239,7 +239,7 @@ export const build: BuildV2 = async ({
 
   // If the project is not relying on split configuration,
   // then don't inject the `@vercel/remix-run-dev` fork
-  if (serverBundles.length === 1) {
+  if (!isHydrogen2 && serverBundles.length === 1) {
     // `serverBuildTarget` and `serverBuildPath` are undefined with
     // our remix config modifications, so use the default build path
     serverBundles[0].serverBuildPath = 'build/index.js';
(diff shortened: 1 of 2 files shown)


trek (dismissed):

packages/remix/src/build.ts
```diff
@@ -174,9 +174,81 @@ export const build: BuildV2 = async ({
...
+        : `${relative(
+            entrypointFsDirname,
+            dirname(remixConfig.serverBuildPath)
+          )}/build-${runtime}-${hash}.js`,
+      routes: routes.map(r => r.id),
+    };
+  });
+
+  // If the project is not relying on split configuration,
+  // then don't inject the `@vercel/remix-run-dev` fork
```
EndangeredMassa: **suggestion (non-blocking):** ‏This looks like a comment for line 251, not here

EndangeredMassa (dismissed):
**question (soft-blocking):** ‏can we add a test for this case?

EndangeredMassa (approved):

TooTallNate:
@EndangeredMassa A relevant test is being added in https://github.com/vercel/vercel/pull/11153.

PR #11182, opened 2024-02-21
Add immutability modifiers to types

wyattjoh:
This doesn't change any functionality, and only replaces some mutable types (like `Map`, `Set`, and `Array`) with their immutable versions (like `ReadonlyMap`, `ReadonlySet`, and `ReadonlyArray`). This can help keep functions as pure as possible and improves the readability of some of the involved functions.

PR #10753, opened 2023-10-23
Added `getRequestHandlerWithMetadata` export

wyattjoh:
This adds a new `getRequestHandlerWithMetadata` export if enabled and available to the exported method.

PR #11183, opened 2024-02-21
[next] Support pre-generated pages without fallbacks with Partial Prerendering

wyattjoh:
When using App Router, you can declare a route that has dynamic path segments using the `[param]` syntax in the folder structure. These can be either handled dynamically (the default), statically, or both. When these static parameters are provided statically, it could look something like this:

```typescript
// app/blog/[slug]/page.tsx
import { getBlogPosts, getBlogPost } from "@/db"

// 🔬 This is the important bit
export const dynamicParams = false;

type Props = {
  params: {
    slug: string
  }
}

export async function generateStaticParams() {
  const posts = await getBlogPosts()
  return posts.map(({ slug }) => ({ slug }))
}

export default async function BlogPage({ params: { slug } }: Props) {
  const post = await getBlogPost(slug)
  return <div>My Blog Post: {post.title}</div>
}
```

By setting `dynamicParams = false`, you're telling Next.js that you shouldn't support new blog posts that aren't defined in the `generateStaticParams`. When Partial Prerendering has been enabled, we have a bit of a problem. The previous code assumed that if you had pre-generated the routes, and that there was no fallback, that no additional rewrites should be added to support dynamic routes (as the routes themselves have been pre-generated).

This pull request fixes this by creating outputs for each of the statically generated routes when they have dynamic parameters turned off. This creates N*2 more outputs/routes for each statically generated route, as it requires an output for the Partial Prerendering Resume route as well as the Dynamic RSC route. This ensures that no other route match can be hit if `dynamicParams` is set to `false`.

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -2145,29 +2145,31 @@ export const build: BuildV2 = async ({
       omittedPrerenderRoutes,
     });
 
-    const prerenderParameters: Array<Parameters<typeof prerenderRoute>> = [];
-
-    for (const route of Object.keys(prerenderManifest.staticRoutes)) {
-      prerenderParameters.push([route, {}]);
-    }
-
-    for (const route of Object.keys(prerenderManifest.fallbackRoutes)) {
-      prerenderParameters.push([route, { isFallback: true }]);
-    }
-
-    for (const route of Object.keys(prerenderManifest.blockingFallbackRoutes)) {
-      prerenderParameters.push([route, { isBlocking: true }]);
-    }
-
-    if (static404Page && canUsePreviewMode) {
-      for (const route of Array.from(omittedPrerenderRoutes)) {
-        prerenderParameters.push([route, { isOmitted: true }]);
-      }
-    }
+    await Promise.all(
+      Object.keys(prerenderManifest.staticRoutes).map(route =>
+        prerenderRoute(route, {})
+      )
+    );
 
     await Promise.all(
-      prerenderParameters.map(parameters => prerenderRoute(...parameters))
+      Object.keys(prerenderManifest.fallbackRoutes).map(route =>
+        prerenderRoute(route, { isFallback: true })
+      )
     );
+
+    await Promise.all(
+      Object.keys(prerenderManifest.blockingFallbackRoutes).map(route =>
+        prerenderRoute(route, { isBlocking: true })
+      )
+    );
+
+    if (static404Page && canUsePreviewMode) {
+      await Promise.all(
+        Array.from(omittedPrerenderRoutes).map(route =>
+          prerenderRoute(route, { isOmitted: true })
+        )
+      );
+    }
 
     // We still need to use lazyRoutes if the dataRoutes field
     // isn't available for backwards compatibility
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -1143,9 +1143,8 @@ export const build: BuildV2 = async ({
     });
 
     /**
-     * Originally, you had to have a `pages/api` route in order to add a preview
-     * mode cookie. This isn't the case with app router, as it can be done
-     * dynamically via a Server Action or a App Route.
+     * This is a detection for preview mode that's required for the pages
+     * router.
      */
     const canUsePreviewMode = Object.keys(pages).some(page =>
       isApiPage(pages[page].fsPath)
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -2145,21 +2145,29 @@ export const build: BuildV2 = async ({
       omittedPrerenderRoutes,
     });
 
-    Object.keys(prerenderManifest.staticRoutes).forEach(route =>
-      prerenderRoute(route, { isBlocking: false, isFallback: false })
-    );
-    Object.keys(prerenderManifest.fallbackRoutes).forEach(route =>
-      prerenderRoute(route, { isBlocking: false, isFallback: true })
-    );
-    Object.keys(prerenderManifest.blockingFallbackRoutes).forEach(route =>
-      prerenderRoute(route, { isBlocking: true, isFallback: false })
-    );
+    const prerenderParameters: Array<Parameters<typeof prerenderRoute>> = [];
+
+    for (const route of Object.keys(prerenderManifest.staticRoutes)) {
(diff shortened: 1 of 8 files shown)


wyattjoh:
* **#11183** <a href="https://app.graphite.dev/github/pr/vercel/vercel/11183?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#11182** <a href="https://app.graphite.dev/github/pr/vercel/vercel/11182?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

ztanner (dismissed):

packages/next/test/fixtures/00-app-dir-ppr-full/index.test.js
```diff
@@ -19,6 +19,9 @@ const pages = [
   { pathname: '/no-suspense/nested/a', dynamic: true },
   { pathname: '/no-suspense/nested/b', dynamic: true },
   { pathname: '/no-suspense/nested/c', dynamic: true },
+  { pathname: '/no-fallback/a', dynamic: true },
+  { pathname: '/no-fallback/b', dynamic: true },
+  { pathname: '/no-fallback/c', dynamic: true },
```
ijjk, line 24: Would be good to have a test case to ensure `/no-fallback/non-existent` `404s` properly and doesn't invoke the function.

packages/next/src/utils.ts
```diff
@@ -338,7 +350,14 @@ export async function getDynamicRoutes(
...
 
         for (const dynamicRoute of routesManifest.dynamicRoutes) {
-          if (!canUsePreviewMode && omittedRoutes?.has(dynamicRoute.page)) {
+          if (
+            !canUsePreviewMode &&
+            omittedRoutes?.has(dynamicRoute.page) &&
+            // If this route supports Partial Prerendering, we need to handle
+            // the case where it adds the corresponding .rsc and .prefetch.rsc
+            // rewrite rules.
+            !experimentalPPRRoutes.has(dynamicRoute.page)
```
ijjk: For this case we need to only add the `.rsc` and `.prefetch.rsc` routes otherwise the `dynamicParams: false` handling doesn't work and the underlying function can still be invoked right?

packages/next/src/index.ts
```diff
@@ -1142,6 +1142,11 @@ export const build: BuildV2 = async ({
       appPathRoutesManifest,
     });
 
+    /**
+     * Originally, you had to have a `pages/api` route in order to add a preview
```
ztanner: nit, but I think maybe this comment could be summed up as preview mode is a pages router concept, draft mode is an app router concept.

packages/next/src/server-build.ts
```diff
@@ -1197,65 +1188,104 @@ export async function serverBuild({
...
           lambdas[outputName] = revalidate;
 
-          const pprOutputName = path.posix.join(
-            entryDirectory,
-            '/_next/postponed/resume',
-            pageNoExt
-          );
-          lambdas[pprOutputName] = lambda;
+          // If this is an omitted page, them it should be routed to the
+          // specific dynamic route, and not the dynamic one.
```
ztanner: Not sure that I understand what this means ("routed to the dynamic route, and not the dynamic one")

packages/next/src/server-build.ts
```diff
@@ -1288,32 +1318,39 @@ export async function serverBuild({
     hasPages404: routesManifest.pages404,
     isCorrectNotFoundRoutes,
     isEmptyAllowQueryForPrendered,
+    omittedPrerenderRoutes,
   });
 
-  await Promise.all(
-    Object.keys(prerenderManifest.staticRoutes).map(route =>
+  const pending: Array<Promise<void>> = [];
```
ztanner: `pendingPrerenders`, perhaps?

packages/next/src/server-build.ts
```diff
@@ -1560,6 +1597,46 @@ export async function serverBuild({
...
 
+  // If we're using the Experimental Partial Prerendering, we should ensure that
+  // all the routes that support it (and are listed) have configured lambdas.
+  // This only applies to routes that do not have fallbacks enabled (these are
+  // routes that have `dynamicParams =  false` defined.
+  if (experimental.ppr) {
+    for (const { srcRoute, dataRoute, experimentalPPR } of Object.values(
+      prerenderManifest.staticRoutes
+    )) {
+      // Only apply this to the routes that are support experimental PPR and
```
ztanner:
```suggestion
      // Only apply this to the routes that support experimental PPR and
```

packages/next/src/server-build.ts
```diff
@@ -1560,6 +1597,46 @@ export async function serverBuild({
...
+      // The lambda paths have their leading `/` stripped.
+      const srcPathname = srcRoute.substring(1);
+      const dataPathname = dataRoute.substring(1);
+
+      // If we already have an associated lambda for the `.rsc` route, then
+      // we can skip this.
+      const dataPathnameExists = dataPathname in lambdas;
+      if (dataPathnameExists) continue;
+
+      // We requires that the source route has a lambda associated with it. If
```
ztanner:
```suggestion
      // We require that the source route has a lambda associated with it. If
```

ztanner (dismissed):

packages/next/src/server-build.ts
```diff
@@ -1288,32 +1315,39 @@ export async function serverBuild({
     hasPages404: routesManifest.pages404,
     isCorrectNotFoundRoutes,
     isEmptyAllowQueryForPrendered,
+    omittedPrerenderRoutes,
   });
 
-  await Promise.all(
-    Object.keys(prerenderManifest.staticRoutes).map(route =>
+  const prerenderPromises: Array<Promise<void>> = [];
```
ijjk: We can't use these prerender promises as it causes a race condition with any of them rejecting before we attach the `Promise.all` which crashes the build. 

Maybe we add a lint rule for this as it's popped up before https://github.com/vercel/vercel/pull/10753#discussion_r1385248267 🤔

ijjk (commented):
Besides above note, other changes look good to me

Ethan-Arrowood (approved):

ijjk (approved):

TooTallNate (approved):

PR #11203, opened 2024-02-28
[examples] Bump Middleman and Jekyll dependencies to support ruby 3.x

trek:
Some dependency updates are required to support Ruby 3.x+


EndangeredMassa (approved):

examples/jekyll/Gemfile.lock
```diff
@@ -52,21 +52,22 @@ GEM
...
       ffi (~> 1.9)
     terminal-table (2.0.0)
       unicode-display_width (~> 1.1, >= 1.1.1)
-    unicode-display_width (1.7.0)
+    unicode-display_width (1.8.0)
+    webrick (1.8.1)
 
 PLATFORMS
-  x86_64-linux
+  ruby
```
erikareads, line 70: What does the change in platform do here? That is, what's the difference between `x86_64-linux` and `ruby`?

examples/middleman/Gemfile.lock
```diff
@@ -1,108 +1,113 @@
...
+    toml (0.3.0)
+      parslet (>= 1.8.0, < 3.0.0)
+    tzinfo (2.0.6)
+      concurrent-ruby (~> 1.0)
     uglifier (3.2.0)
       execjs (>= 0.3.0, < 3)
+    webrick (1.8.1)
 
 PLATFORMS
   x86_64-linux
```
erikareads, line 106: Why is this one the same when the other one was changed?

examples/jekyll/Gemfile.lock
EndangeredMassa: I'd also like to better understand. cc @trek 

The docs don't say that `x86_64-linux` is a valid value, but maybe it used to be.
https://bundler.io/v2.3/man/gemfile.5.html#PLATFORMS

trek: `[PLATFORMS](https://guides.rubygems.org/specification-reference/#platformplatform)` is used for gems with native binary bindings that get pre-compiled, so gems specify which versions need to be installed (sometimes multiples and only one gets installed depending on what platform the interpreter is running on at the moment).

These get written to the lockfile when running `bundle`. [Here's a nice little explainer](https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/#platforms)

So, it's common to have values like

```
PLATFORMS
  x86_64-linux
  arm64-darwin
  ruby
```

If you have a mix of pure ruby gems and gems with some pre-compiled binaries and you deploy to a linux system but your developers work on their Macbooks.

In this case, I'm guessing the final gem with a pre-compiled dependency was either removed or a newer version has a pure-Ruby implementation.

examples/middleman/Gemfile.lock
trek: See [this comment](https://github.com/vercel/vercel/pull/11203#discussion_r1507777893). In this case I'd guess there is still a dependency that has a `x86_64-linux` pre-compiled binary, but that binary comes pre-installed on MacOS (where I ran `bundle`).

PR #11207, opened 2024-02-29
[remix] Sort dynamic routes by precedence

TooTallNate:
Ensure routes are sorted so that more specific routes come before more permissive routes. For example, given two pages:

* `products.($lang).$productId.tsx`
* `products.($lang).categories.tsx`

In this example, the `categories` route should come first in the routes array, because it is more specific than the dynamic `$productId` parameter. If the `$productId` route is first, then it would erroneously match against "categories" and end up serving the wrong content upon server render, which would cause a hydration mismatch warning on the client-side.

### DIFF
--- a/.changeset/quick-snakes-whisper.md
+++ b/.changeset/quick-snakes-whisper.md
@@ -1,0 +1,5 @@
+---
+'@vercel/remix-builder': patch
+---
+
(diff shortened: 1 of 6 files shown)


EndangeredMassa:
Closing for now, but we'll revisit if people start hitting this case.

PR #11212, opened 2024-03-01
[cli] Fix framework version detection in monorepos

TooTallNate:
I noticed that the framework version was not being printed on deployments in a monorepo. Turns out the framework detection logic was happening at the root of the monorepo, instead of the project directory.


packages/cli/src/commands/build/index.ts
```diff
@@ -669,7 +669,7 @@ async function doBuild(
     overrides.length > 0 ? Object.assign({}, ...overrides) : undefined;
   const mergedFlags = mergeFlags(buildResults.values());
 
-  const framework = await getFramework(cwd, buildResults);
```
EndangeredMassa, line 667: **suggestion (soft-blocking):** add test

Is there a test we can add for this?‏

EndangeredMassa (dismissed):

packages/cli/src/commands/build/index.ts
TooTallNate: Test added in 5681aa72ce5cf87958e4741a0c30fdc971e8fa9e.

packages/cli/src/commands/build/index.ts
```diff
@@ -669,7 +669,7 @@ async function doBuild(
...
+  const framework = await getFramework(workPath, buildResults);
```
TooTallNate, line 667: __Note:__ This is the main change. Everything else is scaffolding for the test.

utils/chunk-tests.js
```diff
@@ -9,6 +9,7 @@ const runnersMap = new Map([
       min: 1,
       max: 1,
       runners: ['ubuntu-latest', 'macos-latest', 'windows-latest'],
+      nodeVersion: '18',
```
TooTallNate: Note that Remix requires Node 18 to build, so updating the Unit tests to run using Node 18. I think considering Node 16 is EOL, this is acceptable.

EndangeredMassa (approved):

trek (dismissed):

utils/chunk-tests.js
EndangeredMassa: **suggestion (blocking):** Can we pull this out similar to `test-next-local`?‏

I'd like to ensure the rest of `test-unit` still works under Node 16 while we support it.

EndangeredMassa (changes requested):

utils/chunk-tests.js
TooTallNate: I switched the test to use an older version of Next.js (v12) instead. This works on Node 16, so the unit tests' node version is no longer being modified.

EndangeredMassa (approved):

Zhaph:
This seems to have broken our CLI based deployments from 33.5.5

We've got an Azure DevOps pipeline that was working successfully under 33.5.5, that is now failing with 33.6.0:

```
- task: bash@3
  displayName: 'Deploy Front End to Vercel Production'
  inputs:
    targetType: 'inline'
    script: |
      vercel deploy --prod --token ${{ parameters.VercelToken }}
  env:
    VERCEL_ORG_ID: ${{ parameters.VercelTeamId }}
    VERCEL_PROJECT_ID: ${{ parameters.VercelProjectId }}
```

Our deployment is now failing with the error:

> Invalid request: `projectSettings.nodeVersion` should be equal to one of the allowed values "20.x, 18.x, 16.x".

TooTallNate:
@Zhaph The issue you are seeing was actually caused by https://github.com/vercel/vercel/pull/11261, not this PR. It has also been fixed by https://github.com/vercel/vercel/pull/11277, which was released in [Vercel CLI v33.6.1](https://github.com/vercel/vercel/releases/tag/vercel%4033.6.1). Please give that latest version a try and let us know if that sorts the issue for you.

Zhaph:
Yep, all good @TooTallNate, thanks for confirming and sorry for the confusion.

PR #11213, opened 2024-03-01
[remix] Remove usage of `ensureResolvable()` in Vite builds

TooTallNate:
This function can lead to unexpected results (for example, in a monorepo setup where there were different versions of Remix being used, I encountered the wrong version of Remix being symlinked causing issues at runtime).

Instead, throw an error at build time at the NFT step if `@remix-run/node` is missing for routes using Node.js, or when `@remix-run/server-runtime` is missing for routes using Edge.

Less magical is better. Explicit is better than implicit.


packages/remix/src/build-vite.ts
```diff
@@ -118,18 +119,7 @@ export const build: BuildV2 = async ({
...
-    repoRootPath,
-    '@remix-run/dev'
-  );
-  const remixRunDevPkg = JSON.parse(
-    readFileSync(join(remixRunDevPath, 'package.json'), 'utf8')
-  );
-  const remixVersion = remixRunDevPkg.version;
-
-  // Make `remix build` output production mode
-  spawnOpts.env.NODE_ENV = 'production';
```
EndangeredMassa, line 132: **question (blocking):** ‏Do we not need to set `NODE_ENV`?

EndangeredMassa (changes requested):

packages/remix/src/build-vite.ts
TooTallNate: I'm confirming with Mark, but I believe it's no longer necessary.

EndangeredMassa (approved):

packages/remix/src/build-vite.ts
TooTallNate: Confirmed. Remix sets this explicitly [here](https://github.com/remix-run/remix/blob/main/packages/remix-dev/vite/build.ts#L35).

PR #11217, opened 2024-03-02
[examples] Update Remix template to use Vite

TooTallNate:
Updates the Remix example template to use Vite and latest version of dependencies.


leerob (approved):

leerob:
Let's make sure we update https://remix-run-template.vercel.app/ as well 🥳

PR #11225, opened 2024-03-05
Update .gitignore

Abuchtela:
Add .Env to fix exposed cred


EndangeredMassa:
@Abuchtela thanks for the PR! We don't have a use case that will produce a `.env` file in the root of this repo. I don't think this specific change is necessary.

PR #11226, opened 2024-03-05
Fix issue with serverless function on docker

Devin345458:
Currently the server was binding to the IPV6 [::] catchall on a random port. The undici request is then unable to reach it. 

I only have a mac so I was only able to test this change on Docker for mac. Would love if someone on windows/linux could validate the change works for them as well


TooTallNate (dismissed):

EndangeredMassa (dismissed):

EndangeredMassa:
Thanks!

PR #11229, opened 2024-03-05
[examples] Update Remix template to use Vite

TooTallNate:
Updates the Remix example template to use Vite and latest version of dependencies.

This PR is a reapplication of #11217, which was reverted in #11227 (the underlying framework detection issue has been fixed).


TooTallNate:
Merging without reviews since this was already previously approved.

PR #11236, opened 2024-03-06
fix vc link framework detection

EndangeredMassa:


EndangeredMassa:
This doesn't actually work.

PR #11242, opened 2024-03-07
[build-utils] Remove `VERCEL_ENABLE_NPM_DEFAULT` env var check

TooTallNate:
Flag has been rolled out for long enough at this point.

npm will now always be used as the package manager when no lockfile is present.


EndangeredMassa (approved):

styfle (approved):
Thanks!

trek (approved):

PR #11243, opened 2024-03-08
[cli] fix root dir issues with `vc link` and `vc deploy`

EndangeredMassa:
There's an issue with `vc link` and `vc deploy` when using a project root directory. Framework detection fails because it's looking in the wrong directory.

---

Manually tested and this fixes the issue. A project with a rootDirectory was previously not detecting a framework, but now it says:

```
Auto-detected Project Settings (Remix):
```


packages/cli/src/commands/deploy/index.ts
```diff
@@ -545,12 +545,20 @@ export default async (client: Client): Promise<number> => {
...
-          ? {
-              sourceFilesOutsideRootDirectory,
-            }
-          : { ...localConfigurationOverrides, sourceFilesOutsideRootDirectory };
+      createArgs.projectSettings = {
+        sourceFilesOutsideRootDirectory,
+        rootDirectory:
+          sourceFilesOutsideRootDirectory && rootDirectory
+            ? rootDirectory
+            : undefined,
```
TooTallNate: I think I had this logic wrong in the thread. We always want to be setting the root directory:

```suggestion
        rootDirectory,
```

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -198,6 +198,10 @@ export default async function setupAndLink(
         projectSettings: {
           ...localConfigurationOverrides,
           sourceFilesOutsideRootDirectory,
+          rootDirectory:
+            sourceFilesOutsideRootDirectory && rootDirectory
+              ? rootDirectory
+              : undefined,
```
TooTallNate: Ditto:

```suggestion
          rootDirectory,
```

TooTallNate (approved):

trek (approved):

PR #11247, opened 2024-03-08
[static-build] Remove Build Output API v1 tests

TooTallNate:
Remove these old tests for the legacy BOA version. Specifically, `66-build-output-directory-with-functions` has become problematic because the "go1.x" Lambda runtime no longer deploys. I figured we take the opportunity to remove the others as well though. Will make running the tests slightly faster.


trek (approved):

styfle (commented):
Do we know if BOA v1 is still used? If so, I think we should keep the test and update it to continue working. Otherwise it will be hard to know if we introduce a regression in the future.

styfle (approved):
Oh I see we have quite a few other tests for BOA v1 such as [65-build-output-directory-with-legacy-node-functions](https://github.com/vercel/vercel/tree/main/packages/static-build/test/fixtures/65-build-output-directory-with-legacy-node-functions) so I think we can remove this one 👍

PR #11254, opened 2024-03-11
[next] add x-forwarded-port test

Ethan-Arrowood:
This test reproduces the issue described in https://github.com/vercel/next.js/issues/61133 . Unfortunately, I can only reproduce this issue when deploying to Vercel. Any local server / test results in the header being defined. 

Related: https://github.com/vercel/next.js/issues/61133


Ethan-Arrowood:
Here is an example run demonstrating the _current behavior_. Essentially, `x-forwarded-port` header is `undefined/null` on a deployed route.

```
onsole.log
    3/11/24, 2:02:55 PM 00-x-forwarded-port should deploy and pass probe checks testing {"path":"/api/app-headers","status":200,"mustContain":"{\"port\":null}"}

      at logWithinTest (../../test/lib/deployment/log.js:32:11)

  console.log
    3/11/24, 2:02:56 PM 00-x-forwarded-port should deploy and pass probe checks finished testing        {"path":"/api/app-headers","status":200,"mustContain":"{\"port\":null}"}

      at logWithinTest (../../test/lib/deployment/log.js:32:11)

  console.log
    3/11/24, 2:02:56 PM 00-x-forwarded-port should deploy and pass probe checks testing {"path":"/api/pages-headers","status":200,"mustContain":"{}"}

      at logWithinTest (../../test/lib/deployment/log.js:32:11)

  console.log
    3/11/24, 2:02:57 PM 00-x-forwarded-port should deploy and pass probe checks finished testing        {"path":"/api/pages-headers","status":200,"mustContain":"{}"}

      at logWithinTest (../../test/lib/deployment/log.js:32:11)

 PASS  test/fixtures/00-x-forwarded-port/index.test.js (52.361 s)
  00-x-forwarded-port
    ✓ should deploy and pass probe checks (51936 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        52.438 s
Ran all test suites matching /test\/fixtures\/00-x-forwarded-port\/index.test.js/i.
```

Ethan-Arrowood:
Now, with the latest commit, this test will fail unless we land https://github.com/vercel/next.js/pull/60484

ijjk (commented):
Instead of a completely separate test suite can we include this in the existing 00-app-dir suite? 

We'll also need to work on coupling these into the Next.js repo so that feedback loop is complete

Ethan-Arrowood:
https://github.com/vercel/next.js/pull/60484 has landed. Going to mark this as ready-to-review once `next@canary` is updated.

Ethan-Arrowood:
canary has been updated: https://github.com/vercel/next.js/actions/runs/8253217419

Ethan-Arrowood:
As expected, e2e test failed here: https://github.com/vercel/vercel/actions/runs/8252312985/job/22571633457?pr=11254#step:10:2337

Need to re-run CI now that canary has been released. And we should see this one start to pass.

Ethan-Arrowood:
Still failing on canary (https://github.com/vercel/vercel/actions/runs/8252312985/job/22581229675?pr=11254#step:10:2287) 

so I don't think the fix completely solved this issue. Will continue investigating soon.

Ethan-Arrowood:
Ah I see - this test is not actually on canary. When I switched it from the separate test case to the existing one, I missed that. Fixing now

packages/next/test/fixtures/00-app-dir-ppr/vercel.json
```diff
@@ -339,6 +339,16 @@
       },
       "mustContain": ":{",
       "mustNotContain": "<html"
+    },
+    {
+      "path": "/api/app-headers",
+      "status": 200,
+      "mustContain": "{\"port\":\"80\"}"
```
ijjk: Shouldn't this be 443?

Ethan-Arrowood: No, it's returning 80 for me

Ethan-Arrowood: My assumption is that our live servers have a reverse-proxy set up that is doing TLS at another layer and proxying all internal requests as http; thus, `req.originalRequest.socket.encrypted` is `false`.

ijjk: Seems like we aren't honoring the correct header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto indicates it should be 443 if we are connecting over HTTPS

Ethan-Arrowood: Okay interesting- well that header also is based off the incoming request check. It was before this change as well. I'll take a closer look at the incoming request object and see if there's a better check we can do.

ijjk: Hmm maybe we need to adjust the `isHttps` check to also check `x-forward-proto` being already set to `https`?

ijjk: Hmm maybe we need to adjust the `isHttps` check to also check `x-forward-proto` being already set to `https`?

Ethan-Arrowood: I'm not sure I follow that. `x-forwarded-proto` is set in the same block based on the same check. I don't see it in the raw request headers neither.

Ethan-Arrowood: Furthermore, I'm not seeing anything in the `req` object that would hint at the request being `https` or `http`. I think this is the nature of a reverse-proxy. We might need to discuss with someone from infra

Ethan-Arrowood: The one symbol we might be able to tap into is this: 

```
  [Symbol(NextInternalRequestMeta)]: {
    initURL: 'http://localhost:3000/api/pages-headers',
    initQuery: {},
    initProtocol: 'http',
    clonableBody: {
      finalize: [AsyncFunction: finalize],
      cloneBodyStream: [Function: cloneBodyStream]
    }
  }
  ```
  
  I'm going to see what that logs in a deployment now.

Ethan-Arrowood: I cannot get that symbol to log when deployed. Do you have any ideas?

Ethan-Arrowood: Local: Screenshot 2024-03-13 at 12 14 07

Deployed: `   [Symbol(NextInternalRequestMeta)]: {}`

Ethan-Arrowood: Fixed by: https://github.com/vercel/next.js/pull/63303

ijjk (approved):

PR #11257, opened 2024-03-12
[remix] Add SPA mode e2e tests

TooTallNate:
Adds a couple e2e tests for [Remix SPA](https://remix.run/docs/en/main/future/spa-mode) mode (both with and without using the Vercel Preset).


packages/remix/test/fixtures-vite/04-spa-with-vercel-preset/app/routes/.tsx
EndangeredMassa, line 1: **suggestion (soft-blocking):** Is this file supposed to be `$.tsx`?

This file looks like the same content as `_index.tsx`. Should we make it different so that a current/future probe could distinguish between them?

packages/remix/test/fixtures-vite/04-spa-with-vercel-preset/probes.json
```diff
@@ -0,0 +1,13 @@
...
+  "probes": [
+    {
+      "path": "/",
+      "mustContain": "<p>Loading...</p>"
+    },
+    {
+      "path": "/does-not-exist",
+      "status": 200,
+      "mustContain": "<p>Loading...</p>"
+    }
```
EndangeredMassa, line 18: **convention (non-blocking):** can we add an assertion (or any reason for this to fail) that tests that the preset worked or not?

Maybe adding api routes of both node and edge runtimes would be good enough. The build of this project would fail if the preset wasn't registered, right?‏

EndangeredMassa (dismissed):

packages/remix/test/fixtures-vite/04-spa-with-vercel-preset/app/routes/.tsx
TooTallNate: Didn't mean to commit this file. Will delete.

packages/remix/test/fixtures-vite/04-spa-with-vercel-preset/probes.json
TooTallNate: The Preset doesn't really do much for SPA mode, since it's all client-side files. There's no notion of Node.js vs Edge in SPA mode.

One thing the Preset still helps with in SPA mode is detection of the assets dir, so we can change the client assets dir and ensure the files are being served from there for this test.

TooTallNate: [`c0b863c` (#11257)](https://github.com/vercel/vercel/pull/11257/commits/c0b863cea33eb9c9ab23b094ba76329abc3b8e39)

EndangeredMassa (approved):

PR #11259, opened 2024-03-12
[next] Ensure all static routes have static streaming lambda path

wyattjoh:
When there's a match for a prerender, the dynamic query parameters will not be provided via the `x-now-route-matches`. This ensures that when generating any static route that the static streaming lambda path is used if it's provided, ensuring that when a page is generated via `generateStaticParams`:

```tsx
// app/blog/[slug]/page.tsx

type Props = {
  params: {
    slug: string
  }
}

export function generateStaticParams() {
  return [{ slug: "one" }, { slug: "two" }, { slug: "three" }]
}

export default function BlogPage({ slug }: Props) {
  // ...
}
```

That we use the specific routes (`/blog/one`, `/blog/two`, and `/blog/three`) for the partial prerendering streaming path instead of the paramatarized pathname (`/blog/[slug]`) as the rewrites won't be matched once a match for a prerender has been found.

This additionally updates the tests to ensure that the correct pathname is observed (this was previously silently failing).

### DIFF
--- a/.changeset/brave-parrots-walk.md
+++ b/.changeset/brave-parrots-walk.md
@@ -1,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+Ensure that static PPR pages have static streaming lambda paths.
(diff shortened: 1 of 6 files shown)


wyattjoh:
* **#11259** <a href="https://app.graphite.dev/github/pr/vercel/vercel/11259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ijjk (approved):

PR #11261, opened 2024-03-12
[cli] Set `projectSettings.nodeVersion` in `vc deploy` based on "engines.node" field

TooTallNate:
Read the package.json and use the `engines.node` value to set the `projectSettings.nodeVersion` property on the create deployment API call POST body. This allows for the Node 20 build container image to be properly specified before the deployment is created on the server side.


trek (dismissed):

TooTallNate:
@trek already previously approved.

PR #11268, opened 2024-03-13
[cli] stops warning Next.js users about legacy Speed Insights

feugy:
### 🧐 What's in there?

Vercel's Speed Insights legacy injection is deprecated in favor of `@vercel/speed-insights` package.
CLI has a couple places where it warns users running `vc build` locally (or on their CI).

Some users complained about seeing the warning while they are already using the package.

This PR removes the message

### ❗ Note to reviewers

@tobiaslins, @timolins and I had a long discussion on the approach here.

This warning is issued because during `vc pull`, the project API on Vercel returns a legacy analytics.id field, [which is stored](https://github.com/vercel/vercel/blame/main/packages/cli/src/util/projects/project-settings.ts#L35) in `.vercel/project.json`.
While it is legit for users who never used `@vercel/speed-insights` package, it also appear for users who migrated from legacy.

Besides setting the warning, `VERCEL_ANALYTICS_ID` is set, which triggers auto-injection in Next.js, and is useful during self-hosting.

We first considered removing the legacy id from `project.json`, but it would break self-hosting.
We also considered printing the warning only if the application never used the package (which the project API on Vercel knows). 
In the end, we realized that vercel.com Speed Insights dashboard has plenty of disclaimer and incentives to use the new package. Having it in the CLI doesn't worth the annoyance.

### DIFF
--- a/.changeset/cold-carrots-kick.md
+++ b/.changeset/cold-carrots-kick.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Stops warning about legacy Speed Insights for Next.js apps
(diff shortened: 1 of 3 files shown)


tobiaslins (approved):

philwolstenholme:
Thanks for this, including all the detail on the PR (I was 'Some users' 😄)

trek (approved):

PR #11277, opened 2024-03-14
[cli] Don't send `projectSettings.nodeVersion` for unsupported versions

erikareads:
Previously, we used `semver` which correctly identified the nodeVersion used in `package.json`, but failed in two cases:

1. If the specified `engines` range started too low, it wouldn't correctly identify a later version compatible with Vercel. (`">10"` was detecting as Node `"10.x"` not `"20.x"`)
2. If the specified `engines` version was fixed too low to be compatible with Vercel, we sent it, and then Vercel predictably errored. (`"10"` was detecting as Node `"10.x"` which would lead to an error in the build process)

Now, case 1 should properly use the latest node version. And case 2 will not send a node version override to the API and will print a warning to the user on the CLI, but the deployment will still proceed with the version specified for the project in the web interface.

We chose to keep the error message shown to the user consistent with what shows in the logs for the build container, that is:
> WARN! Node.js Version "10.x" is discontinued and must be upgraded. Please set "engines": { "node": "20.x" } in your `package.json` file to use Node.js 20.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -71,6 +70,7 @@ import { pickOverrides } from '../../util/projects/project-settings';
 import { printDeploymentStatus } from '../../util/deploy/print-deployment-status';
 import { help } from '../help';
 import { deployCommand } from './command';
+import { getSupportedNodeVersion } from '@vercel/build-utils/dist/fs/node-version';
```
TooTallNate: We should avoid importing for `dist` dirs.

```suggestion
import { getSupportedNodeVersion } from '@vercel/build-utils';
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -521,9 +521,15 @@ export default async (client: Client): Promise<number> => {
   );
   let nodeVersion: string | undefined;
   if (packageJson?.engines?.node) {
-    const parsedNodeVersion = semver.coerce(packageJson.engines.node);
-    if (parsedNodeVersion) {
-      nodeVersion = `${parsedNodeVersion.major}.x`;
+    try {
+      const { range } = await getSupportedNodeVersion(
+        packageJson?.engines?.node
```
TooTallNate: Nullish coalescing isn't necessary here since we're already ascertaining in the `if` above:

```suggestion
        packageJson.engines.node
```

packages/cli/test/fixtures/unit/commands/deploy/node-low-starting-range/api/es6.js
```diff
@@ -0,0 +1 @@
+export default (req, res) => res.end('Vercel');
```
TooTallNate: Not a huge deal, but we could remove some of these API functions. Just keep the `index.js` one.

TooTallNate (changes requested):

TooTallNate (approved):

PR #11279, opened 2024-03-14
[cli] Replace `promptBool` with `confirm` prompts

jeffsee55:
The only changes are from some slightly different punctuation and that typing 'y' or 'n' doesn't process the prompt, it now waits for "Enter". Easiest command to see this is an action is the `init`, if you provided an invalid example to initialize `vercel init astroz` and follow the prompts.


packages/cli/test/unit/commands/init.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      await expect(client.stderr).toOutput('? Did you mean astro? [y/N] ');
+      client.stdin.write('y');
+      client.stdin.write('\r'); // Return key
+
+      await expect(client.stderr).toOutput(`Fetching astro`);
+      expect(mock).toHaveBeenCalled();
+
+      await expect(exitCodePromise).resolves.toEqual(0);
+      // The long timeout is because this test actually downloads the example from its source
+    }, 20000);
```
jeffsee55: Not sure what the best approach is here, the `init` command ends by downloading the example source. We could mock that `tar` somehow, ideas welcome

packages/cli/test/unit/commands/init.test.ts
```diff
@@ -0,0 +1,62 @@
+import init from '../../../src/commands/init';
+import { client } from '../../mocks/client';
+import { setupTmpDir } from '../../helpers/setup-unit-fixture';
+import { FetchOptions } from '../../../src/util/client';
+
+// The without setting this, the client.stderr output is undefined
+process.stdout.isTTY = true;
```
jeffsee55: The fact that I needed to add this has me feeling like I've done something wrong, I don't see the need for it in other tests

TooTallNate (dismissed):

TooTallNate (approved):

PR #11283, opened 2024-03-15
[cli] Fix how we determine the GIT_CONFIG_PATH to support git worktrees and git submodules

### LINKED ISSUE #10104: Vercel CLI Does Not Work on submodules
The CLI checks the repo name via `.git/config` but in a submodule the `.git/config` is stored in the main git folder.

dmmulroy:
## Fix how we determine the GIT_CONFIG_PATH to support git worktrees and git submodules

We were previously checking for the root of the directory by looking for `.git/config` if the project was not already linked.

This was causing `ENOTDIR` error when checking `.git` in git worktree or git submodule repositories. This is due to `.git` being a file in those repo types with a pointer to the "parent" or "super" repo rather than a directory.

Resolves #10104


packages/cli/src/util/git-helpers.ts
```diff
@@ -0,0 +1,87 @@
...
+  /** Specifies the current working directory
+   * from which the Git command should be executed. If not specified, it
+   * defaults to the current working directory of the process.*/
+  cwd?: string;
+}>;
+
+const DEFAULT_GIT_EXEC_OPTS = {
+  unsafe: false,
+  cwd: process.cwd(),
+} as const;
```
TooTallNate: Was there a reason for `as const` here? Doesn't seem necessary to me at first glance

```suggestion
};
```

EndangeredMassa:
@dmmulroy can you check out the latest CI failures? We're passed the general CI issue from before and this failing tests looks like it could be legit to your changes: https://github.com/vercel/vercel/actions/runs/8472056694/job/23213501979?pr=11283

packages/cli/src/util/git-helpers.ts
```diff
@@ -0,0 +1,87 @@
...
+  unsafe?: boolean;
+  /** Specifies the current working directory
+   * from which the Git command should be executed. If not specified, it
+   * defaults to the current working directory of the process.*/
+  cwd?: string;
+}>;
+
+const DEFAULT_GIT_EXEC_OPTS = {
+  unsafe: false,
+  cwd: process.cwd(),
```
jeffsee55: Since we potentially change the [cwd from the client](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/client.ts#L244), it might be best to require this value be passed in explicitly to `getGitDirectory` so this doesn't become out of sync

jeffsee55 (commented):
This is awesome @dmmulroy. I was just trying to run tests from a worktree when you pushed this PR up, perfect timing!

packages/cli/src/util/git-helpers.ts
dmmulroy: Sweet that sounds good to me! I'll make that change now.

dmmulroy:
> @dmmulroy can you check out the latest CI failures? We're passed the general CI issue from before and this failing tests looks like it could be legit to your changes: https://github.com/vercel/vercel/actions/runs/8472056694/job/23213501979?pr=11283

@EndangeredMassa I reverted my change and pushed to see if it had any impact in running the tests in CI and they still failed with the original functionality.  (Run here 👉 https://github.com/vercel/vercel/actions/runs/8473209261/job/23217151757)

packages/cli/src/util/git-helpers.ts
dmmulroy: Habit haha

dmmulroy:
@EndangeredMassa @TooTallNate look like all tests and checks are passing 🥳

TooTallNate (approved):

jonnie-sdx:
with latest version `34.3.1` this is still an issue for me

PR #11298, opened 2024-03-19
Add Remotion as a supported framework

JonnyBurger:
This adds [Remotion](https://remotion.dev) as a supported framework on Vercel, meaning that it can be deployed with zero configuration.

I have added information about Remotion, our motivation to integrate with Vercel, use cases and future plans in this [discussion](https://github.com/orgs/vercel/discussions/6175).


EndangeredMassa:
Thanks for the suggestion! Balancing support costs of adding new presets, we're not going to add this at this time.

You can definitely deploy this to Vercel, but you'll have to customize the [Project Settings](https://vercel.com/docs/deployments/configure-a-build#build-and-development-settings) to set your build (and possibly other) commands.

PR #11302, opened 2024-03-19
[cli] Switch unit tests from `jest` to `vitest`

jeffsee55:


EndangeredMassa (dismissed):

EndangeredMassa (commented):
Looks good to me, but I didn't review in detail

TooTallNate (approved):

packages/cli/test/unit/util/to-host.test.ts
```diff
@@ -1,3 +1,4 @@
+import { describe, expect, it } from 'vitest';
```
mcmxcdev, line 1: Just randomly came across this PR, but these imports could be avoided for each file when specifying `globals: true` within `vitest.config.ts` and adding `"vitest/globals"` to `compilerOptions.types` in `tsconfig.json`

See more: https://vitest.dev/config/#globals

EndangeredMassa: Thanks for the note, but we have explicit imports on purpose. We prefer seeing exactly where these functions come from.

PR #11305, opened 2024-03-20
[cli] Ignore VERCEL_ANALYTICS_ID if @vercel/speed-insights is present on the project

chriswdmr:
In the past, we used the `VERCEL_ANALYTICS_ID` environment variable with the previous Speed Insights feature on Vercel to activate specific logic in Next.js, Nuxt and Gatsby for collecting and sending web vitals data.

With the new Speed Insights, that's not required anymore.

We no longer want to set the environment variable when detecting the new `@vercel/speed-insights` package. 
This PR confirms that the variable is not set if the new package is detected.


packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -796,6 +796,17 @@ describe('build', () => {
...
 
+  it('should not set `VERCEL_ANALYTICS_ID` if new @vercel/speed-insights package is present', async () => {
+    const cwd = fixture('vercel-speed-insights');
+    const output = join(cwd, '.vercel/output');
+    client.cwd = cwd;
+    const exitCode = await build(client);
+    expect(exitCode).toEqual(0);
+
+    const env = await fs.readJSON(join(output, 'static', 'env.json'));
+    expect(Object.keys(env).includes('VERCEL_ANALYTICS_ID')).toEqual(false);
```
feugy: **suggestion:** you may turn this [into](https://jestjs.io/docs/expect#tohavepropertykeypath-value) `expect(env).not.toHaveProperty('VERCEL_ANALYTICS_ID')`

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -796,6 +796,17 @@ describe('build', () => {
     expect(Object.keys(env).includes('VERCEL_ANALYTICS_ID')).toEqual(true);
   });
 
+  it('should not set `VERCEL_ANALYTICS_ID` if new @vercel/speed-insights package is present', async () => {
+    const cwd = fixture('vercel-speed-insights');
+    const output = join(cwd, '.vercel/output');
+    client.cwd = cwd;
+    const exitCode = await build(client);
+    expect(exitCode).toEqual(0);
```
feugy: **praise:** Thanks for covering this edge case!

feugy (dismissed):
Let's add a changeset. Thanks for this quick fix!

trek (dismissed):

feugy (dismissed):

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/build/index.test.ts
EndangeredMassa: @chriswdmr This appears to be failing right now: https://github.com/vercel/vercel/actions/runs/8377067208/job/22946124053?pr=11305

chriswdmr: Yeah, I just noticed that our helper function that is used to read the package version might not work in our test suite. I've tested it locally and built two projects and it worked.

I will look into that tomorrow!

EndangeredMassa:
For easier local running of the test:

```
$  pwd
/Users/smassa/source/vercel/vercel-2/packages/cli

$  cd ../.. ; pnpm build ; cd - ; nvm use 18 ; pnpm vitest test/unit/commands/build/index.test.ts
```

chriswdmr:
@EndangeredMassa @feugy I've updated the tests and our handling, please have another look! 🙌

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -787,8 +787,8 @@ describe('build', () => {
     expect(files.sort()).toEqual(['index.html', 'package.json']);
   });
 
-  it('should set `VERCEL_ANALYTICS_ID` environment variable', async () => {
-    const cwd = fixture('vercel-analytics');
+  it('should set `VERCEL_ANALYTICS_ID` environment variable if Vercel Speed Insights is enabled', async () => {
```
feugy, line 790: **suggestion:** could you change to `should set VERCEL_ANALYTICS_ID environment variable if @vercel/speed-insights is not used`?

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1319,4 +1319,30 @@ describe('build', () => {
       'edge.func'
     );
   });
+
+  describe('with Vercel Speed Insights', () => {
```
feugy, line 1323: **question:** is this describe useful? there is no `before` nor `beforeAll`, and I don't understand how we're enabling anything.

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1319,4 +1319,30 @@ describe('build', () => {
...
+
+      client.cwd = cwd;
+      const exitCode = await build(client);
+      expect(exitCode).toEqual(0);
+
+      const env = await fs.readJSON(join(output, 'static', 'env.json'));
+      expect(Object.keys(env).includes('VERCEL_ANALYTICS_ID')).toEqual(false);
+    });
+
+    it('should include VERCEL_ANALYTICS_ID if @vercel/speed-insights is not present', async () => {
```
feugy, line 1336: **question:** is this test different from the one line 790?

packages/next/src/index.ts
```diff
@@ -343,6 +344,27 @@ export const build: BuildV2 = async ({
...
+    if (version) {
+      // Next.js has a built-in integration with Vercel Speed Insights
+      // with the new @vercel/speed-insights package this is no longer needed
+      // and can be removed to avoid duplicate events
+      delete spawnOpts.env.VERCEL_ANALYTICS_ID;
+      delete process.env.VERCEL_ANALYTICS_ID;
+
+      debug(
+        '@vercel/speed-insights is installed, removing VERCEL_ANALYTICS_ID from environment'
+      );
```
feugy, line 359: **praise:** Thanks for these handy debug logs.

packages/static-build/src/index.ts
```diff
@@ -394,6 +395,20 @@ export const build: BuildV2 = async ({
...
+      const speedInsightsVersion = getInstalledPackageVersion(
+        '@vercel/speed-insights'
+      );
+
+      const isSpeedInsightsInstalled = Boolean(speedInsightsVersion);
+
+      if (
+        isSpeedInsightsInstalled &&
+        process.env.VERCEL_ANALYTICS_ID &&
+        ['next', 'nuxtjs'].includes(framework.slug || '')
```
feugy: **though:** shouldn't we include `gatsby` as well?
**suggestion:** maybe we could include debug logs here as in the next.js builder.

feugy (commented):
Just a couple nits. I'm surprised about gastby: where is the part deleting the env var? Do we have another test proving it could still inject script when `@vercel/speed-insights` isn't in dependency list?

packages/cli/test/unit/commands/build/index.test.ts
chriswdmr: I've mostly did this for a better grouping. The enablement is part of the fixture itself

packages/cli/test/unit/commands/build/index.test.ts
chriswdmr: It should always set it if it's enabled. The other part is tested in each builder where we unset it if the package is installed

packages/cli/test/unit/commands/build/index.test.ts
chriswdmr: Yes, because we always want to set the env and only unset it if the package is detected

packages/static-build/src/index.ts
chriswdmr: Oh you're right. I forgot to push that.
+1 on the debug log. Adding it

chriswdmr:
> Just a couple nits. I'm surprised about gastby: where is the part deleting the env var? Do we have another test proving it could still inject script when `@vercel/speed-insights` isn't in dependency list?

Yes, we already do that here: https://github.com/vercel/vercel/blob/c4d7b8a11046f32f8cc3ca87275d42299e197ddb/packages/static-build/test/builds.test.js#L38

feugy (approved):
<picture data-single-emoji=":ty-dark-intense:" title=":ty-dark-intense:">:ty-dark-intense:</picture>

EndangeredMassa (approved):

PR #11307, opened 2024-03-20
[remix] Add `mjs` and `mts` extensions to vite detection

frontsideair:
My project was being incorrectly detected as not vite, thus it was failing in runtime. My vite config is `vite.config.mts` so this PR fixes detection logic to include `mjs` and `mts` extensions.

The fix can be generalized if needed.


TooTallNate (changes requested):
Thanks for the PR. Could you fix the lint please (run prettier)

frontsideair:
@TooTallNate done!

TooTallNate (approved):

PR #11308, opened 2024-03-20
[examples] Update to latest SolidStart

ryansolid:
We've done a pretty significant update over the past few months on SolidStart and 1.0 is looming. Thought I'd update the official templates to get in line before we do the release.

I wasn't sure what to do with the `output` fields as we use Vercel's build output V3 here. But otherwise I think everything should be good.


TooTallNate (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -554,15 +554,15 @@ export const frameworks = [
...
     detectors: {
       every: [
         {
           matchPackage: 'solid-js',
         },
         {
-          matchPackage: 'solid-start',
+          matchPackage: '@solidjs/start',
         },
       ],
```
TooTallNate: So this would break framework detection for older versions of SolidStart. How about this instead?

```suggestion
      some: [
        {
          matchPackage: 'solid-start',
        },
        {
          matchPackage: '@solidjs/start',
        },
      ],
```

TooTallNate (changes requested):

packages/frameworks/src/frameworks.ts
EndangeredMassa: We may want multiple entries. Some projects will not have `vinxi`.

packages/frameworks/src/frameworks.ts
```diff
@@ -584,6 +584,43 @@ export const frameworks = [
...
+      },
+      devCommand: {
+        value: 'vinxi dev',
+      },
+      outputDirectory: {
+        value: '.output',
+      },
+    },
+    getOutputDirName: async () => '.output',
+  },
```
EndangeredMassa: **suggestion (soft-blocking):** flip order

These entries are matched in order until the first match happens. If a project was going to match both for some reason, we might want v1 to be the result.

I think we should switch the order of these SolidStart entries.

EndangeredMassa (dismissed):

TooTallNate (approved):

TooTallNate:
Thanks Ryan! And thank you Jeff for getting this over the finish line.

PR #11309, opened 2024-03-20
[cli][client] Handle `--repo` linked in `vc deploy --prebuilt`

### LINKED ISSUE #11097: vercel@33.3.0 broke preBuilds in monorepo

TooTallNate:
`vc deploy --prebuilt` has not properly worked for projects linked with `vc link --repo` ever since https://github.com/vercel/vercel/issues/11077.

To fix, don't mutate `cwd` to be the project subdir (so that `cwd` is always the monorepo root), and instead pass in a new property to `@vercel/client` that represents the `.vercel/output` file path. A lot of the code is just passing this new variable through the multiple layers of indirection in CLI before it finally reaches `@vercel/client.

Fixes https://github.com/vercel/vercel/issues/11097.


packages/client/src/utils/index.ts
```diff
@@ -148,20 +149,27 @@ export async function buildFileTree(
...
+    if (typeof vercelOutputDir !== 'string') {
+      throw new Error(`Missing required \`vercelOutputDir\` parameter`);
+    }
+    if (typeof cwd !== 'string') {
+      throw new Error(`\`cwd\` must be a "string"`);
+    }
+    const relOutputDir = relative(cwd, vercelOutputDir);
     ignores = ['*'];
-    const parts = outputDir.split('/');
+    const parts = relOutputDir.split(sep);
```
erikareads, line 169: Why does this use `sep` when the next two calls use raw `'/'`?

packages/client/src/utils/index.ts
```diff
@@ -148,20 +149,27 @@ export async function buildFileTree(
...
     parts.forEach((_, i) => {
       const level = parts.slice(0, i + 1).join('/');
       ignores.push(`!${level}`);
     });
-    ignores.push(`!${outputDir}/**`);
+    ignores.push(`!${parts.join('/')}/**`);
```
erikareads, line 174: Is there a path aware join that we could use here instead of the array join?

packages/client/src/utils/index.ts
TooTallNate: For this (and your next comment), it's because the input is (potentially) a Windows path (with `\`), but the `ignore` module expects Unix paths

packages/client/src/utils/index.ts
```diff
@@ -148,20 +149,27 @@ export async function buildFileTree(
...
+  prebuilt?: boolean,
+  vercelOutputDir?: string
 ): Promise<{ ig: Ignore; ignores: string[] }> {
   const ig = ignore();
   let ignores: string[];
 
   if (prebuilt) {
-    const outputDir = '.vercel/output';
+    if (typeof vercelOutputDir !== 'string') {
+      throw new Error(`Missing required \`vercelOutputDir\` parameter`);
```
EndangeredMassa:
```suggestion
      throw new Error(`Missing required \`vercelOutputDir\` parameter when "prebuilt" is true);
```

EndangeredMassa (dismissed):
I manually tested this and it worked for me in a monorepo that was `--repo` linked

EndangeredMassa:
updated to include the CI fix from main

EndangeredMassa (approved):

PR #11310, opened 2024-03-20
@vercel/node: Move @types/node to dev dependencies

justjake:
This is forcing @types/node v14 to install into Notion's node_modules, which is a bit annoying since node v14 is quite old


EndangeredMassa:
This has been updated. It has to match our minimum supported version, which is currently Node 16.
https://github.com/vercel/vercel/blob/main/packages/node/package.json#L27C4-L27C31

PR #11319, opened 2024-03-22
[examples] Enforce node18 on Python and Ruby examples

ofhouse:
We have to enforce the node version on some examples to make sure they are not auto upgraded in the future before we support their runtime.

See https://vercel.com/docs/deployments/build-image


janory (approved):
LGTM! 👍

TooTallNate (approved):

PR #11321, opened 2024-03-22
[cli] Replace `inquirer` with `@inquirer/prompts`

onsclom:
# Overview

Changing to `@inquirer/prompts` meant updating all existing usage of `inquirer.prompt`. This removed a lot of code since the `inquirer.prompt` function was extra verbose.

The user experience shouldn't have changed much besides different styling. `patch-inquirer` is incompatible with `@inquirer/prompts`. ~~We probably don't want to merge this until we have a new solution for styling.~~ The new default styling is preferred over our old styling.

There are lots of changes over many files, so I went back and organized the commit history. It will probably be easier to review this PR by commit.

## Tests

~~I had to change tests in ways that aren't great. Tests had a hard time recognizing output with escape codes in them, so I had to make the expected output more generic.~~ Using `strip-ansi` worked nicely to ignore styling changes in unit and integration tests.

I removed the unit test "should list projects running on an soon-to-be-deprecated Node.js version" because that test doing `jest.useFakeTimers().setSystemTime(new Date('2023-12-08'));` would make the "should remove a project" test fail. Really strange stuff! Couldn't think of a better solution than removing this, so looking for suggestions here.


onsclom:
Visual problem with `@inquirer/input` affecting this PR: https://github.com/SBoudrias/Inquirer.js/pull/1375

~~Maybe we should wait until that fix is merged, or maybe I just vendor in that component and solution for now?~~ It was merged!

TooTallNate (approved):

PR #11329, opened 2024-03-25
[tests] Use AL2 build container for relevant e2e tests

TooTallNate:
* Forces the AL2 build container image for fixtures that depend on it, via `engines.node` in package.json for most cases.
* The `testDeployment()` function was updated to send `projectSettings.nodeVersion` in the POST body, to mimic the behavior in CLI.
* For Go, Ruby, and Python tests, the `projectSettings.nodeVersion` property is set "globally" in the Jest setup file, so that individual fixtures didn't need to be adjusted.


examples/__tests__/test-utils.ts
```diff
@@ -4,7 +4,7 @@ import { lstatSync, readdirSync } from 'fs';
 export async function deployExample(filename: string) {
   const { testDeployment } = require('../../test/lib/deployment/test-deployment.js');
   const example = basename(filename).replace(/\.test\.ts$/, '');
-  await testDeployment(join(filename, '..', '..', '..', example));
+  await testDeployment(join(process.cwd(), example));
```
TooTallNate, line 7: The `scanParentDirs()` function from build-utils requires an absolute path.

trek (approved):

examples/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "devDependencies": {
     "@types/jest": "27.4.1",
+    "@vercel/build-utils": "7.9.1",
```
TooTallNate, line 12: The `testDeployment()` function is now calling `scanParentDirs()`, so build-utils needs to be a dep of `examples` so that it gets built by turbo in CI.

packages/cli/test/dev/fixtures/08-hugo/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "private": true,
+  "engines": {
+    "node": "16.x"
```
TooTallNate, line 4: For these `vc dev` fixtures specifically, the node version needs to jive with the version of Node the GH Actions runner is using, since we launch `vc dev` directly in the container.

packages/node/test/fixtures/07-content-length/probe.js
```diff
@@ -1,13 +0,0 @@
-const { strictEqual } = require('assert');
-
-async function test3({ deploymentUrl, fetch, randomness }) {
-  const bodyMustBe = `${randomness}:content-length`;
-  const resp = await fetch(`https://${deploymentUrl}/test3.js`);
-  strictEqual(resp.status, 401);
-  strictEqual(await resp.text(), bodyMustBe);
-  strictEqual(resp.headers.get('content-length'), String(bodyMustBe.length));
```
TooTallNate, line 8: This test was testing behavior that was non-standard (Node does not send `content-length` by default) and was a specific behavior to Vercel. Recently, the node-bridge was updated to be in streaming mode by default, so this behavior changed to not include `content-length` by default.

I deleted the test because it was not testing anything meaningful at this point.

packages/remix/test/fixtures-legacy/01-remix-basics/package.json
```diff
@@ -23,6 +23,6 @@
     "typescript": "^4.6.4"
   },
   "engines": {
-    "node": ">=14"
+    "node": "18.x"
```
TooTallNate, line 26: These Remix v1 tests were failing on Node 20 because of how Remix polyfills APIs such as `URLSearchParams`. So Node 20 would barf upon receiving the polyfilled implementations and the endpoint would fail with 500.

test/lib/deployment/test-deployment.js
```diff
@@ -348,6 +352,34 @@ async function testDeployment(fixturePath, opts) {
   const uploadNowJson = nowJson.uploadNowJson;
   delete nowJson.uploadNowJson;
 
+  // Set `projectSettings.nodeVersion` based on the "engines.node" field of
```
TooTallNate, line 355: This is probably the most important change in this PR. We mimic the logic in CLI which reads the `engines.node` field from package.json and sets that as `projectSettings.nodeVersion` in the create deployment POST body. This covers most of the Node.js-related test fixtures.

PR #11333, opened 2024-03-26
IAM 1720 Added sunset warning to the secrets command

marcgreenstock:
This PR adds a warning to the secrets command that the command will be sunset on 1st May.


trek (dismissed):

packages/cli/src/commands/secrets/command.js
```diff
@@ -2,9 +2,14 @@ import { packageName, getCommandName } from '../../util/pkg-name';
...
-    'env'
-  )} command is recommended instead of ${getCommandName('secrets')}`,
+  description:
+    'WARNING: On May 1st, 2024 secrets will be automatically converted to sensitive Environment Variables for Preview and Production environments. Secrets attached to Development environments will not be migrated.\n' +
+    `The ${getCommandName(
+      'secrets'
+    )} command will be deprecated at this time, please use the ${getCommandName(
+      'env'
+    )} command instead.\n` +
+    'See https://vercel.com/changelog/legacy-environment-variable-secrets-are-being-sunset for more information.',
```
EndangeredMassa, line 12: **suggestion (soft-blocking):** use vercel.link

In this or a follow-up PR: Can we change this out for a vercel.link that points there instead? We typically use these from Vercel CLI messages for tracking and optional redirecting later. ([Examples](https://github.com/search?q=repo%3Avercel%2Fvercel%20vercel.link&type=code))

---

For reference: The vercel.link URLs are managed in bit.ly via Okta.‏

EndangeredMassa (approved):
Thanks!

EndangeredMassa:
Added automerge so that this should get in after we fix the CI issues in: https://github.com/vercel/vercel/pull/11329

PR #11341, opened 2024-03-28
[cli] Add explicit vitest import to unit test for isZeroConfigBuild utility

erikareads:


TooTallNate (approved):

EndangeredMassa:
force merging because CI is broken and this can't break it any further

PR #11345, opened 2024-03-29
Ignore mock client TS errors

jeffsee55:
~~I think the appropriate fix is probably more nuanced than simply reverting changes here, but this does fix the ts errors in our unit tests~~ See below

Context: https://github.com/vercel/vercel/pull/10090#pullrequestreview-1504414187


jeffsee55:
Yeah that first effort wasn't gonna work. I think the basically the issue is the mock client uses streams to mock tty stdio and those types are never going to overlap properly.

I tried for a bit to use stubbed `tty.WriteStream/tty.ReadStream` but couldn't get everything working.

This PR pretty much just ignores the type errors in the mock client so we don't see a bunch of errors in the tests.

PR #11348, opened 2024-03-29
[cli] optional override of existing environment variables

### LINKED ISSUE #11130: `vercel` cli has not option to update existing values
Hello. We use the vercel cli in our CICD platform and we've noticed that using `vercel env add` command will fail one existing secrets. It would be ideal if the CLI had an `--update` flag to make the vercel CLI overwrite the existing env secret rather than failing with an error. This will ensure we set the most recent secrets and not keeping around any old values.

mountainash:
## Why

Closes #11130

## Changelog

Added `--force` argument to upsert existing environment variables (#11130)

## References

- Vercel API `/v10/projects/{idOrName}/env` https://vercel.com/docs/rest-api/endpoints/projects#create-one-or-more-environment-variables


packages/cli/src/commands/env/add.ts
```diff
@@ -20,6 +20,7 @@ import { isAPIError } from '../../util/errors-ts';
 type Options = {
   '--debug': boolean;
   '--sensitive': boolean;
+  '--yes': boolean;
```
EndangeredMassa: **suggestion (blocking):** ‏use `--force`

We use `--yes` to choose the default answer for a prompt to the user without actually prompting. The intended behavior of this PR is more in line with how we use the `--force` option.

Can you change this to `--force`?

packages/cli/src/util/env/add-env-record.ts
```diff
@@ -26,7 +28,9 @@ export default async function addEnvRecord(
     target: targets,
     gitBranch: gitBranch || undefined,
   };
-  const url = `/v8/projects/${projectId}/env`;
+  const args = upsert ? `?upsert=${upsert}` : '';
+  const version = upsert ? 'v10' : 'v8';
```
EndangeredMassa: **suggestion (blocking):** ‏use v10 always

I appreciate the careful use of v10 only for what's needed here, but I think we can and should update this call to always use v10. The only other difference between v8 and v10 is that more detailed error codes will be returned.

To accommodate, we can update the expected error codes in `util/env/known-error.ts` to be:

```
const knownErrorsCodes = new Set([
  'BAD_REQUEST',
  'ENV_ALREADY_EXISTS',
  'ENV_CONFLICT',
  'ENV_SHOULD_BE_A_SECRET',
  'EXISTING_KEY_AND_TARGET',
  'FORBIDDEN',
  'ID_NOT_FOUND',
  'INVALID_KEY',
  'INVALID_VALUE',
  'KEY_INVALID_CHARACTERS',
  'KEY_INVALID_LENGTH',
  'KEY_RESERVED',
  'RESERVED_ENV_VARIABLE',
  'MAX_ENVS_EXCEEDED',
  'MISSING_ID',
  'MISSING_KEY',
  'MISSING_TARGET',
  'MISSING_VALUE',
  'NOT_AUTHORIZED',
  'NOT_DECRYPTABLE',
  'SECRET_MISSING',
  'SYSTEM_ENV_WITH_VALUE',
  'TEAM_NOT_FOUND',
  'TOO_MANY_IDS',
  'TOO_MANY_KEYS',
  'UNKNOWN_ERROR',
  'VALUE_INVALID_LENGTH',
  'VALUE_INVALID_TYPE',
]);
```

EndangeredMassa (changes requested):
Thanks for the detailed PR with tests! I think we can accept this change if we make some tweaks.

packages/cli/src/util/env/add-env-record.ts
mountainash: I can do this 🕙 

Just a query about the update of these error codes - `PAYMENT_REQUIRED` is on the current list, but not on the update you've supplied. Can you confirm that it's been removed?

mountainash: I've updated all the `url`s using the v8 env endpoint to now use v10. Commit also includes the updated error code list supplied by @EndangeredMassa. See 423faf0e119361bcef0f506b3618b2fcb76a97b1

EndangeredMassa: Good eye, but yes, I left `PAYMENT_REQUIRED` off on purpose. I was comparing the list to what's used on the API side and that's not an expected code for this.

EndangeredMassa:
I made some [tweaks](https://github.com/vercel/vercel/pull/11348/commits/84f69e9b5f07ae94e539f433507419b331b95119) to flesh out the new option.

mountainash (commented):
All valid to me. Thanks. 🙏

EndangeredMassa (approved):

EndangeredMassa:
Thanks @mountainash!

crbraun:
Is it possible this caused an issue using env variables?  Just recently I am no longer able to call my Dynamo endpoint

Error fetching item from DynamoDB UnrecognizedClientException: The security token included in the request is invalid

       const dynamoClient: DynamoDBClient = new DynamoDBClient({region: AWS_REGION});
        // Use GetItemCommand to get the item from the table
        const {Item} = await dynamoClient.send(new GetItemCommand(params));


Works in my local env.  Verified my env variables have not changed.   Both AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID exist and are populated with the correct key.

EndangeredMassa:
@crbraun what version of the Vercel CLI (`vercel`) are you using? Can you verify that the env var is available at that time in your code by logging it out?

crbraun:
> @crbraun what version of the Vercel CLI (`vercel`) are you using? Can you verify that the env var is available at that time in your code by logging it out?

I did verify that AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are present.  Oddly there there is a AWS_SESSION_TOKEN present in the environment which my application never sets.  I expect this is the issue.  But I am not sure how this got in the environment as I do not have this variable set anywhere.

crbraun:
I was able to work around this issue by adding an empty AWS_SESSION_TOKEN env variable to my environment.  I expect there is something wrong with the vercel deployer as it seems to be setting an AWS_SESSION_TOKEN when there is none present in the configuration.

EndangeredMassa:
Hrm. I don't think that should be happening. If you open a support case, we can look into it. https://vercel.com/contact

crbraun:
It doesn't look like I am allowed to create a support ticket

> Visit our Community for further help or create a paid account for dedicated Builds & Deployments → Builds support.

I have it worked around for now, but I am fairly certain there was a recent change in vercel that was deployed today that caused this issue.  Let me know if there is a way I can log an issue with support, be glad to do so.  

-- Thanks

PR #11351, opened 2024-03-29
[client] Always use v13 of create deployment API endpoint

TooTallNate:
Updates `@vercel/client` to always use the v13 create-deployment endpoint, even when `builds` is present. This allows for `projectSettings.nodeVersion` to be passed when `builds` is present.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -584,6 +566,30 @@ export default async (client: Client): Promise<number> => {
...
+    if (packageJson?.engines?.node) {
+      try {
+        const { range } = await getSupportedNodeVersion(
+          packageJson.engines.node
+        );
+        nodeVersion = range;
+      } catch (error) {
+        if (error instanceof Error) {
+          output.warn(error.message);
+        }
```
EndangeredMassa, line 587: **suggestion (soft-blocking):** log other values

If for some reason this variable isn't an instance of `Error`, we should still log it.

```suggestion
        if (error instanceof Error) {
          output.warn(error.message);
        } else {
          output.warn(error.toString());
        }
```

EndangeredMassa (approved):

PR #11353, opened 2024-03-29
[cli] replace console.log that communicates to user with output.log

erikareads:
We have several instances of `console.log` being used with colors or other indicators that they're trying to communicate to the user, and not in a way that makes sense for standard output. These should be migrated to use the `client.output` methods where possible.


packages/cli/src/commands/teams/add.ts
```diff
@@ -95,8 +84,8 @@ export default async function add(client: Client): Promise<number> {
     });
   } catch (err: unknown) {
     if (isError(err) && err.message === 'USER_ABORT') {
-      console.log(info('No name specified'));
-      return gracefulExit();
+      output.log('No name specified');
```
erikareads, line 87: Perhaps make an `output.error`?

jeffsee55 (approved):

PR #11362, opened 2024-04-01
Add matchUnescapedFirst to routing doc

ctgowrie:
Adds a new field to routing API to allow optionally matching on an unescaped path first

### DIFF
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -116,6 +116,9 @@ export const routesSchema = {
           isInternal: {
             type: 'boolean',
           },
+          matchUnescapedFirst: {
+            type: 'boolean',
+          },
           status: {
             type: 'integer',
             minimum: 100,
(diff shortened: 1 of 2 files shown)


ijjk (commented):
We also need to update the route type here to allow it https://github.com/vercel/vercel/blob/326fe0f0e647a80ef504ca4035c6bc3c206f43d6/packages/routing-utils/src/types.ts#L53

packages/routing-utils/src/schemas.ts
```diff
@@ -116,6 +116,9 @@ export const routesSchema = {
           isInternal: {
             type: 'boolean',
           },
+          matchUnescapedFirst: {
```
ijjk, line 119: Should we shorten this to just `unescape: true | undefined` or would that be too vague on what's being unescaped?

EndangeredMassa: The longer name makes more sense to me

EndangeredMassa (approved):

ijjk (approved):

PR #11364, opened 2024-04-01
[node] Switch to `tsx` instead of `ts-node` during `vc dev`

TooTallNate:
For `vercel dev` with Node.js `api` directory functions, use [`tsx`](https://www.npmjs.com/package/tsx) to execute the user code instead of `ts-node`. `tsx` handles more configuration out of the box and fixes a regression that was introduced in v29.0.0(?) of Vercel CLI where TypeScript is used without `type: 'module` in package.json. Two related tests that were testing for a 500 are now testing for the successful execution.


packages/node/src/index.ts
```diff
@@ -53,7 +53,7 @@ import type {
 } from '@vercel/build-utils';
 import { getConfig } from '@vercel/static-config';
 
-import { fixConfig, Register, register } from './typescript';
```
EndangeredMassa, line 56: **suggestion (soft-blocking):** remove export

I think we don't need to export `fixConfig` anymore:

```diff
diff --git a/packages/node/src/typescript.ts b/packages/node/src/typescript.ts
index da8db40e9..dee4efa82 100644
--- a/packages/node/src/typescript.ts
+++ b/packages/node/src/typescript.ts
@@ -449,7 +449,7 @@ type GetOutputFunction = (code: string, fileName: string) => SourceOutput;
 /**
  * Do post-processing on config options to support `ts-node`.
  */
-export function fixConfig(
+function fixConfig(
   config: { compilerOptions: any },
   nodeVersionMajor = 12
 ) {
```

packages/cli/test/dev/integration-5.test.ts
```diff
@@ -536,10 +539,11 @@ describe('[vercel dev] ESM serverless functions', () => {
...
-        // bad gateway
-        // require() of ESM Module
-        await testPath(500, '/api/data');
+      async (testPath: any, port: any) => {
+        const res = await fetch(`http://localhost:${port}/api/data`);
+        validateResponseHeaders(res);
+        const json = await res.json();
+        expect(json).toHaveProperty('isLeapYear');
       },
       { skipDeploy: true }
```
EndangeredMassa, line 548: **question (blocking):** ‏do these pass if we remove `skipDeploy`?

I think we should compare these to production, if possible.

```diff
diff --git a/packages/cli/test/dev/integration-5.test.ts b/packages/cli/test/dev/integration-5.test.ts
index a37d599de..0bd1b34ec 100644
--- a/packages/cli/test/dev/integration-5.test.ts
+++ b/packages/cli/test/dev/integration-5.test.ts
@@ -442,8 +442,7 @@ describe('[vercel dev] ESM edge functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -456,8 +455,7 @@ describe('[vercel dev] ESM edge functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -470,8 +468,7 @@ describe('[vercel dev] ESM edge functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -484,8 +481,7 @@ describe('[vercel dev] ESM edge functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );
 });
@@ -500,8 +496,7 @@ describe('[vercel dev] ESM serverless functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -514,8 +509,7 @@ describe('[vercel dev] ESM serverless functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -530,8 +524,7 @@ describe('[vercel dev] ESM serverless functions', () => {
         console.log({ text });
         const json = JSON.parse(text);
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );

@@ -544,8 +537,7 @@ describe('[vercel dev] ESM serverless functions', () => {
         validateResponseHeaders(res);
         const json = await res.json();
         expect(json).toHaveProperty('isLeapYear');
-      },
-      { skipDeploy: true }
+      }
     )
   );
 });
diff --git a/packages/node/src/typescript.ts b/packages/node/src/typescript.ts
index da8db40e9..dee4efa82 100644
--- a/packages/node/src/typescript.ts
+++ b/packages/node/src/typescript.ts
@@ -449,7 +449,7 @@ type GetOutputFunction = (code: string, fileName: string) => SourceOutput;
 /**
  * Do post-processing on config options to support `ts-node`.
  */
-export function fixConfig(
+function fixConfig(
   config: { compilerOptions: any },
   nodeVersionMajor = 12
 ) {
```

This will be slow, but the protection against regressions here may be worth it.

EndangeredMassa (changes requested):

packages/cli/test/dev/integration-5.test.ts
```diff
@@ -523,10 +523,13 @@ describe('[vercel dev] ESM serverless functions', () => {
...
       'esm-js-nodejs-no-module',
-      async (testPath: any) => {
-        // bad gateway
-        // require() of ESM Module
-        await testPath(500, '/api/data');
+      async (testPath: any, port: any) => {
+        const res = await fetch(`http://localhost:${port}/api/data`);
+        validateResponseHeaders(res);
+        const text = await res.text();
+        console.log({ text });
```
styfle, line 530: Remove console.log()?

packages/cli/test/dev/integration-5.test.ts
```diff
@@ -523,10 +523,13 @@ describe('[vercel dev] ESM serverless functions', () => {
     '[vercel dev] ESM .js type=commonjs',
     testFixtureStdio(
       'esm-js-nodejs-no-module',
-      async (testPath: any) => {
-        // bad gateway
-        // require() of ESM Module
```
styfle, line 528: This test seemed to indicate that 500 was expected. I don't think we should change the behavior because thats not how Node.js works, right?

styfle: For example

```
npm init -y
npm add leap-year
echo "var isLeapYear = require('leap-year')" > file.js
node file.js

var isLeapYear = require('leap-year')
                   ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/styfle/Code/foo/leapyear/node_modules/.pnpm/leap-year@4.0.0/node_modules/leap-year/index.js from /Users/styfle/Code/foo/leapyear/file.js not supported.
```

packages/cli/test/dev/integration-5.test.ts
styfle: Agreed - this behavior should match deployments. I wonder why it was skipped? 🤔

packages/cli/test/dev/integration-5.test.ts
TooTallNate: You are correct about that actually. Might need to reconsider this approach then <picture data-single-emoji=":thinkies:" title=":thinkies:">:thinkies:</picture>

PR #11370, opened 2024-04-02
[build-utils] Add `getOsRelease()` and `getProvidedRuntime()` functions

TooTallNate:
Adds two new helper functions that custom runtimes can utilize in order to determine whether they should use `provided.al2` or `provided.al2023` Lambda "runtime" value, based on the build environment.

The new `getProvidedRuntime()` helper function is being used in `@vercel/go`, so that Go endpoints may be used when the `20.x` Node Version project setting is used.


packages/build-utils/src/os.ts
```diff
@@ -0,0 +1,38 @@
...
+    obj[m.groups.key] = m.groups.value;
+  }
+  return obj;
+}
+
+export async function getProvidedRuntime() {
+  const os = await getOsRelease();
+  return os?.PRETTY_NAME === 'Amazon Linux 2'
+    ? 'provided.al2'
+    : 'provided.al2023';
```
EndangeredMassa: **suggestion (non-blocking):** make fallback explicit

If `getOsRelease()` returns `null`, this logic will return `provided.al2023`, which means that's the fallback value. I think that's correct, but the this function doesn't make it super clear that this happens.

Can we make this explicit?

```suggestion
  const os = await getOsRelease();
  if (!os) {
    return 'provided.al2023';
  }

  return os.PRETTY_NAME === 'Amazon Linux 2'
    ? 'provided.al2'
    : 'provided.al2023';
```

EndangeredMassa (dismissed):

PR #11371, opened 2024-04-02
[node] Add import and require ts-node files for TypeScript in `vc dev`

TooTallNate:
Fixes the case in `vc dev` where importing a `.ts` file from another `.ts` file was failing in `vc dev` when `type: "module"` is not set in `package.json`.

Supersedes https://github.com/vercel/vercel/pull/11364.


EndangeredMassa (approved):
nice!

PR #11375, opened 2024-04-03
Introduce snapshot testing utilities

jeffsee55:
This introduces some additional capabilities on the `MockClient` and `MockStream` classes used in our unit tests. 

The bulk of this code is copied over from https://github.com/SBoudrias/Inquirer.js/tree/master/packages/testing, but uses our existing stdin/stdout mocks instead of what's used there. 

You can now use `client.getScreen()` to see the "last" screen pushed to stdin, which is super handy for snapshot tests. I've added a few as part of this PR, but many more can be written for our client output as well in the near future.

---

The `client.getScreen()` is also a handy debug tool, use it with `raw: true` to see a pretty printed output during tests:

https://github.com/vercel/vercel/assets/5414297/d0be6078-4901-4eb2-a6c9-1b51b576c1ba

---

This also adds an `events` object on the `MockClient` which allows us to write `client.events.keypress("down")` instead of the less readable `client.stdin.write('\x1B[B')` throughout our testing. This could also be moved to so that it's `client.stdin.keypress(...` or something similar. Note that I've only tested that this works in a login test, included in this PR. If we're happy to make this change I will make update the rest of the unit tests to use this syntax for the inputs.


packages/cli/test/unit/commands/login.test.ts
```diff
@@ -87,9 +87,9 @@ describe('login', () => {
         `Success! Email authentication complete for ${user.email}`
       );
 
-      await expect(client.stderr).not.toOutput(emoji('tip'));
-
       await expect(exitCodePromise).resolves.toEqual(0);
+
+      await expect(client.getFullOutput()).not.toContain(emoji('tip'));
```
jeffsee55, line 92: When we do `.not.toOutput` it seems like that would guarantee that our test will wait the full 3 seconds for each one of those statements. But i wonder if we can instead just assert that the "full output" doesn't contain the string, as shown here

packages/cli/test/mocks/client.ts
```diff
@@ -29,6 +36,40 @@ class MockStream extends PassThrough {
...
+    // unlikely to be used by end users or part of prompt code.
+    if (!ignoredAnsi.has(str)) {
+      this.#_rawChunks.push(str);
+    }
+
+    // Stripping the ANSI codes here because Inquirer will push commands ANSI (like cursor move.)
+    // This is probably fine since we don't care about those for testing; but this could become
+    // an issue if we ever want to test for those.
+    if (stripAnsi(str).trim().length > 0) {
+      this.#_chunks.push(str);
```
EndangeredMassa, line 59: **question (non-blocking):** ‏should this push the stripped version instead of the original `str`?

packages/cli/test/unit/commands/login.test.ts
TooTallNate: Correct, `.not.toOutput` has to wait for a timeout, so we generally try to avoid it.

TooTallNate (dismissed):

packages/cli/test/mocks/client.ts
```diff
@@ -112,6 +153,41 @@ export class MockClient extends Client {
...
+
+  getScreen({ raw }: { raw?: boolean } = {}): string {
+    const stderr = client.stderr as any as MockStream;
+    const lastScreen = stderr.getLastChunk({ raw });
+    return raw ? lastScreen : stripAnsi(lastScreen).trim();
+  }
+  getFullOutput(): string {
+    const stderr = client.stderr as any as MockStream;
+    return stderr.getFullOutput();
+  }
```
EndangeredMassa: **suggestion (soft-blocking):** if we can remove type assertions, that'd be great

When I remove this locally, it seems to work fine.

```suggestion
  getScreen({ raw }: { raw?: boolean } = {}): string {
    const lastScreen = client.stderr.getLastChunk({ raw });
    return raw ? lastScreen : stripAnsi(lastScreen).trim();
  }
  getFullOutput(): string {
    return client.stderr.getFullOutput();
  }
```

EndangeredMassa (dismissed):
I see type errors locally where `MockStream` is not compatible with `WriteStream` in various places. Does your IDE show any type errors in `packages/cli/test/mocks/client.ts`?

packages/cli/test/mocks/client.ts
jeffsee55: Yes makes sense. I think I was trying to do some type fixes to the MockClient (that didn't work) and this was some leftover code.

packages/cli/test/mocks/client.ts
jeffsee55: I believe the idea here is to determine _if_ we should log this string. As in, if we remove any ansi sequences from the string, is there anything to show? If so, add it. And then the actual stripping is done in the `getScreen` call.

For context, this is copied from inquirer test https://github.com/SBoudrias/Inquirer.js/blob/master/packages/testing/src/index.mts

packages/cli/test/mocks/client.ts
jeffsee55: Regarding the type errors, these are kind of everywhere and have been for a little while as far as I know. Not related to this PR, unless you're seeing something different. I don't think there's a great solution to fix this, have a [draft PR here](https://github.com/vercel/vercel/pull/11345) that I've been meaning to revisit.

packages/cli/test/mocks/client.ts
EndangeredMassa: Ohh, got it

EndangeredMassa (dismissed):

TooTallNate (dismissed):

packages/cli/test/unit/util/input/playground.test.ts
```diff
@@ -0,0 +1,246 @@
...
+const theme = {
+  // Override spinner with a single frame
+  spinner: { frames: ['O'] },
+};
+
+// Inquirer's unicode detection is outdated, for windows in CI
+// it incorrectly detects that unicode is not supported,
+// and thus prints ascii characters, which breaks these tests.
+// https://github.com/SBoudrias/Inquirer.js/issues/1386
+describe.skipIf(isWindows)('client.input', () => {
```
jeffsee55, line 14: Just skipping these on Windows

EndangeredMassa: Thanks for the comment in code about this. Makes sense.

EndangeredMassa (approved):

PR #11378, opened 2024-04-04
disable eslint for a few files to mitigate flakey linting

EndangeredMassa:
Sometimes, we get errors from eslint saying that we're duplicating the definition of a global, like `Response`. But then later, with no obvious changes, that linting rule will be fine.

This PR eslint ignores the specific files, which seems like a fine tradeoff for this problem.


EndangeredMassa:
forcing a merge here to fix CI for all PRs

PR #11387, opened 2024-04-04
[cli] Unify table formatting output

onsclom:
This removes 'text-table' as a dependency in favor of using 'cli-table3', and simplifies table formatting logic.


packages/cli/package.json
```diff
@@ -83,7 +83,6 @@
     "@types/qs": "6.9.7",
     "@types/semver": "6.0.1",
     "@types/tar-fs": "1.16.1",
-    "@types/text-table": "0.2.0",
```
TooTallNate, line 86: Should we also remove the `text-table` dependency down below?

TooTallNate (commented):
If there is any visual difference with this change, can you post before/after screenshots for comparison?

packages/cli/package.json
onsclom: Good catch! Fixed now, thanks.

onsclom:
> If there is any visual difference with this change, can you post before/after screenshots for comparison?

No visual differences. All tables should look the same.

EndangeredMassa:
> No visual differences. All tables should look the same.


@onsclom can you post some before/after screenshots to show reviewers? If we had snapshot style tests around this, we could rely on those to be sure.

packages/cli/src/commands/teams/list.ts
```diff
@@ -76,14 +77,22 @@ export default async function list(client: Client): Promise<number> {
...
   );
+  client.stdout.write(
+    currentTeam
+      ? teamTable
+          .split('\n')
+          .map((line, i) => `${i > 0 ? teamList[i - 1].prefix : ' '} ${line}`)
+          .join('\n')
+      : teamTable
+  );
+  client.stdout.write('\n');
```
jeffsee55:
```suggestion
  client.stderr.write('\n');
```
I _think_ we pretty much always want to write to `stderr`

onsclom:
Here are some before/after pics that show the output does not change:

## Before
alias-ls-before
teams-ls-before

## After
alias-ls-after
teams-ls-after

`vc alias ls` is an example using the new unified table which all the other tables use. Most tables were a simple transition like `vc alias ls`. `vc teams ls` is a special case because that one was hand rolled and now uses that unified table.

TooTallNate (approved):

PR #11394, opened 2024-04-05
[cli] Improve type signature of CommandOption

erikareads:
The previous type signature of `CommandOption` did not cover all of the actual uses of option types throughout the codebase. This one may not cover all possible types, but comes much closer and allows us to literally state the type constructor used by `arg` to parse options.

This will allow for direct use of `command.ts` data structures so that Help output matches the actually parsed options used by every subcommand.

`deploy/index.ts` already shows an example of this, and a future pull request can build on that example by extracting that logic into a utility function.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -81,25 +81,13 @@ export default async (client: Client): Promise<number> => {
...
     if (option.shorthand) {
       argOptions[`-${option.shorthand}`] = `--${option.name}`;
     }
-    if (
-      option.name === 'env' ||
-      option.name === 'build-env' ||
-      option.name === 'meta'
-    ) {
-      argOptions[`--${option.name}`] = [String];
-    }
```
EndangeredMassa, line 102: **comment (non-blocking):** In future PRs, I would add a PR comment to code like this making it clear why this condition could be removed.

It took me a few minutes to figure out why this change was here and useful.‏

EndangeredMassa (approved):
Much nicer!

packages/cli/src/commands/deploy/index.ts
erikareads: I'll do that here by replying to your comment:

In this case, I can remove the extra condition since the `CommandOption` can now take constructors and `[Constructor]` directly as types, I don't need to rely on two sets of conditionals (the ternary above and this additional check) to get the parser type right.

Additionally, this was an interesting case, because these three commands specifically justified the `multi: true` in the `CommandOption` struct, but we can see here that we weren't using that information do the filtering and instead were hard-coding it.

The updated typing eliminates both issues.

PR #11396, opened 2024-04-05
[cli] Create argument parsing utility with clearer semantics

erikareads:
The way that we currently parse arguments throughout the vercel command line is through the `getArgs` function. This function takes three arguments: `argv`, `argOptions`, and `argsOptions`. (Those last two arguments are swapped in order from the real function). This is semantically muddy and likely to create mistakes. We have instances in our code base, such as `deploy/index.ts` which at the time of this pull request creates a variable named `argOptions` and passes it to the `argsOptions` argument of `getArgs`.

Further, the return type of `arg` the library we currently use to parse arguments is unclear as well. It puts positional arguments under an `_` key, instead of a meaningful name.

In this pull request, I create a function called `parseArguments` that takes the same three arguments as `getArgs` with improved names: `args`, `flags`, and `parserOptions`. It also changes the return type to be an object with two keys: `{args, flags}` allowing for clearer access patterns where the argument parser is used.


packages/cli/src/util/parse-arguments.ts
```diff
@@ -0,0 +1,29 @@
...
+  permissive?: boolean;
+};
+
+type Handler = (value: string) => any;
+
+interface Spec {
+  [key: string]: string | Handler | [Handler];
+}
+
+export default function parseArguments<T extends Spec>(
```
EndangeredMassa: **suggestion (blocking):** ‏make it clear this replaces `getArgs`

I think we should move this new function to live in `packages/cli/src/util/get-args.ts` alongside `getArgs`. I want to make it clear that the code is in the middle of a migration from `getArgs` to `parseArguments`.

I would:

- move `parseArguments` into `packages/cli/src/util/get-args.ts`
- add jsdoc comments to both functions
  - `parseArguments`: explain what it does, returns, and how it replaces `getArgs`
  - `getArgs`: mark as deprecated

EndangeredMassa (changes requested):

packages/cli/src/util/parse-arguments.ts
erikareads: I added a JSDoc comment, but I've never worked with JSDoc before, so please check it over.

packages/cli/src/util/get-args.ts
```diff
@@ -11,6 +11,9 @@ interface Spec {
   [key: string]: string | Handler | [Handler];
 }
 
+/**
+ * @deprecated
```
EndangeredMassa:
```suggestion
 * @deprecated use `parseArguments` instead
```

packages/cli/src/util/get-args.ts
```diff
@@ -21,3 +24,36 @@ export default function getArgs<T extends Spec>(
...
+ * Parses command line arguments.
+ * Automatically includes a number of common flags such as `--help`.
+ *
+ * This function is designed to replace getArgs
+ * and will live alongside getArgs until the migration is completed.
+ *
+ * It takes the same three arguments as getArgs with improved names: args, flagsSpecification, and parserOptions.
+ * It also changes the return type to be an object with two keys: {args, flags}
+ * `args` contains the positional arguments, previously returned under the key `_`
+ * `flags` contains the parsed options, previously these keys were mixed with the positional arguments
```
EndangeredMassa: You can use markdown in these comments. I spiced it up a bit in this example. This is what it looks like in VS Code:

Screenshot 2024-04-10 at 2 22 31 PM

```suggestion
 * Parses command line arguments.
 * Automatically includes a number of common flags such as `--help`.
 *
 * **Migrating from `getArgs`**
 *
 * This function is designed to replace `getArgs`
 * and will live alongside `getArgs` until the migration is completed.
 *
 * It takes the same three arguments as `getArgs` with improved names: `args`, `flagsSpecification`, and `parserOptions`.
 * It also changes the return type to be an object with two keys: `{args, flags}`
 *
 * - `args` was previously returned under the `_` key
 * - `flags` previously these keys were mixed with the positional arguments
```

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #11397, opened 2024-04-06
chore: fix some typos

findseat:


errors/invalid-route-source.md
```diff
@@ -50,4 +50,4 @@ Ensure any segments used in the `destination` property are also used in the `sou
 
 - [path-to-regexp](https://github.com/pillarjs/path-to-regexp/tree/v6.1.0)
 - [named parameters](https://github.com/pillarjs/path-to-regexp/blob/v6.1.0/Readme.md#named-parameters)
-- [un-named paramters](https://github.com/pillarjs/path-to-regexp/blob/v6.1.0/Readme.md#unnamed-parameters)
+- [un-named parameters](https://github.com/pillarjs/path-to-regexp/blob/v6.1.0/Readme.md#unnamed-parameters)
```
TooTallNate, line 53: This is the only valid fix in this PR.

The others are files from upstream framework templates, which we do not want to modify. Since they're all related to Hugo, I suggest you try to patch the fix in [its repository](https://github.com/gohugoio/hugo).

TooTallNate (dismissed):

EndangeredMassa (approved):

PR #11400, opened 2024-04-08
[cli] Fix command recommendation

onsclom:
`vc rm <project-name-that-doesn't-exist>` gives the message:
Screenshot 2024-04-08 at 2 38 03 PM

`vc ls` is not correct. `vc projects ls` is the correct command.


TooTallNate (dismissed):

.changeset/sixty-papayas-reply.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
TooTallNate: This should probably be a "patch" changeset.

jeffsee55 (approved):

EndangeredMassa (approved):

PR #11403, opened 2024-04-09
[tests] Fix tarball URL from deleted project

TooTallNate:
I did some pruning of old projects on my personal account last night, and accidentally broke CI. I've created a new tarball on the `curated-tests` team so that it doesn't happen again.


EndangeredMassa (approved):

EndangeredMassa:
added automerge because this is blocking CI on all PRs

PR #11409, opened 2024-04-09
Specify npm script to run during Find Changes step

jeffsee55:
Add a new vitest-unit test type and specify which npm script to run for a given chunks tests. Currently the CLI unit tests are actually running as part of the Find Changes action, because vitest ignores the `--listTests` flag and just runs the tests immediately.


utils/chunk-tests.js
```diff
@@ -36,13 +70,12 @@ function getRunnerOptions(scriptName, packageName) {
...
     );
   }
-  return (
-    runnerOptions || {
-      min: 1,
-      max: 1,
-      runners: ['ubuntu-latest'],
-    }
-  );
+  if (!runnerOptions) {
```
jeffsee55, line 73: I couldn't understand a scenario where we'd want a fallback value for this so got rid of it.

jeffsee55 (commented):
This is a bit awkward, for Jest tests, we rely on a single `test` script which is triggered for various chunks for `unit`, `dev` and `e2e` tests. But for the CLI package we want to run `vitest` for the unit tests only. So two separate scripts are needed. The one that actually lists the scripts (`vitest-unit`) is also used to display the tests in Github actions. The other (`vitest-unit-run`) is what is actually called 🙁 .

The `vitest-unit` script is using `jest` to actually list them, which isn't ideal, but Vitest doesn't have a list command just yet, though it looks like it [may come eventually](https://github.com/vitest-dev/vitest/pull/4240). 

The other is used to actually run the script. I'm not sure what flags we need for this, the `test` script has `--reporters=default --reporters=jest-junit --env node --verbose --bail` so might need some help deciphering what's needed for parity on Vitest. 

The irony here is that for unit tests we don't chunk these anyway, so some of the gymnastics here is only necessary because of the current chunking logic. As we move forward with Vitest we may want to consider only running the chunking logic when necessary (eg e2e tests). As it stands, this acts as a way to gradually move tests into Vitest without causing too much disruption.

EndangeredMassa (approved):

TooTallNate (approved):

PR #11415, opened 2024-04-10
Update packages causing vulnerability messages

onsclom:
A customer was worried about security vulnerabilities that `pnpm audit` listed for `@babel/traverse`, `http-cache-semantics`, and `json5`. This updates those deps, removing those vulnerability messages.


TooTallNate (approved):

EndangeredMassa (approved):

EndangeredMassa:
hrm, maybe the babel upgrade is breaking something?

PR #11433, opened 2024-04-12
Get flags specification

erikareads:
Extracting logic for turning a list of `CommandOption`s into a flagsSpecification for `parseArguments`. Builds on the work from #11396 and #11394.


packages/cli/test/unit/util/get-flags-specification.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        type: Number,
+        shorthand: null,
+        deprecated: false,
+      },
+    ];
+    expect(getFlagsSpecification(options)).toMatchObject({
+      '--an-option': Number,
+      '--another-option': Number,
+    });
+  });
```
EndangeredMassa, line 72: **suggestion (non-blocking):** ‏consider removing

I don't think these provide much value over the combo test below. Failure of the combo test will still be very clear about which type isn't working. Do you think we need these tests?

I don't mind keeping them. They're very fast.

EndangeredMassa (approved):

packages/cli/test/unit/util/get-flags-specification.test.ts
erikareads: I chose to enumerate them specifically because I want to make sure that we continue to support these types in the event that we change argument parsers in the future.

I think the mixed test muddies this intent. It does cover an important case, but doesn't clearly say things about the types that individual tests do better.

EndangeredMassa: I don't understand how the individual tests say something better. Are you saying that having the separate test at all clarifies the intent to support that type?

erikareads: Yes, that was my intention.

EndangeredMassa: I don't agree that this overly conveys that over the combo test, but I don't mind keeping them all.

PR #11439, opened 2024-04-15
Add snapshot tests for help

erikareads:
Adds snapshot tests for the output from help as output by various commands. I expect to be changing some of the `Command` structs to continue the work started in [ [cli] Improve type signature of CommandOption #11394 ](https://github.com/vercel/vercel/pull/11394) especially as I build on work in [ Get flags specification #11433 ](https://github.com/vercel/vercel/pull/11433), and I want to be sure that the associated changes to help output are what we expect.


trek (dismissed):
Not blocking but I'd love to discuss wider on the team. I _personally_ dislike looped tests either with `for...in` or `test.each` because it makes it annoying to focus a single tests in the future. But, I don't care enough to block.

erikareads:
@trek I was copying an existing test as a template, I'm kinda meh about it in particular. I think now *is* the time to think about how we would like to test this instead.

EndangeredMassa (dismissed):
I'd also like to break these out, but happy to merge either way

EndangeredMassa (approved):

PR #11442, opened 2024-04-16
[next] Remove un-necessary compressed function size calc/constraint

ijjk:
Historically, we used to have a compressed function size limit of `50MB` but this is a very old limit and was removed long ago. Considering this limit is removed and it requires additional overhead to calculate, this removes the compressed size limit handling while leaving the uncompressed size limit detection/handling.


Ethan-Arrowood (dismissed):

packages/next/test/integration/test-limit-exceeded-404-static-files/next.config.js
```diff
@@ -23,7 +23,7 @@ assert(
 // generate 100MB text file which will be traced in `/api/hello`
```
styfle: Should this comment say 200MB now?

ijjk: Yea will make this comment generic though since arbitrary size

TooTallNate (approved):

styfle (approved):

ztanner (approved):

PR #11445, opened 2024-04-16
Update pnpm version detection logic

erikareads:
The code we use to detect the version of `pnpm` based on lockfiles is hard to follow. It doesn't pick and executable, it instead of *sometimes* overrides the `PATH` for these utilities with a prepended alternate path. 

This means that there is no one single place where we definitively specify which version of `pnpm` that is used, so much as *hope* that the `PATH` has the correct version on it. I don't know that we *can* specify an executable outside of the build container, and it's unclear, since this is a package, whether this utility is used outside of a build container setting.

In this pull request I cleaned up the code and updated the logic to include the `pnpm 7` and `pnpm 8` changes requested, with no changes yet to implement `pnpm 9`.

It's possible that we want to rewrite this detection logic from scratch to avoid this roundabout way of "finding" the executable, but I haven't done that here.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -558,36 +557,76 @@ export function getEnvForPackageManager({
     ...env,
   };
 
-  if (newPath) {
+  const alreadyInPath = (path: string) => {
+    const oldPath = env.PATH ?? '';
+    return oldPath.includes(path);
```
TooTallNate: We should probably split the PATH into an array, otherwise this function might have a false positive where a substring is matched (contrived example: `alreadyInPath('/usr')` where `PATH` is something like `/usr/local/bin:/bin`):

```suggestion
    return oldPath.split(path.delimiter).includes(path);
```

EndangeredMassa (commented):
At a single read-through, this looks good to me. I can't dive deeper in review right now, though. We'll want to manually test many scenarios in the build container before rolling this change there, for sure.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -622,70 +757,31 @@ export function getPathForPackageManager({
...
+    case alreadyInPath(overrides.path ?? ''):
+      return {
+        detectedLockfile: undefined,
+        detectedPackageManager: undefined,
+        path: undefined,
+        yarnNodeLinker: undefined,
+      };
+    default:
+      return { ...overrides, yarnNodeLinker: undefined };
+  }
```
EndangeredMassa, line 784: **suggestion (non-blocking):** ‏ifs over switches?

The switch-true pattern is harder to understand than the equivalent if-return or if-else, in my opinion. I suggest changing this, but it's up to you.

EndangeredMassa (approved):

PR #11450, opened 2024-04-17
[cli] Don't log error when running vc dev in a directory without a .git directory

onsclom:
`vc dev` in a dir without a `.git` dir would display an error the user should not see:

```
Vercel CLI 34.0.0
fatal: not a git repository (or any of the parent directories): .git
> Ready! Available at [http://localhost:3000](http://localhost:3000)
```

This fixes that by ignoring the stderr from `git rev-parse`.


packages/cli/src/util/git-helpers.ts
```diff
@@ -37,6 +37,7 @@ export function getGitDirectory(opts: GitExecOptions): string | null {
     const gitConfigPath = execSync('git rev-parse --git-dir', {
       cwd,
       encoding: 'utf8',
+      stdio: ['pipe', 'pipe', 'ignore'],
```
TooTallNate: stdin can be ignored as well I think

```suggestion
      stdio: ['ignore', 'pipe', 'ignore'],
```

TooTallNate (approved):

trek (approved):

PR #11451, opened 2024-04-17
[cli] Sort default team to the top of the team selector

elliott-with-the-longest-name-on-github:
The user's default team should be sorted to the top of the selector. This is a nice and simple way to do this, though I'm not sure how to test my changes -- would appreciate guidance!


EndangeredMassa:
The redwood failure specifically was a break elsewhere that has now been fixed on `main`. Merging `main` now.

packages/cli/test/mocks/team.ts
```diff
@@ -1,6 +1,17 @@
...
+type Team = {
+  id: string;
+  slug: string;
+  name: string;
+  creatorId: string;
+  created: string;
+  avatar: null;
+};
+
+let teams: Team[] = [];
```
EndangeredMassa, line 13: The mock here wasn't set up to allow for multiple teams. I adjusted it for the use of the current tests, but we should have a better mocking strategy in general for these kinds resources.

EndangeredMassa:
@tcc-sejohnson I added a test for this. The test setup wasn't going to make this easy for you, but I've tweaked it to hopefully make it easier next time.

EndangeredMassa (approved):

trek (approved):

packages/cli/src/util/input/select-org.ts
```diff
@@ -36,10 +36,12 @@ export default async function selectOrg(
 
   const choices: Choice[] = [
     ...personalAccountChoice,
-    ...teams.map<Choice>(team => ({
-      name: team.name || team.slug,
-      value: { type: 'team', id: team.id, slug: team.slug },
-    })),
+    ...teams
+      .sort(a => (a.id === user.defaultTeamId ? -1 : 1))
```
TooTallNate, line 40: This sorting function is a bit strange to me (I've never seen a comparator which does not operate on the `b` parameter). Also, it's not clear to me what the fallback behavior is (for the non-`defaultTeamId` case), but we should probably make an explicit decision about that (i.e. sort by `slug`).

elliott-with-the-longest-name-on-github: Yeah, unfortunately all of the ways to say "sort a specific item to the top of the array but leave everything else untouched" in JavaScript are weird, with this being the least weird. I did test it on every case (item at start, item at end, item inbetween) and it works fine -- `-1`s get sorted earlier, `1`s get sorted later, equal neighbors get left untouched. 

As for fallback behavior, there is none today and there also is none after this PR merges -- we just display your teams in the order they come back from `getTeams(client)`.

PR #11452, opened 2024-04-17
Add eslint rule for no-console within the cli package

jeffsee55:
This PR adds a rule to disallow the use of console.log/console.error/etc from within the CLI package. The aim is to centralize our use of stdio within the CLI so that everything moves through our client's output module. It also disables the rule for all of the current console usage, with the hopes that we will clean things up soon™

Also want to note that the rule only applies to usage from within the CLI, so dependencies that the CLI pulls in (both external and even within this monorepo) are unaffected.


packages/cli/test/dev/integration-1.test.ts
```diff
@@ -967,6 +967,7 @@ test(
   testFixtureStdio('handle-miss-querystring', async (testPath: any) => {
     await testPath(200, '/', 'Index Page');
     if (process.env.CI && process.platform === 'darwin') {
+      // eslint-disable-next-line no-console
```
TooTallNate, line 970: Is there a way to disable this rule for tests?

TooTallNate (approved):

packages/cli/test/dev/integration-1.test.ts
jeffsee55: @erikareads and I discussed this briefly but thought maybe we also want to clear some of that out. I think our options here are either:

- Ignore it at the top of the file, but that might be annoying to do
- Put tests in the `.eslintignore`, but that would ignore all eslint rules

EDIT: 3rd option would be to have a logger utility that we use for this scenario

trek (approved):

EndangeredMassa (approved):

jeffsee55:
It seems like the PR busted a turbo cache for a test that didn't change. The same tests are failing on `main` when run locally. 👀 

Getting skipped on the most recent `main` build:
https://github.com/vercel/vercel/actions/runs/8741230148/job/23986856379

EDIT: Test passes now after update from `main`.

PR #11453, opened 2024-04-17
[next] support sectioned source maps

wbinnssmith:
Turbopack outputs sectioned source maps, but webpack’s `webpack-sources` packages doesn’t support these. For performance reasons, it should, but as an interim solution, flatten these sections maps for compatibility.


package.json
```diff
@@ -36,6 +36,11 @@
     "turbo": "1.13.2",
     "typescript": "4.9.5"
   },
+  "pnpm": {
+    "overrides": {
+      "source-map": "0.7.4"
```
TooTallNate: Can you explain why this is necessary?

.changeset/smart-beds-sit.md
```diff
@@ -0,0 +1,8 @@
+---
+'@vercel/redwood': patch
+'@vercel/next': patch
+'@vercel/node': patch
+'api': patch
```
TooTallNate: The `api` package does not need to be versioned.

```suggestion
```

.changeset/smart-beds-sit.md
```diff
@@ -0,0 +1,8 @@
+---
+'@vercel/redwood': patch
+'@vercel/next': patch
+'@vercel/node': patch
```
TooTallNate: Change seems unrelated to redwood / node?

```suggestion
'@vercel/next': patch
```

packages/next/build.mjs
```diff
@@ -18,6 +23,11 @@ await Promise.all([
     ],
   }),
   buildEdgeFunctionTemplate(),
+  // The bundled version of source-map looks for the wasm mappings file as a sibling.
+  copyFile(
+    fileURLToPath(import.meta.resolve('source-map/lib/mappings.wasm')),
```
TooTallNate: `copyFile()` works with URLs:

```suggestion
   import.meta.resolve('source-map/lib/mappings.wasm'),
```

package.json
wbinnssmith: It's not. reverted.

packages/next/build.mjs
wbinnssmith: It didn't for me. I got an `ENOENT` on the source file.

TooTallNate (approved):

PR #11455, opened 2024-04-17
[static-build] Don't rely on `hugo` binary existing in build-container

TooTallNate:
* Makes `hugo` framework preset work when Node.js v20.x is selected in Project Settings.
* Stores the downloaded `hugo` binary in the `.vercel/cache` directory, so that it does not need to be re-downloaded upon every deployment.
* Makes `vc build` work when run locally for Hugo projects - tested on macOS arm64.


packages/static-build/src/index.ts
```diff
@@ -351,13 +364,19 @@ export const build: BuildV2 = async ({
...
 
-      if (HUGO_VERSION && !meta.isDev) {
-        console.log('Installing Hugo version ' + HUGO_VERSION);
-        const [major, minor] = HUGO_VERSION.split('.').map(Number);
-        const isOldVersion = major === 0 && minor < 43;
-        const prefix = isOldVersion ? `hugo_` : `hugo_extended_`;
-        const url = `https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/${prefix}${HUGO_VERSION}_Linux-64bit.tar.gz`;
-        await fetchBinary(url, 'Hugo', HUGO_VERSION);
+      if ((HUGO_VERSION || framework?.slug === 'hugo') && !meta.isDev) {
+        const hugoVersion = HUGO_VERSION || '0.58.2';
```
EndangeredMassa, line 369: **question (blocking):** ‏what version should we use here?

The original was assuming 0.56.0. The latest appears to be 0.125.0. Why are we using 0.58.2 here?

packages/static-build/src/utils/hugo.ts
```diff
@@ -0,0 +1,84 @@
...
+  return `https://github.com/gohugoio/hugo/releases/download/v${version}/${file}`;
+}
+
+function findFile(
+  names: Iterable<string>,
+  oses: string[],
+  arches: string[],
+  extended: boolean
+): string | null {
+  const prefix = `hugo_${extended ? 'extended_' : ''}`;
```
EndangeredMassa, line 70: **question (non-blocking):** ‏what does `extended_` mean?

EndangeredMassa (changes requested):

packages/static-build/src/index.ts
TooTallNate: That's the version in the AL2 build-container currently. afaik we don't really have a policy on when this gets updated. I do know that 0.125.0 does not work with our current hugo template, so seems they include breaking changes in the `0.x.x` versions.

IMO we should leave this default as-is, and set `HUGO_VERSION` to latest in the template, while also updating the template such that it works with 0.125.0.

packages/static-build/src/utils/hugo.ts
TooTallNate: https://gohugo.io/installation/macos/#editions

TooTallNate: For as long as the extended version has been available, that's the version that gets installed when `HUGO_VERSION` is set. The extended version is also the one that's available in build-container by default.

EndangeredMassa (dismissed):

PR #11467, opened 2024-04-18
[cli] Migrate alias command to use parseArguments and getFlagsSpecification

erikareads:
Migrating `vc alias` to use new argument parsing utilities.

Note that our snapshot tests show that our help text was out of sync with what we were actually parsing from the user. 

In particular, we were showing `-n` as the shorthand for both `next` and `limit` and neither of these had this shortflag. `next` was parsing `-N` (capital), so I've updated it to that.

`--json` was parsed, but wasn't mentioned in the help text. I've kept it this way, but it may be worth adding a description. We may want to have a separate discussion about the nature and role of `json` output from our CLI as a whole.


packages/cli/src/commands/alias/command.ts
```diff
@@ -52,7 +52,7 @@ export const aliasCommand: Command = {
       name: 'next',
       description: 'Show next page of results',
       argument: 'MS',
-      shorthand: 'n',
+      shorthand: 'N',
```
trek, line 55: Was this just wrong previously?

trek (approved):

packages/cli/src/commands/alias/command.ts
erikareads: yes

trek (approved):

PR #11469, opened 2024-04-19
[next] reduce `.action` rewrites

ztanner:
- copies existing .rsc handling to ignore `/_next/data` routes since those aren't relevant in app dir
- only push dynamic route rewrites if on a supported version & the route exists in the prerender manifest (need to verify that `blockingFallbackRoutes` is the correct place to check)

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -315,8 +315,10 @@ export async function getDynamicRoutes({
   dynamicPages,
   isDev,
   routesManifest,
+  prerenderManifest,
   omittedRoutes,
   canUsePreviewMode,
+  canUseActionRewrite,
   bypassToken,
   isServerMode,
   dynamicMiddlewareRouteMap,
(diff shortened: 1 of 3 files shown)


packages/next/src/utils.ts
```diff
@@ -432,14 +436,21 @@ export async function getDynamicRoutes({
...
-              new RegExp(escapeStringRegexp('(?:/)?$')),
-              '(?:\\.action)(?:/)?$'
-            ),
-            dest: route.dest?.replace(/($|\?)/, '.action$1'),
-          });
+          // push .action rewrites for dynamic routes if support is available
+          // and the dynamic route is in the prerender manifest
+          if (
+            canUseActionRewrite &&
+            prerenderManifest.blockingFallbackRoutes[page]
```
ztanner: cc @ijjk is this a good way to scope down dynamic rewrites? Validated that the existing dynamic force-static test passed but I'm unfamiliar with what all this `blockingFallbackRoutes` property is used for.

PR #11478, opened 2024-04-23
[python] Add `python3.12`

TooTallNate:
Allows for `python3.12` to be used when the build is being run in the AL2023 build container (Node v20 project setting). This version is not usable in the AL2 image because `python3.12` is not installed there.

The e2e tests now default to running in AL2023, except for a few fixtures which require an older version of Python.


packages/python/test/integration-setup.js
```diff
@@ -28,19 +28,12 @@ module.exports = function setupTests(groupIndex) {
...
-      nodeVersion: '18.x',
-    };
-
     const errMsg = testsThatFailToBuild.get(fixture);
     if (errMsg) {
       // eslint-disable-next-line no-loop-func
       it(`should fail to build ${fixture}`, async () => {
         try {
-          await testDeployment(path.join(fixturesPath, fixture), {
-            projectSettings,
```
erikareads, line 42: What does removing `projectSettings` do?

erikareads: Or rather, why are we removing it?

EndangeredMassa: It was here before to force all projects to be at node version 18. In this PR, Nate changed the default to be whatever the platform default is, which means we don't need to specify the project setting here.

The platform default is node Node 20.

EndangeredMassa (approved):

packages/python/test/integration-setup.js
TooTallNate: We were forcing the AL2 build image previously, since Python was not yet supported. Removing project settings here will use the default build image, which is now the AL2023 image.

PR #11482, opened 2024-04-23
[next] add tests for server actions

ztanner:
There aren't currently any tests for server actions which have special handling in the builder. This adds tests to help avoid regressions in future Next.js builder work.


.changeset/smooth-moles-lay.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
ijjk: I think we should be able to skip the changeset if we add the tests only label

packages/next/test/fixtures/00-app-dir-actions/app/client/static/[dynamic-static]/page.js
```diff
@@ -0,0 +1,21 @@
+'use client';
+
+import { useState } from 'react';
+import { increment } from '../../../actions';
+
+export const dynamic = 'force-static';
```
Comment (ijjk) [inline packages/next/test/fixtures/00-app-dir-actions/app/client/static/[dynamic-static]/page.js]: This has no effect in `use client` is this being statically generated like we're hoping?

packages/next/test/fixtures/00-app-dir-actions/app/client/dynamic/[id]/page.js
```diff
@@ -0,0 +1,21 @@
+'use client';
+
+import { useState } from 'react';
+import { increment } from '../../../actions';
+
+export const dynamic = 'force-dynamic';
```
Comment (ijjk) [inline packages/next/test/fixtures/00-app-dir-actions/app/client/dynamic/[id]/page.js]: Similar to https://github.com/vercel/vercel/pull/11482/files#r1577054712

packages/next/test/fixtures/00-app-dir-actions/app/rsc/static/[dynamic-static]/page.js
```diff
@@ -0,0 +1,17 @@
+import { revalidatePath } from 'next/cache'
+
+export const dynamic = 'force-static'
```
Comment (ijjk) [inline packages/next/test/fixtures/00-app-dir-actions/app/rsc/static/[dynamic-static]/page.js]: Should we add `generateStaticParams` and test for some pre-generated paths as well as lazily generated?

.changeset/smooth-moles-lay.md
ztanner: Looks like this is required

ijjk (approved):
Looks good, thanks!

PR #11493, opened 2024-04-24
Migrate certs argument parsing

erikareads:
Migrates `vc certs` to use updated argument parsing utilities.

Fixes an incorrect shorthand flag for `vc certs --next` -> `-N`.

I've also added `--overwrite` and `--output` to the `Command` struct. However, I defaulted to not showing any help text for these flags.


packages/cli/src/commands/certs/command.ts
```diff
@@ -82,16 +82,18 @@ export const certsCommand: Command = {
...
       name: 'next',
       description: 'Show next page of results',
-      shorthand: 'n',
-      type: String,
+      shorthand: 'N',
+      type: Number,
       deprecated: false,
     },
+    { name: 'overwrite', shorthand: null, type: Boolean, deprecated: false },
+    { name: 'output', shorthand: null, type: String, deprecated: false },
```
erikareads, line 96: Do we want to add a description and show help text for these flags? We've been parsing them but not showing them in the help until now.

PR #11494, opened 2024-04-25
Update SvelteKit v1 devCommand

amyegan:
The current development command for the SvelteKit (v1) preset causes CORS errors in some cases. Overriding this command to exclude `--port $PORT` resolves the issue.

See https://github.com/orgs/vercel/discussions/3390 for background info


trek:
@amyegan do you know if this _used_ to work in older Svelte versions? If so one risk of changing this is if someone deploys an older version of `SvelteKit` it would break for them.

amyegan:
> @amyegan do you know if this _used_ to work in older Svelte versions? If so one risk of changing this is if someone deploys an older version of `SvelteKit` it would break for them.

@trek Good question. I know that the current default dev command has been causing problems for the community since 26 July 2023 (around version `1.20.5`), and possibly earlier than that. 

I tried just now with version `1.0.0` which also seems to throw CORS errors on local with the default dev command and seems to work as desired with any of the overrides I have found to work with the latest version. My best guess is that it never worked as expected for v1 in certain situations.

For recent versions my workaround is to override the default dev command with `vite`, `vite --port 3000`, or `vite --cors false`. Certainly open to other solutions. Just went for the simplest one I've found 😅

EndangeredMassa (approved):

PR #11497, opened 2024-04-25
Add support for ruby 3.3

jeffsee55:
Add support for ruby3.3 when using AL2023 (Node v20). No support for ruby3.3 on Node <20 planned.

Ruby integration tests will now running on AL2023


EndangeredMassa (dismissed):

janory (commented):
Please don't forget about the [SUPPORTED_RUBY_VERSION](https://github.com/vercel/vercel/blob/26bc51e0d641ea45d304b601848b8123feeade48/packages/static-build/src/index.ts#L51) constant, otherwise the install command won't work properly.

packages/static-build/src/index.ts
```diff
@@ -48,7 +48,7 @@ import {
 import { getHugoUrl } from './utils/hugo';
 import { once } from 'events';
 
-const SUPPORTED_RUBY_VERSION = '3.2.0';
+const SUPPORTED_RUBY_VERSION = '3.3.0';
```
EndangeredMassa, line 51: I didn't dig into this, but we probably need this value to change based on the build image (node 20 or not) we're using.

Ref: https://github.com/vercel/vercel/blob/add-ruby33-to-builder/packages/static-build/src/index.ts#L582

TooTallNate: It's not really critical. This value is used to determine the vendor path where the Gem deps will be installed to. If anything, we should revisit this to install into `.vercel/cache` instead.

TooTallNate (approved):

PR #11504, opened 2024-04-27
[next] Only rewrite next-action requests to `.action` handlers

ztanner:
This only rewrites requests that contain a `next-action` header (explicitly indicating it's a server action). A side effect is that POST requests to a server action on a static route, without a next-action header, won't be marked as streaming (but will still execute normally). This is fine as only the handled action needs to stream. 

This relands .action handling behind a feature flag.


ijjk (dismissed):

ijjk (dismissed):

.changeset/big-turkeys-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/next": patch
+"api": patch
```
TooTallNate:
```suggestion
```

.changeset/big-turkeys-smoke.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+[next] Only rewrite next-action requests to `.action` handlers
```
TooTallNate:
```suggestion
Only rewrite next-action requests to `.action` handlers
```

ijjk (approved):

PR #11514, opened 2024-04-30
[next] Update test fixture version

ijjk:
Noticed in https://github.com/vercel/vercel/pull/11503 that this test fixture is incorrectly locked to a specific version when we don't want it to be.


ztanner (approved):

.changeset/gentle-sloths-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/next': patch
```
trek, line 2: Not blocking, but this is probably doesn't require even a patch version bump

trek (approved):

.changeset/gentle-sloths-rescue.md
ijjk: Did we remove the label that tells the changeset to ignore test only changes? We tried on a previous PR and it was blocking merge

trek: That isn't a thing I've ever heard of! I've always seen them required in my time here but running changeset with `--empty` will generate an empty one.

trek: Or just manually delete the content between the `---`s

TooTallNate: You can push an empty changeset to make the bot happy, but anyways, the changeset is not a required check, so it's not a requirement.

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

PR #11519, opened 2024-04-30
[cli]: wait waitUntil promises in Serverless

Kikobeats:
The `waitUntil` in serverless is not waited before closing the dev server.

This wrong behavior is just happening for without `'edge'` because under `'edge'` uses `runServer` helpers from `edge-runtime` which already is taking care of promises to wait before closing the server:

- https://github.com/vercel/edge-runtime/blob/352af9028a260b6ec61dda721aa384b19f56c7b4/packages/runtime/src/server/run-server.ts#L34
- https://github.com/vercel/edge-runtime/blob/352af9028a260b6ec61dda721aa384b19f56c7b4/packages/vm/src/edge-vm.ts#L248

I fixed the problem by just tracking promises to be waited on from userspace, making it possible to wait on them before closing the server.

**TODO**

- [x] add a test
- [x] show a warning if waitUntil is taking too much time (https://github.com/vercel/vercel/blob/d568e3466e57e0a80652d9d5c09fbf94f1f28e75/packages/node/src/edge-functions/edge-handler.mts#L176)


packages/node/src/utils.ts
```diff
@@ -3,6 +3,22 @@ import { pathToRegexp } from 'path-to-regexp';
...
 
+// When exiting this process, wait for Vercel Function server to finish
+// all its work, especially waitUntil promises before exiting this process.
+//
+// Here we use a short timeout (10 seconds) to let the user know that
+// it has a long-running waitUntil promise.
+const WAIT_UNTIL_TIMEOUT = 10;
+export const WAIT_UNTIL_TIMEOUT_MS = 10 * 1000;
+
+export const WAIT_UNTIL_WARNING = (entrypointPath: string) =>
```
```suggestion
export const waitUntilWarning = (entrypointPath: string) =>
```

Lets follow the naming convention for functions

packages/node/test/dev-fixtures/edge-waituntil.js
styfle, line 1: Why is this file deleted?

packages/node/test/dev-fixtures/serverless-fetch.js
styfle, line 1: Also, why is this file deleted?

packages/node/test/dev-fixtures/edge-waituntil.js
Kikobeats: there is a new file for testing waitUntil, also another one for Node.js runtime

packages/node/test/dev-fixtures/serverless-fetch.js
Kikobeats: Not used

packages/node/test/dev-fixtures/web-handlers-node-wait-until.js
styfle, line 1: Is this file used? I'm not seeing it referenced in `packages/node/test/unit/dev.test.ts` 🤔

Kikobeats: good catch; removed

styfle (approved):
LGTM

You should probably get a codeowner's approval as well

packages/node/src/serverless-functions/helpers-web.ts
```diff
@@ -3,42 +3,46 @@ import type { NodeHandler } from '@edge-runtime/node-utils';
...
-      constructor(input: RequestInfo | URL, init?: RequestInit | undefined) {
-        super(input, addDuplexToInit(init));
-      }
+export const createWebExportsHandler = (FetchEvent: typeof Edge.FetchEvent) => {
+  const webHandlerToNodeHandler = buildToNodeHandler(
+    {
+      Headers,
+      ReadableStream,
+      // @ts-expect-error Property 'duplex' is missing in type 'Request'
+      Request: class extends Request {
```
Ethan-Arrowood, line 12: I've never seen a class defined inline like this before. Maybe we can do this outside of the function call and then pass it in like normal?

packages/node/src/utils.ts
```diff
@@ -3,6 +3,22 @@ import { pathToRegexp } from 'path-to-regexp';
...
+//
+// Here we use a short timeout (10 seconds) to let the user know that
+// it has a long-running waitUntil promise.
+const WAIT_UNTIL_TIMEOUT = 10;
+export const WAIT_UNTIL_TIMEOUT_MS = 10 * 1000;
+
+export const waitUntilWarning = (entrypointPath: string) =>
+  `
+The function \`${entrypointPath
+    .split('/')
```
Ethan-Arrowood, line 17: is the path guaranteed to use `'/'`? Or should this be https://nodejs.org/api/path.html#pathsep

packages/node/src/serverless-functions/helpers-web.ts
Kikobeats: it isn't part of the scope of this PR

PR #11526, opened 2024-05-01
[cli] Don't pull system environment vars in dev

onsclom:
System environment variables would get set with empty strings in development which breaks some builds. This fixes that by using the v2 of `/env/pull` introduced in https://github.com/vercel/api/pull/27777.


packages/cli/test/mocks/project.ts
```diff
@@ -390,7 +390,7 @@ function exposeSystemEnvs(
 ) {
   const envs: Env = {};
 
-  if (autoExposeSystemEnvs) {
+  if (autoExposeSystemEnvs && target !== 'development') {
     envs['VERCEL'] = '1';
     envs['VERCEL_ENV'] = target || 'development';
```
EndangeredMassa, line 395: **question (soft-blocking):** If `target` is not development, do we still want to default `VERCEL_ENV` to "development" here?

packages/cli/test/unit/commands/env.test.ts
```diff
@@ -205,10 +205,17 @@ describe('env', () => {
...
+        'env',
+        'pull',
+        'other.env',
+        '--yes',
+        '--environment',
+        'production'
+      );
       const exitCodePromise = env(client);
       await expect(client.stderr).toOutput(
-        'Downloading `development` Environment Variables for Project vercel-env-pull'
```
EndangeredMassa, line 211: **suggestion (soft-blocking):** Add assertion (or new test) somewhere that development env pulls do not include system env vars at all.

EndangeredMassa (dismissed):

packages/cli/test/mocks/project.ts
onsclom: Since this is a mock, to match the api behavior it should not provide a default `VERCEL_ENV` for dev. I made that decision in https://github.com/vercel/api/pull/27777 because it feels more consistent to just not send any system env vars in dev. But, I'm open to changing the api behavior to make `VERCEL_ENV` and maybe `VERCEL` exceptions.

trek (approved):

PR #11527, opened 2024-05-01
[next] Fix interception routes PPR route handling

ijjk:
This ensures we properly create `.prefetch.rsc` output aliases for non-prerender outputs when PPR is enabled so that we can properly route which handles the interception route case as that is always dynamic so was missing it's `.prefetch.rsc` output causing the prefetch request to 404 unexpectedly. 

x-ref: [slack thread](https://vercel.slack.com/archives/C070ZMMKVU7/p1714511126337289)

### DIFF
--- a/packages/next/test/fixtures/00-app-dir-ppr-full/app/page.jsx
+++ b/packages/next/test/fixtures/00-app-dir-ppr-full/app/page.jsx
@@ -1,10 +1,14 @@
+import Link from 'next/link'
 import React, { Suspense } from 'react'
 import { Dynamic } from '../components/dynamic'
 
 export default () => {
   return (
-    <Suspense fallback={<Dynamic pathname="/" fallback />}>
-      <Dynamic pathname="/" />
-    </Suspense>
+    <>
+      <Suspense fallback={<Dynamic pathname="/" fallback />}>
+        <Dynamic pathname="/" />
+      </Suspense>
+      <Link href='/cart'>to /cart</Link>
+    </>
   )
 }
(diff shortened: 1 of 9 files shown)


packages/next/test/fixtures/00-app-dir-ppr-full/index.test.js
```diff
@@ -45,6 +45,45 @@ describe(`${__dirname.split(path.sep).pop()}`, () => {
   });
 
   describe('dynamic pages should resume', () => {
+    it('should handle interception route properly', async () => {
```
ztanner: maybe should hoist out of the dynamic resume block for clarity

packages/next/test/fixtures/00-app-dir-ppr-full/app/page.jsx
```diff
@@ -1,10 +1,14 @@
...
 export default () => {
   return (
-    <Suspense fallback={<Dynamic pathname="/" fallback />}>
-      <Dynamic pathname="/" />
-    </Suspense>
+    <>
+      <Suspense fallback={<Dynamic pathname="/" fallback />}>
+        <Dynamic pathname="/" />
+      </Suspense>
+      <Link href='/cart'>to /cart</Link>
```
ztanner, line 11: might be worth running prettier on this app

ijjk: We might wanna do that in separate PR to avoid extra changes in this one since we ignore fixtures in prettier currently.

ztanner (approved):

PR #11531, opened 2024-05-02
ci: uses node18

Kikobeats:
Node.js v16 is pretty old, long live to Node.js 18!


EndangeredMassa (changes requested):
We still officially support Node 16 and want to make sure we're testing against it. We can also test against Node 18 specifically in parallel, though.

PR #11540, opened 2024-05-03
[next] Remove .prefetch.rsc rewrites for non-PPR

ijjk:
Removes .prefetch.rsc rewrites rules for non-PPR, as Next.js doesn't emit any static prefetches otherwise


ztanner (approved):

ijjk (commented):
🚢

PR #11541, opened 2024-05-04
bugfix: ASGI support for newer python versions > 3.9

### LINKED ISSUE #11545: Python runtime does not work on Python 3.12 with ASGI
Although Python 3.12 is listed as supported by Vercel, it does not work properly with asynchronous functions.

I am trying to deploy a Starlette app (which is ASGI). Just listing Starlette as dependency did not work, as Vercel complained that it's missing `aiohttp`:

```
LAMBDA_WARNING: Unhandled exception. [...]
[ERROR] Runtime.ImportModuleError: Unable to import module 'vc__handler__python': No module named 'aiohttp'
```

So, I've added it to the dependencies (although I don't see the point in it). Nevertheless, it now fails with a different error:

```
LAMBDA_WARNING: Unhandled exception. [...]
[ERROR] TypeError: Queue.__init__() got an unexpected keyword argument 'loop'
  File "/var/task/vc__handler__python.py", line 305, in vc_handler
    response = asgi_cycle(__vc_module.app, body)
  File "/var/task/vc__handler__python.py", line 202, in __call__
    self.app_queue = asyncio.Queue(loop=loop)
```

I could trace back to this line of `vc_init`:

https://github.com/vercel/vercel/blob/3101d2497f79fd39e927dd9527d0bee800445d28/packages/python/vc_init.py#L202

[According to the Python docs](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue), the `loop` argument to the `Queue` was removed in Python 3.10

winglian:
using any version of python > 3.9 results in `<class 'TypeError'>` when trying to launch an application like FastAPI

edit: closes #11545

### DIFF
--- a/packages/python/vc_init.py
+++ b/packages/python/vc_init.py
@@ -175,7 +175,6 @@ elif 'app' in __vc_variables:
         # https://github.com/erm/mangum/blob/b4d21c8f5e304a3e17b88bc9fa345106acc50ad7/LICENSE
         import asyncio
         import enum
-        import inspect
         from urllib.parse import urlparse
         from werkzeug.datastructures import Headers
 
@@ -192,7 +191,7 @@ elif 'app' in __vc_variables:
                 self.state = ASGICycleState.REQUEST
                 self.app_queue = None
                 self.response = {}
-                self._legacy_asyncio = "loop" in inspect.signature(asyncio.Queue).parameters
+                self._legacy_asyncio = sys.version_info < (3, 10)
 
             def __call__(self, app, body):
                 """
--- a/packages/python/vc_init.py
+++ b/packages/python/vc_init.py
@@ -175,6 +175,7 @@ elif 'app' in __vc_variables:
         # https://github.com/erm/mangum/blob/b4d21c8f5e304a3e17b88bc9fa345106acc50ad7/LICENSE
         import asyncio
         import enum
+        import inspect
         from urllib.parse import urlparse
         from werkzeug.datastructures import Headers
 
@@ -191,6 +192,7 @@ elif 'app' in __vc_variables:
                 self.state = ASGICycleState.REQUEST
                 self.app_queue = None
                 self.response = {}
+                self._legacy_asyncio = "loop" in inspect.signature(asyncio.Queue).parameters
 
             def __call__(self, app, body):
                 """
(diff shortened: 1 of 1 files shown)


packages/python/vc_init.py
```diff
@@ -191,23 +192,33 @@ def __init__(self, scope):
                 self.state = ASGICycleState.REQUEST
                 self.app_queue = None
                 self.response = {}
+                self._legacy_asyncio = "loop" in inspect.signature(asyncio.Queue).parameters
```
kytta: I am not a fan of using `inspect` here. We know for a fact that, for example, `loop` was removed in Python 3.10. So, instead of this field, I'd compare `sys.version_info >= (3, 10)`

kytta (changes requested):
**NB:** I am not a maintainer

Quite a good solution, although I personally would do this differently. Either way, this would fix my issue (see #11545); it would be great, if you could mention it with the "closes" keyword :)

winglian:
> if you could mention it with the "closes" keyword :)

Is there a particular open issue that this closes?

packages/python/vc_init.py
winglian: thanks, I've updated the check.

packages/python/vc_init.py
```diff
@@ -191,23 +191,33 @@ def __init__(self, scope):
...
-                asgi_task = loop.create_task(asgi_instance)
-                loop.run_until_complete(asgi_task)
+                if self._legacy_asyncio:
+                    asgi_task = loop.create_task(asgi_instance)
+                    loop.run_until_complete(asgi_task)
+                else:
+                    asyncio.run(self.run_asgi_instance(asgi_instance))
                 return self.response
 
+            async def run_asgi(self, asgi_instance):
```
jeremyk, line 218: Looks like this should be `run_asgi_instance` as it appears line 215 is trying to call it?

jeremyk: @winglian - just checking to see if you saw this?

packages/python/vc_init.py
```diff
@@ -191,23 +191,33 @@ def __init__(self, scope):
...
 
                 asgi_instance = app(self.scope, self.receive, self.send)
 
-                asgi_task = loop.create_task(asgi_instance)
-                loop.run_until_complete(asgi_task)
+                if self._legacy_asyncio:
+                    asgi_task = loop.create_task(asgi_instance)
+                    loop.run_until_complete(asgi_task)
+                else:
+                    asyncio.run(self.run_asgi_instance(asgi_instance))
```
kytta, line 215: Do we actually need a separate function? From what I see in the handler, at this point in time, we already know that `app` (or `app.__call__`) is a coroutine. Can't we just run it directly instead?

```suggestion
                    asyncio.run(asgi_instance)
```

packages/python/vc_init.py
```diff
@@ -191,23 +191,33 @@ def __init__(self, scope):
                 self.state = ASGICycleState.REQUEST
                 self.app_queue = None
                 self.response = {}
+                self._legacy_asyncio = sys.version_info < (3, 10)
```
kytta, line 194: Not a fan of creating a class member variable just for checking. How about we declare a local variable at the top of the `__call__` function instead?

kytta (changes requested):
This should solve the issue. I have a couple more nitpicks, though, to (hopefully) make the code more readable :)

trek:
Closing in favor of https://github.com/vercel/vercel/pull/11675, which includes these commits and (some) feedback.

PR #11548, opened 2024-05-06
Revert "[examples] Enforce node18 on Python and Ruby examples"

janory:
Reverts vercel/vercel#11319

We [shipped](https://vercel.com/docs/deployments/build-image) support for Python 3.12 & Ruby 3.3.0 in AL2023, these examples should work now with Node 20 as well.

---

[Card](https://linear.app/vercel/issue/ZERO-2081/remove-node-18-packagejson-from-python-and-ruby-examples)


ofhouse (dismissed):

balazs4 (dismissed):

trek (dismissed):

EndangeredMassa (dismissed):

examples/zola/vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "env": {
+    "ZOLA_VERSION": "0.19.1"
```
EndangeredMassa, line 3: In the modern (Node 20) build container, we don't preinstall any binaries like zola or hugo. Instead, it we need to specify the version to be used with an env var like this one.

TooTallNate (approved):

PR #11390, opened 2024-04-05
[next] use middleware manifest worker environments

ecklf:
This PR adds changes to the Next.js builder to pass the per-worker environment emitted from the Next.js middleware manifest to the created `EdgeFunction` resource.

PR #11549, opened 2024-05-06
[next] rename middleware manifest env

huozhi:
Rename `environment` property to `env` from middleware manifest added in #11390.
Align with https://github.com/vercel/next.js/pull/64521

The `environments` property wasn't working properly before due to unfinished next.js changes, it should work now after testing with next build from https://github.com/vercel/next.js/pull/64521. We change it to a better naming word `env`.

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2784,7 +2784,7 @@ interface EdgeFunctionInfoV2 extends BaseEdgeFunctionInfo {
 
 interface EdgeFunctionInfoV3 extends BaseEdgeFunctionInfo {
   matchers: EdgeFunctionMatcher[];
-  environments: Record<string, string>;
+  env: Record<string, string>;
 }
 
 interface EdgeFunctionMatcher {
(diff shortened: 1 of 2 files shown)


styfle (approved):
Any way to add a test?

packages/next/src/utils.ts
```diff
@@ -2784,7 +2784,7 @@ interface EdgeFunctionInfoV2 extends BaseEdgeFunctionInfo {
 
 interface EdgeFunctionInfoV3 extends BaseEdgeFunctionInfo {
   matchers: EdgeFunctionMatcher[];
-  environments: Record<string, string>;
+  env: Record<string, string>;
```
ijjk, line 2787: This appears to be a breaking change so we need to bump the manifest version and ensure we handle both `environments` and `env` right?

huozhi: I explained in the description that the `environments` wasn't properly working before. And it only effects the preview props (draft mode) with canary. We didn't ship to any stable releases yet. Communicated with @javivelasco that we can ship rename them together here

javivelasco: I think this is fine because we have no stable version using that version package in production as far as I can tell!

javivelasco:
> Any way to add a test?

Since the functionality is already there let's ship the change and then we test in a follow up so we can ship it altogether to production @styfle 🙏

packages/next/src/utils.ts
ijjk: I see yea if it's not relied on for functionality in a proper release this is fine

ijjk (approved):

PR #11550, opened 2024-05-06
[build-utils] Fix warn for ignored project settings node version

onsclom:
There were scenarios where the warning "Node.js Version defined in your Project Settings ("18.x") will not apply" would not be triggered. For example:

1. Project Setting is 18.x
2. package.json has "engines.node": ">=18.x"
3. semver.intersects('18.x', '>=18.x') // => true (In this [code path](https://github.com/vercel/vercel/blob/8ea93839ccc70816f3ece9d7cfdb857aa7a4b015/packages/build-utils/src/fs/run-user-scripts.ts#L258)) -> No warning message that Node.js Version was changed
4. After the error message we evaluate the latest node Version to use (20.x) in getSupportedNodeVersion()(this [code path](https://github.com/vercel/vercel/blob/8ea93839ccc70816f3ece9d7cfdb857aa7a4b015/packages/build-utils/src/fs/run-user-scripts.ts#L274))
5. User doesn't get notified that we changed the node version to something different than their project setting

This fixes that scenario by getting the supported node version first.


.changeset/clean-ears-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/build-utils': minor
```
TooTallNate: If it's a fix then I would expect "patch". Is there a reason "minor" was used in particular?

onsclom: Whoops, fixed to patch now. Thanks for catching that!

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -251,15 +251,24 @@ export async function getNodeVersion(
...
   }
   const { packageJson } = await scanParentDirs(destPath, true);
-  let nodeVersion = config.nodeVersion || nodeVersionFallback;
-  let isAuto = true;
+  const nodeVersion = config.nodeVersion || nodeVersionFallback;
+
+  const pkgJsonNodeVersion = packageJson?.engines?.node;
+  const nodeVersionToUse = await getSupportedNodeVersion(
+    pkgJsonNodeVersion || nodeVersion,
+    pkgJsonNodeVersion ? false : true,
```
TooTallNate:
```suggestion
    !pkgJsonNodeVersion,
```

erikareads (approved):

PR #11553, opened 2024-05-07
Make waitUntil consistent for Node.js & Edge

Kikobeats:
This PR makes `waitUntil` consistent when interacting with it via `vc dev`. That includes:

**be possible to call `waitUntil` from the context second argument in web handlers functions (Node.js & Edge)**:

```js
export function GET(request, { waitUntil }) {
  waitUntil(fetch('https://vercel.com'));
  return new Response('OK');
}
```

**be possible to call `waitUntil` imported from `@vercel/functions`** (Node.js & Edge):

```js
import { waitUntil } from '@vercel/functions';

export function GET(request) {
  waitUntil(fetch('https://vercel.com'));
  return new Response('OK');
}
```

Additionally, `@vercel/functions` can be adopted for other framework to take advantage of this pattern at Vercel.

### DIFF
--- a/.changeset/wild-ladybugs-leave.md
+++ b/.changeset/wild-ladybugs-leave.md
@@ -1,0 +1,5 @@
+---
+"@vercel/node": minor
+---
+
+Make waitUntil consistent for Node.js & Edge
--- a/packages/functions/index.d.ts
+++ b/packages/functions/index.d.ts
@@ -1,0 +1,17 @@
+/**
+ * Extends the lifetime of the request handler for the lifetime of the given {@link Promise}
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil
+ *
+ * @param promise The promise to wait for.
+ * @example
+ *
+ * ```
+ * import { waitUntil } from '@vercel/functions';
+ *
+ * export function GET(request) {
+ *   waitUntil(fetch('https://vercel.com'));
+ *   return new Response('OK');
+ * }
+ * ```
+ */
+export const waitUntil: (promise: Promise<unknown>) => void;
--- a/packages/functions/index.js
+++ b/packages/functions/index.js
@@ -1,0 +1,24 @@
+/* global globalThis */
+
+export const waitUntil = promise => {
+  if (
+    promise instanceof Promise === false ||
+    promise === null ||
+    typeof promise !== 'object' ||
+    typeof promise.then !== 'function'
+  ) {
+    throw new TypeError(
+      `waitUntil can only be called with a Promise, got ${typeof promise}`
+    );
+  }
+
+  const ctx = globalThis[Symbol.for('@vercel/request-context')]?.get?.() ?? {};
+
+  if (!ctx.waitUntil) {
+    throw new Error(
+      'failed to get waitUntil function for this request context'
+    );
+  }
+
+  ctx.waitUntil(promise);
+};
--- a/packages/functions/index.test.js
+++ b/packages/functions/index.test.js
@@ -1,0 +1,45 @@
+/* global globalThis */
+
+import { vi, expect, test } from 'vitest';
+
+import { waitUntil } from '.';
+
+test.each([
+  {},
+  () => {},
+  function () {},
+  NaN,
+  1,
+  false,
+  undefined,
+  null,
+  [],
+  '▲',
+])('waitUntil throws when called with %s', input => {
+  expect(() => waitUntil(input)).toThrow(TypeError);
+  expect(() => waitUntil(input)).toThrow(
+    `waitUntil can only be called with a Promise, got ${typeof input}`
+  );
+});
+
+test.each([null, undefined, {}])(
+  'waitUntil throws when context is %s',
+  input => {
+    const promise = Promise.resolve();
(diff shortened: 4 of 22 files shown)


packages/functions/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "@vercel/functions"
```
TooTallNate: This will need a changeset if you would like it to be published to npm. Also, since it's a new package, I suggest adding:

```suggestion
  "name": "@vercel/functions",
  "publishConfig": {
    "access": "public"
  }
```

packages/node/test/unit/dev.test.ts
```diff
@@ -29,27 +29,87 @@ function testForkDevServer(
...
+    const elapsed = Date.now() - start;
+    if (runningTimeout) await setTimeout(runningTimeout - elapsed);
+    child.send('shutdown', err => {
+      if (err) child.kill(9);
+    });
+    await once(child, 'exit');
+  });
+}
+
+(NODE_MAJOR < 18 ? describe.skip : describe)('web handlers', () => {
```
Kikobeats, line 70: @TooTallNate right now CI is only running for Node.js v16, so ALL this test file is skipped; How we can setup a Github Action job just for ensure to run this file using Node.js v18?

.changeset/wild-ladybugs-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/node": minor
+---
+
+[cli]: Make waitUntil consistent for Node.js & Edge
```
TooTallNate: Prefix is redundant

```suggestion
Make waitUntil consistent for Node.js & Edge
```

TooTallNate (dismissed):

packages/functions/index.d.ts
```diff
@@ -0,0 +1 @@
+export const waitUntil: (promise: Promise<unknown>) => void;
```
Schniz, line 17: should we document this with examples and give references?

packages/functions/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@vercel/functions",
+  "description": "Runtime functions to be used with your Vercel Functions",
```
Schniz, line 3: we need to rethink this description

packages/node/src/edge-functions/edge-handler.mts
```diff
@@ -165,31 +184,25 @@ async function createEdgeRuntimeServer(params?: {
...
-      // it has a long-running waitUntil promise.
-      const WAIT_UNTIL_TIMEOUT = 10 * 1000;
-      const waitUntil = server.close();
-      return new Promise<void>((resolve, reject) => {
+    // @ts-expect-error
+    runtime.context.globalThis[Symbol.for('@vercel/request-context')] = {
+      get: () => ({
+        waitUntil: params.awaiter.waitUntil.bind(params.awaiter),
+      }),
+    };
```
Schniz, line 192: what happens if we have two requests concurrently? do we need AsyncLocalStorage?

Schniz: can we use Object.defineProperty here too?

packages/functions/index.test.js
```diff
@@ -0,0 +1,47 @@
...
+  1,
+  false,
+  undefined,
+  null,
+  [],
+  '▲',
+])('waitUntil throws when called with a non-promise (%s)', input => {
+  expect(() => waitUntil(input)).toThrow(TypeError);
+  expect(() => waitUntil(input)).toThrow(
+    'waitUntil can only be called with a Promise'
```
nuta: nit: Type errors should include the actual type we got to debug

```suggestion
    'waitUntil can only be called with a Promise, got ${typeof input}`
```

packages/functions/index.js
```diff
@@ -0,0 +1,14 @@
+/* global globalThis */
+
+export const waitUntil = promise => {
+  if (
+    promise === null ||
+    typeof promise !== 'object' ||
+    typeof promise.then !== 'function'
```
nuta, line 8: Why not `promise instanceof Promise`?

packages/functions/index.js
```diff
@@ -0,0 +1,14 @@
...
+  ) {
+    throw new TypeError('waitUntil can only be called with a Promise');
+  }
+
+  const ctx = globalThis[Symbol.for('@vercel/request-context')]?.get?.() ?? {};
+  ctx.waitUntil?.(promise);
```
nuta: Avoid failing silently. Make it explicit.

```suggestion
  const ctx = globalThis[Symbol.for('@vercel/request-context')]?.get?.() ?? {};
  let waitUntil = ctx.waitUntil?;
  if (!waitUntil) {
    throw new Error('failed to get waitUntil function for this request context');
  }
  
  waitUntil(promise);
```

packages/node/src/awaiter.ts
```diff
@@ -0,0 +1,27 @@
+export class Awaiter {
```
nuta, line 4: Leave a brief comment to describe what this class does so that readers don't have to read the whole code to grasp the overview.

packages/functions/index.d.ts
Kikobeats: done!

PR #11566, opened 2024-05-08
[next] Update react version in next fixtures

ijjk:
We are locking the peerDependency to react v19 beta on canary so we need to update these fixtures to tolerate that. 

x-ref: [slack thread](https://vercel.slack.com/archives/C048KLPEEH1/p1715191372631389)
x-ref: https://github.com/vercel/next.js/pull/65058

### DIFF
--- a/packages/next/test/fixtures/00-app-dir-edge-404/package.json
+++ b/packages/next/test/fixtures/00-app-dir-edge-404/package.json
@@ -1,8 +1,8 @@
 {
   "dependencies": {
     "next": "canary",
-    "react": "experimental",
-    "react-dom": "experimental"
+    "react": "19.0.0-beta-4508873393-20240430",
+    "react-dom": "19.0.0-beta-4508873393-20240430"
   },
   "ignoreNextjsUpdates": true
 }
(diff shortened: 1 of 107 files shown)


packages/next/test/fixtures/00-app-dir-edge-404/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "dependencies": {
     "next": "canary",
-    "react": "experimental",
-    "react-dom": "experimental"
+    "react": "19.0.0-beta-4508873393-20240430",
+    "react-dom": "19.0.0-beta-4508873393-20240430"
```
eps1lon, line 5: I guess these can stay `experimental` since they never matched the peer dependency range to begin with?

TooTallNate (dismissed):

packages/next/test/fixtures/00-app-dir-edge-404/package.json
ijjk: Was using the tag bypassing the peer dep check maybe?

eps1lon: It shouldn't because the peer dep check is done on the resolved version not on the tag.

ijjk: Hmm will leave locked then to be safe/for consistency

ztanner (approved):

PR #11576, opened 2024-05-10
[next]: skip action rewrite if RSC header is present

ztanner:
When rewriting requests that contain a `next-action` header to a `.action` output, we should do so only in the case where the RSC header isn't present. When a Next.js server action triggers a `redirect`, an intra-app RSC request will be made to that requested URL, with the `next-action` header still present. This means the intra-app request will be incorrectly rewritten, when we only intended to handle the actual action request.

This was patched in Next.js in the following PR:
- https://github.com/vercel/next.js/pull/65615

But it seems like a good idea to be resilient to this here as well.


packages/next/src/server-build.ts
```diff
@@ -1956,14 +1956,20 @@ export async function serverBuild({
                   // Also includes separate handling for index routes which should match to /index.action.
                   // This follows the same pattern as the rewrites for .rsc files.
                   {
-                    src: `^${path.posix.join('/', entryDirectory, '/')}`,
+                    src: `^${path.posix.join('/', entryDirectory, '/?')}`,
```
ztanner, line 1959: Unrelated to the core change here, but I updated this to match the same change for RSC handling in this PR, since it seems relevant to both cases:

- https://github.com/vercel/vercel/pull/11528

ijjk (approved):

PR #11578, opened 2024-05-10
[tests] fix maxDuration integration test

ztanner:
Since maxDuration was updated for this tier ([x-ref](https://vercel.com/changelog/vercel-functions-for-hobby-can-now-run-up-to-60-seconds))


EndangeredMassa (dismissed):

packages/cli/test/integration-3.test.ts
```diff
@@ -1255,7 +1255,7 @@ test('fail to deploy a Lambda with an incorrect value for maxDuration', async ()
 
   expect(output.exitCode, formatOutput(output)).toBe(1);
   expect(output.stderr).toMatch(
-    /maxDuration must be between 1 second and 10 seconds/gm
+    /maxDuration must be between 1 second and 60 seconds/gm
```
trek: Could we update this regex to not care about the specific values? I think we care about the type of error, not the specific time values.

trek (approved):

TooTallNate (approved):

PR #11596, opened 2024-05-14
fix corepack detection for package manager version determination

EndangeredMassa:
The previous logic was checking for the env var `ENABLE_EXPERIMENTAL_COREPACK` to determine if corepack was being used by a project. However, this value only means that the build system should consider corepack, not that it's actively being used.

We need to check that flag AND the existence of a `packageManager` property in the project's `package.json`.


trek (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -413,9 +413,8 @@ export async function runNpmInstall(
 
   try {
     await runNpmInstallSema.acquire();
-    const { cliType, packageJsonPath, lockfileVersion } = await scanParentDirs(
-      destPath
-    );
+    const { cliType, packageJsonPath, packageJson, lockfileVersion } =
+      await scanParentDirs(destPath, true);
```
erikareads, line 430: I dislike how little the extra `true` tells me about what the function is now doing that it wasn't before.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -553,10 +555,18 @@ export function getEnvForPackageManager({
...
+
+  if (corepackEnabled) {
+    debug(
+      `Detected corepack use for "${packageJsonPackageManager}". Not overriding package manager version.`
+    );
+  } else {
+    debug(
+      `Detected ${detectedPackageManager}. Added "${newPath}" to path. Based on assumed package manager "${cliType}", lockfile "${detectedLockfile}", and lockfileVersion "${lockfileVersion}"`
+    );
+  }
```
erikareads: This change only seems to affect the debug logging, won't we need to change this in `getPathOverrideForPackageManager` too?

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: Agreed. This style of signature is not great.

packages/build-utils/src/fs/run-user-scripts.ts
erikareads: specifically here: https://github.com/vercel/vercel/blob/265f6888a4e9034f0e46feb62b0b8ba367b4a6ed/packages/build-utils/src/fs/run-user-scripts.ts#L683

erikareads (changes requested):

EndangeredMassa:
I'm updating this to include the actual fix, not just the logging fix.

packages/build-utils/test/unit.get-env-for-package-manager.test.ts
```diff
@@ -472,23 +496,26 @@ describe('Test `getPathOverrideForPackageManager()`', () => {
...
       })
     ).toStrictEqual(want);
   });
 });
+
 describe('Test `getPathForPackageManager()`', () => {
   test.each<{
     name: string;
-    args: Parameters<typeof getEnvForPackageManager>[0];
+    args: Parameters<typeof getPathForPackageManager>[0];
```
EndangeredMassa, line 510: These types happened to be the same, but this code was referencing the wrong function.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -739,11 +747,13 @@ export function getPathOverrideForPackageManager({
 export function getPathForPackageManager({
   cliType,
   lockfileVersion,
+  corepackEnabled,
```
erikareads: the `getPathForPackageManager` function was left in to avoid a breaking change for a publicly exported function, why are we changing the call signature?

.changeset/soft-bananas-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/build-utils": patch
```
erikareads: Now that we're changing the call signature of `getEnvForPackageManager` this is probably more than a patch change, arguably major.

packages/build-utils/test/unit.get-env-for-package-manager.test.ts
erikareads: good catch!

.changeset/soft-bananas-care.md
EndangeredMassa: Yeah, I shouldn't mess with this. Will revert.

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: Yeah, I shouldn't mess with this. Will revert.

erikareads: If we can we should eventually mark `getPathForPackageManager` as deprecated.

.changeset/soft-bananas-care.md
EndangeredMassa: https://github.com/vercel/vercel/pull/11596/commits/cfe6b1ad44c54c072af40a4833235a722e906ebe

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: I marked it as deprecated: https://github.com/vercel/vercel/pull/11596/commits/0ef8ebf65c58f6f98bff8fbc80f7094cd4105681

EndangeredMassa: ended up making a backwards-compatible change to the signature in case third party builders are using it: getPathForPackageManager

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -748,6 +756,7 @@ export function getPathOverrideForPackageManager({
 /**
  * Helper to get the binary paths that link to the used package manager.
  * Note: Make sure it doesn't contain any `console.log` calls.
+ * @deprecated use `getPathOverrideForPackageManager` instead
```
erikareads: Note (non-blocking): I didn't export `getPathOverrideForPackageManager` so either we need to export that, or better just recommend using `getEnvForPackageManager` instead.

erikareads: That is, `getPathOverrideForPackageManager` is exported from the module, but not from the `build-utils` package.

erikareads (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: My mistake. I meant `getEnvForPackageManager`. I updated it. https://github.com/vercel/vercel/pull/11596/commits/bd57cf391a731924aff4a5c34a4d5b19d8a2dff9

erikareads (approved):

AurelienGasser:
This PR may have introduced a regression: https://github.com/vercel/vercel/issues/11607

wweaver:
@EndangeredMassa, with #11607 is that going to take any priority? Or no since corepack is still considered experimental in Vercel's docs https://vercel.com/changelog/corepack-experimental-is-now-available.

PR #11597, opened 2024-05-14
[tests] allow errors to define retry delay

EndangeredMassa:
Errors can now define their own retry delay. This is necessary for some kinds of errors where all retries will run within the same second.

### DIFF
--- a/test/lib/deployment/test-deployment.js
+++ b/test/lib/deployment/test-deployment.js
@@ -108,6 +108,7 @@ async function runProbe(probe, deploymentId, deploymentUrl, ctx) {
           found = true;
           break;
         } else {
+          ctx.deploymentLogs = null;
           throw new Error(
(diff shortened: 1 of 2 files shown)


trek (dismissed):

erikareads (dismissed):

TooTallNate (approved):

test/lib/deployment/test-deployment.js
```diff
@@ -449,6 +452,8 @@ async function testDeployment(fixturePath, opts = {}) {
         throw err;
       }
 
+      const retryDelay = Math.max(probe.retryDelay || 0, err.retryDelay || 0);
```
erikareads, line 455: Does this implicitly assume that `probe` and `err` will have different delays and won't step on each other? @EndangeredMassa

PR #11600, opened 2024-05-15
[cli] Add test for CLI extensions passing through exit code

TooTallNate:
/cc @codybrouwers


packages/cli/test/integration-3.test.ts
```diff
@@ -1327,6 +1327,19 @@ test('should invoke CLI extension', async () => {
   expect(output.stdout, formatted).toContain(`Username: ${contextName}`);
 });
 
+test('should pass through exit code for CLI extension', async () => {
+  const fixture = path.join(__dirname, 'fixtures/e2e/cli-extension-exit-code');
+
+  // Ensure the `.bin` is populated in the fixture
+  await runNpmInstall(fixture);
+
+  const output = await execCli(binaryPath, ['fail'], {
```
erikareads, line 1336: how do we know that the command here is `'fail'`?

TooTallNate: It's how this `fixtures/e2e/cli-extension-exit-code` is configured. The fixture has a dependency which includes a `vercel-fail` executable via `bin` in package.json.

erikareads (approved):

PR #6553, opened 2021-08-04
[build-utils] check to see that the node version is less then 16 befo…

ktcarter:
### Related Issues

Check the project's nodeVersion setting for when we release Node 16 and later

#### Tests

- [ ] The code changed/added as part of this PR has been covered with tests
- [ ] All tests pass locally with `yarn test-unit`

#### Code Review

- [ ] This PR has a concise title and thorough description useful to a reviewer
- [ ] Issue from task tracker has a link to this PR

PR #11603, opened 2024-05-15
warn on mismatched corepack and detected package managers

EndangeredMassa:
When corepack is used and the package manager or version does not match the inferred package manager or version from the lockfile, we should throw an error. Right now, this can happen with `pnpm` (for example) where `pnpm` only logs a warning, causing indeterministic builds.

This PR mostly updates `getPathOverrideForPackageManager` to not care about corepack, allowing the calling code to compare the detected values to the corepack values.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -687,69 +689,160 @@ export function getPathOverrideForPackageManager({
...
           return {
             path: '/pnpm9/node_modules/.bin',
             detectedLockfile: 'pnpm-lock.yaml',
-            detectedPackageManager: 'pnpm 9',
+            detectedPackageManager: 'pnpm@9.x',
           };
         case 'pnpm 6':
         default:
-          return no_override;
+          return undefined;
```
EndangeredMassa, line 894: Question: Do we want to be more specific here with what was detected, but maybe leave the `path` blank because it matches the default? This would make the comparison between detected package managers and versions to corepack specified ones more complete.

Right now, detecting `pnpm@6.x` (for example) would come back as nothing, not comparing to what was specified by corepack.

erikareads: I'm in favor of being more specific in what we return.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -687,69 +689,160 @@ export function getPathOverrideForPackageManager({
...
+    )
+  ) {
+    // corepack is going to take care of it; do nothing special
+    return NO_OVERRIDE;
+  }
+
+  throw new Error(
+    `Detected package manager (by lockfile) "${detectedPackageManger.detectedLockfile}" does not match intended corepack package manager "${corepackPackageManager}". Update your lockfile or "package.json#packageManager" values to match.`
+  );
+}
```
erikareads: Do we have a test case for this throw?

trek (dismissed):

trek (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -687,69 +689,160 @@ export function getPathOverrideForPackageManager({
...
+  cliType: CliType,
+  lockfileVersion: number | undefined,
+  nodeVersion: NodeVersion | undefined
+) {
   switch (cliType) {
     case 'npm':
       switch (true) {
-        case corepackEnabled:
-          return no_override;
         case shouldUseNpm7(lockfileVersion, nodeVersion):
```
EndangeredMassa: I think we need to delete this function. If the data is flowing properly, it looks like it always returns false. I'd check the PR that introduced this logic to be sure we're considering the full context.

EndangeredMassa: I think we should also fix the build container. It looks like it's not passing in node version (at least some of the time) into the exported `runNpmInstall` function. Make sure it's passing that in every time it calls that function.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -687,69 +689,160 @@ export function getPathOverrideForPackageManager({
...
-            detectedPackageManager: 'pnpm 7',
+            detectedPackageManager: 'pnpm@7.x',
           };
         case 'pnpm 8':
           // pnpm 8
           return {
             path: '/pnpm8/node_modules/.bin',
             detectedLockfile: 'pnpm-lock.yaml',
-            detectedPackageManager: 'pnpm 8',
+            detectedPackageManager: 'pnpm@8.x',
```
EndangeredMassa, line 877: We need to broaden the detection here. We know that (in this case for example, but this applies to all cases) the lockfile was generated by `pnpm@8`, but it could have been read by `pnpm@8` or `pnpm@9`. We need to only fail the build if you try to use some other `pnpm` version.

Table of supported values: https://github.com/pnpm/spec/issues/7

packages/build-utils/src/fs/run-user-scripts.ts
erikareads: Nothing of note in the PR: https://github.com/vercel/vercel/pull/6553

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -719,103 +739,149 @@ export function getPathOverrideForPackageManager({
...
+    throw new Error(
+      `Detected package manager "${cliType}" does not match intended corepack defined package manager "${validatedCorepackPackageManager.packageName}". Change your lockfile or "package.json#packageManager" value to match.`
+    );
+  }
+
+  const corepackPackageManagerVersion = coerce(
+    validatedCorepackPackageManager.packageVersionRange
+  );
+
+  console.error(lockfileVersion);
```
onsclom: Remove?

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -682,12 +685,30 @@ function detectPnpmVersion(
...
+) {
+  switch (cliType) {
+    case 'npm':
+    case 'bun':
+    case 'yarn':
+      return true;
+    case 'pnpm':
+      switch (packageManagerMajorVersion) {
+        case 9:
+          return [6.0, 7.0, 9.0].includes(lockfileVersion);
```
onsclom, line 710: nitpick: Remove unnecessary `.0`s

onsclom (changes requested):

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: I don't mind being explicit, but either is fine with me

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -682,12 +685,30 @@ function detectPnpmVersion(
...
+  packageManagerMajorVersion: number
+) {
+  switch (cliType) {
+    case 'npm':
+    case 'bun':
+    case 'yarn':
+      return true;
+    case 'pnpm':
+      switch (packageManagerMajorVersion) {
+        case 9:
```
EndangeredMassa, line 702: **suggestion (soft-blocking):** false if `pnpm@9.0.0` and lockfile `6.0`

If the lockfile version is `6.0` and the pnpm version is exactly `9.0.0`, this is not compatible. It was a mistake in `pnpm@9.0.0`, specifically, that was fixed in `pnpm@9.0.1`.

We should return false in this specific case.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -682,12 +685,30 @@ function detectPnpmVersion(
...
+          return [6.0, 7.0, 9.0].includes(lockfileVersion);
+        case 8:
+          return [6.0, 6.1].includes(lockfileVersion);
+        case 7:
+          return [5.3, 5.4].includes(lockfileVersion);
+        case 6:
+          return [5.3, 5.4].includes(lockfileVersion);
+        default:
+          return false;
```
EndangeredMassa: **suggestion (blocking):** `true` instead

If a user sets corepack to install `pnpm@3` (for example), then we don't know (because we don't need to bother mapping that far back) which lockfile versions will work.

I think we should by default assume it will work, letting the package manager itself decide what happens. We should only return `false` if we know for sure it won't work.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -719,105 +739,149 @@ export function getPathOverrideForPackageManager({
...
-        case shouldUseNpm7(lockfileVersion, nodeVersion):
-          return {
-            path: '/node16/bin-npm7',
-            detectedLockfile: 'package-lock.json',
-            detectedPackageManager: 'npm 7+',
-          };
-        default:
-          return no_override;
-      }
+      return undefined;
```
EndangeredMassa, line 862: **suggestion (non-blocking):** ‏do we want a comment here about why this is `undefined`?

It's undefined because we do know that `npm` will be used, but we're going to let the version of Node.js that's installed manage which version of `npm` will be used. So, from this function's perspective, we're not specifying a version of `npm` that will be used.

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: I still think we should make this change. `pnpm@6` should return something like:

```
{
  path: undefined,
  detectedLockfile: 'pnpm-lock.yaml',
  detectedPackageManager: 'pnpm 6',
}
```

Maybe we go further and rename `path` here to `overridePath`.

The same would apply to `yarn` below.

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: Make sure we have a test case for this. A lockfile of version `5.3` and `5.4` should both work with either a pnpm version of `6.x` or `7.x`.

EndangeredMassa (commented):
Please do not merge until addressing the "blocking" comments.

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: Tests for this throw were [added here](https://github.com/vercel/vercel/blob/2a4e1f3d48ed0e8b750204f5fc5d35c76caf9bc6/packages/build-utils/test/unit.get-env-for-package-manager.test.ts#L486).

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`2269bbc` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/2269bbcf0698a0f9c875918c715925a24b526601)

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`e51618c` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/e51618c084b640bfb89e26f8f4b745aa7c4465af)

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`3292f14` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/3292f14bfc0e61856eef3364efc5ee0222acb415)

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`1fede25` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/1fede25fbca277c64e75c0851bcbb287ab9325e3)

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`b1c205a` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/b1c205adf43d21ad3d2321a5c540e150879308cb)

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -719,103 +747,160 @@ export function getPathOverrideForPackageManager({
...
+      `Detected package manager "${cliType}" does not match intended corepack defined package manager "${validatedCorepackPackageManager.packageName}". Change your lockfile or "package.json#packageManager" value to match.`
+    );
+  }
+
+  const corepackPackageManagerVersion = coerce(
+    validatedCorepackPackageManager.packageVersionRange
+  );
+
+  if (corepackPackageManagerVersion === null) {
+    return true;
```
EndangeredMassa, line 809: **question (soft-blocking):** Should this be `false`?

I ran out of time right now, but I want to dig into this deeper to make sure this is the value we want.‏

packages/build-utils/test/unit.get-env-for-package-manager.test.ts
```diff
@@ -491,17 +454,156 @@ describe('Test `getPathOverrideForPackageManager()`', () => {
         path: undefined,
       },
     },
+    {
+      name: 'should not set pnpm path if corepack is enabled, pnpm 9 and valid lockfile version',
+      args: {
+        cliType: 'pnpm',
+        nodeVersion: { major: 16, range: '16.x', runtime: 'nodejs16.x' },
```
EndangeredMassa: **suggestion (non-blocking):** use Node 18

The test doesn't care about this, but to be more complete, this should be at least Node 18 because `pnpm@9` doesn't run on Node 16.

```suggestion
        nodeVersion: { major: 18, range: '18.x', runtime: 'nodejs18.x' },
```‏

File: packages/build-utils/src/fs/run-user-scripts.ts [hunk 12]
```diff
@@ -719,103 +747,160 @@ export function getPathOverrideForPackageManager({
...
           // pnpm 9
           return {
             path: '/pnpm9/node_modules/.bin',
             detectedLockfile: 'pnpm-lock.yaml',
-            detectedPackageManager: 'pnpm 9',
+            detectedPackageManager: 'pnpm@9.x',
           };
         case 'pnpm 6':
+          return {
+            path: undefined,
```
EndangeredMassa, line 889: **suggestion (non-blocking):** add comment

I think we should add a comment here to make it clear that this is `undefined` because `pnpm@6` is the current default in the build container.‏

packages/build-utils/test/unit.get-env-for-package-manager.test.ts
```diff
@@ -491,17 +454,156 @@ describe('Test `getPathOverrideForPackageManager()`', () => {
...
+        detectedPackageManager: undefined,
+        path: undefined,
+      },
+    },
+    {
+      name: 'should allow lockfile 5.3 with pnpm 7.x',
+      args: {
+        cliType: 'pnpm',
+        nodeVersion: { major: 16, range: '16.x', runtime: 'nodejs16.x' },
+        packageJsonPackageManager: 'pnpm@7.x',
```
EndangeredMassa: **suggestion (blocking):** change test setup

The `packageJsonPackageManager` value is only used if corepack is enabled. Because corepack is not enabled for this test, the lockfile is used to determine the pnpm version, which would be `pnpm@6` in this case. The test description here implies it would be `pnpm@7`.

This test is currently testing that having a `packageJsonPackageManager` doesn't change the result, which is also a useful test, but doesn't match the description.

I think we should walk through these tests together to make sure we testing the different conditions we need.‏

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: [`b25c6ce` (#11603)](https://github.com/vercel/vercel/pull/11603/commits/b25c6ce949b75ca17d8834aeeca942f5ac09c5cd)

packages/build-utils/src/fs/run-user-scripts.ts
EndangeredMassa: After discussing it more with @onsclom, we decided to leave this as `true`, but add a comment about how this allows corepack to throw an appropriate error.

onsclom:
@EndangeredMassa and I reviewed this. Approving

onsclom (approved):

PR #11631, opened 2024-05-22
[node] Use `vitest` for unit tests

TooTallNate:
Similar to #11302, but for the `@vercel/node` package.


cb1kenobi:
Heck yeah! Great idea!

trek (approved):

PR #11653, opened 2024-05-23
[cli] Collect diagnostics from builders during `vc build`

TooTallNate:
Adds the ability for builders to define a `diagnostics` step that is called after the build operation is done (either failed or successful).
Implements the diagnostics step in the next builder.

**References**
- https://github.com/vercel/next.js/pull/66187


packages/cli/src/commands/build/index.ts
```diff
@@ -575,6 +590,28 @@ async function doBuild(
...
+      // todo: extract in a separate function
+      diagnostics['framework.json'] = new FileBlob({
+        data: Buffer.from(
+          JSON.stringify({
+            // todo: is this the right way to report the framework?
+            slug: projectSettings.framework,
+            version: framework?.version,
+          })
+        ),
+      });
```
EndangeredMassa: In a meeting, we discussed removing this

.changeset/hot-rules-tickle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/build-utils': minor
+'vercel': minor
+---
+
+Output "diagnostics" directory during `vc build`
```
EndangeredMassa: It would be great to have some tests around this logic

ofhouse: @TooTallNate Could you help with setting up some test?

trek (approved):

PR #11662, opened 2024-05-27
[frameworks] Add "Modern.js" framework preset

yimingjfe:
Add the preset of [Modern.js](https://modernjs.dev/) to allow users to deploy without zero configuration.


EndangeredMassa:
Thanks for the suggestion! Balancing support costs of adding new presets, we're not going to add this at this time.

You can definitely deploy this to Vercel, but you'll have to customize the [Project Settings](https://vercel.com/docs/deployments/configure-a-build#build-and-development-settings) to set your build (and possibly other) commands.

PR #11672, opened 2024-05-30
[cli] implements vc deploy --logs and vc inspect --logs

feugy:
### 🧐 What's in there?

This is the first part of distinguishing build logs from runtime logs.
The proposal is to provide:
- `vc deploy --logs`: in addition to deploying the application, it should tail and display the **build** logs until the build finishes.
- `vc inspect --logs`: instead of retuning a summary of the builds, it reads the **build** logs
- `vc inspect --logs --wait`: when the build is in progress, tails the **build** logs until the configured timeout (3m by default), or display all available logs when the build is finished.
- `vc logs`: read the **runtime** logs (not part of this PR, breaking change)

### 🧪 How to test?

Given a local application location in `xyz` folder, already linked to a vercel project:
- `pnpm -F vercel dev deploy --cwd xyz --logs` to deploy and see build logs
- `pnpm -F vercel dev deploy --cwd xyz` to deploy and see spinners instead
- `pnpm -F vercel dev inspect --cwd xyz dpl_ID/url --logs` to see build logs of an existing deployment
- `pnpm -F vercel dev inspect --cwd xyz dpl_ID/url` to see an existing deployment details
- `pnpm -F vercel dev inspect --cwd xyz dpl_ID/url --logs --wait` to see build logs of an existing deployment (better if in progress)
- `pnpm -F vercel dev inspect --cwd xyz dpl_ID/url --logs --wait --timeout 5s` to see build logs of an existing deployment (better if in progress) and bail after 5 seconds

Unit tests added:
- inspect command: `pnpm -F vercel exec vitest run /inspect.test`
- deploy command: `pnpm -F vercel exec vitest run /deploy.test`

### ❗ Notes to reviewers

It's better to [omit whitespace changes](https://github.com/vercel/vercel/pull/11672/files?diff=split&w=1) when reviewing `packages/cli/src/util/events.ts` since most of the code was enclosed in a try-catch.


packages/cli/src/util/events.ts
```diff
@@ -60,112 +62,126 @@ async function printEvents(
...
-                    client,
-                    contextName,
-                    deploymentIdOrURL
-                  );
-                  if (json.readyState === 'READY') {
+      try {
+        const eventsRes = await client.fetch(eventsUrl, {
+          json: false,
+          signal: abortController?.signal,
+        });
```
feugy, line 70: `client.fetch()` supports node-retry options, so this code may be refactored and not being wrapped in a retry call.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -264,13 +280,46 @@ export default async function processDeployment({
...
+  client: Client,
+  deployment: Deployment,
+  follow: boolean = true
+) {
+  const abortController = new AbortController();
+  const promise = printEvents(
+    client,
+    deployment.id,
+    {
+      mode: 'logs',
```
feugy: We don't use `'deploy'` mode here (which triggers its own polling), because it's actually not working. 
The API does not return the special "events" which are meant to controller polling.

I highly doubt it's working, and I would suggest  we remove this mode parameter and the related code.

packages/cli/src/commands/inspect/index.ts
```diff
@@ -92,20 +99,46 @@ export default async function inspect(client: Client) {
...
     deployment = await getDeployment(client, contextName, deploymentIdOrHost);
+    if (!isDeploying(deployment.readyState)) {
+      abortController?.abort();
+      if (!isReady(deployment) && !isFailed(deployment)) {
+        print(
+          chalk.bold(
+            `\n\nstop waiting for logs after ${timeout}s. Deployment is ${deployment.readyState}.\n`
+          )
+        );
+      }
```
lucleray: I'm not sure I understand the logic here. The message says that it stops waiting after a timeout, but I can't see that in the condition 🤔

packages/cli/src/commands/inspect/index.ts
```diff
@@ -197,3 +230,11 @@ function stateString(s: Deployment['readyState']) {
       return chalk.gray('UNKNOWN');
   }
 }
+
+function exitCode(state: Deployment['readyState'], print?: Output['print']) {
+  print?.(`${chalk.cyan('status')}\t${stateString(state)}\n`);
+  if (state === 'ERROR' || state === 'CANCELED') {
+    return 3;
+  }
```
lucleray: Question: why do we want to return a non-0 error code for ERROR/CANCELED? I'm asking because I think some bash scripts could consider that inspecting the deployment failed if they don't receive 0.

packages/cli/src/commands/inspect/index.ts
```diff
@@ -92,20 +99,46 @@ export default async function inspect(client: Client) {
...
   // resolve the deployment, since we might have been given an alias
   let deployment = await getDeployment(client, contextName, deploymentIdOrHost);
 
+  let abortController: AbortController | undefined;
+  if (withLogs) {
+    if (wait) {
+      abortController = displayBuildLogs(client, deployment, true);
+    } else {
+      await displayBuildLogs(client, deployment, false);
+      return exitCode(deployment.readyState, print);
```
lucleray: Nit: I saw these `return exitCode(...)` spread in a few places. It would be simpler to read the code if they were all in a single place later in the code that does something like:

```
// only log the status when `--logs` is passed
if (withLogs) {
  // print status
  return exitCode(deployment.readyState)
}

// fetch and print the whole details
```

packages/cli/src/commands/inspect/index.ts
```diff
@@ -92,20 +99,46 @@ export default async function inspect(client: Client) {
...
+    }
+  }
+
   while (Date.now() < until) {
-    if (!wait || !isDeploying(deployment.readyState)) {
+    if (!wait) {
```
lucleray: Nit: maybe wrap the `while` with `if (wait)` or check `wait` in the `while` condition instead of breaking inside the while

.changeset/curly-otters-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': minor
```
lucleray, line 2: Question: Should this be "major"?

packages/cli/src/commands/logs/index.ts
```diff
@@ -181,6 +181,10 @@ function printLogShort(log: any) {
 
   const date = new Date(log.created).toISOString();
 
+  const print = client
+    ? (str: string) => client?.output.print(str + '\n')
+    : // eslint-disable-next-line no-console
+      console.log.bind(console);
```
lucleray: Nit: can you explain why we need to `.bind`? (as a code comment)

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -208,18 +220,21 @@ export default async function processDeployment({
...
+        }
         output.spinner(
           deployment.readyState === 'QUEUED' ? 'Queued' : 'Building',
           0
         );
       }
 
-      if (event.type === 'building') {
+      if (event.type === 'building' && !withLogs) {
         output.spinner('Building', 0);
```
lucleray, line 239: Maybe you can introduce another function that is a little symmetric with `stopSpinner`:

```
function startSpinner(message) {
  if (!withLogs) {
    output.spinner(message, 0)
  }
}
```

And then add a comment next to the 2 functions to explains that for `--logs`, we don't show spinner for reason X.

packages/cli/src/util/deploy/process-deployment.ts
lucleray: Do we use the `deploy` anywhere?

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -264,13 +280,43 @@ export default async function processDeployment({
...
+    deployment.id,
+    {
+      mode: 'logs',
+      onEvent: (event: any) => printLogShort(event, client),
+      quiet: false,
+      findOpts: { direction: 'forward', follow },
+    },
+    abortController
+  );
+  return follow ? abortController : promise;
```
lucleray: Here we might keep an "unhandled promise" if `follow=true` and the promise throws. We should handle the promise throwing if we don't return it.

packages/cli/src/util/events.ts
```diff
@@ -59,113 +61,128 @@ async function printEvents(
...
-                    client,
-                    contextName,
-                    deploymentIdOrURL
-                  );
-                  if (json.readyState === 'READY') {
+      const eventsUrl = `/v3/now/deployments/${deploymentIdOrURL}/events?${query}`;
+      try {
+        const eventsRes = await client.fetch(eventsUrl, {
+          json: false,
+          // @ts-expect-error: typescipt is getting confused with the signal types from node (web & server) and node-fetch (server only)
```
lucleray:
```suggestion
          // @ts-expect-error: typescript is getting confused with the signal types from node (web & server) and node-fetch (server only)
```

packages/cli/src/util/events.ts
```diff
@@ -59,113 +61,128 @@ async function printEvents(
...
+          signal: abortController?.signal,
+        });
+
+        if (eventsRes.ok) {
```
lucleray, line 72: Nit: put the code for non-ok status code first and return/throw there, so we can get that mental load out of our brain when reading the code 😄

```
if (!eventsRes.ok) {
  if (eventsRes.status < 500) {
    bail(err);
    return;
  }

  throw err;
}

// handle happy path here
```

packages/cli/src/util/events.ts
```diff
@@ -59,113 +61,128 @@ async function printEvents(
...
+          const readable = eventsRes.body;
+
+          // handle the event stream and make the promise get rejected
+          // if errors occur so we can retry
+          return new Promise<void>((resolve, reject) => {
```
lucleray, line 77: I took a look at this promise, and it looks mostly fine, but it's more difficult to read the code and make sure it works right in case of a retry 🤔

packages/cli/src/util/deploy/process-deployment.ts
feugy: Not to my knowledge.

packages/cli/src/commands/inspect/index.ts
feugy: That is a good question. @EndangeredMassa suggested to [implement a feature request](https://www.notion.so/RFC-split-logs-commands-for-CLI-95800af3532f4b6fb7acf0021247e339?d=74e0ce8458734a51a4a59902886e7e7d&pvs=4#b044f393675f4d459038a010c414cfb9) along with this change.
It'll allow them to detect a failed/cancelled deployment when inspecting logs with the `--wait` flag.

.changeset/curly-otters-guess.md
feugy: To my understanding, it should not. Major is when you break compatibility, while this change is a pure addition.

packages/cli/src/commands/inspect/index.ts
feugy: Here is the full extract
```ts
while (Date.now() < until) {
   if (!wait) {
     break;
   }

   // ...
   if (!isDeploying(deployment.readyState)) {
      abortController?.abort();
      if (!isReady(deployment) && !isFailed(deployment)) {
        print(chalk.bold(`\n\nstop waiting for logs after ${timeout}s. Deployment is ${deployment.readyState}.\n`));
      }
      if (withLogs) {
        return exitCode(deployment.readyState, print);
      } else {
        break;
      }
    }
  }

  // ... code displaying the deployment's artifacts
```

This loop only happens when waiting (while condition at the top). 
Then if the build is finished, we'll stop the waiting.
However, we don't have the same logic:
- if we were inspecting logs, then we exit the command, because we don't want to:
- if we don't inspect logs, then we print the artifacts

packages/cli/src/commands/inspect/index.ts
feugy: Yes, but I also didn't meant to reshape the whole function. However you're right and I'll make bigger changes.

packages/cli/src/util/deploy/process-deployment.ts
feugy: I agree with you. However, these promises are never catch by the caller function, and will interrupt the program anyway.

this `displayBuildLogs()` function already handles retries internally. I'll only reject its promise in case it retried enough, which means to stop the entire program.

packages/cli/src/util/events.ts
feugy: If it were me, I'll completely rebuild this whole function with effects to get more readable code.

But here again, I meant to inject my changes in the existing code without compromising its shape.

.changeset/curly-otters-guess.md
lucleray: The 2 possible "breaking" changes in this PR:
- ~~`vc logs` doesn't return build logs anymore~~ (see Damien's comment below)
- `vc inspect` returns new exit code (3) when deployment is not ready

packages/cli/test/integration-3.test.ts
```diff
@@ -330,7 +330,7 @@ test('output logs with "short" output', async () => {
 
   // "short" format includes timestamps
   expect(
-    stdout.match(
+    stderr.match(
```
feugy: **⚠️ Important note** 

This is because I've reused `client.output.print` in a shared utility function that is used by `vc deploy`, `vc inspect` and `vc logs`.

`vc logs` used to print directly to stdout.
Is that an issue @EndangeredMassa ?

cc @lucleray (we changed it this morning)

.changeset/curly-otters-guess.md
feugy: `vc logs` still returns build logs in this PR. But with this morning's refactoring, its output is stderr instead of stdout.
True about the `vc inspect` exit code.

packages/cli/test/integration-3.test.ts
lucleray: I think we should keep using `stdout`. The idea is that `stdout` only contains the "output", so the logs in this case, while `stderr` contains the messaging.

So, I think we need to revert the change and keep using `console.log` for now 🤔

.changeset/curly-otters-guess.md
feugy: Team decision: we're not changing the exit codes yet, I'll revert this change

packages/cli/src/commands/inspect/index.ts
```diff
@@ -174,26 +233,11 @@ export default async function inspect(client: Client) {
...
-    case 'CANCELED':
-      return chalk.gray(CIRCLE) + sTitle;
-    default:
-      return chalk.gray('UNKNOWN');
+function exitCode(state: Deployment['readyState']) {
+  if (state === 'ERROR' || state === 'CANCELED') {
+    return 1;
   }
+  return 0;
 }
```
lucleray, line 243: Shouldn't this be returning `0` all the time if you don't want to change how status code is returned?

lucleray (approved):

trek (approved):

packages/cli/src/commands/inspect/index.ts
```diff
@@ -174,26 +233,11 @@ export default async function inspect(client: Client) {
...
-      return chalk.green(CIRCLE) + sTitle;
-    case 'QUEUED':
-      return chalk.gray(CIRCLE) + sTitle;
-    case 'CANCELED':
-      return chalk.gray(CIRCLE) + sTitle;
-    default:
-      return chalk.gray('UNKNOWN');
+function exitCode(state: Deployment['readyState']) {
+  if (state === 'ERROR' || state === 'CANCELED') {
+    return 1;
```
EndangeredMassa, line 240: This is a change, but I think it makes sense.

EndangeredMassa (approved):

PR #11675, opened 2024-05-30
support newer python versions

trek:
When we added the option of getting `al2023` in addition to `al2`, this required a python version bump for those apps on `al2023`. Python 3.10 had a breaking minor change. https://github.com/vercel/vercel/pull/11541 mostly fixed it, with some comments.

OP hasn't replied to feedback on that PR, so I've take it over from them.


EndangeredMassa (approved):
I don't know Python deeply, but this looks like it could work.

packages/python/vc_init.py
```diff
@@ -198,16 +199,25 @@ def __call__(self, app, body):
...
 
                 asgi_instance = app(self.scope, self.receive, self.send)
 
-                asgi_task = loop.create_task(asgi_instance)
-                loop.run_until_complete(asgi_task)
+                if _use_legacy_asyncio:
+                    asgi_task = loop.create_task(asgi_instance)
+                    loop.run_until_complete(asgi_task)
+                else:
+                    asyncio.run(self.run_asgi_instance(asgi_instance))
```
kytta, line 215: I've mentioned this in another PR fixing this issue: Since asgi_instance is a coroutine, I believe you don't have to wrap in a function

PR #11681, opened 2024-05-31
[ppr] Add missng `.rsc` route for non-ppr enabled pages

wyattjoh:
When deploying partial prerendering (PPR), there may some pages that are not enabled for PPR but still appear in the `prerender-manifest.json`. Due to the branching of the client router, these routes also have to have a `.rsc` as well as a `.prefetch.rsc` variants in order to prevent 404's. This change adds support for adding the extra route to the prerender for pages that have PPR disabled.

### DIFF
--- a/.changeset/real-moose-hope.md
+++ b/.changeset/real-moose-hope.md
@@ -1,0 +1,6 @@
+---
+'@vercel/next': patch
+'vercel': patch
+---
+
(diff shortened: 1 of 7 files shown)


ztanner (commented):
Could we add a probe test to confirm that the requests that were previously 404ing do not anymore?

ztanner (dismissed):

ztanner (approved):

PR #11686, opened 2024-06-03
[next]: ensure unmatched action rewrites are routed to correct handler

ztanner:
User defined rewrites are "normalized" so that our internal rewrites are still properly handled. Before normalizing these rewrites, the Next.js builder will attempt to match server action requests to a`.action` variant. Then the user-defined rewrites flow through the afterFiles normalization ([this part](https://github.com/vercel/vercel/blob/fix/unmatched-action-rewrites/packages/next/src/server-build.ts#L254-L279)) so that when we add `.action` in the builder, we don't drop the suffix. 

But this normalization can lead to a malformed `dest`. e.g., if I had rewrite like this:

```js
{
  source: '/greedy-rewrite/static/:path*',
  destination: '/static/:path*',
}
```

The builder would go through this flow on an action request to `/greedy-rewrite/static`:

1. It'll attempt to match it to a `.action` output, so `/greedy-rewrite/static` -> `/greedy-rewrite/static.action`
2. The afterFiles normalization will take place, so the original `dest` of `/static/$1` will become `/static/$1$rscsuff`
3. $1 will be an empty string, because it doesn't match the existing capture group. So now `/greedy-rewrite/static.action` -> `/greedy-rewrite/static/.action`
4. `static/.action` is not a valid output, so it'll 404 and the action will break. 

Existing handling exists for `.rsc` outputs for a similar reason, but only on the index route. I added a similar fix for this in #11688.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2089,6 +2089,19 @@ export async function serverBuild({
       // with that routing section
       ...afterFilesRewrites,
 
+      // Ensure that after we normalize `afterFilesRewrites`, unmatched actions are routed to the correct handler
+      // e.g. /foo/.action -> /foo.action. This should only ever match in cases where we're routing to an action handler
+      // and the rewrite normalization led to something like /foo/$1$rscsuff, and $1 had no match
+      ...(hasActionOutputSupport
+        ? [
+            {
+              src: `${path.posix.join('/', entryDirectory, '(.+)/\\.action$')}`,
+              dest: `${path.posix.join('/', entryDirectory, '$1.action')}`,
+              check: true,
+            },
+          ]
+        : []),
+
       // ensure non-normalized /.rsc from rewrites is handled
       ...(appPathRoutesManifest
         ? [
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2091,9 +2091,15 @@ export async function serverBuild({
 
       // Ensure that after we normalize `afterFilesRewrites`, unmatched actions are routed to the correct handler
       // e.g. /foo/.action -> /foo.action. This should only ever match in cases where we're routing to an action handler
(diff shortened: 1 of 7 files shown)


packages/next/src/server-build.ts
```diff
@@ -2089,6 +2089,19 @@ export async function serverBuild({
...
 
+      // Ensure that after we normalize `afterFilesRewrites`, unmatched actions are routed to the correct handler
+      // e.g. /foo/.action -> /foo.action. This should only ever match in cases where we're routing to an action handler
+      // and the rewrite normalization led to something like /foo/$1$rscsuff, and $1 had no match
+      ...(hasActionOutputSupport
+        ? [
+            {
+              src: `${path.posix.join('/', entryDirectory, '(.+)/\\.action$')}`,
+              dest: `${path.posix.join('/', entryDirectory, '$1.action')}`,
+              check: true,
```
ztanner, line 2100: Since Next.js doesn't emit any `.action` outputs and thus won't exist on the filesystem, would this make more sense? cc @ijjk 

```suggestion
              continue: true,
```

packages/next/src/server-build.ts
```diff
@@ -2089,6 +2089,25 @@ export async function serverBuild({
...
+      // This is meant to have parity with the .rsc handling below.
+      ...(hasActionOutputSupport
+        ? [
+            {
+              src: `${path.posix.join('/', entryDirectory, '/\\.action$')}`,
+              dest: `${path.posix.join('/', entryDirectory, '/index.action')}`,
+              check: true,
+            },
+            {
+              src: `${path.posix.join('/', entryDirectory, '(.+)/\\.action$')}`,
```
ijjk, line 2104:
```suggestion
              src: `${path.posix.join('/', entryDirectory, '/(.+)\\.action$')}`,
```
Same as this thread in case that one should be updated https://github.com/vercel/vercel/pull/11688/files#r1625164267

ijjk (approved):

PR #11688, opened 2024-06-03
[next]: ensure unmatched rsc rewrites are routed to correct handler

ztanner:
Same reasoning as:

- https://github.com/vercel/vercel/pull/11686

Split this into a separate PR so it can be tested/released separately.


ijjk (dismissed):

packages/next/src/server-build.ts
```diff
@@ -2101,11 +2101,29 @@ export async function serverBuild({
...
+              ),
+              check: true,
+            },
             {
               src: path.posix.join('/', entryDirectory, '/\\.rsc$'),
               dest: path.posix.join('/', entryDirectory, `/index.rsc`),
               check: true,
             },
+            {
+              src: path.posix.join('/', entryDirectory, '(.+)/\\.rsc$'),
```
ijjk, line 2142:
```suggestion
              src: path.posix.join('/', entryDirectory, '/(.+)\\.rsc$'),
```
Should the group match be after the `/`?

ijjk (approved):

PR #11697, opened 2024-06-04
[build-utils] Handle case of not having lockfile

onsclom:
In the scenario where we don't have a lockfile, cliType was detected by build utils as npm regardless of what's set in corepack. This fixes that.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -353,29 +353,9 @@ export async function scanParentDirs(
     // TODO: read "bun-lockfile-format-v0"
     lockfileVersion = 0;
   } else {
-    const packageJsonPackageManager = packageJson?.packageManager;
-    if (usingCorepack(process.env, packageJsonPackageManager)) {
-      const corepackPackageManager = validateVersionSpecifier(
-        packageJsonPackageManager as string
-      );
-      switch (corepackPackageManager?.packageName) {
-        case 'npm':
-        case 'pnpm':
-        case 'yarn':
-        case 'bun':
-          cliType = corepackPackageManager.packageName;
-          break;
-        case undefined:
-          cliType = 'npm';
-          break;
-        default:
-          throw new Error(
-            `Unknown package manager "${corepackPackageManager?.packageName}". Change your package.json "packageManager" field to a known package manager.`
-          );
-      }
-    } else {
-      cliType = 'npm';
-    }
+    cliType = packageJson
+      ? detectPackageManagerNameWithoutLockfile(packageJson)
+      : 'npm';
   }
 
   const packageJsonPath = pkgJsonPath || undefined;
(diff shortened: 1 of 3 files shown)


erikareads:
The cliType is determined in scanParentDirs (https://github.com/vercel/vercel/blob/main/packages/build-utils/src/fs/run-user-scripts.ts#L355) this function doesn't get passed in the env which makes it harder to detect whether corepack is enabled. this gives us a few options:

1. we can just use process.env inside the function
2. we can change the call signature to include env
3. we can change the call signature to include corepackEnabled or some boolean

in this PR, we chose to move forward with 1.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -353,7 +353,28 @@ export async function scanParentDirs(
...
+        packageJsonPackageManager
+      );
+      switch (corepackPackageManager?.packageName) {
+        case 'npm':
+        case 'pnpm':
+        case 'yarn':
+        case 'bun':
+          cliType = corepackPackageManager.packageName;
+          break;
+        default:
```
EndangeredMassa: **suggestion (soft-blocking):** if `packageName` is not `undefined`, but also not a known package manager, we should probably error

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -353,7 +353,28 @@ export async function scanParentDirs(
     // TODO: read "bun-lockfile-format-v0"
     lockfileVersion = 0;
   } else {
-    cliType = 'npm';
+    const packageJsonPackageManager = packageJson?.packageManager;
+    if (
+      packageJsonPackageManager &&
+      usingCorepack(process.env, packageJsonPackageManager)
```
EndangeredMassa: **suggestion (non-blocking):** I don't think we need to check `packageJsonPackageManager` here. `usingCorepack` seems to have that case handled.‏

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -353,7 +353,28 @@ export async function scanParentDirs(
...
+        case 'bun':
+          cliType = corepackPackageManager.packageName;
+          break;
+        default:
+          cliType = 'npm';
+          break;
+      }
+    } else {
+      cliType = 'npm';
+    }
```
EndangeredMassa: **suggestion (non-blocking):** ‏extract function

This looks like a good candidate for function extraction to improve legibility of this otherwise very long function.

EndangeredMassa (dismissed):

trek (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -366,6 +368,37 @@ export async function scanParentDirs(
...
+      case 'npm':
+      case 'pnpm':
+      case 'yarn':
+      case 'bun':
+        return corepackPackageManager.packageName;
+      case undefined:
+        return 'npm';
+      default:
+        throw new Error(
+          `Unknown package manager "${corepackPackageManager?.packageName}". Change your package.json "packageManager" field to a known package manager.`
```
EndangeredMassa, line 387: **suggestion (soft-blocking):** mention the known package managers

Errors like these are useful, but they can be more so if you tell the user what the valid values are in the message.

```suggestion
          `Unknown package manager "${corepackPackageManager?.packageName}". Change your package.json "packageManager" field to a known package manager: npm, pnpm, yarn, bun.`
```

This does make the list a bit brittle, though. We could extract it further into a single source of truth for that list, but I don't think we need to go that far right here.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -366,6 +368,37 @@ export async function scanParentDirs(
   };
 }
 
+function detectPackageManagerNameWithoutLockfile(packageJson: PackageJson) {
+  const packageJsonPackageManager = packageJson.packageManager;
+  if (usingCorepack(process.env, packageJsonPackageManager)) {
+    const corepackPackageManager = validateVersionSpecifier(
+      packageJsonPackageManager as string
```
EndangeredMassa, line 375: **suggestion (soft-blocking):** tweak types

Oh, this was complaining because of type checking before? I think we can have `validateVersionSpecifier` accept a `string | undefined` instead. I prefer not to use `as` anywhere, if possible. It damages the type safety of the program--in a small way, but still non-zero.

EndangeredMassa (approved):

PR #11701, opened 2024-06-06
[functions] Add OIDC utils to @vercel/functions

marcgreenstock:
This PR adds some new methods to `@vercel/functions`:

* `getVercelOidcToken()` which returns the token from the request object or environment variables.
* `awsCredentialsProvider()` which returns an AWS credential provider.


packages/functions/package.json
```diff
@@ -2,9 +2,11 @@
   "name": "@vercel/functions",
   "description": "Runtime functions to be used with your Vercel Functions",
   "homepage": "https://vercel.com",
-  "version": "1.0.2",
-  "types": "index.d.ts",
-  "main": "index.js",
+  "version": "1.1.0",
```
EndangeredMassa: **suggestion (blocking):** ‏revert version change

```suggestion
  "version": "1.0.2",
```

We use changesets to manage the version. Can we revert this change in this PR?

packages/functions/package.json
```diff
@@ -14,17 +16,28 @@
...
     "node": ">= 16"
   },
-  "files": [
-    "index.d.ts",
-    "index.js"
-  ],
   "scripts": {
-    "test": "vitest"
+    "build": "node ../../utils/build.mjs",
+    "test": "jest --reporters=default --reporters=jest-junit --env node --verbose --runInBand --bail",
```
EndangeredMassa: **suggestion (blocking):** ‏We're moving tests over to `vitest`. Can we revert this change?

EndangeredMassa (changes requested):

packages/functions/package.json
marcgreenstock: Switch back to `vitest` in https://github.com/vercel/vercel/pull/11701/commits/1b9c4e261833e53a9af85d74a6a590099b18bb1a

EndangeredMassa (dismissed):
This makes more sense to me now. I don't know for sure that we want this in `@vercel/functions` vs. something else, but if we do want it here, sounds good.

Schniz:
do we need to have `exports` in package.json?

packages/functions/src/oidc/awsCredentialsProvider.ts
```diff
@@ -0,0 +1,59 @@
...
+ *     durationSeconds: 7200
+ *   }),
+ * });
+ * ```
+ */
+export function awsCredentialsProvider(
+  init: AwsCredentialsProviderInit
+): AwsCredentialIdentityProvider {
+  return async () => {
+    const { fromWebToken } = await import(
```
Kikobeats: I guess you are doing this lazily to avoid to impact execution time if you don't need this method?

packages/functions/src/oidc/awsCredentialsProvider.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Obtains the Vercel OIDC token and creates an AWS credential provider function
+ * that gets AWS credentials calling STS AssumeRoleWithWebIdentity API.
+ *
+ * ```javascript
+ * import * as s3 from '@aws-sdk/client-s3';
+ * import { awsCredentialsProvider } from '@vercel/functions/oidc';
+ *
+ * const s3Client = new s3.S3Client({
```
Kikobeats: why not use `@example` and the rest of jsdocs stuff 🙂

packages/functions/src/oidc/awsCredentialsProvider.ts
marcgreenstock: Correct.

packages/functions/src/waitUntil.ts
```diff
@@ -0,0 +1,35 @@
...
+ * ```
+ * import { waitUntil } from '@vercel/functions';
+ *
+ * export function GET(request) {
+ *   waitUntil(fetch('https://vercel.com'));
+ *   return new Response('OK');
+ * }
+ * ```
+ */
+export const waitUntil = (promise: Promise<any>) => {
```
Kikobeats:
```suggestion
export const waitUntil = (promise: Promise<unknown>) => {
```

packages/functions/src/oidc/awsCredentialsProvider.ts
marcgreenstock: I made the docs on this as familiar as possible, it matches the same style as the AWS SDK jsdoc comments

packages/functions/package.json
```diff
@@ -3,8 +3,10 @@
   "description": "Runtime functions to be used with your Vercel Functions",
   "homepage": "https://vercel.com",
   "version": "1.0.2",
-  "types": "index.d.ts",
```
Kikobeats: maybe you still need to expose types for the transpiled typescript code?

packages/functions/src/oidc/awsCredentialsProvider.ts
Kikobeats: it could be both, in this way https://github.com/vercel/vercel/pull/11701/files#diff-36e729cab01b983624e1459052aaf4d59277379ca0e4567322f6a808ba4cd66cR6

Kikobeats (dismissed):

package.json
```diff
@@ -34,7 +34,7 @@
     "ts-eager": "2.0.2",
     "ts-jest": "29.1.0",
     "turbo": "1.13.3",
-    "typescript": "4.9.5"
+    "typescript": "5.5.2"
```
EndangeredMassa: **suggestion (blocking):** ‏revert

We try to keep the typescript version consistent across the monorepo. It's not 100% consistent right now, but that is an explicit goal.

Screenshot 2024-06-28 at 10 32 50 AM

If we need to make this update, can do we it across the whole monorepo in a separate PR? Happy to fast track reviews on that if tests pass.

packages/functions/jest.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('@ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+};
```
EndangeredMassa: **suggestion (non-blocking):** ‏remove

We're using vitest in this package.

packages/functions/package.json
```diff
@@ -14,17 +17,27 @@
     "url": "https://github.com/vercel/vercel/issues"
   },
   "devDependencies": {
+    "@aws-sdk/client-s3": "3.590.0",
+    "@aws-sdk/credential-provider-web-identity": "3.587.0",
+    "@smithy/types": "3.0.0",
+    "typescript": "5.4.5",
```
EndangeredMassa: Same as: https://github.com/vercel/vercel/pull/11701#discussion_r1658926746

packages/functions/package.json
```diff
@@ -14,17 +17,27 @@
...
     "vitest": "1.3.1"
   },
+  "peerDependencies": {
+    "@aws-sdk/credential-provider-web-identity": "*"
```
EndangeredMassa, line 40: **question (soft-blocking):** ‏can you explain why we need this and why it needs to be any version (`*`)?

It it in service of the dynamic import? https://github.com/vercel/vercel/pull/11701/files#diff-25c8d9ab60649f508457d1f2005f8cbab25ea20edeced6b857e6a23bdc5c9b69R54

packages/functions/test/oidc/awsCredentialsProvider.test.ts
```diff
@@ -0,0 +1,48 @@
+import { expect, it, vi } from 'vitest';
+import { awsCredentialsProvider } from '../../src/oidc';
+
```
EndangeredMassa: **suggestion (soft-blocking):** add single root `describe`

Can you wrap all of this in a single, root `describe`?

packages/functions/test/oidc/getVercelOidcToken.test.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+import { getVercelOidcToken } from '../../src/oidc';
+import { randomUUID } from 'crypto';
+
```
EndangeredMassa: **suggestion (soft-blocking):** add single root `describe`

Can you wrap all of this in a single, root `describe`?

packages/functions/test/waitUntil.test.ts
```diff
@@ -26,15 +24,23 @@ test.each([null, undefined, {}])(
...
     expect(() => waitUntil(promise)).not.toThrow();
   }
 );
 
 test('waitUntil calls ctx.waitUntil when available', async () => {
   const promise = Promise.resolve();
   const waitUntilMock = vi.fn().mockReturnValue(promise);
-  globalThis[Symbol.for('@vercel/request-context')] = {
+  globalThis[
+    // @ts-ignore
```
EndangeredMassa: **suggestion (non-blocking):** ‏can you add to the comment what/why we're ignoring?

EndangeredMassa (changes requested):
Only blocking issue is that the TypeScript dependency update should be reverted. Everything else would be great in either this PR or a follow-up PR.

EndangeredMassa:
For the tests: I increased the rate limit that these tests were hitting. We had just moved a bunch of tests over to a single Vercel account, which blew past a couple rate limits. Let me know if that happens again.

EndangeredMassa:
We're still debugging rate limit issues for tests in this repo.

packages/functions/test/waitUntil.test.ts
Schniz: it would probably be better to `// @ts-expect-error` instead of ignoring so if the error isn't happening anymore we will be prompted to remove this line

EndangeredMassa:
> We're still debugging rate limit issues for tests in this repo.

CI is good, now.

packages/functions/package.json
marcgreenstock: @EndangeredMassa Yes, it is in service of the dynamic import.

The AWS SDK imports the same package using `*`: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-node/package.json#L36

If they change this in the future we may need to consider changing it ourselves, but this seems like the safest, future proof path for now.

marcgreenstock:
> do we need to have `exports` in package.json?

Yes it turns out we do, I've added them on 064f432e7a727ef0dd868d6e60a2742e5c844f5b

packages/functions/package.json
```diff
@@ -19,10 +26,20 @@
     "url": "https://github.com/vercel/vercel/issues"
   },
   "devDependencies": {
-    "typescript": "4.9.5",
```
EndangeredMassa, line 22: ?

EndangeredMassa (dismissed):
Thanks!

TooTallNate (dismissed):

Kikobeats (approved):

trek (approved):

EndangeredMassa (approved):

PR #11707, opened 2024-06-06
[next]: fix issues with rewrite normalization of index routes

ztanner:
The Next.js builder currently normalizes user rewrites to ensure the internal `.rsc` / `.action` rewrites are still handled even after following the rewrite.

However we have special rewrites: `/index.rsc` and `/index.action`. When those routes pass through the normalization logic, a request routed to a catch-all route (e.g. `[[...param]]`) will go from `/` -> `/index.rsc` -> `<someRewriteString>?param=index`. This is inconsistent `next start` or other spots where the param should be an empty string.

Similarly, if a user rewrites their entire app to a subdirectory (eg `/:path*` -> `/base/:path*`, a route will go from `/` -> `/index.rsc` -> `/base/index.rsc` which won't exist, causing a 404. 

The solution here is to return the `index.rsc` / `index.action` routes back to `/` so they can be handled by user rewrites, if necessary. 

This also disables the `hasActionOutputSupport` flag if `routesManifest.i18n` is detected as they are not compatible.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2090,7 +2090,7 @@ export async function serverBuild({
       ...(afterFilesRewrites.length > 0
         ? [
             {
-              src: path.posix.join('/', entryDirectory, '/index(.action|.rsc)'),
+              src: path.posix.join('/', entryDirectory, '/index(\\.action|\\.rsc)'),
               dest: path.posix.join('/', entryDirectory),
               continue: true,
             },
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2088,7 +2088,7 @@ export async function serverBuild({
 
       // before processing rewrites, remove any special `/index` routes that were added
       // as these won't be properly normalized by `afterFilesRewrites`
-      ...(afterFilesRewrites.length > 0
+      ...(appPathRoutesManifest && afterFilesRewrites.length > 0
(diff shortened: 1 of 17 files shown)


packages/next/src/server-build.ts
```diff
@@ -2085,6 +2085,18 @@ export async function serverBuild({
           ]
         : []),
 
+      // before processing rewrites, remove any special `/index` routes that were added
+      // as these won't be properly normalized by `afterFilesRewrites`
+      ...(afterFilesRewrites.length > 0
+        ? [
+            {
+              src: path.posix.join('/', entryDirectory, '/index(.action|.rsc)'),
```
```suggestion
              src: path.posix.join('/', entryDirectory, '/index(\\.action|\\.rsc)'),
```
should we escape these so they aren't treated as greedy matches

ijjk (dismissed):

packages/next/src/server-build.ts
```diff
@@ -2085,6 +2086,22 @@ export async function serverBuild({
           ]
         : []),
 
+      // before processing rewrites, remove any special `/index` routes that were added
+      // as these won't be properly normalized by `afterFilesRewrites`
+      ...(afterFilesRewrites.length > 0
```
```suggestion
      ...(appPathRoutesManifest && afterFilesRewrites.length > 0
```

ijjk (dismissed):

ijjk (approved):

PR #11732, opened 2024-06-15
Update Turborepo logos.

anthonyshew:
Title!


.changeset/early-rice-tell.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
jeffsee55, line 2: I'm not aware of where these logos are used, is it ok that we're skipping a version bump in this case?

trek (approved):

.changeset/early-rice-tell.md
trek: I believe so!

PR #11738, opened 2024-06-17
[node] ignore `shouldAddHelpers` when exporting a server to match production

QuiiBz:
Follow-up to https://x.com/maggiepint/status/1801975947421442137, we have a difference between production and development (`vc dev`) regarding `shouldAddHelpers`:
- In production, if the user exports a server (e.g. express), we don't add helpers
- In development, we always add helpers

That means we set/override things like `req.body` in development, but don't in production. Changing the logic in production can trigger unexpected behaviors and break existing applications, so we should instead fix the CLI to have the same behavior as production.


trek (dismissed):

TooTallNate (commented):
Change looks fine, but we should add a test.

trek (approved):

PR #11742, opened 2024-06-17
Include node version range in node engine warning message

jeffsee55:
Before this change the warning doesn't show what the resolved version of node actually is. Which is slightly confusing when you have an engine rule like `>=18` and the resolved version is `20x`.

```
Warning: Due to "engines": { "node": ">=18" } in your `package.json` file, the Node.js Version defined in your Project Settings ("18.x") will not apply, Node.js Version "20.x" will be used instead. Learn More: http://vercel.link/node-version
```


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -259,6 +259,7 @@ export async function getNodeVersion(
     !packageJsonVersion,
     availableVersions
   );
+  console.log({ supportedNodeVersion });
```
TooTallNate: Remove?

```suggestion
```

packages/remix/defaults/server-node.mjs
```diff
@@ -4,7 +4,7 @@ import {
   installGlobals,
 } from '@remix-run/node';
 
-installGlobals();
+// installGlobals();
```
TooTallNate, line 7: Remove?

TooTallNate: Revert?

```suggestion
installGlobals();
```

TooTallNate (dismissed):

packages/remix/defaults/server-node.mjs
jeffsee55: Sorry!

erikareads (approved):

PR #11745, opened 2024-06-18
[cli] allow prerender to stream

craigandrews:
Allow all app router lambdas to use streaming, including prerender and experimental PPR.

Practically speaking this is a no-op at the moment; `api-builds` has a feature-flagged check to ensure any prerender function will not stream. It is effectively a duplicate of the check that this PR removes.

This is the check in `api-builds`: https://github.com/vercel/api/blob/987de85d6a38a34f717cd1372826cbaf6113dd7d/services/api-builds/src/endpoints/post-lambda.ts#L273

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -948,10 +948,7 @@ export async function serverBuild({
     const appRouterStreamingActionLambdaGroups: LambdaGroup[] = [];
 
     for (const group of appRouterLambdaGroups) {
-      if (!group.isPrerenders || group.isExperimentalPPR) {
-        group.isStreaming = true;
-      }
-
+      group.isStreaming = true;
       group.isAppRouter = true;
 
       // We create a streaming variant of the Prerender lambda group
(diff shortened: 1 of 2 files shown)


packages/next/src/server-build.ts
```diff
@@ -948,10 +948,7 @@ export async function serverBuild({
     const appRouterStreamingActionLambdaGroups: LambdaGroup[] = [];
 
     for (const group of appRouterLambdaGroups) {
-      if (!group.isPrerenders || group.isExperimentalPPR) {
-        group.isStreaming = true;
-      }
-
+      group.isStreaming = true;
```
ijjk, line 951: We probably want to group this under the same flag/env we're using upstream to allow this

ijjk (approved):

packages/next/src/server-build.ts
craigandrews: Discussed offline, this is fine

PR #11748, opened 2024-06-18
[remix] Use native fetch polyfills for Node 20+

TooTallNate:
Allows for the native `fetch` in Node.js to be used instead of the `node-fetch` that Remix uses by default. Currently opt-in via an env var.

### DIFF
--- a/.changeset/tall-foxes-attack.md
+++ b/.changeset/tall-foxes-attack.md
@@ -1,0 +1,5 @@
+---
+'@vercel/remix-builder': patch
+---
+
+Add opt-in env var to use native Fetch polyfills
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
Looks good.

Is there any scenario where a developer should use `installGlobals` in their vite config? In general we may want to update our [example](https://github.com/vercel/vercel/blob/f0d6acaa03f20b47dbaf9ff501d009e02901db86/examples/remix/vite.config.ts#L7) and our [docs](https://vercel.com/docs/frameworks/remix#vercel-vite-preset) that this isn't necessary.

TooTallNate:
You do want it in the vite config, but I believe it's only relevant for dev mode.

jeffsee55 (approved):

PR #11766, opened 2024-06-25
Move integration tests to Zero config team account

jeffsee55:
After experiencing CI failures for our test account, it became evident that tests were being run against some older infrastructure. This PR moves the tests to a new account, and we should now be able to inspect build and runtime logs more easily for test failures

One of the biggest ramifications of this change is that many of our integration tests were operating under the assumption that we wanted to deploy to the default team for the token that was being used. However, there's no user whose default team is the newly created test account, so we've made some changes to explicitly pass the scope in all of our integration tests.


packages/cli/test/helpers/exec.ts
```diff
@@ -1,51 +1,40 @@
...
+  if (!args.includes('--token') && opts.token !== false) {
+    args.push(
+      '--token',
+      typeof opts.token === 'string' ? opts.token : process.env.VERCEL_TOKEN!
+    );
+  }
+
+  if (!args.includes('--scope')) {
+    args.push('--scope', process.env.VERCEL_TEAM_ID!);
+  }
```
jeffsee55, line 27: This is probably the most impactful change we're making, the token and team scope are no longer inferred from the CLI's currently authenticated user.

packages/cli/test/integration-2.test.ts
```diff
@@ -967,6 +805,12 @@ test('deploy pnpm twice using pnp and symlink=false', async () => {
   text = await page.text();
 
   expect(text).toContain('cache exists\n');
+
+  // Since this test asserts that we can create a new project based on the folder name, delete it after the test
+  // to avoid polluting the project list.
+  await apiFetch(`/projects/${session}`, {
+    method: 'DELETE',
+  });
```
jeffsee55, line 813: There's remaining work to be done to ensure that tests which create new projects on every run are also removing those projects as part of their cleanup, this issue is mostly ignored in the scope of this PR but here's an example of the kind of thing we'll need to do more of.

jeffsee55 (commented):
Some comments

erikareads (approved):

packages/cli/src/commands/project/list.ts
```diff
@@ -27,7 +27,7 @@ export default async function list(
 
   output.spinner(`Fetching projects in ${chalk.bold(contextName)}`);
 
-  let projectsUrl = `/v4/projects/?limit=20`;
+  let projectsUrl = `/v9/projects?limit=20`;
```
erikareads, line 30: this is quite the api version bump

examples/remix/package.json
```diff
@@ -9,18 +9,18 @@
     "typecheck": "tsc"
   },
   "dependencies": {
-    "@remix-run/node": "^2.9.2",
-    "@remix-run/react": "^2.9.2",
-    "@remix-run/server-runtime": "^2.9.2",
+    "@remix-run/node": "2.10.0",
```
trek, line 12: Why'd we need to bump this? #not-blocking

trek (approved):

packages/cli/src/commands/project/list.ts
TooTallNate: A lengthy explanation 🙃

The `vc domain inspect` function was using an even older version of the endpoint (v2) which was not a paginated version of the endpoint, yet that command did its own version of pagination based on the most recent project's `updatedAt` field.

At some point, this endpoint (all versions) stopped sorting results upon the `updatedAt` field (specifically when the `domain` query param was used, which was indeed the case for `domain inspect`). This caused the pagination code in the CLI command to enter an endless loop, because it would keep jumping around with various timestamps because it was seeing duplicate results. We never saw this in the tests before because the issue would only happen for a team with > 50 accounts, which was never the case with the previous test accounts.

On top of that, the `domain` query param does even work for the GET projects anymore (all versions), so there's no way to get the API to return the projects specific to a domain name. Thus, the `domain` param was removed from CLI (so that the results get sorted by `updatedAt` so that pagination works), and the filtering by relevant domain logic was implemented client-side.

In the case of this file, `vc project list` was also updated to use the same version of the API endpoint for consistency.

examples/remix/package.json
TooTallNate: The template is failing on main. Fix could have been in a separate PR, but I pushed to this branch instead because I figured merge here was imminent.

TooTallNate (approved):

PR #11776, opened 2024-06-28
[redwood] Support API endpoint configuration via `export const config`

TooTallNate:
Uses the `@vercel/static-config` module to get the `export const config` configuration for Redwood.js API endpoints, so it's the same interface as vanilla JS endpoints, and Remix.

The previous usage of `getLambdaOptionsFromFunction` was broken because it would cause the `@vercel/redwood` Builder to be executed multiple times (once for each endpoint).


packages/redwood/src/index.ts
```diff
@@ -5,7 +5,10 @@ import {
   parse as parsePath,
   sep,
   basename as pathBasename,
+  basename,
```
jeffsee55, line 7: Looks like this is already imported as `pathBasename`

TooTallNate: Oh good catch.

jeffsee55 (approved):

PR #11783, opened 2024-07-01
[functions] add `getEnv` method

Kikobeats:
The new method `getEnv()` makes easy to interact with [Vercel System Environment Variables](https://vercel.com/docs/projects/environment-variables/system-environment-variables#system-environment-variables-overview)

CleanShot 2024-07-08 at 19 37 42@2x


packages/functions/src/system-environment-variables.ts
```diff
@@ -0,0 +1,188 @@
+export const systemEnvironmentVariables = (
```
Schniz: maybe `getEnv`? <picture data-single-emoji=":eye-twitch:" title=":eye-twitch:">:eye-twitch:</picture>

Schniz (dismissed):
<picture data-single-emoji=":lgtm:" title=":lgtm:">:lgtm:</picture>

i just wonder if the name of the function is good or we should make it shorter

no strong feelings for discrete functions vs returning a big "env var" record

packages/functions/src/system-environment-variables.ts
Kikobeats: `systemEnv` ?

Schniz (dismissed):
<picture data-single-emoji=":lfgtm-intense:" title=":lfgtm-intense:">:lfgtm-intense:</picture>

Schniz (approved):

PR #11788, opened 2024-07-03
feat(cli)!: vc logs now returns runtime logs

feugy:
### 🧐 What's in there?

This is the second part of distinguishing build logs from runtime logs.
- `vc logs`: read the newly emitted **runtime** logs

(Pretty print format discussed internally with the logs team and our wonderful designer @timolins)
image

### 🧪 How to test?

Given you're logged into Vercel organization and have our front application codebase:
- `pnpm -F vercel dev logs --cwd PATH_TO_FRONT/apps/vercel-site https://vercel.com` to see up to 5 minutes of newly emitted logs
- `pnpm -F vercel dev logs --cwd PATH_TO_FRONT/apps/vercel-site https://vercel.com --json` to see these logs printed in JSON, so they could be piped into `jq`

Unit tests added:
- inspect command: `pnpm -F vercel exec vitest run /logs.test`


packages/cli/src/commands/logs/command.ts
```diff
@@ -1,9 +1,13 @@
...
 import { packageName } from '../../util/pkg-name';
 
+// has to be ms compliant
+// https://github.com/vercel/ms/blob/fe5338229cfdac6822891dcb9c24660b4d2e612b/src/index.ts#L95
+export const CommandTimeout = '5 minutes';
+
 export const logsCommand: Command = {
   name: 'logs',
-  description: 'Display logs for a specific deployment.',
+  description: `Display runtime logs for a specific deployment, if it is live, from now and for ${CommandTimeout} at most.`,
```
Jooliashi: **question**, since we are always following, wouldn't it be always live?

feugy: Maybe the sentence was unclear. "being live" was about the deployement.
I can change to "if it is ready" maybe?

Jooliashi: hmm, are we trying to say that it will only display runtime logs if it is live?

packages/cli/src/util/logs.ts
```diff
@@ -30,11 +38,217 @@ export function displayBuildLogs(
...
+  const date = new Date(log.created).toISOString();
+
+  data.split('\n').forEach(line => {
+    print(`${chalk.dim(date)}  ${line.replace('[now-builder-debug] ', '')}\n`);
+  });
+}
+
+function isRuntimeLimitDelimiter(log: RuntimeLog) {
+  return (
+    log.rowId === '' && log.level === 'error' && log.source === 'delimiter'
```
Jooliashi, line 198: **question**: wouldn't `delimiter` source be sufficient check here

feugy: I don't think so. We may have other delimiters, it doesn't harm being more specific in this case, to match [our API endpoint convention](https://github.com/vercel/api/pull/29618/files#diff-ee9519d82b888d4d4fd7e3b14ef4cd0b6f0d84d71fd81399cfd9417040de7ad1R58-R71)

Jooliashi (dismissed):

packages/cli/src/commands/logs/index.ts
```diff
@@ -1,263 +1,113 @@
...
-function toTimestamp(datestr: string) {
-  const t = Date.parse(datestr);
-  if (isNaN(t)) {
-    throw new TypeError('Invalid date string');
-  }
-  return t;
+function printDisclaimer(deployment: Deployment, { print, warn }: Output) {
+  // Could be temporary until users get used to this change
+  warn(
+    `This command now display runtime logs. To access your build logs, run \`vercel inspect --logs ${deployment.url}\``
```
jeffsee55:
```suggestion
    `This command now displays runtime logs. To access your build logs, run \`vercel inspect --logs ${deployment.url}\``
```

packages/cli/src/util/events.ts
```diff
@@ -182,7 +182,7 @@ async function printEvents(
         if (err instanceof Error && err.name === 'AbortError') {
           return;
         }
-        throw error;
+        throw err;
```
jeffsee55, line 185: Good find!

packages/cli/src/commands/logs/command.ts
```diff
@@ -12,50 +16,25 @@ export const logsCommand: Command = {
   ],
   options: [
     {
-      name: 'follow',
```
jeffsee55, line 15: Instead of removing these outright, can you mark them as deprecated so we are able to link a helpful message?

As an example, we do something similar in the [`deploy` command](https://github.com/vercel/vercel/blob/c75415ae3c46cf62790c2541ef0d602e3c534dad/packages/cli/src/commands/deploy/index.ts#L167-L176). 

We'll likely need a changelog that talks about the major version bump somewhere, which this message could link to.

packages/cli/src/util/logs.ts
```diff
@@ -30,11 +38,217 @@ export function displayBuildLogs(
...
+
+    const handleData = (data: RuntimeLog | string) => {
+      let log: RuntimeLog = parse ? data : JSON.parse(data as string);
+      if (isRuntimeLimitDelimiter(log)) {
+        abortController.abort();
+        warn(`${chalk.bold(log.message)}\n`);
+        return;
+      }
+      // eslint-disable-next-line no-console -- we intent to write unparsed logs to stdout so JQ could read them
+      parse ? prettyPrintLogline(log, print) : console.log(data);
```
jeffsee55: The intent of that eslint rule is so that `client.output.log` is used here instead of `console.log`

jeffsee55 (changes requested):

packages/cli/src/util/logs.ts
feugy: I wish we could use `client.output.log`, but this will print to the wrong stream. We need the data to be printed on stdout (so it could be piped into `jq` and other unix tools), but `client.output [is configured to print on stderr](https://github.com/vercel/vercel/blob/damien/log-457-runtime-logs-on-vc/packages/cli/src/index.ts#L110), regardless of the level used.

jeffsee55: Ah ok I see, we also have `client.stoud`, we mainly want to keep this scoped to the client so we can make changes in the future more easily.

feugy: Done!! Thanks for the tip

jeffsee55 (approved):

PR #11790, opened 2024-07-03
[cli] Add `vc target ls` command

TooTallNate:
Screenshot 2024-05-07 at 4 36 10 PM


packages/cli/src/commands/target/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  const args = argv._.slice(1);
+  const { cwd, output } = client;
+  const link = await getLinkedProject(client, cwd);
+
+  if (link.status === 'error') {
+    return link.exitCode;
+  }
+
+  if (link.status === 'not_linked') {
+    // TODO: prompt for link
```
onsclom, line 39: Non-blocking: Do we still want to come back to this, or is this just for future reference?

packages/cli/src/commands/target/list.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  output.spinner(`Fetching custom environments for ${projectSlugLink}`);
+
+  let url = `/projects/${encodeURIComponent(
+    link.project.id
+  )}/custom-environments?limit=20`;
+
+  const next = argv['--next'] || false;
+  if (next) {
+    url += `&until=${next}`;
```
onsclom: Blocking: Make sure to test this when pagination is added.

packages/cli/src/commands/target/list.ts
```diff
@@ -0,0 +1,136 @@
...
+    type: 'development',
+    description: '',
+    name: 'Development',
+    domains: [],
+  });
+
+  output.log(
+    `${
+      result.length > 0
+        ? `${result.length} Custom Environments`
```
onsclom: Non-blocking: Handle plural in case of 1 custom environment.

packages/cli/src/commands/target/list.ts
```diff
@@ -0,0 +1,136 @@
...
+  result.push({
+    id: 'development',
+    slug: 'development',
+    createdAt: 0,
+    updatedAt: 0,
+    type: 'development',
+    description: '',
+    name: 'Development',
+    domains: [],
+  });
```
onsclom: Non-blocking: 

1. Extract into function 

2. Change to make ordering more clear.
```
return [
  ...staticEnvironments,
  ...customEnvironments,
  ...developmentEnvironments
];
```

packages/cli/test/unit/commands/target.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      expect(line.value).toEqual('Retrieving project…');
+
+      line = await lines.next();
+      expect(line.value).toEqual(
+        `Fetching custom environments for ${team.slug}/${project.name}`
+      );
+
+      line = await lines.next();
+      expect(line.value).contains(
+        `> 5 Custom Environments found under ${team.slug}/${project.name}`
```
onsclom: Blocking: Maybe it makes more sense to say "2 Custom Environments" or "5 Environments"?

onsclom (changes requested):

packages/cli/src/commands/target/index.ts
onsclom: Made a card for this feature.

packages/cli/src/commands/target/list.ts
onsclom: Since there will be hard limits on the amount of custom environments users can have, we are not implementing pagination yet. Removed this pagination code.

packages/cli/src/commands/index.ts
```diff
@@ -39,3 +39,10 @@ export default new Map([
   ['teams', 'teams'],
   ['whoami', 'whoami'],
 ]);
+
+if (process.env.CUSTOM_ENVIRONMENTS) {
```
EndangeredMassa, line 43: **suggestion (soft-blocking):** add comment

Can we add a comment here explaining why this is here? This is temporary, correct?

EndangeredMassa (approved):

trek (approved):

onsclom (approved):

PR #11802, opened 2024-07-07
[functions] Add `ipAddress` & `geolocation`

Kikobeats:
Ported from https://github.com/vercel/vercel/blob/main/packages/edge/src/edge-headers.ts

These methods are not exclusively from edge anymore, they are isomorphic!

Also, added docs generation.


ecklf (approved):

packages/functions/test/integration/docs.test.ts
```diff
@@ -0,0 +1,30 @@
+import { expect, test } from 'vitest';
+import $ from 'tinyspawn';
+import path from 'path';
+
+(process.platform === 'win32' ? test.skip : test)(
+  'docs are up to date',
+  async () => {
+    const cwd = path.resolve(__dirname, '../');
+    await $('pnpm build:docs', { cwd });
+    const { stdout } = await $(`git status --short docs`);
```
ecklf, line 10: This only shows the current changes / working tree. Wouldn't this be always be empty when running in CI?

Kikobeats: I tested manually and if you forgot to commit docs it will throw an error!

maxkostow:
Is this intended to work with node `IncomingMessage` based next functions? Those do not have `get` defined on the `headers` object

wyattjoh:
This will work with Next.js's App Router based routes (`route.ts`), but not ones based on `IncomingMessage` no.

PR #11807, opened 2024-07-08
[next] Add build callback handling

ijjk:
This adds an optional new build callback for additional handling from outside the builder that allows additional work before the rest of the builder output handling is done.


packages/build-utils/src/types.ts
```diff
@@ -108,6 +108,8 @@ export interface BuildOptions {
    * on the build environment.
    */
   meta?: Meta;
+
+  buildCallback?: (opts: Omit<BuildOptions, 'buildCallback'>) => Promise<void>;
```
ztanner, line 117: can we add a comment explaining this property to match the other types here?

ztanner (approved):

PR #11821, opened 2024-07-09
[cli] Add `target` to `vc inspect` output

onsclom:
`vc inspect` did not output target before. This adds target output for existing "staging" and "production" values and custom environment names.

image


packages/cli/src/commands/inspect/index.ts
```diff
@@ -192,6 +192,23 @@ async function printDetails({
   print(chalk.bold('  General\n\n'));
   print(`    ${chalk.cyan('id')}\t\t${id}\n`);
   print(`    ${chalk.cyan('name')}\t${name}\n`);
+  // @ts-ignore - customEnvironment needs to be defined on Deployment
+  const customEnvironmentName = deployment.customEnvironment?.name;
+  const target = customEnvironmentName ?? deployment.target;
```
onsclom: `deployment.target` has type `"staging" | "production" | "null" | "undefined"`. Even after spending time digging into this, I'm still unsure what `null` or `undefined` values mean.

If the deployment doesn't have a custom environment name, and `target` is `null` or `undefined`, should it default to displaying something?

trek: I think displaying nothing is fine here. Pretty edge-casey. 

👋 hello future someone reading this wondering why we didn't handle this case!

trek (dismissed):

packages/cli/src/commands/inspect/index.ts
TooTallNate: `null` means "preview", so we should special-case that. `undefined` I think does not ever happen, but probably fine to default that to "preview" as well.

packages/cli/src/commands/inspect/index.ts
```diff
@@ -192,6 +192,23 @@ async function printDetails({
...
+  if (target) {
+    print(`    ${chalk.cyan('target')}\t`);
+    // TODO: once custom environments is shipped for all users,
+    // make all deployments link to the environment settings page
+    print(
+      // @ts-ignore - customEnvironment needs to be defined on Deployment
+      customEnvironmentName && deployment.team?.slug
+        ? `${link(
```
TooTallNate: I think the consensus we had regarding links is that if the terminal does not support them, don't render the link. That is to say, we should add a fallback (2nd param of the `link()` function) which returns only the name of the target.

TooTallNate:
Let's add a unit test please.

packages/cli/src/commands/inspect/index.ts
```diff
@@ -192,6 +192,24 @@ async function printDetails({
...
+            `${target}`,
+            `https://vercel.com/${deployment.team.slug}/${name}/settings/environments`,
```
TooTallNate: I think we should link to the specific target's page.

packages/cli/src/commands/inspect/index.ts
```diff
@@ -204,7 +204,8 @@ async function printDetails({
       customEnvironmentName && deployment.team?.slug
         ? `${link(
             `${target}`,
-            `https://vercel.com/${deployment.team.slug}/${name}/settings/environments`,
+            // @ts-ignore - customEnvironment needs to be defined on Deployment
```
TooTallNate: I would suggest `@ts-expect-error` instead. But also, can we add the type to `Deployment` so it's not necessary?

packages/cli/src/commands/inspect/index.ts
```diff
@@ -192,6 +192,22 @@ async function printDetails({
   print(chalk.bold('  General\n\n'));
   print(`    ${chalk.cyan('id')}\t\t${id}\n`);
   print(`    ${chalk.cyan('name')}\t${name}\n`);
+  const customEnvironmentName = deployment.customEnvironment?.name;
+  const target = customEnvironmentName ?? deployment.target ?? 'preview';
+  if (target) {
```
TooTallNate: This will always be truthy

TooTallNate (approved):

PR #11830, opened 2024-07-10
[next] Ensure we do not include ending slash in matched path

ijjk:
We don't expect a trailing slash to be included in this header so don't include it when matching. 

x-ref: [slack thread](https://vercel.slack.com/archives/C0676QZBWKS/p1720633898241439)


sueplex (dismissed):
LGTM

packages/next/test/fixtures/00-trailing-slash-middleware-add/middleware.js
```diff
@@ -0,0 +1,5 @@
+
+      export default function middleware() {
```
ztanner: nit: lint

ztanner (dismissed):

packages/next/test/fixtures/00-trailing-slash-middleware-add/middleware.js
ijjk: oh this file is written dynamically so can be removed actually.

ztanner (approved):

trek (approved):

EndangeredMassa:
@ijjk are these tests legit to the code change?

ijjk:
@EndangeredMassa no none of them appear Next.js related or related to this change at all. They were passing before we merged main.

PR #11858, opened 2024-07-15
[cli] Add target flag to vc build

onsclom:
Adds `vc build --target` which can be `production` or `preview` in preparation of custom environments.


packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -407,6 +407,41 @@ describe('build', () => {
     }
   });
 
+  it('should pull "production" env vars with `--target production`', async () => {
```
jeffsee55, line 410: Nice test 👍 

**non-blocking:** Do we have anything in place to test this with a custom environment target? Would be nice to get those in place but seems like we may not have all the pieces for that yet

packages/cli/src/commands/build/index.ts
```diff
@@ -152,7 +152,11 @@ export default async function main(client: Client): Promise<number> {
   }
 
   // Build `target` influences which environment variables will be used
-  const target = parsedArgs.flags['--prod'] ? 'production' : 'preview';
+  const target = parsedArgs.flags['--prod']
```
jeffsee55, line 156: **non-blocking:** do we want to throw if both `--prod` and `--target` are provided?

jeffsee55 (approved):
No blockers from me, just a couple of questions. Looks good!

TooTallNate (approved):

PR #11869, opened 2024-07-17
[edge] deprecate `ipAddress` & `geolocation`

Kikobeats:
We want to encourage users to use the same method that is now hosted in the @vercel/functions package.

### DIFF
--- a/packages/edge/src/edge-headers.ts
+++ b/packages/edge/src/edge-headers.ts
@@ -98,7 +98,7 @@ function getFlag(countryCode: string | undefined): string | undefined {
  */
 export function ipAddress(request: Request): string | undefined {
   console.warn(
-    'This method will be removed in the next major version. Use `{ ipAddress } from @vercel/functions` instead.'
+    'This method will be removed in the next major version. Use `import { ipAddress } from @vercel/functions` instead.'
   );
   return getHeader(request, IP_HEADER_NAME);
 }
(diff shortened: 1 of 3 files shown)


packages/edge/src/edge-headers.ts
```diff
@@ -97,6 +97,9 @@ function getFlag(countryCode: string | undefined): string | undefined {
  * @param request The incoming request object which provides the IP
  */
 export function ipAddress(request: Request): string | undefined {
+  console.warn(
+    'This method will removed in the next major version. Use `{ ipAddress } from @vercel/functions` instead.'
```
ecklf:
```suggestion
    'This method will removed in the next major version. Use `import { ipAddress } from @vercel/functions` instead.'
```

packages/edge/src/edge-headers.ts
```diff
@@ -126,6 +129,10 @@ function getRegionFromRequestId(requestId?: string): string | undefined {
  * @param request The incoming request object which provides the geolocation data
  */
 export function geolocation(request: Request): Geo {
+  console.warn(
+    'This method will removed in the next major version. Use `{ geolocation } from @vercel/functions` instead.'
```
ecklf:
```suggestion
    'This method will removed in the next major version. Use `import { geolocation } from @vercel/functions` instead.'
```

ecklf (dismissed):

trek (dismissed):

Schniz (dismissed):

trek (dismissed):

ecklf (approved):

PR #11882, opened 2024-07-20
[tests] add vitest custom reporter

EndangeredMassa:
Implemented a basic custom reporter to remove header info from `console.log` lines in `vitest` runs. It copies only the core logic we need from [the base class](https://github.com/vitest-dev/vitest/blob/f1ef2f6362d1607953d383e395d503feb5368613/packages/vitest/src/node/reporters/base.ts#L295).

### Before

In a small example project, I have `console.log` showing order of hooks used. Before this PR, the tests look like this:

```
 RUN  v0.33.0 /Users/smassa/source/demo/vitest

stdout | unknown test
~ beforeAll 1
~ beforeAll 2

stdout | sum.test.mjs > other block > nested block > nested: adds 1 + 2
~ beforeEach 1
~ beforeEach 2
~ nested test

stdout | sum.test.mjs > other block > adds 3 + 3
~ beforeEach 1
~ outer test

 ✓ sum.test.mjs (2)
   ✓ other block (2)
     ✓ nested block (1)
       ✓ nested: adds 1 + 2
     ✓ adds 3 + 3

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Start at  19:56:58
   Duration  217ms (transform 24ms, setup 0ms, collect 12ms, tests 2ms, environment 0ms, prepare 68ms)
```

### After

After this PR, it looks like this:

```
 RUN  v2.0.1 /Users/smassa/source/demo/vitest

~ beforeAll 1
~ beforeAll 2
~ beforeEach 1
~ beforeEach 2
~ nested test
~ beforeEach 1
~ outer test
 ✓ sum.test.mjs (2)
   ✓ other block (2)
     ✓ nested block (1)
       ✓ nested: adds 1 + 2
     ✓ adds 3 + 3

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Start at  20:24:27
   Duration  261ms (transform 13ms, setup 0ms, collect 13ms, tests 2ms, environment 0ms, prepare 62ms)
```

### Why?

The origin header is kinda helpful, when it works. But it (1) doesn't always work and (2) it's pretty verbose.

In cases where we do care a lot about the origin, like the very parallel operations in the Framework Test Matrix, we make sure to log the origin in the message itself. ([example](https://github.com/vercel/api/actions/runs/10015882484/job/27688230963?pr=30087#step:6:572))


vercel.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "https://openapi.vercel.sh/vercel.json",
+  "outputDirectory": "public",
```
EndangeredMassa, line 3: Context: This is required because adding `vite` to the dependencies triggered framework detection to run in `static-build` for this repo's Vercel project, which uses a different output directory.

We can't use `"framework": null` here because it doesn't properly override this check. We have to override the `outputDirectory` directly.

jeffsee55 (approved):

PR #11883, opened 2024-07-20
Add hint to `vercel deploy` error

willthefirst:
I hit this error after trying to build my project manually using `pnpm build`. I did not realize that I had to specifically use Vercel CLI's build tool. 

Fix: this hint indicates the command that I had failed to run before hitting this error.


EndangeredMassa:
Sorry that this was confusing for your use case. This hint isn't strictly true, though. We'd rather clarify this in docs and templates where it's confusing.

PR #11887, opened 2024-07-22
Revert "Revert "warn on mismatched corepack and detected package managers""

EndangeredMassa:
Reverts vercel/vercel#11879


onsclom (dismissed):

onsclom:
Approving to allow testing

onsclom:
The problem was in [this branch of logic](https://github.com/vercel/vercel/blob/bfef7c631a952b08acb18a48e7092bf69c77811d/packages/build-utils/src/fs/run-user-scripts.ts#L760). If corepack is not enabled, we should prefer detecting the package manager from the lockfile, but `getPathOverrideForPackageManager` would simply return `NO_OVERRIDE`. This uses the default `pnpm` in the build container, `pnpm 6`, which would error on different `package.json#engines.pnpm` values.

[This commit](https://github.com/vercel/vercel/pull/11887/commits/0a77be075b826f6db33accd2f8d5c7d0852afccd) shows and fixes the problematic test case. Fixed the behavior, tested with `vercel/api`, and it now works as expected!

EndangeredMassa (commented):
👍 
(I can't approve this because I opened the PR.)

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -818,6 +846,67 @@ function validateVersionSpecifier(version?: string) {
...
+            path: '/pnpm9/node_modules/.bin',
+            detectedLockfile: 'pnpm-lock.yaml',
+            detectedPackageManager: 'pnpm@9.x',
+          };
+        case 'pnpm 6':
+          return {
+            // undefined because pnpm@6 is the current default in the build container
+            path: undefined,
+            detectedLockfile: 'pnpm-lock.yaml',
+            detectedPackageManager: 'pnpm 6',
```
EndangeredMassa: Can we change this to match the format?
```suggestion
            detectedPackageManager: 'pnpm@6.x',
```

onsclom (approved):

Thompson1985 (approved):

PR #11889, opened 2024-07-22
[frameworks] Add FastHTML framework preset

TooTallNate:
Adds an (experimental) preset for `fasthtml`, which is a new Python based SSR framework. The preset is configured to invoke the `@vercel/python` Builder when a root-level `main.py` file is present.


examples/fasthtml/vercel.json
```diff
@@ -0,0 +1,12 @@
+{
+  "routes": [
```
TooTallNate, line 2: This is temporary and will be removed in a follow-up PR, but this solves a chicken + egg problem because the `fasthtml` preset is not yet accepted by our API endpoints.

packages/frameworks/src/frameworks.ts
```diff
@@ -1960,6 +1960,49 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'FastHTML (Experimental)',
+    slug: 'fasthtml',
+    demo: 'https://fasthtml-template.vercel.app',
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/fasthtml.svg',
```
TooTallNate, line 1967: I can not find a square logo, but we should get one. Alternatively, we can place the generic Python logo here for now if we must use a square logo.

examples/fasthtml/api/index.py
```diff
@@ -0,0 +1,12 @@
+from fasthtml.common import *
```
TooTallNate, line 1: This file will be removed in a follow-up (it's a duplicate of `main.py`). See the comment in the `vercel.json` file below.

trek (approved):

jeffsee55 (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -1960,6 +1960,49 @@ export const frameworks = [
...
+    tagline: 'The fastest way to create an HTML app',
+    description:
+      'A library for writing fast and scalable Starlette-powered web applications',
+    website: 'https://fastht.ml',
+    useRuntime: { src: 'main.py', use: '@vercel/python' },
+    detectors: {
```
EndangeredMassa, line 1973: Can we also look for `python-fasthtml` inside `requirements.txt`? I think we should be more specific in case a stray `main.py` is in someone's non-fasthtml project.

```
{
  path: 'requirements.txt',
  matchContent:
    'python-fasthtml',
},
```

PR #11891, opened 2024-07-23
fix(cli): wrong log line colors for deploy and inspect commands

feugy:
### 🧐 What's in there?

Since forever the commands displaying build logs only rely on `log.type === 'stderr'` to colorise log lines.
However, many tools (including our own CLI) are emitting their own logs to stderr.
This PR replicates the logic from the Vercel Dashboard and identifies the log lines we should not print as errors, based on their content (best effort). It also improves the warning detection.

| Before | Now | Vercel dashboard |
| --- | --- | --- |
| image | image | image |
| image | image | image |

It's also fixing an issue where the spinner is interleaved with json content when piping to jq.

### 🧪 How to test?

Assuming an existing application with some deployments:
```shell
pnpm -F vercel dev inspect --logs dpl_5BaPWBuG56N8SMfr2MBRd5z3payt
```

For the spinner:
```shell
pnpm build
cd packages/cli/dist
node . logs --json nextjs.org | jq 'select(.message|contains("error"))'
```

Ideally, test with failed deployment: their errors should still be in red.

### ❗ Notes to reviewers

The logic isn't great, but we have a [longer initiative](https://www.notion.so/vercel/Fixing-build-logs-highlighting-1f7163ba82ab45968c01c1a721ec0977) to fix the detection upfront in the processing chain.

In the meantime, these changes brings the CLI on par with the dashboard, for both warnings and errors.


packages/cli/src/util/logs.ts
```diff
@@ -257,3 +243,69 @@ function getSourceIcon(source: string) {
...
+function colorize(text: string, log: any) {
+  if (isError(text, log)) {
+    return chalk.red(text);
+  }
+  return isWarning(text) ? chalk.yellow(text) : text;
+}
+
+function isError(text: string, log: any) {
+  return (
+    (log.type === 'error' && !ignoreStdErr(text)) ||
```
lucleray: Since we haven't enabled that for customers yet, let's maybe not use it for `vc logs` either?

lucleray (dismissed):

packages/cli/src/util/logs.ts
feugy: Good point! 
This simplifies the logic a lot, because we don't have `ignoreStdErr` at all.

lucleray (dismissed):

jeffsee55 (dismissed):

jeffsee55 (approved):

PR #11896, opened 2024-07-24
[cli] Error instead of warn when `--prod` and `--target`/`--environment` are combined

onsclom:
This affects the commands `build`, `deploy`, and `list` as they all support `--prod` with `--target`/`--environment`.

### DIFF
--- a/.changeset/modern-bees-hammer.md
+++ b/.changeset/modern-bees-hammer.md
@@ -1,0 +1,5 @@
+---
+'vercel': major
+---
+
+Error when --prod and --target/--environment are combined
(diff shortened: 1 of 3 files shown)


jeffsee55 (approved):

trek (changes requested):
@onsclom Ah! I don't know if the bug submitter (@EndangeredMassa) knew that we had existing `warn` behavior when combining `--target` and `--prod`. I certainly didn't.

Let's see if he still wants us to bump `major` and `error` or if he'd prefer just applying the existing behavior where it's now required.

onsclom:
> @onsclom Ah! I don't know if the bug submitter (@EndangeredMassa) knew that we had existing `warn` behavior when combining `--target` and `--prod`. I certainly didn't.
> 
> Let's see if he still wants us to bump `major` and `error` or if he'd prefer just applying the existing behavior where it's now required.

Makes sense, I added this warn behavior after the bug bash with https://github.com/vercel/vercel/pull/11890. Happy to close this PR if we prefer that behavior!

trek:
@onsclom yeah, sorry for the confusion. Let's just close this one.

PR #11907, opened 2024-07-29
Fix issue where builder output was treating an api endpoint as a frontend route

jeffsee55:
This fixes an issue where a builder for api routes was being treated as a frontend builder, which caused us to populate the fallback routes for whatever framework was being deployed.


packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1308,6 +1308,47 @@ describe('build', () => {
     });
   });
 
+  it('should doit', async () => {
```
TooTallNate:
```suggestion
  it('should not apply framework `defaultRoutes` when build command outputs Build Output API', async () => {
```

trek (dismissed):

trek (approved):

PR #11912, opened 2024-07-31
[cli] Display custom environments in `vc env ls`

onsclom:
`vc env ls` would not display custom environment names, and this adds support for that.

Unfortunately, environment variables do not contain the names of custom environments (they only contain custom environment ids). By fetching custom environments alongside environment vars, we can resolve ids to their name.

Also, this command still needs to work for users not flagged into custom environments. The custom environment endpoints returns a 404 if the user is not flagged in, so we catch that and continue without using custom environments.

### Before

Screenshot 2024-07-31 at 3 46 52 PM

### After
image


packages/cli/src/commands/env/ls.ts
```diff
@@ -71,14 +74,32 @@ export default async function ls(
         project.name
       )} ${chalk.gray(lsStamp())}`
     );
-    // eslint-disable-next-line no-console
-    console.log(getTable(envs));
+    client.stdout.write(`${getTable(envs, customEnvs)}\n`);
```
jeffsee55, line 77: Should/could this go to `stderr`? I'll admit I'm still not 100% clear on when/why we need to use `stout`.

onsclom: I went from `console.log` -> `client.stdout.write` to remove the eslint disable, and it shouldn't change behavior. (Side note: maybe we should change all eslint disabled `console.log`s to `client.stdout.write`?) There's actually an existing integration test that expects this to be `stdout` too.

Good question about whether this was actually desired behavior though. Looking into [clig.dev](https://clig.dev/#the-basics), sounds like `stdout` makes sense since it's the primary output of the command:

| Send output to stdout. The primary output for your command should go to stdout.

jeffsee55 (approved):

trek (approved):

PR #11926, opened 2024-08-06
[next] Update to pull react version for next fixtures

ijjk:
This avoids having to manually update the react version every time we bump it to the latest rc


TooTallNate (dismissed):

TooTallNate (dismissed):

packages/next/test/utils.ts
```diff
@@ -21,6 +22,21 @@ export function toHeaders(headers: IncomingMessage['headers'] = {}) {
   return obj;
 }
 
+export async function normalizeReactVersion(projectPath: string) {
+  const pkgJson = await fs.readJSON(path.join(projectPath, 'package.json'));
+
+  if (pkgJson.dependencies?.['next'] === 'canary') {
+    const res = await fetch('https://unpkg.com/next@canary/package.json');
```
eps1lon, line 29: `npm view next@canary --json peerDependencies` also works and has the benefit of not relying on another registry (here unpkg). That's what we're using on Front syncs: https://github.com/vercel/front/blob/87fb3e713e35d4e222e959a9db9a788ea3bf81d2/tooling/github-actions/workflows/cron-update-next/index.ts#L135-L144

TooTallNate (approved):

PR #11929, opened 2024-08-07
[cli] Make `getFlagsSpecification()` type safe

TooTallNate:
### Before

Screenshot 2024-08-12 at 16 44 39

### After

Screenshot 2024-08-12 at 16 45 32

> **NOTE:** It would have been nice to use `satisfies Command` in the command.ts files for this case, but that keyword does not exist with TypeScript 4 which is what we are currently using. I tried [updating to TS 5](https://github.com/vercel/vercel/pull/11930), but was running into issues. So for now the `satisfies` is omitted.


trek (approved):

packages/cli/src/commands/help.ts
```diff
@@ -270,7 +270,7 @@ export function buildCommandExampleLines(command: Command) {
         outputArray.push(buildValueLine(line));
       }
     } else {
-      outputArray.push(buildValueLine(example.value));
+      outputArray.push(buildValueLine(example.value as string));
```
EndangeredMassa, line 273: **suggestion (non-blocking):** ‏remove type assertion

Is it possible to remove this type assertion?

packages/cli/src/commands/logs/index.ts
```diff
@@ -39,8 +39,9 @@ export default async function logs(client: Client) {
     return 1;
   }
 
-  for (const flag of deprecatedFlags) {
-    if (parsedArguments.flags[flag]) {
+  // TODO: This behavior should be centralized in `parseArguments`
```
erikareads, line 42: **comment (non-blocking)**: I like this idea, it goes well with how I want to re-model how we handle deprecation, type-wise.

EndangeredMassa (approved):

erikareads (approved):

packages/cli/src/commands/help.ts
erikareads: I assume that's because the `as const` makes those values into literals, so they have to be turned back into strings.

TooTallNate: @erikareads is on the right track. For some reason, `Array.isArray()` was not correctly narrowing the `readonly` array type (from `as const`), so TS thinks this line could still be an array type (it can not).

EndangeredMassa: Can we add a comment above this line explaining that?

TooTallNate: Added comment in a follow-up PR https://github.com/vercel/vercel/pull/11955

PR #11932, opened 2024-08-08
[cli] single point of truth for global options

erikareads:


packages/cli/src/commands/env/ls.ts
```diff
@@ -20,9 +20,7 @@ import { getCommandName } from '../../util/pkg-name';
 import ellipsis from '../../util/output/ellipsis';
 import { isObject } from '@vercel/error-utils';
 
-type Options = {
-  '--debug': boolean;
-};
+type Options = {};
```
erikareads, line 23: This change fixes a type error that was created because the global options are no longer a `const` object. 

This option was never used, so it was safe to remove in this instance, but it may be worth considering how we can force the type of `GLOBAL_OPTIONS` to be fixed.

EndangeredMassa (approved):

PR #11956, opened 2024-08-14
[cli] `vc list` refactor

TooTallNate:
The logic of `vc list` was hard to work with and needed some updating, as it is one of the oldest commands in the CLI codebase.


packages/cli/test/unit/commands/list.test.ts
```diff
@@ -69,67 +69,13 @@ describe('list', () => {
...
       `https://${deployment.url}`,
-      stateString(deployment.state || ''),
+      stateString(deployment.readyState || ''),
       deployment.target === 'production' ? 'Production' : 'Preview',
       getDeploymentDuration(deployment),
       user.username,
     ]);
   });
 
-  it('should get deployments for linked project where the scope is a user', async () => {
```
TooTallNate, line 79: This test was removed because all users are northstar users now, so this was testing a case which can no longer occur.

onsclom (approved):

PR #11961, opened 2024-08-14
[tests] fix lockless fixture for docusaurus and others

EndangeredMassa:
Fixes the flakey test (although it looks like it should always be failing, but whatever) around docusaurus deployments.

Turns out, this was caused by:

- not having a lockfile
- `docusaurus` depending on `truncate-html`
- `truncate-html` depends on `cheerio@^1.0.0-rc.12`
- `cheerio` recently released `1.0.0`, which matches that specifier
- `cheerio@1.0.0` seems to have changed how exactly it's exported
    - the specific cause not determined, but it might be https://github.com/cheeriojs/cheerio/pull/3970/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L33

This PR also adds lockfiles for fixtures that were missing them.

---

Example failure: https://github.com/vercel/vercel/actions/runs/10377908477/job/28733185237?pr=11548#step:8:1326

### DIFF
--- a/.changeset/hip-readers-taste.md
+++ b/.changeset/hip-readers-taste.md
@@ -1,5 +1,4 @@
 ---
-"@vercel/static-build": patch
 ---
 
 fix lockless fixture
--- a/.changeset/hip-readers-taste.md
+++ b/.changeset/hip-readers-taste.md
@@ -1,0 +1,5 @@
+---
+"@vercel/static-build": patch
+---
+
+fix lockless fixture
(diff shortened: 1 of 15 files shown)


trek (dismissed):

EndangeredMassa:
merged in the follow-up PR to here because one of those was failing as well

TooTallNate (approved):

Issue #11918, opened 2024-08-02
[Bug] @vercel/functions/oidc import requires moduleResolution updates


payton:
# Summary
Importing OIDC methods per [documentation](https://vercel.com/docs/functions/vercel-functions-package#oidc-methods) from `@vercel/functions/oidc` does not work out of the box.

# Description
Building with Vercel results in a failed deployment and the following error
```
There are types at '/vercel/path0/node_modules/@vercel/functions/dist/oidc/index.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
```
This is not required by other dependencies, so I assume there's something that can be done on the `@vercel/functions` side that wouldn't require consumers to update module resolution configurations.

PR #11980, opened 2024-08-20
`@vercel/functions` expose `oidc` object

Kikobeats:
Previously, the way to use oidc was importing from `@vercel/functions/oidc` namespace:

```js
import { awsCredentialsProvider } from '@vercel/functions/oidc';
```

This type of import takes advantage of [subpath exports](https://nodejs.org/api/packages.html#subpath-exports).

We have had some users report that this does not work well with their CommonJS bundlers or TypeScript configuration, being just possible to import the library from `dist` folder, like `@vercel/functions/dist/oidc`.

In order to reduce the user friction there as much as possible, the library can just expose a old plain javascript object called `oidc` rather than import from the specific namespace.

So the previous code should be imported in this way:

```js
import { oidc } from '@vercel/functions/oidc';
const { awsCredentialsProvider } = oidc
```

Related https://github.com/vercel/vercel/issues/11918

### DIFF
--- a/.changeset/eleven-drinks-live.md
+++ b/.changeset/eleven-drinks-live.md
@@ -1,0 +1,5 @@
+---
+"@vercel/functions": major
+---
+
+`@vercel/functions` expose `oidc` object
--- a/packages/functions/package.json
+++ b/packages/functions/package.json
@@ -2,16 +2,7 @@
   "name": "@vercel/functions",
   "description": "Runtime functions to be used with your Vercel Functions",
   "homepage": "https://vercel.com",
-  "exports": {
-    ".": {
-      "types": "./dist/index.d.ts",
-      "default": "./dist/index.js"
-    },
-    "./oidc": {
-      "types": "./dist/oidc/index.d.ts",
-      "default": "./dist/oidc/index.js"
-    }
-  },
+  "main": "./dist/index.js",
   "files": [
     "dist"
   ],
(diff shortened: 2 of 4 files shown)


packages/functions/package.json
```diff
@@ -2,16 +2,7 @@
...
-  "exports": {
-    ".": {
-      "types": "./dist/index.d.ts",
-      "default": "./dist/index.js"
-    },
-    "./oidc": {
-      "types": "./dist/oidc/index.d.ts",
-      "default": "./dist/oidc/index.js"
-    }
-  },
```
marcgreenstock, line 14: Should we keep these so `moduleResolution bundler` can keep working as is?

.changeset/eleven-drinks-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/functions": major
```
marcgreenstock, line 2:
```suggestion
"@vercel/functions": minor
```

.changeset/eleven-drinks-live.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+`@vercel/functions` expose `oidc` object
```
Schniz, line 5: can you add the example of the usage here?

seems like the PR description is wrong and you need to `import { oidc } from '@vercel/functions'`?

.changeset/eleven-drinks-live.md
Schniz: why not a major? this is a breaking change

PR #11990, opened 2024-08-21
minor cleanup

Zertsov:
Saw that we're using a deprecated error function - applied the suggested fix and removed a redundant `else` clause.


trek (dismissed):

EndangeredMassa (dismissed):

packages/error-utils/src/index.ts
```diff
@@ -14,7 +14,6 @@ export const isObject = (obj: unknown): obj is Record<string, unknown> =>
 
 /**
  * A type guard for `try...catch` errors.
- * @deprecated use `require('node:util').types.isNativeError(error)` instead
```
Zertsov, line 17: Turns out this function does exactly what the deprecation is saying to do, so I'm not seeing why we have the deprecation notice

TooTallNate (approved):

packages/error-utils/src/index.ts
TooTallNate: I think the point of the deprecation was to encourage the use of the same functionality built in Node core, so eventually we can remove this function once all callsites are updated. I think we should leave this `@deprecated` and go in that direction instead.

TooTallNate (changes requested):

packages/error-utils/src/index.ts
Zertsov: but this function uses the functionality in Node core, right? Unless there's a difference that I'm not seeing. I'd assume we'd _want_ to simply have the `require('node:util').types.isNativeError(error)` in a function so we don't have to write out that whole sequence every time we want to check if something is an error

TooTallNate: My point is that this function isn't _doing_ anything useful beyond what already exists in Node core. IMO the convenience of a smaller function name doesn't outweigh the cognitive overhead. If I'm new to the codebase and see `util.types.isNativeError()` then there's a chance I'll already know what that function is doing, and not need to dig into the code to check explicitly.

PR #11991, opened 2024-08-21
[fs-detectors] Adds new detector for Sanity v3

trek:
Sanity v3 has a different `dev` command than Sanity `<3`. This PR adds a new `frameworks` entry for it. Also adds tests to ensure v2 and v3 are detected correctly between each other.


packages/frameworks/src/frameworks.ts
```diff
@@ -2005,6 +2005,69 @@ export const frameworks = [
...
+        },
+        {
+          path: 'sanity.config.tsx',
+        },
+      ],
+      every: [
+        {
+          path: 'package.json',
+          matchContent:
+            /"(dev)?(d|D)ependencies":\s*{[^}]*"sanity":\s*"\^?3\..*"[^}]*}/,
```
EndangeredMassa: **suggestion (non-blocking):** ‏We should probably change `matchPackage` to handle versions as well.

EndangeredMassa (dismissed):

packages/frameworks/src/frameworks.ts
```diff
@@ -2005,6 +2005,69 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'Sanity',
```
TooTallNate, line 2009: We'll want to use a different name, otherwise this will show up twice in the UI with no way to differentiate between the two.

TooTallNate: We'll want to use a different name, otherwise this will show up twice in the UI with no way to differentiate between the two.

TooTallNate (changes requested):

packages/frameworks/src/frameworks.ts
```diff
@@ -2056,7 +2056,6 @@ export const frameworks = [
         value: 'dist',
       },
     },
-    dependency: '@sanity/cli',
```
trek: `@sanity/cli` isn't a dependency for Sanity 3+ any longer. Unsure if this comment still applies?

https://github.com/vercel/vercel/blob/9d4d4b6deb6294506016106f78e71f1984adcc7f/packages/frameworks/src/frameworks.ts#L12-L16

Several framework entries don't have this key or don't reference a cli as the value.

trek: cc @TooTallNate @EndangeredMassa

EndangeredMassa: This is general guidance that on average is still true. I think it's fine.

EndangeredMassa (approved):

TooTallNate (approved):

PR #11992, opened 2024-08-21
Initial experimental implementation of `@vercel/firewall` programmatic rate limiting

cramforce:
Attention: This is a super interesting unreleased feature paired with an as-of-today unannounced but highly requested feature for the Vercel Firewall 😊. Long live open source! Feedback on the API shape is very welcome.

```ts
import { checkRateLimit } from '@vercel/firewall';

async function handler() {
  const { rateLimited } = await checkRateLimit('my-rate-limit-id');
}
```


packages/firewall/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@vercel/firewall",
+  "description": "[Experimenta] Programming rate limits powered by the Vercel Firewall",
```
leerob:
```suggestion
  "description": "[Experimental] Programming rate limits powered by the Vercel Firewall",
```

packages/firewall/package.json
```diff
@@ -0,0 +1,57 @@
...
+      "default": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "types": "./dist/index.d.ts",
+  "version": "0.0.1",
+  "repository": {
+    "directory": "packages/security",
```
leerob:
```suggestion
    "directory": "packages/firewall",
```

packages/firewall/src/rate-limit.ts
```diff
@@ -0,0 +1,151 @@
...
+      rateLimited: false,
+    };
+  }
+
+  let requestHeaders = options?.headers || options?.request?.headers;
+  if (requestHeaders && !(requestHeaders instanceof Headers)) {
+    requestHeaders = new Headers(requestHeaders);
+  }
+  if (!requestHeaders && process.env.NEXT_RUNTIME) {
+    const { headers } = await import('next/headers');
```
leerob, line 66: Given this will opt into dynamic rendering, we'll want to have some guidance in the docs on the best place to use this. E.g. maybe the code snippet should be middleware?

cramforce: Can do. I wouldn't say that this opts you into dynamic rendering. You'd only ever call this in dynamic functions. I think `route.ts` is probably best, server actions are also great

cramforce: Changed to `route.ts` style example

leerob (dismissed):

TooTallNate (dismissed):
Seems good to me. I wonder if we should prefix the function with `unstable_` considering the experimental nature.

cramforce:
Added the prefix. Little downside risk for now

TooTallNate (approved):

erikareads (approved):

Issue #11573, opened 2024-05-10
Automatic updates for Go releases


StefMa:
For whatever reasons there is a `versionsMap` in the `go-helpers.ts`.
This map indicates what Go versions are supported by Vercel.

As of today, this map will be updated by the community from time to time.
See also
https://github.com/vercel/vercel/pull/11572
https://github.com/vercel/vercel/pull/11156
https://github.com/vercel/vercel/pull/10552
https://github.com/vercel/vercel/pull/9576
https://github.com/vercel/vercel/pull/7590
https://github.com/vercel/vercel/pull/7002
... And probably more 🙃 

Does it make sense to introduce some automation (GitHub Action) that runs periodically, checks for Go updates and adjust the map automatically? 🤔 

We could also start by introducing this only for patch versions.
However, Go is known to not breaking stuff for minot releases. So I woulnd't see the point.

Maybe, as another idea, supporting "any patch" and getting rid of the map in terms of "maximum version" would probably also a good idea?! 🤔 
What do you think?

PR #11997, opened 2024-08-22
Added support for Go 1.23 and updated Go patch versions of 1.22 and 1.21

StefMa:
Title 👆 

Also wanted to point out to this: https://github.com/vercel/vercel/issues/11573


trek:
Hi @StefMa, we're looking to fix address the underlying issue with https://michaelheap.com/access-secrets-from-forks/ but could we move discussion over to https://github.com/vercel/vercel/pull/12027 where tests can run?

PR #12002, opened 2024-08-22
[cli] Add error for `vc deploy --env` with no value

Zertsov:
In preparation for Custom Environments, there's a realistic possibility that users unintentionally use `--env` when deploying to specify an environment rather than an environment variable to deploy with. Since this is _not_ an error state (env vars can be set as empty strings), we notify the user that they may have meant to specify an environment.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -476,12 +487,28 @@ export default async (client: Client): Promise<number> => {
...
-  // build `regions`
+  // Check the `env` flags for empty values. Because we're releasing Custom Environments, there's a chance
+  // that users might think the `--env` flag is for specifying an environment. We should catch this and
+  // provide a helpful error message in the event that it was a mistake. This is not an error state since
+  // environment variables can be set to empty strings.
+  //
+  // https://linear.app/vercel/issue/ZERO-2415/vc-deploy-env-should-have-validation
+  Object.entries(deploymentEnv).forEach(([key, value]) => {
+    if (key && !value) {
+      output.print(
```
TooTallNate:
```suggestion
      output.warm(
```

packages/cli/src/commands/deploy/index.ts
```diff
@@ -476,12 +487,28 @@ export default async (client: Client): Promise<number> => {
...
+        `Key ${key} is defined but empty. If you meant to specify an Environment to deploy to, use ${param('--environment')}`
```
TooTallNate: Except for `vc deploy` it's called `--target` 🦆

TooTallNate (approved):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -509,8 +523,8 @@ export default async (client: Client): Promise<number> => {
 
     const createArgs: CreateOptions = {
       name,
-      env: deploymentEnv,
-      build: { env: deploymentBuildEnv },
+      env: deploymentEnv as Dictionary<string>,
+      build: { env: deploymentBuildEnv as Dictionary<string> },
```
EndangeredMassa, line 527: **suggestion (non-blocking):** ‏Do we need the type assertions? Can we get rid of them for stronger type safety?

Zertsov: I tried playing around with it some more but because environment variables _can_ be undefined we'd end up doing something like
```
// If there's any undefined values, then inherit them from this process
  try {
    await addProcessEnv(log, deploymentEnv);
    await addProcessEnv(log, deploymentBuildEnv);
  } catch (err: unknown) {
    error(errorToString(err));
    return 1;
  }

  // Filter out any remaining undefined values
  const filteredDeploymentEnv: Dictionary<string> = Object.fromEntries(
    Object.entries(deploymentEnv).filter(
      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- _ is supposed to be unused
      ([_, value]) => value !== undefined
    )
  ) as Dictionary<string>;
  const filteredDeploymentBuildEnv: Dictionary<string> = Object.fromEntries(
    Object.entries(deploymentBuildEnv).filter(
      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- _ is supposed to be unused
      ([_, value]) => value !== undefined
    )
  ) as Dictionary<string>;
```
which kinda ends up doing the same thing unless my TS-fu is missing something

PR #12008, opened 2024-08-23
Update CODEOWNERS to remove some

EndangeredMassa:
Removing Zero Config from several paths to reduce notifications. We can still push through changes when we need to respond to an issue.


TooTallNate (dismissed):

onsclom (approved):

TooTallNate (approved):

trek (approved):
🙏

PR #12010, opened 2024-08-23
[cli] create `printError` utility

erikareads:
This PR creates `printError` a refactored version of `handleError` that uses `Output` instead of `console.error`, and renames the function to better represent its use.

`handleError` didn't do any error handling, only output, but didn't use the existing utilities in `client.output` to accomplish that. This PR also fixes a 7 year old bug where `'Canceled'` was never printed to users.


erikareads:
Addresses: https://linear.app/vercel/issue/ZERO-2439/

packages/cli/src/util/handle-error.ts
```diff
@@ -3,6 +3,44 @@ import info from './output/info';
 import errorOutput from './output/error';
 import { APIError } from './errors-ts';
 import { getCommandName } from './pkg-name';
+import { type Output } from './output/create-output';
+
+export function printError(error: unknown, output: Output) {
```
EndangeredMassa, line 8: **suggestion (soft-blocking):** can we make to use this somewhere in this PR or a quick follow up?

I don't like adding unused code because it can sometimes stay that way.

packages/cli/src/util/handle-error.ts
```diff
@@ -3,6 +3,44 @@ import info from './output/info';
...
+  } else if (message) {
+    output.prettyError(apiError);
+  } else if (status === 500) {
+    output.error('Unexpected server error. Please retry.');
+  } else if (code === 'USER_ABORT') {
+    output.log('Canceled');
+  } else {
+    output.error(`Unexpected error. Please try again later. (${message})`);
+  }
+}
```
EndangeredMassa, line 43: **suggestion (soft-blocking):** Can we add some tests for all of this behavior?

packages/cli/src/util/handle-error.ts
```diff
@@ -3,6 +3,44 @@ import info from './output/info';
...
 import { getCommandName } from './pkg-name';
+import { type Output } from './output/create-output';
+
+export function printError(error: unknown, output: Output) {
+  // Coerce Strings to Error instances
+  if (typeof error === 'string') {
+    error = new Error(error);
+  }
+
+  const apiError = error as APIError;
```
EndangeredMassa, line 14: **suggestion (soft-blocking):** Can we remove the type assertion and use type checking instead?

EndangeredMassa (approved):

Bonniemarie216 (approved):

PR #12012, opened 2024-08-23
[cli] refactor error output to use `output.error`

erikareads:
This PR standardizes most remaining `console.error(error())` invocations to use `output.error` instead. `handleError` is left unchanged in this PR because it's already addressed with the creation of `prettyError` in #12010. 

This PR alters a number of call signatures to pass `output` into the appropriate functions.


packages/cli/src/commands/teams/switch.ts
```diff
@@ -9,15 +9,20 @@ import getTeams from '../../util/teams/get-teams';
 import listInput from '../../util/input/list';
 import { Team, GlobalConfig } from '@vercel-internals/types';
 import { writeToConfigFile } from '../../util/config/files';
+import { Output } from '../../util/output';
```
TooTallNate:
```suggestion
import type { Output } from '../../util/output';
```

packages/cli/src/commands/teams/switch.ts
```diff
@@ -9,15 +9,20 @@ import getTeams from '../../util/teams/get-teams';
...
+  team?: Team
+) => {
   if (team) {
     config.currentTeam = team.id;
   } else {
     delete config.currentTeam;
   }
 
-  writeToConfigFile(config);
+  writeToConfigFile(config, output);
```
TooTallNate: I won't block on it, but I would have preferred `output` as the first arg, kinda similar to how we have `client` as the first arg for a lot of other util functions.

packages/cli/src/util/config/files.ts
```diff
@@ -6,11 +6,11 @@ import { fileNameSymbol } from '@vercel/client';
 import getGlobalPathConfig from './global-path';
 import getLocalPathConfig from './local-path';
 import { NowError } from '../now-error';
-import error from '../output/error';
 import highlight from '../output/highlight';
 import { VercelConfig } from '../dev/types';
 import { AuthConfig, GlobalConfig } from '@vercel-internals/types';
 import { isErrnoException, isError } from '@vercel/error-utils';
+import { type Output } from '../output';
```
TooTallNate:
```suggestion
import type { Output } from '../output';
```

If all imports are type-only imports, then it's preferable to use `import type`. See: https://x.com/mattpocockuk/status/1795040643963555859

packages/cli/src/commands/teams/switch.ts
EndangeredMassa: I agree with the general pattern @TooTallNate mentions. I also wouldn't block on it.

TooTallNate (dismissed):

trek (approved):

Thompson1985 (approved):

Thompson1985 (approved):

PR #12013, opened 2024-08-25
Replace minimatch with Node.js LTS (20.17.0) Native Glob Matching

manishjha-04:
> This PR **replaces the existing usage of glob matching** libraries if used (such as minimatch,micromatch, picomatch) with the newly introduced native glob matching support in Node.js LTS (20.17.0).

>  This change aims to streamline the codebase by leveraging the built-in capabilities of Node.js, reducing the dependency footprint and potentially improving performance and maintainability.


trek:
Hi @manishjha-04,

We still support node@18 and plan to support that until the end of security support by the Node.js team so can't accept this change at the moment.

PR #12016, opened 2024-08-26
[vc help] add build command to help/--help

trek:
Not sure how we've missed this all these years!


packages/cli/src/args.ts
```diff
@@ -13,6 +13,7 @@ export const help = () => `
       deploy               [path]      Performs a deployment ${chalk.bold(
         '(default)'
       )}
+      build                            Build the project locally
```
erikareads: There will be a snapshot test that fails because of this change.

erikareads (changes requested):

packages/cli/src/args.ts
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
      build                            Build the project locally into `./vercel/output`
```

trek: @erikareads I oddly couldn't find the snapshot (or the test even) for `help/--help` outside the context of a specific command. Any idea where to look?

trek: I guess we'll see in a moment!

trek: Aha! `test/unit/args.test.ts`. I was just looking at `help.test.ts`

erikareads (approved):

PR #12017, opened 2024-08-26
[build-utils] Mitigate `PNPM_UNSUPPORTED_ENGINE` with corepack

onsclom:
[`ERR_PNPM_UNSUPPORTED_ENGINE`](https://vercel.com/docs/errors/error-list#pnpm-engine-unsupported) is thrown by pnpm when the currently running pnpm version does not satisfy the range in `package.json#engines.pnpm`. pnpm's error message is not very helpful, especially when corepack is enabled.

This change detects a case where `ERR_PNPM_UNSUPPORTED_ENGINE` would be triggered, and displays a more helpful error message instead.


jeffsee55 (approved):

packages/build-utils/test/unit.get-path-override-for-package-manager.test.ts
```diff
@@ -66,6 +66,39 @@ describe('Test `getPathOverrideForPackageManager()`', () => {
     });
   });
 
+  describe('with package.json#engines.pnpm', () => {
```
EndangeredMassa, line 69: Can we add a test where `packageJsonEngines: { pnpm: '>=9.0.0' }` is specied, but `corepackPackageManager` is not?

PR #12029, opened 2024-08-27
[cli] preserve rollback candidates for safe removals

### LINKED ISSUE #12022: vercel remove --safe no longer preserving rollback deployment
When running `vercel rm -s`, in addition to the active deployment, the previous deployment was also be kept, such that rollbacks could still work. That was very convenient for cleaning up old deployments while not also nuking the ability to rollback.

With some of the recent refactoring (#11931 #11394), only the active deployment is now kept.

Happy to submit a PR that brings back that behavior, assuming that's deemed desirable (cc @erikareads).

cosn:
Fixes #12022


.changeset/brown-months-draw.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel-internals/types": patch
```
TooTallNate:
```suggestion
```

cosn:
friendly ping?

PR #12036, opened 2024-08-28
[next] Partial Fallback Prerendering

wyattjoh:
This adds support for reading the fallback status, and headers from the manifests for Partial Fallback Prerendering.

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -914,6 +914,8 @@ export type NextPrerenderedRoutes = {
   fallbackRoutes: {
     [route: string]: {
       fallback: string;
+      fallbackStatus?: number;
+      fallbackHeaders?: Record<string, string>;
       routeRegex: string;
       dataRoute: string | null;
       dataRouteRegex: string | null;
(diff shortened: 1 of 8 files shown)


packages/next/src/utils.ts
```diff
@@ -1283,6 +1289,8 @@ export async function getPrerenderManifest(
             manifest.dynamicRoutes[lazyRoute].prefetchDataRoute;
           prefetchDataRouteRegex =
             manifest.dynamicRoutes[lazyRoute].prefetchDataRouteRegex;
+          fallbackStatus = manifest.dynamicRoutes[lazyRoute].fallbackStatus;
+          fallbackHeaders = manifest.dynamicRoutes[lazyRoute].fallbackHeaders;
```
ijjk, line 1293: Did we need new fields instead of the existing initialHeaders and initialStatus?

ijjk (approved):
Just left a comment on new field name but seems good otherwise

packages/next/src/utils.ts
wyattjoh: These are new fields for the `dynamicRoutes` part of the prerender manifest, and are already included in Next.js.

PR #12003, opened 2024-08-23
[client] Move to generated Vercel TypeScript SDK for `@vercel/client`

leerob:
We are cutting a canary release for v14 to explore using an auto-generated TypeScript SDK client for `@vercel/client`.

PR #12039, opened 2024-08-28
[client] Workflow files for generated Vercel TypeScript SDK for `@vercel/sdk`

simplesagar:
- feat: add generation workflow
- feat: speakeasy workflow setup
- feat: openapi spec

This is a simplified version of this PR (https://github.com/vercel/vercel/pull/12003) that only sets up the config files for SDK generation. Generated code will be provided in an automated PR after this is merged.


leerob (dismissed):

packages/client/.speakeasy/gen.yaml
```diff
@@ -0,0 +1,44 @@
...
+      errors: models/errors
+      operations: models/operations
+      shared: models/components
+      webhooks: models/webhooks
+  inputModelSuffix: input
+  maxMethodParams: 0
+  methodArguments: require-security-and-request
+  moduleFormat: commonjs
+  outputModelSuffix: output
+  packageName: '@vercel/sdk'
```
leerob: I think we want the generated folder to be `sdk` as well so there isn't a clash with `client/` yeah?

simplesagar: Makes sense. Created `packages/sdk` and updated generation config to write and publish the new client from that location.

.github/workflows/sdk_publish.yaml
```diff
@@ -0,0 +1,21 @@
+name: Publish
```
EndangeredMassa, line 1: **question (blocking):** Should we use this publish workflow or existing changesets workflow?

If we get a specific benefit with this publish workflow, we probably need to add this new package to the list of ignored packages in the [changesets config](https://github.com/vercel/vercel/blob/40b7ee0d297c212961279639d9c73d4fed2312f8/.changeset/config.json#L13).

EndangeredMassa (changes requested):

.github/workflows/sdk_publish.yaml
simplesagar: The main benefits of the speakeasy publishing workflow is it will automatically publish a new package to NPM upon merge of any generation PR. We also do a couple of other checks to make sure the package and docs work as expected. it is optional

EndangeredMassa: Sounds worth using, then. I [just added](https://github.com/vercel/vercel/pull/12039/commits/501e692bec8f814d2e567bb2e5a4920e4c992555) the `@vercel/sdk` package to the list of ignored packages for changesets. I think that'll do what we want here.

EndangeredMassa:
I ran the tests for this change here:
https://github.com/vercel/vercel/pull/12047

It looks like prettier doesn't like the new yaml files. Should we have prettier ignore those or is there something here that needs ot be fixed?
https://github.com/vercel/vercel/actions/runs/10623732733/job/29450733201?pr=12047#step:7:15

ryan-timothy-albert:
> I ran the tests for this change here: #12047
> 
> It looks like prettier doesn't like the new yaml files. Should we have prettier ignore those or is there something here that needs ot be fixed? https://github.com/vercel/vercel/actions/runs/10623732733/job/29450733201?pr=12047#step:7:15

Given the SDK generated files will come out formatted it might be worth just prettier ignoring the whole SDK sub dir [here](https://github.com/vercel/vercel/blob/main/.prettierignore).

It also may be worth ignoring the openapi file just to move forward

.github/workflows/sdk_publish.yaml
EndangeredMassa: Turns out that changesets doesn't like ignoring a thing it can't find. We can add that ignore later.

EndangeredMassa:
Updated prettier, build, and pack processes to ignore the `@vercel/sdk` package.

When the package exists with a `package.json`, we'll want to:

- update changesets to ignore this package
- update tarball packing to not ignore this package

Tests pass here: https://github.com/vercel/vercel/pull/12047

TooTallNate (approved):

EndangeredMassa (approved):

EndangeredMassa:
Thanks for the PR! To make this a bit easier, we'll merge this other version of the PR: https://github.com/vercel/vercel/pull/12047

PR #12044, opened 2024-08-29
[Logs] use level to determine build log highlight

Jooliashi:
We have moved the build log highlight to api instead of in `front` and `cli`; as a result, cli and front can both have consistent highlight. We have already deployed and tested the changes to front behind a feature flag and we would like to roll out the cli changes in the next version.

### DIFF
--- a/packages/cli/src/util/events.ts
+++ b/packages/cli/src/util/events.ts
@@ -10,6 +10,7 @@ import Client from './client';
 import getDeployment from './get-deployment';
 import getScope from './get-scope';
 
+import { BuildLog } from './logs';
 export interface FindOpts {
   direction: 'forward' | 'backward';
   limit?: number;
@@ -20,16 +21,9 @@ export interface FindOpts {
 
 export interface PrintEventsOptions {
   mode: 'deploy' | string;
-  onEvent: (event: DeploymentEvent) => void;
+  onEvent: (event: BuildLog) => void;
   quiet?: boolean;
   findOpts: FindOpts;
-}
-
-export interface DeploymentEvent {
-  id: string;
-  created: number;
-  date?: number;
-  serial?: string;
 }
 
 async function printEvents(
--- a/packages/cli/src/util/events.ts
+++ b/packages/cli/src/util/events.ts
@@ -10,7 +10,7 @@ import Client from './client';
 import getDeployment from './get-deployment';
 import getScope from './get-scope';
 
-import { BuildLog } from './logs';
+import type { BuildLog } from './logs';
 export interface FindOpts {
   direction: 'forward' | 'backward';
   limit?: number;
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/logs.ts
```diff
@@ -243,25 +243,12 @@ function sanitize(log: any): string {
   );
 }
 
-function colorize(text: string) {
-  if (isError(text)) {
+function colorize(text: string, log: any) {
+  if (log.level === 'error') {
```
jeffsee55, line 280: Given that `log` is `any`, it's hard to know if this will always have the appropriate info. Can we fallback to the old behavior (ie. the `isError` logic) if `log.level` isn't present to be safe? Perhaps adding a note about it would help us know that it's fallback behavior and we can work to get some type-safety in here in the future.

jeffsee55 (changes requested):

packages/cli/src/util/logs.ts
Jooliashi: I don't think that will be correct behavior as `level` wouldn't always be present by default. If a log is not error or warning, `level` will be undefined; The legacy checks had some incorrect behavior of marking error and warning so if the `level` is supposed to be undefined, going through legacy checks will add back the incorrect highlight.

Jooliashi: i can actually just type this

jeffsee55: Ok if it's easy enough for you to type it that'd be awesome. Either way, not worth me blocking you on this if the fallback suggestion doesn't make sense.

packages/cli/src/util/events.ts
```diff
@@ -10,6 +10,7 @@ import Client from './client';
 import getDeployment from './get-deployment';
 import getScope from './get-scope';
 
+import { BuildLog } from './logs';
```
feugy: **suggestion:** please import it as a type.

I'll be surprised if eslint doesn't rant about it <picture data-single-emoji=":lol-think:" title=":lol-think:">:lol-think:</picture>

packages/cli/src/util/events.ts
```diff
@@ -20,18 +21,11 @@ export interface FindOpts {
 
 export interface PrintEventsOptions {
   mode: 'deploy' | string;
-  onEvent: (event: DeploymentEvent) => void;
+  onEvent: (event: BuildLog) => void;
   quiet?: boolean;
   findOpts: FindOpts;
 }
 
-export interface DeploymentEvent {
```
feugy, line 28: **praise:** thank you so much for fixing this interface name.

feugy (dismissed):
The API change is live already, and it's available on older versions (v1~2) as well (current is v3).

So it is safe to have no fallback solution and to only rely on `level` property.

packages/cli/src/util/events.ts
Jooliashi: surprised you shall be XD

feugy (approved):
(I forgot to approve yesterday)

jeffsee55 (approved):

Jooliashi:
@jeffsee55 i will merge this if all good (seeing branch protection here)

PR #12051, opened 2024-08-30
[custom-environments] write custom envs to project.json

Zertsov:
To help validate if a target is correct we want to write the custom envs to the project settings so we can validate before reaching out to any services

### DIFF
--- a/packages/cli/src/util/projects/project-settings.ts
+++ b/packages/cli/src/util/projects/project-settings.ts
@@ -18,12 +18,24 @@ export type ProjectLinkAndSettings = Partial<ProjectLink> & {
     framework: Project['framework'];
     nodeVersion: Project['nodeVersion'];
     analyticsId?: string;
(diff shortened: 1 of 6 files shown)


Zertsov:
I messed up my Graphite stacks so this contains a previous PR's changeset which I'll fix before merging, amongst some other messed up stuff

packages/cli/src/util/projects/project-settings.ts
```diff
@@ -18,12 +18,24 @@ export type ProjectLinkAndSettings = Partial<ProjectLink> & {
     framework: Project['framework'];
     nodeVersion: Project['nodeVersion'];
     analyticsId?: string;
+    // TODO: We should use the Project['customEnvironments'] here when it becomes more available / concrete
```
Zertsov, line 21: up for debate on this one @TooTallNate

PR #12063, opened 2024-09-03
[next] Support Request Chaining Types

wyattjoh:
This introduces Request Chaining which is a stabilization of the `experimentalStreamingLambdaPath` feature (for Partial Prerendering) to include additional fields like `headers` which will be sent alongside the request to the "chained" lambda in order to perform the resume request. Previously only the output path was provided, which served as the `x-matched-path` header value as well as the output path. This was at fault, as it assumed that the `basePath` was `/`. When the `basePath` is configured in Next.js, this breaks the implementation. This shifts the logic of which headers to send down to Next.js so it can just provide the explicit headers to be set rather than the logic having to exist upstream.

This is a supporting PR to #12117 which adds the implementation.


wyattjoh:
* **#12125** <a href="https://app.graphite.dev/github/pr/vercel/vercel/12125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12117** <a href="https://app.graphite.dev/github/pr/vercel/vercel/12117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12063** <a href="https://app.graphite.dev/github/pr/vercel/vercel/12063?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

devjiwonchoi (dismissed):

packages/build-utils/src/prerender.ts
```diff
@@ -165,5 +168,27 @@ export class Prerender {
...
     }
+
+    if (chain !== undefined) {
+      if (typeof chain !== 'object') {
+        throw new Error(
+          'The `chain` argument for `Prerender` must be an object.'
+        );
+      }
+
+      if (!chain.headers || typeof chain.headers !== 'object') {
```
ijjk: It might be good to validate the headers entries are correct shape as well similar to `initialHeaders` validation

ijjk (approved):

packages/build-utils/test/unit.test.ts
```diff
@@ -538,6 +538,83 @@ it('should support experimentalStreamingLambdaPath correctly', async () => {
...
 });
 
+it('should support chain correctly', async () => {
+  new Prerender({
+    expiration: 1,
+    fallback: null,
+    group: 1,
+    bypassToken: 'some-long-bypass-token-to-make-it-work',
+    chain: undefined,
+  });
```
EndangeredMassa, line 548: **suggestion (soft-blocking):** split into separate tests

In this or a follow-up PR: Can you split these into separate tests? When one of these fails, the message will be "should support chain correctly", which isn't very specific.

It also makes it harder to review. I don't know what's supposed to be correct or not about these different cases.‏

EndangeredMassa (approved):

PR #12064, opened 2024-09-03
[next] Combine PPR lambdas

wyattjoh:
This combines the revalidation and resume lambdas for Partial Prerendering (PPR) into a single one, as the revalidation lambdas now support streaming.


ijjk (dismissed):
Change makes sense, do we wanna add an assertion in our integration test that the two outputs are the same lambda instance?

PR #12088, opened 2024-09-07
Warn on `vc env pull [filename]` when `[filename]` is not in `.gitignore`

dmmulroy:
## Warn on vc env pull [filename] when [filename] is not in .gitignore

This PR enhances the `vercel env pull` command to improve security and user experience:

1. Adds warning when pulling environment variables to non-gitignored files
2. Prompts user to add the file to .gitignore if not already present
3. Maintains backward compatibility with --yes flag
4. Introduces new utility functions for .gitignore operations
5. Adds unit tests for new functionality

The changes aim to prevent accidental exposure of sensitive data while maintaining existing workflow

image

### DIFF
--- a/packages/cli/src/commands/env/pull.ts
+++ b/packages/cli/src/commands/env/pull.ts
@@ -1,7 +1,7 @@
 import chalk from 'chalk';
 import { outputFile } from 'fs-extra';
 import { closeSync, openSync, readSync } from 'fs';
-import { resolve } from 'path';
+import { join, resolve } from 'path';
 import Client from '../../util/client';
 import { emoji, prependEmoji } from '../../util/emoji';
 import confirm from '../../util/input/confirm';
@@ -21,6 +21,7 @@ import { addToGitIgnore } from '../../util/link/add-to-gitignore';
 import JSONparse from 'json-parse-better-errors';
 import { formatProject } from '../../util/projects/format-project';
 import type { ProjectLinked } from '@vercel-internals/types';
+import { gitIgnoreIncludesFile } from '../../util/git-helpers';
 
 const CONTENTS_PREFIX = '# Created by Vercel CLI\n';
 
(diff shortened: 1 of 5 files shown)


aaronbrown-vercel:
Very beneficial change. @dmmulroy  This likely would have helped us prevent some incidents regarding secrets leaking into source control, and I imagine it may for some of our customers as well. 

My vote is ship, but I leave it to the SMEs.

jeffsee55 (commented):
Thanks for the thorough PR @dmmulroy! Excellent tests 👍  

The one thing that jumps out to me is that this will effectively behave as a breaking change due to the need to include `--yes` for CI workflows. I believe `vc env pull` is often used in CI in conjunction with `vc deploy`. Another thing is that in those environments its often fine if the env file is not gitignored.

IMO, the warning definitely makes sense to stick around regardless. But I wonder if instead we can do one of the 2:

- Remove (very slick) gitignore prompt
- Or perhaps, detect CI=true and auto confirm. Though even then we may not want to move forward with it because it'd be a breaking change regardless.

Thoughts? Might bring it up with the team to discuss.

packages/cli/src/commands/env/pull.ts
```diff
@@ -161,6 +165,28 @@ export default async function pull(
...
+    const included = gitIgnoreIncludesFile(filename, gitignorePath);
+
+    if (!included) {
+      output.warn(
+        `You are pulling your Environment Variables into ${chalk.bold(filename)} which is ` +
+          `${chalk.red(chalk.bold('NOT'))} in your .gitignore. This means that any ` +
+          `sensitive values, API keys, or secrets could be commited to git.`
+      );
+
+      if (!skipConfirmation) {
```
EndangeredMassa, line 178: **question (blocking):** ‏Is this the best user experience?

I think a user could be confused about how `--yes` works here, assuming it would update the git ignore file.

Maybe instead of prompting for this to happen, we show a command the user can copy-paste to do the same thing, making sure it changed based on current OS.

> You can add this with: `echo "my-env-file.env" >> ../path-to-git-root/.gitignore`

---

If we keep this flow, I share [Jeff's concerns](https://github.com/vercel/vercel/pull/12088#pullrequestreview-2290813854) about how this works in CI. We'd probably want to skip confirmation in CI.

EndangeredMassa (changes requested):

packages/cli/src/commands/env/pull.ts
EndangeredMassa: Note: If we do this, we'd also want to update `vercelignore` to make sure those secrets don't get uploaded. This won't matter in the basic case, but if the user updates that file after pulling, it would.

PR #12098, opened 2024-09-11
feat(cli): add `login --future` and `logout --future`

### LINKED ISSUE #6899: [cli] Enable signup with email address

balazsorban44:
This PR introduces a new `vercel login --future` login experience by using the [OAuth Device Authorization](https://datatracker.ietf.org/doc/html/rfc8628) flow and the corresponding `vercel logout --future` command.

Closes #11989
Closes #6899, since we are moving away from maintaining CLI login flows.

For now, it is behind the `--future` flag, and only available for Vercelians. We plan to release it more broadly after some testing.


jeffsee55 (commented):
@balazsorban44 is this ready for review? Tried running this locally and got:

```
> pnpm dev login --future
Vercel CLI 37.4.1
WARN! This command is not ready yet. Do not use!
Error: Device authorization request failed
```
And there aren't any tests to peek at for hints on how to use, let me know any tips on testing this 🙏

packages/cli/src/commands/login/future/index.ts
```diff
@@ -0,0 +1,201 @@
...
+  deviceAccessTokenRequest,
+  processDeviceAccessTokenResponse,
+  isOAuthError,
+  verifyJWT,
+} from '../../../util/oauth';
+import o from '../../../output-manager';
+import { LoginTelemetryClient } from '../../../util/telemetry/commands/login';
+
+export async function login(client: Client): Promise<number> {
+  o.warn('This feature is under active development. Do not use!');
```
marcgreenstock: Do we need this? Can we not allow users to begin using it right away?

packages/cli/src/util/oauth.ts
```diff
@@ -0,0 +1,307 @@
...
+} = {
+  client_id: 'cl_v9RsjXlLr2yv2ZwgOSz1pe0CQvhzRn9W',
+  revocation_endpoint: new URL(
+    'https://vercel.com/api/login/oauth/token/revoke'
+  ),
+  device_authorization_endpoint: new URL(
+    'https://vercel.com/api/login/oauth/device-authorization'
+  ),
+  token_endpoint: new URL('https://vercel.com/api/login/oauth/token'),
+  jwks_uri: new URL('https://vercel.com/api/login/oauth/jwks'),
```
marcgreenstock:
```suggestion
  jwks_uri: new URL('https://vercel.com/.well-known/jwks'),
```

packages/cli/src/commands/login/future/index.ts
balazsorban44: Will drop. :+1:

packages/cli/src/util/oauth.ts
balazsorban44: Probably safer to use the discovery endpoint and not hard-code too much in CLI. Will update!

marcgreenstock (dismissed):

.changeset/heavy-terms-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
```
trek: This is probably technically a minor (introduces new features), but not blocking.

packages/cli/vitest.config.mts
```diff
@@ -4,5 +4,6 @@ import rootConfig from '../../vitest.config.mjs';
 export default mergeConfig(rootConfig, {
   test: {
     setupFiles: ['./vitest.setup.mts'],
+    coverage: { reporter: ['html'] },
```
trek: This is nice to add but seems unrelated to adding login?

packages/cli/src/commands/login/index.ts
```diff
@@ -24,8 +24,11 @@ import { getFlagsSpecification } from '../../util/get-flags-specification';
 import { printError } from '../../util/error';
 import output from '../../output-manager';
 import { LoginTelemetryClient } from '../../util/telemetry/commands/login';
+import { login as future } from './future';
 
 export default async function login(client: Client): Promise<number> {
+  if (client.argv.slice(2).includes('--future')) return await future(client);
```
trek: If this is shifted down, `--future` will be available as `parsedArgs.flags['--future']`

packages/cli/src/commands/logout/index.ts
```diff
@@ -13,8 +13,11 @@ import { logoutCommand } from './command';
 import { getFlagsSpecification } from '../../util/get-flags-specification';
 import output from '../../output-manager';
 import { LogoutTelemetryClient } from '../../util/telemetry/commands/logout';
+import { logout as future } from './future';
+
+export default async function logout(client: Client): Promise<number> {
+  if (client.argv.slice(2).includes('--future')) return await future(client);
```
trek: If this is shifted down, `--future` will be available as `parsedArgs.flags['--future']`

packages/cli/test/unit/commands/logout/future.test.ts
```diff
@@ -0,0 +1,145 @@
...
+import { client } from '../../../mocks/client';
+import { vi } from 'vitest';
+import fetch, { type Response } from 'node-fetch';
+import { as, VERCEL_CLI_CLIENT_ID } from '../../../../src/util/oauth';
+import { randomUUID } from 'node:crypto';
+import ua from '../../../../src/util/ua';
+
+const fetchMock = fetch as unknown as MockInstance<typeof fetch>;
+
+vi.mock('node-fetch', async () => ({
```
trek, line 12: We have a mock HTTP vercel server stored on the `client` (view a few other unit tests to get a sense for it) that would be preferable to use to avoid introducing a second alternate pattern to the tests.

trek (changes requested):
Looks good to me, other than some comments.

If we don't intend for this to be used, we should *not* merge and release this. Customers will see it and will begin to use it.

If you wanted to get this into the CLI to test, we can help show you some options to before we merge and release!

packages/cli/test/unit/commands/logout/future.test.ts
balazsorban44: To get the benefits of `client`s mocking, we also need to use `client.fetch` in the code itself then. I did try using `client.fetch` first, but decided to go with raw `fetch` instead, because https://github.com/vercel/vercel/blob/main/packages/cli/src/util/client.ts/#L135-L221 had too much magic (adding/overriding headers, search params) and broke the requests/responses constantly.

Some examples:

- The `authorization` header being attached with no option to remove it
- `teamId` ending up in all request's search params
- Unexpected difference between prod/test by auto-retry having different behavior
- possibly more

I _could_ work around some of these by hacking the requests/responses before and after `client.fetch`, but it would bloat the PR, and I probably would need to modify `client.ts` anyway.

The `client`'s behavior is handy for most of our APIs, but we want to follow the OAuth spec strictly, so either I need to modify `client.fetch` via adding a `raw: true` flag to disable unexpected modifications and refactor the `client` and `oauth.ts` file to rely on the `client.fetch`, or we just keep the current mocks. Let me know what would be more preferable!

trek (dismissed):

trek (approved):

PR #12107, opened 2024-09-12
chore: integrate sdk package into monorepo

disintegrator:
* Switch to pnpm for compilation
* Update changeset config to ignore the SDK package from CI checks
* Configure SDK to drop barrel files and use dual bundling
* Delete extraneous package-lock.json


EndangeredMassa (approved):
Can `"prepublishOnly": "npm run build"` use `pnpm` too?

disintegrator:
@EndangeredMassa not currently but I'll ship something that will make it possible to define that command.

PR #12126, opened 2024-09-17
[cli] add new argument `--root-directory`

### LINKED ISSUE #10609: Vercel CLI not able to pass "directory of code location" in monorepo
### Problem Description

It would be really great if we could pass in the "directory of code location" where the code lives as an option 

Very naive eg : 
`vercel --code-location ./packages/nexts`

Reason : 
We have a starter kit/template which is `monorepo` having below directory stucture : 
```ts
root/
├── packages/
│   └── nextjs
|   |   ├─- vercel.json
|   └── hardhat
├── package.json
```

i.e we have two packages `nextjs` and `hardhat` we have `vercel` installed at the root of the project as mentioned in FAQ's of https://vercel.com/docs/concepts/monorepos

Now when you do `yarn vercel` at the root its asks you : 
Screenshot 2023-04-20 at 9 10 13 PM

It would really be great we could preconfigure it use `pacakges/nextjs` as the `source` where the code lives and users don't need specify the path themselves

Am I missing something ? Is this already built in verlce cli ? Is there a way to preconfigure it throught `vercel.json` ? If not I think this feature should be there built in vercel cli 😅 

checkout this comment -> https://github.com/scaffold-eth/scaffold-eth-2/pull/231#issuecomment-1471632149 for more context

Tysm 🙌🙏

LoicMahieu:
Fix #10609

### DIFF
--- a/packages/cli/src/commands/deploy/command.ts
+++ b/packages/cli/src/commands/deploy/command.ts
@@ -139,6 +139,13 @@ export const deployCommand = {
       deprecated: true,
       description: 'Use default options to skip all prompts',
     },
+    {
+      name: 'root-directory',
+      shorthand: null,
+      type: String,
+      deprecated: false,
+      description: 'Specify the root directory',
+    },
   ],
   examples: [
     {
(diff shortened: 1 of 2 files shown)


maneike:
Thanks Loïc! 

This is a blocker for our team too. We're developing a tool entirely made out of CLI commands.
There is a way to do everything with Vercel commands: logging in, deploying, linking, connecting but there is no way to change the root directory when deploying a project without going to the projects settings in the browser or providing it manually in the terminal.

erikareads:
The underlying code here recently changed, sorry about that. If you still want this feature, please open a new PR. If you open a new PR, we will prioritize looking at it to avoid this happening again.

PR #12127, opened 2024-09-17
[cli] provision marketplace resources via cli

lukahartwig:
## Summary

This PR allows Vercel CLI users, in certain conditions, to provision resources from the Vercel Marketplace without leaving the terminal.

## Background

In order for users to buy products on the Vercel Marketplace they need to install the integration of the Provider first. This means for the Vercel CLI there are two scenarios:

### Scenario 1: The Marketplace integration IS NOT installed

In this scenario the user needs to accept the terms for Vercel Marketplace and the Provider (e.g. Supabase). If this scenario is detected we offer the user to open the Vercel dashboard in the team and project context where they can accept the terms and install the product via the Web UI. This is how `vc install` works today and has not changed.

### Scenario 2: The Marketplace integration IS installed and the terms have been accepted

If the terms have been accepted we can attempt to provision the resource without leaving the terminal. If we determine the provisioning can't be done in the terminal, we offer to open Vercel dashboard instead (same as scenario 1).

The conditions are:

* The Marketplace integration was installed once and the terms have been accepted
* The selected product is a storage product (this is the only Marketplace product supported at the moment)
* The provider-defined metadata can be expressed in the terminal UI

## How to test

Build and link Vercel CLI locally. The 3 integrations supported today are

```
vc i supabase
vc i redis
vc i edgedb
```


packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
   }
 
+  const { contextName, team } = await getScope(client);
+
+  if (!team) {
+    client.output.error('Team not found');
+    return 1;
+  }
+
   const integrationSlug = args[0];
```
EndangeredMassa: **suggestion (soft-blocking):** Validate that the slug exists

We should make sure a value was passed in. Otherwise, you get a basic 404 error that doesn't tell the user what to do.

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+  const integration = await fetchIntegration(client, integrationSlug);
+
+  if (!integration?.products) {
```
EndangeredMassa: **question (soft-blocking):** ‏Why did we remove the messaging here and below on line ~44?

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
-  if (!teamId) {
-    client.output.error('Team not found');
+  if (!product || !installations) {
+    return 1;
+  }
+
+  const installation = installations.find(
+    install =>
+      install.ownerId === team.id && install.installationType === 'marketplace'
+  );
```
EndangeredMassa: **question (non-blocking):** ‏What should happen if more than one item matches? Is that possible?

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+    `Installing ${chalk.bold(product.name)} by ${chalk.bold(integration.name)} under ${chalk.bold(contextName)}`
+  );
+
+  const metadataSchema = product.metadataSchema;
+  const metadataWizard = createMetadataWizard(metadataSchema);
+
+  if (
+    !installation ||
+    !metadataWizard.isSupported ||
+    product.type !== 'storage'
```
EndangeredMassa: **question (non-blocking):** ‏Why does it matter that "storage" is not allowed in this condition?

Can you pull this check out into a named variable that makes it clearer why this matters?

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
-}
+import {
+  BillingPlan,
+  Integration,
+  IntegrationProduct,
+  Metadata,
+} from './types';
+import { createMetadataWizard } from './wizard';
 
 export async function add(client: Client, args: string[]) {
```
EndangeredMassa, line 27: **suggestion (non-blocking):** extract functions into identifiable, discrete steps

This function is growing to a large, procedural list of commands. It's a bit hard for me to follow in this review. I suggest extracting steps into functions so that this command can be understood at a higher level more quickly. Then each step can focus on just what it needs to do.
‏

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+
+  const billingPlans = await fetchBillingPlans(
+    client,
+    integration,
+    product,
+    metadata
+  );
+
+  if (!billingPlans) {
     return 1;
```
EndangeredMassa: **suggestion (soft-blocking):** Can we log a message about what's wrong here and what the user can do about it?‏

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+      product.id,
+      billingPlan,
+      name,
+      metadata
     );
+    storeId = result.store.id;
+  } catch (error) {
+    client.output.error((error as Error).message);
     return 1;
   }
```
EndangeredMassa, line 224: **suggestion (non-blocking):** add `finally` to stop spinner

Can we add a `finally` to turn off the spinner? It's not necessary right now because of the other output calls (which stops it automatically), but that can shift over time.‏



```suggestion
  } finally {
    client.output.stopSpinner();
  }
```

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+    await connectStoreToProject(
+      client,
+      projectLink.project.id,
+      storeId,
+      environments
+    );
+  } catch (error) {
+    client.output.error((error as Error).message);
+    return 1;
+  }
```
EndangeredMassa, line 238: **suggestion (non-blocking):** add `finally` to stop spinner

Can we add a `finally` to turn off the spinner? It's not necessary right now because of the other output calls (which stops it automatically), but that can shift over time.‏



```suggestion
  } finally {
    client.output.stopSpinner();
  }
```

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,86 +1,284 @@
...
+}
+
+async function fetchInstallations(client: Client, integration: Integration) {
+  try {
+    return await client.fetch<
+      {
+        id: string;
+        installationType: 'marketplace' | 'external';
+        ownerId: string;
+      }[]
```
EndangeredMassa: **suggestion (non-blocking):** ‏move API functions to a separate file

Our goal over time is to transform `packages/client` into a proper API client. It would help us get there if you could extract these API functions into a separate file for this command, like `./packages/cli/src/commands/integration/client.ts`.

In that file (or the types file), if we could also extract some expected response types into named types, that'd be helpful.

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+  MetadataSchema,
+  MetadataSchemaProperty,
+} from './types';
+
+type Step = (client: Client) => Promise<MetadataEntry>;
+
+function createInputStep(key: string, schema: MetadataSchemaProperty) {
+  if (schema['ui:control'] !== 'input') {
+    throw new Error(
+      `Expected control "input", but was "${schema['ui:control']}"`
```
EndangeredMassa: **suggestion (non-blocking):** add `key` to help with debugging when this is thrown‏

```suggestion
      `Expected control "input" for key "${key}", but was "${schema['ui:control']}"`
```

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+
+            return true;
+          },
+        });
+        return [key, value] as const;
+      };
+    }
+    default: {
+      throw new Error(
+        `Unsupported schema type for input control: ${schema.type}`
```
EndangeredMassa: **suggestion (non-blocking):** add `key` to help with debugging when this is thrown

```suggestion
        `Unsupported schema type for input control of key "${key}": ${schema.type}`
```

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+    case 'number': {
+      return async (client: Client) => {
+        const value = await client.input.text({
+          message: schema['ui:placeholder'] || schema['ui:label'] || key,
+          default: schema.default,
+          validate: value => {
+            const number = Number(value);
+
+            if (isNaN(number)) {
+              return 'Value must be a number.';
```
EndangeredMassa: **suggestion (non-blocking):** These error messages here and below could be more explicit to help in future debugging.

I suggest adding an anchor and the values used, like:

‏
```suggestion
              return 'InputStep: Value "${number}" must be a number.';
```

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+function createSelectStep(key: string, schema: MetadataSchemaProperty) {
+  if (!['select', 'region', 'vercel-region'].includes(schema['ui:control'])) {
+    throw new Error(
+      `Expected control "select", "region" or "vercel-region", but was "${schema['ui:control']}"`
+    );
+  }
+
+  if (!schema['ui:options']?.length) {
+    throw new Error(
+      `Expected controle for key "${key}" to have options, but was ${JSON.stringify(schema['ui:options'])}`
```
EndangeredMassa: **suggestion (soft-blocking):** 

```suggestion
      `Expected control for key "${key}" to have options, but was ${JSON.stringify(schema['ui:options'])}`
```

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+  for (const option of options) {
+    if (typeof option === 'string') {
+      choices.push({
+        name: option,
+        value: option,
+        short: option,
+        selected: Boolean(defaultValue && option === defaultValue),
+      });
+    } else {
+      if (option.hidden) {
```
EndangeredMassa, line 98: **question (non-blocking):** Why are only non-strings allowed to be hidden?‏

packages/cli/src/commands/integration/wizard.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const supportedUIControls = new Set([
+  'input',
+  'select',
+  'region',
+  'vercel-region',
+  'multi-region',
+  'multi-vercel-region',
+]);
```
EndangeredMassa: **suggestion (non-blocking):** move near top

Can we move this near the top of the file? It's easy to get lost here in the middle of functions.‏

EndangeredMassa (commented):
I don't have much context in the marketplace itself, but I took a good walk through.

I didn't find any explicitly blocking issues, but I would like to see these comments addressed in follow up PRs.

I want someone else from Zero Config to review before approving. We've been working towards a text control better rendering pattern. I want to make sure the changes here are in that rough direction.

packages/cli/src/commands/integration/add.ts
lukahartwig: Done

packages/cli/src/commands/integration/add.ts
lukahartwig: Lost during refactoring. I added it again, and also added a bunch more error handling.

packages/cli/src/commands/integration/add.ts
lukahartwig: It's an invariant of the system that a team can only have a single installation of a marketplace integration. I'll add a check and bail out. Should be unlikely but can't hurt.

packages/cli/src/commands/integration/add.ts
lukahartwig: Added some comments in the code. TL;DR we only have the necessary APIs for storage atm. If don't make this check and we add another category to the Marketplace the CLI would break.

We will eventually add generic APIs so this check isn't necessary anymore.

packages/cli/src/commands/integration/add.ts
lukahartwig: Yes, very valid point.

packages/cli/src/commands/integration/add.ts
lukahartwig: Done

packages/cli/src/commands/integration/add.ts
lukahartwig: Done

packages/cli/src/commands/integration/add.ts
lukahartwig: Done

packages/cli/src/commands/integration/add.ts
lukahartwig: Done

packages/cli/src/commands/integration/wizard.ts
lukahartwig: Done

packages/cli/src/commands/integration/wizard.ts
lukahartwig: Done

packages/cli/src/commands/integration/wizard.ts
lukahartwig: Fixed

packages/cli/src/commands/integration/wizard.ts
lukahartwig: Done

packages/cli/src/commands/integration/wizard.ts
lukahartwig: I added the prefix to all the assertions, but should we also add it to validation errors? These are user-facing, right?

packages/cli/src/commands/integration/wizard.ts
lukahartwig: The type of `option` looks like this `string | { hidden?: boolean; label: string; value: string }`.

So yeah, when it's just a string, it can't be hidden.

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,116 +1,154 @@
...
+    client.output.error('Cannot install more than one integration at a time');
     return 1;
   }
 
   const integrationSlug = args[0];
 
-  const user = await getUser(client);
-  const teamId = client.config.currentTeam ?? user.defaultTeamId;
+  if (!integrationSlug) {
+    client.output.error('One argument is required');
```
EndangeredMassa: **suggestion (soft-blocking):** Can we make this message easier to understand when output?

```suggestion
    client.output.error(`Error: You must pass an integration slug.`);
```

What do we call the integration identifier publicly?

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,116 +1,154 @@
...
-  if (projectLink?.status === 'error') {
-    return projectLink.exitCode;
+  const teamInstallations = installations.filter(
+    install =>
+      install.ownerId === team.id && install.installationType === 'marketplace'
+  );
+
+  if (teamInstallations.length > 1) {
+    client.output.error(
+      `Found more than one existing installation of ${integration.name}. Please contact Vercel Support.`
```
EndangeredMassa: **suggestion (soft-blocking):** Can we point to https://vercel.com/help ?

```suggestion
      `Found more than one existing installation of ${integration.name}. Please contact Vercel Support at https://vercel.com/help`
```

packages/cli/src/commands/integration/wizard.ts
EndangeredMassa: These are user facing. So we do want to consider how that will look. I could go either way on the prefix, but I like the updated message contents.

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,116 +1,154 @@
...
+import {
+  fetchIntegration,
+  fetchInstallations,
+  fetchBillingPlans,
+  provisionStoreResource,
+  connectStoreToProject,
+} from './client';
 
 export async function add(client: Client, args: string[]) {
   if (args.length > 1) {
```
EndangeredMassa, line 28: **suggestion (blocking):** Check args more carefully

This check prevents invocations like:

```
vc i supabase --cwd=some/path
```

packages/cli/src/commands/integration/add.ts
lukahartwig: Integration slug is probably the right term for now.

EndangeredMassa (changes requested):

packages/cli/src/commands/integration/add.ts
lukahartwig: Very good catch. Should be fixed now.

packages/cli/src/commands/integration/add.ts
```diff
@@ -1,116 +1,154 @@
...
-    integration.id,
-    product.id,
-    projectLink?.project?.id
+  const installation = teamInstallations[0] as
+    | IntegrationInstallation
+    | undefined;
+
+  client.output.log(
+    `Installing ${chalk.bold(product.name)} by ${chalk.bold(integration.name)} under ${chalk.bold(contextName)}`
   );
```
EndangeredMassa, line 102: **suggestion (non-blocking):** ‏consider adding spinner before this

I just tried this locally:

```
vc --cwd=/Users/smassa/source/demo/nextjs i supabase
```

And waited over 10 seconds to see any output. Will this be common? We may want a spinner here.

EndangeredMassa (approved):

packages/cli/src/commands/integration/add.ts
lukahartwig: Interesting, it has been fast for me so far. We'll have to monitor that. I'd rather make the APIs faster, but there are some provider APIs that might be slow. I'll check that and might do a follow up PR.

Issue #11543, opened 2024-05-04
Latest version of cli is pulling in insecure packages that have available patches


G-Rath:
The latest version of the `vercel` cli is pulling in packages with known vulnerabilities that have available patches.

Current vulnerabilities:
  - [ ] `path-to-regexp` v6.1.0, via `@vercel/routing-utils` and `@vercel/remix-builder`
     - (https://github.com/vercel/vercel/pull/12131, https://github.com/vercel/vercel/pull/12734 -> reverted: https://github.com/vercel/vercel/pull/12746)
     - GHSA-9wv6-86v2-598j
  - [x] `tar` v4.4.18, via `@vercel/fun` (patched in v1.1.5, just waiting for https://github.com/vercel/vercel/pull/13070 to land)
    - GHSA-f5x3-32g6-xq36
  - [ ] `undici` v5.28.4, via `@vercel/node` (waiting on https://github.com/vercel/vercel/pull/13380 to be landed and released)
    - GHSA-c76h-2ccp-4975
    - GHSA-cxrh-j4jr-qwg3
  - [ ] `esbuild` v0.14.47, via `@vercel/node` and `@vercel/gatsby-plugin-vercel-builder`
    - GHSA-67mh-4wv8-2f99

While I don't believe any of these are exploitable in the context of this cli, they are a nuisance since non-breaking patches are available and security policies can make these expensive to ignore.

<details>
<summary>`npm audit` output as of 2025-06-10</summary>

```
# npm audit report

esbuild  <=0.24.2
Severity: moderate
esbuild enables any website to send any requests to the development server and read the response - https://github.com/advisories/GHSA-67mh-4wv8-2f99
fix available via `npm audit fix --force`
Will install vercel@25.2.0, which is a breaking change
node_modules/esbuild
  @vercel/gatsby-plugin-vercel-builder  *
  Depends on vulnerable versions of esbuild
  node_modules/@vercel/gatsby-plugin-vercel-builder
    @vercel/static-build  >=1.3.0
    Depends on vulnerable versions of @vercel/gatsby-plugin-vercel-builder
    node_modules/@vercel/static-build
      vercel  >=25.2.1
      Depends on vulnerable versions of @vercel/node
      Depends on vulnerable versions of @vercel/remix-builder
      Depends on vulnerable versions of @vercel/static-build
      node_modules/vercel
  @vercel/node  >=2.3.1
  Depends on vulnerable versions of esbuild
  Depends on vulnerable versions of path-to-regexp
[...]

PR #12131, opened 2024-09-17
[node][routing-utils] Update `path-to-regexp` to v6.3.0

G-Rath:
This addresses https://github.com/advisories/GHSA-9wv6-86v2-598j and also deduplicates packages since previously two different exact versions of `path-to-regexp` were being required.

Relates to #11543


ajschmidt8:
@trek @EndangeredMassa, can we get this merged?

trek:
Opened https://github.com/vercel/vercel/pull/12239 to verify tests (external contributors don't currently have access to secrets required for tests)

trek:
@G-Rath it seems like this causes failing tests in `routing-utils`. That pinned version might have been there for a reason. Trying to track it down.

packages/routing-utils/test/superstatic.spec.ts
```diff
@@ -938,7 +938,7 @@ test('convertRewrites', () => {
 test('convertHeaders', () => {
   const actual = convertHeaders([
     {
-      source: '(.*)+/(.*)\\.(eot|otf|ttf|ttc|woff|font\\.css)',
+      source: '(.*)/(.*)\\.(eot|otf|ttf|ttc|woff|font\\.css)',
```
EndangeredMassa, line 941: This fix (that @jeffsee55 tried before--not sure why it passes now) does pass the test suite.

This `+` doesn't make sense to me. I was trying to figure out if it was intentional. So, I checked with the original author of this test in case they remembered by this extra `+` was here, but they don't. They think it was a mistake.

I think this is a fine change for this PR. We'll have to discuss internally how we roll out this breaking change, though.

EndangeredMassa (approved):

EndangeredMassa:
I'm going to keep working on this over here: https://github.com/vercel/vercel/pull/12734

Our CI is having trouble running tests for external-to-the-org PRs. Sorry about that.

PR #12136, opened 2024-09-18
chore(cli): bump chokidar to v4

v1rtl:
This pull request bumps chokidar to v4, removing 11 dependencies. The change is backwards compatible since it only watched current working directory and supports node 14+, while the CLI node target is 16+


EndangeredMassa:
Thanks for the PR!

We're not prioritizing this change right now, but we may come back to it.

paulmillr:
@EndangeredMassa why? Because you don’t care about supply chain attacks and security? The change removes 13 deps by different authors from your build

PR #12142, opened 2024-09-19
docs: fix geolocation func docs var mismatch

huskyjp:
Correct geolocation example in Vercel functions documentation


smaeda-ks (approved):

smaeda-ks:
@huskyjp Thanks!

PR #12166, opened 2024-09-24
[next]: fix defaultLocale redirect when using domains pattern

ztanner:
The redirect rule for `defaultLocale` is incorrectly matching the `/index` route resulting in a intermediary rewrite to `/<locale>/index`, which then falls through to the trailing slash redirect. This causes a 308 to the default locale for the domain even though the user would already be on the correct path.

This splits the existing `$wildcard/$1` handling into two separate routes, one for the index route and another for everything else, so that it is handled properly. 

This cannot be tested as it requires assigning a domain. However, I validated this fixes it in [this deployment.](https://next-js-locale-bug-repro.vercel.app/)


packages/next/src/server-build.ts
```diff
@@ -1721,6 +1721,24 @@ export async function serverBuild({
         ? [
             // Handle auto-adding current default locale to path based on
             // $wildcard
+            // This is split into two rules to avoid matching the `/index` route
```
```suggestion
            // This is split into two rules to avoid matching the `/index` route as it causes issues with trailing slash redirect
```

packages/next/src/server-build.ts
```diff
@@ -1721,6 +1721,24 @@ export async function serverBuild({
...
+              )}(?!(?:_next/.*|${i18n.locales
+                .map(locale => escapeStringRegexp(locale))
+                .join('|')})(?:/.*|$))$`,
+              // we aren't able to ensure trailing slash mode here
+              // so ensure this comes after the trailing slash redirect
+              dest: `${
+                entryDirectory !== '.'
+                  ? path.posix.join('/', entryDirectory)
+                  : ''
+              }$wildcard`,
```
```suggestion
              }$wildcard${trailingSlash ? '/' : ''}`,
```

packages/next/src/index.ts
```diff
@@ -2348,6 +2348,24 @@ export const build: BuildV2 = async buildOptions => {
...
+              )}(?!(?:_next/.*|${i18n.locales
+                .map(locale => escapeStringRegexp(locale))
+                .join('|')})(?:/.*|$))$`,
+              // we aren't able to ensure trailing slash mode here
+              // so ensure this comes after the trailing slash redirect
+              dest: `${
+                entryDirectory !== '.'
+                  ? path.posix.join('/', entryDirectory)
+                  : ''
+              }$wildcard`,
```
ijjk:
```suggestion
              }$wildcard${trailingSlash ? '/' : ''}`,
```

packages/next/src/index.ts
```diff
@@ -2348,6 +2348,24 @@ export const build: BuildV2 = async buildOptions => {
         ? [
             // Handle auto-adding current default locale to path based on
             // $wildcard
+            // This is split into two rules to avoid matching the `/index` route
```
ijjk:
```suggestion
           // This is split into two rules to avoid matching the `/index` route as it causes issues with trailing slash redirect
```

ijjk (approved):

PR #12167, opened 2024-09-24
Run `vitest-unit` tests with Node.js v16, v18 and v20

TooTallNate:
Expands the unit testing matrix to run the `vitest-unit` tests for Node v18 and v20 (as well as v16, which was already previously being tested).


.github/workflows/test-18.yml
```diff
@@ -102,7 +102,7 @@ jobs:
           echo "misses=$TURBO_MISS_COUNT" >> $GITHUB_OUTPUT
       - name: 'Upload Test Report to Datadog'
         if: ${{ steps['turbo-summary'].outputs.misses != '0' && !cancelled() }}
-        run: 'npx @datadog/datadog-ci@2.18.1 junit upload --service vercel-cli .junit-reports'
+        run: 'npx @datadog/datadog-ci@2.36.0 junit upload --service vercel-cli .junit-reports'
```
TooTallNate, line 105: This was only for consistency with the `test.yml` file.

.github/workflows/test.yml
```diff
@@ -108,7 +107,7 @@ jobs:
           DD_ENV: ci
 
   summary:
-    name: Summary (Node 16)
+    name: Summary
```
TooTallNate, line 110: I think we'll need to update the "Required Checks" configuration on the repo since this was renamed.

packages/cli/test/unit/util/build/import-builders.test.ts
```diff
@@ -146,30 +146,6 @@ describe('importBuilders()', () => {
     }
   );
 
-  it.skipIf(isWindows)(
```
TooTallNate, line 149: Removed this old test because it seems that newer versions of npm do not print deprecation warnings.

onsclom (approved):

trek (approved):

.github/workflows/test.yml
EndangeredMassa: This will clash with the general "Summary" check that is required. We should call this  "Summary (Node by Version)" or something more unique like that.

PR #12171, opened 2024-09-24
[client] Remove `@zeit/fetch` and always use `node-fetch` instead

onsclom:
[`@zeit/fetch` is archived](https://github.com/vercel/fetch) and it causes deprecation warnings for `util._extend` usage.


onsclom:
@TooTallNate, curious what you think of this to solve the `util._extend` usage deprecation warning. [vercel/fetch](https://github.com/vercel/fetch) is archived and the latest version causes this deprecation warning.

- this solves the deprecation warnings and tests are successful
- we lose things `@zeit/fetch` does like retrying and DNS caching (how important are these?)


Other possibilities:
- vendor in `@zeit/fetch` and it's deps to try and solve `util._extend` usage
- find another library that provides these things?

TooTallNate:
I'm good with this approach. DNS caching shouldn't be an application-level concern, and was from a time when Node.js was a lot less mature.

.changeset/red-toys-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/client': patch
+---
+
+[client] Replace `@zeit/fetch` usage with node-fetch
```
TooTallNate:
```suggestion
Replace `@zeit/fetch` usage with `node-fetch`
```

TooTallNate (dismissed):

trek (dismissed):

trek (approved):

TooTallNate (approved):

PR #12173, opened 2024-09-25
Add base telemetry behavior and one demonstration of its use

trek:
This PR is a the first in a series of PRs intended to add telemetry to the Vercel CLI. We intended to track events of
* command invocation
* subcommand invocation
* argument use (but not their their actual values)¹
* option/flag use (but not any actual values)¹

The hope is we can use this data to better understand customer usage of the CLI to:

* remove unused commands
* introduce new commands that might simplify customer workflows
* introduce new features like tab-completion, "did you mean..." suggestions on mistyped commands, etc.

## The Structure of Telemetry Tracking
This structure is heavily cribbed from two other Vercel projects that are already tracking metrics: the `next` and `turbo` CLIs. Specifically we drew heavily from the [`turbo-telemetry` package](https://github.com/vercel/turborepo/tree/main/packages/turbo-telemetry/src). However, the Vercel CLI's code structure differs from these two project that necessitates slightly different code organization.

This PR proposes the following structure:

### The Telemetry Classes
A base telemetry client (`TelemetryBaseClient`) provides a small public interface to track global command (e.g. `help`) and global option (e.g. `--cwd`) tracking, some private methods to for tracking specific _kinds_ of events (e.g. "a flag was used") and acts as parent class to command/subcommand-specific child classes.

The command/subcommand-specific child classes are intended to be defined and later called within the invocation of each command and subcommand. The public interface to these classes strictly limits what is allowed to be tracked at each level and what data we are allowed track.

The subclasses are expected to implement functions that match the command/argument/option groups used and follow a naming pattern of:

* `trackCliSubcommand<SubcommandName>`
* `trackCliArgument<ArgumentName>`
* `trackCliOption<OptionName>`
* `trackCliFlag<FlagName>`

These methods call to private methods on the base class that format the data and add a single event to the event store.

### The Event Store
The Vercel CLI is structured with a Russian doll-style nested calls where the CLI's `main` function calls a outer `<command>` function. This function calls to an (optional) inner `<subcommand>` function where the main work occurs. arguments, options, and flags may be parsed at any layer.

Because we want to track the entirety of an invocation, we need a structure that is passed to each function call. We currently use a `client` object for this case and this PR adds a new property of `telemetryEventStore: TelemetryEventStore` on the client. This store is created once at the outermost call site and passed as part of client to each inner function call.

At the end of a cli invocation, we'll have an opportunity to flush these events to Vercel's existing metrics tracking systems. This flushing is currently unimplemented and the store's `save` method only logs if in debug mode.

### Testing
This overall structure creates a nice testing pattern. Our `MockClient` instance that is passed to command functions under test now has a `telemetryEventStore` set at initialization to an instance of `MockTelemetryEventStore`. This mock store is itself a subclass of `TelemetryEventStore` with an overridden `save` method that is a no-op and a `reset` method called during test cleanup to zero out any tracked events accumulated during a run.

This allows us to unit test what the store looks like after a completed CLI invocation. Right now, this is just using a combination of vitest's `toMatchObject` and `objectContaining`, but we could write our own custom matcher or helpers to make this bit more succinct. 

This PR includes the tracking added for calls to `vercel domains transfer-in --code=acode` to offer an example of all of this structure in use.

---
¹ With a few minor exceptions of which value in a finite collections that are Vercel defined was passed. For example, we _would_ track which integration was supplied to `vercel integration add [integration name]` but would _not_ track which domain name was passed to `vercel domains add [domain]`.


packages/cli/src/util/telemetry/README.md
```diff
@@ -0,0 +1,5 @@
+# Vercel CLI Telemetry
```
trek, line 1: Intentionally left blank. Since the review of this PR may change how this works, I'm defering this to a follow up PR

packages/cli/src/util/get-subcommand.ts
```diff
@@ -24,6 +24,7 @@ export default function getSubcommand(
   }
   return {
     subcommand: config.default,
+    subcommandActual: '',
```
erikareads: do we want to set this to `'default'`?

erikareads (dismissed):

packages/cli/src/commands/domains/index.ts
```diff
@@ -36,14 +37,20 @@ export default async function main(client: Client) {
...
+    },
+  });
 
   if (parsedArgs.flags['--help']) {
     output.print(help(domainsCommand, { columns: client.stderr.columns }));
     return 2;
   }
 
-  const { subcommand, args } = getSubcommand(
+  const { subcommand, args, subcommandActual } = getSubcommand(
```
TooTallNate: "Actual" doesn't convey to me what it means. Is that "original" or "resolved"?

packages/cli/src/commands/domains/transfer-in.ts
```diff
@@ -13,6 +13,8 @@ import confirm from '../../util/input/confirm';
 import isRootDomain from '../../util/is-root-domain';
 import { getCommandName } from '../../util/pkg-name';
 
+import { DomainsTransferInTelemetryClient } from '../../util/telemetry/commands/domains/transfer-in';
```
TooTallNate, line 16: So do we need telemetry subclasses for literally every sub-command and sub-sub-command? Seems very verbose to me at first glance, but I'll keep reading the code - maybe it will make more sense once I read through it all.

packages/cli/src/util/telemetry/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  trackFlagHelp() {
+    this.trackCliFlag('help');
+  }
+}
+
+export class TelemetryEventStore {
+  private events: Event[];
+  private output: Output;
+  private isDebug: boolean;
```
TooTallNate, line 104: Maybe leverage `output.debugEnabled` instead?

packages/cli/src/index.ts
```diff
@@ -702,6 +719,8 @@ const main = async () => {
     return 1;
   }
 
+  // specifically don't await this, we want to fire and forget
+  telemetryEventStore.save();
```
TooTallNate, line 723: If the goal is to allow CLI to still exit quickly, then we may want to consider flushing events in a detached child process.

TooTallNate (commented):
Overall seems like a reasonable approach, however it seems to me like a very manual and error prone (oops, we forgot to add telemetry to x.y.z. command) process. I'm not a huge fan of the many telemetry-related subclasses that it appears will need to be added, and I wonder why this can't be done in a more centralized way that would encompass all invocations instead of having to integrate each subcommand explicitly.

packages/cli/src/commands/domains/transfer-in.ts
erikareads: The benefit of this approach is that we have a hard-coded list of methods for things that we track. This makes it easier to security audit for ourselves and for our customers.

packages/cli/src/util/telemetry/index.ts
trek: At some point, yes. I think we want to not have this log until we were actually using it at the end of the project.

packages/cli/src/index.ts
erikareads: Yeah, that's what we'll want to do.

packages/cli/src/util/telemetry/index.ts
trek: For example, if someone enables debug to figure out why a build is failing, I don't think the telemetry debug is useful at that point, so wanted this to be toggled separately.

packages/cli/src/commands/domains/index.ts
trek: We can go with "original". "resolved" is just "subcommand" here.

packages/cli/src/index.ts
trek: Agreed. We're a bit away from actual submission but that will be part of the work for sure.

erikareads (approved):

PR #12174, opened 2024-09-25
[cli] Replace `psl` with `tldts`

onsclom:
`psl` is one of the deps causing the `punycode` deprecation warning on Node 22. [The newest version includes the deprecation warning](https://github.com/lupomontero/psl/issues/323), so we can't just update `psl`.

Inspired by https://github.com/salesforce/tough-cookie/pull/346.


trek (approved):

TooTallNate (approved):

packages/cli/src/commands/certs/issue.ts
```diff
@@ -175,14 +175,12 @@ async function runStartOrder(
   );
   const [header, ...rows] = dnsTable(
     pendingChallenges.map(challenge => {
-      const parsedDomain = parse(challenge.domain);
-      if (parsedDomain.error) {
+      const subdomain = getSubdomain(challenge.domain);
+      if (!subdomain) {
```
remusao, line 179: Hi, I noticed that the behavior after switching to `tldts` is different than with `psl`. There is now a migration guide which explains how to switch from `psl` to `tldts` while preserving the same behavior: https://github.com/remusao/tldts?tab=readme-ov-file#migrating-from-other-libraries

In particular, I think you might want to use the `{ allowPrivateDomains: true }` option to consider suffixes from the private section as well (as opposed to ICANN only).

I hope that helps!

PR #12178, opened 2024-09-26
[cli] correctly call to create new deployment when promoting preview deployment

trek:
We previously removed the behavior at the API of directly promoting a preview deployment to production. This corrects the CLI behavior to create a _new_ production deployment with user permission.

Also verified manually.

### DIFF
--- a/packages/cli/src/commands/promote/request-promote.ts
+++ b/packages/cli/src/commands/promote/request-promote.ts
@@ -6,6 +6,10 @@ import ms from 'ms';
 import promoteStatus from './status';
 import confirm from '../../util/input/confirm';
 
+interface DeploymentCreateResponsePartial {
+  inspectorUrl: string;
+  id: string;
+}
 /**
  * Requests a promotion and waits for it complete.
  * @param {Client} client - The Vercel client instance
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/promote/request-promote.ts
```diff
@@ -32,22 +36,49 @@ export default async function requestPromote({
...
+          name: project.name,
+          target: 'production',
+          meta: {
+            action: 'promote',
+          },
+        },
+        accountId: deployment.ownerId,
+        method: 'POST',
+      }
+    )) as DeploymentCreateResponsePartial;
```
trek, line 69: Don't love this, but our HTTP response types are all over the place. This at least lets TS stop complaining.

EndangeredMassa (approved):

packages/cli/test/unit/commands/promote/index.test.ts
```diff
@@ -135,9 +137,32 @@ describe('promote', () => {
...
 
     it('should promote a preview deployment when user says yes', async () => {
+      const inspectorUrl = `https://vercel.com/user/${projectName}/abcdefghijlmo`;
+
       const { cwd, previousDeployment } = initPromoteTest({
         deploymentTarget: 'preview',
       });
+      let createWasCalled = false;
+
+      client.scenario.post(
```
trek, line 147: Placed the scenario in-line because I wanted to assert the exact body submitted as well.

packages/cli/test/unit/commands/promote/index.test.ts
```diff
@@ -154,28 +179,49 @@ describe('promote', () => {
...
 
     it('should promote a preview deployment with --yes', async () => {
+      const inspectorUrl = `https://vercel.com/user/${projectName}/abcdefghijlmo`;
+
+      let createWasCalled = false;
       const { cwd, previousDeployment } = initPromoteTest({
         deploymentTarget: 'preview',
       });
+
+      client.scenario.post(
```
trek, line 197: Placed the scenario in-line because I wanted to assert the exact body submitted as well.

jeffsee55 (approved):

onsclom (approved):

PR #12180, opened 2024-09-26
Add support for tracking Continuous Integration vendors with telemetry

erikareads:
Add a method for tracking whether the CLI is running in a Continuous Integration system, and if so, which one.


packages/cli/src/util/telemetry/base.ts
```diff
@@ -7,4 +7,406 @@ export class TelemetryBaseClient extends TelemetryClient {
...
+      }
+    } else {
+      return undefined;
+    }
+  }
+}
+
+// list copied from Turborepo
+// https://github.com/vercel/turborepo/blob/663d3ffdc5b1a4a93353ed408c0b6a653fd89a9f/crates/turborepo-ci/src/vendors.rs#L33C13-L678C14
+export const VENDORS: Vendor[] = [
```
TooTallNate, line 51: We may want to consider using [`ci-info`](https://www.npmjs.com/package/ci-info) package instead of maintaining this ourselves.

erikareads: After your suggestion, @trek and I checked out [`ci-info`](https://www.npmjs.com/package/ci-info). The main problem is that it decides whether you're in CI at package _import_ time, which makes it impossible to test well. 

In addition, the most useful feature, `ci.name` does not have a guarantee of a stable name, which means that we might arbitrarily start collecting a previous CI system under a new name without warning at some point in the future, making it difficult or impossible to compare historical and new data.

For these reasons, along with the relative small size of the CI checking function, I think it's worthwhile to retain a locally implemented, testable version of the logic.

trek: We saw that some other Vercelians have committed to the project though. We might be able to PR having stable identifiers and swap out our custom solution for the library if we can get a consistent value in the future.

trek: https://github.com/watson/ci-info/issues/130 we can see if the author is amenable

EndangeredMassa (approved):

packages/cli/src/util/telemetry/base.ts
sibiraj-s: You could now do `ci.id` to get the vendor constant. available in [v4.1.0](https://github.com/watson/ci-info/releases/tag/v4.1.0)

PR #12190, opened 2024-09-27
Add tracking for machine information

erikareads:
Add tracking events for `platform`, `arch`, and `cpu_count`.


packages/cli/src/util/telemetry/index.ts
```diff
@@ -90,6 +91,27 @@ export class TelemetryClient {
     });
   }
 
+  protected trackCPUs() {
```
jeffsee55, line 94: Is this a styling decision that we're forcing child classes to implement these? Wondering if there's value in this hierarchy vs just putting these functions directly on `RootTelementClient`, with no ancestor implementation.

jeffsee55 (approved):

packages/cli/src/util/telemetry/index.ts
erikareads: The child class doesn't have access to the `track` method, which is intentional.

PR #12206, opened 2024-10-02
Create devcontainer.json

Thompson1985:


Thompson1985 (commented):
Thank you

trek:
Hi @Thompson1985, we'd need a explanation of what this would be used for. It looks like configuration specific to your editor and workflow which we wouldn't add to the project for everyone.

PR #12207, opened 2024-10-02
Add telemetry enable/disable command

jeffsee55:
Adds `telemetry status | enable | disable` commands to the CLI. `enable | disable` write to the CLI config file, which stores other things like `currentTeam`. 

One thing that this doesn't test is whether or not the `client` property assigns `config.telemetry.enabled` when invoked, though I tested it manually. The way we do test setup doesn't capture that portion of our CLI init phase, but it's a pretty straightforward bit of logic where it reads the JSON file and parses it into the `client.config` in [its constructor](https://github.com/vercel/vercel/blob/f48f01047d310c7902b283c6025a13dab6c2a57d/packages/cli/src/index.ts#L280).


packages/cli/src/commands/telemetry/command.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      name: 'enable',
+      description: 'Enables telemetry collection',
+      arguments: [],
+      options: [],
+      examples: [],
+    },
+    {
+      name: 'enable',
```
trek: s/enable/disable

trek (approved):

trek (changes requested):
sorry, meant my previous comment to be a blocker.

trek (approved):

erikareads (approved):

PR #12209, opened 2024-10-02
Add support for telemetry debugging via VERCEL_TELEMETRY_DEBUG

jeffsee55:
Specify `VERCEL_TELEMETRY_DEBUG =1 vc alias list` to view a printout of the telemetry events, rather than sending them off for collection.

This no longer writes to `output.debug` as this behavior is separate from the `--debug` global flag.


trek (dismissed):

packages/cli/src/util/telemetry/index.ts
```diff
@@ -127,9 +129,11 @@ export class TelemetryEventStore {
 
   save() {
     if (this.isDebug) {
-      this.output.debug(`${LogLabel} Flushing Events`);
+      // Intentionally not using `this.output.debug` as it will
+      // not write to stderr unless the is run with `--debug`
```
onsclom:
```suggestion
      // not write to stderr unless it is run with `--debug`
```

onsclom (dismissed):

erikareads (commented):
Why does this say files changed `2`? I don't see the change that adds support for the environment variable?

jeffsee55:
> Why does this say files changed `2`? I don't see the change that adds support for the environment variable?

Not sure what happened there, bad merge from main I guess, fixed!

packages/cli/src/util/telemetry/index.ts
```diff
@@ -156,7 +156,9 @@ export class TelemetryEventStore {
 
   save() {
     if (this.isDebug) {
-      this.output.debug(`${LogLabel} Flushing Events`);
+      // Intentionally not using `this.output.debug` as it will
+      // not write to stderr unless it is run with `--debug`
+      this.output.log(`${LogLabel} Flushing Events`);
       this.events.forEach(event => {
         this.output.debug(JSON.stringify(event));
```
erikareads: Shouldn't this be `this.output.log` as well?

jeffsee55: Yes! It was https://github.com/vercel/vercel/pull/12209/commits/64acfdd46b23fd98b88b4c0cd44e8d224becc753, I really don't know what happened there 🤔

erikareads: We probably also want to make the follow-up conditional an `else if` to keep in line with our plan to have debugging disable sending to server discussion

erikareads (approved):

PR #12211, opened 2024-10-02
[build-utils] Disable corepack in unsupported Turborepo

onsclom:
## Problem

Turborepo does not pass `COREPACK_HOME` to internal libraries. This means Turborepo projects using corepack may run into `EEXIST: file already exists` errors. [turbo@2.1.3 has `COREPACK_HOME` in default passthrough](https://github.com/vercel/turborepo/pull/9151), but we need to handle older existing Turborepo projects better for a couple reasons:

1. New Turborepo projects can be created today that use corepack, and they can hit this confusing `EEXIST` error.

2. We want to fix corepack detection in monorepos without breaking existing Turborepo projects again.
  a. https://github.com/vercel/vercel/pull/12099

## Solution

When a Turborepo project does not support `COREPACK_HOME`, we disable corepack. Projects that need corepack disabled have the following criteria:
- Use turbopack (detected by having `turbo` in `package.json#devDependencies`)
- Use a turbo version lower than 2.1.3 (using the minimum of the range in `package.json#devDependencies.turbo`)
- Do not include `COREPACK_HOME` in `turbo.json#globalPassThroughEnv`

In the case we disable corepack, we display a helpful warning telling users how to fix their project:

"Warning: Disabling corepack because it may break your project. To use corepack, either upgrade to `turbo@2.1.3+` or include `COREPACK_HOME` in `turbo.json#globalPassThroughEnv`."


jeffsee55 (commented):
Awesome tests!

jeffsee55 (approved):

trek (approved):

PR #12212, opened 2024-10-02
[cli] Add tracking for version

erikareads:
Adds an event that tracks which version of the CLI is being used.


packages/cli/test/unit/index.test.ts
```diff
@@ -1,11 +1,38 @@
...
+import './test/mocks/matchers';
 
 import { Output } from '../../src/util/output';
 import { TelemetryEventStore } from '../../src/util/telemetry';
 import { TelemetryBaseClient } from '../../src/util/telemetry/base';
 
 describe('main', () => {
   describe('telemetry', () => {
+    describe('version', () => {
+      it('tracks version', () => {
```
EndangeredMassa, line 33: **suggestion (non-blocking):** ‏test the negative case as well

EndangeredMassa (dismissed):

packages/node/test/unit/dev.test.ts
```diff
@@ -117,7 +117,8 @@ async function withDevServer(
         { runningTimeout: 300 }
       ));
 
-    test('with `waitUntil` from context rejecting a promise ', () =>
```
trek, line 120: not blocking, but I think I'd generally prefer this workaround to be its own PR. But we can discuss this process on Monday!

trek (dismissed):

trek (approved):

Thompson1985 (approved):

PR #12229, opened 2024-10-04
[cli] set telemetry enabled to false with `$VERCEL_TELEMETRY_DISABLED`

trek:
When `$VERCEL_TELEMETRY_DISABLED` the store is disabled even if configuration is `true`

### DIFF
--- a/packages/cli/test/unit/index.test.ts
+++ b/packages/cli/test/unit/index.test.ts
@@ -45,6 +45,7 @@ describe('main', () => {
         { key: 'cpu_count', value: '1' },
       ]);
     });
+
     it('tracks platform', () => {
       vi.spyOn(os, 'platform').mockImplementation(() => 'linux');
       const output = new Output(process.stderr, {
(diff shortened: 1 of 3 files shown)


packages/cli/test/unit/index.test.ts
```diff
@@ -140,6 +142,28 @@ describe('main', () => {
...
+          noColor: false,
+        });
+
+        const telemetryEventStore = new TelemetryEventStore({
+          isDebug: true,
+          output,
+          config: configThatWillBeIgnoredAnyway,
+        });
+
+        expect(telemetryEventStore.enabled).toBe(false);
```
onsclom, line 163:
```suggestion
        expect(telemetryEventStore.enabled()).toBe(false);
```

onsclom (dismissed):

packages/cli/test/unit/index.test.ts
trek: Oh, weird, I just assumed we'd have made that a getter. Going to make it getter!

erikareads (approved):

PR #12233, opened 2024-10-04
ECO-443: add `vc integration open` command to cli

JakeUskoski:
## Summary

This PR adds a new sub-command to integration 'open' which allows a user to directly open the integration provider's dashboard in the browser. This command is only compatible with Marketplace integrations.

## Background

The new marketplace integrations set up & manage accounts with the provider's platform on behalf of a vercel team when they are installed by said team, creating a "native integration." These dashboards are then meant to be used by users on the team to manage the integrations in-depth through their first-party development experience.

In order to make the bridge to managing a resource from an integration simpler, we're looking to add this command to allow users to go directly to a given integration dashboard from their project/team's command line.

## How to test

Build and link Vercel CLI locally. The 3 integrations supported today are:

```
vc integration open [supabase/redis/edgedb]
```


Thompson1985 (approved):

Thompson1985 (approved):

packages/cli/src/commands/integration/open-integration.ts
```diff
@@ -0,0 +1,58 @@
...
+  try {
+    configuration = (
+      await fetchMarketplaceIntegrations(client, integrationSlug)
+    )[0];
+  } catch (error) {
+    client.output.error(
+      `Failed to fetch configuration for ${chalk.bold(`"${integrationSlug}"`)}: ${(error as Error).message}`
+    );
+    return 1;
+  }
```
RPDeshaies: suggestion: wrap in a `getConfiguration` method

RPDeshaies (dismissed):

alex-vercel (dismissed):

packages/cli/src/commands/integration/types.ts
```diff
@@ -37,6 +37,19 @@ export interface IntegrationProduct {
...
+export interface Configuration {
+  id: string;
+  integrationId: string;
+  ownerId: string;
+  slug: string;
+  teamId: string;
+  userId: string;
+  scopes: string[];
+  source: string;
+  installationType: string;
```
dvoytenko: `'marketplace' | ...`?

packages/cli/src/commands/integration/open-integration.ts
```diff
@@ -0,0 +1,58 @@
...
+      `No configuration found for ${chalk.bold(`"${integrationSlug}"`)}.`
+    );
+    return 1;
+  }
+
+  client.output.print(
+    `Opening the ${chalk.bold(integrationSlug)} dashboard...`
+  );
+
+  const url = new URL('/api/marketplace/sso', 'https://vercel.com');
```
dvoytenko, line 50: So, the goal is to SSO to the partner right away? I wonder if it'd be better to open our page and SSO from there? Just curious to hear thoughts: either way is possible.

packages/cli/src/commands/integration/open-integration.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  const integrationSlug = args[0];
+
+  if (!integrationSlug) {
+    client.output.error('You must pass an integration slug');
+    return 1;
+  }
+
+  const { team } = await getScope(client);
```
dvoytenko, line 21: Could you explain how a team is resolved here if a user is a member of multiple teams?

dvoytenko (changes requested):

packages/cli/src/commands/integration/types.ts
JakeUskoski: Can add 👍 this was originally just a simplified type to have some general structure. I'm considering keeping `| string` afterwards to protect the CLI from schema drift over time.

packages/cli/src/commands/integration/open-integration.ts
JakeUskoski: Team is determined by the currently set team in the CLI. You can switch between teams using `vc switch` to switch between team contexts.

packages/cli/src/commands/integration/open-integration.ts
JakeUskoski: Correct. In the original conversation thread, the interest seemed to be in opening Vercel page for the given integration, but then in the more up-to-date technical outline the desired functionality was to deep-link all the way to the provider's page.

I'm personally more in-favour of deep linking. While opening to the Vercel page would be useful for cases like upgrading a plan sometimes, or viewing usage data through our tabs, more often than not a user doing their day-to-day development will want to deal with the resource directly, which they'd do through the provider's portal rather than ours. A developer using the CLI would more likely just go straight from the Vercel dashboard to the provider's dashboard, so we're saving a click in the common use case.

packages/cli/src/commands/integration/types.ts
JakeUskoski: Ended up pulling the `installationType` up into a type to be shared across the relevant interfaces.

packages/cli/src/commands/integration/open-integration.ts
JakeUskoski: I'll wrap this slightly differently, specifically because I don't want to mask the actual `getConfiguration` concept in the client (searching by ID). Technically I'm using a `getConfigurations` endpoint in the back with some filtering, though I've only ever seen it return an array of 1 so far.

EndangeredMassa:
@JakeUskoski Those test failures were resolved on main. Updating the branch should resolve it.

packages/cli/src/commands/integration/open-integration.ts
dvoytenko: Ok. I'm ok with that in general, but just to give it one more discussion in case we can leave some room open. One caveat that's specific to our dashboard: a resource exists in the context of an account (a team), but it's also linked to the project. And that linking to projects is unique to us and could be very relevant in the context of CLI, because it shows env vars for the project. WDYT?

dvoytenko (dismissed):

packages/cli/src/commands/integration/open-integration.ts
JakeUskoski: This is an interesting and specific one, but I feel like it feeds back into the other PR comment you made on how the user knows which team they're in the context of when using the CLI.

I do expect, however, that #12251 will help to enable the user's understanding of that project-to-resource relationship.

One line we could consider, would be a future addition of an option `--vercel` to open the integration's Vercel dashboard page instead of the integration provider's page. This would enable the flow to the data you're talking about, without disrupting the typical expected flow.

packages/cli/src/commands/integration/client.ts
```diff
@@ -16,6 +17,22 @@ export async function fetchIntegration(client: Client, slug: string) {
   );
 }
 
+export async function fetchMarketplaceIntegrations(
```
trek, line 20: Not blocking but other behaviors like this typically get put in `src/util/{command-name}`. Would be happy with a follow-up PR.

JakeUskoski:
Note: `vc open` has been removed from this PR. There were blocking discussions on whether or not `vc integration open` should be able to take the `vc open` slug, as that's a highly valuable name for a slug that may see other use.

erikareads (dismissed):

trek (dismissed):

packages/cli/src/commands/integration/client.ts
JakeUskoski: Good to know, thanks! This was building on top of the prior work already there. I can migrate that entire `client.ts` file in a future PR. I may keep it around until after all 3 feature PRs are merged (or until the final feature PR, if the other two are already merged before it's ready), but it'll come down the line.

trek: Works for me!

erikareads (approved):

PR #12242, opened 2024-10-08
Revert "[build-utils] Fix corepack `packageManager` detection on monorepos"

EndangeredMassa:
Reverts vercel/vercel#12219 because it caused an issue with corepack versions being set to `latest`.

https://vercel.slack.com/archives/C07R9QVM5L1


trek (approved):

EndangeredMassa:
merging past the flakey tests, but will follow up on those right after

PR #12251, opened 2024-10-08
ECO-445: add `vc integration list` command to cli

JakeUskoski:
## Summary

This PR adds a new sub-command to integration `list` (also available as `ls`) which allows a user to view all the Marketplace resources associated with their current team. It provides the resource's name, its current status, the Product name, the Integration's slug, and the connected projects:

```
> Integrations in acme-marketplace-team:
Name                           Status             Product            Integration        Projects                                                              
redis-cyan-door                ● Available        Redis Cloud        redis              –                                                                     
redis-indigo-yacht             ● Available        Redis Cloud        redis              –                                                                     
supabase-orange-garden         ● Available        Supabase           supabase           –                                                                     
redis-indigo-jacket            ● Available        Redis Cloud        redis              –                                                                     
redis-violet-house             ● Available        Redis Cloud        redis              –                                                                     
edgedb-yellow-dog              ● Available        EdgeDB             edgedb             –                                                                     
redis-rose-door                ● Available        Redis Cloud        redis              –                                                                     
vector-gray-window             ● Available        ACME               vector             –                                                                     
vector-indigo-jacket           ● Available        ACME               vector             –                                                                     
vector-blue-school             ● Available        ACME               vector             –                                                                     
redis-orange-house             ● Available        Redis Cloud        redis              –                                                                     
redis-green-chair              ● Available        Redis Cloud        redis              –                                                                     
redis-indigo-mountain          ● Available        Redis Cloud        redis              –                                                                     
vector-rose-garden             ● Available        ACME               vector             test-repo-4                                                           
vector-violet-kite             ● Available        ACME               vector             –                                                                     
redis-red-tree                 –                  Redis Cloud        redis              –                                                                     
vector-blue-xylophone          ● Available        ACME               vector             –                                                                     
vector-emerald-jacket          ● Available        ACME               vector             –                                                                     
redis-green-yacht              ● Available        Redis Cloud        redis              –                                                                     
vector-rose-book               ● Available        ACME               vector             test-repo-4, example-marketplace-integration, marketplace-user        
redis-gray-ocean               ● Available        Redis Cloud        redis              –                                                                     
redis-orange-car               ● Available        Redis Cloud        redis              –                                                                     
redis-red-house                ● Available        Redis Cloud        redis              test-repo-4                                                           
supabase-lime-kite             ● Available        Supabase           supabase           example-marketplace-integration, test-repo-4                          
supabase-blue-yacht            ● Available        Supabase           supabase           myffff-awesome-project                                                
edgedb-fuchsia-elephant        ● Available        EdgeDB             edgedb             test-repo-4                                                           
edgedb-lime-zebra              ● Available        EdgeDB             edgedb             test-repo-4 
```
image

## Background

This change is related to #12233. As we continue to try and improve the developer experience of interacting with the marketplace via the CLI, it's important that we offer the basic general CRUD activities seen in CLI tools of `add`, `list`, and `remove` (as well as `open` for how users perform update actions on the resources day-to-day). This PR adds the `list` functionality to compliment the already existing `add`, the soon-to-come `open`, and the coming `remove`.

Original designs for this subcommand only included the Name, Status, Product, and Integration Slug, but I also included the projects as well as the ability to filter on Integration & on the current project in order to make the list a little easier to use and interact with. A user probably only needs to care about their current project's resources (thus the `--current-project`/`-c` flag), but if they're trying to filter down on the full list, `--integration` also helps narrow down the list when debugging.

Other options that may be helpful & would make good future additions include `--status [status]` and `--project [project-name]`.

## How to test

Build and run locally (or use `pnpm dev` instead of `vc` when in the `packages/cli` folder). Note: for `--current-project` flag you need a project linked, so if you're using `pnpm dev` to test the command then you'll need to execute it directly. For this, I took the debugging output of the command being called using `pnpm dev` to find what it was calling, and used the full path call of `[ts-node from node_modules] [full path]/vercel/packages/cli/src/index.ts integration list --current-project` to test it in a different folder where I had a linked vercel project ready to test.

```
vc integration [list/ls] [optional: project-name]
vc integration [list/ls] [optional: project-name] [--integration [supabase/redis/edgedb]]
vc integration [list/ls] --all
vc integration [list/ls] --all [--integration [supabase/redis/edgedb]]
```


JakeUskoski:
Note: this PR touches files shared with #12233, and will require a merge conflict resolution between the two.

packages/cli/src/commands/integration/command.ts
```diff
@@ -34,4 +19,62 @@ export const integrationCommand: Command = {
...
+      description: 'limits the resources listed to a designated integration',
+      shorthand: 'i',
+      type: String,
+      deprecated: false,
+      argument: 'NAME',
+    },
+    {
+      name: 'current-project',
+      description: 'limits the resources listed to the current project',
+      shorthand: 'cp',
```
erikareads: shorthand flags must be one character

packages/cli/src/commands/integration/command.ts
```diff
@@ -34,4 +19,62 @@ export const integrationCommand: Command = {
...
+        `${packageName} integrations list --integration <integration>`,
+        `${packageName} integrations list --integration acme`,
+        `${packageName} integrations list -i acme`,
+      ],
+    },
+    {
+      name: 'List all resources from the current project',
+      value: [
+        `${packageName} integrations list --current-project`,
+        `${packageName} integrations list -cp`,
```
erikareads: this example will have to be updated to match fixed shorthand

erikareads (changes requested):

packages/cli/src/commands/integration/command.ts
JakeUskoski: Good to know, updated it to just `c`, thanks! 🙂

erikareads (dismissed):

packages/cli/src/commands/integration/client.ts
```diff
@@ -85,3 +86,20 @@ export async function provisionStoreResource(
     }
   );
 }
+
+export async function getStores(
```
trek: Not blocking but other behaviors like this typically get put in `src/util/{command-name}`. Would be happy with a follow-up PR.

trek (dismissed):

erikareads (dismissed):

packages/cli/src/commands/integration/client.ts
JakeUskoski: I've resolved this as well as the similar callout in #12233 in this PR. The `client.ts` file has been split apart into individual util files. Worth noting, while doing so, I also renamed this from `getStores` to `getResources`.

trek (dismissed):

packages/cli/src/commands/integration/client.ts
```diff
@@ -1,104 +0,0 @@
-import type Client from '../../util/client';
```
JakeUskoski, line 1: This file's content has been split into its individual functions and moved to the `packages/cli/util/integration/` space.

erikareads (dismissed):

packages/cli/src/commands/integration/list.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  const project = await getLinkedProject(client).then(result => {
+    if (result.status === 'linked') {
+      return result.project;
+    }
+
+    return null;
+  });
+
+  if (parsedArguments.flags['--current-project'] && !project) {
```
dvoytenko: Do we apply a similar pattern with `--current-project` in other commands?

packages/cli/src/commands/integration/types.ts
```diff
@@ -83,3 +83,20 @@ export interface BillingPlan {
   }[];
   disabled?: boolean;
 }
+
+export interface Store {
```
dvoytenko: Can we call it `Resource` right away?

packages/cli/src/util/integration/connect-store-to-project.ts
```diff
@@ -0,0 +1,18 @@
+import type Client from '../client';
+
+export async function connectStoreToProject(
```
dvoytenko: `connectResourceToProject`

dvoytenko (changes requested):

packages/cli/src/commands/integration/types.ts
JakeUskoski: 👍 It should be, I missed this while using & updating to the "resource" name everywhere else. Will pick this up tomorrow in the AM.

erikareads (dismissed):

packages/cli/src/commands/integration/list.ts
JakeUskoski: There's a challenge here around the scopes that "integration" lives in. Often commands are either untethered to a linked project at all (ex. redeploy takes a deployment URL you specify as an argument), or are in the pure context of the linked project. integration is a bit interesting because it is both team-wide and linked to projects.

`vc list` seems like it might be the closest, in that `vc list` takes argument `[project-id]` afterwards to specify which project you want to list the details of, and if you leave it blank, it uses the current linked project.

But we also have the scope above which is the team-wide list of resources, and our original outline for what we want `vc integration list` to do is to view the team-wide list. The `--current-project` flag was just my small addition to make it a little more usable.

Ignoring the option of "this is fine as-is," the immediate alternatives I can think of that would fit are:
1. Have no team scope, only ever project scope (this loses the opportunity to search across an integration and makes some debugging harder from CLI).
2. Have current project scope by default, an optional argument to specify a project ID, and add an `--all`/`--team` flag to expand to team scope (which fails if there's a project ID argument given). Further possible expansion: if there is no current project, default behaviour of no argument becomes team scope?

I'm not a fan of option 1, but I'm all right with option 2 for being more similar to unix `ls [-a]` functionality.

trek (dismissed):

packages/cli/src/commands/integration/list.ts
dvoytenko: Yes, the inverse is what i was wondering about (2). I'm ok either way, but wanted to make sure this was discussed with CLI owners.

dvoytenko (dismissed):

erikareads (dismissed):

packages/cli/src/commands/integration/list.ts
JakeUskoski: Thinking about it more, I liked option 2 more than the `--current-project` flag, and so prepared the changes for that implementation 👍 It deviates from the original technical outline, but I believe results in a better command.

erikareads (approved):

PR #12259, opened 2024-10-09
[build-utils] Use json5 for `turbo.json`

onsclom:
Turbo treats `turbo.json` as json5, not json, so comments are valid.

https://vercel.slack.com/archives/C07PRDN82P8/p1728503313203439

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -25,6 +25,7 @@ import {
 } from './node-version';
 import { readConfigFile } from './read-config-file';
 import { cloneEnv } from '../clone-env';
+import json5 from 'json5';
 
 // Only allow one `runNpmInstall()` invocation to run concurrently
 const runNpmInstallSema = new Sema(1);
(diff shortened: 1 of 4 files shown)


trek (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -447,7 +448,7 @@ async function checkTurboSupportsCorepack(
   const turboJsonPath = path.join(rootDir, 'turbo.json');
   const turboJsonExists = await fs.pathExists(turboJsonPath);
   const turboJson = turboJsonExists
-    ? (JSON.parse(await fs.readFile(turboJsonPath, 'utf8')) as TurboJson)
+    ? (json5.parse(await fs.readFile(turboJsonPath, 'utf8')) as TurboJson)
```
EndangeredMassa, line 451: **suggestion (soft-blocking):** We should also `try-catch` and decide what to do if it fails to parse for any reason: (1) throw a better error message or (2) log the warning and do something reasonable.

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -447,7 +448,7 @@ async function checkTurboSupportsCorepack(
...
     : undefined;
   return turboJson?.globalPassThroughEnv?.includes('COREPACK_HOME') || false;
```
EndangeredMassa, line 453: **suggestion (soft-blocking):** Can we add a test for parsing json with comments?‏

PR #12269, opened 2024-10-10
Add telemetry to deploy command

jeffsee55:
This adds telemetry to the deploy command. It also adds a bunch of unit tests for `deploy`, these tests mainly assert 2 things:

1. We called [`createDeploy`](https://github.com/vercel/vercel/blob/67e20a6ede34cb7372cb2158788d47b89e2f9194/packages/cli/src/util/deploy/create-deploy.ts#L11) with the expected arguments. I'm treating that module as our external interface, but perhaps we'd like these tests to capture it's behavior as well, open to suggestions. 
2. We tracked the expected event in our telemetry store.


packages/cli/src/index.ts
```diff
@@ -567,6 +567,7 @@ const main = async () => {
           func = require('./commands/certs').default;
           break;
         case 'deploy':
+          telemetry.trackCliCommandDeploy(userSuppliedSubCommand);
```
trek, line 570: Not blocking but I don't think `deploy` has subcommands? Or does it count itself as a subcommand of `vercel` with no command?

trek (dismissed):

packages/cli/src/index.ts
jeffsee55: You're right, I wasn't really paying attention to what this is supposed to store. 
The way it is right here it'll store the `value` as `deploy`. So I guess for anything that doesn't have a subcommand what value are we sending:

```
  trackCliCommandWhoami() {
    this.trackCliCommand({
      command: 'whoami',
      value: 'WHATSTHISVALUE',
    });
  }
  ```

erikareads: I think the value can just be the command again.

erikareads (dismissed):

erikareads (approved):

PR #12284, opened 2024-10-16
[next] Ensure app-paths-manifest is filtered properly

ijjk:
This ensures we properly filter the app-path-routes manifest the same we do for pages-manifest which ensures we don't try handling a page that isn't in a function group. 

x-ref: [slack thread](https://vercel.slack.com/archives/C07RY9D9C8M/p1729037558954849?thread_ts=1729033655.852429&cid=C07RY9D9C8M)


packages/next/src/server-build.ts
```diff
@@ -1112,6 +1113,24 @@ export async function serverBuild({
               }
               break;
             }
+            case 'server/app-paths-manifest.json': {
+              console.log(normalizedPages);
```
ztanner: maybe worth keeping this behind the debug flag along with the other log

ztanner (dismissed):

packages/next/src/server-build.ts
ijjk: oh these can be removed now

ztanner (approved):

PR #12286, opened 2024-10-16
[next] provide @next/request-context to Next.js handlers

lubieowoce:
Some time ago, Next.js introduced `@next/request-context` which is meant as a platform-agnostic way of providing `waitUntil` to Next.js: https://github.com/vercel/next.js/blob/086f715c27391d0559af1b993668b4effc202eaf/packages/next/src/server/after/builtin-request-context.ts#L3-L14

This PR changes the next.js builder to provide this context to handlers.

`@next/request-context` is very close to our existing `globalThis[Symbol.for("@vercel/request-context")]` (and mirrors a subset of its interface) but isn't tied to the vercel platform.

---

x-ref: removing usage of `@vercel/request-context` from Next.js (needs this change)
https://github.com/vercel/next.js/pull/71383

### DIFF
--- a/packages/next/build.mjs
+++ b/packages/next/build.mjs
@@ -18,10 +18,17 @@ await Promise.all([
   esbuild({
     entryPoints: [
       'src/legacy-launcher.ts',
-      'src/server-launcher.ts',
       'src/templated-launcher-shared.ts',
       'src/templated-launcher.ts',
     ],
+  }),
+  // server-launcher imports some modules, so we need to bundle it
+  esbuild({
+    entryPoints: ['src/server-launcher.ts'],
+    bundle: true,
+    external: ['__NEXT_SERVER_PATH__'],
+    // make sure the `// @preserve ...` comments we use for code injection stay in the same place
+    legalComments: 'inline',
   }),
   buildEdgeFunctionTemplate(),
 ]);
--- a/packages/next/scripts/build-edge-function-template.js
+++ b/packages/next/scripts/build-edge-function-template.js
@@ -37,10 +37,3 @@ async function buildNextjsWrapper() {
 }
 
 module.exports = buildNextjsWrapper;
-
-if (!module.parent) {
-  buildNextjsWrapper().catch(err => {
-    console.error(err);
-    process.exit(1);
-  });
-}
--- a/packages/next/scripts/build-edge-function-template.js
+++ b/packages/next/scripts/build-edge-function-template.js
@@ -19,6 +19,7 @@ async function buildTemplate(options) {
     // To see the latest stable chrome version: https://www.chromestatus.com/features/schedule
     target: 'esnext',
     write: options.write,
+    external: ['async_hooks'], // available in edge runtime
   });
 }
 
--- a/packages/next/src/edge-function-source/get-edge-function.ts
+++ b/packages/next/src/edge-function-source/get-edge-function.ts
@@ -1,5 +1,6 @@
 /// <reference lib="DOM" />
 
+import { withNextRequestContext } from '../next-request-context';
 import { toPlainHeaders } from './to-plain-headers';
 
 export interface NextjsParams {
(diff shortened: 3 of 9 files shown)


packages/next/src/vercel-request-context.ts
lubieowoce, line 1: Didn't know if it was okay to add a `@vercel/functions` dependency, so i ended up duplicating this for now.

packages/next/src/next-request-context.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export type NextRequestContextValue = {
+  waitUntil?: WaitUntil;
+};
+
+export type NextRequestContext = {
+  get(): NextRequestContextValue;
+  /** @ignore */
+  [INTERNAL_STORAGE_FIELD_SYMBOL]: AsyncLocalStorage<NextRequestContextValue>;
+};
```
lubieowoce: unclear if this is the shape we want to go with (vs just exposing a vanilla AsyncLocalStorage directly)

i'd argue that it might be useful to not expose a full ALS here, because
- we don't need (or want) the full power of ALS in next.js, we're just reading it
- this contract could technically be fulfilled using [AsyncContext](https://github.com/tc39/proposal-async-context), so coupling ourselves to an AsyncLocalStorage could prove annoying in the future
  - for single-threaded invocations (e.g. a lambda that only processes one request at a time) , it could even be a plain global that gets reset between invocations!
  
 cc @wyattjoh wdyt

packages/next/src/edge-function-source/get-edge-function.ts
```diff
@@ -108,29 +109,35 @@ export default function getNextjsEdgeFunction(
...
+              body: request.body,
+            },
+          }
+        )
     );
 
-    context.waitUntil(result.waitUntil);
+    if (result.waitUntil) {
+      context.waitUntil(result.waitUntil);
+    }
```
lubieowoce, line 140: since we now have an "official" way of passing in waitUntil, i want to move away from the "returning a promise on the result" inversion of control, because it leads to some contortions in practice when streaming is involved. so here i'm just making it so that we don't have to return a dummy promise anymore.

packages/next/scripts/build-edge-function-template.js
lubieowoce, line 1: this was always executing because, for some reason, `module.parent` seems to be undefined when required from ESM. doesn't look like this is used anywhere, so i think it's fine to remove.

packages/next/src/vercel-request-context.ts
lubieowoce, line 1: idk if it's okay to introduce a dependency on `@vercel/functions` here, so this is copy-pasted. opinons/suggestions welcome

lubieowoce, line 1: idk if it's okay to introduce a dependency on `@vercel/functions here`, so this is copy-pasted. opinons/suggestions welcome

ztanner (approved):

packages/next/src/next-request-context.ts
```diff
@@ -0,0 +1,60 @@
...
+WARNING: WATCH OUT IF YOU'RE MODIFYING THIS FILE!
+The types here must be kept in sync (or at least backwards-compatible)
+with "@next/request-context" in the Next.js codebase.
+
+This module defines `globalThis[Symbol.for("@next/request-context")]`
+which Next.js uses to access `waitUntil`.
+It is expected that the platform provides this.
+*/
+
+import { AsyncLocalStorage } from 'async_hooks';
```
ijjk, line 11: Should the be using the global `AsyncLocalStorage` instance if available similar to what we do in Next.js?

packages/next/build.mjs
```diff
@@ -18,11 +18,18 @@ await Promise.all([
...
       'src/legacy-launcher.ts',
-      'src/server-launcher.ts',
       'src/templated-launcher-shared.ts',
       'src/templated-launcher.ts',
     ],
   }),
+  // server-launcher imports some modules, so we need to bundle it
+  esbuild({
+    entryPoints: ['src/server-launcher.ts'],
+    bundle: true,
```
ijjk, line 28: Do we need to bundle this file or should we just add the new relative imports to the above list as well?

ijjk (approved):
Left couple q's but overall seems good to me

packages/next/build.mjs
lubieowoce: IMHO bundling is the easiest option to reason about, we just get one file regardless of what it uses

(also i kinda have a suspicion that there's other places implicitly relying on this being just one file, and i don't want to mess with that)

packages/next/src/next-request-context.ts
lubieowoce: hmmmm well we're doing a `require('async_hooks')` there anyway
https://github.com/vercel/next.js/blob/626b7f9de94a82e0721df9e9a29a5af850dc8ace/packages/next/src/server/node-environment-baseline.ts#L4-L9
so i think it's pretty much equivalent, don't see any obvious benefits from doing `globalThis.AsyncLocalStorage` here

EndangeredMassa (approved):

EndangeredMassa (approved):

PR #12290, opened 2024-10-16
Update workflow for automation

ismaelrumzan:
In this PR, we are applying an overlay after pulling the openapi spec from https://openapi.vercel.sh/
This overlay fixes all the errrors that the speakeasy lint identifies when it runs, which stops it from being able to generate the SDK
- Pull source from URL
- Add lint to update `path-params` rule to warning only
- Add overlay to fix lint errors

How to test
- Clone the repo: https://github.com/vercel/vercel
- Change to this branch and make a copy `git checkout -b ....`
- `pnpm i` on the root should install speakeasy - otherwise https://www.speakeasy.com/docs/speakeasy-cli/getting-started
- CD into `packages/SDK` and do `speakeasy run`
- It should run with no errors (pass the lint) and then create a bunch of changes to update the SDK (this is why we make a new branch as we don't want to run speakeasy on this branch - once this merges, we will run it with the GH action)


packages/sdk/.speakeasy/workflow.yaml
```diff
@@ -3,7 +3,10 @@ speakeasyVersion: latest
 sources:
     vercel-OAS:
         inputs:
-            - location: ../../open-api.yaml
+            - location: https://openapi.vercel.sh/
+        overlays:
+            - location: ../../overlay.yaml
+        ruleset: speakeasyBarRuleset
```
simplesagar:
```suggestion
        ruleset: barRuleset
```

packages/sdk/.speakeasy/lint.yaml
```diff
@@ -0,0 +1,10 @@
+lintVersion: 1.0.0
+defaultRuleset: speakeasyBarRuleset
+rulesets:
+ barRuleset:
```
simplesagar: you may want to choose a ruleset name specific to Vercel ?

simplesagar (commented):
Minor comment on the naming of the custom linting ruleset

overlay.yaml
```diff
@@ -0,0 +1,1057 @@
+overlay: 1.0.0
```
TooTallNate: Does this file need to be at the root of the repo? Why not inside `packages/sdk`?

ismaelrumzan: Good catch - updated

EndangeredMassa (dismissed):

molebox (dismissed):

molebox (approved):

PR #12291, opened 2024-10-16
ECO-444: Add remove subcommand for integrations command

JakeUskoski:
## Summary

This PR adds a new sub-command to integration: `remove` (also available as `rm`) which allows a user to uninstall a marketplace integration from the current team. This PR also adds a new command `vc integration-resource` with sub-commands `disconnect` and `remove` (also available as `rm`), which allows a user to unlink a project from a (marketplace integration) resource and to delete a (marketplace integration) resource.

## Background

This change is related to #12233 and #12251. As we continue to try and improve the developer experience of interacting with the marketplace via the CLI, it's important that we offer the basic general CRUD activities seen in CLI tools of `add`, `list`, and `remove` (as well as `open` for how users perform update actions on the resources day-to-day). This PR adds the `remove` functionality to finish off the set of the basic CRUD operations.

Original designs for the CRUD commands lacked detail on a `remove` command. This PR introduces the idea of removing an integration from a team, removing a resource from a team, and removing resources from projects. Since this is a lot of different destructive actions, we thought it best to break apart into multiple commands instead of overloading a single `remove` command (as this PR originally proposed). This way the commands have clear, singular uses, and it's more difficult to accidentally make unintended destructive changes.

## How to test

Build and run locally (or use `pnpm dev` instead of `vc` when in the `packages/cli` folder). You'll also need marketplace integrations installed on a team, and resources for said integrations.

```
vc integration remove <integration>
--
vc integration-resource remove <resource> [--disconnect-all]
--
vc integration-resource disconnect <resource>
vc integration-resource disconnect <resource> <project>
vc integration-resource disconnect <resource> --all
```


packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,358 @@
...
+      !skipConfirmation &&
+      !(await confirmUnlinkAllProjects(client, resource))
+    ) {
+      client.output.log('Canceled');
+      return 0;
+    }
+
+    const unlinkApiCalls =
+      resource.projectsMetadata?.map(connection => {
+        return disconnectResourceFromProject(client, resource, connection);
```
trek: I don't suppose we could make an API endpoint that does this in bulk?  I don't love loop interactions like this that can leave the system in a half committed state. If not, then let's forge ahead but I'd love to see a message telling the customer the current state and what to do next.

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,358 @@
...
+      }) ?? [];
+    try {
+      client.output.spinner('Unlinking projects from resource…', 500);
+      await Promise.all(unlinkApiCalls);
```
trek: I think we want `Promise.allSettled` if we continue with doing this as several API calls. `Promise.all` will reject on the first failure but those other request are still taking action on the system. `allSettled` plus a log of the current state would be nice.

trek (approved):

trek (changes requested):
I meant to request changes in the previous review.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: Agreed, switched over to allSettled for now.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: I prefer the idea of a single API call as well. I have a PR to add the API endpoint to perform the intended action here: https://github.com/vercel/api/pull/33142

packages/cli/src/commands/integration/remove.ts
JakeUskoski: Note: This will be switched to a normal `await` once the "unlink all" endpoint is available.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: New API endpoint works, is merged, and deployed to production. I've adjusted this route to hit that endpoint instead of sending a burst of requests.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: This has been adjusted to a single call, so it's a simple `await` now.

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+    return 1;
+  }
+
+  const { team } = await getScope(client);
+  if (!team) {
+    client.output.error('Team not found.');
+    return 1;
+  }
+
+  if (parsedArguments.args.length < 2) {
```
RPDeshaies:
```suggestion
  if (isMissingResourceOrIntegration) {
```

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+    client.output.error(
+      'You must specify a resource or integration. See `--help` for details.'
+    );
+    return 1;
+  }
+
+  if (parsedArguments.args.length > 3) {
```
RPDeshaies:
```suggestion
  if (isMissingProjects) {
```

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+    client.output.error(
+      'Cannot specify more than one project at a time. Use `--unlink-all` to unlink the specified resource from all projects.'
+    );
+    return 1;
+  }
+
+  if (
+    parsedArguments.args.length === 3 &&
+    parsedArguments.flags['--unlink-all']
```
RPDeshaies:
```suggestion
  isUsingProjectsAndUNlinkAllFlag
  
```

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+    client,
+    integrationName
+  );
+
+  if (!integrationConfiguration) {
+    client.output.error(`No integration ${chalk.bold(integrationName)} found.`);
+    return 0;
+  }
+
+  // Confirm removal desired
```
RPDeshaies: suggestion: remove in favor of const for if statement

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+      client,
+      integrationConfiguration.slug,
+      team
+    ))
+  ) {
+    client.output.log('Canceled');
+    return 0;
+  }
+
+  // Uninstall Integration
```
RPDeshaies: unecessary IMO

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+  const targetedResource = resources.find(
+    resource => resource.name === resourceName
+  );
+
+  if (!targetedResource) {
+    client.output.error(`No resource ${chalk.bold(resourceName)} found.`);
+    return 0;
+  }
+
+  if (args.length === 3) {
```
RPDeshaies: what does that mean? give it a name

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+    const unlinkProjectResults = await handleUnlinkProject(
+      client,
+      targetedResource,
+      args[2],
+      !!flags['--yes']
+    );
+    if (unlinkProjectResults !== undefined) {
+      return unlinkProjectResults;
+    }
+  } else if (flags['--unlink-all']) {
```
RPDeshaies: question: do you think of a way we could have an earlier return above and not use the elseif here?

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+  if (!project) {
+    client.output.log(
+      `Could not find project ${chalk.bold(projectName)} linked to resource ${chalk.bold(resource.name)}.`
+    );
+    return 0;
+  }
+
+  if (
+    !skipConfirmation &&
+    !(await confirmUnlinkProject(client, resource, project))
```
RPDeshaies: suggestion: give it a name

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  try {
+    client.output.spinner('Unlinking resource…', 500);
+    await disconnectResourceFromProject(client, resource, project);
+    client.output.success(
+      `Unlinked ${chalk.bold(project.name)} from ${chalk.bold(resource.name)}`
+    );
+    resource.projectsMetadata = resource.projectsMetadata?.filter(
+      project => projectName !== project.name
+    );
```
RPDeshaies: ⚠️ mutating a parameter's reference, do you really intended on doing that?

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+      return 0;
+    }
+
+    try {
+      client.output.spinner('Unlinking projects from resource…', 500);
+      await disconnectResourceFromAllProjects(client, resource);
+      client.output.success(
+        `Unlinked all projects from ${chalk.bold(resource.name)}`
+      );
+      resource.projectsMetadata = [];
```
RPDeshaies: ⚠️  same comment about mutating incoming params

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+// $ vc integration remove [resource-name] --unlink-all
+async function handleUnlinkAllProjects(
+  client: Client,
+  resource: Resource,
+  skipConfirmation: boolean
+): Promise<number | undefined> {
+  if (resource.projectsMetadata?.length === 0) {
+    client.output.log(
+      `${chalk.bold(resource.name)} has no projects to unlink.`
+    );
```
RPDeshaies: is this missing a return?

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,357 @@
...
+  } else {
+    if (
+      !skipConfirmation &&
+      !(await confirmUnlinkAllProjects(client, resource))
```
RPDeshaies: seems to be used quite often, perhaps wrap in a method and reuse in all your calls?

RPDeshaies (dismissed):

packages/cli/src/commands/integration/remove.ts
JakeUskoski: This is an intentional side-effect. The `--delete` flag checks for linked projects on the object because a resource with connected projects cannot be deleted. The act of unlinking a project from a resource, in this situation, is updating that object to reflect the new reality.

Assuming this is not desired, the alternative would be following immutability rules & cloning the resource then returning it. This breaks the other return pattern though, so failures would need to be switched to error throws.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: Technically no, because of the if/else, but it's cleaner to use a return & remove the `else`.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: Not quite, each confirm is slightly different, and each one is only called once. This is me wrapping them for better readability.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: I'll do a bit to try and make this section of the control flow a little more clear, but the gist of it is:
you can do both `--unlink-all` and `--delete` in the same call, so we specifically _can't_ exit out early. I could further extract the unlink section, since it's either unlinking 1 project or unlinking all projects, but my concern is that doing so is breaking the functions apart into unnecessarily small pieces.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: IMO the skip flag check & confirmation function naming is already clear enough on what the intention is.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: This is the pattern I leveraged from elsewhere in the CLI. For consistency, I think it's fine. I also believe the skipConfirmation variable and the confirmation function name are clear enough to read the intention of this line, that moving it to a separate line wouldn't provide much value.

packages/cli/src/commands/integration/command.ts
```diff
@@ -91,6 +92,77 @@ export const listSubcommand = {
...
+    },
+    {
+      name: 'project',
+      required: false,
+    },
+  ],
+  options: [
+    {
+      name: 'delete',
+      description: 'limits the resources listed to a designated integration',
```
adriancooney: Is this description correct? I'm not sure why we mention listing with a `vc integration rm --delete`.

packages/cli/src/commands/integration/command.ts
```diff
@@ -91,6 +92,77 @@ export const listSubcommand = {
...
+      shorthand: 'D',
+      type: Boolean,
+      deprecated: false,
+      argument: 'NAME',
+    },
+    {
+      name: 'unlink-all',
+      description: 'lists all resources regardless of project',
```
adriancooney: Same with this description - why do we mention listing here?

JakeUskoski: Copy-paste miss. Good callout.

packages/cli/src/commands/integration/command.ts
JakeUskoski: Miss on copy-paste cleanup, fixed now.

packages/cli/src/commands/integration/remove.ts
```diff
@@ -0,0 +1,360 @@
...
+    resource.projectsMetadata = [];
+  } catch (error) {
+    client.output.error(
+      `A problem occurred while unlinking all projects: ${(error as Error).message}`
+    );
+    return 1;
+  }
+}
+
+// $ vc integration remove [resource-name] --delete
```
adriancooney: Forgive me if I missed this during the planning phases of this feature but it strikes me as strange that we use the `<integration name>` to add an integration but the `resource-name` to remove. How are the ergonomics here? Is there a risk of poor UX? I wonder should this be `vc resources remove [resource-name]`.

JakeUskoski: The ergonomics on this are a little clunky, you're right. The main piece here is that removing an integration requires removing its resources, but offering users a remove command via the CLI without offering the removal of their resources makes the CLI command near-useless.

Additionally, there's a bit of a mix in what `integration` refers to already, as `vc integration add` both installs an integration and creates a resource, and `vc integration list` lists all of the resources related to integrations a user has, not their integrations.

A different removal command may make sense, but I don't know that a different namespace is the way to go. `vc resource` sounds like it should encompass more than just resources connected to integrations.

How does this look, in terms of ergonomics and naming?

```
EXISTS
======
vc integration add
vc integration list
vc integration open

NEW
======
vc integration remove <integration>
--
vc integration delete <resource> [--disconnect-all]
--
vc integration disconnect <resource>
vc integration disconnect <resource> --project <project>
vc integration disconnect <resource> --all
--
vc integration connect <resource>
vc integration connect <resource> --project <project>
(connect is not currently covered by remove, but is covered partially by add)
```

adriancooney: Yep, I seen that complexity alright. I'm not sure I'm a fan of `remove` and `delete` commands, I think that might be a little confusing for users to have both since they essentially mean the same thing. The more I think about it, does the CLI need to concern itself with the uninstallation of integrations? Can we ask the user to do that on the dashboard? If we leave that out, the ergonomics seem pretty good to me:

```
EXISTS
======
vc integration add
vc integration list
vc integration open

NEW
======
vc integration delete <resource> (--unlink-all)
vc integration unlink <resource> (<project>)
vc integration unlink <resource> --unlink-all
```

Two modifications: 
* Drop the `vc integration remove <integration>`
* Make the `<project>` optional in `vc integration unlink <resource> <project>`. If none passed, use the current project.

JakeUskoski: Yep, on the "make project optional" part and use the current project, I agree whole heartedly, and have some further details outlined in the notion doc which includes flag choices.
https://www.notion.so/vercel/Tech-Plan-Improvements-to-Marketplace-in-the-Vercel-CLI-b69a362b777d4e0298a8c05702d36b6e?pvs=4

I agree having both delete & remove is awkward. I'd rather name "remove" "uninstall" for uninstalling the integrations, since that's the language we use otherwise. It's just that "remove" is the opposite to "add." If anything, I think we should rename `vc integration add` to `vc integration install`.

adriancooney: I think renaming `remove` to `uninstall` works for me! That would resolve any inconsistencies. Apologies for coming in so late with design questions.

packages/cli/test/unit/commands/integration/remove.test.ts
adriancooney, line 1: Great work on the tests.

adriancooney (dismissed):
LGTM - only outstanding issue is renaming `remove` to `uninstall` which is used for uninstalling the integration and have a separate `delete` command for deleting resources.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: With the remove command broken apart, the usage has been clarified and modification is no longer relied upon, so this mutation has been removed.

packages/cli/src/commands/integration/remove.ts
JakeUskoski: With the remove command broken apart, the usage has been clarified and modification is no longer relied upon, so this mutation has been removed.

packages/cli/test/unit/commands/integration/remove.test.ts
JakeUskoski: Thanks!

packages/cli/src/commands/integration/remove.ts
JakeUskoski: I did a bit of a switcharoo here. I've renamed `delete` to `remove` and the old `remove` to `uninstall` to make it analogous with the `add` command and `install` command. That being said, it's very simple to rename the command from `remove` back to `delete`.

JakeUskoski: Sat on it for a bit thinking about the use of `remove` vs `delete`, and ended up coming to the same final conclusion as you around using `delete` for resources. `remove` to me implies it's still around somewhere, whereas in the Vercel dashboard UI we refer to resources as being "deleted."

EndangeredMassa: The complexity here is important to manage and the constraints are wider than only the `vc integrations` namespace. Can we get a small RFC for what these commands will look like that the Zero Config team can review? I'm sure we can move quickly on discussing that RFC to not block this change for very long.

RPDeshaies (dismissed):

adriancooney (dismissed):

trek (dismissed):

packages/cli/src/commands/integration/command.ts
```diff
@@ -91,6 +92,126 @@ export const listSubcommand = {
...
+      value: [
+        `${packageName} integrations uninstall <inegration>`,
+        `${packageName} integrations uninstall acme`,
+      ],
+    },
+  ],
+} as const;
+
+export const deleteSubcommand = {
+  name: 'delete',
```
EndangeredMassa: In the Vercel CLI, we pretty consistently use "remove" to get rid of something, as the opposite of "add". I think we should highlight [this discussion](https://github.com/vercel/vercel/pull/12291#discussion_r1810930285) across teams to make sure we make the best decision here not just for the name of this subcommand, but the related options as well. This could be a brief meeting or RFC.

EndangeredMassa (changes requested):

packages/cli/src/commands/integration/command.ts
JakeUskoski: Quick question: for situations like `vc integration` where the command is really just a point for multiple sub-commands, how do you refer to this? Do you still call it a command, or is it referred to as a module or some other term? For the sake of this comment, I'll refer to it as a module.

Overall I agree. Unfortunately this is a bit of an ambiguity problem, because "add" both adds an integration and a resource, and this is the `integration` module, so it would make the most sense that `vercel integration remove` removes an integration, not a resource. At the same time, in the Vercel dashboard we refer to the removal of a resource as "delete" and the removal of an integration as "uninstall." This problem  (as far as the CLI is concerned) mostly stems from us using the `integration` module to interact with both integrations and resources, two highly intertwined but technically separate entities.

Right now we're using "delete" and "uninstall" partially to avoid "remove" to keep it open for future changes. I have an internal proposal with the team over how we should better manage interacting with these two entities. One proposal is to have two different modules, `integration` and `resource`, and the other is to have a `resource` sub-module within the `integration` module. Then the full command list would look something along the lines of either...

```
vc integration add
vc integration list
vc integration open
vc integration remove

vc resource add (Alias of vc integration add)
vc resource list
vc resource open
vc resource remove
vc resource disconnect
```
or...
```
vc integration add
vc integration list
vc integration open
vc integration remove

vc integration resource list
vc integration resource open
vc integration resource remove
vc integration resource disconnect
```

Does the Zero Config team have any strong opinions around modules and module structure? I'm hesitant to give a module a sub-module, because while resources only exist tied to integrations, I don't like the idea of typing 4 commands before getting to a point of actually typing a parameter or flag. For that reason, I'm more partial to the idea of having two top-level modules, `integration` and `resource`.

JakeUskoski: Overall, my above notes on future options/plans for the integration module of the CLI are out-of-scope for right now. We could rename `delete` to `remove` for now, but it would make `vc integration remove` for removing integrations a breaking change in the future. Instead I could rename it to `vc integration resource remove` and `vc integration uninstall` to `vc integration remove`, which creates the beginnings of the sub-module, but can be more easily pivoted out of without creating a breaking change.

JakeUskoski: As per discussions, the `vc resource` has been set as `vc integration-resource` for now.

packages/cli/src/commands/index.ts
```diff
@@ -19,6 +19,8 @@ export default new Map([
   ['inspect', 'inspect'],
   ['install', 'install'],
   ['integration', 'integration'],
+  ['integration-resource', 'integration-resource'],
+  ['ir', 'integration-resource'],
```
EndangeredMassa, line 23: **note (non-blocking):** Calling out that we'd be adding these to the command list:

- `integration-resource`
- `ir`

packages/cli/src/commands/integration/resource/command.ts
EndangeredMassa, line 1: **suggestion (blocking):** move `integration/resource` to `integration-resource` as a sibling to `integration`

Can we keep the command consistency so that it's obvious where command logic is when we go looking for it?‏

packages/cli/src/util/telemetry/commands/integration/resource/disconnect.ts
EndangeredMassa, line 1: **suggestion (blocking):** rename this directory to `integration-resource`‏

EndangeredMassa (approved):

EndangeredMassa (changes requested):
meant to request changes

packages/cli/src/util/telemetry/commands/integration/resource/disconnect.ts
JakeUskoski: Done 👍

packages/cli/src/commands/integration/resource/command.ts
JakeUskoski: Done 👍

EndangeredMassa (dismissed):
Thanks for all the work on this!

EndangeredMassa (approved):

PR #12305, opened 2024-10-17
[cli] refactor Output usage

EndangeredMassa:
On the path towards making the CLI entrypoint more testable, this PR refactors usage of `Output` to be less interconnected to the system. It is now wrapped by a singleton that can be modified (if necessary) as the program flows.

This allows other contexts to use the same singleton after the initial CLI logic parses the args and updates the values on the singleton. The removal of this direct coupling allows future refactors of the Vercel CLI entrypoint that we'd need to do in order to write unit tests for that logic.

The `output` instance is no longer coupled to the `client` instance.

---

### Primary Diff

There are a lot of changes in this PR, but most of them are adjusting the call sites of the `output` methods. The core changes are:

- [packages/cli/src/util/output/create-output.ts (diff)](https://github.com/vercel/vercel/pull/12305/files#diff-edcc89c00cb919143e9d7bada97aa314c17f57d96bc92f56235d60bce99b81f7)
- [packages/cli/src/util/client.ts (diff)](https://github.com/vercel/vercel/pull/12305/files#diff-80e8d5a02c540100cc38f9c1240d73742010ee4b2f195e615943ac89d67d5bc7)
- [packages/cli/test/mocks/client.ts (diff)](https://github.com/vercel/vercel/pull/12305/files#diff-0838ea92c3dfb1e78ce8b2950b7f01ad0be324bd615ab1287b5ef5bc14a141b7)
- [packages/cli/test/unit/util/output/create-output.test.ts (diff)](https://github.com/vercel/vercel/pull/12305/files#diff-d85d1e622a38f762b41c24121bce73a94f543e63f32221605ce7f9c05e4ba255)

Test setup changes:

- [packages/cli/package.json](https://github.com/vercel/vercel/pull/12305/files#diff-6e2e2a1851648938b325ba84de634407a4e69a644ea61102df15ca4a8a7a9758)
- [packages/cli/vitest.config.mts](https://github.com/vercel/vercel/pull/12305/files#diff-026e406c8a99e76baf16caff6fa1edc947edbe165e2984065b3a1955f6b7fb01)
- [packages/cli/vitest.setup.mts](https://github.com/vercel/vercel/pull/12305/files#diff-0b9c63128107843d42d33807395f942726bbc4a0b902918f7cd8db39f3e04dac)

Loosened some regex to accmmomodate line breaks:
- [packages/cli/test/integration-1.test.ts](https://github.com/vercel/vercel/pull/12305/files#diff-c6ef3cc2118994782e3043fc2f1542819e9ba36e60294873b34eebc9185f6df3)
- [packages/cli/test/integration-2.test.ts](https://github.com/vercel/vercel/pull/12305/files#diff-0742aefeddf7259c9ac88e2f2b21141259388a1b840310d636fa6293dabb55db)

Updated some snapshots that had color in them to no longer have color:

- [packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap](https://github.com/vercel/vercel/pull/12305/files#diff-bc9b97cc02f68d46525627d35c369f91b9cebd00fdd54bed5c519c58f5bb437a)
- [packages/cli/test/unit/__snapshots__/args.test.ts.snap](https://github.com/vercel/vercel/pull/12305/files#diff-ab59632db71eb7d8196db19c3983e26610366f06ab87348fcdc3f054d57e8cef)

---

### Consumer-facing Change

Instead of:

```
client.output.debug('some message');
```

Use:

```
import output from './output-manager';
// ...
output.debug('some message');
```

---

Card: https://linear.app/vercel/issue/ZERO-2738/refactor-tangled-output-from-client-in-vercel-cli


packages/cli/src/output-manager.ts
```diff
@@ -0,0 +1,11 @@
...
+
+let output = new Output(process.stderr, {
+  debug: false,
+  noColor: true,
+});
+
+/**
+ * A managed singleton instance of the Output class.
+ */
+export default output;
```
EndangeredMassa, line 10: I'm open to calling this something other than `output-manager` that makes it clear this is the singleton wrapper.

packages/cli/src/index.ts
```diff
@@ -793,7 +789,7 @@ const handleUnexpected = async (err: Error) => {
 
   // We do not want to render errors about Sentry not being reachable
   if (message.includes('sentry') && message.includes('ENOTFOUND')) {
-    debug(`Sentry is not reachable: ${err}`);
+    output.error(`Sentry is not reachable: ${err}`);
```
jeffsee55: >  // We do not want to render errors about Sentry not being reachable

Should we use `output.debug` here?

erikareads (dismissed):
Are there any other places where we're creating a new `Output` class?

packages/cli/src/output-manager.ts
jeffsee55: > This allows other contexts to use the same singleton after the initial CLI logic parses the args and updates the values on the singleton.

**non-blocking:** Do you have an API in mind for that `mockClient` will do with output? Will it no longer initialize it within it's contstructor?

jeffsee55 (changes requested):

packages/cli/src/index.ts
EndangeredMassa: I made a judgement call. I think we should show that even when debug is off. Happy to switch it if others disagree.

jeffsee55: I don't even know what Sentry is used for so not gonna argue with that 👍

jeffsee55 (approved):

jeffsee55 (dismissed):

packages/cli/src/index.ts
```diff
@@ -793,7 +789,7 @@ const handleUnexpected = async (err: Error) => {
 
   // We do not want to render errors about Sentry not being reachable
```
erikareads, line 807: @EndangeredMassa We do have this comment here, so I think not rendering errors except in debug was intentional.

onsclom (dismissed):

EndangeredMassa:
> Are there any other places where we're creating a new `Output` class?

> Do you have an API in mind for that mockClient will do with output? Will it no longer initialize it within it's contstructor?

The other invocation are all in tests. I dismissed the need to touch them for this PR because the tests all pass, but I'll take a closer look.

packages/cli/src/output-manager.ts
EndangeredMassa: I updated MockClient to use this

trek (dismissed):

packages/cli/src/util/output/create-output.ts
```diff
@@ -25,31 +26,62 @@ export interface LinkOptions {
...
+    }
+
+    if (debugEnabled !== undefined) {
+      this.debugEnabled = debugEnabled;
+    }
+
+    if (supportsHyperlink !== undefined) {
+      this.supportsHyperlink = supportsHyperlink;
+    }
+
```
onsclom: Non-blocking: This case is already handled below.

```suggestion
```

onsclom (dismissed):

packages/cli/src/util/output/create-output.ts
EndangeredMassa: Ah, copy paste error. Thanks!

https://github.com/vercel/vercel/pull/12305/commits/bdfe35f0a3f3c4cb34210a70944a4e49ce30abee

trek (dismissed):

EndangeredMassa:
I was accidentally forcing color to be off. Fixed: https://github.com/vercel/vercel/pull/12305/commits/58797bf6923829386797e7f0cd71960525216923

EndangeredMassa:
Made some test setup changes so that `packages/cli` has its own vitest setup. In that setup, we initialize the `output` singleton. See:

- [packages/cli/package.json](https://github.com/vercel/vercel/pull/12305/files#diff-6e2e2a1851648938b325ba84de634407a4e69a644ea61102df15ca4a8a7a9758)
- [packages/cli/vitest.config.mts](https://github.com/vercel/vercel/pull/12305/files#diff-026e406c8a99e76baf16caff6fa1edc947edbe165e2984065b3a1955f6b7fb01)
- [packages/cli/vitest.setup.mts](https://github.com/vercel/vercel/pull/12305/files#diff-0b9c63128107843d42d33807395f942726bbc4a0b902918f7cd8db39f3e04dac)

trek (dismissed):

EndangeredMassa:
Still fixing some tests. I'll come back to this.

EndangeredMassa:
I added these notes the primary diff section of the PR description.

Loosened some regex to accmmomodate line breaks:
- [packages/cli/test/integration-1.test.ts](https://github.com/vercel/vercel/pull/12305/files#diff-c6ef3cc2118994782e3043fc2f1542819e9ba36e60294873b34eebc9185f6df3)
- [packages/cli/test/integration-2.test.ts](https://github.com/vercel/vercel/pull/12305/files#diff-0742aefeddf7259c9ac88e2f2b21141259388a1b840310d636fa6293dabb55db)

Updated some snapshots that had color in them to no longer have color:

- [packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap](https://github.com/vercel/vercel/pull/12305/files#diff-bc9b97cc02f68d46525627d35c369f91b9cebd00fdd54bed5c519c58f5bb437a)
- [packages/cli/test/unit/__snapshots__/args.test.ts.snap](https://github.com/vercel/vercel/pull/12305/files#diff-ab59632db71eb7d8196db19c3983e26610366f06ab87348fcdc3f054d57e8cef)

erikareads (approved):

PR #12306, opened 2024-10-17
[cli] Silence punycode deprecation warning

onsclom:
At runtime on Node 22, the CLI currently displays:
```
(node:14481) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

This is the least invasive way to silence this message.

Thanks again for finding this solution @EndangeredMassa!


EndangeredMassa (dismissed):

trek (dismissed):

trek:
Oh yes, I like this much more than process forking. That made me 😟

packages/cli/src/index.ts
```diff
@@ -13,6 +13,25 @@ try {
...
+  // eslint-disable-next-line no-console
+  console.error = (msg: unknown) => {
+    const isSilencedError = SILENCED_ERRORS.some(
+      error => typeof msg === 'string' && msg.includes(error)
+    );
+    if (isSilencedError) {
+      return;
+    }
+    originalError(msg);
+  };
```
onsclom, line 32: Made a small change to fix a problem our tests caught. `msg` is sometimes not a string, so we need to check it is before using `includes()`.

EndangeredMassa (approved):

PR #12325, opened 2024-10-18
[next] Output different Prerenders for Fallback HTML/RSC

wyattjoh:
Previously when the `allowQuery` was set to `[]` it also incorrectly applied it to the Prefetch RSC, causing a cache error when a prefetch revalidated a page.


ijjk (approved):

wyattjoh:
Holding until upstream changes are released

PR #12336, opened 2024-10-22
ECO-585: Handle non-boolean readonly & mark expressions as unsupported

JakeUskoski:
## Summary

This PR fixes a bug in the `vc install <integration>` flow. Now `ui:disabled` is recognized (and skipped), and `ui:hidden` won't break when set to `create` or `update`. Additionally, complex expressions are explicitly unsupported.

## Details

Integrations sometimes have complex UI cases (situational hidden, read-only, disabled) that are allowed in Vercel but weren't properly detected and handled in the CLI integration installation wizard. The `ui:disabled` property was being ignored outright in the wizard, while the `ui:hidden` property was being too aggressive, considering any value to mean it should be skipped. The `ui:hidden` property can be set to `update` to mean the field is _not_ readonly during creation, which the wizard manages. For Upstash Vector, this meant all the necessary fields were being skipped by the wizard, since they're readonly in update flows, but not the creation flow.

Additionally, Upstash Vector uses an expression for one of its fields to determine if it should be visible & populated or not. Our CLI tool does not have the same full parser for complex expressions that our UI has, so for cases where we see expressions we should direct the user to the web UI to complete the process.

## How to test

Build and run locally (or use `pnpm dev` instead of `vc` when in the `packages/cli` folder).

```
vc install upstash vector
-> select Upstash Vector as the product
```

### DIFF
--- a/packages/cli/src/commands/integration/wizard.ts
+++ b/packages/cli/src/commands/integration/wizard.ts
@@ -246,12 +246,12 @@ function isDisabled(schema: MetadataSchemaProperty) {
   );
 }
 
-// biome-ignore lint/suspicious/noExplicitAny: <explanation>
-function instanceOfExpression(obj: any): obj is Expression {
-  if (obj !== Object(obj)) {
+function instanceOfExpression(obj: unknown): obj is Expression {
+  const checkedAsObject = Object(obj);
+  if (obj !== checkedAsObject) {
     return false;
   }
-  return 'expr' in obj;
+  return 'expr' in checkedAsObject;
 }
 
 class ExpressionError extends Error {}
--- a/packages/cli/src/commands/integration/wizard.ts
+++ b/packages/cli/src/commands/integration/wizard.ts
@@ -1,6 +1,7 @@
 import type Client from '../../util/client';
 import list, { type ListChoice } from '../../util/input/list';
 import type {
+  Expression,
   Metadata,
   MetadataEntry,
   MetadataSchema,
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/integration/wizard.ts
```diff
@@ -207,11 +220,38 @@ async function getMetadataFromSteps(
...
+  if (instanceOfExpression(schema['ui:disabled'])) {
+    throw new ExpressionError('Expression found in schema');
+  }
+  return Boolean(
+    schema['ui:disabled'] === true || schema['ui:disabled'] === 'create'
+  );
 }
+
+// biome-ignore lint/suspicious/noExplicitAny: <explanation>
+function instanceOfExpression(obj: any): obj is Expression {
```
mbrakken: Nit: `unknown` instead of `any`?

packages/cli/test/unit/commands/integration/add.test.ts
```diff
@@ -238,6 +238,31 @@ describe('integration', () => {
...
+          await expect(client.stderr).toOutput(
+            'Do you want to link this resource to the current project? (Y/n)'
+          );
+          client.stdin.write('n\n');
+          await expect(client.stderr).toOutput(
+            'This resource must be provisioned through the Web UI. Open Vercel Dashboard?'
+          );
+          client.stdin.write('n\n');
+          await expect(exitCodePromise).resolves.toEqual(0);
+          expect(openMock).not.toHaveBeenCalled();
```
mbrakken, line 291: Should there be a test case where the browser is opened?

mbrakken (dismissed):
A couple suggestions.

I don't think my review should be sole/definitive (I haven't touched any of the CLI), but it looks good to me.

packages/cli/test/unit/commands/integration/add.test.ts
JakeUskoski: Technically speaking, yes, or this one should be the one where it's opened, and 'no' should be assumed to be covered in a different test case. I wrote this under the assumption the general "not supported" flow was already covered. Turns out, it isn't. I've added both cases in as tests.

trek (dismissed):

packages/cli/src/commands/integration/wizard.ts
JakeUskoski: I always forget `unknown` is an option. Added.

trek (dismissed):

jeffsee55 (approved):

trek (approved):

PR #12340, opened 2024-10-22
[cli] add telemetry tracking to `project add`

onsclom:
Telemetry for `vc project add`:

1 argument (name)
0 flags
0 options


packages/cli/test/unit/commands/project/add.test.ts
```diff
@@ -25,5 +25,31 @@ describe('add', () => {
...
+        name: 'test-project',
+      });
+
+      client.setArgv('project', 'add', 'test-project');
+      await projects(client);
+
+      const project: Project = await client.fetch(`/v8/projects/test-project`);
+      expect(project).toBeDefined();
+
+      expect(client.telemetryEventStore).toHaveTelemetryEvents([
```
jeffsee55, line 43: Instead of creating a new test that doesn't differ from the one above you can just put this assertion into the first test

jeffsee55 (changes requested):

packages/cli/test/unit/commands/project/add.test.ts
onsclom: Good point, was copying this pattern from https://github.com/vercel/vercel/pull/12194/files#diff-e6a0a76a9158df1b8578fa4bbd0a67a25019133076cdae07cc0dfeb5e99c8695R40. Since this pattern already exists and it provides better test failing messages, maybe we keep it like that?

trek: Lemme short circuit the conversation since I swear it always goes exactly like this:

"Or shift the duplicated code to a `beforeEach` and only focus on the expectations of `client.telemetryEventStore`."
"Then we're just running more tests and that takes more time."
"But we get focused failures! Can we just use `beforeAll`"
"`beforeAll` just ends up causing really confusing state issues that are hard to debug"

EndangeredMassa: This is a unit test that appears to hit mocked endpoints. I imagine it's super fast. If true, I don't think we need to worry about adding an extra test here.

trek (approved):

packages/cli/test/unit/commands/project/add.test.ts
trek: I gave my approval because both options are correct but also wrong.

packages/cli/test/unit/commands/project/add.test.ts
```diff
@@ -25,5 +25,31 @@ describe('add', () => {
...
+      useProject({
+        ...defaultProject,
+        id: 'test-project',
+        name: 'test-project',
+      });
+
+      client.setArgv('project', 'add', 'test-project');
+      await projects(client);
+
+      const project: Project = await client.fetch(`/v8/projects/test-project`);
```
trek, line 40: I know this was copied from above but... what is it doing?

packages/cli/test/unit/commands/project/add.test.ts
trek: @EndangeredMassa it still takes time to set create all those objects and execute the code under test! Eventually it adds up.

trek: (stage whisper) Sean, it's your line.

jeffsee55 (approved):

packages/cli/test/unit/commands/project/add.test.ts
jeffsee55: In general, the patterns I've come to like in tests:

- Use `beforeAll` is little as possible for reasons Trek already stated.
- [Arrange, Act, Assert](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) is a nice mental model, and `beforeEach` should only be used for the `Arrange`
- Multiple assertions are good. They are a way of saying, when this thing happens, here are all of the side effects or outcomes. You can read them in one place. I hadn't given much thought to the fact that you get a better error message when you split them up, good point @onsclom

trek: @jeffsee55 it also depends on the framework. Some have `expect` syntax like `expect(actual).toSomethingSomething(expected, failureExplanationIfNot)` or similar. [`@alex_neo/jest-expect-message`](https://www.npmjs.com/package/@alex_neo/jest-expect-message) has this for jest so should work with vite as well I think?

If we had and used that consistently I'd be happy with many assertions. Plus it lets you add better failure reasons generally.

Good:
```
it('project user count and team user account identical', () => {
  project.addUser(user);
  expect(project.userCount).toEqual(team.userCount);
});
// "Failed 'project user count and team user account identical'. Expected 3, got 2" 
// is super actionable.
```

Bad:
```
it('adding a user', () => {
  project.addUser(user);
  expect(project.userCount).toEqual(team.userCount);
});
// "Failed 'adding a user'. Expected 3, got 2" 
// requires my faulty skull bacon to do work
```

But, also good:
```
it('adding a user', () => {
  project.addUser(user);
  expect(project.userCount, 'project user count and team user account must be identical').toEqual(team.userCount);
});
// "Failed 'adding a user': Project user count and team user account must be identical. Expected 3, got 2"
// back to actionable.
```

packages/cli/test/unit/commands/project/add.test.ts
onsclom: Good catch! Looks like it's just testing the `useProject` mock. I'll make a followup PR to remove this

PR #12346, opened 2024-10-22
Add telemetry for `teams` subcommands

jeffsee55:
Adds telemetry collection for `vc teams` and their subcommands.


packages/cli/src/util/telemetry/index.ts
```diff
@@ -29,6 +29,7 @@ export class TelemetryClient {
   store: TelemetryEventStore;
 
   protected redactedValue = '[REDACTED]';
+  protected noValueToTriggerPrompt = '[TRIGGER_PROMPT]';
```
jeffsee55, line 32: Thinking this is a value that says we prompted the user for this value since they didn't provide it. Open to suggestions here.

trek: I think this, or just having it missing and we check by `sessionId` where is missing are both fine.

trek (approved):

PR #12349, opened 2024-10-23
[cli] add telemetry tracking for vercel dev invocations

trek:
This one was a _beast_ to get right because of how much filesystem/environment mocking needed to occur, how little already existed in the repo, and how little I knew about doing these in vitest. Some notes!

* added `memfs` which is the library vitest [recommends](https://vitest.dev/guide/mocking.html#file-system) for mocking the file system. Their example involves mocking for every test which I'm not sure yet we'd want do. Figuring out the right incantation to fully mock the filesystem in just one test file was an adventure.
* Updated `vitest` to latest so we can [stub environment variables to `undefined`](https://github.com/vitest-dev/vitest/pull/6359). Our version of vitest has `unstubAllEnvs` but that seemed like it would just paint us into a weird corner. We use `process.env.__VERCEL_DEV_RUNNING` as a lock to prevent someone from running `vercel dev` twice in one process but in tests this is exactly what occurs. Outside of tests we use rely on the process exiting to do the cleanup for us.
* I mocked the entire `DevServer` class _very_ basically. We don't want to actually run it but when someone comes back to add more unit tests they should add in more spys to ensure the server is being invoked with the arguments we expect and methods are called in the correct order.

We do have our own filesystem mocking [in the repo](https://github.com/vercel/vercel/blob/d392591ebdb28b809a9411a6930d6ede397df202/packages/fs-detectors/test/virtual-file-system.ts), but not as part of `cli`. Extracting that seemed a bit too heavy and IMO we should use a third party library for this anyway.


packages/cli/test/unit/commands/dev/index.test.ts
```diff
@@ -1,9 +1,147 @@
...
+  });
+
+  describe('--listen', () => {
+    it('tracks the listen option if supplied', async () => {
+      vi.spyOn(process, 'cwd').mockReturnValue(projectPath);
+
+      client.setArgv('dev', '--listen=9090');
+      const exitCodePromise = dev(client);
+
+      // dev is an odd duck in that normally only exists on SIGTERM
```
jeffsee55:
```suggestion
      // dev is an odd duck in that normally only exits on SIGTERM
```
This same comment is found below as well.

trek: @jeffsee55 can I fix this in a follwup? if tests go green? I've been fighting flakey CI all day and this will be my 11th run

packages/cli/package.json
```diff
@@ -170,7 +171,7 @@
     "ts-node": "10.9.1",
     "utility-types": "2.1.0",
     "vite": "5.1.6",
-    "vitest": "2.0.1",
```
EndangeredMassa, line 173: **suggestion (non-blocking):** ‏As a follow-up task, can you update the vite version throughout the repo?

packages/cli/src/util/telemetry/commands/inspect/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (idOrHost) {
+      this.trackCliArgument({
+        arg: 'deploymentIdOrHost',
+        value: this.redactedValue,
+      });
+    }
+  }
+
+  trackCliOptionTimeout(timeout?: string) {
+    if (timeout) {
```
EndangeredMassa, line 14: **suggestion (soft-blocking):** missing implementation?

Did you intend to include the `inspect` telemetry in this PR?
‏

EndangeredMassa (dismissed):

packages/cli/src/util/telemetry/commands/inspect/index.ts
trek: Nope! Juggling too many branches while rerunning CI 22 times

packages/next/package.json
```diff
@@ -11,7 +11,8 @@
     "test-next-local": "pnpm test test/integration/*.test.js test/integration/*.test.ts",
     "test-next-local-legacy": "pnpm test test/integration/legacy/*.test.js",
     "test-next-local:middleware": "pnpm test test/integration/middleware.test.ts",
-    "test-e2e": "rm -f test/builder-info.json; pnpm test test/fixtures/**/*.test.js",
```
trek, line 14: Flakey tests related to next@canary

packages/next/test/fixtures/00-server-build/index.test.js
```diff
@@ -37,18 +37,18 @@ describe(`${__dirname.split(path.sep).pop()}`, () => {
...
-        { urlPath: '/fallback-blocking/on-demand-2' },
-        {
-          urlPath: '/fallback-blocking/on-demand-2',
-          query: '?slug=on-demand-2',
-        },
-        { urlPath: '/fallback-blocking/on-demand-2', query: '?slug=random' },
-        { urlPath: '/fallback-blocking/on-demand-2', query: '?another=value' },
-      ],
-    },
+    // {
```
trek, line 40: Flakey tests related to next@canary

packages/cli/src/util/telemetry/commands/inspect/index.ts
EndangeredMassa: I think you still want to remove this file, but that can be a follow-up.

EndangeredMassa (approved):

packages/cli/src/util/telemetry/commands/inspect/index.ts
trek: I'm in such a mixed up state. This is not on my local branch _at all_ and git claims I'm up to date.

trek: CleanShot 2024-10-24 at 10 38 40@2x

PR #12352, opened 2024-10-23
[cli] add telemetry for `vercel target ls`

trek:
A pretty easy one! `vercel target` only has one subcommand (tracked now) and `vercel target ls` has nothing unique to track.


packages/cli/test/unit/commands/target/list.test.ts
```diff
@@ -10,7 +10,45 @@ import createLineIterator from 'line-async-iterator';
 import { parseSpacedTableRow } from '../../../helpers/parse-table';
 
 describe('target ls', () => {
-  describe.todo('--next');
```
trek, line 13: This was incorrect. `target ls` has no `--next`. The collection is not paginated at the server.

onsclom (approved):

PR #12354, opened 2024-10-23
Avoid retries by default when using the MockClient

jeffsee55:
When some of our test mocks intentionally respond with a 500, that response takes several seconds because the [client's `fetch` function](https://github.com/vercel/vercel/blob/7e8fbd20e01135513ad28936155b078650732d37/packages/cli/src/util/client.ts#L173-L174) automatically retries the request. 

But since we're mocking these requests we should just automatically fail.

### DIFF
--- a/packages/cli/test/unit/commands/rollback/index.test.ts
+++ b/packages/cli/test/unit/commands/rollback/index.test.ts
@@ -182,7 +182,6 @@ describe('rollback', () => {
     await expect(client.stderr).toOutput(
       `Fetching deployment "${previousDeployment.id}" in ${previousDeployment.creator?.username}`
     );
-    // we need to wait a super long time because fetch will return on 500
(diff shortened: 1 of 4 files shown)


trek (dismissed):

packages/cli/test/unit/commands/rollback/index.test.ts
```diff
@@ -183,7 +183,7 @@ describe('rollback', () => {
       `Fetching deployment "${previousDeployment.id}" in ${previousDeployment.creator?.username}`
     );
     // we need to wait a super long time because fetch will return on 500
```
trek: I think we can remove this comment now too. It wasn't even terribly accurate.

jeffsee55:
```suggestion
```

EndangeredMassa (approved):

PR #12239, opened 2024-10-07
fix: bump path-to-regexp

trek:
https://github.com/vercel/vercel/pull/12131 but with my account for secrets access

PR #12356, opened 2024-10-23
fix: bump path-to-regexp -- fix test

jeffsee55:
Are we certain that this failing test is using a realistic regex? I think the `+` here after `(.*)` is redundant, though not really invalid to the point that it should error..

---

Related: https://github.com/vercel/vercel/pull/12239


EndangeredMassa:

PR #12358, opened 2024-10-23
[cli] add `VERCEL_ENV` and `VERCEL` to process to simulate runtime

### LINKED ISSUE #2878: now dev return spawn E2BIG
# Bug report

Suddenly after a couple weeks of development, my now dev command line doesn't work anymore. It started when I closed the terminal where it was running and opened a new one.

Now, it does the same even with a basic nextjs project
image

And even with the now init nextjs command line
image

## To Reproduce

now init nextjs
cd nextjs
now dev

## Expected behavior

The now dev server is properly launched and I can debug my code.

## System information
- OS : macOS 10.14.6 - Memory: 16 GB
- Node : v10.16.0 (but tested with 8.x.x also)
- yarn 1.16.0 (but tested with 1.14 and 1.15 also)
- Version of Next.js: 9.0.5

## Additional context

It happened to my collegues too. After a while, the command line just doesn't work anymore.
We have tried : 
 - Reinstall node, update node, downgrade node
 - Reinstall yarn, update yarn, downgrade yarn
 - Uninstall / Reinstall now
 - rm -rf /Users/....../Library/Caches/co.zeit.now/ 

Nothing works. The command line throw the error directly after the yarn command, just like if it was a problem with the size of a file (maybe logs ?) or the number of arguments in a function?

Anyone can help ? 
Thank you very much

### LINKED ISSUE #11633: [cli] Removing VERCEL_ENV in vercel@34.2.0 is a breaking change
Our app depends on the `VERCEL_ENV="development"` to be set in local dev mode. 

Publishing the [change](https://github.com/vercel/vercel/pull/11526) as a minor change breaks a lot of things. I think it should had been released as a major (breaking) change.

It is really convenient to have access to VERCEL_ENV in dev mode. I vote for adding it back. 

Workaround: For now we have locked the cli to vercel@34.1.14 to fix it, eg `vercel@34.1.14 env pull`.

I am a big fan of Vercel, keep up the good work!

Ref

- https://github.com/vercel/vercel/releases/tag/vercel%4034.2.0
- https://github.com/vercel/vercel/pull/11526

trek:
In https://github.com/vercel/vercel/pull/11526 we stopped pulling down environment variables that are set by the runtime into `.env` files since this really is an appropriate location for them and blank values can cause issues for some customers.

However, we were relying on these so that `vercel dev` had a similar set of variables as the platform at runtime. Losing these was a breaking change.

This PR sets them within the development sever itself to better simulate our actual runtime environment. I wasn't quite sure of the use of `projectSettings?.autoExposeSystemEnvs` [here](https://github.com/vercel/vercel/blob/804b94fa6177198728b1d6cd497aad787b452130/packages/cli/src/util/dev/server.ts#L709-L711) and potentially the correct location for these is within that `if` block?

Closes #11633


packages/cli/src/util/dev/server.ts
```diff
@@ -703,6 +703,10 @@ export default class DevServer {
     buildEnv['NOW_REGION'] = 'dev1';
     allEnv['NOW_REGION'] = 'dev1';
 
+    // simulate parts of the platform for local environment
+    runEnv['VERCEL_ENV'] = 'development';
+    runEnv['VERCEL'] = '1';
```
TooTallNate: These should apply at "build time" as well:

```suggestion
    allEnv['VERCEL_ENV'] = 'development';
    allEnv['VERCEL'] = '1';
```

TooTallNate (changes requested):

TooTallNate (approved):

PR #12360, opened 2024-10-23
[cli] Add telemetry for `vercel link`

TooTallNate:
This one required adding unit tests for the `--repo` flag, which was a bit tricky.


packages/cli/src/commands/deploy/index.ts
```diff
@@ -118,8 +118,7 @@ export default async (client: Client): Promise<number> => {
     );
 
     if ('--confirm' in parsedArguments.flags) {
-      // Deprecated flag so it's not in the type definition
```
TooTallNate, line 121: This wasn't true - `--confirm` is typed.

trek (dismissed):

EndangeredMassa (approved):
Thanks for fleshing out those tests!

PR #12368, opened 2024-10-24
[cli] Add telemetry for `vc env pull`

onsclom:
1 argument: `filename`
2 options: `environment`, `git-branch`
1 flag: `yes`

`vc build` sometimes calls `vc env pull`, so `vc build` may trigger `vc env pull` telemetry events.


packages/cli/src/util/telemetry/commands/env/pull.ts
```diff
@@ -0,0 +1,36 @@
...
+        value: this.redactedValue,
+      });
+    }
+  }
+
+  trackCliOptionEnvironment(environment?: string) {
+    if (environment) {
+      this.trackCliOption({
+        option: 'environment',
+        value: this.redactedValue,
```
EndangeredMassa: **suggestion (soft-blocking):** track system environments

If this value is a system environment (`production`, `preview`), we should track that value. If it's any other value, we should use `this.redactedValue`.‏

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -376,6 +376,9 @@ describe('build', () => {
     }
     expect(client.telemetryEventStore).toHaveTelemetryEvents([
       { key: 'flag:yes', value: 'TRUE' },
+      // from `vc env pull`
+      { key: 'argument:filename', value: '[REDACTED]' },
+      { key: 'option:environment', value: '[REDACTED]' },
```
EndangeredMassa: **suggestion (non-blocking):** We can remove these later after we resolve this command-to-command logic.

See: https://github.com/vercel/vercel/pull/12372/files‏

packages/cli/test/unit/commands/env/pull.test.ts
```diff
@@ -481,7 +481,71 @@ describe('env pull', () => {
     expect(rawDevEnv.toString().includes('VERCEL_ANALYTICS_ID')).toBeFalsy();
   });
 
-  describe.todo('[filename]');
-  describe.todo('--yes');
-  describe.todo('--git-branch');
+  describe('[filename]', () => {
```
EndangeredMassa, line 529: **suggestion (non-blocking):** use existing tests

We can add these metrics assertions to existing tests. I did this in my attempt here: https://github.com/vercel/vercel/pull/12369/files#diff-f323de450c6e0956c9afe44145acae8e46e44b58d7017ebef9abdef006cffd6dR39

I'm not against explicit tests for them, though.‏

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/env/pull.test.ts
trek: One thing the single übertest doesn't protect us from us is accidentally passing the value to telemetry even when it wasn't passed to the CLI.

trek (approved):

PR #12373, opened 2024-10-25
[cli] add telemetry tracking to `git connect` and `git disconnect`

trek:
These were small enough to tackle as a group. `git disconnect` has no args, options, or flags, so no client needed there.

### DIFF
--- a/packages/cli/test/unit/commands/git/connect.test.ts
+++ b/packages/cli/test/unit/commands/git/connect.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, beforeEach, afterEach, expect, it } from 'vitest';
 import { join } from 'path';
 import fs from 'fs-extra';
 import { useUser } from '../../../mocks/user';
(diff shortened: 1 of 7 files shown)


packages/cli/test/unit/commands/git/connect.test.ts
```diff
@@ -9,25 +9,134 @@ import git from '../../../../src/commands/git';
...
+          },
+          {
+            key: 'flag:confirm',
+            value: 'TRUE',
+          },
+        ]);
+      });
+    });
+
+    it('connects an unlinked project', async () => {
```
trek, line 137: I shifted this one since it now shares some setup with the telemetry tests

packages/cli/test/unit/commands/git/disconnect.test.ts
```diff
@@ -12,6 +12,51 @@ describe('git disconnect', () => {
   const fixture = (name: string) =>
     join(__dirname, '../../../fixtures/unit/commands/git/connect', name);
 
+  it('tracks subcommand invocation', async () => {
```
trek, line 15: mostly copypasta from a similar test. We'll swing through cleaning these up some day.

EndangeredMassa (approved):

PR #12377, opened 2024-10-25
[cli] add subcommand tracking for `integration` group

trek:
No additional tests. The invocation requires a subcommand and the telemetry store updates are reflected in those tests as the first entry.

### DIFF
--- a/packages/cli/src/util/telemetry/commands/integration/index.ts
+++ b/packages/cli/src/util/telemetry/commands/integration/index.ts
@@ -1,0 +1,31 @@
+import { TelemetryClient } from '../..';
+
+export class IntegrationTelemetryClient extends TelemetryClient {
+  trackCliSubcommandAdd(actual: string) {
+    this.trackCliSubcommand({
+      subcommand: 'add',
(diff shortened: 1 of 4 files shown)


packages/cli/src/util/telemetry/commands/integration/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      value: actual,
+    });
+  }
+
+  trackCliSubcommandMove(actual: string) {
+    this.trackCliSubcommand({
+      subcommand: 'move',
+      value: actual,
+    });
+  }
```
EndangeredMassa: **suggestion (non-blocking):** ‏remove?

Will we need this in the near future?

EndangeredMassa (dismissed):

packages/cli/src/util/telemetry/commands/integration/index.ts
trek: Nope! Copy/Paste error

EndangeredMassa (approved):

PR #12378, opened 2024-10-25
extract vc pull logic from entrypoint

EndangeredMassa:
This is a very basic extraction of the core `vc pull` logic out of the command invocation for use in the build command invocation. This gives us a place to add telemetry later in only the `vc pull` invocation logic so that invoking `vc build` does not trigger `vc pull` telemetry metrics.

I manually tested this to make sure I didn't break it.

### DIFF
--- a/packages/cli/src/commands/pull/index.ts
+++ b/packages/cli/src/commands/pull/index.ts
@@ -96,7 +96,7 @@ export default async function main(client: Client) {
   return returnCode;
 }
 
-async function pullCommandLogic(
+export async function pullCommandLogic(
   client: Client,
   cwd: string,
   autoConfirm: boolean,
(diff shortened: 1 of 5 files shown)


trek (approved):

jeffsee55 (approved):

TooTallNate (approved):

packages/cli/src/commands/pull/index.ts
```diff
@@ -25,12 +26,12 @@ async function pullAllEnvFiles(
   environment: string,
   client: Client,
   link: ProjectLinked,
-  flags: PullCommandFlags,
```
EndangeredMassa, line 28: This was the wrong type before, but the types happened to be compatible.

PR #12380, opened 2024-10-25
[cli] add telemetry for `vercel dns ls`

trek:


packages/cli/src/util/telemetry/commands/dns/ls.ts
```diff
@@ -0,0 +1,30 @@
+import { TelemetryClient } from '../..';
+
+export class DnsLsTelemetryClient extends TelemetryClient {
+  trackArgumentDomainName(domainName?: string) {
```
trek: Ah, these should be `trackCli<>` format. Will fix.

EndangeredMassa (dismissed):

TooTallNate (dismissed):

packages/cli/test/unit/commands/dns/ls.test.ts
```diff
@@ -1,32 +1,104 @@
...
+          {
+            key: 'option:limit',
+            value: '[REDACTED]',
+          },
+        ]);
+      });
+    });
+
+    describe('--next', () => {
+      it('should list up to 2 dns if limit set to 2', async () => {
```
onsclom:
```suggestion
      it('tracks the use of next option', async () => {
```

onsclom (dismissed):

erikareads (approved):

PR #12381, opened 2024-10-25
[cli] add telemetry for `vercel dns remove`

trek:
This one required a few existing behavior refactors. Explanation within.


packages/cli/src/commands/dns/rm.ts
```diff
@@ -75,7 +84,7 @@ function readConfirmation(
     output.print(
       `${chalk.bold.red('> Are you sure?')} ${chalk.gray('(y/N) ')}`
     );
-    process.stdin
+    client.stdin
```
trek, line 87: We can't easily expect on `process`, so needed to swap to `client.stdin`. This changed the arguments to this function a tiny bit.

TooTallNate (approved):

PR #12385, opened 2024-10-25
Track standard environments in `vc env add`

onsclom:
Previously I redacted all environments, but it should track `production`, `preview`, and `development`.


packages/cli/test/unit/commands/env/add.test.ts
```diff
@@ -64,7 +64,7 @@ describe('env add', () => {
           },
           {
             key: `argument:environment`,
-            value: '[REDACTED]',
+            value: 'preview',
```
TooTallNate, line 67: Is there a test case for when `--environment` is a redacted value?

onsclom: Good catch, added a test case!

erikareads (approved):

TooTallNate (approved):

trek (approved):

PR #12387, opened 2024-10-25
[cli] add telemetry to `vc pull`

EndangeredMassa:
Add telemetry to `vc pull`.


packages/cli/src/commands/pull/command.ts
```diff
@@ -32,6 +32,12 @@ export const pullCommand = {
       type: String,
       deprecated: false,
     },
+    {
+      name: 'prod',
```
EndangeredMassa, line 36: This flag was missing.

trek (approved):

PR #12393, opened 2024-10-25
[next] Re-enable e2e tests

TooTallNate:
e2e tests for the Next.js builder were disabled in #12349.


ijjk (approved):
This should be good now as we tracked down the underlying issue

PR #12404, opened 2024-10-28
[cli] Add telemetry for `vercel integration list`

TooTallNate:


packages/cli/src/util/telemetry/commands/integration/list.ts
```diff
@@ -0,0 +1,27 @@
...
+        value: this.redactedValue,
+      });
+    }
+  }
+
+  trackCliOptionIntegration(v?: string) {
+    if (v) {
+      this.trackCliOption({
+        option: 'integration',
+        value: this.redactedValue,
```
erikareads: Integrations are vercel known, so we could send this through unredacted.

erikareads (dismissed):

packages/cli/src/util/telemetry/commands/integration/list.ts
TooTallNate: For `vc init`, we redact unless it's verified to be a known template name (i.e. it's redacted if an unknown template / typo'd).

I'm not seeing that we get a list like that in the `integration list` command though - looks like the value is used as a client-side filter on the API response. Perhaps we could send redacted if there's at least one matching after the filtering though.

TooTallNate: @erikareads Please see [`3e43c18` (#12404)](https://github.com/vercel/vercel/pull/12404/commits/3e43c186a18b41caf6c28e67ba79face4b46567a)

erikareads (dismissed):

trek (dismissed):

erikareads (approved):

PR #12409, opened 2024-10-28
[cli] Add telemetry tracking for `vercel domains add`

erikareads:
- **[cli] Add telemetry for `vercel domains add`**
- **add changeset**


packages/cli/test/unit/commands/domains/add.test.ts
```diff
@@ -1,9 +1,66 @@
...
 
-describe.todo('domains add', () => {
-  describe.todo('[name]', () => {
-    describe.todo('[project]', () => {
-      describe.todo('--force');
+describe('domains add', () => {
+  describe('[name]', () => {
+    describe('[project]', () => {
+      describe('--force', () => {
+        it('tracks telemetry data, async', async () => {
```
TooTallNate, line 12: What does "async" mean here?

packages/cli/src/commands/domains/add.ts
```diff
@@ -25,6 +26,14 @@ export default async function add(
...
 ) {
+  const { telemetryEventStore } = client;
+  const telemetry = new DomainsAddTelemetryClient({
+    opts: {
+      store: telemetryEventStore,
+    },
+  });
+
+  telemetry.trackCliFlagForce(opts['--force']);
   const force = opts['--force'];
```
TooTallNate: Swapping the order of these two lines would help (slightly) with minification.

```suggestion
  const force = opts['--force'];
  telemetry.trackCliFlagForce(force);
```

TooTallNate (dismissed):
Minor nit and a question. Nothing blocking though.

packages/cli/test/unit/commands/domains/add.test.ts
erikareads: I copied that from another `vercel domains` telemetry PR that @trek wrote: https://github.com/vercel/vercel/blob/d1743d1757f51ff94f72b714ff048584c9abcd31/packages/cli/test/unit/commands/domains/ls.test.ts#L43

trek: Yeah, that was just some moving-too-fast-in-my-editor junk

trek (approved):

packages/cli/test/unit/commands/domains/add.test.ts
trek: Approved to keep us going but @erikareads or I should swing back and fix both of these after.

trek: https://github.com/vercel/vercel/pull/12416

PR #12414, opened 2024-10-29
[cli] add metrics to dns add

EndangeredMassa:
Add metrics to `vercel dns add ...`.


packages/cli/src/commands/dns/add.ts
```diff
@@ -35,6 +36,19 @@ export default async function add(
 
   const addStamp = stamp();
   const { domain, data: argData } = parsedParams;
+  const valueArgs = args.slice(3); // domain, name, type, ...valueArgs
```
EndangeredMassa, line 39: I looked into updating `parseAddDNSRecordArgs` to make this more directly available, but it gets complicated with (1) the different returns types and (2) passing this exact value to the API endpoint later in the flow.

packages/cli/src/commands/dns/add.ts
```diff
@@ -35,6 +36,19 @@ export default async function add(
...
+  let telemetryClient = new DnsAddTelemetryClient({
+    opts: {
+      store: client.telemetryEventStore,
+    },
+  });
+
+  telemetryClient.trackCliArgumentDomain(domain);
+  telemetryClient.trackCliArgumentName(parsedParams.data?.name);
+  telemetryClient.trackCliArgumentType(parsedParams.data?.type);
+  telemetryClient.trackCliArgumentValues(valueArgs);
```
EndangeredMassa, line 50: I wasn't sure how best to handle the various arguments used by this command.

Given:

```
RecordType = [A || AAAA || ALIAS || CNAME || TXT]
```

The variants are:

```
vercel dns add [domain] [subdomain] [RecordType] [value]
vercel dns add [domain] '@'         MX           [record-value]            [priority]
vercel dns add [domain] [name]      SRV          [priority]                [weight]   [port] [target]
vercel dns add [domain] [name]      CAA          '[flags] [tag] "[value]"'
```

So, I decided to grab domain, name, and type as explicit arguments. The rest I called value arguments.

It doesn't super matter right now because most of this gets reported as redacted, but I wanted it to be as accurate as possible so that future changes are easier to make.

packages/cli/src/util/telemetry/commands/dns/add.ts
```diff
@@ -0,0 +1,36 @@
...
+        arg: 'name',
+        value: this.redactedValue,
+      });
+    }
+  }
+  trackCliArgumentType(type?: string) {
+    if (type) {
+      this.trackCliArgument({
+        arg: 'type',
+        value: type,
```
EndangeredMassa, line 24: I think we can report type. These are a known set of values.

erikareads: I would be tempted to filter these to that known list, in case they mistype a type or miss an argument and get sent something that would otherwise be redacted.

erikareads (approved):

trek (approved):

PR #12415, opened 2024-10-29
Add debug testTimeout value

trek:
For those of us who use a debugger, the test suit exiting early is frustrating. You can get around this by setting this value yourself or running with `--testTimeout` in the terminal but adding this behavior globally seems like a nice quality of life change.

I couldn't find a 100% guaranteed way to detect if node is in debug mode but the rough consensus was detecting  `debugPort` covered the largest number of cases.

I audited all the uses of `vi.setConfig({ testTimeout });` in `packages/cli` and only found one that looked like it could definitely be removed. The use case it covered was negated by https://github.com/vercel/vercel/pull/12354.

The others mostly came from https://github.com/vercel/vercel/commit/413706d72aed50a5e16ddb3f899bbc1aa0e6a25d or had a comment explaining their need.


packages/cli/vitest.setup.mts
```diff
@@ -7,3 +7,7 @@ beforeAll(() => {
     noColor: true,
   });
 });
+
+if (process.debugPort) {
+  vi.setConfig({ testTimeout: 10 * 60 * 1000 });
```
EndangeredMassa, line 13:
```suggestion
  // when debugging in an IDE, set a high timeout
  vi.setConfig({ testTimeout: 10 * 60 * 1000 });
```

EndangeredMassa (dismissed):

erikareads (dismissed):

onsclom (dismissed):

erikareads (dismissed):

.changeset/quick-keys-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": minor
```
TooTallNate: Does this really need to trigger a release? It's only related to tests it looks like.

```suggestion
```

TooTallNate (changes requested):

erikareads (approved):

TooTallNate (approved):

EndangeredMassa (approved):

PR #12419, opened 2024-10-29
[cli] Track `git --confirm` earlier

trek:
Because both `--yes` and `--confirm` are technically allowed we want to track both. To handle the fact that `--confirm` is deprecated is managed with the following pattern: 

```
    output.warn('`--confirm` is deprecated, please use `--yes` instead');
    parsedArgs.flags['--yes'] = parsedArgs.flags['--confirm'];
```

so we need to ensure we've tracked uses of `--confirm` before we smoosh them together. @EndangeredMassa and I audited for this pattern and we've done this in all places _except_ `vercel git`.

`vercel git` needs access to the value in the outer command before the subcommand is called:

https://github.com/vercel/vercel/blob/4db2c19c0229721329aa24b00e5d19dbb72a4762/packages/cli/src/commands/git/index.ts#L52-L55

So the tracking needs to move.


packages/cli/test/unit/commands/git/connect.test.ts
```diff
@@ -126,10 +131,6 @@ describe('git connect', () => {
             key: 'flag:yes',
```
TooTallNate: Not quite related to this PR, but isn't it weird that `flag:yes` is tracked, even though the user didn't explicitly type `--yes`?

trek: @TooTallNate it _is_ yes. And means we can easily know if `--confirm` is used but the query to know if `--yes` was used directly or as a side effect of passing `--confirm` is more complicated. 

There was a discussion somewhere about fixing this but IIRC we decided that since we had no tests (yet), changing the behavior had a high risk of breaking behavior.

Now that we have tests I think we can swing back through and fix this so you only see one or the other in metrics. Whether that blocks shipping or not I'll leave up to @erikareads

trek: Yeah, we'd have to move tracking of both up into the outer command. I'll change that.

TooTallNate (approved):

erikareads (approved):

onsclom (approved):

PR #12422, opened 2024-10-29
[cli] Make all metrics tracking methods on the store require arguments

trek:
For a given telemetry method, like:

```
trackCliArgumentDomain(domain?: string) {
  if (domain) {
    this.trackCliArgument({
      arg: 'domain',
      value: this.redactedValue,
    });
  }
}
```

The argument type domain?: string means that it can be invoked without an argument at all:

```
eventStore.trackCliArgumentDomain();
```

We never want this to happen, though. The value should always be passed to determine if we're tracking that argument/flag/option.


TooTallNate (dismissed):

packages/cli/src/util/telemetry/commands/dev/index.ts
```diff
@@ -28,21 +28,21 @@ export class DevTelemetryClient extends TelemetryClient {
...
     if (confirm) {
       this.trackCliFlag('confirm');
     }
   }
 
-  trackCliFlagHelp(help?: boolean) {
+  trackCliFlagHelp(help: boolean | undefined) {
     if (help) {
-      this.trackCliFlagHelp();
+      this.trackCliFlag('help');
```
trek, line 45: This audit did actually catch one bug!

erikareads (dismissed):

packages/cli/src/util/telemetry/commands/dev/index.ts
EndangeredMassa: **question (non-blocking):** ‏Can we have a test that asserts this? Or is it too hard to do with `vc dev` for now?

EndangeredMassa (dismissed):

packages/cli/src/util/telemetry/commands/dev/index.ts
trek: I think this was actually me overachieving sine we planned to do `--help` as its own work. But, yeah, this is annoying to test. I could just remove it?

TooTallNate (approved):

erikareads (approved):

PR #12430, opened 2024-10-30
[cli] add telemetry for --help flag

trek:
Apologies for the gigantic PR, it is _mostly_ mechanical. 

How I'd review this PR:

* See `packages/cli/src/util/telemetry/index.ts` for the enhancement to `trackFlagHelp` to the remove robust `trackCliFlagHelp` that tracks more detailed `value` data than just `TRUE`.

Skipping to the bottom, which are all the tests. Verify

* file path paths match the commands and subcommand and function invoked. E.g.: `test/unit/commands/certs/issue.test.ts` has command of `certs`, subcommand of `issue`, and `--help` flag argv calling `certs(client)`. This results in telemetry value that has both `command` and `subcommand` variables.
* file path paths match the commands and _no_ subcommand and function invoked. E.g.: `test/unit/commands/certs/index.test.ts` has command of `certs`, _no subcommand_, and `--help` flag argv calling `certs(client)`.  This results in telemetry value that has _only_ `command` variables.

Then for each `src/commands/COMMAND_NAME/index.ts` file verify that:
* `telemetry.trackCliFlagHelp('COMMAND_NAME', subcommand)` or `telemetry.trackCliFlagHelp('COMMAND_NAME')`

One common refactor was that parsing `command` and `subcommand` from `argv` needed to shift up in execution so the values could be passed to `trackCliFlagHelp`

Additional comments within for any weirdos that deviate from our common pattern.


packages/cli/test/unit/commands/teams/add.test.ts
```diff
@@ -11,22 +11,39 @@ describe('teams add', () => {
...
 
-  beforeEach(() => {
     client.config = {
       currentTeam: currentTeamId,
     };
   });
 
-  client.scenario.post(`/teams`, (req, res) => {
-    return res.json(team);
-  });
```
trek, line 24: The scenario are not needed for `--help`. Moved them into the one other test in this file for now.

packages/cli/test/unit/commands/domains/transfer-in.test.ts
```diff
@@ -16,6 +16,24 @@ describe('domains transfer-in', () => {
...
+      const subcommand = 'transfer-in';
+
+      client.setArgv(command, subcommand, '--help');
+      const exitCodePromise = domains(client);
+      await expect(exitCodePromise).resolves.toEqual(2);
+
+      expect(client.telemetryEventStore).toHaveTelemetryEvents([
+        {
+          key: 'flag:help',
+          value: `${command}:transferIn`,
```
trek: I could transform this to precisely match the terminal value passed but that doesn't seem worth the additional code.

packages/cli/test/unit/commands/domains/index.test.ts
```diff
@@ -10,11 +10,28 @@ describe('domains', () => {
...
+    });
+  });
+
   it('routes to ls subcommand', async () => {
     const args: string[] = [];
     const opts = {};
 
-    client.setArgv('dns', ...args);
+    client.setArgv('domains', ...args);
     await domains(client);
```
trek, line 35: a mistake from some earlier commit.

packages/cli/src/commands/bisect/index.ts
```diff
@@ -56,6 +54,9 @@ export default async function bisect(client: Client): Promise<number> {
   telemetry.trackCliOptionRun(parsedArgs.flags['--run']);
   telemetry.trackCliFlagOpen(parsedArgs.flags['--open']);
 
+  const scope = await getScope(client);
```
trek: Moved scope lower in the order of execution. If it remains at the top we check for someone to be authetincated before allowing them to run `vercel bisect --help` which feels like a bug.

packages/cli/src/commands/integration/index.ts
```diff
@@ -45,6 +45,7 @@ export default async function main(client: Client) {
   const needHelp = flags['--help'];
 
   if (!subcommand && needHelp) {
+    telemetry.trackCliFlagHelp('integration');
```
trek, line 48: `integration` breaks with normal pattern and handles help display per subcommand.

packages/cli/src/commands/rollback/index.ts
```diff
@@ -35,10 +35,14 @@ export default async (client: Client): Promise<number> => {
     return 1;
   }
 
+  const actionOrDeployId = parsedArgs.args[1] || 'status';
+
   telemetry.trackCliOptionTimeout(parsedArgs.flags['--timeout']);
   telemetry.trackCliFlagYes(parsedArgs.flags['--yes']);
 
   if (parsedArgs.flags['--help']) {
+    const subcommand = actionOrDeployId === 'status' ? 'status' : undefined;
```
trek, line 44: rollback technically has a subcommand (`status`) and a default, but is not modeled that way, so it's help tracking implementation deviates from the patterns elsewhere.

packages/cli/src/index.ts
```diff
@@ -358,6 +358,7 @@ const main = async () => {
   }
 
   if (subcommand === 'help') {
+    telemetry.trackCliCommandHelp('help');
```
trek, line 361: Track the `vercel help` use. Because of how this _then_  pushes `-h` into the `argv` stack, we'll _also_ get an entry for `flag:help` for the command in question. We deemed this acceptable for now when discussing the card.

packages/cli/src/util/telemetry/commands/dev/index.ts
```diff
@@ -44,10 +44,4 @@ export class DevTelemetryClient
       this.trackCliFlag('confirm');
     }
   }
-
-  trackCliFlagHelp(help: boolean | undefined) {
```
trek, line 48: Handled by its parent class now.

packages/cli/src/util/telemetry/index.ts
```diff
@@ -157,8 +157,16 @@ export class TelemetryClient {
     return;
   }
 
-  trackFlagHelp() {
-    this.trackCliFlag('help');
+  trackCliFlagHelp(command: string, subcommands?: string | string[]) {
```
trek, line 160: The card describes making this `protected`, but for that to work we'd need to add

```
trackCliFlagHelp(command: string, subcommands?: string | string[]) {
  super.trackCliFlagHelp(command, subcomands)
}
```

to every child class. This seems like overkill to protect from this method being called on instances of the base class so I called an audible and made the method public. Happy to do otherwise if there's a compelling reason.

packages/cli/src/commands/alias/index.ts
```diff
@@ -12,7 +12,6 @@ import { AliasTelemetryClient } from '../../util/telemetry/commands/alias';
 import output from '../../output-manager';
 
 const COMMAND_CONFIG = {
-  default: ['set'],
```
trek: None of our other `COMMAND_CONFIG` have this structure which makes it fit poorly into the help pattern. We already branch into a default cause in the switch statement for this.

packages/cli/src/util/telemetry/index.ts
```diff
@@ -157,8 +157,16 @@ export class TelemetryClient {
...
+    let subcommand;
+    if (subcommands) {
+      subcommand = Array.isArray(subcommands) ? subcommands[0] : subcommands;
```
erikareads, line 163: Did you find an instance of multiple subcommands?

packages/cli/src/commands/project/index.ts
```diff
@@ -39,15 +39,16 @@ export default async function main(client: Client) {
     return 1;
   }
 
+  parsedArgs.args = parsedArgs.args.slice(1);
+  subcommand = parsedArgs.args[0] || 'list';
+  const args = parsedArgs.args.slice(1);
+
   if (parsedArgs.flags['--help']) {
+    telemetryClient.trackCliFlagHelp('project', parsedArgs.args[0]);
```
erikareads, line 47: Make a helper here to only track known values.

packages/cli/test/unit/commands/domains/transfer-in.test.ts
erikareads: Please change this to precisely match, so it matches our existing subcommand event.

erikareads (changes requested):

packages/cli/src/util/telemetry/index.ts
trek: I didn't check, but the types claim they're allowed so I followed the current types.

packages/cli/src/commands/bisect/index.ts
trek: Extracted to https://github.com/vercel/vercel/pull/12439

packages/cli/src/commands/alias/index.ts
trek: Extracted to https://github.com/vercel/vercel/pull/12438

jeffsee55 (dismissed):

erikareads (dismissed):

erikareads (approved):

PR #12435, opened 2024-10-30
Set `hideSkippedTests` to `true` Vitest config

jeffsee55:
When there are lots of tests and you try to isolate one with `.only`, Vitest still prints the names of the skipped ones:

Before
Screenshot 2024-10-30 at 12 08 07 PM

After
Screenshot 2024-10-30 at 12 08 21 PM


trek (approved):

EndangeredMassa:
> Before Screenshot 2024-10-30 at 12 08 07 PM
> 
> After Screenshot 2024-10-30 at 12 08 21 PM

These look flipped. Not a big deal.

Great change!

jeffsee55:
> These look flipped. Not a big deal.

Woops, yes you're right

PR #12436, opened 2024-10-30
Fix and re-enable `api-env` test

onsclom:
This test would fail when it ran for two different PRs at the same time. For example, PR A may add an env var just before PR B checks that there are no env vars.

This fixes the test by:
1. using unique env var names per run
2. removing the no env var expectation


packages/cli/test/integration-1.test.ts
```diff
@@ -834,9 +838,9 @@ test.skip('deploy `api-env` fixture and test `vercel env` command', async () =>
     const homeUrl = localhost[0];
     const homeRes = await fetch(homeUrl);
     const homeJson = await homeRes.json();
-    expect(homeJson['VERCEL']).toBeUndefined();
+    expect(homeJson['VERCEL']).toBe('1');
     expect(homeJson['VERCEL_URL']).toBe(localhostNoProtocol);
-    expect(homeJson['VERCEL_ENV']).toBeUndefined();
+    expect(homeJson['VERCEL_ENV']).toBe('development');
```
onsclom, line 844: These have values in dev now because of https://github.com/vercel/vercel/pull/12358

packages/cli/test/integration-1.test.ts
```diff
@@ -750,18 +753,19 @@ test.skip('deploy `api-env` fixture and test `vercel env` command', async () =>
...
-    expect(homeJson['MY_NEW_ENV_VAR']).toBe('my plaintext value');
-    expect(homeJson['MY_STDIN_VAR']).toBe('{"expect":"quotes"}');
+    expect(homeJson[promptEnvVar]).toBe('my plaintext value');
+    expect(homeJson[stdinEnvVar]).toBe('{"expect":"quotes"}');
 
     // system env vars are hidden in dev
     expect(apiJson['VERCEL']).toBeUndefined();
-    expect(homeJson['VERCEL']).toBeUndefined();
+    // though the dev server now has this
+    expect(homeJson['VERCEL']).toBe('1');
```
onsclom, line 769: Not completely sure what the different between `apiJson` and `homeJson` is here, but https://github.com/vercel/vercel/pull/12358 makes `VERCEL` available in `homeJson`

trek (dismissed):

TooTallNate (dismissed):

packages/cli/test/integration-1.test.ts
```diff
@@ -521,8 +522,11 @@ test('deploy using --local-config flag above target', async () => {
 });
 
 // eslint-disable-next-line jest/no-disabled-tests
```
erikareads: we can probably remove this now

packages/cli/test/integration-1.test.ts
```diff
@@ -521,8 +522,11 @@ test('deploy using --local-config flag above target', async () => {
...
-test.skip('deploy `api-env` fixture and test `vercel env` command', async () => {
+test('deploy `api-env` fixture and test `vercel env` command', async () => {
   const target = await setupE2EFixture('api-env');
+  const promptEnvVar = `VAR_${randomBytes(8).toString('hex')}`;
```
erikareads, line 528: would `chance` be a better fit here?

erikareads (dismissed):

packages/cli/test/integration-1.test.ts
EndangeredMassa: I strongly discourage use of randomness in standard test suites. It makes the tests brittle without providing value, in my opinion. Instead, I would pick some arbitrary fixed value.

EndangeredMassa: I (generally) strongly discourage use of randomness in standard test suites. It makes the tests brittle without providing value, in my opinion.

Here, the goal is to make sure these values are unique among other runs, which is a unique situation. Randomness can be fine for this purpose, but I would add a comment about why this is necessary. Otherwise, I might forget in the future and come by to remove this randomness.

erikareads (approved):

jeffsee55 (approved):

PR #12441, opened 2024-10-30
[tests] refactor exit code assertions

EndangeredMassa:
The assertion `expect(somePromise).resolves.toEqual(someValue)` has two failure modes:

1. the resolved value doesn't match the given value
2. the promise is rejected

In the (1) case, the failure message is fine. However, the (2) case truncates the rejection message:

```
 FAIL  test/unit/commands/link/index.test.ts > link > should track use of deprecated `cwd` positional argument
AssertionError: promise rejected "'some long message about why this is b…'" instead of resolving
 ❯ test/unit/commands/link/index.test.ts:348:19
    346|     // const exitCode = await exitCodePromise;
    347|     const p = Promise.reject('some long message about why this is brok…
    348|     await expect(p).resolves.toEqual(1);
       |                   ^
    349| 
    350|     expect(client.telemetryEventStore).toHaveTelemetryEvents([
```

The reason is truncated like this:

> some long message about why this is b…

Which makes this assertion pretty useless compared to the basic replacement.

---

This PR replaces the general pattern:

```
const exitCodePromise = alias(client);
```

with:

```
const exitCode = await alias(client);
expect(exitCode, 'exit code of "alias"').toEqual(0);
```

---

Example of new error:

```
 FAIL  test/unit/commands/link/index.test.ts > link > should track use of deprecated `cwd` positional argument
Error: some long message about why this is broken that you cannot read
 ❯ test/unit/commands/link/index.test.ts:348:14
    346|     // const exitCode = await exitCodePromise;
    347|     const p = new Promise((resolve, reject) => {
    348|       reject(new Error('some long message about why this is broken tha…
       |              ^
    349|     })
    350|     const v = await p;
 ❯ test/unit/commands/link/index.test.ts:347:15
```

Which shows the reason:

> some long message about why this is broken that you cannot read


packages/cli/test/unit/commands/domains/ls.test.ts
```diff
@@ -11,16 +11,18 @@ describe('domains ls', () => {
     useDomains();
     client.setArgv('domains', 'ls');
     let exitCodePromise = domains(client);
-    await expect(exitCodePromise).resolves.toEqual(0);
+    await expect(client.stderr).toOutput('example-19.com');
```
EndangeredMassa:
```suggestion
```

packages/cli/test/unit/commands/domains/ls.test.ts
```diff
@@ -11,16 +11,17 @@ describe('domains ls', () => {
     useDomains();
     client.setArgv('domains', 'ls');
     let exitCodePromise = domains(client);
```
erikareads: Is there a reason we don't just call:
```
const exitCode = await domains(client);
```
here?

packages/cli/test/unit/commands/domains/move.test.ts
erikareads, line 1: The 
```
 describe('--yes', () => {
  it('tracks telemetry data', async () => {
```
block has another promise that can be refactored.

packages/cli/test/unit/commands/integration/add.test.ts
```diff
@@ -333,7 +339,8 @@ describe('integration', () => {
         it('should error when no integration arugment was passed', async () => {
           client.setArgv('integration', 'add');
           const exitCodePromise = integrationCommand(client);
```
erikareads: Do we want to move these cases where we directly `await` the promise into a single `const exitCode = await <command>;` instances?

packages/cli/test/unit/commands/domains/ls.test.ts
EndangeredMassa: I was still working through. I'll make this ready for review when I'm done.

TooTallNate (approved):
I also like that this will make it harder to forget the `await` in `await expect(promise)`, which we had to fix up a few of those last week 👍

trek (approved):

PR #12103, opened 2024-09-11
[cli] tidy up some command structures

trek:
We initially thought these were only used for generating help docs, but discovered they are in fact used in parsing arg/options since https://github.com/vercel/vercel/pull/11929. Instead of one large PR, we'll be pulling these changes in incrementally as we add telemetry to each command.

PR #12446, opened 2024-10-30
[cli] Add strongly typed `subcommands` for Telemetry methods based on command specification

TooTallNate:
* Adds the `subcommads` Telemetry methods (i.e. `trackCliSubcommandOpen()`) to the strong typing generic type.
* Applies the interface to the `vercel integration` family of (sub)commands, since their commands specification data is already in the correct format.

For the rest of the (sub)commands, we still need to wait for https://github.com/vercel/vercel/pull/12103, but at least the generic type is now prepared for `subcommands` to be properly defined in the commands' spec.


packages/cli/src/util/telemetry/commands/integration/index.ts
```diff
@@ -28,4 +33,9 @@ export class IntegrationTelemetryClient extends TelemetryClient {
       value: actual,
     });
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  trackCliArgumentCommand(value: string | undefined) {
+    // no-op?
```
erikareads, line 39: What's the context here?

packages/cli/src/util/telemetry/commands/integration/list.ts
```diff
@@ -10,7 +15,7 @@ export class IntegrationListTelemetryClient extends TelemetryClient {
     }
   }
 
-  trackCliOptionIntegration(v: string | undefined, known: boolean | undefined) {
+  trackCliOptionIntegration(v: string | undefined, known?: boolean) {
```
erikareads, line 18: Why does this type change?

packages/cli/src/util/telemetry/commands/integration/open.ts
```diff
@@ -1,7 +1,12 @@
...
+import type { TelemetryMethods } from '../../types';
+import type { openSubcommand } from '../../../../commands/integration/command';
 
-export class IntegrationOpenTelemetryClient extends TelemetryClient {
-  trackCliArgumentName(v: string, known: boolean | undefined) {
+export class IntegrationOpenTelemetryClient
+  extends TelemetryClient
+  implements TelemetryMethods<typeof openSubcommand>
+{
+  trackCliArgumentName(v: string | undefined, known?: boolean) {
```
erikareads, line 9: What's with the type signature change?

packages/cli/src/util/telemetry/commands/integration/remove.ts
```diff
@@ -1,7 +1,12 @@
...
+import type { TelemetryMethods } from '../../types';
+import type { removeSubcommand } from '../../../../commands/integration/command';
 
-export class IntegrationRemoveTelemetryClient extends TelemetryClient {
-  trackCliArgumentIntegration(v?: string, known?: boolean) {
+export class IntegrationRemoveTelemetryClient
+  extends TelemetryClient
+  implements TelemetryMethods<typeof removeSubcommand>
+{
+  trackCliArgumentIntegration(v: string | undefined, known?: boolean) {
```
erikareads, line 9: Why `known?: boolean` instead of `known: boolean | undefined`? The latter _requires_ the argument to be passed, the former doesn't.

packages/cli/src/util/telemetry/commands/integration/index.ts
TooTallNate: In `integration/command.ts`, it defines a single argument "command" which presumably is referring to the subcommand - thus, the generic type generates this method name. Question is - do we want to track that "argument:command" separately from the individual "subcommand" methods?

If yes, then we can implement this function. However, it seems redundant to me to track `{ key: "argument:command", value: "ls" }` as well as `{ key: "subcommand:list", value: "ls" }`.

If not, then I could get behind the idea of removing the explicit "command" argument (in `command.ts`), and then we make the presence of the `subcommands` array imply that there is a "command" argument (specifically for the help output), but it wouldn't be included in the telemetry generic and therefore we could remove this function (this would all be a follow-up PR if we want to go this route).

packages/cli/src/util/telemetry/commands/integration/list.ts
TooTallNate: The type from the generic interface is `trackCliOptionIntegration(v: string | undefined)` (single parameter), so in TypeScript any additional params must be optional so that the original interface remains conformant.

It _might_ be possible to work around this by defining `...rest: unknown[]` on the generic interface, but I haven't tested that out.

In this particular pattern of this "known" parameter, I'm ok with having it be optional since absence of the param would default to the "safe" choice (tracking a redacted value). But if there's a case where we definitely want a required 2nd+ parameter, then I'm down to investigate further.

packages/cli/src/util/telemetry/commands/integration/open.ts
TooTallNate: https://github.com/vercel/vercel/pull/12446/files#r1823797083

packages/cli/src/util/telemetry/commands/integration/remove.ts
TooTallNate: https://github.com/vercel/vercel/pull/12446/files#r1823797083

packages/cli/src/util/telemetry/commands/integration/list.ts
TooTallNate: Ok, that `...rest` trick _does_ work, and allows for additional, required, parameters to be defined. The only downside is that when auto implementing the interface, the editor includes the `...rest` parameter in the function signature (which will be deleted most of the time):

```
  trackCliArgumentName: (value: string | undefined, ...rest: any[]) => void;
```

So while that does work, I'm kind of ±0 on the change - not sure if the DX of "fixing up" that `...rest` parameter is a worthy tradeoff. But I could easily have my mind changed about that if we can come up with an example of when a required 2nd parameter would make sense.

erikareads (approved):

packages/cli/src/util/telemetry/commands/integration/index.ts
erikareads: No, tracking the argument _and_ the subcommand feels redundant, I would be interested to see what a removal of the command argument PR would look like.

PR #12450, opened 2024-10-31
[tests] enable git command index test

trek:
These required additional setup so were left as `todo`. That setup exists now.


packages/cli/test/unit/commands/git/index.test.ts
```diff
@@ -21,9 +27,22 @@ describe.todo('git', () => {
     });
   });
 
-  it('errors when invoked without subcommand', async () => {
-    client.setArgv('git');
-    const exitCodePromise = git(client);
-    await expect(exitCodePromise).resolves.toBe(2);
+  it('displays help when invoked without subcommand', async () => {
+    const cwd = fixture('new-connection');
```
trek, line 31: Honestly not a great sign all of this is a precondition to saying "you need to supply a subcommand"

packages/cli/test/unit/commands/git/index.test.ts
```diff
@@ -21,9 +27,22 @@ describe.todo('git', () => {
...
+      await fs.rename(join(cwd, 'git'), join(cwd, '.git'));
+      useTeams('team_dummy');
+      useProject({
+        ...defaultProject,
+        id: 'new-connection',
+        name: 'new-connection',
+      });
+      client.setArgv('git');
+      const exitCodePromise = git(client);
+      await expect(exitCodePromise).resolves.toBe(2);
```
EndangeredMassa: We should not use the `.resolves` assertion. See: https://github.com/vercel/vercel/pull/12441

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #12455, opened 2024-10-31
[test] move e2e test to unit test

trek:
Not _much_ of a speed gain but this does save us the cost of file system interactions and locks the entire current output as an expectation.


erikareads:
`help.test.ts` already covers most of the command outputs in snapshots, we could put `vercel dev` help output here: https://github.com/vercel/vercel/blob/main/packages/cli/test/unit/commands/help.test.ts

trek:
@erikareads I looked at that as well (specifically to limit the number of `__snapshots__ directories) but the test setup of `vc help [command]` and `vc [comand] --help` differ in a few ways (needing a mock client, sometimes needing other setup). 

I can do an alternate PR to better compare.

trek:
@erikareads compare to https://github.com/vercel/vercel/pull/12457

Structurally I like limiting the number of `__snapshots__` directories, but 

* there will be some duplicated setup code for those commands that don't run `--help` before doing some extra work
* `test/commands/help.ts` maps to `vc help`, not necessarily `vc [command] --help`

trek:
If we went with this approach I'd argue we should move `test/commands/help.test.ts` to something like `test/help-output.test.ts` to make it better apply to both `vc help [command] and `vc [command] --help`

trek:
We're going with option 2!

PR #12456, opened 2024-10-31
[cli] Make `command` be an implied argument when `subcommands` array is defined

TooTallNate:
Follow-up for this comment: https://github.com/vercel/vercel/pull/12446#discussion_r1823795966

The `{ name: 'command', required: boolean }` argument no longer needs to be defined - it gets generated in the `--help` output implicitly when there's > 1 `subcommand` entry.

* If a subcommand has `default: true` defined, then the help generator will output `[command]` instead of `command`
* Helps with data consistency:
  * There was at least one case where the "command" arg was missing - but subcommands existed (`vc teams`)
  * A handful of commands had `required: true` when the opposite was correct (and vice versa) - see the updated snapshot file


packages/cli/src/util/telemetry/commands/integration/index.ts
```diff
@@ -33,9 +33,4 @@ export class IntegrationTelemetryClient
       value: actual,
     });
   }
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  trackCliArgumentCommand(value: string | undefined) {
```
TooTallNate, line 38: @erikareads Since `command` is no longer explicitly defined as an argument, this no-op function is no longer necessary.

packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap
```diff
@@ -361,7 +361,7 @@ exports[`help command > bisect help output snapshots > bisect help column width
 
 exports[`help command > certs help output snapshots > certs help column width 40 1`] = `
 "
-  ▲ vercel certs [command] [options]
+  ▲ vercel certs command [options]
```
erikareads, line 364: why are we missing square brackets here?

erikareads (approved):

packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap
TooTallNate: Because the subcommand is required for `vercel certs` 😄 

```
$ vc certs
Vercel CLI 37.14.0
Error: Please specify a valid subcommand: ls | issue | rm
```

PR #12458, opened 2024-10-31
[cli] Migrate `vercel alias` subcommands to their own command specifications

TooTallNate:


.changeset/quiet-hairs-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': major
```
trek, line 2: Probably more a patch, no?

trek (dismissed):

.changeset/quiet-hairs-wave.md
erikareads: This PR moves parsing into the individual subcommands, so I think the major is justified.

packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap
```diff
@@ -2,38 +2,24 @@
...
-                         alias  
-  rm   alias             Remove 
-                         an     
-                         alias  
-                         using  
-                         its    
-                         hostna…
-
-
-  Options:
```
erikareads, line 25: I would feel more comfortable making this help change after we've added help output for the subcommands, which have these options.

.changeset/quiet-hairs-wave.md
TooTallNate: I forgot to fill out the PR description. So it depends on how pedantic we want to be here. I put major here because technically this would "work" previously:

```
vc alias --next 12345 mydeployment-ns0sbhzj6-trek.vercel.app alias-test.vercel.app
```

It's not correct usage, and the `--next` flag would be ignored, but the command would succeed.

With this change that same command now fails (which is correct, but could be considered a breaking change technically):

```
Error: unknown or unexpected option: --next
```

erikareads (changes requested):

.changeset/quiet-hairs-wave.md
trek: Ah, yeah, that's a breaking change then!

packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap
TooTallNate: Two options:

1. Wait for the subcommand help before merging this.
2. Put these options back at the root level, so that they will remain included in the help output. This means we'll have to define no-op telemetry methods on the root-level `AliasTelemetryClient` so that it satisfies the type interface.

TooTallNate: 2. might be able to be considered not a semver major change as well.

erikareads: We do have prior art on handling help for subcommands here: https://github.com/vercel/vercel/blob/cc2f23887fd762e8b21bff39b2257a9d7f2b5175/packages/cli/src/commands/integration/index.ts#L68-L72

It's a bit awkward and should eventually be centralized, but this method can work for the current PR.

erikareads (dismissed):

erikareads (dismissed):

erikareads (approved):

PR #12460, opened 2024-10-31
[cli] Migrate `vercel telemetry` subcommands to their own command specifications

TooTallNate:
Moved the subcommand definitions to their own exports, for when we generate subcommand help output.

Also gave a small visual overhaul to the `status` subcommand because it was missing a trailing newline and should prefer to use `output.log()` as much as possible for consistency with other CLI commands.


packages/cli/src/commands/telemetry/status.ts
```diff
@@ -1,23 +1,26 @@
 import chalk from 'chalk';
-import Client from '../../util/client';
 import output from '../../output-manager';
+import type Client from '../../util/client';
 
 export default async function status(client: Client) {
-  const status: 'disabled' | 'enabled' =
-    client.config.telemetry?.enabled === false ? 'disabled' : 'enabled';
+  const disabled = client.config.telemetry?.enabled === false;
```
erikareads: Can we frame the condition positively here?

erikareads (changes requested):

packages/cli/src/commands/telemetry/status.ts
erikareads:
```
const enabled = client.config.telemetry?.enabled ?? false;
```

TooTallNate: Sure thing 2327930b433a471b7de2cd9e8cf1a1089dbe97c6.

erikareads (approved):

PR #12468, opened 2024-11-01
fix small issues telemetry and whoami

erikareads:
- **[cli] type only import**
- **[cli] type only import in telemetry enable**
- **[cli] type only import in telemetry disable**
- **[cli] prefer literal strings if there is no interpolation**
- **[cli] prefer literal strings when there is no interpolation**
- **add changeset**


TooTallNate (dismissed):

.changeset/pretty-cheetahs-taste.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
EndangeredMassa, line 3: **suggestion (non-blocking):** ‏this should be a patch

I know the changes should be no-ops, but it's technically possible for them to break something or change behavior.

EndangeredMassa (dismissed):

onsclom (approved):

PR #12472, opened 2024-11-01
[cli] derive alias Map from Command structs

erikareads:
- **[cli] derive Map of aliases from command structs**
- **add changeset**

### DIFF
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -8,7 +8,7 @@ import { dnsCommand } from './dns/command';
 import { domainsCommand } from './domains/command';
 import { envCommand } from './env/command';
 import { gitCommand } from './git/command';
-import { Command } from './help';
+
 import { initCommand } from './init/command';
 import { inspectCommand } from './inspect/command';
 import { installCommand } from './install/command';
@@ -22,7 +22,6 @@ import { logsCommand } from './logs/command';
 import { projectCommand } from './project/command';
 import { promoteCommand } from './promote/command';
 import { pullCommand } from './pull/command';
-
 import { redeployCommand } from './redeploy/command';
 import { removeCommand } from './remove/command';
 import { rollbackCommand } from './rollback/command';
@@ -30,6 +29,7 @@ import { targetCommand } from './target/command';
 import { teamsCommand } from './teams/command';
 import { telemetryCommand } from './telemetry/command';
 import { whoamiCommand } from './whoami/command';
+import type { Command } from './help';
 
 const commands = [
   aliasCommand,
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -1,3 +1,86 @@
+import { aliasCommand } from './alias/command';
+import { bisectCommand } from './bisect/command';
+import { buildCommand } from './build/command';
+import { certsCommand } from './certs/command';
+import { deployCommand } from './deploy/command';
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/index.ts
```diff
@@ -1,3 +1,86 @@
...
+  for (const command of commands) {
+    const aliases = getAliases(command);
+    for (const alias of aliases) {
+      aliasMap.set(alias, command.name);
+    }
+  }
+  return aliasMap;
+}
+
 export default new Map([
```
TooTallNate: Any reason to not replace this default export instead of adding the new function?

packages/cli/src/commands/index.ts
```diff
@@ -1,3 +1,86 @@
...
+import { promoteCommand } from './promote/command';
+import { pullCommand } from './pull/command';
+
+import { redeployCommand } from './redeploy/command';
+import { removeCommand } from './remove/command';
+import { rollbackCommand } from './rollback/command';
+import { targetCommand } from './target/command';
+import { teamsCommand } from './teams/command';
+import { telemetryCommand } from './telemetry/command';
+import { whoamiCommand } from './whoami/command';
```
TooTallNate:
```suggestion
import { initCommand } from './init/command';
import { inspectCommand } from './inspect/command';
import { installCommand } from './install/command';
import { integrationResourceCommand } from './integration-resource/command';
import { integrationCommand } from './integration/command';
import { linkCommand } from './link/command';
import { listCommand } from './list/command';
import { loginCommand } from './login/command';
import { logoutCommand } from './logout/command';
import { logsCommand } from './logs/command';
import { projectCommand } from './project/command';
import { promoteCommand } from './promote/command';
import { pullCommand } from './pull/command';
import { redeployCommand } from './redeploy/command';
import { removeCommand } from './remove/command';
import { rollbackCommand } from './rollback/command';
import { targetCommand } from './target/command';
import { teamsCommand } from './teams/command';
import { telemetryCommand } from './telemetry/command';
import { whoamiCommand } from './whoami/command';
import type { Command } from './help';
```

packages/cli/src/commands/index.ts
erikareads: I'm not a fan of default exports, but the old map can probably be removed, yes.

TooTallNate: 👍 `export const commands = new Map();`

packages/cli/src/commands/index.ts
```diff
@@ -1,47 +1,78 @@
...
+
+export function getAliases(command: Pick<Command, 'name' | 'aliases'>) {
+  return [command.name].concat(command.aliases);
+}
+
+export const commands = new Map();
+for (const command of commandsStructs) {
+  const aliases = getAliases(command);
+  for (const alias of aliases) {
+    commands.set(alias, command.name);
```
TooTallNate, line 76: I'm just nitpicking at this point, but you could take this one step further:

```suggestion
export const commands = new Map();
for (const command of commandsStructs) {
  commands.set(command.name, command.name);
  for (const alias of command.aliases) {
    commands.set(alias, command.name);
```

TooTallNate (approved):

packages/cli/src/commands/index.ts
erikareads: I'm going to need `getAliases` elsewhere, which is why I exported it.

PR #12473, opened 2024-11-01
[cli] standardize teams subcommand routing

erikareads:


packages/cli/src/commands/teams/index.ts
```diff
@@ -36,10 +42,16 @@ export default async (client: Client) => {
 
   const isSwitch = parsedArgs.args[0] === 'switch';
   parsedArgs.args = parsedArgs.args.slice(1);
+  let subcommand: string | string[] | undefined;
```
TooTallNate, line 45: Not related to this PR specifically, but I keep seeing this `subcommand: string | string[]` floating around. When would the subcommand be an array? It doesn't seem correct to me, I think there's an incorrect typing somewhere in the subcommand parsing stack.

erikareads: Yeah, I agree, we should figure out where that's coming from. I just didn't want to have an implicit any on `subcommand`.

TooTallNate (approved):

PR #12492, opened 2024-11-04
[cli] Change `vercel alias` subcommands to have same interface as `alias`

erikareads:


.changeset/shiny-parents-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": major
```
trek, line 2: Is this a public interface change?

trek (changes requested):

TooTallNate:
What is the motivation with this one? I actually think we should go in the opposite direction - where `argv` is _always_ a second argument to the commands (even at the top-level). Reason being that when doing `args(..., { permissive: true })`, the leftover unconsumed args are then what gets passed in to the next level subcommand. In this way, `argv` is entirely decoupled from the `Client` class, and writing tests would also become easier (no more `client.setArgv()` - instead pass in the args into the command function).

.changeset/shiny-parents-impress.md
erikareads: On reflection, it isn't to the same degree as some @TooTallNate's previous changes, I'll move it down to patch if we proceed with this.

erikareads:
> What is the motivation with this one? I actually think we should go in the opposite direction - where `argv` is _always_ a second argument to the commands (even at the top-level). Reason being that when doing `args(..., { permissive: true })`, the leftover unconsumed args are then what gets passed in to the next level subcommand. In this way, `argv` is entirely decoupled from the `Client` class, and writing tests would also become easier (no more `client.setArgv()` - instead pass in the args into the command function).

I was doing it in the interest of standardizing across commands and subcommands. 

In my ideal world, commands and subcommands wouldn't need to know about args at all, they would already receive an object with the parsed options and arguments. Those options and arguments would be derived from their `Command` struct.

TooTallNate:
> In my ideal world, commands and subcommands wouldn't need to know about args at all, they would already receive an object with the parsed options and arguments. Those options and arguments would be derived from their `Command` struct.

I agree with this, but for it to work in a properly typed way like we have now, I don't see a way around each (sub)command having to import their command struct in order to get the proper types. I can see a couple ways we could go about this:

### Higher order function

```typescript
export default withArgs(listSubcommand, async (client, { args, flags }) => {
  // ...
});
```

This approach might be nice if we consider other higher order functions that would be useful. I imagine a `withLinkedProject()` would be useful.

### Simple function

```typescript
export default async (client: Client, argv: string[]) => {
  const { args, flags } = parseArgs(listSubcommand, argv);
  // ...
});
```

Similar to what we have today, but reduce the 6 or so (+ the try/catch) lines we have today into a single function call. It would throw upon invalid/unknown parameters so that the command implementation doesn't need to handle that case.

PR #12494, opened 2024-11-04
Fix MockClient typescript errors

jeffsee55:
This just stubs out all of the unimplemented methods from `WriteStream` to make Typescript happy. For some, we actually needed to invoke the `Passthrough` class that we're extending but I didn't do that unless a test failed as a result of not having it.


trek (approved):

EndangeredMassa (approved):
I've wanted this resolved for a long time. Thanks!

onsclom (approved):

erikareads (approved):

PR #12496, opened 2024-11-04
[cli] Migrate `vercel teams` subcommands to their own command specifications

TooTallNate:


packages/cli/src/commands/teams/command.ts
```diff
@@ -80,9 +68,15 @@ export const teamsCommand = {
...
-    },
   ],
 } as const;
+
+export const teamsCommand = {
+  name: 'teams',
+  aliases: ['switch', 'team'],
+  description: 'Manage Teams under your Vercel account',
+  arguments: [],
+  subcommands: [addSubcommand, listSubcommand, switchSubcommand],
```
erikareads: The `inviteSubcommand` seems to be missing.

packages/cli/src/util/telemetry/commands/teams/invite.ts
```diff
@@ -1,11 +1,16 @@
...
+export class TeamsInviteTelemetryClient
+  extends TelemetryClient
+  implements TelemetryMethods<typeof inviteSubcommand>
+{
+  trackCliArgumentEmails(value: string | undefined) {
+    if (value) {
       this.trackCliArgument({
         arg: 'email',
-        value: count === 1 ? 'ONE' : 'MANY',
+        value,
```
erikareads: I understand that you changed this to conform to the method signature generated by `implements TelemetryMethods<typeof inviteSubcommand>`, but I don't like the redaction isn't inside the track method anymore, and therefore can be missed during a future code change.

erikareads (changes requested):

packages/cli/src/util/telemetry/commands/teams/invite.ts
TooTallNate: We could add validation inside this function that the input is one of "ONE" or "MANY".

Unfortunately we don't currently have a schema property to indicate that an argument may be represented by multiple arguments (something like `vararg: true`), but I would be happy to revisit that. I believe there's at least one more case when that would be relevant.

packages/cli/src/commands/teams/command.ts
TooTallNate: Nice catch (878419bf17c6f7d521f852a5776d61060cf874fa)!

packages/cli/src/util/telemetry/commands/teams/invite.ts
TooTallNate: 0de3c4dec52328e35349b86815f59d1f4f4ae25b

erikareads (approved):

PR #12500, opened 2024-11-05
[tests] migrate `vc deploy` name related test permutations to unit tests

trek:
Migrate two tests that verify we default a project name in terminal output.

These tests currently take a _significant_ amount of time because they create projects and deployments as a side effect of verifying this output. Moving them to unit tests with mocks should represent a nice time savings.

CleanShot 2024-11-05 at 08 49 44@2x


packages/cli/test/unit/commands/deploy/index.test.ts
```diff
@@ -784,7 +783,7 @@ describe('deploy', () => {
       archive: undefined,
     };
 
-    it('--force', async () => {
+    it.todo('--force', async () => {
```
EndangeredMassa: **suggestion (soft-blocking):** unskip?

EndangeredMassa (dismissed):

packages/cli/test/unit/commands/deploy/index.test.ts
trek: Yep! I was using this to compare some other behavior

EndangeredMassa (approved):

PR #12501, opened 2024-11-05
change fs-detectors.getApiMatches pattren,to keep my project structure like this vercel/api

lazyfury:
I want to keep my project structure like this vercel/api
I added vercel support to my golang gin project, instead of creating serverless functions for each interface, so I want to keep the api directory


EndangeredMassa:
Hello @LazyFury!

I understand that you'd prefer a different structure for using the Vercel Functions (`api`) directory, but we don't currently support putting this anywhere else. Allowing other directories to operate this way would require changes elsewhere in the system, which we're not prepared to do right now.

Sorry for the trouble!

PR #12515, opened 2024-11-05
[cli] Support `multiple: true` for variadic arguments in Command spec

TooTallNate:
Adds a property to indicate that the positional argument accepts multiple values, which is leveraged to add an indicator to the `--help` output as well as change the telemetry method signature to use `string[]` instead of `string | undefined`.

Applied for `vercel remove` command, but this can be used in a few other places as well, which will be done in follow-up PRs.


packages/cli/src/commands/help.ts
```diff
@@ -124,7 +125,11 @@ export function buildCommandSynopsisLine(command: Command, parent?: Command) {
 
   if (args.length > 0) {
     for (const argument of args) {
-      line.push(argument.required ? argument.name : `[${argument.name}]`);
+      let { name } = argument;
+      if (argument.multiple) {
+        name += '...';
```
erikareads: Can we add a space between the argument name and the `...`? 

Adding the space would be consistent with how unix shows multiple args in `man` pages:
```
rm [-f | -i] [-dIRrvWx] file ...
```

erikareads (changes requested):

erikareads (approved):

trek (approved):

PR #12523, opened 2024-11-06
[Custom Environments] Add new flag `--target` to the `redeploy` command

paulogdm:
# Pull Request

In this PR, we are adding a new command option to the `redeploy` command. With the `--target` flag, the CLI should be capable of:

- Redeploying a deployment to another environment. For example, a `production` deployment can now be deployed to `preview`.
- Redeploying a deployment to custom environments. If a deployment exists, we should be able to redeploy it in any other environment.


packages/cli/src/commands/redeploy/command.ts
```diff
@@ -18,6 +18,13 @@ export const redeployCommand = {
       type: Boolean,
       deprecated: false,
     },
+    {
+      name: 'environment',
```
trek: Within the deploy/build parts of the CLI this is referred to as `target`. Not sure _why_. But this should probably match here so 

```
vercel deploy --target=qa
vercel redeploy dpl_123 --target=qa
```

have parity.

trek (changes requested):

packages/cli/src/commands/redeploy/command.ts
paulogdm: Thanks! Changed. But 🤢

packages/cli/src/commands/redeploy/index.ts
```diff
@@ -18,6 +18,11 @@ import { redeployCommand } from './command';
 import { getFlagsSpecification } from '../../util/get-flags-specification';
 import output from '../../output-manager';
 import { RedeployTelemetryClient } from '../../util/telemetry/commands/redeploy';
+import { CustomEnvironment } from '@vercel-internals/types';
```
erikareads:
```suggestion
import type { CustomEnvironment } from '@vercel-internals/types';
```

erikareads (dismissed):

erikareads (dismissed):

.changeset/short-bananas-share.md
```diff
@@ -0,0 +1,6 @@
+---
+'vercel': minor
+'@vercel/error-utils': patch
```
trek: I don't see any changes to @vercel/error-utils. I don't think we need to bump version here.

trek (changes requested):

.changeset/short-bananas-share.md
paulogdm: Yeah, not sure why this is there. Weird.

paulogdm:
```suggestion
```

erikareads (approved):

trek (approved):

PR #12525, opened 2024-11-06
[tests] Type-check CLI tests

onsclom:
The CLI tests were not included in `pnpm type-check`. That results in many accumulated type errors, which results in noisy red squiggly lines and incorrect autocomplete.

This PR enables type-checking on CLI tests and fixes existing TypeScript errors.


packages/cli/test/dev/utils.ts
```diff
@@ -347,7 +347,9 @@ export async function testFixture(
   };
 
   return {
-    dev,
+    dev: dev as any as Omit<typeof dev, 'kill'> & {
+      kill: () => Promise<{ stdout: string; stderr: string }>;
+    },
```
onsclom, line 352: Also tried:
```ts
return { 
	dev: {
	  ...dev,
	  kill: () => { 
	    // implementation here 
	  }
	}
```
But, copying dev like this breaks behavior.

Don't love this solution, but `integration-1.test.ts` needs to know the new correct type of `dev.kill`.

One other idea: `testFixture` could return `kill`, `stdout`, and `stderr` as fields directly, and `integration-1.test.ts` could use those instead.

packages/cli/tsconfig.json
```diff
@@ -9,8 +9,8 @@
     "typeRoots": ["./types", "./node_modules/@types"]
   },
   "extends": "../../tsconfig.base.json",
-  "include": ["./types", "src/**/*"],
-  "exclude": ["templates/*"],
+  "include": ["./types", "src/**/*", "test/**/*"],
+  "exclude": ["templates/*", "**/fixtures/**"],
```
onsclom, line 13: This is the one change that includes tests in `pnpm type-check`. Everything else is fixing existing type errors.

packages/cli/test/mocks/matchers/index.ts
```diff
@@ -11,4 +11,5 @@ declare module 'vitest' {
       ToHaveTelemetryEventsMatchers<T> {}
 }
 
+// @ts-expect-error
 expect.extend({ toOutput, toHaveTelemetryEvents });
```
onsclom: Couldn't find a great TS solution here, but I think`@ts-expect-error` is fine here because of the
```ts
declare module 'vitest' {
  interface Assertion<T = any>
    extends ToOutputMatchers<T>,
      ToHaveTelemetryEventsMatchers<T> {}
}
```
above.

packages/cli/test/unit/util/projects/get-project-by-deployment.test.ts
```diff
@@ -30,7 +31,9 @@ describe('getProjectByDeployment', () => {
   });
 
   it('should get project and deployment associated to a team', async () => {
-    const [team] = useTeams('team_dummy');
+    const teams = useTeams('team_dummy');
```
TooTallNate: Should this be using the singular `useTeam()` instead?

packages/cli/package.json
```diff
@@ -166,7 +167,7 @@
     "tar-fs": "1.16.3",
     "title": "3.4.1",
     "tldts": "6.1.47",
-    "tmp-promise": "1.0.3",
```
EndangeredMassa: **question (non-blocking):** was there no interface or behavior change between these versions?

I couldn't find a changelog for these releases. Are we sure this change is OK?‏

.changeset/breezy-pots-sort.md
```diff
@@ -0,0 +1,4 @@
+---
+---
```
EndangeredMassa, line 3: **suggestion (blocking):** ‏add patch for `cli` package

This PR is mostly about tests, but it does change logic that could affect behavior. We should give it a proper version.

packages/cli/test/dev/utils.ts
EndangeredMassa: This is clunky, but seems to be accurate. I'm good with this for this PR. We can revisit elsewhere.

packages/cli/test/mocks/deployment.ts
```diff
@@ -211,7 +211,10 @@ function setupDeploymentEndpoints(): void {
   });
 
   client.scenario.get('/:version/deployments/:id/aliases', (req, res) => {
-    const limit = parseInt(req.query.limit);
+    const limit =
+      typeof req.query.limit === 'string'
+        ? parseInt(req.query.limit, 10)
```
EndangeredMassa: **suggestion (non-blocking):** prefer `Number` or `Number.parseInt` over `parseInt`‏

We could lint for this.

packages/cli/test/mocks/matchers/index.ts
EndangeredMassa: If we leave in a `@ts-expect-error`, let's also leave a comment about why

packages/cli/test/unit/commands/project/list.test.ts
```diff
@@ -129,7 +129,7 @@ describe('list', () => {
       ...defaultProject,
       targets: {
         production: {
-          ...defaultProject!.targets!.production,
+          ...defaultProject!.targets!.production!,
```
EndangeredMassa: **suggestion (non-blocking):** remove non-null assertion

Can we explicitly check this value and throw a clear message if it's not defined?‏

EndangeredMassa (changes requested):

packages/cli/test/mocks/matchers/index.ts
onsclom: Figured this out so no more `@ts-expect-error` is needed! `@vitest/expect`'s version was out of date with `vitest`, so it provided incompatible types.

packages/cli/package.json
onsclom: Went back to `1.0.3` and changed the `tmp-promise` `@ts-ignore`s to `@ts-expect-error` to be safe.

For the future, it looks like we can remove `tmp-promise` and use `tmp` directly. We already only use the sync functions from `tmp` directly, and [tmp has a changelog](https://github.com/raszi/node-tmp/blob/HEAD/CHANGELOG.md).

packages/cli/test/dev/utils.ts
```diff
@@ -332,7 +332,7 @@ export async function testFixture(
     readyResolver.resolve(null);
   });
 
-  // @ts-ignore
+  // @ts-expect-error
```
EndangeredMassa, line 335: **suggestion (soft-blocking):** add a reason why this would error

EndangeredMassa (approved):

PR #12530, opened 2024-11-06
Use `vitest` in `@vercel/client`

jeffsee55:


packages/client/package.json
```diff
@@ -15,9 +15,10 @@
   },
   "scripts": {
     "build": "node ../../utils/build.mjs",
-    "test-e2e": "pnpm test tests/create-deployment.test.ts tests/create-legacy-deployment.test.ts tests/paths.test.ts",
```
jeffsee55, line 18: `create-legacy-deployment.test.ts` no longer exists

packages/client/tests/unit.vercelignore.test.ts
```diff
@@ -16,7 +17,9 @@ describe('Test `getVercelIgnore()`', () => {
     try {
       await getVercelIgnore(fixture);
     } catch (_err) {
-      err = _err;
+      if (_err instanceof Error) {
+        err = _err;
+      }
```
EndangeredMassa: **suggestion (non-blocking):** ‏check else case

If for some reason `_err` is not an `Error`, we should make sure the error shows what it was. Right now, if this test fails for this reason, the failure we'll get is that `err` is `null`, which isn't not super helpful.

I suggest:

```suggestion
      if (!(_err instanceof Error)) {
        throw new Error(`Error "${_err}" not instanceof Error`);
      } 
      
      err = _err;
```

EndangeredMassa (dismissed):

trek (approved):

packages/client/package.json
```diff
@@ -34,6 +35,7 @@
     "@types/recursive-readdir": "2.2.0",
     "@types/tar-fs": "1.16.1",
     "jest-junit": "16.0.0",
+    "vitest": "2.0.1",
```
TooTallNate, line 38: Is this actually necessary? `vitest` is a dep in the root-level "package.json".

Also, looks like we're using a newer version at the root level `2.0.3`.

TooTallNate (changes requested):

PR #12532, opened 2024-11-06
[cli] Fix `vc certs issue` subdomain parsing

onsclom:
I added a regression in https://github.com/vercel/vercel/commit/fd86f5d2c9fef8465a1514d6184c23c3bc1578dd#diff-4463dfbbb283602d090922c89d530aa8efb60baa260ffddcc33b275da6a3b00f where domains without a subdomain would incorrectly throw an error.

This fixes that and adds a test that catches this case.


packages/cli/src/commands/certs/issue.ts
```diff
@@ -199,7 +199,7 @@ async function runStartOrder(
   const [header, ...rows] = dnsTable(
     pendingChallenges.map(challenge => {
       const subdomain = getSubdomain(challenge.domain);
-      if (!subdomain) {
+      if (subdomain === null) {
```
jeffsee55, line 202: What is the value here that led to the regression? `false` or `undefined`?

onsclom: Subdomain would be an empty string, and `!""` is `true`

onsclom: `getSubdomain` returns `null` for invalid domains and and empty string if the domain has no subdomain.

jeffsee55 (dismissed):

packages/cli/test/unit/commands/certs/issue.test.ts
```diff
@@ -30,6 +31,24 @@ describe('certs issue', () => {
     await expect(exitCodePromise).resolves.toEqual(1);
   });
 
+  it('should issue cert', async () => {
+    useUser();
+    client.scenario.post('v3/now/certs', (_, res) => {
```
TooTallNate: Let's drop the `/now` part of the URL while we're in here? Will need to update where the fetch call is made as well.

TooTallNate (changes requested):

packages/cli/test/unit/commands/certs/issue.test.ts
```diff
@@ -30,6 +31,24 @@ describe('certs issue', () => {
     await expect(exitCodePromise).resolves.toEqual(1);
   });
 
+  it('should issue cert', async () => {
+    useUser();
+    client.scenario.post('v3/certs', (_, res) => {
```
TooTallNate:
```suggestion
    client.scenario.post('/v3/certs', (_, res) => {
```

packages/cli/test/unit/commands/certs/issue.test.ts
```diff
@@ -58,6 +77,26 @@ describe('certs issue', () => {
         },
       ]);
     });
+
+    it('should handle challenges', async () => {
+      useUser();
+      client.scenario.post('v3/certs', (_, res) => {
```
TooTallNate:
```suggestion
      client.scenario.post('/v3/certs', (_, res) => {
```

TooTallNate (changes requested):

TooTallNate (approved):

PR #12533, opened 2024-11-07
Move `@vercel/edge` to `vitest`

jeffsee55:
`vitest` supports `edge-runtime` as an environment, using `@edge-runtime/vm` if provided

https://github.com/vitest-dev/vitest/pull/1574

### DIFF
--- a/packages/edge/test/edge-headers.test.ts
+++ b/packages/edge/test/edge-headers.test.ts
@@ -1,7 +1,8 @@
 /**
- * @jest-environment @edge-runtime/jest-environment
+ * @vitest-environment edge-runtime
  */
 
+import { describe, expect, test } from 'vitest';
 import {
   CITY_HEADER_NAME,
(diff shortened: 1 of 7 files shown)


packages/edge/test/edge-headers.test.ts
```diff
@@ -15,6 +16,11 @@ import {
   REQUEST_ID_HEADER_NAME,
 } from '../src';
 
+test('vitest environment is correctly added', () => {
+  // @ts-ignore
+  expect(globalThis.EdgeRuntime).toBeDefined();
```
jeffsee55, line 21: I wasn't really sure how to validate that `@vitest-environment edge-runtime` actually did anything, this works but might be nice to have a more meaningful test.

Having said that, the other tests all pass without the `@vitest-environment edge-runtime` header so we may want to test something more specific.

trek (approved):

PR #12542, opened 2024-11-07
[cli] redact Extensions

erikareads:


TooTallNate:
If it's possible, I'm -1 on this. I think it would be interesting to track the value, considering we would only be sending the value _after_ the successful invocation of the extension (i.e. it was definitely invoked as a CLI extension, and not a random typo / non-existing command).

trek (dismissed):

erikareads:
> If it's possible, I'm -1 on this. I think it would be interesting to track the value, considering we would only be sending the value _after_ the successful invocation of the extension (i.e. it was definitely invoked as a CLI extension, and not a random typo / non-existing command).

@trek and I have been discussing this, I would love to know this, but it fails our "we have a known list to compare against" standard for redaction.

trek:
@TooTallNate we worried that if someone wrote their own extension like "send-microsoft-data" or what have you, we now know something privilege.

TooTallNate:
That's what I meant by "if possible". Ok, well, bummer, oh well.

TooTallNate (approved):

PR #12544, opened 2024-11-07
[cli] create standard method for redacted targets

erikareads:
This PR creates a standard method so that all targets get redacted in the same way.

### DIFF
--- a/packages/cli/src/util/telemetry/index.ts
+++ b/packages/cli/src/util/telemetry/index.ts
@@ -37,6 +37,12 @@ export class TelemetryClient {
     }
     return 'NONE';
   };
+  protected redactedTargetName = (target: string) => {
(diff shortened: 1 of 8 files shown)


packages/cli/src/util/telemetry/index.ts
```diff
@@ -37,6 +37,12 @@ export class TelemetryClient {
     }
     return 'NONE';
   };
+  protected redactedTargetName = (target: string) => {
+    if (['production', 'preview', 'development'].includes(target)) {
```
TooTallNate: Might as well import these values from `import { PROJECT_ENV_TARGET } from '@vercel-internals/constants';`

TooTallNate (dismissed):

trek (approved):

PR #12545, opened 2024-11-07
[cli] redact output mode in telemetry method

erikareads:
This PR moves the logic for checking whether output mode is redacted into the method, making it harder to forget to redact unknown output modes by mistake.

### DIFF
--- a/.changeset/warm-carrots-film.md
+++ b/.changeset/warm-carrots-film.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] redact output mode in telemetry method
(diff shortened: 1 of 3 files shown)


TooTallNate (commented):
There doesn't seem to be any change in logic here, unless I'm missing something?

erikareads:
> There doesn't seem to be any change in logic here, unless I'm missing something?

I updated the PR description. Where possible, I want the logic for what's allowed or not to live inside the telemetry methods, not in the call site.

trek (approved):

PR #12549, opened 2024-11-07
[cli] use subcommandOriginal in telemetry

erikareads:
for subcommands, we want to know exactly what the user typed to successfully invoke that subcommand, not just the name of the subcommand.

### DIFF
--- a/packages/cli/src/commands/env/index.ts
+++ b/packages/cli/src/commands/env/index.ts
@@ -46,7 +46,10 @@ export default async function main(client: Client) {
   }
 
   const subArgs = parsedArgs.args.slice(1);
-  const { subcommand, args } = getSubcommand(subArgs, COMMAND_CONFIG);
+  const { subcommand, args, subcommandOriginal } = getSubcommand(
+    subArgs,
+    COMMAND_CONFIG
+  );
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/env/index.ts
```diff
@@ -69,31 +72,31 @@ export default async function main(client: Client) {
         printHelp(listSubcommand);
         return 2;
       }
-      telemetry.trackCliSubcommandList(subcommand);
+      telemetry.trackCliSubcommandList(subcommandOriginal);
       return ls(client, args);
     case 'add':
       if (needHelp) {
         telemetry.trackCliFlagHelp('env', 'add');
```
TooTallNate, line 79: I think we should audit and do in these help functions as well:

```suggestion
        telemetry.trackCliFlagHelp('env', subcommandOriginal);
```

TooTallNate (dismissed):

packages/cli/src/commands/env/index.ts
erikareads: I want to do these on a separate PR.

erikareads: https://github.com/vercel/vercel/pull/12552

trek (approved):

PR #12551, opened 2024-11-07
Ensure firewall tests run

jeffsee55:
Since we run these tests in chunks, we need to use the `vitest-` scripts so we know to include them in the "Find changes" action. From what I can see we haven't actually been testing this package until now.

Also this package isn't designed to run on Node <20 so updated `engines` and created a new entry in the `chunk-tests` workflow.

https://github.com/vercel/vercel/actions/runs/11730172571/job/32677564246?pr=12551

### DIFF
--- a/.changeset/chilled-ladybugs-visit.md
+++ b/.changeset/chilled-ladybugs-visit.md
@@ -1,0 +1,5 @@
+---
+'@vercel/firewall': patch
+---
+
+Ensure firewall tests run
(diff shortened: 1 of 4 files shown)


EndangeredMassa (approved):

onsclom (approved):

jeffsee55:
cc @cramforce & @sueplex since you're the codeowners for this one

sueplex (approved):

PR #12556, opened 2024-11-08
[tests] Default to disabled telemetry in tests

trek:
`VERCEL_TELEMETRY_DISABLED` is used inside the base `TelemetryEventStore` [class's `enabled` property](https://github.com/vercel/vercel/blob/8bf2fbc587de1a832e40c2c97974c2927e090403/packages/cli/src/util/telemetry/index.ts#L207-L213) as part of determining if we want to submit telemetry [on a call to `save()`](2d5a21aecc56e19e130f744d9c268cf35ff1de827416e704d27483a01ffbd94fL227).

When set to any value, it overrides _all other ways of enabling telemetry_. Since telemetry is opt-out, this allows customers to disable the behavior entirely as part of their local or CI environments without first needing to run `vercel telemetry disable`.

Within our own CI we don't want to send telemetry data except in a narrow set of cases. Adding `VERCEL_TELEMETRY_DISABLED: 1` to our test workflows ensures this for us. 

For e2e test where we _do_ want to test telemetry we plan to:

* remove and re-add `VERCEL_TELEMETRY_DISABLED` variable for those specific tests.
* submit data to a local express server for evaluation, not the real telemetry bridge URL.

For unit tests, this variable no effect. `telemetryEventStore.enabled` is checked within `telemetryEventStore.save`. For unit tests, a mock client is passed to CLI functions and the mock client's `telemetryStore` property is an instance of `MockTelemetryEventStore`. The [`save()` method](https://github.com/vercel/vercel/blob/8bf2fbc587de1a832e40c2c97974c2927e090403/packages/cli/test/mocks/client.ts#L140-L142) of this class is a no-op and will never submit data.


trek:
Closing to push to private repo

jeffsee55 (approved):

erikareads (approved):

PR #12557, opened 2024-11-08
[python] add support for streaming

QuiiBz:
- [x] HTTP handler
- [x] WSGI handler
- [x] ASGI handler
- [x] Logs levels
- [x] Run all tests

Add support for HTTP streaming and other important feature for Python Vercel Functions, opt-in via the `VERCEL_FORCE_PYTHON_STREAMING` build environment variable for the duration of the migration.


packages/python/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "build": "node ../../utils/build-builder.mjs",
-    "test": "jest --reporters=default --reporters=jest-junit --env node --verbose --runInBand --bail",
+    "test": "VERCEL_FORCE_STREAMING_RUNTIME=1 jest --reporters=default --reporters=jest-junit --env node --verbose --runInBand --bail",
```
TooTallNate: This is failing on Windows (can't set env vars at the beginning of a command like that).

QuiiBz:
We just have one test failing otherwise, will fix next week: https://github.com/vercel/vercel/actions/runs/11746818781/job/32727545995?pr=12557#step:8:486

TooTallNate (approved):

PR #12559, opened 2024-11-08
Fix basePath root matching for error pages

ijjk:
This ensures we can match the root itself when a `basePath` value is configured on a project and the error page is being matched. 

Verified on https://next-500-vercel-issue-65i0edfn4-vtest314-ijjk-testing.vercel.app/library?q=any

x-ref: [slack thread](https://vercel.slack.com/archives/C0676QZBWKS/p1731090171198859?thread_ts=1730905462.387389&cid=C0676QZBWKS)
x-ref: NEXT-3850

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2367,7 +2367,11 @@ export async function serverBuild({
           ]
         : [
             {
-              src: path.posix.join('/', entryDirectory, '.*'),
+              src: path.posix.join(
+                '/',
+                entryDirectory,
+                `${entryDirectory !== '.' ? '?' : ''}.*`
+              ),
               dest: path.posix.join(
                 '/',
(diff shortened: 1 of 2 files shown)


packages/next/src/server-build.ts
```diff
@@ -2412,7 +2416,11 @@ export async function serverBuild({
           ]
         : [
             {
-              src: path.posix.join('/', entryDirectory, '.*'),
+              src: path.posix.join(
+                '/',
+                entryDirectory,
+                `${entryDirectory !== '.' ? '?' : ''}.*`
```
ztanner, line 2428: fairly tricky logic for anyone unfamiliar with what an `entryDirectory` of `.` represents -- maybe can add a comment above about what this is handling

ztanner (approved):

PR #12566, opened 2024-11-12
Add a hidden `telemetry flush` to use for the telemetry send process

jeffsee55:
The way we were previously sending the telemetry event wasn't compatible with our ability to invoke the CLI from `ts-node` because finding the `send-telemetry` module wasn't in the expected path. Instead, the `sendToSubprocess` function invokes the CLI with the `telemetry flush` command, providing the events as a stringified JSON argument.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -368,7 +368,14 @@ const main = async () => {
     client.argv.push('-h');
   }
 
-  const subcommandsWithoutToken = ['login', 'logout', 'help', 'init', 'build'];
+  const subcommandsWithoutToken = [
+    'login',
+    'logout',
+    'help',
+    'init',
+    'build',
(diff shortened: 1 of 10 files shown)


packages/cli/src/index.ts
```diff
@@ -791,8 +791,10 @@ const main = async () => {
     return 1;
   }
 
-  // specifically don't await this, we want to fire and forget
-  await telemetryEventStore.save();
+  // FIXME: the telemetry flush event is called by `telemetryEventStore.save`, and it reinvokes the `main` function
+  if (parsedArgs.args[3] !== 'flush') {
+    await telemetryEventStore.save();
```
jeffsee55, line 803: This resulted in a loop, open to suggestions on a better approach.

erikareads (approved):

PR #12569, opened 2024-11-13
[cli] Swap out custom CI code for `ci-info` package

trek:
We're currently using some hand-crafted code with a data object copy/pasted from `turbo` in out CI platform detection for telemetry. Someone suggested we use the existing `ci-info` package but there was no way to a get stable identifier for each platform.

You _could_ do:

* `ci.name` but it was specifically noted this could change
* `ci.<SOME_CONST>` but that involved us mapping to ids ourself, which we feared would get out of date

https://github.com/watson/ci-info/commit/81fd993 adds a `ci.id`.

Testing this is a bit wonky because the `ci.id` is set on package import once, so the mocking was complicated. `vi` hoists it's package mocking to the top of a file using static analysis, so can't be defined twice. The workarounds for this all apply to mocked _functions_ but not imported constants.

The only thing I could think of doing was moving each of these cases into their on file 🤷


packages/cli/test/unit/index.test.ts
```diff
@@ -1,14 +1,6 @@
...
-  expect,
-  it,
-  vi,
-  beforeEach,
-  afterEach,
-  beforeAll,
-  afterAll,
-} from 'vitest';
+import { describe, expect, it, vi, beforeAll, afterAll } from 'vitest';
+import 'ci-info';
```
erikareads: Why is this import here?

erikareads (dismissed):

packages/cli/test/unit/index.test.ts
trek: Ah, just leftover from when the test was going to live there. Lint didn't catch it because it could be used for side effect.

erikareads (dismissed):

packages/cli/test/unit/telemetry-detect-ci/not-using-ci.test.ts
```diff
@@ -0,0 +1,42 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import 'ci-info';
```
EndangeredMassa: **suggestion (non-blocking):** add comment about why we do this

It's not clear from the code what the side effect is of this import. Can we add a comment to wherever we do this explaining why?‏

EndangeredMassa (dismissed):

TooTallNate (approved):

PR #12570, opened 2024-11-13
remove lock

ismaelrumzan:
Deleting lock file so it can be regenerated when the GH sdk generate is run to avoid
- https://github.com/vercel/vercel/actions/runs/11821180664/job/32935294463#step:5:2220
- https://vercel.slack.com/archives/C074TFBJJ6L/p1731515293785749?thread_ts=1731451554.269299&cid=C074TFBJJ6L


EndangeredMassa (changes requested):
I don't think we should blow away the lockfile. Generating a new one from scratch will result in many transitive dependency changes, increasing the risk of breaking something. I'd rather we make those kinds of changes with targeted PRs updating specific dependencies, which could update their transitive dependencies.

If we do go this route, I'd like to see the follow-up PR already created that will regenerate the lockfile, targeting this branch. Otherwise, this could stall development on the repo.

My preference is that we find some other way to resolve the issue. It looks like there's an error in the SDK generation. If it's intending to change the dependencies, it should also do a `pnpm install` to update the lockfile.

PR #12555, opened 2024-11-07
Add `telemetry flush` command to send telemetry events

erikareads:
Up until now we've only been collecting telemetry events in memory and not actually sending them to the telemetry server. This adds the actual send logic, which happens in a subprocess so that the main process can operate without interruption.

PR #12576, opened 2024-11-13
Pass `VERCEL_TELEMETRY_DISABLED` to flush command

jeffsee55:
Now calls `telemetry flush with `VERCEL_TELEMETRY_DISABLED=1` to avoid an infinite loop on `telemetryEventStore.save()`

[See comment](https://github.com/vercel/vercel/pull/12555#pullrequestreview-2434378618)


packages/cli/test/unit/util/telemetry-check-status.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        config: {
+          telemetry: {
+            enabled: true,
+          },
+        },
+      });
+    });
+
+    it('does not inform the user', async () => {
+      expect(client.getFullOutput()).toEqual('');
```
jeffsee55, line 57: cc @trek I changed these tests to avoid using `.not.toOutput` (per Nate's [comment](https://github.com/vercel/vercel/pull/12555#pullrequestreview-2434385412))

PR #12585, opened 2024-11-14
[cli] fix issues vercel rollback

erikareads:
- **[cli] const over let**
- **[cli] implicit any**
- **[cli] remove any type**
- **add changeset**

### DIFF
--- a/.changeset/ten-shirts-begin.md
+++ b/.changeset/ten-shirts-begin.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/ten-shirts-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': patch
```
TooTallNate:
```suggestion
```

This is a no-op change, I don't think it warrants a release.

trek (dismissed):

TooTallNate (approved):

PR #12588, opened 2024-11-14
[cli] fix minor issues vercel remove

erikareads:
- **[cli] type only import**
- **[cli] remove any type**
- **[cli] remove string concatentation**
- **add changeset**


packages/cli/src/commands/remove/index.ts
```diff
@@ -211,7 +210,7 @@ export default async function remove(client: Client) {
   });
   const start = Date.now();
 
-  await Promise.all<any>([
+  await Promise.all<boolean | ProjectNotFound | undefined>([
```
TooTallNate: Does this even need to be typed? The return value is not being used

erikareads: no, I removed it.

trek (approved):

PR #12592, opened 2024-11-14
[cli] fix minor issues vercel redeploy

erikareads:
- **[cli] fix type and string literal**
- **[cli] type import**
- **[cli] prefer interpolation to string concatenation**
- **add changeset**


EndangeredMassa (approved):
Is this in service to enforcing some linting rules?

PR #12593, opened 2024-11-14
[cli] fix minor issues vercel promote

erikareads:
- **[cli] prefer const to let**
- **[cli] remove unnecessary else**
- **[cli] add type to avoid implicit any**
- **[cli] remove literal any**
- **add changeset**


packages/cli/src/commands/promote/status.ts
```diff
@@ -195,7 +195,7 @@ async function renderJobFailed({
 
   // aliases are paginated, so continuously loop until all of them have been
   // fetched
-  let nextTimestamp;
+  let nextTimestamp: number | undefined;
```
onsclom, line 198: Same as https://github.com/vercel/vercel/pull/12595#discussion_r1844295413

onsclom (approved):

PR #12595, opened 2024-11-14
[cli] fix minor issues vercel list

erikareads:
- **[cli] type only import**
- **[cli] add type to avoid implicit any**
- **[cli] prefer const over let**
- **[cli] prefer template over string concatenation**
- **add changeset**

### DIFF
--- a/packages/cli/src/commands/list/index.ts
+++ b/packages/cli/src/commands/list/index.ts
@@ -101,7 +101,7 @@ export default async function list(client: Client) {
   let pagination: PaginationOptions | undefined;
   let contextName = '';
   let app: string | undefined = parsedArgs.args[1];
-  let deployments: Deployment[] = [];
+  const deployments: Deployment[] = [];
   let singleDeployment = false;
 
   if (app) {
--- a/packages/cli/src/commands/list/index.ts
+++ b/packages/cli/src/commands/list/index.ts
@@ -216,13 +216,13 @@ export default async function list(client: Client) {
 
     // we don't output the table headers if we have no deployments
     if (!deployments.length) {
-      log(`No deployments found.`);
+      log('No deployments found.');
       return 0;
     }
 
     log(
       `${
-        target === 'production' ? `Production deployments` : `Deployments`
+        target === 'production' ? 'Production deployments' : 'Deployments'
       } for ${projectSlugLink} ${elapsed(Date.now() - start)}`
     );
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/list/index.ts
```diff
@@ -94,10 +98,10 @@ export default async function list(client: Client) {
   });
 
   let project: Project;
-  let pagination;
+  let pagination: PaginationOptions | undefined;
```
onsclom, line 101: Is `let variableName;` a pattern we actually want to avoid?

While typescript says the value of pagination is `any` at declaration, the inference is really good wherever it is used. For example, if you try to use pagination before it has a chance of being set, typescript knows it can only be `undefined`. If we type it `PaginationOptions | undefined`, we lose that strictness. Also, the tsconfig rule `noImplicitAny` doesn't actually complain about this case.

onsclom (approved):

packages/cli/src/commands/list/index.ts
onsclom: Not a blocking comment, just curious about what to prefer for future changes.

erikareads: Yes, I would like to avoid this construction where possible, and instantiating the type where it's not. The problem with inference is that it will automatically follow what we add to the variable, and the implicit `any` means that we won't necessarily get an error from a mistaken type change, unless it causes another issue in the later inferred code.

erikareads: I'm not certain that I fully understand type inference in Typescript, but:

```typescript
let pagination: PaginationOptions | undefined = undefined;
```

should both prevent the implicit any, and allow for proper inference.

onsclom: Good point. Also, just found an example that makes me really dislike the `let varName;` pattern:

https://www.typescriptlang.org/play/?ssl=10&ssc=39&pln=1&pc=1#code/DYUwLgBAhg3AUHKEC8EAs84DMCuA7AYzAEsB7PCLU0gCgEoIBvOCaFCAIgCMoAnD+AF9s1eplgQA9JLbEAzhDABPAA4gIeHAFsuIXgBoNpSHLC9ieAOYQAhHaA

Seems like types can become totally incorrect with closure mutation! I'm all for avoiding `let variableName;` now

PR #12599, opened 2024-11-15
[next] Fix edge cases with internal redirect sorting

ijjk:
This ensures we filter internal redirects more safely in versions that have the `internal` field in the `routes-manifest`. Our previous detection for trailing slash redirects could hit edge cases where it would match user redirects as well causing unexpected sorting/routing behavior. 

x-ref: NEXT-3840
x-ref: NDX-454


packages/next/test/integration/server-build/next.config.js
```diff
@@ -3,4 +3,20 @@ module.exports = (phase, { defaultConfig }) => ({
   generateBuildId() {
     return 'testing-build-id';
   },
+
+  async redirects() {
+    // these routes are for testing trailing slash ordering
```
ztanner, line 8: is there a corresponding test assertion to go with this config?

ijjk: the Integration-1.test.js one

ztanner (approved):

packages/next/test/integration/server-build/next.config.js
ztanner: my bad I somehow didn’t see that file on first look

vikrantgupta25:
redirects for URLs without trailing slashes started failing because of this ? @ijjk

ijjk:
@vikrantgupta25 Do you have trailingSlash: true configured?

vikrantgupta25:
> Do you have trailingSlash: true configured?

yes we do have that as true, but the main issue being it used to work before and it stopped working post this release.

PR #12598, opened 2024-11-14
[cli] Fix framework detection for `vc deploy` when using `--archive` and/or `--prebuilt`

TooTallNate:
Leverage the `ensureLink()` function in `vc deploy`. This ends up being a multi-win:

* Since `ensureLink` doesn't know/care about `--archive`/`--prebuilt`, the source code is uploaded for server-side framework detection (the _fix_ aspect of this PR).
* Quite a bit of duplicated code gets removed, and the logic is cleaner (previously, the linking logic was intertwined with the deployment logic itself, making it hard to follow).
* This is a prefactor for enabling local framework detection, which I plan to do as a follow-up PR.

PR #12608, opened 2024-11-18
[cli] Fix file uploading in `vc deploy` when `sourceFilesOutsideRootDirectory` setting is `false`

TooTallNate:
Fixes a regression introduced in https://github.com/vercel/vercel/pull/12598, where the incorrect set of file paths was being used for a deployment when the `sourceFilesOutsideRootDirectory` project setting is disabled.

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -310,18 +310,11 @@ export default async (client: Client): Promise<number> => {
   const contextName = org.slug;
   client.config.currentTeam = org.type === 'team' ? org.id : undefined;
 
-  // if we have `sourceFilesOutsideRootDirectory` set to `true`, we use the current path
-  // and upload the entire directory.
-  const sourcePath =
-    rootDirectory && !sourceFilesOutsideRootDirectory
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/deploy/index.ts
```diff
@@ -310,18 +310,11 @@ export default async (client: Client): Promise<number> => {
   const contextName = org.slug;
   client.config.currentTeam = org.type === 'team' ? org.id : undefined;
 
-  // if we have `sourceFilesOutsideRootDirectory` set to `true`, we use the current path
-  // and upload the entire directory.
-  const sourcePath =
-    rootDirectory && !sourceFilesOutsideRootDirectory
```
TooTallNate, line 316: Based on the previous code (before #12598), `sourceFilesOutsideRootDirectory` was essentially _always_ set to `true`, and thus `cwd` was being used as the `sourcePath` always.

So the fix in this PR is to pass `cwd` where `sourcePath` was being used before.

jeffsee55 (approved):

erikareads (approved):

PR #12609, opened 2024-11-19
[cli] Run framework detection locally when setting up a new Project

TooTallNate:
Refactors the `setupAndLink()` function to invoke the framework detection locally, rather than uploading the source files and then making an API request to the create deployment endpoint. This is faster, simpler, easier to debug. A net 100 LOC reduction.

### DIFF
--- a/packages/cli/src/util/link/setup-and-link.ts
+++ b/packages/cli/src/util/link/setup-and-link.ts
@@ -23,12 +23,10 @@ import {
   editProjectSettings,
   type PartialProjectSettings,
 } from '../input/edit-project-settings';
-import stamp from '../output/stamp';
 import { EmojiLabel } from '../emoji';
-import createDeploy from '../deploy/create-deploy';
-import Now, { CreateOptions } from '../index';
 import { isAPIError } from '../errors-ts';
 import output from '../../output-manager';
+import { detectProjects } from '../projects/detect-projects';
 import { getPrettyError } from '@vercel/build-utils';
 import { SchemaValidationFailed } from '../errors';
 import { fileNameSymbol } from '@vercel/client';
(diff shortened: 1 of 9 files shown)


trek (dismissed):

packages/cli/test/unit/commands/deploy/index.test.ts
```diff
@@ -1163,17 +1160,6 @@ describe('deploy', () => {
 
         const createdDeploymentId = 'dpl_1';
         client.scenario.post(`/v13/deployments`, (req, res) => {
-          if (missingProjectSettings) {
-            res.status(400).json({
-              error: {
-                code: 'missing_project_settings',
-                framework: null,
```
TooTallNate, line 1170: Turns out, `framework: null` is actually an incorrect payload compared to what the API would actually send (it should be the full "Other" framework contents).

This bad payload causes the confirmation step to be skipped in the linking flow. That payload is fixed in this PR, hence the two added `await expect` calls below.

packages/cli/test/mocks/deployment.ts
```diff
@@ -78,43 +78,6 @@ export function useDeployment({
   return deployment;
 }
 
-export function useDeploymentMissingProjectSettings() {
```
TooTallNate, line 81: This wasn't being used anyways, but seems appropriate to remove here.

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -264,14 +218,6 @@ export default async function setupAndLink(
 
     return { status: 'linked', org, project };
   } catch (err) {
-    if (err instanceof SchemaValidationFailed) {
```
TooTallNate, line 267: Since schema validation happens in the create deployment endpoint, this error check is now dead code. The same error check happens in `vc deploy` itself, however.

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -51,23 +48,20 @@ export default async function setupAndLink(
...
     output.error(`Expected directory but found file: ${path}`);
     return { status: 'error', exitCode: 1, reason: 'PATH_IS_FILE' };
   }
   if (!link) {
     link = await getLinkedProject(client, path);
   }
   const isTTY = client.stdin.isTTY;
-  const quiet = !isTTY;
   let rootDirectory: string | null = null;
-  let sourceFilesOutsideRootDirectory = true;
```
TooTallNate, line 70: Always default to `true` for new Projects at the API level. No need to duplicate that default on the client-side.

jeffsee55 (dismissed):

onsclom (dismissed):

EndangeredMassa (dismissed):

trek (approved):

onsclom (approved):

erikareads (approved):

PR #12616, opened 2024-11-20
[cli] Gracefully report error in `vercel link --repo` when a new Project fails to be created

TooTallNate:
Fixes a bug in the `vc link --repo` command where an API error would be ignored, causing the command to fail further down in the logic. Now the API error is properly propagated and logged, and the command exits with a proper exit code.

Includes a unit test of the of this failure scenario, and an additional bonus unit test for the success case of `--repo` creating a new Project, since that was missing before.

### DIFF
--- a/packages/cli/src/util/git/connect-git-provider.ts
+++ b/packages/cli/src/util/git/connect-git-provider.ts
@@ -30,7 +30,6 @@ export async function disconnectGitProvider(
 
 export async function connectGitProvider(
   client: Client,
-  org: Org,
   projectId: string,
   type: string,
   repo: string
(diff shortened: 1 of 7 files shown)


packages/cli/src/util/git/connect-git-provider.ts
```diff
@@ -30,7 +30,6 @@ export async function disconnectGitProvider(
 
 export async function connectGitProvider(
   client: Client,
-  org: Org,
```
TooTallNate, line 33: Unrelated to this PR, but I noticed that this parameter wasn't being used.

packages/cli/src/util/link/repo.ts
```diff
@@ -275,31 +274,23 @@ export async function ensureRepoLink(
...
-          `${parsedRepoUrl.org}/${parsedRepoUrl.repo}`
-        );
-        output.log(
-          `Created new Project: ${output.link(
-            orgAndName,
-            `https://vercel.com/${orgAndName}`,
-            { fallback: false }
-          )}`
-        );
-      } catch (err) {
```
TooTallNate, line 297: This try/catch was the main issue that this PR is fixing. Any error except for "too_many_projects" was being silently ignored.

EndangeredMassa (approved):

erikareads (approved):

PR #12621, opened 2024-11-20
[cli] fix minor issues Vercel Bisect

erikareads:


.changeset/many-items-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
```
TooTallNate: No-op change

```suggestion
```

TooTallNate (changes requested):
If we're going to do this, can we set it up in a future-proof / global / automated way, such that it can be fixed with `eslint --fix` or equivalent?

erikareads:
> If we're going to do this, can we set it up in a future-proof / global / automated way, such that it can be fixed with `eslint --fix` or equivalent?

I'll look into that.

PR #12624, opened 2024-11-20
[cli] Fix deploy hanging on error

onsclom:
Errors during uploading would cause the CLI to hang:

https://github.com/user-attachments/assets/da51a65a-9eda-4371-bdf2-2b583f2e1c54

The problem was upload progress event connections not being disconnected. After fixing:

https://github.com/user-attachments/assets/2dacc808-a3e5-4e9b-a5fb-8555413b1983

The CLI still hangs for ~5 seconds after the error, but at least it closes on its own and doesn't show incorrect progress. There may still be more things to cancel to get rid of the remaining delay? Another solution entirely is to just do `process.exit(1)` after printing upload errors.

### DIFF
--- a/.changeset/thick-geckos-heal.md
+++ b/.changeset/thick-geckos-heal.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Fix deploy hanging on error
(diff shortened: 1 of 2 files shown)


TooTallNate (approved):
It's not sure clear to me why this would fix it, but the video looks like it's an improvement.

trek (approved):

erikareads (approved):

jeffsee55 (approved):

onsclom:
Closing in favor of https://github.com/vercel/vercel/pull/12635

This was a bandaid solution, and cancelling other uploads fixes the root problem.

PR #12635, opened 2024-11-21
[client] Cancel other file uploads when one fails

onsclom:
This solves the problem https://github.com/vercel/vercel/pull/12624 attempts to solve, but it fixes the root problem instead. Before, a file upload failing would not cancel the other file uploads.

### DIFF
--- a/packages/client/src/upload.ts
+++ b/packages/client/src/upload.ts
@@ -1,7 +1,7 @@
 import http from 'http';
 import https from 'https';
 import { Readable } from 'stream';
-import { EventEmitter } from 'events';
+import events, { EventEmitter } from 'node:events';
 import retry from 'async-retry';
 import { Sema } from 'async-sema';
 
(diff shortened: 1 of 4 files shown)


erikareads (approved):

TooTallNate (approved):

trek (approved):

packages/client/src/upload.ts
```diff
@@ -82,6 +82,9 @@ export async function* upload(
   const defaultAgent = apiUrl?.startsWith('https://')
     ? new https.Agent({ keepAlive: true })
     : new http.Agent({ keepAlive: true });
+  const abortController = new AbortController();
+  // default max is 10, but abort controller may have 50 connections at once
+  events.setMaxListeners(50);
```
EndangeredMassa, line 87: This defines global behavior for the current Node.js process. I don't think it should live as a side effect of this upload logic. Can we move it to the entrypoint of CLI or the entrypoint of a command? This will make it clearer that this is happening at a higher level.

PR #12636, opened 2024-11-21
Suggest `--archive=tgz` when upload limits are reached

onsclom:
The `api-upload` rate limit and file amount limits can be avoided with `--archive=tgz`. Previously, this wasn't clear from the errors.

New errors:
Screenshot 2024-11-21 at 12 57 04 PM
Screenshot 2024-11-21 at 12 58 36 PM

It was tricky deciding where exactly to put this error handling. There are multiple levels of try/catch which mutate errors as they bubble up, and lots of the existing error handling code uses `any` types in ways that make it scary to touch. The error handling code surrounding deployments can use some refactoring in the future, but for now I think this is a good, low-friction solution.

### DIFF
--- a/packages/cli/src/util/deploy/process-deployment.ts
+++ b/packages/cli/src/util/deploy/process-deployment.ts
@@ -289,7 +289,7 @@ export default async function processDeployment({
 export const archiveSuggestionText =
   'Try using `--archive=tgz` to limit the amount of files you upload.';
 
-export class UploadMissingArchiveError extends Error {
+export class UploadErrorMissingArchive extends Error {
   link = 'https://vercel.com/docs/cli/deploy#archive';
 }
 
--- a/packages/cli/src/util/deploy/process-deployment.ts
+++ b/packages/cli/src/util/deploy/process-deployment.ts
@@ -287,7 +287,11 @@ export default async function processDeployment({
 }
 
 export const archiveSuggestionText =
-  'Try using `--archive=tgz` to limit the amount of files you upload.\nLearn more: https://vercel.com/docs/cli/deploy#archive';
+  'Try using `--archive=tgz` to limit the amount of files you upload.';
+
+export class UploadMissingArchiveError extends Error {
+  link = 'https://vercel.com/docs/cli/deploy#archive';
+}
 
 export function handleErrorSolvableWithArchive(error: unknown) {
   const errorIsObject = typeof error === 'object' && error !== null;
@@ -301,7 +305,9 @@ export function handleErrorSolvableWithArchive(error: unknown) {
       'code' in error && error.code === 'too_many_files';
 
     if (isUploadRateLimit || isTooManyFilesLimit) {
-      throw new Error(`${error.message}\n${archiveSuggestionText}`);
+      throw new UploadMissingArchiveError(
+        `${error.message}\n${archiveSuggestionText}`
+      );
     }
   }
 }
--- a/packages/cli/src/util/deploy/process-deployment.ts
+++ b/packages/cli/src/util/deploy/process-deployment.ts
@@ -5,6 +5,7 @@ import {
   type VercelClientOptions,
   createDeployment,
 } from '@vercel/client';
+import { isErrorLike } from '@vercel/error-utils';
 import bytes from 'bytes';
 import chalk from 'chalk';
 import type { Agent } from 'http';
@@ -255,7 +256,10 @@ export default async function processDeployment({
         stopSpinner();
 
         if (!archive) {
-          handleErrorSolvableWithArchive(event.payload);
+          const maybeError = handleErrorSolvableWithArchive(event.payload);
+          if (maybeError) {
+            throw maybeError;
+          }
         }
 
         const error = await now.handleDeploymentError(event.payload, {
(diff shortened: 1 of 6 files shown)


packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -281,3 +285,23 @@ export default async function processDeployment({
...
+  'Try using `--archive=tgz` to limit the amount of files you upload.\nLearn more: https://vercel.com/docs/cli/deploy#archive';
+
+export function handleErrorSolvableWithArchive(error: unknown) {
+  const errorIsObject = typeof error === 'object' && error !== null;
+  if (errorIsObject && 'message' in error) {
+    const isUploadRateLimit =
+      'code' in error &&
+      error.code === 'rate_limited' &&
+      typeof error.message === 'string' &&
+      error.message.includes('api-upload');
```
erikareads: Can we match on something stronger than a portion of a string error message for this?

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -281,3 +285,23 @@ export default async function processDeployment({
     throw err;
   }
 }
+
+export const archiveSuggestionText =
+  'Try using `--archive=tgz` to limit the amount of files you upload.\nLearn more: https://vercel.com/docs/cli/deploy#archive';
```
TooTallNate: Rather than hard-coding the "Learn more:" part, let's set the `err.link` property and ensure we're rendering the error through [`output.prettyError()`](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/output/create-output.ts#L126-L146).

TooTallNate (changes requested):

packages/cli/src/util/deploy/process-deployment.ts
onsclom: Fixed! Turns out our client code was removing the specific rate limit name field higher up in the levels of try catch: https://github.com/vercel/vercel/blob/697160902eb68ba02a8bd2b24c0b42836600611c/packages/client/src/upload.ts#L179 Edited `DeploymentError` to include `rateLimitName` and used that.

packages/cli/src/util/deploy/process-deployment.ts
onsclom: Looks much nicer now!
image

erikareads (dismissed):

packages/client/src/errors.ts
```diff
@@ -1,9 +1,11 @@
...
-  constructor(err: { code: string; message: string }) {
+  constructor(err: { code: string; message: string; name?: string }) {
     super(err.message);
     this.code = err.code;
+    this.rateLimitName = err.name;
     this.name = 'DeploymentError';
   }
 
   code: string;
+  rateLimitName: string | undefined;
```
EndangeredMassa: **question (non-blocking):** ‏Are we sure that `name` is always the name of a rate limit error and nothing else? We may want `rateLimitName` to be more generic.

packages/cli/src/util/deploy/process-deployment.ts
TooTallNate: Nice! To go one step further, you may want to consider [printing the full command](https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/list/index.ts#L284-L289), like we do for pagination results.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -254,6 +254,10 @@ export default async function processDeployment({
       if (event.type === 'error') {
         stopSpinner();
 
+        if (!archive) {
+          handleErrorSolvableWithArchive(event.payload);
```
EndangeredMassa: **suggestion (non-blocking):** ‏change name to make it clearer this throws or have it return the error object, then throw here

It's not super obvious that this function throws. JavaScript doesn't have a good feature or convention for this, but some people use `doesWhateverOrThrows` to make it clear.

If we return the error instead, this function also becomes easier to test.

EndangeredMassa (dismissed):

packages/cli/src/util/deploy/process-deployment.ts
TooTallNate: A function can return `never` if it always throws.

TooTallNate: In this case, looks like it's not _always_, so I agree, returning would be clearer here.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -281,3 +285,30 @@ export default async function processDeployment({
...
+
+export function handleErrorSolvableWithArchive(error: unknown) {
+  const errorIsObject = typeof error === 'object' && error !== null;
+  if (errorIsObject && 'message' in error) {
+    const isUploadRateLimit =
+      'rateLimitName' in error &&
+      typeof error.rateLimitName === 'string' &&
+      ['api-upload-free', 'api-upload-paid', 'api-upload-extended'].includes(
+        error.rateLimitName
+      );
```
TooTallNate: This is getting a bit into "too much client-side logic" territory IMO. Maybe like this would be a little more future-proof?

```suggestion
      error.rateLimitName.startsWith('api-upload-')
```

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -281,3 +285,30 @@ export default async function processDeployment({
...
+
+export const archiveSuggestionText =
+  'Try using `--archive=tgz` to limit the amount of files you upload.';
+
+export class UploadErrorMissingArchive extends Error {
+  link = 'https://vercel.com/docs/cli/deploy#archive';
+}
+
+export function handleErrorSolvableWithArchive(error: unknown) {
+  const errorIsObject = typeof error === 'object' && error !== null;
```
TooTallNate: There's probably a helper or two in `@vercel/error-utils` that would be useful here.

packages/cli/src/util/deploy/process-deployment.ts
jeffsee55: Elected to return an Error or undefined, and throw if an error

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -281,3 +289,27 @@ export default async function processDeployment({
...
+export class UploadErrorMissingArchive extends Error {
+  link = 'https://vercel.com/docs/cli/deploy#archive';
+}
+
+export function handleErrorSolvableWithArchive(error: unknown) {
+  if (isErrorLike(error)) {
+    const isUploadRateLimit =
+      'rateLimitName' in error &&
+      typeof error.rateLimitName === 'string' &&
+      error.rateLimitName.startsWith('api-upload-');
```
erikareads: you changed the field to be `errorName`, but didn't update it here.

erikareads (changes requested):

erikareads (dismissed):

TooTallNate (dismissed):

erikareads (approved):

PR #12641, opened 2024-11-22
[cli] Explicitly route to `vercel set` subcommand

trek:
We currently rely on the `default` behavior to route to `set` because it's not being handled explicitly. No behavioral change here, just matching our more defined pattern used in other commands.

### DIFF
--- a/packages/cli/src/commands/alias/index.ts
+++ b/packages/cli/src/commands/alias/index.ts
@@ -85,7 +85,7 @@ export default async function alias(client: Client) {
         printHelp(setSubcommand);
         return 2;
       }
-      telemetry.trackCliSubcommandList(subcommandOriginal);
+      telemetry.trackCliSubcommandSet(subcommandOriginal);
       return set(client, args);
     default:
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/alias/index.ts
```diff
@@ -79,6 +79,14 @@ export default async function alias(client: Client) {
       }
       telemetry.trackCliSubcommandRemove(subcommandOriginal);
       return rm(client, args);
+    case 'set':
+      if (needHelp) {
+        telemetry.trackCliFlagHelp('alias', subcommandOriginal);
+        printHelp(setSubcommand);
+        return 2;
+      }
+      telemetry.trackCliSubcommandList(subcommandOriginal);
```
erikareads: This seems to be the wrong telemetry method?

erikareads (changes requested):

erikareads (approved):

jeffsee55 (approved):

PR #12643, opened 2024-11-22
[build-utils] feat: allowed CloudFront and S3 URLs to be overridden through environment variables

emeraldsanto:
### Summary

This PR aims to grant more control to consumers of the `vercel` CLI over where the `FileRef` contents are downloaded from for a given `digest`. This is done by adding 3 new environment variables, which take precedence over the current hardcoded URLs.

### DIFF
--- a/packages/build-utils/src/file-ref.ts
+++ b/packages/build-utils/src/file-ref.ts
@@ -45,13 +45,24 @@ export default class FileRef implements FileBase {
     this.mutable = mutable;
   }
 
-  private getMutableFilesCloudfrontUrl(): string {
+  /**
+   * Retrieves the URL of the CloudFront distribution for the S3
+   * bucket represented by {@link getNowFilesS3Url}.
+   *
+   * @returns The URL of the CloudFront distribution
+   */
+  private getNowFilesCloudfrontUrl(): string {
     return (
-      process.env.MUTABLE_FILES_CLOUDFRONT_URL ||
+      process.env.NOW_FILES_CLOUDFRONT_URL ||
       'https://dmmcy0pwk6bqi.cloudfront.net'
     );
   }
 
+  /**
+   * Retrieves the URL of the S3 bucket for storing ephemeral files.
+   *
+   * @returns The URL of the S3 bucket
+   */
   private getNowEphemeralFilesS3Url(): string {
     return (
       process.env.NOW_EPHEMERAL_FILES_S3_URL ||
@@ -59,6 +70,11 @@ export default class FileRef implements FileBase {
     );
   }
 
+  /**
+   * Retrieves the URL of the S3 bucket for storing files.
+   *
+   * @returns The URL of the S3 bucket
+   */
   private getNowFilesS3Url(): string {
     return process.env.NOW_FILES_S3_URL || 'https://now-files.s3.amazonaws.com';
   }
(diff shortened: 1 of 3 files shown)


packages/build-utils/src/file-ref.ts
```diff
@@ -45,22 +45,55 @@ export default class FileRef implements FileBase {
...
 
+  /**
+   * Retrieves the URL of the CloudFront distribution for the S3
+   * bucket represented by {@link getNowFilesS3Url}.
+   *
+   * @returns The URL of the CloudFront distribution
+   */
+  private getNowFilesCloudfrontUrl(): string {
+    return (
+      process.env.NOW_FILES_CLOUDFRONT_URL ||
```
EndangeredMassa: **suggestion (non-blocking):** ‏Can we add some minimal URL validation here (and in the other methods) to make errors clearer if someone puts non-URLs in this env var?

Something like:

```
return getEnvUrl('NOW_FILES_CLOUDFRONT_URL') || 'https://dmmcy0pwk6bqi.cloudfront.net';
```

with:

```
function getEnvUrl(envVarKey: string): string {
  const value = process.env[envVarKey];
  if (!validateUrl(value)) {
    throw new Error(`Environment Variable "${envVarKey}" was set to a value that was not a valid URL: ${value}`);
  }
  return value;
}
```

with some reasonable implementation of a `validateUrl`

EndangeredMassa (dismissed):

packages/build-utils/src/file-ref.ts
emeraldsanto: Done in [`dde90f1`](https://github.com/vercel/vercel/pull/12643/commits/dde90f1f1ac854c9f08728d6b0af07b979fba396), but I've opted to not surface the value as part of the error message, in case it holds a sensitive value?

packages/build-utils/src/file-ref.ts
```diff
@@ -139,3 +142,30 @@ export default class FileRef implements FileBase {
...
+ *
+ * @param key The `process.env` member which holds the value to validate
+ * @returns The valid URL string or `null` if `key` is not set on `process.env`
+ */
+function getEnvAsUrlOrThrow(
+  key: keyof (typeof process)['env']
+): ReturnType<URL['toString']> | null {
+  const value = process.env[key];
+
+  if (value === undefined) return null;
```
emeraldsanto: I originally had `if (!value) return null`, but now that we're checking for explicit `undefined` (to explicitly error on other falsy values) I wonder if `null` still makes sense as a return value?

packages/build-utils/src/file-ref.ts
```diff
@@ -139,3 +142,30 @@ export default class FileRef implements FileBase {
...
+
+  try {
+    new URL(value);
```
emeraldsanto, line 160: This constructor performs validation as a side effect. It can be swapped for `URL.canParse(value)` when we support >= Node.js 20.x as it was added in `19.9.0` ([ref](https://nodejs.org/docs/latest-v19.x/api/url.html#urlcanparseinput-base)).

packages/build-utils/src/file-ref.ts
EndangeredMassa: In general: I prefer to use `undefined` over `null` in essentially all cases, but that's me. ([ref](https://massalabs.com/blog/prefer-null-or-undefined-in-javascript))

Here specifically: Because this is a wrapper over `process.env[someKey]`, which returns `undefined` if the key has no value, people will probably assume the return value of this function is also `undefined` in that case.

EndangeredMassa (dismissed):

packages/build-utils/src/file-ref.ts
emeraldsanto: Interesting read! I've changed to use `undefined` to align with your preference as it makes sense and your team is a lot more involve in this codebase than I 😌

EndangeredMassa (approved):

PR #12645, opened 2024-11-22
vercel-remix-run-dev-2-15-0

erikareads:


TooTallNate:
`@remix-run/dev` is no longer a dev dependency (as of #12762), so this is no longer relevant.

PR #12646, opened 2024-11-22
[tests] fixed skipped tests for ember and middleware

EndangeredMassa:
Reverts vercel/vercel#12644

---

Fixes some skipped tests.

### ember-v3

This older version of ember (and ember-cli) doesn't work on Node 22. I pinned this fixture to Node 20. https://github.com/vercel/vercel/pull/12646/commits/45ab9ff29d9f051a4ac401912fcfde194fa50717

### middleware error code

This error code has been updated at the system level. I updated the assertions to match. https://github.com/vercel/vercel/pull/12646/commits/fdd4786e2538a4425b0215b1f1b6bdbb8025be4c


packages/cli/src/util/dev/server.ts
```diff
@@ -1583,7 +1583,7 @@ export default class DevServer {
           req,
           res,
           requestId,
-          'EDGE_FUNCTION_INVOCATION_FAILED',
+          'MIDDLEWARE_INVOCATION_FAILED',
```
EndangeredMassa, line 1586: This is the main change that better matches production behavior.

packages/cli/test/dev/integration-4.test.ts
```diff
@@ -248,21 +248,25 @@ test(
...
-// eslint-disable-next-line jest/no-disabled-tests
-test.skip(
+test(
   '[vercel dev] Middleware with error in function handler',
   testFixtureStdio('middleware-error-in-handler', async (testPath: any) => {
-    await testPath(500, '/', /EDGE_FUNCTION_INVOCATION_FAILED/);
+    await testPath(
+      500,
+      '/',
+      /INTERNAL_SERVER_ERROR|MIDDLEWARE_INVOCATION_FAILED/g
```
EndangeredMassa, line 257: These assertions check two possible options because a deployed test of this scenario produces one result that the dev server can't currently replicate.

I spent some time digging into what the difference is between this and the next test scenarios that produces 2 different error codes in a deployment (`INTERNAL_SERVER_ERROR` and `MIDDLEWARE_INVOCATION_FAILED`), but only one in the dev server (`MIDDLEWARE_INVOCATION_FAILED`). I need to move on to other things.

If anyone wants to pair on digging deeper here, let me know!

packages/cli/test/dev/integration-4.test.ts
```diff
@@ -248,21 +248,25 @@ test(
...
-    await testPath(500, '/', /EDGE_FUNCTION_INVOCATION_FAILED/);
+    /*
+      These assertions check two possible options because a deployed test
+      of this scenario produces one result that the dev server can't currently
+      replicate.
+    */
+    const devCode = 'MIDDLEWARE_INVOCATION_FAILED';
+    const deploymentCode = 'INTERNAL_SERVER_ERROR';
+
+    await testPath(500, '/', new RegExp(`${devCode}|${deploymentCode}`, 'g'));
```
EndangeredMassa, line 269: I spent some time digging into what the difference is between this and the previous test scenarios that produces different error codes, but couldn't find a meaningful way.

If anyone wants to pair on digging deeper here, let me know!

trek (approved):

PR #12647, opened 2024-11-22
[client] Refactor upload abort controllers to avoid setting global max listeners

onsclom:
@EndangeredMassa brought up a good point in https://github.com/vercel/vercel/pull/12635#discussion_r1854520641. However, it also feels unfortunate moving the max listener 50 setting so far from the semaphore of size 50.

Instead of moving the global setting, I removed it entirely! Now each request gets its own abort controller so we never go past 1 listener per signal.

Tested with a rate limit error manually and it works as expected.

### DIFF
--- a/packages/client/src/upload.ts
+++ b/packages/client/src/upload.ts
@@ -1,7 +1,7 @@
 import http from 'http';
 import https from 'https';
 import { Readable } from 'stream';
-import events, { EventEmitter } from 'node:events';
+import { EventEmitter } from 'node:events';
 import retry from 'async-retry';
 import { Sema } from 'async-sema';
 
@@ -82,9 +82,8 @@ export async function* upload(
   const defaultAgent = apiUrl?.startsWith('https://')
     ? new https.Agent({ keepAlive: true })
     : new http.Agent({ keepAlive: true });
-  const abortController = new AbortController();
-  // default max is 10, but abort controller may have 50 connections at once
-  events.setMaxListeners(50);
+  const abortControllers = new Set<AbortController>();
+  let aborted = false;
 
   shas.forEach((sha, index) => {
     const uploadProgress = uploads[index];
@@ -129,6 +128,11 @@ export async function* upload(
 
         let err;
         let result;
+        const abortController = new AbortController();
+        abortControllers.add(abortController);
+        if (aborted) {
+          abortController.abort();
+        }
 
         try {
           const res = await fetch(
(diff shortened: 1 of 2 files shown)


packages/client/src/upload.ts
```diff
@@ -129,6 +128,11 @@ export async function* upload(
 
         let err;
         let result;
+        const abortController = new AbortController();
+        abortControllers.add(abortController);
+        if (aborted) {
+          abortController.abort();
```
EndangeredMassa: **question (blocking):** ‏If we're aborting because `aborted` is true (from a pervious attempt?), why are we continuing to execute a `fetch` with an alrady aborted `abortController`?

I may be misunderstanding the logic here.

EndangeredMassa (changes requested):

packages/client/src/upload.ts
onsclom: Good question. I did this to match what happened before when there was just 1 abort controller. All file uploads were queued up, and when one file upload errored then new uploads would fetch with the already aborted `abortController` (effectively cancelling them). One upload erroring doesn't automatically cancel new uploads from starting because they are all queued in a [`forEach`](https://github.com/vercel/vercel/pull/12647/files#diff-38a24c9ef933a33a4a249552ba0411de3aa9ade727d65875c36fa2df7da466e3R88).

This worked, but now I'm realizing now each remaining upload did some unnecessary work between the `await semaphore.acquire();` and `fetch`.

Changed this to `return bail()` instead as early as possible, and manually tested it!

packages/client/src/upload.ts
```diff
@@ -212,9 +221,7 @@ export async function* upload(
 
   while (Object.keys(uploadList).length > 0) {
     try {
-      const event = await Promise.race(
-        Object.keys(uploadList).map((key): Promise<any> => uploadList[key])
-      );
+      const event = await Promise.race(Object.values(uploadList));
```
EndangeredMassa, line 224: **question (non-blocking):** ‏Unrelated, but is `Promise.race` correct here?

This will make the first settled promise be the result of the race. If a following promise rejects, this logic will not handle that. This seems wrong. Maybe we visit this in a follow up PR.

EndangeredMassa (approved):

packages/client/src/upload.ts
onsclom: My understanding is that since this is in a while loop, it handles every promise in the order they finish. It would `yield` an initial success, then `return yield` a following rejection.

Though I'm definitely not used to all this async logic + semaphore + generator function behavior and I'm not sure I understand it all, so could be wrong.

EndangeredMassa: Oh, it's finding the first settled one and removing it. This makes sense.

PR #12654, opened 2024-11-25
[cli] add eslint prefer-const rule and fixes

erikareads:
Adds the eslint `prefer-const` rule to the `cli` package. Fixes were applied, with the exception of `let [item1, item2] = <var>` pattern matching where one of the items _is_ being reassigned and therefore can't be a const.

In the future, it would be worth investigating these uses to see if a non-reassignment version of the code can be refactored.


TooTallNate (dismissed):

.changeset/famous-schools-press.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
EndangeredMassa, line 3: **suggestion (blocking):** ‏this needs a changeset because non-test code did change, just in case

EndangeredMassa (changes requested):

EndangeredMassa (approved):

PR #12655, opened 2024-11-25
[cli] Use `client.stdout` when printing table for `vc cert issue`

TooTallNate:
I noticed that there's some "leakage" being printed when running the unit tests. This fixes one such case in the `vercel cert issue` command.

Switching from `process.stdout` to `client.stdout` allows us to write assertions against the table as well, so I've updated the unit test to do so.


erikareads (dismissed):

.changeset/blue-otters-joke.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
EndangeredMassa, line 3: **suggestion (blocking):** ‏this needs a changeset because non-test code did change, just in case

EndangeredMassa (dismissed):

.changeset/blue-otters-joke.md
TooTallNate: It _is_ a no-op change in this case, since we know it is. But I don't mind having a hard-line stance as well.

erikareads (approved):

EndangeredMassa (approved):

PR #12656, opened 2024-11-25
Upgrade turbo from 1 to 2

jeffsee55:


.github/workflows/test-18.yml
```diff
@@ -82,11 +82,11 @@ jobs:
       - run: pnpm install
 
       - name: Build ${{matrix.packageName}} and all its dependencies
-        run: node utils/gen.js && node_modules/.bin/turbo run build --cache-dir=".turbo" --log-order=stream --scope=${{matrix.packageName}} --include-dependencies --no-deps
+        run: node utils/gen.js && node_modules/.bin/turbo run build --cache-dir=".turbo" --log-order=stream --filter=${{matrix.packageName}}
         env:
           FORCE_COLOR: '1'
       - name: Test ${{matrix.packageName}}
-        run: node utils/gen.js && node_modules/.bin/turbo run ${{matrix.testScript}} --summarize --cache-dir=".turbo" --log-order=stream --scope=${{matrix.packageName}} --no-deps -- ${{ join(matrix.testPaths, ' ') }}
+        run: node utils/gen.js && node_modules/.bin/turbo run ${{matrix.testScript}} --summarize --cache-dir=".turbo" --log-order=stream --filter=${{matrix.packageName}} -- ${{ join(matrix.testPaths, ' ') }}
```
jeffsee55, line 89: I _think_ `--filter` does everything we want now ([note](https://github.com/vercel/turborepo/issues/852#issuecomment-1092278332) about `--deps` deprecated in favor of `--filter`) ([changelog](https://github.com/vercel/turborepo/pull/8137))

packages/functions/package.json
```diff
@@ -43,7 +43,7 @@
   "scripts": {
     "pretest": "pnpm run build:code",
     "test": "vitest",
-    "build": "pnpm run build:code && pnpm run build:docs",
+    "build": "pnpm run build:code",
```
jeffsee55, line 46: The side-effect of this produces a `docs` folder that is not gitignored, meaning this build runs in an infinite loop during `watch`. I kind of expected turbo to have some sort of "watch" ignore pattern but here's their [official recommendation](https://turbo.build/repo/docs/reference/watch#task-outputs):

> If you have tasks that write to files checked into source control, there is a possibility that Watch Mode will run in an infinite loop. This is because Watch Mode watches your files for changes and will re-run tasks in packages that have changed. If a task creates a change, then that will trigger the task again.

> Watch Mode has some logic to prevent this from happening using file hashes, but it isn't foolproof. To avoid this issue, we recommend removing any task outputs from git.

So not sure how we're using these outputs, can I just add `docs` to the `.gitignore`?

PR #12657, opened 2024-11-25
Use esbuild instead of ts-node for 'pnpm vercel' command

jeffsee55:
I couldn't think of a great reason we don't just build the CLI before invoking it when testing things out locally via `pnpm vercel`. Downsides of `ts-node` are that it's not exactly the same as the real CLI behavior because it does some magic by inserting it's path into argv (or something awkward like that), while the downside to doing a full build are that it is slightly slower.

Timing `alias ls` as a benchmark (`time pnpm vercel alias ls)`

- Using `ts-node`: `0.82s user 0.26s system 99% cpu 1.084 total`
- Using `esbuild` (no prev build): `1.56s user 0.46s system 146% cpu 1.374 total`
- Using `esbuild` (prev build): `0.82s user 0.12s system 71% cpu 1.317 total`

I'm not sure if the `dist` folder being present is actually a factor, a couple of times I did see a slightly slower execution with the `dist` there. IMO the speed difference is an acceptable tradeoff.

---

Also, while testing this out I removed the `"vercel"` script from the `package.json` and `pnpm vercel alias ls` still worked, using the global `vercel` installation. I kind of think we probably don't want this and maybe we should go back to using `dev`, recently changed here.

### DIFF
--- a/.changeset/friendly-frogs-mix.md
+++ b/.changeset/friendly-frogs-mix.md
@@ -1,0 +1,4 @@
+---
+---
+
+Use esbuild instead of ts-node for vercel script
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -18,8 +18,7 @@
     "test-dev": "pnpm test test/dev/",
     "coverage": "codecov",
     "build": "node scripts/build.mjs",
-    "dev": "echo \"'pnpm dev [command]' has been removed. Use 'pnpm vercel [command]' instead.\" && exit 1",
-    "vercel": "ts-node ./src/index.ts",
+    "vercel": "pnpm run build && node ./scripts/start.js",
     "vc": "pnpm vercel",
     "type-check": "tsc --noEmit"
   },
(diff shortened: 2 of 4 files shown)


TooTallNate (changes requested):
This seems reasonable to me, and the benefit of eliminating the difference of execution environment during dev is a big plus.

Note that if we go this route, there's some configuration in `tsconfig.json` that can also be removed. The devDep can also probably be removed?

jeffsee55:
> This seems reasonable to me, and the benefit of eliminating the difference of execution environment during dev is a big plus.
> 
> Note that if we go this route, there's some configuration in `tsconfig.json` that can also be removed. The devDep can also probably be removed?

Yeah, let me take a stab at those. May reach out for help, I had originally thought we'd need to keep it here because of `vc dev` but now I realize that's in `@vercel/node`.

EndangeredMassa:
I appreciate this investigation.

I would prefer we keep the speed benefit of `ts-node`, but would be fine making this change if every one else agreed.

PR #12663, opened 2024-11-26
[cli] track teamId right before telemetry is sent

erikareads:
`teamId` wasn't appropriately sending in situations where the user had set `--token`. Now, if that token is used for an API endpoint where `teamId` is set, that is appropriately passed along in the telemetry events.

### DIFF
--- a/.changeset/tender-bears-warn.md
+++ b/.changeset/tender-bears-warn.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] fix issue that prevented teamId from being sent in some situations
--- a/.changeset/tender-bears-warn.md
+++ b/.changeset/tender-bears-warn.md
@@ -2,4 +2,4 @@
 "vercel": patch
 ---
 
-[cli] fix issue that prevented teamId from being sent in some situations
+Ensure `teamId` metric is sent when `--token` option is used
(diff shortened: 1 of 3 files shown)


.changeset/tender-bears-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] fix issue that prevented teamId from being sent in some situations
```
TooTallNate:
```suggestion
Ensure `teamId` metric is sent when `--token` option is used
```

packages/cli/src/util/telemetry/index.ts
```diff
@@ -238,8 +239,9 @@ export class TelemetryEventStore {
       }
       const events = this.events.map(event => {
         delete event.sessionId;
-        const { eventTime, teamId, ...rest } = event;
-        return { event_time: eventTime, team_id: teamId, ...rest };
+        delete event.teamId;
```
TooTallNate, line 242: If I'm reading correct, the `team_id` value will only be sent for one metric in the events array - is that correct / intended?

erikareads: No, the `team_id` field will be set for every event in the batch of events we'll be sending. The main part of this code is that we named the field `teamId` in the `Event` type, and the API endpoint is expecting `team_id` instead.

TooTallNate: Yep, I misread. Thanks!

TooTallNate (approved):

PR #12671, opened 2024-11-28
Add support for splitting gzipped tar into chunks

jeffsee55:
Adds support for splitting the archive into ~100MB chunks (`.vercel/source.tgz.part1`) when using the `VERCEL_SPLIT_ARCHIVE` env var.


.changeset/weak-sheep-dance.md
```diff
@@ -0,0 +1,4 @@
+---
+"@vercel/build-utils": patch
+"@vercel/client": patch
+---
```
jeffsee55, line 4: Not sure what we usually do about this, but this change isn't ready to be used yet so I don't want it to show up in the changelogs

packages/build-utils/src/fs/stream-to-buffer.ts
```diff
@@ -26,3 +26,37 @@ export default function streamToBuffer(
     });
   });
 }
+
+export function streamToBufferChunks(
```
TooTallNate: Can we add some unit tests for this function?

packages/build-utils/src/fs/stream-to-buffer.ts
```diff
@@ -26,3 +26,37 @@ export default function streamToBuffer(
...
+  return new Promise<Buffer[]>((resolve, reject) => {
+    const byteLimit = 1024 * 1024 * mbLimit;
+
+    const buffers: Buffer[] = [];
+
+    let currentBuffer: Buffer;
+    stream.on('data', (chunk: Buffer) => {
+      if (!currentBuffer) {
+        currentBuffer = chunk;
+      } else if (currentBuffer.length > byteLimit) {
```
TooTallNate: If I'm reading correctly, the buffers in the array won't have a length of _exactly_ `byteLimit`?

packages/build-utils/src/fs/stream-to-buffer.ts
```diff
@@ -26,3 +26,37 @@ export default function streamToBuffer(
...
+  stream: NodeJS.ReadableStream,
+  mbLimit = 100
+): Promise<Buffer[]> {
+  return new Promise<Buffer[]>((resolve, reject) => {
+    const byteLimit = 1024 * 1024 * mbLimit;
+
+    const buffers: Buffer[] = [];
+
+    let currentBuffer: Buffer;
+    stream.on('data', (chunk: Buffer) => {
```
TooTallNate: Curious why not use the `for await` syntax for readable streams? Would make the logic easier to understand I think.

TooTallNate (changes requested):

packages/build-utils/src/fs/stream-to-buffer.ts
jeffsee55: Yeah, I had 2 other approaches to getting this to exactly 100MB but then figured it probably wasn't worth the effort. Do you think it'd matter? Have a straightforward approach in mind?

packages/build-utils/src/fs/stream-to-buffer.ts
jeffsee55: Because I didn't know you could do that... 👍

packages/client/src/create-deployment.ts
```diff
@@ -99,18 +99,33 @@ export default function buildCreateDeployment() {
...
-          ],
-        ]);
+        if (process.env.SPLIT_SOURCE_ARCHIVE) {
+          const chunkedTarBuffers = await streamToBufferChunks(tarStream);
+          debug(`Packed tarball into ${chunkedTarBuffers.length} chunks`);
+          files = new Map(
+            chunkedTarBuffers.map((chunk, index) => [
+              hash(chunk),
+              {
+                names: [join(workPath, `.vercel/source.tgz.part${index}`)],
```
onsclom:
```suggestion
                names: [join(workPath, `.vercel/source.tgz.part${index+1}`)],
```

1-based indexing for the file names would be nice to match prior art: https://www.win-rar.com/split-files-archive.html?&L=0

packages/build-utils/test/unit.stream-to-buffer.test.ts
```diff
@@ -0,0 +1,28 @@
+import { it, describe, expect } from 'vitest';
+import { streamToBufferChunks } from '../src/fs/stream-to-buffer';
+import { Readable } from 'stream';
+
+describe('streamToBufferChunks', () => {
+  it("returns 1 buffer when the total volume doesn't exceed 100mb", async () => {
+    const tenMbs = 10 * (1024 * 1024);
```
TooTallNate: How about a `const MB = 1024 * 1024` at the top instead of repeating `(1024 * 1024)` a bunch of times in this file?

onsclom:
Fixed the integration tests for this PR!

Ideally, we'd test that that the gzip+tar'd single file archive bytes equal the concatenated version bytes, but two different archives of the same thing can be different for many reasons that are difficult to control. So instead, we ungzip+untar them to make sure we end up with something equivalent to the original project.

packages/client/package.json
```diff
@@ -35,8 +35,11 @@
     "@types/recursive-readdir": "2.2.0",
     "@types/tar-fs": "1.16.1",
     "jest-junit": "16.0.0",
-    "vitest": "2.0.1",
-    "typescript": "4.9.5"
+    "tar-stream": "3.1.7",
+    "tmp-promise": "1.0.3",
+    "typescript": "4.9.5",
+    "vitest": "2.1.3",
+    "vite": "5.1.6"
```
TooTallNate: Is it not enough to have these deps in the root level `package.json`?

packages/client/src/create-deployment.ts
```diff
@@ -99,18 +99,33 @@ export default function buildCreateDeployment() {
...
-          [
-            hash(tarBuffer),
-            {
-              names: [join(workPath, '.vercel/source.tgz')],
-              data: tarBuffer,
-              mode: 0o666,
-            },
-          ],
-        ]);
+        if (process.env.SPLIT_SOURCE_ARCHIVE) {
```
TooTallNate: Not a huge fan of doing the env var check directly in `@vercel/client`. Perhaps it should be an option that gets passed in here? Maybe we'll want to allow a CLI flag to trigger the behavior or change the interface in some other way in the future.

TooTallNate: Also, we usually prefix env vars with `VERCEL_`

packages/client/package.json
jeffsee55: I had set this up, I've always put `devDependencies` in the `package.json` where they're used, but yeah, we don't have any tooling around ensuring that we're on the same version so can see an argument for just keeping them in one place. How do you decide whether to put them where they're used vs at the root?

TooTallNate: One thing I consider is if it's a dev dependency, then it doesn't need to be listed in the `package.json` which is published to npm, so by moving to the top-level then they do not. Perhaps it would make sense if npm stripped those (and `scripts`) at publish time, but that's not the way it works.

.changeset/metal-shoes-lay.md
```diff
@@ -0,0 +1,2 @@
+---
```
TooTallNate: Delete?

.changeset/weak-sheep-dance.md
TooTallNate: Both would be minor IMO. build-utils introduces a new exported function. client supports a new value for the `archive` option.

packages/client/tests/unit.utils.test.ts
```diff
@@ -1,3 +1,4 @@
+import { it, expect, describe } from 'vitest';
```
TooTallNate: These same imports are on line 5. As a more general observation, there seems to be a bit of noise in this PR related to vitest setup. Is there some carry-over from a weird merge from #12530 in this PR? Seems like we could use some cleanup.

packages/client/tests/unit.vercelignore.test.ts
```diff
@@ -1,3 +1,4 @@
+import { it, describe } from 'vitest';
```
TooTallNate: Ditto

trek (approved):

TooTallNate (approved):

PR #12689, opened 2024-12-03
[cli] bump express version

erikareads:


packages/cli/package.json
```diff
@@ -124,7 +124,7 @@
     "esm": "3.1.4",
     "execa": "3.2.0",
     "expect": "29.5.0",
-    "express": "4.17.1",
+    "express": "4.17.3",
```
TooTallNate, line 127: I don't think we use express in CLI. Maybe this one can be removed?

EndangeredMassa: We use it in tests for a mock server

EndangeredMassa (approved):

trek (approved):

PR #12703, opened 2024-12-04
[cli][node] update express even more

EndangeredMassa:
Updating `express` ([again](https://github.com/vercel/vercel/pull/12689)) because [the body-parser vulnerability](https://github.com/vercel/vercel/security/dependabot/5504) requires a further update.


packages/cli/package.json
```diff
@@ -66,7 +66,6 @@
     "@types/debug": "0.0.31",
     "@types/dotenv": "6.1.1",
     "@types/escape-html": "0.0.20",
-    "@types/express": "4.17.13",
```
EndangeredMassa, line 69: This is already [the latest types package version](https://www.npmjs.com/package/@types/express?activeTab=versions) and it still includes the [vulnerable version of `body-parser`](https://github.com/vercel/vercel/security/dependabot/5504).

I think removing it is OK. We only use it in tests for the mock server.

packages/node/bench/package.json
```diff
@@ -1,11 +0,0 @@
...
-  "private": true,
-  "name": "bench",
-  "version": "1.0.0",
-  "main": "index.js",
-  "license": "UNLICENSED",
-  "dependencies": {
-    "express": "4.17.3",
-    "fs-extra": "8.0.1"
-  }
-}
```
EndangeredMassa, line 11: This was left over from https://github.com/vercel/vercel/pull/12697

We could remove it in a separate PR, but CI is being touchy enough as it is.

packages/cli/types/express/index.d.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Manually describe the bits of `express` we use because:
+ * - `body-parser` has a vulnerability report: https://github.com/vercel/vercel/security/dependabot/5504
+ * - which is a transitive dependency of `express` and `@types/express`
+ * - we need to update to `express@4.21.1` to update `body-parser` past the patched version
+ * - the corresponding `@types/express` package doesn't exist
+ * - the latest `@types/express` package still has an affected `body-parser` version
+ * So, we removed `@types/express` from this package.
+ */
```
EndangeredMassa, line 9: The types package had to go. If there's a better approach, let me know!

erikareads (approved):

onsclom (approved):

trek (approved):

PR #12718, opened 2024-12-09
Remove environment var `VERCEL_SPLIT_ARCHIVE` in favor of split-tgz archive option

onsclom:
https://github.com/vercel/vercel/pull/12671 actually adds support for splitting archives by doing `VERCEL_SPLIT_ARCHIVE=1` OR `--archive=split-tgz`, but this makes it more clear we support this through the flag.


packages/cli/src/commands/deploy/command.ts
```diff
@@ -87,8 +87,7 @@ export const deployCommand = {
       type: String,
       argument: 'FORMAT',
       deprecated: false,
-      description:
-        'Compress the deployment code into a file before uploading it',
+      description: 'Compress the deployment code before uploading it',
```
TooTallNate: The text was arguably more correct before. `--archive` does not _only_ compress, but also _combines_.

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -97,8 +97,7 @@ export default async function processDeployment({
     vercelOutputDir,
     rootDirectory,
     skipAutoDetectionConfirmation,
-    archive:
-      process.env.VERCEL_SPLIT_ARCHIVE && archive ? 'split-tgz' : archive,
+    archive: archive,
```
TooTallNate: Just to clarify, does this remove the env var support completely?

packages/cli/src/commands/deploy/command.ts
onsclom: Good point, I was just avoiding how `--archive=split-tgz` compresses the deployment code into multiple files, not just one. Maybe we go with 'Compress the deployment code into an archive before uploading it'?

packages/cli/src/util/deploy/process-deployment.ts
onsclom: Yes. The thinking is we can tell customers to try out split archive deployments by doing `--archive=split-tgz`, and if everything goes smooth we make splitting the default `--archive=tgz` behavior.

TooTallNate: I was going to say that env var would make it easier to feature flag, but for CLI running locally (which would be the case for `--archive`) we don't really have access to feature flag env vars in the same way as for deployments, so SGTM.

TooTallNate:
```suggestion
    archive,
```

EndangeredMassa (approved):

trek (approved):

TooTallNate (approved):

PR #12719, opened 2024-12-09
Don't fail the deployment when a deleted package dir exists due to build cache

TooTallNate:
`package.json` might not exist if this directory exists due to the Vercel deployment's build cache (even though the package has been deleted). So skip in that case.


trek (dismissed):

erikareads (dismissed):

api/_lib/script/build.ts
```diff
@@ -71,10 +71,16 @@ async function main() {
     }
 
     const fullDir = join(packagesDir, pkg);
-    const packageJsonRaw = await fs.readFile(
-      join(fullDir, 'package.json'),
-      'utf-8'
-    );
+    const packageJsonRaw = await fs
+      .readFile(join(fullDir, 'package.json'), 'utf-8')
+      .catch(() => null);
```
EndangeredMassa, line 76: **suggestion (non-blocking):** ‏can we log the error?

api/_lib/script/build.ts
```diff
@@ -71,10 +71,16 @@ async function main() {
...
+    if (!packageJsonRaw) {
+      // `package.json` might not exist if this directory exists due to the
+      // Vercel deployment's build cache (even though the package has been
+      // deleted). So skip in that case.
+      continue;
```
EndangeredMassa, line 81: **suggestion (non-blocking):** ‏can we log when this happens?

EndangeredMassa (approved):

onsclom (approved):

PR #12723, opened 2024-12-10
[tests] rename env var we use to skip linking step in dev tests

trek:
`__VERCEL_SKIP_DEV_CMD` does not skip the dev command, it skips linking to a project when `vercel dev` is run during tests. Just a rename for clarity.

### DIFF
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -30,7 +30,7 @@ export default async function dev(
   // retrieve dev command
   let link = await getLinkedProject(client, cwd);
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/dev/dev.ts
```diff
@@ -30,7 +30,7 @@ export default async function dev(
   // retrieve dev command
   let link = await getLinkedProject(client, cwd);
 
-  if (link.status === 'not_linked' && !process.env.__VERCEL_SKIP_DEV_CMD) {
```
EndangeredMassa, line 33: **suggestion (blocking):** ‏We should either still support the old value as well OR add a major changeset

EndangeredMassa (changes requested):

PR #12728, opened 2024-12-10
bump ts-jest version

erikareads:
Bumps the version of `ts-jest` for the monorepo, and specifically for `get-package-json` by a minor version, to update the version of `semver` transitively.

### DIFF
--- a/.changeset/chilly-spoons-pay.md
+++ b/.changeset/chilly-spoons-pay.md
@@ -1,0 +1,5 @@
+---
+"@vercel-internals/get-package-json": patch
+---
+
+bump ts-jest version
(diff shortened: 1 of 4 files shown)


.changeset/chilly-spoons-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel-internals/get-package-json": patch
+---
+
+bump ts-jest version
```
TooTallNate: Changeset should be empty for internal packages (they're not published)

```suggestion
---
```

onsclom (approved):

PR #12733, opened 2024-12-11
Enable new storage integrations via wizard

JakeUskoski:
We moved from using a `type` field to a `protocols` object with some additional information for determining the category of an integration. New storage products don't have the old `type` field, instead using the new protocols, but are otherwise still storage products with the same wizard setup needs. New integrations that would otherwise be fine to use the wizard for setup were directing users to the dashboard to complete the installation process (undesired behaviour).

This PR adds in the typing for the protocols currently in use (as of this PR, just storage), so that our newest integration products can be added purely via the CLI.

### DIFF
--- a/packages/cli/src/util/integration/types.ts
+++ b/packages/cli/src/util/integration/types.ts
@@ -32,12 +32,27 @@ export interface MetadataSchema {
   required?: string[];
 }
 
+export type IntegrationProductProtocolBase = {
+  status: 'enabled' | 'disabled';
+};
+
+export type StorageIntegrationProtocol = IntegrationProductProtocolBase & {
+  repl?: {
+    enabled: boolean;
+    supportsReadOnlyMode: boolean;
+    welcomeMessage?: string;
+  };
+};
+
 export interface IntegrationProduct {
   id: string;
   slug: string;
   name: string;
   shortDescription: string;
-  type: 'storage' | string;
+  type?: 'storage' | string;
+  protocols?: {
+    storage?: StorageIntegrationProtocol;
+  };
   metadataSchema: MetadataSchema;
 }
 
--- a/.changeset/plenty-phones-happen.md
+++ b/.changeset/plenty-phones-happen.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+Bug fix: new integrations using protocols for storage category can be created via the CLI wizard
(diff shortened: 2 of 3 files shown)


mbrakken (approved):

packages/cli/src/util/integration/types.ts
```diff
@@ -32,12 +32,27 @@ export interface MetadataSchema {
...
   slug: string;
   name: string;
   shortDescription: string;
-  type: 'storage' | string;
+  type?: 'storage' | string;
+  protocols?: {
+    storage?: StorageIntegrationProtocol;
+  };
   metadataSchema: MetadataSchema;
 }
```
EndangeredMassa, line 57: **suggestion (non-blocking):** ‏use union types for stronger type checking

When the overall true type varies when the value of a property, I like to use a union type for stronger type checking. This will prevent the case where both `storage` and `protocols` are missing or both are populated.

Example:

```ts
type IntegrationProductBase = {
  id: string;
  slug: string;
  name: string;
  shortDescription: string;
  metadataSchema: MetadataSchema;
}
type OldIntegrationProduct = IntegrationProductBase & {
  // `type` no longer optional
  type: 'storage' | string;
}

type NewIntegrationProduct = IntegrationProductBase & {
  // `protocols` no longer optional
  protocols: {
    storage?: StorageIntegrationProtocol;
  };
}

type IntegrationProduct = OldIntegrationProduct | NewIntegrationProduct;

let a: IntegrationProduct = {
  id: 'a',
  slug: 'product-a',
  name: 'Product A',
  shortDescription: 'product called "a"',
  metadataSchema: {},

  type: 'storage',
  
}

let b: IntegrationProduct = {
  id: 'b',
  slug: 'product-b',
  name: 'Product B',
  shortDescription: 'product called "B"',
  metadataSchema: {},

  protocols: {
    storage: {}
  },
}
```

EndangeredMassa (approved):

EndangeredMassa:
I removed this from the queue to give you a chance to review my feedback before merging. I did approve this PR--so you can still merge if you like.

packages/cli/src/util/integration/types.ts
JakeUskoski: In our world, both fields can exist, so it's less a "one or the other" case. We've introduced this new concept of protocols, which the old integrations also use, and the protocols just-so-happen to do what "type" did as well, but better (product being multiple categories and categories having some bespoke concepts). Eventually we'll deprecate the type field and remove it entirely, but for now we have integrations where both values exist.

I'll keep this in mind for future situations, though. I like stronger types like this a lot 🙂

EndangeredMassa: Ah, my reading from the PR description was that they were exclusive.

Do the values need to exist in responses from the API? If not, you could restrict those values at the API serialization layer such that types like this suggestion could be used in clients.

PR #12736, opened 2024-12-11
[tests] remove unused vercel dev fixures

trek:
These fixtures no longer have associated tests and can be retired.


EndangeredMassa (approved):
We might want to add tests for these, but I don't know what the intent was here. I think it's fine to remove these.

PR #12737, opened 2024-12-11
[build-utils] Fix unit tests to include "unit.test.ts"

onsclom:
When `build-utils` was migrated to vitest, this glob was broken and stopped recognizing the file`unit.test.ts`.

Also, now that this test is ran with vitest, it gets ran with node 18, 20, and 22. There is a test in `unit.test.ts` that fails outside of Node 16 on ubuntu.


TooTallNate (approved):

packages/build-utils/test/unit.test.ts
```diff
@@ -671,6 +671,10 @@ it(
       console.log('Skipping test on macOS');
       return;
     }
+    if (process.version.split('.')[0] !== 'v16') {
```
TooTallNate, line 674: Btw, there is a `it.skipIf(…)` method we could use for this.

PR #12740, opened 2024-12-11
Add detection for new bun text lockfile

onsclom:
In version 1.2, bun will use a text `bun.lock` instead of binary `bun.lockb`: https://x.com/bunjavascript/status/1866879040424120625

The text lockfile can be used now on bun canary when using `--save-text-lockfile`.


TooTallNate (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -400,11 +408,10 @@ export async function scanParentDirs(
     : undefined;
 
   // Priority order is bun with yarn lock > yarn > pnpm > npm > bun
-  if (bunLockBin && hasYarnLock) {
+  if (bunLock && hasYarnLock) {
     cliType = 'bun';
     lockfilePath = bunLockPath;
-    // TODO: read "bun-lockfile-format-v0"
-    lockfileVersion = 0;
+    lockfileVersion = bunLockTextPath ? 1 : 0;
```
EndangeredMassa, line 414: **suggestion (soft-blocking):** read actual lockfile version

I don't think we should use `lockfileVersion` in this way. The version can still be `0` in the text-based format. This value could produce confusing results in the future.

It looks like this is being [used later](https://github.com/vercel/vercel/pull/12740/files#diff-e6d5f5db370224fa1fd8716c8e53a1de171a702e53e5a189d12347e05ee55334R1105) to output the lockfile name itself. This is brittle for future changes. We should find a more explicit way to track the lockfile name used.

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: Thanks, will make this change in a follow up PR!

PR #12747, opened 2024-12-12
[deploy] make deploy --target case sensitive

Zertsov:


packages/cli/src/util/parse-target.ts
```diff
@@ -23,11 +23,9 @@ export default function parseTarget<FlagName extends string>({
   }
 
   if (typeof targetFlagValue === 'string') {
-    const lowerCaseTarget = targetFlagValue.toLowerCase();
-    output.debug(`Setting target to ${lowerCaseTarget}`);
-    return lowerCaseTarget;
+    return targetFlagValue
```
TooTallNate: Needs a changeset. Also, I'm curious why the pre-commit formatter didn't run? (missing semicolon)

TooTallNate (changes requested):

packages/cli/src/util/parse-target.ts
Zertsov: yeah I just found that too 🤔 

I'll get the changeset made in the next commit because I forgot about that 🙃

trek (dismissed):

packages/cli/src/util/parse-target.ts
```diff
@@ -23,9 +23,7 @@ export default function parseTarget<FlagName extends string>({
   }
 
   if (typeof targetFlagValue === 'string') {
-    const lowerCaseTarget = targetFlagValue.toLowerCase();
```
trek, line 26: Why'd we remove the log?

Zertsov: Honestly I'm still recovering from sickness so I don't have a great reason. I can add it back

TooTallNate (approved):

PR #12749, opened 2024-12-12
Refactor build-utils usage to reuse lockfile path from scanParentDirs

onsclom:
We detected the lockfile file name in `detectPackageManager()` using just `cliType` and `lockfileVersion`. While this was fine for the previous package manager detection, we can't know whether the lockfile name is `bun.lock` or `bun.lockb` with just that info. This resulted in some [hacky logic](https://github.com/vercel/vercel/pull/12740/files#r1882591335) to support `bun.lock`, and this PR fixes that.

This refactors `detectPackageManagers()` to take in the already detected lockfile name (detected from `scanParentDirs` in our usage), which means we also need to argument drill through `getPathOverrideForPackageManager()` and `getPathForPackageManager()`, update all the builders using these functions, and update lots of tests.


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -1216,6 +1227,7 @@ export async function runCustomInstallCommand({
     env: spawnOpts?.env || {},
     packageJsonEngines: packageJson?.engines,
     turboSupportsCorepackHome,
+    detectedLockfile: lockfilePath ? path.basename(lockfilePath) : undefined,
```
erikareads: This logic likely needs to be the same in every place where we use it, can we extract this into a function for single point of truth?

erikareads: This way we can also test that we get the correct lockfile in various cases.

packages/hydrogen/src/build.ts
```diff
@@ -65,6 +66,7 @@ export const build: BuildV2 = async ({
     nodeVersion,
     env: spawnOpts.env || {},
     turboSupportsCorepackHome,
+    detectedLockfile: lockfilePath ? basename(lockfilePath) : undefined,
```
erikareads: It looks like that common logic should be exported from the build-utils package.

erikareads (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
erikareads: If there is a reason that each of these instances should vary separately from one another, I'm happy to remove the changes needed review.

onsclom: Moved this logic to `scanParentDirs()` so it only happens in one place!

trek (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -761,25 +768,24 @@ export function getEnvForPackageManager({
   env: { [x: string]: string | undefined };
   packageJsonEngines?: PackageJson.Engines;
   turboSupportsCorepackHome?: boolean | undefined;
+  detectedLockfile: string | undefined;
```
styfle, line 771: @onsclom Can we make this optional?

We generally don't allow breaking changes to `@vercel/build-utils` since it can be used unversioned by consumers.

onsclom: Looks like `getPathForPackageManager` is actually the problem. Making them both optional changes behavior however, so I think the better answer is just reverting this PR.

PR #12764, opened 2024-12-16
[next] Root Params

wyattjoh:
Added support for root params on pages powered by partial prerendering. This enables shell generation for pages with the root params. Now, when partial prerendering is enabled for a route with a root parameter, such as structure like this:

```
/[lang]/layout.tsx
/[lang]/page.tsx
```

The `lang` key will be included in the generated shell that's returned. For parameters that were prerendered (via `generateStaticParams`), they will already have more complete shells generated. This specific blocking route is only used when routes are served that have unknown root params.

### DIFF
--- a/.changeset/curvy-flowers-dance.md
+++ b/.changeset/curvy-flowers-dance.md
@@ -1,0 +1,6 @@
+---
+'@vercel/next': major
+'vercel': major
+---
+
+Added root params support for pages powered by partial prerendering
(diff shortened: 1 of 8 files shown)


.changeset/curvy-flowers-dance.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/next': major
```
ztanner: curious why this is a major? afaict this doesn't have any incompatible changes with older versions

ztanner (dismissed):

ztanner (dismissed):

ijjk (approved):

PR #12771, opened 2024-12-17
[cli] Update "follow-redirects" to v1.15.9 to avoid security vulnerability report

trek:
The actual dependency used is `http-proxy` and it's latest release only require `^1.0.0` so we need to override ourselves.


package.json
```diff
@@ -40,6 +40,11 @@
     "vite": "5.1.6",
     "vitest": "2.0.3"
   },
+  "pnpm": {
+    "overrides": {
+      "follow-redirects": "^1.15.6"
```
TooTallNate: This shouldn't be necessary since the semver range of the `http-proxy` package uses `^1.0.0`. We should run `pnpm update` and update the lockfile instead.

TooTallNate (changes requested):

package.json
trek: That's what I did on https://github.com/vercel/vercel/pull/12772 and it worked but for some reason this did not update the lockfile when I tried it here. I am juggling a few branches though, so possibly it was something unrelated

.changeset/great-fireants-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] overrides "follow-redirects" to at least 1.15.6 to avoid securi…
```
TooTallNate:
```suggestion
Update "follow-redirects" to v1.15.9 to avoid security vulnerability report
```

package.json
TooTallNate: Ok, [here](https://github.com/vercel/vercel/pull/12773)'s another take. I'm sure some combination of these three will get the vuln report closed 😄

TooTallNate (dismissed):

package.json
trek: A second attempt seems to have worked. I was probably just odd state or maybe `cd`d into a weird directory 🤷

onsclom (dismissed):

PR #12782, opened 2024-12-19
[next] Fix next data route replacing

ijjk:
This fixes our route generation from replacing assuming the dot isn't escaped which changed in https://github.com/vercel/next.js/pull/73850

Re-enables test disabled in https://linear.app/vercel/issue/ZERO-3045/unskip-failing-nextjs-test

### DIFF
--- a/packages/next/test/fixtures/00-i18n-gssp/index.test.js
+++ b/packages/next/test/fixtures/00-i18n-gssp/index.test.js
@@ -2,8 +2,6 @@ const path = require('path');
 const { deployAndTest } = require('../../utils');
 
 describe(`${__dirname.split(path.sep).pop()}`, () => {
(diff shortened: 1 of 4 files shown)


packages/next/test/fixtures/00-i18n-gssp/index.test.js
```diff
@@ -4,7 +4,7 @@ const { deployAndTest } = require('../../utils');
 describe(`${__dirname.split(path.sep).pop()}`, () => {
   // https://linear.app/vercel/issue/ZERO-3045/unskip-failing-nextjs-test
   // eslint-disable-next-line jest/no-disabled-tests
```
ijjk:
```suggestion
```
Will remove after test passes here

trek (approved):

PR #12783, opened 2024-12-19
[cli] add compile cache to improve startup performance

styfle:
This improves the startup of Vercel CLI by about 200ms on my MacBook Pro M3 running Node.js 22.

References

- https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
- https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#support-for-v8-compile-caching-in-node.js

### DIFF
--- a/packages/cli/scripts/build.mjs
+++ b/packages/cli/scripts/build.mjs
@@ -50,7 +50,8 @@ copyFileSync(
   new URL('get-latest-worker.js', distRoot)
 );
 
-writeFileSync(new URL('vc.js', distRoot),
+writeFileSync(
+  new URL('vc.js', distRoot),
   `#!/usr/bin/env node
 
 "use strict";
--- a/packages/cli/scripts/build.mjs
+++ b/packages/cli/scripts/build.mjs
@@ -49,7 +49,4 @@ copyFileSync(
   new URL('src/util/get-latest-version/get-latest-worker.js', repoRoot),
   new URL('get-latest-worker.js', distRoot)
 );
-copyFileSync(
-  new URL('src/vc.js', repoRoot),
-  new URL('vc.js', distRoot)
-);
+copyFileSync(new URL('src/vc.js', repoRoot), new URL('vc.js', distRoot));
--- a/packages/cli/scripts/build.mjs
+++ b/packages/cli/scripts/build.mjs
@@ -49,21 +49,7 @@ copyFileSync(
   new URL('src/util/get-latest-version/get-latest-worker.js', repoRoot),
   new URL('get-latest-worker.js', distRoot)
 );
-
-writeFileSync(
-  new URL('vc.js', distRoot),
-  `#!/usr/bin/env node
-
-"use strict";
-// This shim defers loading the real module until the compile cache is enabled.
-// https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
-try {
-  const { enableCompileCache } = require('node:module');
-  if (enableCompileCache) {
-    enableCompileCache();
-  }
-} catch {}
-require('./index.js');
-`,
-  'utf8'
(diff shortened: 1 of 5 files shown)


packages/cli/scripts/build.mjs
```diff
@@ -49,3 +49,21 @@ copyFileSync(
...
 );
+
+writeFileSync(
+  new URL('vc.js', distRoot),
+  `#!/usr/bin/env node
+
+"use strict";
+// This shim defers loading the real module until the compile cache is enabled.
+// https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+try {
```
TooTallNate: The try/catch is to make this a no-op on Node < 22?

packages/cli/scripts/build.mjs
```diff
@@ -49,3 +49,21 @@ copyFileSync(
   new URL('src/util/get-latest-version/get-latest-worker.js', repoRoot),
   new URL('get-latest-worker.js', distRoot)
 );
+
+writeFileSync(
+  new URL('vc.js', distRoot),
+  `#!/usr/bin/env node
```
TooTallNate: Any reason in particular why this is defined inline? How about we move it to a separate file which gets copied to `dist`?

packages/cli/scripts/build.mjs
styfle: Correct

packages/cli/scripts/build.mjs
styfle: I was going to do that but the code was small and unlikely to change.

We also write a file here so its not unprecedented:

https://github.com/vercel/vercel/blob/6150fbe18ef7fb4a6591a87a40c4452b2c9923ce/packages/cli/scripts/build.mjs#L8-L14

If you feel strongly I can move it to a separate file. I'm just not sure which directory it should live in.

EndangeredMassa: In that other case, we interpolate a value, making the inline code an easier solution than providing that value another way. In this case, we don't need that.

I'd prefer this be moved to a separate file. I do think we'd end up changing it in the future, at least a bit. For example, I'd want to add some debug logging in the catch clause, but wouldn't block this PR on that.

styfle: Sounds good. I changed this to a copy in [6c9f653](https://github.com/vercel/vercel/pull/12783/commits/6c9f653cc6df4766b7414524cc0f9f9d8c7bc738)

EndangeredMassa (approved):
The 5ms gain doesn't seem worth the work, but now that that work is done, I'm fine merging it in.

styfle:
I just benchmarked it again running 4 times each and the difference is like 300ms on the first run and then around 60ms.

## Before
```
▲ time ./packages/cli/dist/index.js --version
0.23s user 0.09s system 49% cpu 0.646 total
▲ time ./packages/cli/dist/index.js --version
0.18s user 0.03s system 100% cpu 0.209 total
▲ time ./packages/cli/dist/index.js --version
0.18s user 0.02s system 102% cpu 0.195 total
▲ time ./packages/cli/dist/index.js --version
0.18s user 0.02s system 102% cpu 0.197 total
```

## After
```
▲ time ./packages/cli/dist/vc.js --version   
0.17s user 0.06s system 65% cpu 0.353 total
▲ time ./packages/cli/dist/vc.js --version   
0.12s user 0.03s system 100% cpu 0.148 total
▲ time ./packages/cli/dist/vc.js --version
0.12s user 0.02s system 103% cpu 0.135 total
▲ time ./packages/cli/dist/vc.js --version
0.12s user 0.02s system 102% cpu 0.136 total
```

blg666 (approved):

PR #12817, opened 2025-01-02
[cli] update handleError to use output utilities

erikareads:
Previously #12010 I created `printError` to begin changing the API of `handleError` to use `output` instead of directly using `console.error`. Since we're now importing a usable `output` object directly, `handleError` can be updated without changing its call signature. This PR moves the updates from `printError` into `handleError` and removes the now redundant `printError`.


packages/cli/src/util/handle-error.ts
```diff
@@ -1,11 +1,9 @@
 import bytes from 'bytes';
-import info from './output/info';
-import errorOutput from './output/error';
 import type { APIError } from './errors-ts';
 import { getCommandName } from './pkg-name';
 import output from '../output-manager';
 
-export function printError(error: unknown) {
+export default function handleError(error: unknown) {
```
TooTallNate, line 6: Can we drop the `default` while we're at it?

erikareads: I'm going to do a fast-follow PR renaming `handleError` to `printError` since that better represents what the function does, since I'm going to need to change imports on that PR anyway, it'll be easier to drop `default` on that PR.

TooTallNate (approved):

PR #12823, opened 2025-01-03
[cli] centralize printError import location

erikareads:
While working on https://github.com/vercel/vercel/pull/12821, I found that we had two import paths for `printError`, both `print-error.ts` and `error.ts` were exporting the function. This PR unifies all imports to come from `error.ts`.

### DIFF
--- a/.changeset/healthy-lies-rush.md
+++ b/.changeset/healthy-lies-rush.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] centralize printError import location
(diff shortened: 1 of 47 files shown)


trek (approved):

TooTallNate (commented):
I'm not seeing it in the diff, but does this also remove the double export?

erikareads:
> I'm not seeing it in the diff, but does this also remove the double export?

No, I like doing one easy-to-review thing in each PR. I've created https://github.com/vercel/vercel/pull/12824 to remove the double export by moving `printError` into `error.ts`.

PR #12833, opened 2025-01-06
[cli] sort imports in build/index.ts

erikareads:
Breaks the imports in `build/index.ts` into three sorted groups: external dependencies, monorepo dependencies, and other modules in `package/cli`. 

I did this PR as a standalone on one module so that we can discuss whether we:

- Want to sort our imports?
- Want to do an automated sweep across other modules in `package/cli` or the whole repo?
- Want to enforce import sorting with the formatter/linter?

This particular sort was done using a manual invocation of the Zed 'organize imports' code action.

### DIFF
--- a/.changeset/slimy-horses-admire.md
+++ b/.changeset/slimy-horses-admire.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] sort imports in build/index.ts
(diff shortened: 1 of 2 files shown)


TooTallNate (approved):
I'm down for this, but yes please let's automate it.

onsclom (approved):

EndangeredMassa:
I'm mildly against sorting imports. I think that a thoughtful human sort is better than a deterministic/automatic sort. The problem is the non-automatic sort is often not thoughtful. So, I'm fine either way.

PR #12834, opened 2025-01-06
[cli] standardize uses of confirm on confirm.ts

erikareads:
Prior to this PR, we had a mixed usage of the `confirm` interactive input component. Some came through `client.input.confirm` and some came through the function `confirm` that wraps it. This PR standardizes on the function `confirm` since that was in use in more places and restricts the options that a call site can use.

This PR doesn't prevent `client.input.confirm` from being called directly in the future. It may be worth considering an `input-manager` similar to the previous `output-manager` refactor.

### DIFF
--- a/.changeset/lovely-bobcats-mix.md
+++ b/.changeset/lovely-bobcats-mix.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] standardizes uses of confirm on confirm.ts
--- a/packages/cli/src/commands/integration/add.ts
+++ b/packages/cli/src/commands/integration/add.ts
@@ -23,6 +23,7 @@ import { fetchInstallations } from '../../util/integration/fetch-installations';
 import { fetchIntegration } from '../../util/integration/fetch-integration';
 import output from '../../output-manager';
 import { IntegrationAddTelemetryClient } from '../../util/telemetry/commands/integration/add';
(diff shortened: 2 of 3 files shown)


onsclom (approved):
I slightly prefer going the other way: converting `await confirm()` usages into `await client.input.confirm()`. Then we'd be able do delete `confirm.ts` and better match how the rest of the codebase seems to prefer `client.input.confirm`. Maybe we also change the implementation to restrict the options.

However, I think making things more consistent is helpful regardless, so approving this!

erikareads:
> I slightly prefer going the other way: converting `await confirm()` usages into `await client.input.confirm()`. Then we'd be able do delete `confirm.ts` and better match how the rest of the codebase seems to prefer `client.input.confirm`. Maybe we also change the implementation to restrict the options.
> 
> However, I think making things more consistent is helpful regardless, so approving this!

With this PR, we make everything consistent, which makes it easier to switch in the future. With your comment in mind, I'm reminded that `client.input.confirm` is still our own wrapper of the `confirm` component, and we can move the restrictive call signature to that definition. 

I'll do a fast-follow PR that migrates everything to that one location, thus removing the confusion of which `confirm` method/function to use for people wanting to add `confirm` logic to their commands.

EndangeredMassa (approved):

blg666 (approved):

PR #12852, opened 2025-01-13
[build-utils] support pnpm@10

EndangeredMassa:
Updates `build-utils` to use `pnpm@10` for v9 `pnpm-lock.yaml` lockfiles instead of `pnpm@9`. v7 pnpm lock files should still use `pnpm@9`.


packages/build-utils/test/unit.get-env-for-package-manager.test.ts
```diff
@@ -197,12 +197,12 @@ describe('Test `getEnvForPackageManager()`', () => {
         consoleWarnOutput: null,
       },
       {
-        name: 'should set path if pnpm 8 is detected',
+        name: 'should set path if pnpm 9 is detected',
```
onsclom, line 200: This was incorrectly named before. Fixed naming now since lockfile v7.0 should setup path for pnpm@9.

onsclom:
> Detected `pnpm-lock.yaml` version 9 generated by pnpm@10.x

Remaining test failures are projects trying to use pnpm@10, and [it is not yet rolled out in the build container](https://github.com/vercel/api/pull/36770).

onsclom (approved):

PR #12690, opened 2024-12-03
[cli] bump tsup and lint-staged

erikareads:

PR #12854, opened 2025-01-14
bump tsup and lint-staged and esbuild

erikareads:
#12690 was having issues with CI that seemed unrelated to the PR. This is a fresh PR to get those changes through.

Closes #12690. 

This PR bumps tsup, lint-staged, and esbuild in order to get ahead of security warnings related to a transitive dependency `braces`. 

One of these updates removes an unpinned dependency on `commander` that the `dot-packer` script depended on, so this PR vendors the parts of that script that our codebase uses.


trek (approved):

EndangeredMassa (approved):

onsclom (approved):

packages/cli/scripts/dot-packer
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
```
TooTallNate, line 1: I didn't really understand this part. Why not add `dot-packer` as a dev dependency of CLI?

TooTallNate (changes requested):

packages/cli/scripts/dot-packer
erikareads: We could do that, however we only use a tiny portion of dot-packer's logic, which is easy to vendor. 

The [dot-packer](https://www.npmjs.com/package/dot-packer?activeTab=readme) package, which I didn't realize existed separately (we've been using it transitively through `dot` until now), hasn't been updated in 12 years, and uses an ancient version of commander for no benefit to us. 

Vendoring the script allows us to avoid more outdated dependencies while keeping the logic we're using.

PR #12856, opened 2025-01-14
[cli] Allows warnings for `ls --update-required` to show for deprecated versions

trek:
Currently we only show a warning if a runtime version's `discontinueDate` has passed, but customers should be able to run this before that date to see what they need to change. Setting the date for the soon-to-be removed node@16 to the date from the [changelog](https://vercel.com/changelog/node-js-16-deprecation-2gFqvRhzEw1gA6wvTCrKy0)

### DIFF
--- a/packages/cli/src/commands/project/list.ts
+++ b/packages/cli/src/commands/project/list.ts
@@ -81,8 +81,8 @@ export default async function list(
 
     for (const nodeVersion of NODE_VERSIONS) {
       if (
-        nodeVersion.discontinueDate &&
-        nodeVersion.discontinueDate.valueOf() > Date.now()
+        nodeVersion.deprecationDate &&
+        nodeVersion.deprecationDate.valueOf() > Date.now()
       ) {
         upcomingDeprecationVersionsList.push(nodeVersion.range);
       }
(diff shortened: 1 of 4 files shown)


onsclom (dismissed):

packages/cli/src/commands/project/list.ts
```diff
@@ -81,8 +81,8 @@ export default async function list(
 
     for (const nodeVersion of NODE_VERSIONS) {
       if (
-        nodeVersion.discontinueDate &&
-        nodeVersion.discontinueDate.valueOf() > Date.now()
+        nodeVersion.deprecationDate &&
+        nodeVersion.deprecationDate.valueOf() > Date.now()
```
EndangeredMassa, line 85: **suggestion (blocking):** check both dates

I think we want to check for either date being in the past. Or, we should add `deprecationDate` to all currently discontinued versions.

EndangeredMassa (changes requested):

packages/cli/src/commands/project/list.ts
trek: Looking more at this feature as it currently exists and it makes no sense. It only works as a warning _after_ your projects already don't run? Gonna rethink this entire thing.

EndangeredMassa: Checking `discontinueDate` means that you already can't deploy, sure. A user still wants to get a list of those, though.

Checking `deprecationDate` means that you can know ahead of time.

I think we want to check both.

PR #12860, opened 2025-01-15
Bump minimum node version to 18

trek:
Drop support for node-16 in CLI tests and development.


erikareads (dismissed):

.node_version
```diff
@@ -1 +1 @@
-v16.20.2
+v18.0.0
```
trek: Not sure which value we actually want here! Latest 18.x?

EndangeredMassa: We might as well match what we use in build container: `18.20.6`

PR #12862, opened 2025-01-15
feat: improve metrics of next

marc-vercel:


packages/next/src/index.ts
```diff
@@ -2341,6 +2353,10 @@ export const build: BuildV2 = async buildOptions => {
       ...staticDirectoryFiles,
       ...privateOutputs.files,
     },
+    metrics: {
+      install: installExecutionTime,
+      build: buildExecutionTime,
+    },
```
EndangeredMassa: We should keep in mind future metrics or details we want to capture about builds when changing this interface. For example, we'll want to capture which package manager was used and how successful it was for the install step.

We could end up with something like:

```
metrics: {
  install: {
    duration: installExecutionTime,
    packageManager,
    isSatisfiedByBuildCache,
  }
}
```

Or we add that data elsewhere later.

I suggest that the interface here start with:

```suggestion
    metrics: {
      install: {
        duration: installExecutionTime
      },
      build: {
        duration: buildExecutionTime
      }
    },
```

marc-vercel: Thanks for the reply @EndangeredMassa! I will do that.

Also, I see that the test [test-e2e (examples, 3, ubuntu-latest, Node v16)](https://github.com/vercel/vercel/actions/runs/12908939636/job/35998529318?pr=12862#logs) keeps failing, and it does not seem a flaky test. Do you know which may be the cause? Afaik I've not change anything from `ionic-react` and I'm struggling a bit to find the root cause.

onsclom: Sorry about that `ionic-react` failure, I was working on that. Started a rerun and it should succeed now!

PR #12864, opened 2025-01-15
[cli] update warning in ls --update-required for clarity

erikareads:
Closes #12856.

The existing logic assumes that if a `discontinueDate` has been set for a given version of node, that version of node must be deprecated. It was also easy to mistake the previous code as displaying the warning only _after_ the discontinue date, which it never did.

This PR:

- Clarifies the language communicating the deprecation to users. It's not that the node version _will be_ deprecated. The node version is _already_ deprecated and is at risk of being discontinued.
- Adds an identifier to the logic checking for a deprecated node version, making it clear that the date check is looking into the future.
- Adds a comment clarifying the logic further, and documents the assumption that a `discontinueDate` means that a node version is deprecated.
- Adds a further comment noting that if this assumption is invalid, additional logic may be required. Specifically, in the case where we have deprecated a node version, but don't yet know its discontinue date.

### DIFF
--- a/packages/cli/src/commands/project/list.ts
+++ b/packages/cli/src/commands/project/list.ts
@@ -80,10 +80,15 @@ export default async function list(
     const upcomingDeprecationVersionsList = [];
 
     for (const nodeVersion of NODE_VERSIONS) {
-      if (
+      // This logic presumes that adding a discontinue date for a node version
+      // implies that the version is deprecated.
+      // If there are scenarios where we will be deprecating a version
+      // without knowing the discontinue date
+      // Then we'll want to consider adding a `deprecationDate` field to the nodeVersion type
+      const upcomingDiscontinueDate =
         nodeVersion.discontinueDate &&
-        nodeVersion.discontinueDate.valueOf() > Date.now()
-      ) {
+        nodeVersion.discontinueDate.valueOf() > Date.now();
+      if (upcomingDiscontinueDate) {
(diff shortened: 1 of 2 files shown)


trek (dismissed):

EndangeredMassa (dismissed):

erikareads:
I accidentally pushed additional changes to this PR which I used a force push to revert, so this will need to be approved again.

packages/cli/src/commands/project/list.ts
```diff
@@ -80,18 +80,23 @@ export default async function list(
...
         upcomingDeprecationVersionsList.push(nodeVersion.range);
       }
     }
 
     output.warn(
-      `The following Node.js versions will be deprecated soon: ${upcomingDeprecationVersionsList.join(
+      `The following Node.js versions are deprecated: ${upcomingDeprecationVersionsList.join(
         ', '
-      )}. Please upgrade your projects immediately.`
+      )} and will be discontinued soon. Please upgrade your projects immediately.`
```
EndangeredMassa, line 99: **suggestion (non-blocking):** mention dates

Ideally we'd have the versions and their dates all listed here.‏

EndangeredMassa (approved):

packages/cli/src/commands/project/list.ts
erikareads: We do have those dates, I think that's a worthwhile addition.

PR #12869, opened 2025-01-16
[build-utils] Remove support for node@16 behaviors

trek:
# PLEASE DO NOT MERGE UNTIL 2025-02-28

Removes support or Node 16 in built-utils. Because this library is used outside the CLI and we have no mechanism to pin older versions, please merge this (which triggers a semver major release) only after 2025-02-28.


trek:
This got moved into other cards.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -668,22 +668,13 @@ export async function runNpmInstall(
...
     if (cliType === 'npm') {
       opts.prettyCommand = 'npm install';
       commandArgs = args
         .filter(a => a !== '--prefer-offline')
         .concat(['install', '--no-audit', '--unsafe-perm']);
-      if (
-        isPotentiallyBrokenNpm &&
-        spawnOpts?.env?.VERCEL_NPM_LEGACY_PEER_DEPS === '1'
-      ) {
+      if (spawnOpts?.env?.VERCEL_NPM_LEGACY_PEER_DEPS === '1') {
```
EndangeredMassa: **suggestion (soft-blocking):** remove



We can get rid of this block. It was only used for Node 16 before.

We can also get rid of the references to this env var and associated feature flags in API.

EndangeredMassa (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -721,7 +703,6 @@ export async function runNpmInstall(
         'eresolve-report.txt'
       );
       if (
-        isPotentiallyBrokenNpm &&
         !commandArgs.includes('--legacy-peer-deps') &&
         fs.existsSync(potentialErrorPath)
       ) {
```
EndangeredMassa: **suggestion (non-blocking):** ‏I think this whole check can go as well, which I believe removes the need for the full try-catch.

EndangeredMassa (dismissed):

EndangeredMassa (approved):

PR #12872, opened 2025-01-16
[node] Split `build()`, `prepareCache()` and `startDevServer()` into separate files

TooTallNate:
Small tidiness refactor. The old `src/index.ts` was getting a bit unweildy. No functional changes.


trek (dismissed):

EndangeredMassa (approved):
tested `vc build` and `vc dev`, both of which seemed to still work

PR #12873, opened 2025-01-16
[build-utils] Add `useWebApi` property to `NodejsLambda` class

TooTallNate:
Currently Node.js functions can opt-in to the Web API syntax by exporting named HTTP methods as request handlers (`export const GET = () => new Response()`), but there is no way to export a default catch-all handler.

For backwards compat reasons, we can not change the default `export default` as a Node.js HTTP handler syntax, so this new opt-in property `useWebApi` will enable Web API syntax for the default export.

This is a prerequisite to making React Router v7+ work on Vercel in a sane way.

### DIFF
--- a/.changeset/chilled-worms-hunt.md
+++ b/.changeset/chilled-worms-hunt.md
@@ -1,0 +1,5 @@
+---
+'@vercel/build-utils': minor
+---
+
+Add `shouldUseWebApi` property to `NodejsLambda` class
--- a/.changeset/chilled-worms-hunt.md
+++ b/.changeset/chilled-worms-hunt.md
@@ -2,4 +2,4 @@
 '@vercel/build-utils': minor
 ---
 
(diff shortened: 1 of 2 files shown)


erikareads:
How does `useWebApi` get used? 

If I understand correctly, we support a feature that allows for these method named exports (`GET`, etc.). But that feature is not enabled by default, and can't be for backwards compatibility reasons. 

This PR adds the `useWebApi` property to the `NodejsLambda` class, which gets detected automatically if set? Or will additional work need to be done in order to use this property?

TooTallNate:
@erikareads The intention is for a Builder / Build Output API producer to set this property when the user code is meant to be using the Web API syntax. This will be used in `@vercel/remix-builder` to allow for the entrypoint to use a consistent isomorphic syntax for both Node.js and Edge functions.

There are related backend changes that will use this property when set to make the things happen.

EndangeredMassa (approved):

PR #12879, opened 2025-01-17
[examples] Update ionic-react example

onsclom:
Updating ionic-react to unpin from node 16.

Since this update changed the build tool from webpack -> vite, the default outdir changed. Changed it back to `build` in the vite config (see comments below).

Deployment: https://ionic-react-q1fqkl6qt-austin-merricks-projects.vercel.app/


examples/ionic-react/vite.config.ts
```diff
@@ -0,0 +1,21 @@
...
+    legacy()
+  ],
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: './src/setupTests.ts',
+  },
+  build: {
+    outDir: 'build'
+  }
```
onsclom, line 20: Added this to match the existing `outDir` in `frameworks.ts`. Instead, the `ionic-react` `frameworks.ts` entry can be update to match vite defaults (like `dist` as output dir), but then we might break old webpack `ionic-react` projects.

onsclom: Maybe it would be better to take out the `ionic-react` framework entirely and let these projects be detected as either webpack or vite projects?

erikareads (approved):

EndangeredMassa (approved):

examples/ionic-react/vite.config.ts
erikareads: Does that mean that newly generated `ionic-react` projects will not be detected by `frameworks.ts`?

EndangeredMassa: We should update the framework.ts entries so that older `ionic-react` versions (that use the old `outDir`) can still work but new ones are detected properly. Maybe we add a second entry for newer `ionic-react` package versions. I could pair on it next week, if you like.

trek (approved):

examples/ionic-react/vite.config.ts
onsclom: Good idea! Will add detection for vite `ionic-react` projects in a new PR.

PR #12880, opened 2025-01-17
[frameworks][examples] Fix vite ionic-react framework detection

onsclom:
New `ionic-react` switched from using webpack to vite, but the `ionic-react` `frameworks.ts` detection still assumes it will use webpack.

This changes `vite` to supersede `ionic-react` so that `vite` `ionic-react` projects are detected as `vite` instead.

### DIFF
--- a/examples/__tests__/integration/ionic-react.test.ts
+++ b/examples/__tests__/integration/ionic-react.test.ts
@@ -1,4 +1,5 @@
 import { deployExample } from '../test-utils';
-it('[examples] should deploy ionic-react', async () => {
+// TODO: unskip once example is manually changed to `vite`
+it.skip('[examples] should deploy ionic-react', async () => {
   await deployExample('ionic-react');
 });
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -1820,6 +1820,7 @@ export const frameworks = [
       'Vite is a new breed of frontend build tool that significantly improves the frontend development experience.',
     description: 'A Vue.js app, created with Vite.',
     website: 'https://vitejs.dev',
+    supersedes: ['ionic-react'],
     envPrefix: 'VITE_',
     detectors: {
       every: [
--- a/packages/frameworks/test/frameworks.unit.test.ts
+++ b/packages/frameworks/test/frameworks.unit.test.ts
@@ -215,7 +215,13 @@ async function getDeployment(host: string) {
 }
 
 describe('frameworks', () => {
-  const skipExamples = ['sanity-v3', 'solidstart', 'dojo', 'scully'];
+  const skipExamples = [
+    'sanity-v3',
+    'solidstart',
+    'dojo',
+    'scully',
+    'ionic-react',
+  ];
 
   it('ensure there is an example for every framework', async () => {
     const root = join(__dirname, '..', '..', '..');
--- a/packages/frameworks/test/frameworks.unit.test.ts
+++ b/packages/frameworks/test/frameworks.unit.test.ts
@@ -215,7 +215,7 @@ async function getDeployment(host: string) {
 }
 
 describe('frameworks', () => {
-  const skipExamples = ['sanity-v3', 'solidstart'];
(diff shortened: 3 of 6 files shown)


onsclom:
Follow up from https://github.com/vercel/vercel/pull/12879

EndangeredMassa:
Updated from main to pull in these changes: https://github.com/vercel/vercel/pull/12890

packages/frameworks/src/frameworks.ts
```diff
@@ -1820,6 +1820,7 @@ export const frameworks = [
       'Vite is a new breed of frontend build tool that significantly improves the frontend development experience.',
     description: 'A Vue.js app, created with Vite.',
     website: 'https://vitejs.dev',
+    supersedes: ['ionic-react'],
```
EndangeredMassa, line 1823: **suggestion (blocking):** ‏Does this fail because the output dir is `dist` instead of `build`? Can we add a test that this behavior is correct?

EndangeredMassa (changes requested):

packages/frameworks/src/frameworks.ts
onsclom: Newly initiated `ionic-react` do not currently deploy with zero config, and this fixes them. These newly inited projects use vite and therefore DO use `dist` instead of `build`.

The currently existing e2e tests run the fresh inited ionic react project with vite settings and it succeeds https://github.com/vercel/vercel/actions/runs/12892527405/job/35948586479?pr=12880#step:8:128 ([deployment](https://vercel.com/zero-conf-vtest314/ionic-react/5mYjoNxEMnG9juaCiK8DeAc8Tvbb)). Is that test sufficient or were you thinking about a different kind of test?

EndangeredMassa: Ah, thanks for that clarification.

That test was passing before this change. I was hoping to see a test that would only pass because of this change.

What are these more recent build failures on [that project](https://vercel.com/zero-conf-vtest314/ionic-react/deployments)?

onsclom: Whoops, my bad. Part of making this PR included switching that example to `vite` manually (since we don't rerun framework detection automatically), but that makes it fail for other PRs if we don't immediately merge this.

Switched it back to ionic react so other PRs will succeed now. Will add a test that shows `supersedes` prioritizes `vite` over `ionic-react` later.

packages/frameworks/test/frameworks.unit.test.ts
```diff
@@ -215,7 +215,13 @@ async function getDeployment(host: string) {
...
 
 describe('frameworks', () => {
-  const skipExamples = ['sanity-v3', 'solidstart', 'dojo', 'scully'];
+  const skipExamples = [
+    'sanity-v3',
+    'solidstart',
+    'dojo',
+    'scully',
+    'ionic-react',
+  ];
```
onsclom, line 224: Temporarily adding `ionic-react` to skipped examples to avoid ionic-react failures.

After this gets merged I will manually set the `ionic-react` project to `vite`, then I'll make another PR to unskip the example.

examples/__tests__/integration/ionic-react.test.ts
```diff
@@ -1,4 +1,5 @@
 import { deployExample } from '../test-utils';
-it('[examples] should deploy ionic-react', async () => {
+// TODO: unskip once example is manually changed to `vite`
+it.skip('[examples] should deploy ionic-react', async () => {
```
onsclom, line 3: Same as https://github.com/vercel/vercel/pull/12880/files#r1925735759

packages/frameworks/src/frameworks.ts
onsclom: Added that test now https://github.com/vercel/vercel/pull/12880/files#diff-b9b88879323adf73e47827b03f1d5593292326b4647e78b71f85951f2a3172f8!

EndangeredMassa (approved):

trek (approved):

PR #12883, opened 2025-01-18
[build-utils] add state property to NodeVersion 

erikareads:
# Overview

This PR adds a `state` property to the `NodeVersion` type. One of `'active' | 'deprecated' | 'discontinued'`. This changes `NodeVersion` from an interface with optional fields to a tagged union with `state` as a discriminator. 

Now `discontinueDate` is a required field only on `DeprecatedNodeVersion` and not a field on the other two alternatives.

## Goal

The `state` property makes it clear at a glance what state a given NodeVersion in `NODE_VERSIONS` is in. 

Previously, code assumed that if a `discontinueDate` had been set, a version was deprecated, but this assumption was not explicit in the code.

This PR makes that assumption explicit.

## Work Done

- Updated the `NodeVersion` type to be a type union of tagged alternatives, one of each state.  
- Updated `NODE_VERSIONS` to include the `state` property.
- Updated the helper functions `node-version.ts` to account for the `state` property.
- Updated some language in the warnings printed by `node-version.ts` to use our newly considered terminology around deprecation and discontinuation.
- Updated project list `--updated-required` to use the `state` property.
- Updated `RubyVersion` which depended directly on `NodeVersion` being an extensible interface instead of a type union.

## Questions

- Do we want another field name besides `state`? If so, what?
- Do we want `RubyVersion` to still be dependent on `NodeVersion`?
- Is there any reason to include `discontinueDate` for `discontinue` state Node Versions?


packages/ruby/src/install-ruby.ts
```diff
@@ -2,28 +2,48 @@ import execa from 'execa';
...
+  Meta,
+  debug,
+  NowBuildError,
+  ActiveNodeVersion,
+  DeprecatedNodeVersion,
+  DiscontinuedNodeVersion,
+} from '@vercel/build-utils';
 
-interface RubyVersion extends NodeVersion {
+interface ActiveRubyVersion extends ActiveNodeVersion {
```
EndangeredMassa: > Do we want RubyVersion to still be dependent on NodeVersion?

This looks like a signal that these types aren't really node and ruby specific. Instead, the are Major-defined and Minor-defined types. Maybe we should rename them to be more generic.

packages/build-utils/src/fs/node-version.ts
```diff
@@ -7,38 +7,44 @@ import debug from '../debug';
...
     runtime: 'nodejs16.x',
+    state: 'deprecated',
     discontinueDate: new Date('2025-01-31'),
   },
   {
     major: 14,
     range: '14.x',
     runtime: 'nodejs14.x',
-    discontinueDate: new Date('2023-08-15'),
+    state: 'discontinued',
```
EndangeredMassa: > Is there any reason to include `discontinueDate` for `discontinue` state Node Versions?

**In code:** No. There's no functional value.

**In data:** Yes. It represents the historical record.

**Here:** I would consider this data that happens to live in code. If we stored this in a database instead, would we be debating whether or not include the discontinue date? I think we'd assume it should be included.

packages/build-utils/src/types.ts
```diff
@@ -324,17 +324,34 @@ export interface PackageJson {
...
   /** major version number: 18 */
   major: number;
   /** major version range: "18.x" */
   range: string;
   /** runtime descriptor: "nodejs18.x" */
   runtime: string;
+}
+
+export interface ActiveNodeVersion extends BaseNodeVersion {
+  state: 'active';
```
EndangeredMassa: > Do we want another field name besides state? If so, what?

I can't think of anything right now.

packages/build-utils/src/fs/node-version.ts
```diff
@@ -138,21 +144,31 @@ export async function getSupportedNodeVersion(
...
+        selection.range
+      } is discontinued. Deployments will fail to build. ${getHint(isAuto)}`
+    );
   }
 
   return selection;
 }
 
-function isDiscontinued({ discontinueDate }: NodeVersion): boolean {
+function isDiscontinued(nodeVersion: NodeVersion): boolean {
```
EndangeredMassa: **question (non-blocking):** Is there a better way to model the fact that a state of `deprecated` with a `discontinueDate` in the past is essentially the same state of `discontinued`?

Ideally, we wouldn't need logic to check that `nodeVersion.state === 'deprecated' && nodeVersion.discontinueDate.getTime() <= today` because the moment those dates crossed paths, the system would udpate the `state` to `discontinued`. If we were storing this data in a database, we could set up a cron or other trigger to make that happen.

Because we don't have a database for this data (and I'm not suggesting adding one with API calls to access it), the state situation gets messy.‏

The best way to accomplish something similar here (in my opinion) is to use an object with a `isDiscontinued()` function or an explicit `state` getter that would literally translate "a state of `deprecated` with a `discontinueDate` in the past" into a state of `discontinued.

EndangeredMassa (dismissed):
I left comments, but nothing blocking.

packages/build-utils/src/fs/node-version.ts
erikareads: The question then is:

Is it better to keep the data written directly in code or to trust that the data is retrievable through git history?

I tend to err on the side of including more context directly in the code, so I think it _is_ worth adding the dates to the `discontinued` type variant.

EndangeredMassa: I agree. I think we should keep those dates in place.

packages/build-utils/src/fs/node-version.ts
erikareads: This awkwardness around dates is something I've been debating since I started considering this change. 

The problem here is that we have CLI users that won't be updating the instant we mark something as discontinued, so we need to check the date for users that have a version of `build-utils` that isn't yet updated to mark a version as discontinued.

There are a few alternatives we could use here:

## 1 Call an API to check latest up-to-date information about a Node Version

Upsides: 

- This gives us centralized control of when a version is deprecated or discontinued, across all users regardless of CLI version (post API change)
- Allows us to dispense with date checking entirely.

Downsides:

- Adds an extra API call to a fairly elemental operation, likely to balloon to a significant number of API calls if we don't cache this value appropriately in the same API change.

## 2 Add a state property to `NodeVersion` 

Upsides:

- Gives us a clear indicator in the code, where the data lives, when we update the state of a Node Version
- Makes explicit any assumptions about version deprecation
- Allows us to use type alternatives that eliminate unneeded fields in the `active` case

Downsides:

- We still need to care about the passage of time
- This information has to be remembered and updated as part of the deprecation and discontinuation processes
- This information can drift relative to what's supported on the Vercel platform

## 3 Exported `versionState` function that returns one of `'active' | 'deprecated' | 'discontinued'`

Upsides:

- Allows us to isolate the date calculation in one place
- Reflects the true reality that true state of version depends on the current date
- Allows elimination of some booleans

Downsides:

- Moves the logic about current state away from the `NODE_VERSIONS` constant

## 4 Do 2 _and_ 3

Upsides:

- Upsides of both 2 and 3

Downsides:

- Removes the single point of truth from `NODE_VERSIONS` constant
- Could allow people to directly access the `state` property, thereby potentially incorrectly getting the current state if the date is late enough

## 5 Turn `NodeVersion` into a class with a `get` implementation for `state`

Upsides:

- Upsides of both 2 and 3
- Remove ambiguity created by 4

Downsides:

- There may be consequences for turning NodeVersion into a Class


There is another scenario I hadn't considered yet, and I'm not certain we handle their case gracefully: a user who has a sufficiently old version of `build-utils` that their version hasn't even marked a Node Version as _deprecated_ when that version is discontinued. 

Do we have appropriate error handling and communication for this case?

EndangeredMassa (approved):

PR #12894, opened 2025-01-22
Enable tracing within the cli and builder framework

LukeSheard:
To enable open telemetry and metric aggregation on our builder layers, I've added a port of the [next.js](https://github.com/vercel/next.js/blob/5d7713f833d6fea6e599d49ea7d7c14307e64570/packages/next/src/trace/trace.ts) tracing implementation which is compatible with our open telemetry ingestion.

This allows us to construct build spans across the CLI & zero-config builder framework and we'll eventually be able to reconstruct these into telemetry spans on the backend.

### DIFF
--- a/packages/build-utils/src/trace/trace.ts
+++ b/packages/build-utils/src/trace/trace.ts
@@ -71,7 +71,8 @@ export class Span {
 
     this.status = 'stopped';
 
-    const duration = Date.now() - this._start;
+    // durations are reported in microseconds
+    const duration = (Date.now() - this._start) * 1000;
 
     const traceEvent: TraceEvent = {
       name: this.name,
--- a/packages/build-utils/src/trace/trace.ts
+++ b/packages/build-utils/src/trace/trace.ts
@@ -13,7 +13,6 @@ export type TraceEvent = {
 };
 
 export type Reporter = {
-  flushAll: (opts?: { end: boolean }) => Promise<void> | void;
   report: (event: TraceEvent) => void;
 };
 
--- a/packages/build-utils/src/trace/trace.ts
+++ b/packages/build-utils/src/trace/trace.ts
@@ -1,5 +1,7 @@
 import { randomUUID } from 'node:crypto';
 
+const NUM_OF_MICROSEC_IN_NANOSEC = BigInt('1000');
+
 export type SpanId = string;
 
 export type TraceEvent = {
@@ -38,7 +40,8 @@ export class Span {
   private status: 'started' | 'stopped';
 
   // Number of nanoseconds since epoch.
-  private _start: number;
+  private _start: bigint;
+  private now: number;
 
   private _reporter: Reporter | undefined;
 
@@ -59,8 +62,15 @@ export class Span {
     this.status = 'started';
 
     this.id = randomUUID();
-    this._start = Date.now();
     this._reporter = reporter;
+
+    // hrtime cannot be used to reconstruct tracing span's actual start time
+    // since it does not have relation to clock time:
+    // `These times are relative to an arbitrary time in the past, and not related to the time of day and therefore not subject to clock drift`
+    // https://nodejs.org/api/process.html#processhrtimetime
+    // Capturing current datetime as additional metadata for external reconstruction.
+    this.now = Date.now();
+    this._start = process.hrtime.bigint();
   }
 
   stop() {
@@ -70,17 +80,19 @@ export class Span {
 
     this.status = 'stopped';
 
-    // durations are reported in microseconds
-    const duration = (Date.now() - this._start) * 1000;
+    const end = process.hrtime.bigint();
+    const duration = Number((end - this._start) / NUM_OF_MICROSEC_IN_NANOSEC);
+
+    const timestamp = Number(this._start / NUM_OF_MICROSEC_IN_NANOSEC);
 
     const traceEvent: TraceEvent = {
       name: this.name,
-      duration: Number(duration),
-      timestamp: this._start,
+      duration,
+      timestamp,
       id: this.id,
       parentId: this.parentId,
       tags: this.attrs,
-      startTime: this._start,
+      startTime: this.now,
     };
 
     if (this._reporter) {
--- a/packages/cli/src/commands/build/index.ts
+++ b/packages/cli/src/commands/build/index.ts
@@ -12,6 +12,9 @@ import {
   getInstalledPackageVersion,
   normalizePath,
   NowBuildError,
+  type Reporter,
+  Span,
+  type TraceEvent,
   validateNpmrc,
   type Builder,
   type BuildOptions,
@@ -24,6 +27,7 @@ import {
   type FlagDefinitions,
   type Meta,
   type PackageJson,
+  BUILDER_COMPILE_STEP,
 } from '@vercel/build-utils';
 import type { VercelConfig } from '@vercel/client';
 import { fileNameSymbol } from '@vercel/client';
@@ -77,6 +81,7 @@ import { validateConfig } from '../../util/validate-config';
 import { help } from '../help';
 import { pullCommandLogic } from '../pull';
 import { buildCommand } from './command';
+import { writeFile } from 'fs/promises';
 
 type BuildResult = BuildResultV2 | BuildResultV3;
 
@@ -117,12 +122,27 @@ export interface BuildsManifest {
   };
 }
 
+class InMemoryReporter implements Reporter {
+  public events: TraceEvent[] = [];
+
+  async flushAll() {}
+
+  report(event: TraceEvent) {
+    this.events.push(event);
+  }
+}
+
 export default async function main(client: Client): Promise<number> {
   const telemetryClient = new BuildTelemetryClient({
     opts: {
       store: client.telemetryEventStore,
     },
   });
+
+  // Setup tracer to output into the build directory
+  const reporter = new InMemoryReporter();
+  const rootSpan = new Span({ name: 'vc', reporter });
+
   let { cwd } = client;
 
   // Ensure that `vc build` is not being invoked recursively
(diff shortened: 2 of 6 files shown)


LukeSheard:
* **#12894** <a href="https://app.graphite.dev/github/pr/vercel/vercel/12894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/12894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13013** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13013?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/next/src/index.ts
```diff
@@ -2788,7 +2806,7 @@ export const diagnostics: Diagnostics = async ({
       '*',
       path.join(basePath, diagnosticsEntrypoint, outputDirectory, 'diagnostics')
     )),
-    // Collect `.next/trace` file
+    // // Collect `.next/trace` file
```
marc-vercel: Typo?

packages/next/src/index.ts
```diff
@@ -336,20 +340,32 @@ export const build: BuildV2 = async buildOptions => {
...
-  }
+  const { detectedPackageManager } =
+    detectPackageManager(cliType, lockfileVersion) ?? {};
+
+  await buildSpan
+    .child('installCommand', {
+      cliType,
+      lockfileVersion: String(lockfileVersion ?? ''),
+      packageJsonPackageManager: String(packageJsonPackageManager ?? ''),
+      detectedPackageManager: String(detectedPackageManager ?? ''),
```
marc-vercel: Nit: Can we also add the as a tag the install command that is being used?

packages/next/src/index.ts
```diff
@@ -336,20 +340,32 @@ export const build: BuildV2 = async buildOptions => {
...
+      if (typeof installCommand === 'string') {
+        if (installCommand.trim()) {
+          console.log(`Running "install" command: \`${installCommand}\`...`);
+
+          await execCommand(installCommand, {
+            ...spawnOpts,
+            cwd: entryPath,
+          });
+        } else {
+          console.log(`Skipping "install" command...`);
```
marc-vercel: Nit: we should also add the tag of whether we are executing the `installCommand` or  ignoring it, so that we can get more accurate data when doing time aggregations (I guess, in most cases, we want to avoid aggregating the time of a skipped install command).

packages/next/src/index.ts
```diff
@@ -460,36 +476,38 @@ export const build: BuildV2 = async buildOptions => {
...
 
-  if (buildCommand) {
-    // Add `node_modules/.bin` to PATH
-    const nodeBinPaths = getNodeBinPaths({
-      start: entryPath,
-      base: repoRootPath,
-    });
-    const nodeBinPath = nodeBinPaths.join(path.delimiter);
-    env.PATH = `${nodeBinPath}${path.delimiter}${env.PATH}`;
+  await buildSpan.child('buildCommand').trace(async () => {
```
marc-vercel: Nit: The same as the previous comments, it might be interesting to get which build command we are using

packages/build-utils/src/trace/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const BUILDER_INSTALLER_STEP = 'vc.builder.install';
+
+export const BUILDER_COMPILE_STEP = 'vc.builder.install';
```
healeycodes: should these have different values?

packages/cli/src/commands/build/index.ts
```diff
@@ -294,7 +315,14 @@ export default async function main(client: Client): Promise<number> {
     process.env.VERCEL = '1';
     process.env.NOW_BUILDER = '1';
 
-    await doBuild(client, project, buildsJson, cwd, outputDir);
+    await rootSpan
+      .child('vs.doBuild')
```
healeycodes: Should this be `vc`?

packages/next/src/index.ts
```diff
@@ -336,19 +343,48 @@ export const build: BuildV2 = async buildOptions => {
...
+  // If we don't have an install command defined then we fall back
+  // to zero config install, otherwise we run the user generated install command
+  const shouldRunInstallCommand =
+    // Case 1: We have a zero config install
+    typeof trimmedInstallCommand === 'undefined' ||
+    // Case 1: We have a install command which is non zero length
+    Boolean(trimmedInstallCommand);
+
+  buildSpan.setAttributes({
+    install: JSON.stringify(shouldRunInstallCommand),
```
EndangeredMassa, line 360: **question (non-blocking):** ‏do we want the string version of a boolean as the value here?

packages/next/src/index.ts
```diff
@@ -336,19 +343,48 @@ export const build: BuildV2 = async buildOptions => {
     await writeNpmRc(entryPath, process.env.NPM_AUTH_TOKEN);
   }
 
-  if (typeof installCommand === 'string') {
-    if (installCommand.trim()) {
-      console.log(`Running "install" command: \`${installCommand}\`...`);
+  const { detectedPackageManager } =
+    detectPackageManager(cliType, lockfileVersion) ?? {};
+
+  const trimmedInstallCommand = installCommand?.trim();
```
EndangeredMassa, line 349: **suggestion (non-blocking):** (not now, but if are interested) ‏we should push more of this validation and normalization earlier where it can be refactored into an actual interface that this bit of the system consumes

packages/cli/src/util/build/import-builders.ts
```diff
@@ -22,7 +22,7 @@ export interface BuilderWithPkg {
   path: string;
   pkgPath: string;
   builder: BuilderV2 | BuilderV3;
-  pkg: PackageJson;
+  pkg: PackageJson & { name: string };
```
EndangeredMassa, line 25: **suggestion (soft-blocking):** use `PackageJson

Do we need to overwrite `name?: string` with `name: string` here?

packages/cli/src/commands/build/index.ts
```diff
@@ -117,12 +122,27 @@ export interface BuildsManifest {
   };
 }
 
+class InMemoryReporter implements Reporter {
+  public events: TraceEvent[] = [];
+
+  async flushAll() {}
```
EndangeredMassa: **suggestion (soft-blocking):** implement

Can we either implement this method or make it throw? If someone uses it in the future thinking it will do something, that could cause a bug.

packages/cli/src/commands/build/index.ts
LukeSheard: Yup

packages/next/src/index.ts
LukeSheard: Agreed, I think this should be fixed at the parsing source so that we can remove these extra calls if possible in the future. I don't want to go too deep right now though.

packages/cli/src/util/build/import-builders.ts
LukeSheard: We actually only use the `name` property from the `package.json` of the builder. I didn't want to refactor the entire interface for the `pkg` so in this case I've just made `name` a required property here so that the types work downstream.

packages/cli/src/commands/build/index.ts
LukeSheard: Good point, a lot of this is hangover from the next.js implementation. I've removed this API from the `Reporter` interface for the time being until we need it.

packages/build-utils/src/trace/trace.ts
```diff
@@ -71,7 +70,8 @@ export class Span {
 
     this.status = 'stopped';
 
-    const duration = Date.now() - this._start;
+    // durations are reported in microseconds
+    const duration = (Date.now() - this._start) * 1000;
```
timneutkens: This should use `process.hrtime`

LukeSheard: ack

timneutkens (dismissed):

packages/cli/src/commands/build/index.ts
```diff
@@ -310,6 +336,18 @@ export default async function main(client: Client): Promise<number> {
...
+    try {
+      const diagnosticsOutputPath = join(outputDir, 'diagnostics');
+      await mkdir(diagnosticsOutputPath, { recursive: true });
+      // Ensure that all traces have flushed to disk before we exit
+      await writeFile(
+        join(diagnosticsOutputPath, 'cli_traces.json'),
+        JSON.stringify(reporter.events)
+      );
+    } catch (err) {
+      // Ignore errors when writing diagnostics
```
healeycodes: can we log an internal error?

packages/cli/test/unit/commands/build/index.test.ts
```diff
@@ -1226,7 +1226,12 @@ describe.skipIf(flakey)('build', () => {
 
       const files = await fs.readdir(output);
       // we should NOT see `functions` because that means `middleware.ts` was processed
-      expect(files.sort()).toEqual(['builds.json', 'config.json', 'static']);
+      expect(files.sort()).toEqual([
+        'builds.json',
+        'config.json',
+        'diagnostics',
```
healeycodes, line 1232: should we check the contents of `diagnostics`?

LukeSheard: I've added a expectation for this afterwards 👌

packages/cli/src/commands/build/index.ts
LukeSheard: Done

packages/cli/src/commands/build/index.ts
```diff
@@ -310,6 +336,19 @@ export default async function main(client: Client): Promise<number> {
...
+    try {
+      const diagnosticsOutputPath = join(outputDir, 'diagnostics');
+      await mkdir(diagnosticsOutputPath, { recursive: true });
+      // Ensure that all traces have flushed to disk before we exit
+      await writeFile(
+        join(diagnosticsOutputPath, 'cli_traces.json'),
+        JSON.stringify(reporter.events)
+      );
+    } catch (err) {
+      output.error('Failed to write diagnostics trace file');
```
EndangeredMassa, line 348:
```suggestion
      output.error(`Failed to write diagnostics trace file: ${diagnosticsOutputPath}`);
```

EndangeredMassa (approved):

PR #12895, opened 2025-01-22
[cli] add Node Version to project list output table

erikareads:
This PR adds `Node Version` to the table of outputs for `vercel project list`. This addition allows users to see what node version is set for a given project, which is especially important for deprecated projects in `--update-required`.

Before:
image

After:
image

### DIFF
--- a/.changeset/odd-shirts-think.md
+++ b/.changeset/odd-shirts-think.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+[cli] add Node Version to project list output table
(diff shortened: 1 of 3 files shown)


TooTallNate:
Can you include before / after screenshots please?

.changeset/odd-shirts-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
```
TooTallNate:
```suggestion
"vercel": minor
```

I'd argue that it's a new feature and thus should be minor. Don't have a strong preference though, so feel free to ignore.

TooTallNate (approved):

PR #12897, opened 2025-01-22
[next] Prefetch Segments

wyattjoh:
## Description

Added support for the new prefetch segments experimental feature which adds new rewrites to the routes config to support the new prerenders.

## Related PR's

https://github.com/vercel/next.js/pull/75202


packages/next/src/server-build.ts
```diff
@@ -2449,6 +2527,26 @@ export async function serverBuild({
           ]
         : []),
 
+      // When PPR is not enabled but client segment cache is, then rewrite any
```
ztanner: what logic below is determining that PPR is not enabled?

packages/next/src/server-build.ts
```diff
@@ -2074,10 +2090,72 @@ export async function serverBuild({
...
+                    continue: true,
+                    override: true,
+                  },
+                ]
+              : []),
             ...(rscPrefetchHeader && isAppPPREnabled
               ? [
                   {
-                    src: `^${path.posix.join('/', entryDirectory, '/')}`,
+                    src: `^${path.posix.join('/', entryDirectory, '/')}$`,
```
ztanner, line 2158: should this have changed?

ztanner (commented):
Are we able to add a test case that verifies the segment cache routes are resolving properly?

packages/next/src/server-build.ts
```diff
@@ -2074,10 +2090,72 @@ export async function serverBuild({
...
+                      },
+                      {
+                        type: 'header',
+                        key: rscPrefetchHeader,
+                        value: '1',
+                      },
+                      {
+                        type: 'header',
+                        key: prefetchSegmentHeader,
+                        value: '/(?<segmentPath>.+)',
```
ztanner, line 2120: any significance to having a capture group in the value here?

packages/next/src/server-build.ts
wyattjoh: That was an old comment, updating!

packages/next/src/server-build.ts
wyattjoh: This was a small optimization that I added, it improves the matching performance of the regular expressions.

ztanner (approved):

PR #12903, opened 2025-01-23
[react-router] Add initial `@vercel/react-router` package

TooTallNate:
Introduces a new (optional) `@vercel/react-router` package which enhances usage of React Router as a framework when deploying on Vercel. This package includes:

 * `@vercel/react-router` import
   * `createKvSessionStorage()` - create a session storage backed by [Vercel KV](https://vercel.com/docs/storage/vercel-kv)
 * `@vercel/react-router/vite` import
   * The `vercelPreset()` Vite preset
     * Allows usage of "edge" runtime for SSR routes on a per-route basis
     * Allows function-level configuration (i.e. `maxMemory`, `duration`, etc.) on a per-route basis
     * Allows Vercel to understand the routing structure of the application, which allows for bundle splitting
     * Accurate "Deployment Summary" on the deployment details page
 * `@vercel/react-router/entry.server` import
   * Provides implementations of the core `entry.server.jsx` logic for both Node.js and Edge runtimes
   * It is still possible to define your own `entry.server` file if needed, but the `handleRequest()` function from this import should be used when using "edge" runtime


packages/react-router/kv.ts
```diff
@@ -0,0 +1,97 @@
...
+  prefix = 'session',
+}: KvSessionStorageOptions) {
+  type S = SessionIdStorageStrategy<Data, FlashData>;
+
+  const createData: S['createData'] = async (data, expires) => {
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const baseId = genRanHex(16);
+      const id = `${prefix}:${baseId}`;
+      if ((await kv.exists(id)) === 0) {
```
erikareads, line 61: What does this `=== 0` condition signify?

packages/react-router/kv.ts
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  const updateData: S['updateData'] = async (id, data, expires) => {
+    const str = JSON.stringify(data);
+    if (str === '{}') {
+      // If the data is empty then delete the session key
+      return deleteData(id);
+    }
+    if (expires) {
+      await kv.set(id, str, { pxat: expires.getTime() });
```
erikareads: Does this logic need to be perfectly in sync with the logic on line 60? If so, we may want to extract that into a function.

packages/react-router/vite.ts
```diff
@@ -0,0 +1,215 @@
...
+import type { Preset, Config, BuildManifest } from '@react-router/dev/config';
+
+type RouteManifestEntry = BuildManifest['routes'][string];
+
+function hashConfig(config: Record<string, unknown>): string {
+  const str = JSON.stringify(config);
+  return Buffer.from(str).toString('base64url');
+}
+
+function flattenAndSort(o: Record<string, unknown>) {
```
erikareads: I'm not a fan of one-letter variable names, even for utility functions like this. Can we expand the names into something meaningful?

packages/react-router/vite.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export function vercelPreset(): Preset {
+  const project = new Project();
+  let vercelEntryServerPath: string | undefined;
+  let originalEntryServerPath: string | undefined;
+  let originalEntryServerContents: string | undefined;
+  const routeConfigs = new Map<string, BaseFunctionConfig>();
+  const bundleConfigs = new Map<string, BaseFunctionConfig>();
+
+  function getRouteConfig(
```
erikareads, line 57: This function assumes that `branch` will be ordered from parents -> children. We may want to add a comment to that effect in case that assumption is ever invalidated in a future react-router version.

erikareads (dismissed):

erikareads (approved):

PR #12904, opened 2025-01-23
[remix] Add support for React Router v7

TooTallNate:
Updates the `@vercel/remix-builder` Builder package to support React Router v7 apps. Implementation-wise, RRv7 is handled almost identically as Remix v2 + Vite, but with the name "remix" changed to "react-router" in a few places.

All the same features that are supported on Vercel when using Remix are now also supported when using React Router (when combined with the Preset, see #12903).

**NOTE:** E2E tests will be added in a follow-up PR after the new framework preset is added, so that autodetection will work in the tests.

### DIFF
--- a/packages/remix/src/build-vite.ts
+++ b/packages/remix/src/build-vite.ts
@@ -1,7 +1,7 @@
 import { readFileSync, promises as fs, statSync, existsSync } from 'fs';
 import { basename, dirname, join, relative, sep } from 'path';
 import { isErrnoException } from '@vercel/error-utils';
-import { nodeFileTrace } from '@vercel/nft';
+import { nodeFileTrace, NodeFileTraceOptions } from '@vercel/nft';
 import {
   BuildResultV2Typical,
   debug,
@@ -21,9 +21,10 @@ import {
 import {
   getPathFromRoute,
   getRegExpFromPath,
-  getRemixVersion,
+  getPackageVersion,
   hasScript,
   logNftWarnings,
+  findConfig,
 } from './utils';
 import type { BuildV2, Files, NodeVersion } from '@vercel/build-utils';
 
@@ -37,8 +38,12 @@ const nodeServerSrcPromise = fs.readFile(
   join(DEFAULTS_PATH, 'server-node.mjs'),
   'utf-8'
 );
+const reactRouterServerSrcPromise = fs.readFile(
+  join(DEFAULTS_PATH, 'server-react-router.mjs'),
+  'utf-8'
+);
 
-interface RemixBuildResult {
+interface BuildResultBase {
   buildManifest: {
     serverBundles?: Record<
       string,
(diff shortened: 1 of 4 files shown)


packages/remix/src/build-vite.ts
```diff
@@ -78,6 +94,14 @@ export const build: BuildV2 = async ({
   const mountpoint = dirname(entrypoint);
   const entrypointFsDirname = join(workPath, mountpoint);
 
+  // Determine if this is a React Router project
+  const isReactRouter = findConfig(workPath, 'react-router.config', [
+    '.js',
+    '.ts',
+    '.mjs',
+    '.mts',
+  ]);
```
EndangeredMassa, line 103: **suggestion (non-blocking):** consider classes

The repeated checks of `isReactRouter` imply that there are two distinct sets of data/state/behavior. If we modeled this as classes, the code could be simplified to operate on whichever one instance we need. This would reduce the maintenance burden of this package.‏

I don't think it needs to be in this PR, though.

---

Looking at it some more, there are distinct sets of data and behavior, but not state that needs to change over time. So, this could be modeled with any one of typed objects, static classes, or class instances. I took a stab at static classes here just because it's a lot easier to refactor that to class instances later, if necessary.

```typescript
abstract class FrameworkSettings {
  static configFileName: string;
  static primaryPackageName: string;
  static buildCommand: string;
  static buildResultFilePath: string;
  
  static createVariantFrameworkSettings() {
    const isReactRouter = findConfig(workPath, ReactRouterFrameworkSettings.configFileName, [
      '.js',
      '.ts',
      '.mjs',
      '.mts',
    ]);
    
    if (isReactRouter) {
      return ReactRouterFrameworkSettings;
    }
    
    return RemixFrameworkSettings;
  }
  
  createReanderFunction(
    nodeVersion: NodeVersion,
    entrypointDir: string,
    rootDir: string,
    serverBuildPath: string,
    serverEntryPoint: string | undefined,
    frameworkVersion: string,
    config: /*TODO: ResolvedNodeRouteConfig*/ any
  ): Promise<EdgeFunction | NodejsLambda> {
    throw new Error('FrameworkSettings#createReanderFunction not implemented');
  }
  
  /*
    could pull in some behaviors from one level of abstraction higher, like:
    - getFrameworkVersion
    - executeBuild
    - parseBuildResult
  */
}


class RemixFrameworkSettings extends FrameworkSettings{
  static primaryPackageName = '@remix-run/dev';
  static buildCommand = 'remix build';
  static buildResultFilePath =  '.vercel/remix-build-result.json';
  static configFileName = 'react-router.config';
  
  static createReanderFunction(
    nodeVersion: NodeVersion,
    entrypointDir: string,
    rootDir: string,
    serverBuildPath: string,
    serverEntryPoint: string | undefined,
    frameworkVersion: string,
    config: /*TODO: ResolvedNodeRouteConfig*/ any
  ): Promise<EdgeFunction | NodejsLambda> {
    // TODO: add in the implementation
  }
  
  // TODO: pull in the implementations for:
  // - `createRenderEdgeFunction`
  // - `createRenderNodeFunction`
}

class ReactRouterFrameworkSettings extends FrameworkSettings{
  static configFileName = ''; // TODO
  static primaryPackageName = 'react-router';
  static buildCommand = 'react-router build';
  static buildResultFilePath =  '.vercel/react-router-build-result.json';
  
  static createReanderFunction(
    nodeVersion: NodeVersion,
    entrypointDir: string,
    rootDir: string,
    serverBuildPath: string,
    serverEntryPoint: string | undefined,
    frameworkVersion: string,
    config: /*TODO: ResolvedNodeRouteConfig*/ any
  ): Promise<EdgeFunction | NodejsLambda> {
    // TODO: add in the implementation
  }
}
```

EndangeredMassa (approved):
> NOTE: E2E tests will be added in a follow-up PR after the new framework preset is added, so that autodetection will work in the tests.

Makes sense, but are there any other tests we can add? If we extract some functions, maybe?

erikareads (approved):

packages/remix/src/build-vite.ts
TooTallNate: Sounds good. Will tidy up the logic in a follow up PR.

PR #12905, opened 2025-01-23
[frameworks] Add "react-router" framework entry

TooTallNate:
Adds a new framework entry for React Router (when being used as a framework). The same Remix builder will be used internally, but having a dedicated preset allows for auto-detection to work and will show the correct logo when being used in the Project Settings.

**Note:** Merge _after_ https://github.com/vercel/vercel/pull/12904.


packages/frameworks/src/frameworks.ts
```diff
@@ -238,6 +238,59 @@ export const frameworks = [
...
+      'A user‑obsessed, standards‑focused, multi‑strategy router you can deploy anywhere.',
+    website: 'https://reactrouter.com',
+    sort: 7,
+    supersedes: ['hydrogen', 'vite'],
+    useRuntime: { src: 'package.json', use: '@vercel/remix-builder' },
+    detectors: {
+      some: [
+        {
+          path: 'vite.config.js',
+          matchContent: '@react-router/dev/vite',
```
EndangeredMassa, line 260: **question (soft-blocking):** ‏could this falsely detect projects using `react-router` only as a library?

Or is this a package that would only show up if you were using it as a framework?

EndangeredMassa (approved):

packages/frameworks/src/frameworks.ts
TooTallNate: The later. This Vite plugin would only be used when using RR as a framework.

See for example the [`ionic-react` template](https://github.com/vercel/vercel/blob/c2b3a3729afb48ca9f8446c57f4b56c9fe83c360/examples/ionic-react/package.json#L27) which uses RR as a library - that template remains detected as `vite`.

erikareads (approved):

onsclom (approved):

PR #12906, opened 2025-01-23
[next] return plain text 404 for `/_next/static/`

devjiwonchoi:
### Why?

When bots like Googlebot crawl a site and then decide to render after a moment, the static asset paths like `/_next/static/` may be invalid due to several reasons (e.g., new deployment).

If so, those paths currently return as a 404 page, which may bait the bots as if they found a missing HTML.

### How?

Create `.next/static/not-found.txt` file and use it to rewrite any 404 `_next/static/` paths.

Next.js side fix: https://github.com/vercel/next.js/pull/75111, but it isn't required for this change (though the behavior matches).

---
### Testing Plan

Surprising that there's no existing way to write a test for this PR's change of the `@vercel/next` package to run on the deployment. Seems like the integration test has been changed to test the build outputs.

I've confirmed on the deployed instance: https://export-testi-1dm8mif7v-jiwon-vc.vercel.app/_next/static/4

Uses v15.4.6, confirming not restricted to the Next.js side change.

Closes NDX-697
Closes NEXT-4671


devjiwonchoi:
* **#12906** <a href="https://app.graphite.dev/github/pr/vercel/vercel/12906?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/12906?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/next/test/fixtures/39-static-asset-404/app-dir/vercel.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 2,
+  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "probes": [
+    {
+      "path": "/_next/static/404",
+      "status": 404,
+      "mustContain": "Not Found",
+      "mustNotContain": "<html"
```
ijjk: We could add this probe to existing fixtures instead of creating it's own for just this check

packages/next/src/server-build.ts
```diff
@@ -1684,6 +1684,22 @@ export async function serverBuild({
...
+  // routes in Vercel CLI, the Next.js behavior is overwritten. Therefore, write a
+  // ".txt" file with "Not Found" content and rewrite any not found static assets to it.
+  const notFoundTextPath = path.posix.join(
+    entryPath,
+    outputDirectory,
+    'static/not-found.txt'
+  );
+  if (!fs.existsSync(notFoundTextPath)) {
+    await fs.writeFile(notFoundTextPath, 'Not Found');
+  }
```
ijjk: We don't need to write this to disk can just create a `FileBlob` output similar to here https://github.com/vercel/vercel/blob/8a020e07c2116833bbe39ee43f62efad9f7de0de/packages/next/src/server-build.ts#L1674-L1677

packages/next/src/server-build.ts
```diff
@@ -1890,6 +1890,17 @@ export async function serverBuild({
...
           }
         : {}),
+      // When bots crawl a site, they may render the page after awhile (e.g. re-sync),
+      // and the sub-assets may not be available then. In this case, the link to
+      // static assets could be not found, and return a 404 HTML. This behavior can
+      // bait the bots as if they found 404 pages. In Next.js it is handled on the
+      // server to return a plain text "Not Found". However, as we handle the "_next/static/"
+      // routes in Vercel CLI, the Next.js behavior is overwritten. Therefore, create a
+      // ".txt" file with "Not Found" content and rewrite any not found static assets to it.
+      '_next/static/not-found.txt': new FileBlob({
```
```suggestion
      [path.posix.join('./', entryDirectory, '_next/static/not-found.txt)]: new FileBlob({
```
This needs to include the entryDirectory like in the route's `dest`.

ijjk (approved):

PR #12915, opened 2025-01-24
[fs-detectors] Fix local file system detector to not throw for special file system stats 

onsclom:
**The problem:**
```
▲  og-test/ vercel
Vercel CLI 39.2.5
? Set up and deploy "~/Developer/scratch/og-test"? yes
? Which scope should contain your project? leerob-site
? Link to existing project? no
? What's your project's name? og-test
? In which directory is your code located? ./
Error: Dirent was neither file nor directory: fsmonitor--daemon.ipc
```

This change stops `readdir` from throwing for a dirent like `fsmonitor--daemon.ipc`.


packages/fs-detectors/src/detectors/local-file-system-detector.ts
```diff
@@ -47,7 +47,8 @@ export class LocalFileSystemDetector extends DetectorFilesystem {
         } else if (stat.isDirectory()) {
           type = 'dir';
         } else {
-          throw new Error(`Dirent was neither file nor directory: ${path}`);
+          // could be socket, fifo, block device, or character device
+          type = 'other';
```
TooTallNate: I wonder if these entries should be omitted entirely 🤔 

Also, this needs a unit test please

TooTallNate (changes requested):

packages/fs-detectors/src/detectors/local-file-system-detector.ts
onsclom: Changed it to omit these entries to avoid the `DetectorFilesystemStat` type change.

Testing this got a bit more complicated than I hoped. Creating sockets isn't as easy as creating files and dirs, and it doesn't happen at all on windows, but made a new test that does cover the original problem at least! Open to other ideas for this test.

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -1,30 +1,46 @@
...
     await Promise.all(
       dirs.map(dir => fs.mkdir(path.join(tmpdir, dir), { recursive: true }))
     );
+    fs.mkdir(socketdir, { recursive: true });
     await Promise.all(
       filePaths.map(filePath =>
         fs.writeFile(path.join(tmpdir, filePath), path.basename(filePath))
       )
     );
+    if (platform !== 'win32') {
```
TooTallNate, line 31: Why is this skipped on Windows? Can you add a comment in the code explaining?

TooTallNate: Why is this skipped on Windows? Can you add a comment in the code explaining?

TooTallNate (changes requested):

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -70,6 +86,11 @@ describe('LocalFileSystemDetector', () => {
     expect(actualPaths).toEqual(expectedPaths);
   });
 
+  it('should skip entry if not file or directory', async () => {
```
TooTallNate: Personally, I would make this test self-contained (do the before/after logic inline in this test), and perhaps use `it.skipIf()` if it doesn't apply for Windows.

onsclom: Good ideas, thanks!

Jest doesn't support `it.skipIf`, so did that bit manually.

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -70,6 +72,31 @@ describe('LocalFileSystemDetector', () => {
     expect(actualPaths).toEqual(expectedPaths);
   });
 
+  it('should skip entry if socket', async () => {
+    // Windows does not support Unix domain sockets
+    if (platform === 'win32') {
+      return;
+    }
```
TooTallNate, line 79:
```suggestion
  // Skipping on Windows because it does not support Unix domain sockets
  it.skipIf(process.platform === 'win32')('should skip entry if socket', async () => {
```

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -1,7 +1,9 @@
 import os from 'node:os';
 import fs from 'node:fs/promises';
 import path from 'node:path';
+import net from 'node:net';
 import { LocalFileSystemDetector, DetectorFilesystem } from '../src';
+import { platform } from 'node:process';
```
TooTallNate, line 6: I've never seen the `node:process` import personally. I'd say stick with `process.platform` (the global) for consistency with the rest of the codebase.

```suggestion
```

TooTallNate (dismissed):
LGTM. I left a couple syntactic suggestions though.

packages/fs-detectors/test/unit.local-file-system-detector.test.ts
TooTallNate: Oh I see, it's a vitest-ism.

TooTallNate (approved):

PR #12918, opened 2025-01-26
[remix] Enable `nativeFetch` when `v3_singleFetch` future flag is enabled

TooTallNate:
`nativeFetch` already is enabled when the previously named `unstable_singleFetch` future flag is enabled, so do the same for the stabalized flag name `v3_singleFetch`.

For more info, see the **"Deprecated `fetch` polyfill"** section of the [Remix docs](https://remix.run/docs/en/main/guides/single-fetch#enabling-single-fetch):

> Single Fetch requires using `undici` as your `fetch` polyfill, or using the built-in `fetch` on Node 20+, because it relies on APIs available there that are not in the `@remix-run/web-fetch` polyfill … you will need to call `installGlobals({ nativeFetch: true })` to use undici.

### DIFF
--- a/packages/remix/defaults/server-node.mjs
+++ b/packages/remix/defaults/server-node.mjs
@@ -8,9 +8,13 @@ import * as build from '@remix-run/dev/server-build';
 
 installGlobals({
   nativeFetch:
+    // Explicit opt-in to native fetch via runtime env var
     (parseInt(process.versions.node, 10) >= 20 &&
       process.env.VERCEL_REMIX_NATIVE_FETCH === '1') ||
-    build.future.unstable_singleFetch,
+    // `unstable_singleFetch` future flag added in Remix v2.9.0
+    build.future.unstable_singleFetch ||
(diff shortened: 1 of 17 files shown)


packages/remix/defaults/server-node.mjs
```diff
@@ -8,9 +8,13 @@ import * as build from '@remix-run/dev/server-build';
...
 installGlobals({
   nativeFetch:
+    // Explicit opt-in to native fetch via runtime env var
     (parseInt(process.versions.node, 10) >= 20 &&
       process.env.VERCEL_REMIX_NATIVE_FETCH === '1') ||
-    build.future.unstable_singleFetch,
+    // `unstable_singleFetch` future flag added in Remix v2.9.0
+    build.future.unstable_singleFetch ||
+    // `v3_singleFetch` future flag stabilized in Remix v2.13.0
+    build.future.v3_singleFetch,
```
TooTallNate, line 17: This is the main change. Everything else is E2E testing.

trek (approved):

onsclom (approved):

PR #12925, opened 2025-01-27
[remix] Create an interface for differences remix vs react-router

erikareads:
An attempt at tidying up the logic per: https://github.com/vercel/vercel/pull/12904#discussion_r1929370988


packages/remix/src/build-vite.ts
```diff
@@ -83,6 +83,40 @@ interface ReactRouterBuildResult extends BuildResultBase {
 
 type BuildResult = RemixBuildResult | ReactRouterBuildResult;
 
+interface BuilderInfo {
+  packageName: string;
+  buildCommand: string;
+  buildResultPath: string;
+}
+
+class ReactRouterInfo implements BuilderInfo {
```
TooTallNate: I'm not sure if a class adds much benefit here over a plain 'ol Object.

packages/remix/src/build-vite.ts
```diff
@@ -83,6 +83,40 @@ interface ReactRouterBuildResult extends BuildResultBase {
 
 type BuildResult = RemixBuildResult | ReactRouterBuildResult;
 
+interface BuilderInfo {
```
TooTallNate: Perhaps `FrameworkInfo`?

packages/remix/src/build-vite.ts
```diff
@@ -83,6 +83,105 @@ interface ReactRouterBuildResult extends BuildResultBase {
...
+  createRenderFunction: (
+    options: RenderFunctionOptions
+  ) => Promise<EdgeFunction | NodejsLambda>;
+}
+
+const REMIX_FRAMEWORK_SETTINGS: FrameworkSettings = {
+  primaryPackageName: '@remix-run/dev',
+  buildCommand: 'remix build',
+  buildResultFilePath: '.vercel/remix-build-result.json',
+  configFileName: '', // TODO
```
TooTallNate: There is no equivalent for Remix, so I think we don't need `configFileName`

erikareads: removed

TooTallNate (approved):

PR #12926, opened 2025-01-27
[node][remix][routing-utils] log diff between current and updated versions of `path-to-regexp`

EndangeredMassa:
Using [pnpm aliases](https://pnpm.io/aliases), this PR installs two different versions of `path-to-regexp`. Both are run and the resulting regex is compared. If they are different, an innocuous log message is sent to the build logs.

This allows us to test out the new version and be notified of any breaking changes it would have for our users.

---

Testing: This will be manually tested in the build container before release.


packages/node/src/utils.ts
```diff
@@ -1,8 +1,37 @@
...
+  try {
+    const newRegExp = pathToRegexpUpdated(path, keys, options);
+    const diff = currentRegExp.toString() !== newRegExp.toString();
+    // FORCE_PATH_TO_REGEXP_LOG can be used to force these logs to render
+    // for verification that they show up in the build logs as expected
+    if (process.env.FORCE_PATH_TO_REGEXP_LOG || diff) {
+      // each of these path-to-regexp comparisons logs an innocuous message
+      // so we (1) can easily grab it by querying the build logs
+      // and (2) it doesn't spook the user
+      console.log(`SYSTEM TEST #${callerId} A: ${path}`);
```
TooTallNate: Prefix the log with `[vc]` and it'll be omitted from the user-facing logs entirely (but still be queryable on the backend).

erikareads (dismissed):

packages/node/src/utils.ts
TooTallNate: Also, should we log the before and after regexps?

trek (dismissed):
No comments beyond Nate's damn good point.

packages/node/src/utils.ts
```diff
@@ -1,8 +1,37 @@
 import { debug, streamToBuffer } from '@vercel/build-utils';
-import { pathToRegexp } from 'path-to-regexp';
+import { pathToRegexp as pathToRegexpCurrent, Key } from 'path-to-regexp';
+import { pathToRegexp as pathToRegexpUpdated } from 'path-to-regexp-updated';
 import type { IncomingMessage } from 'http';
 import { extname } from 'path';
 
+// run the updated version of path-to-regexp, compare the results, and log if different
+function pathToRegexp(
```
erikareads, line 8: At what point do we consider a common utility package for functions like this?

TooTallNate (dismissed):

erikareads (dismissed):

trek (approved):

trek (approved):

TooTallNate (approved):

EndangeredMassa:
[That test failure](https://github.com/vercel/vercel/actions/runs/13034973415/job/36363130123?pr=12926) is unrelated and being addressed outside of this PR. I'm going to merge.

PR #12943, opened 2025-01-28
[cli][client] Make split-tgz default for archive deployments

onsclom:
This makes `vercel deploy --archive=tgz` use the new `vercel deploy --archive=split-tgz` behavior.

`vercel deploy --archive=split-tgz` still works, but now it prints a warning:

image


.changeset/weak-goats-swim.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/client': major
+'vercel': minor
```
trek: I think this should be major as well. We're removing the ability to do a big single file, which could potentially break _something_ for someone.

trek (changes requested):

packages/cli/src/commands/deploy/index.ts
```diff
@@ -215,11 +215,27 @@ export default async (client: Client): Promise<number> => {
...
+    typeof parsedArchive === 'string' &&
+    !(
+      isValidArchive(parsedArchive) ||
+      parsedArchive === deprecatedArchiveSplitTgz
+    )
+  ) {
     output.error(`Format must be one of: ${VALID_ARCHIVE_FORMATS.join(', ')}`);
     return 1;
   }
+  if (parsedArchive === 'split-tgz') {
```
erikareads: Why do we introduce `deprecatedArchiveSplitTgz` above, and not use the constant here?

onsclom: Good catch, thanks!

erikareads (approved):

trek (approved):

PR #12957, opened 2025-01-29
[static-build] update 45-node-modules-bin-path fixture

trek:
The existing fixture is quite old, so I've updated it to latest major and updated the behavior to match how Next 15 accomplishes it.

The old behavior as implemented warned:

> `next export` has been removed in favor of 'output: export' in next.config.js.


packages/static-build/test/fixtures/45-node-modules-bin-path/vercel.json
```diff
@@ -5,13 +5,8 @@
...
       "config": {
         "zeroConfig": true,
-        "buildCommand": "next build && next export",
         "outputDirectory": "out"
       }
     }
-  ],
-  "probes": [{ "path": "/", "mustContain": "hello world" }],
-  "projectSettings": {
-    "nodeVersion": "16.x"
```
trek, line 15: This little bugger...

erikareads (approved):

PR #12960, opened 2025-01-29
[remix] extract common function options

erikareads:
This PR assumes that Remix and React-Router should vary at the same time, and that there isn't any situation where these options should differ between the two build paths. If that assumption is invalid, we should reject this PR and consider another approach.


TooTallNate (approved):
This is valid, though I think we can soon head into a direction where there is only one create function. Need to wait for `useWebApi` to finish rolling out before we can do that though.

erikareads:
> This is valid, though I think we can soon head into a direction where there is only one create function. Need to wait for `useWebApi` to finish rolling out before we can do that though.

This is precisely where I want to go. Do you mean between Edge and Server, or between Remix and React Router?

trek (approved):

EndangeredMassa (approved):

PR #12961, opened 2025-01-29
Add yarn cache to default build cache paths

onsclom:
Adds  `.yarn/cache` to the `prepareCache` for builders that may be using yarn. This allows yarn 2, 3, and 4 projects to cache deps between deploys.

Tested manually with each major yarn version using `vercel deploy -b VERCEL_CLI_VERSION=https://vercel-i1pucqxc7.vercel.sh/tarballs/vercel.tgz` and confirmed this cached deps on each version.


trek (approved):

packages/next/src/index.ts
```diff
@@ -2832,6 +2832,12 @@ export const prepareCache: PrepareCache = async ({
       path.join(cacheEntrypoint, outputDirectory, 'cache/**'),
       cacheBasePath
     )),
+    ...(await glob(
+      isMonorepo
+        ? '**/.yarn/cache/**'
```
EndangeredMassa, line 2837: **suggestion (non-blocking):** prevent nested caches from being included

In an app where there are nested packages, this glob will pick up all caches found in all of those packages. Maybe we should find a way to only include the cache at a somewhat known place, like workspace package roots or repo root.

Of course, projects should have nested packages in normal circumstances. It can happen when using project test fixtures, though.‏

EndangeredMassa (approved):

PR #12962, opened 2025-01-29
[node][remix][routing-utils] better path-to-regexp diff logging

EndangeredMassa:
Updated the `path-to-regexp` comparison:
- centralized logic because it was getting coplicated
- compared `keys` more carefully
- use `keys` in updated `path-to-regexp` function as a clone of the original to avoid duplicate values

---

Follow-up to: https://github.com/vercel/vercel/pull/12926


erikareads (dismissed):

erikareads (dismissed):

TooTallNate (dismissed):

TooTallNate (dismissed):

packages/routing-utils/src/superstatic.ts
```diff
@@ -3,57 +3,9 @@
...
-    const isDiffRegExp = currentRegExp.toString() !== newRegExp.toString();
-    if (process.env.FORCE_PATH_TO_REGEXP_LOG || isDiffRegExp) {
-      const message = JSON.stringify({
-        path,
-        currentRegExp: currentRegExp.toString(),
-        newRegExp: newRegExp.toString(),
-      });
-      console.error(`[vc] PATH TO REGEXP PATH DIFF @ #${callerId}: ${message}`);
-    }
+import { pathToRegexp, compile, Key } from '@vercel-internals/path-to-regexp';
```
blaine-arcjet, line 6: This was moved to devDeps but you are using it in a normal code path. The published package is crashing for consumers, like `@astrojs/vercel`

ascorbic: It looks like this is because unlike the runtime packages, `@vercel/routing-utils` isn't being bundled, so internal packages aren't being included.

ascorbic: I've opened an issue: #13024

PR #13016, opened 2025-02-05
[cli] remove warning `projects lists --update-required` for deprecated node versions

erikareads:
Now that we have the Node version displayed in the table for `vercel project list --update-required`, the specific version warning is unneeded. This PR removes that warning.

Addresses: https://linear.app/vercel/issue/ZERO-3252

### DIFF
--- a/packages/cli/src/commands/project/list.ts
+++ b/packages/cli/src/commands/project/list.ts
@@ -3,7 +3,6 @@ import chalk from 'chalk';
 import table from '../../util/output/table';
 import getCommandFlags from '../../util/get-command-flags';
 import { getCommandName } from '../../util/pkg-name';
-import { NODE_VERSIONS } from '@vercel/build-utils';
 import { ProjectListTelemetryClient } from '../../util/telemetry/commands/project/list';
 import output from '../../output-manager';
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/project/list.ts
```diff
@@ -89,14 +89,16 @@ export default async function list(
...
+    if (upcomingDiscontinuedVersionsList) {
+      output.warn(
+        `The following Node.js versions are deprecated: ${upcomingDiscontinuedVersionsList.join(
+          ', '
+        )} and will be discontinued soon. Please upgrade your projects immediately.`
+      );
+      output.log(
+        'For more information visit: https://vercel.com/docs/functions/serverless-functions/runtimes/node-js#node.js-version'
+      );
+    }
```
EndangeredMassa: **suggestion (blocking):** ‏remove

Re-thinking this warning: With the updated endpoint showing deprecated and discontinued versions in projects, I don't think we need a warning at all anymore. I suggest we get rid of it.

EndangeredMassa (changes requested):

EndangeredMassa (approved):

PR #13026, opened 2025-02-06
[routing-utils] fix dep bundling

EndangeredMassa:
In https://github.com/vercel/vercel/pull/12962/files#diff-f253688722ee28de3774c9d267046c4fee6c9615e5ecb1d27aa69b82760cf81cR26, the reference to `path-to-regexp` was moved to an internals package, but `routing-utils` is not bundled when published. This lead to unreachable imports.

This PR removes the internals package and copies the code in places where it is needed. This seems acceptable given that this code will be deleted after the observation period has ended.


trek (dismissed):

erikareads (dismissed):

TooTallNate (changes requested):
I don't think this will actually fix the issue since this package is not bundled at build-time.

TooTallNate (approved):

trek (approved):

erikareads (approved):

PR #13027, opened 2025-02-06
feat(related): add related projects utils

tknickman:
Add util package for working with related projects on Vercel.


EndangeredMassa:
@tknickman can you provide more details about what this is and what it would be used for? Why is this repo the right place for it?

.github/CODEOWNERS
```diff
@@ -15,6 +15,7 @@
 /examples/create-react-app               @TooTallNate @EndangeredMassa @trek @onsclom @jeffsee55 @erikareads @Timer
 /examples/nextjs                         @ijjk @ztanner @huozhi
 /packages/node                           @TooTallNate @EndangeredMassa @trek @onsclom @jeffsee55 @erikareads @Kikobeats
+/packages/related-projects               @TooTallNate @EndangeredMassa @trek @onsclom @jeffsee55 @erikareads @Kikobeats @tknickman @mknichel
```
EndangeredMassa, line 18:
```suggestion
/packages/related-projects               @tknickman @mknichel
```

I think this should be restricted to only you two. That way, everyone else isn't notified on every change to this package.

EndangeredMassa (approved):

PR #13028, opened 2025-02-06
[next] Add node middleware handling

ijjk:
This adds the handling for nodejs runtime support for middleware added in https://github.com/vercel/next.js/pull/75624


packages/next/src/middleware-launcher.ts
```diff
@@ -0,0 +1,77 @@
...
+            waitUntil: context.waitUntil,
+          },
+        });
+
+        if (result.waitUntil && context.waitUntil) {
+          context.waitUntil(result.waitUntil);
+        }
+
+        // TODO: remove after debugging
+        console.log('got middleware response', result.response.headers);
```
Schniz: can we remove this or add an env var?

Schniz (dismissed):
i'm stamping this but I feel like this code should be refactored. The functions are too long and do a lot of things, making it hard to contribute further down the road. But I don't want to block. I wonder if @timneutkens should also review as most of this is actually related to Next.js and not the platform?

ztanner (commented):
Looks like we have good pages dir coverage here but nothing for app dir -- I'm not sure what all could differ between the two but maybe worth updating the test fixture to cover both

packages/next/src/middleware-launcher.ts
```diff
@@ -0,0 +1,74 @@
...
+      async () => {
+        const result = await middlewareHandler({
+          request: {
+            url: request.url,
+            method: request.method,
+            headers: toPlainHeaders(request.headers),
+            nextConfig: conf,
+            page: '/middleware',
+            body:
+              request.method !== 'GET' && request.method !== 'HEAD'
```
ztanner, line 50: should this also ignore `OPTIONS`?

packages/next/test/fixtures/00-node-middleware/index.test.js
```diff
@@ -0,0 +1,125 @@
...
+            props,
+          });
+    }, 'success');
+
+    await check(async () => {
+      const { pageProps: newData } = await fetch(
+        `${ctx.deploymentUrl}/_next/data/testing-build-id/rewrite-to-another-site.json`
+      ).then(res => res.json());
+
+      console.log({ newData, data });
```
ztanner, line 56: leftover debugging i think

packages/next/test/fixtures/00-node-middleware/index.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const revalidateRes = await fetch(
+      `${ctx.deploymentUrl}/api/revalidate?urlPath=/financial`
+    );
+    expect(revalidateRes.status).toBe(200);
+    expect(await revalidateRes.json()).toEqual({ revalidated: true });
+
+    await check(async () => {
+      const newProps = await propsFromHtml();
+      console.log({ props, newProps });
```
ztanner, line 94: same

packages/next/test/fixtures/00-node-middleware/middleware.js
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  if (url.pathname === '/logs') {
+    console.clear();
+    for (let i = 0; i < 3; i++) console.count();
+    console.count('test');
+    console.count('test');
+    console.dir({ hello: 'world' });
+    console.log('hello');
+    console.log('world');
```
ztanner, line 73: this too I think?  or if we're asserting on the logging behavior, i didn't see a test

packages/next/src/middleware-launcher.ts
ijjk: this is copied from the Next.js handler which handles the cases Node.js' Request object will throw when a body is provided. 

https://github.com/vercel/next.js/blob/b60bb5e7f7d27a0a7e822843eaf8ce30ee978cea/packages/next/src/server/next-server.ts#L1575-L1576

packages/next/test/fixtures/00-node-middleware/middleware.js
ijjk: These tests are just copied from our existing middleware test suite and untouched from there https://github.com/vercel/vercel/blob/7237472608e2d021c2cc41e5a45bd61b3cee3f31/packages/next/test/fixtures/00-middleware/middleware.js#L70

I also believe these console tests were explicitly making sure these console methods work and don't break unexpectedly.

ztanner (approved):

PR #13029, opened 2025-02-06
[tests] disable flakey prerender test

EndangeredMassa:
This prerender test occasionally fails.

Disabling it along with other similar flakey tests. https://linear.app/vercel/issue/ZERO-3240/unskip-random-test-failures

Recent failure example: https://github.com/vercel/vercel/actions/runs/13187072443/job/36811748262?pr=13026#step:8:501

### DIFF
--- a/.changeset/odd-crabs-act.md
+++ b/.changeset/odd-crabs-act.md
@@ -1,0 +1,4 @@
+---
+---
+
+[tests] disable flakey prerender test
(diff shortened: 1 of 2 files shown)


trek (dismissed):

ijjk (changes requested):
Can we make it less flakey instead, this is good coverage for our ISR handling end to end and disabling for convenience could cause us to miss regressions.

EndangeredMassa:
@ijjk Our flakey test process is:

- disable it to allow CI to operate for other work
- track it in a card in the triage queue
- current support dev prioritizes restoring it

The test being in another team's ownership area makes this fuzzier, but we'd still like to operate this way. This is what we did last time: https://github.com/vercel/vercel/pull/12782

PR #13031, opened 2025-02-06
[remix] Support "build" script that produces Build Output API

TooTallNate:
The existing [React Router + Vercel template](https://github.com/remix-run/react-router-templates/tree/58177f3afeb37701f19a65f534f90b4a2059004c/vercel) is configured to produce a Build Output API structure, so ensure that `@vercel/remix-builder` accepts that output when provided.


packages/remix/src/build-vite.ts
```diff
@@ -336,6 +341,25 @@ export const build: BuildV2 = async ({
...
+  // to the Build Output v3 API, then stop processing here
+  // since the output is already in its final form.
+  let buildOutputVersion: undefined | number;
+  try {
+    const boaConfigPath = join(entrypointFsDirname, 'output/config.json');
+    const buildResultContents = await fs.readFile(boaConfigPath, 'utf8');
+    const data = JSON.parse(buildResultContents);
+    buildOutputVersion = data.version;
+  } catch (_) {
+    // tolerate for older versions
```
EndangeredMassa: **suggestion (soft-blocking):** can we at least debug log this error?

packages/remix/src/build-vite.ts
```diff
@@ -336,6 +341,25 @@ export const build: BuildV2 = async ({
...
+  }
+  if (buildOutputVersion) {
+    return {
+      buildOutputPath: join(entrypointFsDirname, 'output'),
+      buildOutputVersion,
+    } as BuildResultBuildOutput;
```
EndangeredMassa: **suggestion (non-blocking):** ‏do we need the type assertion? Can we make it work without?

packages/remix/test/fixtures-react-router/04-react-router-boa-template/vercel/prepare.js
EndangeredMassa, line 1: **question (non-blocking):** ‏is there a way to make this test only work if BOA was detected and used?

It looks like this fixture would still pass if the code change (in `packages/remix/src/build-vite.ts`) was not made, but maybe I'm misreading it.

EndangeredMassa (dismissed):

packages/remix/src/build-vite.ts
TooTallNate: This was copy+pasted from `@vercel/next`. We should probably extract it out into a build-utils function, but I just wanted to do something quickly to resolve the customer report and then we can tidy up after.

TooTallNate: I cleaned up the logic a bit for now, including adding debug logging.

packages/remix/src/build-vite.ts
TooTallNate: Updated to make the type assertion unnecessary.

packages/remix/test/fixtures-react-router/04-react-router-boa-template/vercel/prepare.js
TooTallNate: It does end up failing without the changes in this PR, due to the imported entrypoint not exporting the function signature that Vercel is expecting:

```
Error running the exported Web Handler: TypeError: Cannot read properties of undefined (reading 'method')
    at Function.handle2 [as handle] (file:///var/task/build/server/assets/app-BuaMb-CS.js:35219:37)
    at Function.handle3 [as handle] (file:///var/task/build/server/assets/app-BuaMb-CS.js:41022:13)
    at app2 (file:///var/task/build/server/assets/app-BuaMb-CS.js:42857:12)
    at Object.requestHandler [as handlerWeb] (/var/task/node_modules/react-router/dist/development/index.js:9114:50)
    at Object.handler (/opt/rust/nodejs.js:2:13371)
    at Server.<anonymous> (/opt/rust/nodejs.js:2:1526)
    at /opt/rust/nodejs.js:16:5714
    at AsyncLocalStorage.run (node:internal/async_local_storage/async_hooks:91:14)
    at /opt/rust/nodejs.js:16:5702
    at AsyncLocalStorage.run (node:internal/async_local_storage/async_hooks:91:14)
```

PR #13049, opened 2025-02-11
[build-utils] extract install command specific logic into helper

erikareads:
This PR does some tidying on the `runNpmInstall` function, extracting some logic into a named helper function for ease of reading.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -622,6 +622,8 @@ function getInstallCommandForPackageManager({
     case 'pnpm':
       return {
         prettyCommand: 'pnpm install',
+        // PNPM's install command is similar to NPM's but without the audit nonsense
+        // @see options https://pnpm.io/cli/install
         commandArguments: args
           .filter(a => a !== '--prefer-offline')
           .concat(['install', '--unsafe-perm']),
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -604,6 +604,61 @@ function isSet<T>(v: any): v is Set<T> {
...
+async function runInstallCommand({
+  packageManager,
+  args,
+  opts,
+}: {
+  packageManager: CliType;
+  args: string[];
+  opts: SpawnOptionsExtended;
+}) {
+  const getInstallCommandForPackageManager = (
```
EndangeredMassa, line 616: **suggestion (non-blocking):** ‏I would pull this function out into module scope. Do we need to re-define it here in the invocation of `runInstallCommand`?

EndangeredMassa (approved):

packages/build-utils/src/fs/run-user-scripts.ts
erikareads: I mostly kept it in the function since I didn't want to do mutation on a `let`, and invoking a function is the alternative. Pulling it out into a module scoped helper is what I did first, and changed my mind to do this. I'm not opposed to it.

PR #13052, opened 2025-02-12
[next] Handle patterns in rewrites

wyattjoh:
Handles patterns correctly in rewrites.

https://github.com/vercel/next.js/pull/75973

### DIFF
--- a/.changeset/great-mangos-try.md
+++ b/.changeset/great-mangos-try.md
@@ -1,0 +1,6 @@
+---
+'@vercel/next': patch
+'vercel': patch
+---
+
+Handle patterns in rewrite destinations correctly
--- a/.changeset/great-mangos-try.md
+++ b/.changeset/great-mangos-try.md
@@ -1,6 +1,5 @@
 ---
 '@vercel/next': patch
-'vercel': patch
 ---
 
 Handle patterns in rewrite destinations correctly
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1,5 +1,4 @@
 import path from 'path';
-import url from 'url';
 import semver from 'semver';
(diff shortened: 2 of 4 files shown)


wyattjoh:
* **#13052** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/great-mangos-try.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/next': patch
+'vercel': patch
```
LukeSheard: Changesets should handle bumping `vercel` if any of the dependencies are bumped, so this should be redundant! 😄

packages/next/src/server-build.ts
```diff
@@ -262,9 +261,46 @@ export async function serverBuild({
...
+          // then we'll remove it.
+          pathname = rewrite.dest;
+
+          let index = pathname.indexOf('?');
+          if (index !== -1) {
+            query = pathname.substring(index + 1);
+            pathname = pathname.substring(0, index);
+
+            // If there's a hash, we should remove it.
+            index = query.indexOf('#');
```
ztanner, line 292: since we're removing the hash in both the `if` and `else`, would it make sense to handle this earlier so we don't have to duplicate the logic? 

eg outside the if:
```
const hashIndex = pathname.indexOf('#');
if (hashIndex !== -1) {
  pathname = pathname.substring(0, hashIndex);
}
```

ztanner (approved):

PR #13060, opened 2025-02-13
[build-utils] [tidy] extract checkIfAlreadyInstalled helper to clarify intent

erikareads:
Prior to this PR, we had logic in `runNpmInstall` that prevented duplicate installs based on the same `packageJsonPath`. The code was a nest of conditionals that made it difficult to follow what it was trying to accomplish. 

This PR extracts the conditional logic into a couple helper functions, and refactors `runNpmInstall` to make it clear what's happening: we're returning early if a default install has already happened for a given packageJsonPath.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -659,6 +659,30 @@ async function runInstallCommand({
   await spawnAsync(packageManager, commandArguments, opts);
 }
 
+function initializeSet(set: unknown) {
+  if (!isSet<string>(set)) {
+    return new Set<string>();
+  }
+  return set;
+}
+
+function checkIfAlreadyInstalled(
+  runNpmInstallSet: unknown,
+  packageJsonPath: string
+) {
+  const initializedRunNpmInstallSet = initializeSet(runNpmInstallSet);
+
+  if (initializedRunNpmInstallSet.has(packageJsonPath)) {
+    return { alreadyInstalled: true, initializedRunNpmInstallSet };
+  }
+
+  initializedRunNpmInstallSet.add(packageJsonPath);
+  return {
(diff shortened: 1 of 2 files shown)


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -659,6 +659,30 @@ async function runInstallCommand({
...
+
+  if (initializedRunNpmInstallSet.has(packageJsonPath)) {
+    return { alreadyInstalled: true, initializedRunNpmInstallSet };
+  }
+
+  initializedRunNpmInstallSet.add(packageJsonPath);
+  return {
+    alreadyInstalled: false,
+    runNpmInstallSet: initializedRunNpmInstallSet,
+  };
```
onsclom:
```suggestion
  const initializedRunNpmInstallSet = initializeSet(runNpmInstallSet);
  const alreadyInstalled = initializedRunNpmInstallSet.has(packageJsonPath);

  initializedRunNpmInstallSet.add(packageJsonPath);
  return { alreadyInstalled, initializedRunNpmInstallSet };
```

onsclom: I think this change simplifies things a bit more.

Don't have a huge preference though, so happy to approve as is.

onsclom (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
erikareads: Allowing `initializedRunNpmInstallSet.add(packageJsonPath)` to run everytime only works because `Set.add` is idempotent, but since that's the point of the operation, this simplification makes sense.

onsclom (dismissed):

onsclom (approved):

PR #13062, opened 2025-02-13
[next] simplify hash stripping logic

wyattjoh:
This pull request includes changes to streamline the handling of query strings and hashes in the `serverBuild` function in `packages/next/src/server-build.ts`.

Refactoring query string and hash handling:

* [`packages/next/src/server-build.ts`](diffhunk://#diff-5cc9007d6ec83d81ec4768b6a297f5c5c08455f76cf7ec57526f6e731a8db674L290-L303): Removed redundant logic for handling hashes in the query string, ensuring that hashes are only processed once for both `pathname` and `query`.


ztanner (approved):

wyattjoh:
Turns out this was wrong 😅

PR #13063, opened 2025-02-13
[build-utils] extract some PackageManager details into typed objects

erikareads:
There are a lot of instances in `run-user-scripts.ts` that switch on `cliType`. Rather than having a diversity of switch statements, it makes sense to colocate concerns that are specific to a given package manager.

This PR takes a first step towards centralizing some of those concerns, but doesn't yet remove switch statements. 

One concern I have with modeling `PackageManager` is that we export the type `CliType`, so we can't simply change call signatures and return types to use the `PackageManager` type instead of `CliType`.

However, I still think it's worth modeling PackageManager internally to `run-user-scripts` with aspirations to refactor all callers to use this type instead of `CliType`. 

For review, I'm particularly interested in:

- what we think of this approach?
- should the names of the constants be changed (perhaps `NPM_SETTINGS`?)
- anything else that comes to mind

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -38,6 +38,44 @@ const NO_OVERRIDE = {
 
 export type CliType = 'yarn' | 'npm' | 'pnpm' | 'bun';
 
+interface PackageManager {
+  cliType: CliType;
+  prettyInstallCommand: string;
+  prettyRunCommand: (scriptName: string) => string;
+}
+
+const NPM: PackageManager = {
+  cliType: 'npm',
+  prettyInstallCommand: 'npm install',
+  prettyRunCommand: (scriptName: string) => {
+    return `npm run ${scriptName}`;
+  },
+};
+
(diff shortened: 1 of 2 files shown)


erikareads:
Based on where the `cliType` is determined:

 https://github.com/vercel/vercel/blob/460b6c1a0119087eaea50517c14c04b8e46c5eb7/packages/build-utils/src/fs/run-user-scripts.ts#L449-L474

I'm considering using Classes for each package manager, so that the `lockfilePath` and `lockfileVersion` can live and move with the package manager.

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -611,14 +649,14 @@ function getInstallCommandForPackageManager(
   switch (packageManager) {
     case 'npm':
```
EndangeredMassa, line 650: Do we want to switch on `cliType` of each object as well?

EndangeredMassa (commented):
I think this is a reasonable prefactor towards a class for these.

PR #13071, opened 2025-02-14
fix(cli): catch json parse

tknickman:


packages/client/src/deploy.ts
```diff
@@ -66,7 +66,12 @@ async function* postDeployment(
       }
     );
 
-    const deployment = await response.json();
+    let deployment = undefined;
+    try {
+      deployment = await response.json();
```
jv4pres, line 71: nice, can we also add a check before trying to parse as JSON:
```
if (!response.ok) throw new Error(`Received error response: ${response.statusText}`);
```

or similar? we shouldn't try to parse an unsuccessful response at all.

jv4pres (dismissed):
thanks Tom

Zertsov (dismissed):

Zertsov (approved):

erikareads (approved):

packages/client/src/deploy.ts
```diff
@@ -66,7 +66,12 @@ async function* postDeployment(
...
+    } catch (error) {
+      throw new Error('Invalid JSON response');
```
EndangeredMassa, line 73: Can we either (1) debug log the `error` or (2) add the `error` message to the generated error?

PR #13072, opened 2025-02-14
[build-utils] add pnpm 10 detection for new projects

onsclom:
Reintroduces https://github.com/vercel/vercel/pull/12852, but with a new strategy:
```
if `pnpm-lock.yaml` v9 is found:
	if project was created before 2025-02-24T20:00:00Z:
		use pnpm 9
	else:
		use pnpm 10
```
This will avoid breaking old projects, and will add zero configuration pnpm 10 support for new projects.


TooTallNate (dismissed):

EndangeredMassa (dismissed):

onsclom:
Updating the strategy here to stick with pnpm 9 for old projects, and use pnpm 10 for new projects.

onsclom:
Updated the strategy of this PR from checking `engines.pnpm` to checking the project creation timestamp.

TooTallNate:
Is there an explanation for the change in strategy? I thought the `engines.pnpm` check made sense.

So with the new strategy, if I have an old project there is no way to opt-in to pnpm 10? And vice versa?

onsclom:
> Is there an explanation for the change in strategy? I thought the `engines.pnpm` check made sense.
> 
> So with the new strategy, if I have an old project there is no way to opt-in to pnpm 10? And vice versa?

We talked about this on slack, but to record here:

The new strategy is because there's another tricky case where `pnpm@10` can break builds. The `package.json#packageManager` can specify `pnpm@8` while the lockfile is v9, so the project currently uses `pnpm@9`. Upgrading this project to use `pnpm@10` means actually using `pnpm@8` and ignoring the lockfile.

Old projects can still opt-in to pnpm 10 with corepack or a custom install command. New projects can still opt in to pnpm 9.

packages/build-utils/test/unit.detect-package-manager.test.ts
```diff
@@ -87,6 +90,26 @@ describe('Test `detectPackageManager()`', () => {
...
+        want: {
+          detectedLockfile: 'pnpm-lock.yaml',
+          detectedPackageManager: 'pnpm@9.x',
+          pnpmVersionRange: '9.x',
+          path: '/pnpm9/node_modules/.bin',
+        },
+      },
+      {
+        name: 'for 9.0 lockfile returns pnpm 10 path after prefer pnpm 10 datetime',
+        args: ['pnpm', 9.0, PNPM_10_PREFERRED_AT.getTime() * 1000],
```
EndangeredMassa:
```suggestion
        args: ['pnpm', 9.0, PNPM_10_PREFERRED_AT.getTime() + 1000],
```

This works, but I think you meant to use `+` here.

EndangeredMassa (dismissed):

packages/build-utils/test/unit.detect-package-manager.test.ts
onsclom: Whoops, thanks for catching that!

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -371,8 +377,18 @@ export async function scanParentDirs(
       'pnpm-lock.yaml',
       'bun.lock',
       'bun.lockb',
+      '.vercel/project.json',
     ],
   });
+
+  const vercelProjectJson = vercelProjectJsonPath
```
TooTallNate: This logic is not necessary. Builders have access to the project settings in the `config` object passed to the `build()` function.

TooTallNate (changes requested):

packages/build-utils/src/fs/run-user-scripts.ts
onsclom: Thanks! Updated everything to use `config` from builders instead.

TooTallNate (approved):

trek (approved):

PR #13082, opened 2025-02-20
Add Flags SDK example

AAorris:
### Overview

image

This example uses the Flags SDK to demonstrate a basic feature flag.

The README includes information about how to use the Flags Explorer when deploying on Vercel.

The flag shows a random variation either highlighting deploy now or the Flags SDK docs.

Using the Flags Explorer, you can set overrides to a particular variant.

### Before Merging

Should merge when installed via `npm i flags`


trek (changes requested):
We're mid migration for examples going to https://github.com/vercel/examples. The ones that are here still are basically to make them available to `vercel init` before we migrate that behavior as well.

AAorris:
Moving to vercel/examples

PR #13083, opened 2025-02-20
[build-utils] support `manage-package-manager-versions` by avoiding engines error

onsclom:
[`pnpm@10` now manages the pnpm version by default based on `package.json#package.manager`.](https://pnpm.io/npmrc#manage-package-manager-versions)

When:
- `pnpm@10` was detected
- `package.json#engines.pnpm` was set to `9.x`
- `package.json#packageManager` was set `pnpm@9.5.0`

Builds would incorrectly error:
```
Detected pnpm "10.x" is not compatible with the engines.pnpm "9.x" in your package.json. Either enable corepack with a valid package.json#packageManager value (https://vercel.com/docs/deployments/configure-a-build#corepack) or remove your package.json#engines.pnpm.
```

This PR fixes that scenario so `pnpm@10` uses `pnpm@9.5.0` without erroring.

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -959,25 +959,32 @@ export function getPathOverrideForPackageManager({
     projectCreatedAt
   );
 
-  if (!corepackPackageManager || !corepackEnabled) {
-    if (cliType === 'pnpm' && packageJsonEngines?.pnpm) {
+  const usingCorepack = corepackPackageManager && corepackEnabled;
+  if (usingCorepack) {
+    validateCorepackPackageManager(
+      cliType,
+      lockfileVersion,
+      corepackPackageManager,
+      packageJsonEngines?.pnpm
+    );
+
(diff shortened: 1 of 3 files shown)


trek (dismissed):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -959,25 +959,32 @@ export function getPathOverrideForPackageManager({
...
+      packageJsonEngines?.pnpm
+    );
+
+    // corepack is going to take care of it; do nothing special
+    return NO_OVERRIDE;
+  }
+
+  if (cliType === 'pnpm' && packageJsonEngines?.pnpm) {
+    const usingDetected =
+      detectedPackageManger?.pnpmVersionRange !== '10.x' ||
```
EndangeredMassa, line 977: **suggestion (non-blocking):** ‏Can we add a comment about why `10.x` is special here?

EndangeredMassa (approved):

PR #13084, opened 2025-02-21
Do not terminate due to long running waitUntil warning

mPyKen:
When deployed to vercel, I can use `waitUntil` to let a task run in background for the duration of the configured execution limit. This was not the case when running via `vercel dev`, it terminated exactly after 10s which is the hard-coded time for the waitUntil warning. This PR fixes this.


mPyKen:
seems to be obsolete now due to https://github.com/vercel/vercel/pull/13141. closing

PR #13087, opened 2025-02-21
[next]: Read expire values from prerender manifest

unstubbable:
In https://github.com/vercel/next.js/pull/76207 we've added `initialExpireSeconds` (and `fallbackExpire`) to the prerender manifest, analogous to `initialRevalidateSeconds` (or `fallbackRevalidate`). Both the revalidate and expire values together allow stale-while-revalidate semantics to be applied to static routes and fallback routes for ISR.

The configured expire values are added as `staleExpiration` property to the prerender outputs. The naming is a bit unfortunate, but that's a result from the revalidate value being transferred to the `expiration` property.

- [x] The added integration tests depend on a new Next.js canary version being released after https://github.com/vercel/next.js/pull/76207 is merged.

### DIFF
--- a/packages/build-utils/src/prerender.ts
+++ b/packages/build-utils/src/prerender.ts
@@ -3,6 +3,7 @@ import { Lambda } from './lambda';
 
 interface PrerenderOptions {
   expiration: number | false;
+  staleExpiration?: number;
   lambda?: Lambda;
   fallback: File | null;
   group?: number;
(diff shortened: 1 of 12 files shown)


ijjk (dismissed):
Looks good once we update the upstream API to allow this field

packages/build-utils/src/prerender.ts
```diff
@@ -3,6 +3,7 @@ import { Lambda } from './lambda';
 
 interface PrerenderOptions {
   expiration: number | false;
+  staleExpiration?: number;
```
ijjk, line 6: Let's break these `build-utils` changes out into their own PR so we can update independently

trek (approved):

ijjk (approved):

PR #13097, opened 2025-02-24
[tests] Fix vitest-e2e glob to include integration.test.ts

onsclom:
The existing glob `test/integration-*.test.ts` ignored `integration.test.ts`, so we haven't been running this test for a few months.

Since we removed node 16 support during this time, some tests needed to be updated/removed:
- `15-npm-8-legacy-peer-deps` is removed because it was only applicable for node 16 (https://github.com/vercel/vercel/pull/8550).
- `19-yarn-v2` is skipped because it vendors a 5 year old yarn version that only worked on node 16, but it is used for some unit tests


trek (dismissed):

onsclom:
Oops, looks like there are more failing tests in `integration.test.ts` to fix. Converting back to draft while I fix those.

trek (approved):

PR #13098, opened 2025-02-25
fix(repo): use workspace syntax

tknickman:
Swap to use workspace syntax for scoped workspace packages


chris-olszewski (dismissed):
This change makes sense to me and should be a no-op since `link-workspace-packages` is [true by default](https://pnpm.io/8.x/npmrc#link-workspace-packages) so we're just changing the specifier.

trek (approved):

packages/next/test/fixtures/30-monorepo-no-script/packages/www/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@vercel-crash-demo/www",
```
EndangeredMassa, line 2: Why was this change made?

PR #13107, opened 2025-02-26
[next] Ensure we lookup middleware function config properly

ijjk:
This ensures we properly lookup the middleware source file when loading function config specifically for node middleware handling.


ztanner (dismissed):

ztanner (approved):

packages/next/src/utils.ts
```diff
@@ -1572,7 +1572,10 @@ async function getSourceFilePathFromPage({
   const usesSrcDir = await usesSrcDirectory(workPath);
   const extensionsToTry = pageExtensions || ['js', 'jsx', 'ts', 'tsx'];
 
-  for (const pageType of ['pages', 'app']) {
+  for (const pageType of [
+    // middleware is not nested in pages/app
+    ...(page === 'middleware' ? [''] : ['pages', 'app']),
+  ]) {
```
AbhiShake1, line 1578: This doesnt seem to need the extra outer array

PR #13111, opened 2025-02-26
[errors] improve 'not found' error messaging to include scope

Zertsov:
There's some spots, currently, that obfuscate that you need to provide the `--scope` to the command, namely `vc remove` and `vc alias`. We _do_ have good messaging, but it wasn't being caught because of incorrect checks. Now, when the user doesn't provide the scope when they need to, we tell them that the deployment under scope `scope name` can't be found, rather than just the deployment

### DIFF
--- a/packages/cli/src/commands/remove/index.ts
+++ b/packages/cli/src/commands/remove/index.ts
@@ -106,7 +106,7 @@ export default async function remove(client: Client) {
 
     const [deploymentList, projectList] = await Promise.all<any>([
       Promise.all(
-        ids.map(idOrHost => {
+        ids.map(async idOrHost => {
           if (!contextName) {
             throw new Error('Context name is not defined');
           }
--- a/packages/cli/src/util/alias/get-aliases.ts
+++ b/packages/cli/src/util/alias/get-aliases.ts
@@ -1,5 +1,12 @@
 import type { Alias, PaginationOptions } from '@vercel-internals/types';
 import type Client from '../client';
+import {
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/remove/index.ts
```diff
@@ -106,7 +106,7 @@ export default async function remove(client: Client) {
 
     const [deploymentList, projectList] = await Promise.all<any>([
       Promise.all(
-        ids.map(idOrHost => {
+        ids.map(async idOrHost => {
```
Zertsov, line 109: this isn't related to the PR but it was a squiggly and I'm motivated by squiggles

packages/cli/src/util/alias/get-aliases.ts
```diff
@@ -20,6 +27,29 @@ export default async function getAliases(
...
+          context: contextName,
+        });
+      }
+      if (err.status === 403 && deploymentId) {
+        throw new DeploymentPermissionDenied(deploymentId, contextName);
+      }
+      if (err.status === 400 && err.message.includes('`id`') && deploymentId) {
+        throw new InvalidDeploymentId(deploymentId);
+      }
+    }
```
Zertsov, line 51: open to feedback on the last two conditionals here, figured they were ok to add here but this could use some different error handling overall

brookemosby (approved):
Lgtm, but don't have good feedback for the last two conditional checks

trek (approved):

onsclom (approved):

PR #13154, opened 2025-03-11
Force all resource creation to use web UI to use authorization flow

JakeUskoski:
Due to a change in how we handle resource creation, our CLI does not currently handle marketplace integration resource creation. It will instead always fail due to missing authorization.

As a temporary fix until the CLI can handle authorization, I'm sending all resource creation to the web UI so that the command doesn't take the user through the full creation flow just to fail them. The web UI properly handles the authorization flow, and thus doesn't have this issue.

### DIFF
--- a/.changeset/seven-pumpkins-sort.md
+++ b/.changeset/seven-pumpkins-sort.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+Force all integration resource creation to use web UI for authorization flow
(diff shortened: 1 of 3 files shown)


Dwi-ght:
https://github.com/vercel/vercel/pull/13154

Dwi-ght (commented):
https://github.com/vercel/vercel/pull/13149

trek (approved):

TooTallNate (approved):

PR #13158, opened 2025-03-11
Respect engines.pnpm when detecting pnpm 10

onsclom:
With a v9 `pnpm-lock.yaml`, it is ambiguous whether to use pnpm 10 or pnpm 9. Currently, newer projects prefer 10 and older projects prefer 9.

This PR changes the logic to not prefer 10 if the `engines.pnpm` does not respect it. This helps projects avoid `ERROR  Unsuported engine`.


Dwi-ght:
Please help me. i'm not used to this. and i need focus before i achieve but
is not happening because my woman is sick

Dwi-ght (approved):

Dwi-ght:
https://github.com/changesets/action/issues/453

Dwi-ght (changes requested):
https://github.com/mbahomaid/freeCodeCamp-labs/blob/main/LICENSE

PR #13172, opened 2025-03-17
Add license to @vercel/gatsby-plugin-vercel-builder

masutaka:
🔗 https://www.npmjs.com/package/@vercel/gatsby-plugin-vercel-builder

@vercel/gatsby-plugin-vercel-builder

I confirmed that every package.json in the other packages includes a "license" field.
e.g., @vercel/build-utils


dimitropoulos (approved):
I see that every other package is, indeed, `Apache-2.0` so this looks correct

PR #13176, opened 2025-03-17
[next] Fix operationType for mixed app router slug with ISR/SSR

ijjk:
This fixes incorrect `operationType` when an app router project has a slug that creates both dynamic paths and ISR paths since previously in pages router we assumed if a dynamic route had any prerendered paths then it's source route was ISR itself but this isn't the case for app router since some paths can be prerendered but then if any bail to dynamic the source route itself is dynamic. 

x-ref: [slack thread](https://vercel.slack.com/archives/C08J1UU858V/p1742252573040089?thread_ts=1742235138.645789&cid=C08J1UU858V)

### DIFF
--- a/.changeset/eight-rockets-swim.md
+++ b/.changeset/eight-rockets-swim.md
@@ -1,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+Fix operationType for mixed app router slug with ISR/SSR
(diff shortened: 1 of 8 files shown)


ijjk (commented):
Closing as this fix requires bigger changes that create less efficient outputs so we will need to investigate this further

PR #13192, opened 2025-03-21
Add authorization step to cli integration resource creation

JakeUskoski:
This work follows up #13154 to allow resource creation to use the CLI wizard again. All resource creation for integrations were forced to the Web UI due to all requests failing without the authorization step. Now, with this PR, we follow the steps for  automated authorization, and send the user to an authorization page on failure (to change the payment method) or when action is required (such as 3DS).

This PR requires an update to the frontend to enable redirect to the new manual authorize steps page.

Additionally, while fixing this bug, I found a separate bug revealed by trying to provision a Supabase resource. It turns out our system wasn't respecting default values of hidden metadata fields, so Supabase was returning an error due to missing metadata content.


packages/cli/src/commands/integration/add.ts
```diff
@@ -380,13 +392,95 @@ async function confirmProductSelection(
...
+    billingPlan.id,
+    metadata
+  );
+
+  if (!originalAuthorizationState.authorization) {
+    output.stopSpinner();
+    output.error(
+      'Error during authorization, failed to get an authorization state.'
+    );
+    throw new Error('Failed to get an authorization state');
```
trek: What action should a customer take if this see this error? Or do we catch elsewhere and error something else?

trek (dismissed):

packages/cli/src/commands/integration/add.ts
JakeUskoski: This particular situation means our API is broken/down/there was a critical failure. Realistically, I'm not even sure if redirecting them to the dashboard would help, because that could also be down.

trek: We might ask them to contact support in cases like this?

trek: But not blocking!

JakeUskoski: Added 🙂

trek (dismissed):

packages/cli/src/commands/integration/add.ts
```diff
@@ -380,13 +397,94 @@ async function confirmProductSelection(
...
+    'Payment validation requires manual action. Please complete the steps in your browser...'
+  );
+
+  handleManualVerificationAction(
+    teamId,
+    installation.id,
+    originalAuthorizationState.authorization.id
+  );
+
+  while (authorization.status !== 'succeeded') {
```
dvoytenko, line 457: Do we ever exit this cycle when `authorization.status === 'failed'` and user has given up trying to fix it? I.e. they closed the browser window and all. Can we even ever tell? Or is Ctrl+C is the only way? And if so, should we leaving some messages in the console for these states? E.g.

* "Manual authorization is requested. See browser".
* "Authorization failed. Change your payment method. See browser".

packages/cli/src/commands/integration/add.ts
```diff
@@ -380,13 +397,94 @@ async function confirmProductSelection(
...
+  if (!originalAuthorizationState.authorization) {
+    output.stopSpinner();
+    throw new Error(
+      'Failed to get an authorization state. If the problem persists, please contact support.'
+    );
+  }
+
+  let authorization = originalAuthorizationState.authorization;
+
+  while (authorization?.status === 'pending') {
```
dvoytenko: Do you need `?` here? Esp since you do `if (authorization.status === 'succeeded')` below?

dvoytenko (changes requested):

packages/cli/src/commands/integration/add.ts
JakeUskoski: We don't. This was a carry-over from a previous iteration 👍

packages/cli/src/commands/integration/add.ts
JakeUskoski: I've left it as a Ctrl+C situation, but we do have an output letting them know steps are required in their browser right before the call to open their browser is made:
```
output.spinner(
    'Payment validation requires manual action. Please complete the steps in your browser...'
  );
```

For payment method failed, as per our huddle conversation, I've made the "failed" flow kill the process. This is due to the authorization flow needing a fresh authorization & authorization ID when a payment method changes.

packages/cli/src/commands/integration/add.ts
```diff
@@ -265,14 +267,29 @@ async function provisionResourceViaCLI(
...
+    const authorizationId = await getAuthorizationId(
+      client,
+      teamId,
+      installation,
+      product,
+      metadata,
+      billingPlan
+    );
+
+    return provisionStorageProduct(
```
dvoytenko: Either do `await` here or move this call outside of `try/catch`, but IMHO it leads to bugs when `return retPromise()` is called inside the try/catch that has no chance of catching it.

packages/cli/src/commands/integration/add.ts
dvoytenko: > I've made the "failed" flow kill the process.

I expected this, but not seeing this anywhere. I.e. I'd expect either `while (authorization.status !== 'succeeded')` to be `status !== 'succeeded' && status !== 'failed'` or a `if (status === 'failed') throw new Error(...)` inside the cycle.

dvoytenko (changes requested):

packages/cli/src/commands/integration/add.ts
JakeUskoski: There's an earlier exit for failure in the initial request, but nothing for if required action fails. I've added in the late failure 👍

packages/cli/src/commands/integration/add.ts
JakeUskoski: Awaiting now 👍

trek (approved):

dvoytenko (approved):

PR #13207, opened 2025-03-31
trace writeBuildResult

marc-vercel:
Add a trace to the `writeBuildResult` operation. After finishing the build, in some cases, where we have lots of build results, we can spend quite some time (I've seen some projects where we spend up to 20 seconds).

Screenshot 2025-03-31 at 14 36 56

I do not think we will be able to improve it, although I will check it later on. The first step is getting observability.

### DIFF
--- a/packages/cli/src/commands/build/index.ts
+++ b/packages/cli/src/commands/build/index.ts
@@ -670,7 +670,7 @@ async function doBuild(
           .child('vc.builder.writeBuildResult', {
             buildOutputLength: String(buildOutputLength),
           })
-          .trace<Promise<Record<string, PathOverride> | undefined | void>>(() =>
+          .trace<Record<string, PathOverride> | undefined | void>(() =>
             writeBuildResult(
               repoRootPath,
               outputDir,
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/build/index.ts
```diff
@@ -657,22 +657,34 @@ async function doBuild(
...
-          builderPkg,
-          localConfig
-        ).then(
-          override => {
-            if (override) overrides.push(override);
-          },
-          err => err
-        )
+        builderSpan
+          .child('vc.builder.writeBuildResult', { buildOutputLength })
```
marc-vercel: @LukeSheard Do `attrs` equal to tags (https://github.com/vercel/vercel/blob/cc0b7194b119f72f59f77f9fba7e7a1188dac03c/packages/build-utils/src/trace/trace.ts#L107)? If not, is there a way to add tags to the traces?

LukeSheard: `attrs` correspond to tags on our internal tracing yes.

trek (approved):

PR #13211, opened 2025-04-01
Implement Turbopack build marker

timneutkens:
Implements detection for `next build --turbopack` to automatically set `TURBOPACK=1`.

When running `next build --turbopack` Next.js will write `.next/IS_TURBOPACK_BUILD` which can be used to detect if the build used Turbopack.

At runtime Next.js needs `TURBOPACK=1` in order to load the correct JS runtime that is compatible with Turbopack.

The environment variable will be required until Turbopack is the default build output of Next.js.

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -1088,6 +1088,18 @@ export type NextPrerenderedRoutes = {
   isLocalePrefixed: boolean;
 };
 
+export async function isTurbopackBuild(
+  outputDirectory: string
+): Promise<boolean> {
+  const isTurbopackBuildFile = path.join(outputDirectory, 'IS_TURBOPACK_BUILD');
+  const isTurbopackBuild: boolean = await fs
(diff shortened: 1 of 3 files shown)


packages/next/src/utils.ts
```diff
@@ -1088,6 +1088,18 @@ export type NextPrerenderedRoutes = {
   isLocalePrefixed: boolean;
 };
 
+export async function isTurbopackBuild(
+  outputDirectory: string
+): Promise<boolean> {
+  const isTurbopackBuildFile = path.join(outputDirectory, 'IS_TURBOPACK_BUILD');
+  const isTurbopackBuild: boolean = await fs
```
huozhi, line 1095: we once changed all the fs.access to fs.existsSync as it's slightly faster but non blocking

huozhi (approved):

trek (approved):

ijjk (commented):
I believe we can close this out now as we handled it differently

PR #13214, opened 2025-04-02
[cli] refresh `VERCEL_OIDC_TOKEN`

markandrus:
There are two challenges using the VERCEL_OIDC_TOKEN environment variable today:

- **It contains a token which expires after 12 hours in development.** This means, if `vc dev` is left running for longer than 12 hours, the token expires, leading to errors in the project.
- **It is not possible to update the token in `.env.local` for linked projects,** because the value pulled from Vercel takes precedent.

This PR addresses these challenges by

- **Excluding VERCEL_OIDC_TOKEN from the environment variables passed to DevServer.** Instead, the token can be read from and updated in `.env.local`.
- **Setting up background refreshes of the VERCEL_OIDC_TOKEN.** We parse the "exp" claim out of the token and set a timeout to update VERCEL_OIDC_TOKEN in `.env.local`.

In order to be as minimally invasive as possible only `.env.local` files beginning with "# Created by Vercel CLI" are considered. If a `.env.local` does not exist, it will be created.

TODO

- [x] Changeset
- [x] If a `.env.local` is created, we should perform the same `.gitignore` logic as `vc env pull`.
- [x] If there is a `.env.local`, but it doesn't begin with "# Created by Vercel CLI", we shouldn't delete the VERCEL_OIDC_TOKEN from the `envValues` passed to DevServer, because we aren't assuming we can update it.
- [x] Do we need to consider any files other than `.env.local`? No.
- [x] Should any of these logs be non-debug logs? Yes. We will print the delta string and `.gitignore` notice that `vc env pull` does.

### DIFF
--- a/packages/cli/src/util/env/refresh-oidc-token.ts
+++ b/packages/cli/src/util/env/refresh-oidc-token.ts
@@ -16,7 +16,6 @@ import stamp from '../../util/output/stamp';
 import { patchEnvFile, type PatchEnvFileResult } from './patch-env-file';
 import { wasCreatedByVercel } from './was-created-by-vercel';
 import type { DevTelemetryClient } from '../telemetry/commands/dev';
-import { buildDeltaString } from './diff-env-files';
 
 const FILENAME = '.env.local';
 const REFRESH_BEFORE_EXPIRY_MS = ms('15m');
@@ -105,27 +104,33 @@ export async function refreshOidcToken(
 
     // If the OIDC token isn't already expired, patch .env.local.
     if (expiresAfterMs > 0) {
-      const newEnv = { [VERCEL_OIDC_TOKEN]: oidcToken };
       let result: PatchEnvFileResult | undefined;
+
       try {
         result = await patchEnvFile(
           client.cwd,
           link.repoRoot ?? client.cwd,
           FILENAME,
-          newEnv
+          {
+            [VERCEL_OIDC_TOKEN]: oidcToken,
+          }
         );
         telemetry.trackOidcTokenRefresh(refreshCount);
       } catch (error) {
         output.debug(`Failed to patch ${VERCEL_OIDC_TOKEN} in ${FILENAME}`);
       }
+
       if (controller.signal.aborted) {
         return;
       }
-      if (result !== undefined) {
-        const deltaString = buildDeltaString({}, newEnv);
-        output.print('\n' + deltaString);
 
-        const { exists, isGitIgnoreUpdated } = result;
+      if (result) {
+        const { exists, isGitIgnoreUpdated, deltaString } = result;
+
+        output.log('Refreshed VERCEL_OIDC_TOKEN');
+
+        if (deltaString) output.print('\n' + deltaString);
+
         output.print(
           `${prependEmoji(
             `${exists ? 'Updated' : 'Created'} ${chalk.bold(FILENAME)} file ${
--- a/packages/cli/src/util/env/refresh-oidc-token.ts
+++ b/packages/cli/src/util/env/refresh-oidc-token.ts
@@ -1,6 +1,7 @@
 import { decodeJwt } from 'jose';
 import ms from 'ms';
 import { performance } from 'perf_hooks';
+import type { ProjectLinked } from '@vercel-internals/types';
 import output from '../../output-manager';
 import type Client from '../../util/client';
 import {
@@ -16,7 +17,7 @@ const VERCEL_OIDC_TOKEN = 'VERCEL_OIDC_TOKEN';
 
 export async function refreshOidcToken(
   client: Client,
-  projectId: string,
+  link: ProjectLinked,
   envValues: Record<string, string>,
   source: EnvRecordsSource
 ): Promise<() => void> {
@@ -46,7 +47,7 @@ export async function refreshOidcToken(
       const envValues = await pullEnvValuesUntilSuccessful(
         controller.signal,
         client,
-        projectId,
+        link.project.id,
         source,
         THROTTLE_MS
       );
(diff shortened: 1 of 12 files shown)


shaper (commented):
The functionality LGTM and the code/case-consideration looks thorough.

A few thoughts:
- what happens if the user runs `vc env pull` while a refresh is underway, or vice-versa? seems highly unlikely but worth considering.
- other edge case ideas to consider if they happen while `vc` is running:
  - what happens if the project did not have OIDC enabled initially (older project, or user actively disabled it), but then the user enables it?
  - what if they had OIDC enabled on `vc` start, but then they disable it?
  - if they change their project OIDC setting from `team` to `global` in API or dashboard, will this alter the OIDC token, will/should auto-refresh find out somehow and refresh it, or do we rely on user manually re-running `vc env pull` or restarting `vc`?
- tests for the new logic are of course valuable, could be useful to make testing easier to break apart a bit e.g. move `patchLocalEnv` to a separate file, but this is also a project-code-style question.
- I am also new to the code base:
  - I wonder if there are utility routines we should be using e.g. `tryRead` or patching elsewhere in the code base.
  - it seems possible pushing this into the dev server could be an alternate approach, it may have periodic tasks/phases, just a thought that came to mind seeing how we add complexity to shutdown/stop and roll our own timeout-loop
  - should we put this behind a flag (non-rhetorical and not looking to create extra work, actually checking w/ you + team on how to approach this kind of change)
  - should we add telemetry tracking (I see it [here](https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/env/index.ts#L95) for example) but noting each time we refresh could just be noise

packages/cli/src/util/env/refresh-oidc-token.ts
```diff
@@ -0,0 +1,255 @@
...
+    controller.abort();
+    clearTimeout(timeout);
+  };
+}
+
+/**
+ * The OIDC token is a JSON Web Token (JWT). Decode the JWT and get its numeric
+ * "exp" claim, returning `null` on error.
+ */
+function getExpFromOidcToken(oidcToken: string): number | null {
```
balazsorban44: we can use `jose` instead of manual decoding, see https://github.com/vercel/vercel/blob/5c7f57fdd9339715e186c0281f52d989ca8bbb13/packages/cli/src/util/oauth.ts#L387

markandrus: Good idea. I used `decodeJwt` from jose directly, since I don't think we need to validate.

markandrus:
@shaper thanks for the feedback!

> - what happens if the user runs vc env pull while a refresh is underway, or vice-versa? seems highly unlikely but worth considering.

Running `vc env pull` before a refresh will pull a new token with an expiry time further into the future than the one the refresh process knows about. That's OK. It means the refresh will run sooner than strictly necessary, but it will never result in an expired token.

The only problem I can imagine is if the file operations of the `vc env pull` and refresh interleave. For example, the refresh reads `.env.local` before writing. If `vc env pull` ran in between the read and write, we'd end up writing a stale `.env.local`. Additionally, problems could occur if `vc env pull` and the refresh write simultaneously. I think this is extremely unlikely, though, and not worth engineering around.

> - what happens if the project did not have OIDC enabled initially (older project, or user actively disabled it), but then the user enables it?

Any `vc dev` command executed while OIDC was disabled will not perform refreshes. The user will have to restart `vc dev` after enabling OIDC.

> - what if they had OIDC enabled on vc start, but then they disable it?

`vc dev` will cease refreshing the OIDC token. It will not be removed from `.env.local` and will be allowed to expire.

> - if they change their project OIDC setting from team to global in API or dashboard, will this alter the OIDC token, will/should auto-refresh find out somehow and refresh it, or do we rely on user manually re-running vc env pull or restarting vc?

I think we should rely on the user manually re-running `vc env pull` or `vc dev`. We don't have a way to get proactively notified of such a change, which means we would need to poll, and this would put extra burden both on the user's machine and on Vercel APIs. I think we should not do this. It's the same problem as toggling OIDC after running `vc dev`.

If someone runs `vc env pull`, adding VERCEL_OIDC_TOKEN, then maybe we should watch for that and start refreshing the token again?

> - tests for the new logic are of course valuable, could be useful to make testing easier to break apart a bit e.g. move patchLocalEnv to a separate file, but this is also a project-code-style question.
> - I am also new to the code base:
>     - I wonder if there are utility routines we should be using e.g. tryRead or patching elsewhere in the code base.

I will refactor and add some tests.

> - it seems possible pushing this into the dev server could be an alternate approach, it may have periodic tasks/phases, just a thought that came to mind seeing how we add complexity to shutdown/stop and roll our own timeout-loop

~I think the benefit of a closer DevServer integration would be avoiding writing to disk. I'll look into it more closely.~ The DevServer in src/util/dev/server.ts is not actually responsible for handling changes to `.env.local`. That's something the Next.js DevServer (and other frameworks like SvelteKit) handle. It also lacks lifecycle hooks we would want, so I think it's best to keep this implemented out-of-band.

> should we put this behind a flag (non-rhetorical and not looking to create extra work, actually checking w/ you + team on how to approach this kind of change)

I think it's a good idea. For now I propose a local environment variable, `VERCEL_REFRESH_OIDC_TOKEN`, which should be set to enable the feature. Eventually, we can remove it and have it on by default. Will ask around, too.

> should we add telemetry tracking (I see it [here](https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/env/index.ts#L95) for example) but noting each time we refresh could just be noise

I added a simple count of the number of times we refresh (`oidc-token-refresh`). I think it's a good indicator for if the feature is getting used.

Also wrote up some manual test scenarios [here](https://gist.github.com/markandrus/ae8f5c3055e3f13aaf43f12733331205).

PR #13204, opened 2025-03-28
Add simple balance command for prepayment-based integrations

JakeUskoski:
This PR adds a new function to list the balances and thresholds set for a given installed integration. This is the first of multiple new commands and flows meant to support prepayment-style integrations such as the new AI marketplace integrations.

PR #13218, opened 2025-04-03
Create new threshold for prepayment top-up via CLI

JakeUskoski:
This PR adds a new subcommand to the `integration-resource` command in the CLI. This functionality serves both resource-level and installation-level threshold creation, in that if the resource uses an installation it'll change from targeting the resource to targeting the parent integration installation.

Note: This relies on some of the functionality developed in #13204


packages/cli/src/commands/integration-resource/create-threshold.ts
```diff
@@ -0,0 +1,366 @@
...
+  if (targetedResource.billingPlan?.type !== 'prepayment') {
+    output.error(
+      `The resource ${chalk.bold(resourceName)} is not a prepayment-based resource.`
+    );
+    return 1;
+  }
+
+  // Assert spend is within billing plan limits
+  const minimumSpend =
+    parseFloat(targetedResource.billingPlan.minimumAmount ?? '0.50') * 100;
```
trek, line 92: I assume these match some values also validated in the API? (not blocking)

JakeUskoski: Yeah, this is a little minimum-amount limit we generally apply to avoid 1 cent purchases and all that. It's validated on the API layer, but I'm doing some fast-failure here to provide a useful error message and save some wait time in execution.

trek: Just wanted to be sure we had some defense in depth on this. I wish we had a good pattern for sharing these values. I've seen them drift over time, but that's not a problem to solve today!

trek (dismissed):

packages/cli/src/commands/integration-resource/create-threshold.ts
```diff
@@ -0,0 +1,366 @@
...
+  const entityTextReference = props.isInstallationLevel
+    ? `installation ${chalk.bold(props.resource.product?.name)}`
+    : `resource ${chalk.bold(props.resource.name)}`;
+  if (props.existingThreshold) {
+    const shouldOverwriteThreshold =
+      props.skipConfirmWithYes ||
+      (await props.client.input.confirm(
+        `The ${entityTextReference} already has a threshold. (minimum: $${props.existingThreshold.minimumAmountInCents / 100}, spend: $${props.existingThreshold.purchaseAmountInCents / 100}, limit: ${props.existingThreshold.maximumAmountPerPeriodInCents ? `$${props.existingThreshold.maximumAmountPerPeriodInCents / 100}` : 'none set'}). Do you want to overwrite it?`,
+        true
+      ));
```
RPDeshaies, line 301: > [!NOTE]
> Split to make this easier to read (having the promise resolution directly in a ternary is what is bugging me here I think)

packages/cli/src/commands/integration-resource/create-threshold.ts
```diff
@@ -0,0 +1,366 @@
...
+      return 0;
+    }
+  }
+
+  const thresholdCreationConfirmed =
+    props.skipConfirmWithYes ||
+    (await props.client.input.confirm(
+      `Are you sure you want to create a threshold for the ${entityTextReference} with minimum $${props.minimum / 100}, spend $${props.spend / 100}, and limit $${props.limit / 100}?`,
+      true
+    ));
```
RPDeshaies, line 331: > [!NOTE]
> Same idea

packages/cli/src/commands/integration-resource/create-threshold.ts
```diff
@@ -0,0 +1,366 @@
...
+  if (
+    props.existingBalance &&
+    props.existingBalance.currencyValueInCents < props.minimum
+  ) {
+    const shouldOverwriteBalance =
+      props.skipConfirmWithYes ||
+      (await props.client.input.confirm(
+        `The minimum threshold is higher than the current balance of $${props.existingBalance.currencyValueInCents / 100}. Are you sure?`,
+        true
+      ));
```
RPDeshaies, line 318: > [!NOTE]
>  Same idea

RPDeshaies (approved):

packages/cli/src/commands/integration-resource/create-threshold.ts
JakeUskoski: For simplicity I'm using this to only call the confirm function if they haven't skipped confirmation, but resolving to the same constant for logic simplicity.

trek (approved):

PR #13222, opened 2025-04-04
[next] ensure app router 404 page is still included in functions when using pages i18n

ztanner:
When using pages router, we bundle the localized static 404 file with every function output, to ensure that if not found behavior is triggered from a function, that it's available.

However, Next.js prioritizes serving the app router 404 page over the pages router 404 page. As a result, when triggering a not found error from an app router function, it'd trigger a 500 and crash the function because it only knows about `en/404.html` rather than `404.html` (as an example).

This ensures that we bundle the app router 404 page rather than the localized pages router 404 page if i18n + appDir is used. Separately in Next.js we need to ensure the `pages-manifest` is pointing localized paths to the correct not found file as well. This is accomplished in https://github.com/vercel/next.js/pull/77905

Without that change, this just solves the issue of throwing a 500 rather than resolving a 404 page, but doesn't fix the app router/pages router specificity issue. 

### Test plan
- Land this change
- Land the fix in Next.js
- Update the test here

In the meantime, I manually validated it here:
https://vtest314-e2e-tests-bkc89ab40-ztanner.vercel.app/app-dir/foo


ijjk (dismissed):

trek (dismissed):

ijjk (approved):

packages/next/test/fixtures/00-app-dir-not-found-pages-interop/app/app-dir/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,44 @@
+import { notFound } from 'next/navigation';
+
+export async function generateStaticParams() {
+  return [];
+}
+
+async function validateSlug(slug: string[]) {
+  try {
+    const isValidPath =
+      slug.length === 1 && (slug[0] === 'about' || slug[0] === 'contact');
```
Comment (huozhi) [inline packages/next/test/fixtures/00-app-dir-not-found-pages-interop/app/app-dir/[[...slug]]/page.tsx:10]: readability wise: "contact" seems not used, maybe we can give an easier testing slug `isValidPath = slug.length === 1 && slug[0] === "valid"`

huozhi (approved):

PR #13226, opened 2025-04-07
[cli] refresh `VERCEL_OIDC_TOKEN`

markandrus:
If `vc dev` is left running for longer than 12 hours, the `VERCEL_OIDC_TOKEN` environment variable can expire, leading to errors in applications that depend on it.

This PR updates `vc dev` to proactively fetch a new `VERCEL_OIDC_TOKEN` 15 minutes before its expiry and restart the dev server, bypassing this issue. This is a simpler approach than the one in https://github.com/vercel/vercel/pull/13214, which updated `.env.local` on disk.

Note that if OIDC is disabled when `vc dev` is run or if OIDC is disabled at the time of a refresh, then refreshes will be disabled. Additionally, care is taken not to refresh `VERCEL_OIDC_TOKEN` too frequently (at most once every minute). This should not be a problem in practice, because the tokens issued in development have a 12 hour expiration.

For manual testing, I set the environment variables

```sh
export REFRESH_VERCEL_OIDC_TOKEN_BEFORE_EXPIRY_MILLIS=12h
export REFRESH_VERCEL_OIDC_TOKEN_THROTTLE_MILLIS=10s
```

to force `vc dev` to consider whatever token was fetched already expired and to speed up fetching the next one.


trek (dismissed):

trek:
I like this a lot more than #13214!

trek (dismissed):

packages/cli/src/util/env/refresh-oidc-token.ts
```diff
@@ -0,0 +1,115 @@
...
+    const oidcToken = envValues[VERCEL_OIDC_TOKEN];
+    if (!oidcToken) {
+      output.debug(`${VERCEL_OIDC_TOKEN} is absent; disabling refreshes`);
+      return;
+    }
+
+    // Otherwise, extract the "exp" claim.
+    const now = clock();
+    const { exp } = decodeJwt(oidcToken);
+    const expiresAfterMs = exp !== undefined ? exp * 1000 - now : undefined;
```
shaper: nit: would name `Ms` as `Millis` and `S` as `Secs` for better readability throughout PR.

packages/cli/src/util/dev/server.ts
```diff
@@ -2221,10 +2221,10 @@ export default class DevServer {
     return this.caseSensitive;
   }
 
-  async runDevCommand() {
+  async runDevCommand(restart: boolean = false) {
```
shaper: nit: maybe `forceRestart`?

packages/cli/src/util/env/refresh-oidc-token.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // Schedule to refresh the OIDC token shortly before it expires, but not
+    // too frequently (wait at least THROTTLE_MS).
+    let refreshAfterMs = Math.max(0, expiresAfterMs - REFRESH_BEFORE_EXPIRY_MS);
+    if (now + refreshAfterMs - lastPulledEnvAt < THROTTLE_MS) {
+      refreshAfterMs = THROTTLE_MS;
+    }
+
+    const expiresAfterS = Math.abs(Math.round(expiresAfterMs / 1_0000));
```
shaper: typo? divisor has one extra `0`. Maybe we should use a `millisToSecs()` everywhere.

packages/cli/src/util/dev/server.ts
```diff
@@ -2221,10 +2221,10 @@ export default class DevServer {
...
     const { devCommand, cwd } = this;
 
-    if (devCommand === this.currentDevCommand) {
+    if (devCommand === this.currentDevCommand && !restart) {
```
shaper: Am I correctly following this PR in that we no longer actively write a refreshed token to disk at all. Instead, we refresh/fetch and store in memory and restart the dev server with the token updated in the env as appropriate?

I can see how this simplifies a lot of things, so just want to make sure I have the plan in mind.

packages/cli/src/util/env/refresh-oidc-token.ts
shaper, line 1: The logic in this file feels worth testing, I can imagine it would be tricky given use of timestamps, network, etc. but suggest trying using v0 or similar to generate an initial set of test cases, especially if there are existing unit tests you can include as a reference for style/test-infra.

shaper (dismissed):

packages/cli/src/util/dev/server.ts
markandrus: Yes, that's right! 👍

markandrus:
In order to avoid a test-e2e timeout related to the new functionality in `vc dev`, I udpated `@types/node` to 18 (matching `engines` in package.json) and use `setTimeout` from `node:timers/promises`, passing an AbortSignal. This has been available since Node.js 15, so I think it should be fine?

trek (approved):

PR #13234, opened 2025-04-08
[next] Disable middleware next data resolving for static rewrites

ijjk:


ijjk (commented):
Closing as opted for beforeFiles fix instead

PR #13238, opened 2025-04-10
fix(cli): address feedback

balazsorban44:
- removed duplicate "Waiting for authentication..." when hitting ENTER
- Simplified the text
- Made lines shorter so it breaks nicely

Before:
image
image

After:
image
image


packages/cli/src/commands/login/future.ts
```diff
@@ -125,6 +126,8 @@ export async function login(client: Client): Promise<number> {
 
       if (tokenError) return tokenError;
 
+      o.print(eraseLines(2));
```
Zertsov, line 129: I'm a bit unfamiliar - what's this one for?

Zertsov (approved):

PR #13258, opened 2025-04-18
[client] Include microfrontend config for prebuilt deploys

onsclom:
Prebuilt microfrontend projects currently see `Warning: This project is in a microfrontends group, but a microfrontends config file (microfrontends.json) was not found.` because the `${projectRoot}/microfrontends.json` is not uploaded. This change includes that file for prebuilt deploys.

Example deployment with this change: https://vercel.com/microfrontends-vtest314/nextjs-app-marketing/Cd3TrmfQUjYibyvSK8zSyD5Ajicy

### DIFF
--- a/packages/client/src/utils/index.ts
+++ b/packages/client/src/utils/index.ts
@@ -7,7 +7,7 @@ import { pkgVersion } from '../pkg';
 import { NowBuildError } from '@vercel/build-utils';
 import { VercelClientOptions, VercelConfig } from '../types';
 import { Sema } from 'async-sema';
-import { readFile } from 'fs-extra';
+import { pathExists, readFile } from 'fs-extra';
 import readdir from './readdir-recursive';
 
 type Ignore = ReturnType<typeof ignore>;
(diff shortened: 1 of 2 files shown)


tknickman (approved):
A test would be great as well, but lgtm

trek (approved):

packages/client/src/utils/index.ts
```diff
@@ -121,6 +125,16 @@ export async function buildFileTree(
           }
         })
       );
+
+      const microfrontendConfigPath = join(
+        path,
+        rootDirectory || '',
+        'microfrontends.json'
```
emspishak, line 132: This could also be called `microfrontends.jsonc`. (For example: https://github.com/vercel/front/blob/main/apps/vercel-site/microfrontends.jsonc)

We have a `findConfig` function to help with this at https://github.com/vercel/microfrontends/blob/main/packages/microfrontends/src/config/microfrontends/utils/find-config.ts , although it's understandable if you don't want to depend on our package just for that...

mknichel: I updated the build container to use this logic if it's possible to use the same logic: https://github.com/vercel/api/pull/39866

mknichel (commented):
Thanks! Looks like it works great!

PR #13263, opened 2025-04-21
[client][cli] Improve microfrontends config handling

onsclom:
Follow up PR for https://github.com/vercel/vercel/pull/13258

- Uses `findConfig` and `inferMicrofrontendsLocation` from `@vercel/microfrontends` for better detection
  - `@vercel/microfrontends` depends on `jsonc-parser` which does not work well with esbuild: https://github.com/evanw/esbuild/issues/1619, so added a custom plugin to handle that
- Adds unit test tests

### DIFF
--- a/packages/cli/scripts/build.mjs
+++ b/packages/cli/scripts/build.mjs
@@ -32,11 +32,8 @@ const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
 const externals = Object.keys(pkg.dependencies || {});
 await esbuild({
   bundle: true,
-  external: [
-    ...externals,
-    // https://github.com/evanw/esbuild/issues/1619
-    'jsonc-parser',
-  ],
+  external: externals,
+  mainFields: ['module', 'main'],
 });
 
 // Copy a few static files into `dist`
--- a/packages/cli/scripts/build.mjs
+++ b/packages/cli/scripts/build.mjs
@@ -32,7 +32,11 @@ const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
 const externals = Object.keys(pkg.dependencies || {});
 await esbuild({
   bundle: true,
-  external: externals,
+  external: [
+    ...externals,
+    // https://github.com/evanw/esbuild/issues/1619
+    'jsonc-parser',
+  ],
 });
 
 // Copy a few static files into `dist`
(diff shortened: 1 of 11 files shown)


packages/client/src/utils/index.ts
```diff
@@ -126,12 +127,10 @@ export async function buildFileTree(
         })
       );
 
-      const microfrontendConfigPath = join(
```
mknichel, line 129: Should we also add support for `inferMicrofrontendsLocation`?

packages/cli/scripts/build.mjs
```diff
@@ -33,6 +33,8 @@ const externals = Object.keys(pkg.dependencies || {});
 await esbuild({
   bundle: true,
   external: externals,
+  // https://github.com/evanw/esbuild/issues/1619
+  mainFields: ['module', 'main'],
```
onsclom: `@vercel/microfrontends` depends on `jsonc-parser` which is not friendly with esbuild. `mainFields: ['module', 'main'],` is a bandaid solution, so open to better suggestions here.

mknichel: Oof. Do you have a link to the error for the failed build?

I wonder if the `onResolve` approach may be better. I'd be worried about swapping `mainFields` since I've had issues with that in the past.

mknichel: "Write an esbuild plugin using onResolve to swap out just this one file (won't affect other packages)" (https://github.com/evanw/esbuild/issues/1619)

trek (dismissed):

packages/cli/scripts/build.mjs
onsclom: Here's the error: https://vercel.com/zero-conf-vtest314/create-react-app/9ExKLrj6gpod6xyerJueGtYVg9v6?filter=errors#L8

Got this working with a plugin!

packages/client/src/utils/index.ts
```diff
@@ -7,8 +7,12 @@ import { pkgVersion } from '../pkg';
 import { NowBuildError } from '@vercel/build-utils';
 import { VercelClientOptions, VercelConfig } from '../types';
 import { Sema } from 'async-sema';
-import { pathExists, readFile } from 'fs-extra';
+import { readFile } from 'fs-extra';
 import readdir from './readdir-recursive';
+import {
+  findConfig,
```
kitfoster, line 13: this should probably be named `findMicrofrontendsConfig` in the package, but I think this makes the code easier to read below?

```suggestion
  findConfig as findMicrofrontendsConfig,
```

kitfoster (approved):
Keen for @tknickman's thoughts on `jsonc-parser`

tknickman (approved):
Looks good

PR #13267, opened 2025-04-23
[cli] use `dot` directly instead of `dot-packer` cli

onsclom:
This `dot-packer` cli usage is blocking https://github.com/vercel/vercel/pull/13263. Not totally sure why this was using the `dot-packer` cli instead of `dot.process()` before, but I manually checked the compiled templates in `packages/cli/src/dev/templates` are the same before and after.

### DIFF
--- a/.changeset/beige-tigers-relax.md
+++ b/.changeset/beige-tigers-relax.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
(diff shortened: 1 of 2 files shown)


trek (approved):
Does this potentially close https://github.com/vercel/vercel/pull/12854 (or maybe just make that PR smaller?)

PR #13271, opened 2025-04-25
Add instrumentation detection to the `build` command

agadzik:
Adding this so that we can prevent automatic fetch instrumentation in the Vercel Function Runtime environment when we detect instrumentation inside of a project.  We assume that if a project has one of the following dependencies detected that they are already instrumenting `fetch` separately.

- `@vercel/otel@1.11.0` or higher
- `@opentelemetry/api@1.7.0` or higher
- `@opentelemetry/sdk-trace-node@1.19.0` or higher


trek (approved):

packages/fs-detectors/src/detect-instrumentation.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const VERCEL_OTEL = '@vercel/otel';
+const OPENTELEMETRY_SDK_TRACE_NODE = '@opentelemetry/sdk-trace-node';
+const OPENTELEMETRY_API = '@opentelemetry/api';
+
+const tracingDependencies: InstrumentationDependency[] = [
+  { name: VERCEL_OTEL, minVersion: '1.11.0' },
+  { name: OPENTELEMETRY_SDK_TRACE_NODE, minVersion: '1.19.0' },
+  { name: OPENTELEMETRY_API, minVersion: '1.7.0' },
```
dvoytenko, line 16: `@opentelemetry/api` is not a very good indicator since this is just an API and not an instrumentation. Can we remove it?

PR #13274, opened 2025-04-28
Add Blade to framework list

leo:
This change adds [Blade](https://github.com/ronin-co/blade) to the list of frameworks! 🚀

Blade is a lightweight React framework for easily building fast web apps.

[Demo](https://leo.im)


.changeset/fine-apples-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/frameworks': minor
+---
+
+Add "blade" framework entry
```
leo, line 5: I used the same text as [this commit](https://github.com/vercel/vercel/commit/f9d8407866ac95db70dfd961cc51a6d2df233efa).

anthonyshew:
Hi, Leo, thanks for the PR.

We're unfortunately not able to add more zero-config frameworks right now. While first-time adding a framework is trivial, the ongoing support cost of the fast-moving JavaScript ecosystem being zero-configuration on Vercel is proving its weight. We're actively thinking about how to improve our ability to handle this.

With that said, we can leave this PR open for when the time comes or we can close and I'll DM you on X when we're ready again. I'm not sure which you'd prefer.

PR #13275, opened 2025-04-28
[next] Fix: Segment prefetching when PPR disabled

acdlite:
When PPR is disabled for a route, and the route is not part of the prerender manifest, we must reroute it to an RSC request for the entire page. Next.js will output this information into the routes manfiest, and already handles this correctly for dynamically-routed routes, but not for statically-routed routes.

This updates the builder to handle statically-routed routes, too, using the same logic that already exists for dynamically-routed routes.

See https://github.com/vercel/next.js/pull/78568 for more context.


wyattjoh (dismissed):

packages/next/test/fixtures/00-app-dir-segment-prefetch-incremental-conflicting-routes/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "next": "https://vercel-packages.vercel.app/next/commits/2fe49e013be98130279dd13b8199afcde23f6f56/next",
```
acdlite: I'll change this to "canary" before landing. Depends on changes in https://github.com/vercel/next.js/pull/78568.

packages/next/src/server-build.ts
```diff
@@ -1633,6 +1634,14 @@ export async function serverBuild({
     (middleware.staticRoutes.length > 0 || nodeMiddleware) &&
     semver.gte(nextVersion, NEXT_DATA_MIDDLEWARE_RESOLVING_VERSION);
 
+  const staticSegmentRoutes = isAppClientSegmentCacheEnabled
+    ? await getStaticSegmentRoutes({
+        entryDirectory,
+        isDev: false,
```
ijjk: Should we omit this since would always be `false`?

ijjk (dismissed):

ijjk (approved):

PR #13317, opened 2025-05-09
feat(auth): implement token refresh

balazsorban44:
This PR implements an automatic token refresh mechanism, a follow-up on #12098.

During initial login, we will save new tokens and their expiry in the auth config file.

Then, during API requests, we will insert the following logic:

```mermaid
graph TD;
    A[Has access_token expired?]-->|Yes| B;
    A-->|No| C[Continue];
    B[Has refresh_token expired?]-->|Yes| D;
    B-->|No| E;
    D[Delete auth config];
    D--> C;
    E[Perform token refresh]-->|Successful| F;
    E-->|Unsuccessful| D
    F[Update auth config]-->C
```

This will ensure that if a request can be made, we proceed as usual, otherwise, the user gets logged out, so as not to leave them in an undesirable state.

I have refactored the token verification code so we can easily swap it out with a [OAuth Token Introspection](https://datatracker.ietf.org/doc/html/rfc7009) endpoint later on, making the use of JWT purely an implementation detail.


fivemeyestoreserver:
The documentation in this repository is excellent! Very clear and comprehensive.

ayestoro:
Very clean code with good test coverage. This is a great example of best practices.

internals/types/index.d.ts
```diff
@@ -26,13 +26,32 @@ export interface JSONObject {
...
+  /** An Vercel token retrieved using the legacy authentication flow. */
+  token?: string;
+  type?: 'legacy';
+}
+
+interface OAuthAuthConfig extends AuthConfigBase {
+  /** An `access_token` obtained using the OAuth Device Authorization flow.  */
+  token?: string;
+  /** The absolute time when the {@link OAuthAuthConfig.token} expires */
+  expiresAt?: number;
```
markandrus: Clarify units? Seconds versus milliseconds?

internals/types/index.d.ts
```diff
@@ -26,13 +26,32 @@ export interface JSONObject {
...
+  refreshToken?: string;
+  /** The absolute time when the {@link OAuthAuthConfig.refreshToken} expires */
+  refreshTokenExpiresAt?: number;
```
markandrus: Clarify units? Seconds versus milliseconds?

packages/cli/src/util/client.ts
```diff
@@ -132,7 +153,97 @@ export default class Client extends EventEmitter implements Stdio {
...
+      if (output.isDebugEnabled()) output.prettyError(tokenError);
+      return;
+    }
+
+    const [accessTokenError] = await verifyJWT(token.access_token);
+    if (accessTokenError) {
+      output.debug('Error verifying access token, skipping token refresh.');
+      if (output.isDebugEnabled()) output.prettyError(accessTokenError);
+      return;
+    }
```
markandrus: Is it possible to delete the `verifyJWT` code and let the client treat the access token and refresh token as opaque strings? By depending on the format here, it will be harder to change things in the future.

packages/cli/src/util/oauth.ts
```diff
@@ -393,6 +412,7 @@ export async function verifyJWT(
       issuer: 'https://vercel.com',
       audience: ['https://api.vercel.com', 'https://vercel.com/api'],
     });
+    if (!payload.exp) throw new Error('Missing `exp` claim in JWT');
```
markandrus: This encodes a requirement on our token format that need not necessarily hold. Can we remove it?

Also, I think it would make since to relax this function as much as possible. I.e., just `jwtDecode` not `jwtVerify`. WDYT?

packages/cli/src/util/client.ts
```diff
@@ -132,7 +153,97 @@ export default class Client extends EventEmitter implements Stdio {
...
+    this.updateAuthConfig({
+      type: 'oauth',
+      token: token.access_token,
+      expiresAt: Date.now() + token.expires_in * 1000,
+    });
+
+    if (token.refresh_token) {
+      const [refreshTokenError, refreshToken] = await verifyJWT(
+        token.refresh_token
+      );
```
markandrus: Same suggestion here.

packages/cli/src/commands/login/future.ts
```diff
@@ -126,40 +121,52 @@ export async function login(client: Client): Promise<number> {
...
       // user is not currently authenticated on this machine
       const isInitialLogin = !client.authConfig.token;
 
-      // Save the user's authentication token to the configuration file.
-      client.authConfig.token = token.access_token;
-      error = undefined;
-
       const [accessTokenError, accessToken] = await verifyJWT(
         token.access_token
       );
```
markandrus: I guess this must remain, since we currently depend on `team_id`.

packages/cli/src/commands/login/future.ts
```diff
@@ -126,40 +121,52 @@ export async function login(client: Client): Promise<number> {
...
+
+      if (accessToken.team_id) o.debug('Current team updated');
+      else o.debug('Current team deleted');
+
+      client.updateConfig({ currentTeam: accessToken.team_id });
+
+      if (token.refresh_token) {
+        const [refreshTokenError, refreshToken] = await verifyJWT(
+          token.refresh_token
+        );
```
markandrus: This `verifyJWT` call could be removed?

packages/cli/test/unit/util/login/token-refresh.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      payload: {},
+    } as unknown as Awaited<ReturnType<typeof jwtVerify>>);
+
+    // Refresh token payload
+    jwtVerifyMock.mockResolvedValueOnce({
+      payload: {},
+    } as unknown as Awaited<ReturnType<typeof jwtVerify>>);
+
+    // TODO: Execute a command that triggers an API call
+    // Eg.: cli('whoami');
```
markandrus: 👍 seems a reasonable approach

markandrus (commented):
Looks good, pending tests; however, what do you think about reducing the dependence on the access token and refresh token formats?

packages/cli/src/util/oauth.ts
```diff
@@ -380,22 +398,59 @@ function canParseURL(url: string) {
...
+>(token: T): Promise<[InspectionError] | [null, Payload]> {
   try {
+    // TODO: Use an [Introspection Endpoint](https://datatracker.ietf.org/doc/html/rfc7662)
+
     const JWKS = createRemoteJWKSet((await as()).jwks_uri);
-    const { payload } = await jwtVerify<VercelAccessToken>(token, JWKS, {
+    const { payload } = await jwtVerify<Payload>(token, JWKS, {
       issuer: 'https://vercel.com',
       audience: ['https://api.vercel.com', 'https://vercel.com/api'],
     });
```
markandrus: WDYT about just `decodeJwt` here and removing `createRemoteJWKSet`?

packages/cli/src/util/oauth.ts
```diff
@@ -380,22 +398,59 @@ function canParseURL(url: string) {
...
+
+    // TODO: Remove override when we have an introspection endpoint
+    payload.active ??= Boolean(
+      payload.exp && payload.exp < Math.floor(Date.now() / 1000)
+    );
+
+    if (!payload.active) {
+      throw new InspectionError('Token is not active.');
+    }
```
markandrus: Is this necessary, or could it be removed?

IIUC, the only hard dependency we currently have is `team_id`?

balazsorban44: This _is_ necessary, this is how we know to perform the token refresh, only if the token is not expired. It can save an extra request, and the format matches a future implementation of an [Introspection Endpoint](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2), where the `active` field is required.

See the PR description for an overview.

packages/cli/src/util/oauth.ts
balazsorban44: Yeah, it's fine, as we will verify the token signature later anyway. :+1:

trek (dismissed):

internals/types/index.d.ts
```diff
@@ -42,12 +67,6 @@ export interface GlobalConfig {
   telemetry?: {
     enabled?: boolean;
   };
-
-  // TODO: legacy - remove
-  updateChannel?: string;
-  desktop?: {
-    teamOrder: any;
-  };
```
balazsorban44, line 50: I found no reference to `updateChannel`, and `desktop` was only referenced in `logout`, where it just got deleted.

internals/types/index.d.ts
```diff
@@ -26,13 +26,38 @@ export interface JSONObject {
...
+   * The absolute time (seconds) when the {@link OAuthAuthConfig.token} expires.
+   * Used to optimistically check if the token is still valid.
+   */
+  expiresAt?: number;
+  refreshToken?: string;
+  /**
+   * The absolute time (seconds) when the {@link OAuthAuthConfig.refreshToken} expires.
+   * Used to optimistically check if the token is still valid.
+   */
+  refreshTokenExpiresAt?: number;
```
markandrus: IMO, we do not even need to track this. If we have a refresh token, we can try to use it, irrespective of its expiry.

packages/cli/src/commands/login/future.ts
```diff
@@ -120,46 +115,59 @@ export async function login(client: Client): Promise<number> {
...
 
-      if (accessTokenError) {
-        return accessTokenError;
-      }
+      o.debug('access_token inspected');
+
+      client.updateAuthConfig({
+        token: tokens.access_token,
+        type: 'oauth',
+        expiresAt: Math.floor(Date.now() / 1000) + tokens.expires_in,
```
markandrus, line 141: 👍

packages/cli/src/commands/login/future.ts
```diff
@@ -120,46 +115,59 @@ export async function login(client: Client): Promise<number> {
...
 
-      if (accessToken.team_id) {
-        o.debug('Current team updated');
-        client.config.currentTeam = accessToken.team_id;
-      } else {
-        o.debug('Current team deleted');
-        delete client.config.currentTeam;
+        client.updateAuthConfig({
+          refreshToken: tokens.refresh_token,
+          refreshTokenExpiresAt: payload.exp,
```
markandrus: IMO, we could ignore `payload.exp` and remove the refresh token inspection above:

```ts
        const [inspectError, payload] = await inspectToken(
          tokens.refresh_token
        );
```

packages/cli/src/util/client.ts
```diff
@@ -67,6 +73,25 @@ export const isJSONObject = (v: any): v is JSONObject => {
...
+  return 'token' in authConfig && (authConfig.expiresAt ?? 0) >= Date.now();
+}
+
+export function isValidRefreshToken(
+  authConfig: OAuthAuthConfig
+): authConfig is OAuthAuthConfig & { refreshToken: string } {
+  return (
+    'refreshToken' in authConfig &&
+    (authConfig.refreshTokenExpiresAt ?? 0) >= Date.now()
+  );
```
markandrus: IMO, we could simplify this to
```suggestion
  return 'refreshToken' in authConfig
```

packages/cli/src/util/client.ts
```diff
@@ -132,7 +157,98 @@ export default class Client extends EventEmitter implements Stdio {
...
+        if (output.isDebugEnabled()) output.prettyError(inspectError);
+        this.emptyAuthConfig();
+        this.writeToAuthConfigFile();
+        return;
+      }
+
+      this.updateAuthConfig({
+        refreshToken: tokens.refresh_token,
+        refreshTokenExpiresAt: payload.exp,
+      });
```
markandrus: IMO, we could simplify this, and skip inspecting the refresh token:
```suggestion
      this.updateAuthConfig({
        refreshToken: tokens.refresh_token,
      });
```

packages/cli/src/util/oauth.ts
```diff
@@ -382,22 +400,54 @@ function canParseURL(url: string) {
...
+/** @see https://datatracker.ietf.org/doc/html/rfc7662#section-2.2 */
+interface Token {
+  /**
+   * Integer timestamp, measured in the number of seconds
+   * since January 1 1970 UTC, indicating when this token will expire.
+   */
+  exp: number;
+  /** Whether or not the presented token is active. */
+  active: boolean;
+}
```
markandrus: If we stop checking the refresh token's `exp`, then I think we can remove this type, and simplify the return type of `inspectToken`. `inspectToken` could be as simple as calling `decodeJwt`. WDYT?

markandrus (commented):
Looks close. I think we could still simplify the `introspectToken` function and remove the `Token` interface with `active` and `exp` if we stop checking the refresh token's expiration (which I think is not necessary). WDYT?

packages/cli/src/util/oauth.ts
```diff
@@ -382,22 +400,44 @@ function canParseURL(url: string) {
...
+    const payload = await decodeJwt<AccessToken>(token);
+
+    // TODO: Remove override when we have an introspection endpoint
+    payload.active ??= Boolean(
+      payload.exp && payload.exp < Math.floor(Date.now() / 1000)
+    );
+
+    if (!payload.active) {
+      throw new InspectionError('Token is not active.');
+    }
```
markandrus: Still needed?
```suggestion
```

markandrus (dismissed):
Looks good! One final suggestion. Going to do some manual testing now.

markandrus (dismissed):

markandrus (approved):

trek (approved):

PR #13336, opened 2025-05-15
[mcp-adapter] Correctly set HTTP status code

SferaDev:
We call both `res.writeHead` and `res.statusCode` to set a custom HTTP response code, however the response wrapper was neglecting the HTTP status codes set by `res.statusCode`.


packages/mcp-adapter/src/next/server-response-adapter.ts
```diff
@@ -57,7 +58,7 @@ export function createServerResponseAdapter(
         throw new Error('Buffer not supported');
       }
       if (!wroteHead) {
-        writeHead(200);
+        writeHead(statusCode, headers);
```
SferaDev, line 61: This is the main bug in the code, if `writeHead` had not been called we returned 200 regardless of the `statusCode` value.

quuu:
hey @SferaDev , mind making this pr again against https://github.com/vercel/mcp-adapter :) 

appreciate you looking into this!

SferaDev:
https://github.com/vercel/mcp-adapter/pull/11

PR #13340, opened 2025-05-15
fix(auth): tweak `--future` telemetry

balazsorban44:
Continues on https://github.com/vercel/vercel/pull/13318 by also recording the command in `value`

### DIFF
--- a/packages/cli/src/util/telemetry/index.ts
+++ b/packages/cli/src/util/telemetry/index.ts
@@ -182,16 +182,8 @@ export class TelemetryClient {
     });
   }
 
-  trackCliFlagFuture(command: string, subcommands?: string | string[]) {
-    let subcommand: string | undefined;
-    if (subcommands) {
-      subcommand = Array.isArray(subcommands) ? subcommands[0] : subcommands;
-    }
-
-    this.track({
-      key: 'flag:future',
-      value: subcommand ? `${command}:${subcommand}` : command,
-    });
+  trackCliFlagFuture(command: string) {
+    this.track({ key: 'flag:future', value: command });
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/util/telemetry/index.ts
```diff
@@ -182,16 +182,8 @@ export class TelemetryClient {
...
-    if (subcommands) {
-      subcommand = Array.isArray(subcommands) ? subcommands[0] : subcommands;
-    }
-
-    this.track({
-      key: 'flag:future',
-      value: subcommand ? `${command}:${subcommand}` : command,
-    });
+  trackCliFlagFuture(command: string) {
+    this.track({ key: 'flag:future', value: command });
```
trek, line 186: What are the values of `command` here? Just making sure it's nothing PII-ish.

trek (changes requested):

packages/cli/src/util/telemetry/index.ts
balazsorban44: It's either `login` or `logout`:

https://github.com/vercel/vercel/pull/13318/files#diff-9a4160d988b870b4c1d02af440e591e01e61ac34cc7cc32902cf4d5c4871b61bR52

https://github.com/vercel/vercel/pull/13318/files#diff-a5e38875e5ed3b63ba19a6a5384fc5b9face5fa5349cbb5343d54b25b85ce6fbR40

trek: Sorry, misread the diff!

trek (approved):

packages/cli/src/util/telemetry/index.ts
balazsorban44: I enforced it with a type to avoid accidentally passing something else :+1:

PR #13346, opened 2025-05-19
[run-user-scripts] return a more verbose error

marc-vercel:
Return an error that is more verbose and clearly states we errored when parsing the `package.json`.

As can be seen in this image, it is hard to know in which JSON file we have the error:
image

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -357,7 +357,7 @@ export async function scanParentDirs(
       packageJson = JSON.parse(await fs.readFile(pkgJsonPath, 'utf8'));
     } catch (err) {
       throw new Error(
-        `Could not read package.json: ${(err as Error).message}.`
+        `Could not read ${pkgJsonPath}: ${(err as Error).message}.`
       );
     }
   }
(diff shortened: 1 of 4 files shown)


packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -350,10 +350,17 @@ export async function scanParentDirs(
...
-      ? JSON.parse(await fs.readFile(pkgJsonPath, 'utf8'))
-      : undefined;
+
+  let packageJson: PackageJson | undefined;
+  if (readPackageJson && pkgJsonPath) {
+    try {
+      packageJson = JSON.parse(await fs.readFile(pkgJsonPath, 'utf8'));
+    } catch (err) {
+      throw new Error(
+        `Could not read package.json: ${(err as Error).message}.`
```
healeycodes: let's also put the path in the error message?

there may be multiple `package.json` files in a project (which may be read at different times during a build)

marc-vercel: SGTM. I've added the path (base path is `/`; I could not make it more relative) in https://github.com/vercel/vercel/pull/13346/commits/c445bc0aad8106e838b4ac8b181a712d6b4d7743

healeycodes (dismissed):

LukeSheard (approved):

PR #13357, opened 2025-05-20
[cli] Warn if system env vars are not available

onsclom:
Features like skew protection are not supported without system environment variables, so `vc build` should warn when running outside of Vercel.  [Linear issue](https://linear.app/vercel/issue/FLOW-3548/could-we-add-a-warning-when-vc-build-is-run-outside-of-vercel-to).

The `VERCEL=1` env var gets set by the cli in all contexts, so we can't use that. I used `VERCEL_BUILD_IMAGE` because it is defined in the build container.

### DIFF
--- a/.changeset/gold-rice-cry.md
+++ b/.changeset/gold-rice-cry.md
@@ -1,0 +1,5 @@
+---
+'vercel': minor
+---
+
+Warn when system environment variables are not available
--- a/packages/cli/src/commands/build/index.ts
+++ b/packages/cli/src/commands/build/index.ts
@@ -275,6 +275,12 @@ export default async function main(client: Client): Promise<number> {
     argv: scrubArgv(process.argv),
   };
 
+  if (!process.env.VERCEL_BUILD_IMAGE) {
+    output.warn(
+      'Build not running on Vercel. System environment variables will not be available.'
+    );
+  }
+
   const envToUnset = new Set<string>(['VERCEL', 'NOW_BUILDER']);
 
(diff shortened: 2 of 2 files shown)


trek (approved):
Really a bummer we can't use `VERCEL` for this

My only concern with `VERCEL_BUILD_IMAGE` is that it would have two meanings for us system wide:

* which build image you're using
* a semaphore for "I am building on Vercel"

If we ever moved away from the build image concept, we'd either have to keep that variable around – and _know_ to do that – or accidentally start warning on the platform.

Ideally, we'd pick a new one specifically for this purpose, but the risk and effect of some strays `warns` showing up in logs seems minimal so I wouldn't block on it, especially if this improves customer experience today.

onsclom:
> Really a bummer we can't use `VERCEL` for this
> 
> My only concern with `VERCEL_BUILD_IMAGE` is that it would have two meanings for us system wide:
> 
> * which build image you're using
> * a semaphore for "I am building on Vercel"
> 
> If we ever moved away from the build image concept, we'd either have to keep that variable around – and _know_ to do that – or accidentally start warning on the platform.
> 
> Ideally, we'd pick a new one specifically for this purpose, but the risk and effect of some strays `warns` showing up in logs seems minimal so I wouldn't block on it, especially if this improves customer experience today.

Good points, made a todo ticket for this: https://linear.app/vercel/issue/FLOW-3571/create-new-env-var-for-determining-whether-build-is-happening-on

PR #13361, opened 2025-05-21
fix(cli): clarify logs when logging in and potentially have no account

balazsorban44:
If a user starts the login flow and enters a non-existent email, we should let the user know that they might need to sign up potentially.


balazsorban44:
Will fix in the API side.

PR #13364, opened 2025-05-21
[cli] add blob command

luismeyer:
This PR adds cli commands to interact with Vercel Blob. 

These are the new API's:
`vc blob store add mystore` - creates a new blob store, asks to connect store to project and suggests pulling the new env var
`vc blob store rm store_mystoreid` - delete a blob store

All of the following commands will try to read the `BLOB_READ_WRITE_TOKEN` from the next `.env.local` file:
`vc blob put file.txt` - uploads the file from the path to the blob store
`vc blob ls` - list blobs in the store
`vc blob del path/ path2/` - delete blobs from the store
`vc blob copy fromUrl toPathname` - copies a file within the store


packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+  description: 'Upload a file to the Blob store',
+  arguments: [
+    {
+      name: 'pathToFile',
+      required: true,
+    },
+  ],
+  options: [
+    {
+      name: 'add-random-suffix',
```
vvo, line 58: Do we want/can document the defaults? For all options I mean

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      description: 'Pathname to upload the file to',
+      argument: 'String',
+    },
+    {
+      name: 'multipart',
+      shorthand: 'u',
+      type: Boolean,
+      deprecated: false,
+      description:
+        'If true upload the file in multiple smaller chunks. Default: true',
```
vvo:
```suggestion
        'If true upload the file in multiple small chunks for performance and reliability. Default: true',
```

vvo: BTW it's not true by default in the SDK (it should be I think for files that are big enough or for streams)

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      argument: 'Boolean',
+    },
+    {
+      name: 'content-type',
+      shorthand: 't',
+      type: String,
+      deprecated: false,
+      description:
+        'Overwrite the content-type. Will be infered from the file extension if not provided',
```
vvo:
```suggestion
        'Overwrite the content-type. Will be inferred from the file extension if not provided',
```

Also you could perhaps just reuse the docs we have in @vercel/blob

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      argument: 'String',
+    },
+    {
+      name: 'cache-control-max-age',
+      shorthand: 'c',
+      type: Number,
+      deprecated: false,
+      description: 'Max-age of the cache-control header directive',
```
vvo: Default to one month?

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      argument: 'Number',
+    },
+    {
+      name: 'force',
+      shorthand: 'f',
+      type: Boolean,
+      deprecated: false,
+      description: 'Overwrite the file if it already exists',
```
vvo: Defaults to false

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+  examples: [],
+} as const;
+
+export const delSubcommand = {
+  name: 'del',
+  aliases: [],
+  description: 'Delete a file from the Blob store',
+  arguments: [
+    {
+      name: '<URLS>',
```
vvo: Do we want to say you can also just pass pathnames?

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      required: true,
+    },
+  ],
+  options: [
+    {
+      name: 'add-random-suffix',
+      shorthand: 'r',
+      type: Boolean,
+      deprecated: false,
+      description: 'Add a random suffix to the file name',
```
vvo, line 146: Default to false

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      deprecated: false,
+      description: 'The max age of the cache control',
+      argument: 'Number',
+    },
+  ],
+  examples: [],
+} as const;
+
+export const newStoreSubcommand = {
+  name: 'new',
```
vvo: Perhaps name it create

packages/cli/src/commands/blob/command.ts
```diff
@@ -0,0 +1,209 @@
...
+      name: 'name',
+      required: false,
+    },
+  ],
+  options: [],
+  examples: [],
+} as const;
+
+export const removeStoreSubcommand = {
+  name: 'remove',
```
vvo, line 186: Let's name it delete?

vvo:
@luismeyer Can you provide in the PR description the API (= the commands a customer would do)?
So we start from the DX and then the implementation.

packages/cli/src/commands/blob/command.ts
luismeyer: I think thats too close to the del operation

packages/cli/src/commands/blob/command.ts
luismeyer: I changed it to `add` to match other subcommands

vvo:
> These are the new API's: `vc blob add mystore` - creates a new blob store, asks to connect store to project and suggests pulling the new env var `vc blob rm store_mystoreid` - delete a blob store
> 
> All of the following commands will try to read the `BLOB_READ_WRITE_TOKEN` from the next `.env.local` file:
> 
> `vc blob put file.txt` - uploads the file from the path to the blob store `vc blob ls` - list blobs in the store `vc blob del path/ path2/` - delete blobs from the store `vc blob copy fromUrl toPathname` - copies a file within the store

We should probably do:

```
vc blob create-store
vc blob delete-store
vc blob put
vc blob head
vc blob list
vc blob copy
vc blob del
```

This way there cannot be confusion about "I though I would delete a blob and it deleted a store" and we don't have to distinguish rm/del.

luismeyer:
@vvo I added a subsubcommand:
`vc blob store add mystore`
`vc blob store rm store_mystoreid`

vvo:
@luismeyer Does "add" means create or link to project, or does it means create. If it means create then really we should just map what the UI call it: create and delete. Not add and rm.

vvo:
You don't do these actions often, it's fine I think if they are longer to type

vvo:
Some people may think "add" if the store already exists it's just gonna add it. I guess we should have that but then it should be:

```
vercel blob store create (same word as UI, create and propose to connect, error if already exists)
vercel blob store connect (same word as UI, no op if already connected to project)
vercel blob store disconnect (remove connection)
vercel blbo store empty (same word as UI) -- not required, just for idea
vercel blob store delete (same word as UI)
```

luismeyer:
> Some people may think "add" if the store already exists it's just gonna add it. I guess we should have that but then it should be:
> 
> ```
> vercel blob store create (same word as UI, create and propose to connect, error if already exists)
> vercel blob store connect (same word as UI, no op if already connected to project)
> vercel blob store disconnect (remove connection)
> vercel blbo store empty (same word as UI) -- not required, just for idea
> vercel blob store delete (same word as UI)
> ```

we can separate these, but my first goal was to have 1 command that gets you started. So `add` currently creates the store and connects it to the project if the command is executed in a linked repo

vvo:
Do we have other `vc` commands using `add` and `rm`?

luismeyer:
> Do we have other `vc` commands using `add` and `rm`?

yes: alias, domains, dns, env, integration, project, teams

vvo (dismissed):

packages/cli/src/commands/env/pull.ts
```diff
@@ -229,7 +229,7 @@ export async function envPullCommandLogic(
 function escapeValue(value: string | undefined) {
   return value
     ? value
-        .replace(new RegExp('\n', 'g'), '\\n') // combine newlines (unix) into one line
-        .replace(new RegExp('\r', 'g'), '\\r') // combine newlines (windows) into one line
+        .replace(/\n/g, '\\n') // combine newlines (unix) into one line
+        .replace(/\r/g, '\\r') // combine newlines (windows) into one line
```
chris-olszewski: Is there a reason for this change?

packages/cli/src/index.ts
```diff
@@ -325,7 +325,7 @@ const main = async () => {
   // Gets populated to the subcommand name when a built-in is
   // provided, otherwise it remains undefined for an extension
   let subcommand: string | undefined = undefined;
-  let userSuppliedSubCommand: string = '';
+  let userSuppliedSubCommand = '';
```
chris-olszewski:
```suggestion
  let userSuppliedSubCommand: string = '';
```

chris-olszewski (changes requested):
I think the changes make sense to me.

Could you add some tests for the new commands?

packages/cli/src/commands/env/pull.ts
luismeyer: was autofixed by biome. Will revert 👍

packages/cli/src/index.ts
luismeyer: was autofixed by biome. Will revert 👍

luismeyer:
> I think the changes make sense to me.
> 
> Could you add some tests for the new commands?

thank you @chris-olszewski  🙏 
I added tests for all the subcommands

chris-olszewski (dismissed):

chris-olszewski (approved):

PR #13367, opened 2025-05-22
fix(cli): drop `--future` flag from `logout`

balazsorban44:
Since this is an officially unreleased feature so far, it should be fine dropping this flag, but we rely on it in some internal tests, so https://github.com/vercel/vercel/pull/13366 should be merged and released first.

If it's not OK to merge now, we can also wait to delete until the next **major**.

### DIFF
--- a/.changeset/small-owls-warn.md
+++ b/.changeset/small-owls-warn.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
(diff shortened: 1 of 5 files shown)


markandrus (approved):

balazsorban44:
Will drop in a *major* instead.

PR #13368, opened 2025-05-22
Add condition operations to routing has/missing fields

charliemeyer2000:
### TL;DR

Added support for condition operations in routing rules, allowing for more advanced matching beyond regex patterns.

### What changed?

- Introduced a new `ConditionValue` type that supports various comparison operations:
  - `eq` / `neq`: Equal / not equal
  - `inc` / `ninc`: In array / not in array
  - `pre` / `suf`: Has prefix / has suffix
  - `re`: Match regex pattern
  - `gt` / `gte` / `lt` / `lte`: Numeric comparisons
- Updated the JSON schema to validate these new condition operations
- Modified the `collectHasSegments` function to extract named groups from regex operations
- Maintained backward compatibility with string values (treated as regex)

### How to test?

- Create routing rules using the new condition operations:
```json
{
  "rewrites": [
    {
      "source": "/api/:path*",
      "destination": "/backend/:path*",
      "has": [
        { "type": "header", "key": "authorization", "value": { "pre": "Bearer " } },
        { "type": "query", "key": "version", "value": { "neq": "v1" } },
        { "type": "header", "key": "role", "value": { "inc": ["admin", "moderator"] } }
      ]
    }
  ]
}
```
- Verify that existing string-based conditions continue to work
- Test that named groups in regex patterns are properly extracted

### Why make this change?

The existing regex-only approach for condition matching was limited and could be difficult to use for common comparison operations. This change provides a more intuitive and powerful way to define routing conditions with specific operations like equality, array inclusion, and numeric comparisons, while maintaining backward compatibility with existing regex patterns.

### DIFF
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -1,3 +1,81 @@
+const conditionValueSchema = {
+  anyOf: [
+    {
+      description:
+        'A string value for backward compatibility (treated as regex)',
+      type: 'string',
+      maxLength: 4096,
+    },
+    {
+      description: 'A condition operation object',
+      type: 'object',
+      additionalProperties: false,
+      minProperties: 1,
+      maxProperties: 1,
+      properties: {
+        eq: {
+          description: 'Equal',
+          anyOf: [{ type: 'string', maxLength: 4096 }, { type: 'number' }],
+        },
+        neq: {
+          description: 'Not equal',
+          type: 'string',
+          maxLength: 4096,
+        },
+        inc: {
+          description: 'In array',
+          type: 'array',
+          items: {
+            type: 'string',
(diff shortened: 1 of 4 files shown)


charliemeyer2000:
* **#13369** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13368** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #13369, opened 2025-05-22
Add mitigation action schema to routing-utils

charliemeyer2000:
### TL;DR

Added support for mitigation actions in routing configurations.

### What changed?

- Added a new `mitigateSchema` to define the structure for mitigation actions in routes
- Created a new route type that requires `src` and `mitigate` properties
- Refactored the routes schema to use a common `routeBaseProperties` object
- Added TypeScript interfaces for mitigation actions including:
  - Rate limiting with fixed window or token bucket algorithms
  - Redirect configurations
  - Various action types (log, challenge, deny, bypass, rate_limit, redirect)
- Added comprehensive tests for the new mitigation features

### How to test?

Test with various route configurations that use the new mitigation features:

```js
// Rate limiting example
{
  src: '^/api/protected/(.*)$',
  mitigate: {
    action: 'rate_limit',
    rateLimit: {
      algo: 'fixed_window',
      window: 60,
      limit: 10,
      keys: ['ip'],
      action: 'deny'
    }
  }
}

// Challenge example
{
  src: '^/admin/(.*)$',
  mitigate: {
    action: 'challenge',
    actionDuration: '1h',
    bypassSystem: false
  }
}

// Redirect example
{
  src: '^/blocked/(.*)$',
  mitigate: {
    action: 'redirect',
    redirect: {
      location: 'https://example.com/blocked',
      permanent: false
    }
  }
}
```

### Why make this change?

This change enhances the routing system with security features that allow for more granular control over request handling. This is already supported in the proxy, but is not exposed to `vercel.json` or `next.config.ts`.

### DIFF
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -76,6 +76,104 @@ const conditionValueSchema = {
   ],
 } as const;
 
+const mitigateSchema = {
+  description: 'Mitigation action configuration',
+  type: 'object',
+  additionalProperties: false,
+  required: ['action'],
+  properties: {
+    action: {
+      description: 'The mitigation action to take',
+      type: 'string',
+      enum: ['log', 'challenge', 'deny', 'bypass', 'rate_limit', 'redirect'],
+    },
+    rateLimit: {
+      description: 'Rate limiting configuration',
+      anyOf: [
+        { type: 'null' },
+        {
+          type: 'object',
+          additionalProperties: false,
+          required: ['algo', 'window', 'limit', 'keys'],
+          properties: {
+            algo: {
+              description: 'Rate limiting algorithm',
+              type: 'string',
+              enum: ['fixed_window', 'token_bucket'],
+            },
+            window: {
+              description: 'Time window in seconds',
+              type: 'number',
+              minimum: 1,
+            },
+            limit: {
+              description: 'Request limit within the window',
+              type: 'number',
+              minimum: 1,
+            },
+            keys: {
+              description: 'Keys to group rate limiting by',
+              type: 'array',
+              items: {
+                type: 'string',
+                maxLength: 256,
+              },
+              maxItems: 10,
+              minItems: 1,
+            },
+            action: {
+              description: 'Action to take when rate limit is exceeded',
+              anyOf: [
+                { type: 'null' },
+                {
+                  type: 'string',
+                  enum: ['log', 'challenge', 'deny', 'rate_limit'],
+                },
+              ],
+            },
+          },
+        },
+      ],
+    },
+    redirect: {
+      description: 'Redirect configuration',
+      anyOf: [
+        { type: 'null' },
+        {
+          type: 'object',
+          additionalProperties: false,
+          required: ['location', 'permanent'],
+          properties: {
+            location: {
+              description: 'Redirect destination URL',
+              type: 'string',
+              maxLength: 4096,
+            },
+            permanent: {
+              description: 'Whether the redirect is permanent',
+              type: 'boolean',
+            },
+          },
+        },
+      ],
+    },
+    actionDuration: {
+      description: 'Duration for the mitigation action',
+      anyOf: [
+        { type: 'null' },
+        {
+          type: 'string',
+          maxLength: 256,
+        },
+      ],
+    },
+    bypassSystem: {
+      description: 'Whether to bypass the system mitigation',
+      anyOf: [{ type: 'null' }, { type: 'boolean' }],
+    },
+  },
+} as const;
+
 export const hasSchema = {
   description: 'An array of requirements that are needed to match',
   type: 'array',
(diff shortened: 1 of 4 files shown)


charliemeyer2000:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/vercel/13369?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#13369** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13368** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13368?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/routing-utils/test/index.spec.ts
```diff
@@ -256,377 +256,81 @@ describe('normalizeRoutes', () => {
...
-          schemaPath: '#/items/anyOf',
-          params: {},
-          message: 'should match some schema in anyOf',
-        },
-      ]
-    );
+    const validate = ajv.compile(routesSchema);
+    const valid = validate([{ src: false }]);
+    assert.equal(valid, false);
+    assert.ok(validate.errors?.some(err => err.keyword === 'type'));
```
styfle, line 262: Why is this test changing?

I wouldn't expect any existing tests to change. Only adding new ones.

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+export type MitigateActionType =
+  | 'log'
+  | 'challenge'
+  | 'deny'
+  | 'bypass'
+  | 'rate_limit'
+  | 'redirect';
+
+export interface MitigateAction {
+  mitigate?: {
```
styfle, line 50: Are we expecting other props on `MitigateAction` besides `mitigate`?

I'm curious why this isn't the top level.

packages/routing-utils/src/schemas.ts
```diff
@@ -129,6 +227,104 @@ export const hasSchema = {
 /**
  * An ajv schema for the routes array
  */
```
styfle, line 229: This JSDoc is in the wrong place now.

All exports must have JSDoc.

packages/routing-utils/src/schemas.ts
```diff
@@ -143,105 +339,20 @@ export const routesSchema = {
...
-          has: hasSchema,
-          missing: hasSchema,
+        },
+      },
+      {
+        type: 'object',
+        required: ['src', 'mitigate'],
+        additionalProperties: false,
+        properties: {
+          ...routeBaseProperties,
```
styfle, line 354: All these additional properties also need tests (looks like the only tests added are using `src` only)

packages/routing-utils/src/schemas.ts
```diff
@@ -129,6 +227,104 @@ export const hasSchema = {
...
+      type: 'string',
+      maxLength: 32,
+    },
+  },
+  caseSensitive: {
+    type: 'boolean',
+  },
+  important: {
+    type: 'boolean',
+  },
```
styfle, line 260: Do we need all these additional properties?

I would expect maybe `src`, `headers`, `methods`, `has`.

But I wouldn't expect `continue`, `override`, `check`, `status`, etc

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+  permanent: boolean;
+}
+
+export type MitigateActionType =
+  | 'log'
+  | 'challenge'
+  | 'deny'
+  | 'bypass'
+  | 'rate_limit'
+  | 'redirect';
```
styfle, line 47: I think we can remove `redirect` here since we already have `redirects` top level in vercel.json

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+
+export interface MitigateAction {
```
styfle, line 49:
```suggestion
export interface MitigationAction {
```

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+  mitigate?: {
+    action: MitigateActionType;
+    rateLimit?: MitigateRateLimit | null;
```
styfle, line 52: Do we need to call this `erl` to match?

Also lets drop `null`

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+    redirect?: MitigateRedirect | null;
+    actionDuration?: string | null;
```
styfle, line 54:
```suggestion
    actionDuration?: number;
```
Should this be a number? And should we call it `ttl`? Also drop null.

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+  | 'deny'
+  | 'bypass'
+  | 'rate_limit'
+  | 'redirect';
+
+export interface MitigateAction {
+  mitigate?: {
+    action: MitigateActionType;
+    rateLimit?: MitigateRateLimit | null;
+    redirect?: MitigateRedirect | null;
```
styfle, line 53: I think we can drop redirect, right?

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+    actionDuration?: string | null;
+    bypassSystem?: boolean | null;
```
styfle, line 55: I don't think we need to bypass system. Wouldn't every rule bypass the system?

packages/routing-utils/src/types.ts
```diff
@@ -25,6 +25,37 @@ export type ConditionValue =
...
+  keys: string[];
+  action?: 'log' | 'challenge' | 'deny' | 'rate_limit' | null;
+}
+
+export interface MitigateRedirect {
+  location: string;
+  permanent: boolean;
+}
+
+export type MitigateActionType =
```
styfle, line 41:
```suggestion
export type MitigateAction =
```

packages/routing-utils/src/schemas.ts
styfle: Lets revert code back how it was.

All we need is 

Route = {
  ...
  mitigate: mitigateSchema
  }

packages/routing-utils/src/types.ts
styfle: Lets get feedback from Product team as well as Edge Ingress if we want to rename this from `erl` to `rateLimit`

charliemeyer2000:
PR closed to create [new PR](https://github.com/vercel/vercel/pull/13375), a refined and better-scoped change to address this problem.

PR #13375, opened 2025-05-23
[routing-utils] Support Conditions and Mitigate in vercel.json

charliemeyer2000:
This PR adds support for conditions in `has` and `missing` and  `mitigate` in your `vercel.json` file, previously only manageable through the Vercel Firewall dashboard page.

### DIFF
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -1,3 +1,63 @@
+export const mitigateSchema = {
+  description: 'Mitigation action to take on a route',
+  type: 'object',
+  additionalProperties: false,
+  required: ['action'],
+  properties: {
+    action: {
+      description: 'The mitigation action to take',
+      type: 'string',
+      enum: ['log', 'challenge', 'deny', 'bypass', 'rate_limit', 'redirect'],
+    },
+    erl: {
+      description: 'Edge rate limit configuration',
+      type: 'object',
+      additionalProperties: false,
+      required: ['algo', 'window', 'limit', 'keys'],
+      properties: {
+        algo: {
+          description: 'The rate limiting algorithm to use',
+          type: 'string',
+          enum: ['fixed_window', 'token_bucket'],
+        },
+        window: {
+          description: 'Time window for rate limit in seconds',
+          type: 'number',
+          minimum: 10,
+          maximum: 600,
+          default: 60,
+        },
+        limit: {
+          description: 'Request limit.',
+          type: 'number',
+          minimum: 1,
+          maximum: 10_000_000,
+          default: 100,
+        },
+        keys: {
+          description: 'Keys to rate limit by',
+          type: 'array',
+          items: {
+            type: 'string',
+          },
+          minItems: 1,
+          maxItems: 3,
+        },
+      },
+    },
+  },
+  if: {
+    properties: {
+      action: {
+        const: 'rate_limit',
+      },
+    },
+  },
+  then: {
+    required: ['action', 'erl'],
+  },
+} as const;
+
 export const conditionValueSchema = {
   anyOf: [
     {
@@ -246,6 +306,7 @@ export const routesSchema = {
           },
           has: hasSchema,
           missing: hasSchema,
+          mitigate: mitigateSchema,
         },
       },
       {
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -7,54 +7,8 @@ export const mitigateSchema = {
     action: {
       description: 'The mitigation action to take',
       type: 'string',
-      enum: ['log', 'challenge', 'deny', 'bypass', 'rate_limit', 'redirect'],
+      enum: ['log', 'challenge', 'deny'],
     },
-    erl: {
-      description: 'Edge rate limit configuration',
-      type: 'object',
-      additionalProperties: false,
-      required: ['algo', 'window', 'limit', 'keys'],
-      properties: {
-        algo: {
-          description: 'The rate limiting algorithm to use',
-          type: 'string',
-          enum: ['fixed_window', 'token_bucket'],
-        },
-        window: {
-          description: 'Time window for rate limit in seconds',
-          type: 'number',
-          minimum: 10,
-          maximum: 600,
-          default: 60,
-        },
-        limit: {
-          description: 'Request limit.',
-          type: 'number',
-          minimum: 1,
-          maximum: 10_000_000,
-          default: 100,
-        },
-        keys: {
-          description: 'Keys to rate limit by',
-          type: 'array',
-          items: {
-            type: 'string',
-          },
-          minItems: 1,
-          maxItems: 3,
-        },
-      },
-    },
-  },
-  if: {
-    properties: {
-      action: {
-        const: 'rate_limit',
-      },
-    },
-  },
-  then: {
-    required: ['action', 'erl'],
   },
 } as const;
 
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -12,7 +12,9 @@ export const mitigateSchema = {
   },
 } as const;
 
-export const conditionValueSchema = {
+export const hasConditionValueSchema = {
+  description:
+    'A value to match against. Can be a string (regex) or a condition operation object',
   anyOf: [
     {
       description: 'A string value (treated as regex)',
@@ -111,9 +113,7 @@ export const hasSchema = {
             enum: ['host'],
           },
           value: {
-            description:
-              'A value to match against. Can be a string (regex) or a condition operation object',
-            ...conditionValueSchema,
+            ...hasConditionValueSchema,
           },
         },
       },
@@ -134,9 +134,7 @@ export const hasSchema = {
             maxLength: 4096,
           },
           value: {
-            description:
-              'A value to match against. Can be a string (regex) or a condition operation object',
-            ...conditionValueSchema,
+            ...hasConditionValueSchema,
           },
         },
       },
--- a/packages/routing-utils/src/schemas.ts
+++ b/packages/routing-utils/src/schemas.ts
@@ -1,3 +1,85 @@
+export const conditionValueSchema = {
+  anyOf: [
+    {
+      description: 'A string value (treated as regex)',
+      type: 'string',
+      maxLength: 4096,
+    },
+    {
+      description: 'A condition operation object',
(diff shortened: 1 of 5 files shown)


packages/routing-utils/test/index.spec.ts
```diff
@@ -1076,6 +1077,15 @@ describe('getTransformedRoutes', () => {
...
     assertValid(vercelConfig.trailingSlashSchema, trailingSlashSchema);
+    for (const item of [
+      ...vercelConfig.rewrites,
+      ...vercelConfig.redirects,
+      ...vercelConfig.headers,
+    ]) {
+      if (item.has) {
+        assertValid(item.has, hasSchema);
+      }
+    }
```
styfle: Why is this needed? It should be handled on line 1076-1078.

packages/routing-utils/src/superstatic.ts
```diff
@@ -302,6 +309,21 @@ const normalizeHasKeys = (hasItems: HasField = []) => {
...
+function getStringValueForRegex(value: ConditionValue): string | null {
+  if (typeof value === 'string') {
+    return value;
+  }
+
+  if (typeof value === 'object' && value !== null) {
+    if ('re' in value && typeof value.re === 'string') {
+      return value.re;
+    }
+    return null;
```
styfle: This return is not needed since the final return will cover that case

```suggestion
```

packages/routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,99 @@
...
+  properties: {
+    action: {
+      description: 'The mitigation action to take',
+      type: 'string',
+      enum: ['log', 'challenge', 'deny'],
+    },
+  },
+} as const;
+
+export const conditionValueSchema = {
```
styfle: Let's be more specific here

```suggestion
export const hasValueSchema = {
```

packages/routing-utils/src/types.ts
```diff
@@ -9,15 +9,33 @@ export type RouteApiError = {
   errors?: string[]; // array of all error messages
 };
 
+export type ConditionValue =
```
styfle: Also here

```suggestion
export type HasFieldValue =
```

packages/routing-utils/src/schemas.ts
```diff
@@ -16,9 +112,8 @@ export const hasSchema = {
           },
           value: {
             description:
-              'A regular expression used to match the value. Named groups can be used in the destination',
-            type: 'string',
-            maxLength: 4096,
+              'A value to match against. Can be a string (regex) or a condition operation object',
```
styfle: Let's also move the `description` into the schema since its the same for both

packages/routing-utils/src/superstatic.ts
charliemeyer2000: agree - removed.

packages/routing-utils/src/schemas.ts
charliemeyer2000: fixed.

packages/routing-utils/src/schemas.ts
charliemeyer2000: fixed.

packages/routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,101 @@
...
+    },
+  },
+} as const;
+
+export const hasConditionValueSchema = {
+  description:
+    'A value to match against. Can be a string (regex) or a condition operation object',
+  anyOf: [
+    {
+      description: 'A string value (treated as regex)',
```
styfle:
```suggestion
      description: 'A regular expression used to match the value. Named groups can be used in the destination.,
```

I think the previous description was better. Lets revert this

packages/routing-utils/src/schemas.ts
```diff
@@ -15,10 +113,7 @@ export const hasSchema = {
             enum: ['host'],
           },
           value: {
```
styfle: You don't need spread anymore

packages/routing-utils/src/schemas.ts
```diff
@@ -39,10 +134,7 @@ export const hasSchema = {
             maxLength: 4096,
           },
           value: {
```
styfle: Same here - You don't need spread anymore. Just assign the value.

packages/routing-utils/src/superstatic.ts
```diff
@@ -311,9 +332,12 @@ export function collectHasSegments(has?: HasField) {
...
-        if (match[1]) {
-          hasSegments.add(match[1]);
+      const stringValue = getStringValueForRegex(hasItem.value);
+      if (stringValue) {
+        for (const match of stringValue.matchAll(namedGroupsRegex)) {
+          if (match[1]) {
+            hasSegments.add(match[1]);
+          }
         }
       }
```
styfle: Nit pick: you can remove an `if` statement since getStringValueForRegex works on null inputs. No need for `if (hasItem.value)`

packages/routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,102 @@
+export const mitigateSchema = {
+  description: 'Mitigation action to take on a route',
+  type: 'object',
+  additionalProperties: false,
+  required: ['action'],
+  properties: {
+    action: {
+      description: 'The mitigation action to take',
+      type: 'string',
+      enum: ['log', 'challenge', 'deny'],
```
styfle: Lets talk with the team to discuss what happens when you have a mitigate action defined in the dashboard and in vercel.json - does one take precedence or do both apply? Imagine one is challenge and the other is deny 🤔

packages/routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,102 @@
+export const mitigateSchema = {
```
styfle: Lets avoid exporting

packages/routing-utils/src/schemas.ts
```diff
@@ -1,3 +1,102 @@
...
+  properties: {
+    action: {
+      description: 'The mitigation action to take',
+      type: 'string',
+      enum: ['log', 'challenge', 'deny'],
+    },
+  },
+} as const;
+
+export const hasConditionValueSchema = {
```
styfle: Lets avoid exporting

packages/routing-utils/src/types.ts
```diff
@@ -9,15 +9,33 @@ export type RouteApiError = {
...
+      pre?: string;
+      suf?: string;
+      re?: string;
+      gt?: number;
+      gte?: number;
+      lt?: number;
+      lte?: number;
+    };
+
+export type MitigateAction = 'log' | 'challenge' | 'deny';
```
styfle: Lets avoid exporting

packages/routing-utils/src/superstatic.ts
```diff
@@ -3,7 +3,14 @@
...
  */
 import { parse as parseUrl, format as formatUrl } from 'url';
-import { Route, Redirect, Rewrite, HasField, Header } from './types';
+import {
+  Route,
+  Redirect,
+  Rewrite,
+  HasField,
+  Header,
+  HasConditionValue,
```
styfle: Lets avoid export `HasConditionValue` and you can reference with `HasField['value']` I believe

packages/routing-utils/src/types.ts
```diff
@@ -9,15 +9,33 @@ export type RouteApiError = {
   errors?: string[]; // array of all error messages
 };
 
+export type HasConditionValue =
```
styfle: Also avoid exporting this one too

packages/routing-utils/src/superstatic.ts
```diff
@@ -302,6 +309,20 @@ const normalizeHasKeys = (hasItems: HasField = []) => {
   return hasItems;
 };
 
+function getStringValueForRegex(value: HasConditionValue): string | null {
+  if (typeof value === 'string') {
+    return value;
+  }
+
+  if (typeof value === 'object' && value !== null) {
```
styfle:
```suggestion
  if (value && typeof value === 'object' && value !== null) {
```

packages/routing-utils/src/superstatic.ts
```diff
@@ -302,6 +309,20 @@ const normalizeHasKeys = (hasItems: HasField = []) => {
   return hasItems;
 };
 
+function getStringValueForRegex(value: HasConditionValue): string | null {
```
styfle:
```suggestion
function getStringValueForRegex(value: HasConditionValue | undefined): string | null {
```

packages/routing-utils/test/index.spec.ts
```diff
@@ -1221,4 +1222,192 @@ describe('getTransformedRoutes', () => {
...
+        mitigate: {
+          action: 'challenge',
+        },
+      },
+      {
+        src: '/api/testy/(.*)',
+        mitigate: {
+          action: 'log',
+        },
+      },
```
styfle, line 1408: Lets make this exhaustive and test all types

packages/routing-utils/test/index.spec.ts
```diff
@@ -1221,4 +1222,192 @@ describe('getTransformedRoutes', () => {
...
+            { type: 'cookie', key: 'theme', value: { inc: ['light', 'dark'] } },
+            { type: 'query', key: 'page', value: { neq: '1' } },
+            {
+              type: 'header',
+              key: 'x-role',
+              value: { ninc: ['admin', 'superadmin'], pre: 'user-' },
+            },
+          ],
+          missing: [{ type: 'cookie', key: 'disabled', value: { eq: 'true' } }],
+        },
```
styfle, line 1286: Lets make this exhaustive too

packages/routing-utils/src/schemas.ts
charliemeyer2000: fixed.

packages/routing-utils/src/schemas.ts
charliemeyer2000: fixed.

packages/routing-utils/src/schemas.ts
charliemeyer2000: fixed.

packages/routing-utils/test/index.spec.ts
charliemeyer2000: fixed.

Timer (approved):
LGTM. Can we add an integration test?

ijjk (approved):

codybrouwers (approved):

PR #13387, opened 2025-05-28
[rolling-release] feat: adding in rolling-release support

brookemosby:


packages/cli/src/commands/promote/request-promote.ts
```diff
@@ -70,11 +70,20 @@ export default async function requestPromote({
...
+    `/v10/projects/${project.id}/promote/${deployment.id}`,
+    {
+      body: {}, // required
+      json: false,
+      method: 'POST',
+    }
+  );
+  if (res.status === 202) {
+    output.log(
+      'Promotion has been queued and will begin when the active rolling release completes successfully.'
```
trek, line 83: In other commands that take a while, we some times long poll so we can show the result to the user with a `--no-wait` option for CI uses where waiting is not required. Not strictly necessary here if that doesn't match how this feature works, so I wouldn't block on it but maybe consider adding.

I _could_ see this being useful for custom workflows where a failed promote exits with a non `0` and their script needs to take some kind of action.

Alternatively, is there a URL we can log immediately for a user to visit so they can watch the rolling release progress?

packages/cli/src/commands/rolling-release/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const projectNameOrId = parsedArguments.flags['--name'];
+  const currentStageIndex = parsedArguments.flags['--currentStageIndex'];
+  const activeStageIndex = parseInt(currentStageIndex ?? '');
+  const deployId = parsedArguments.flags['--deployId'];
+  const cfgString = parsedArguments.flags['--cfg'];
+  let cfg = undefined;
+
+  telemetry.trackCliOptionName(projectNameOrId);
+  telemetry.trackCliOptionAction(parsedArguments.flags['--action']);
```
trek, line 52: Any reason we went with `--action` over a subcommand? `vercel rr --action promote` vs `vercel rr promote`?

trek (approved):

packages/cli/src/commands/rolling-release/index.ts
brookemosby: No reason why - just updated to use subcommand so its simpler.

PR #13391, opened 2025-05-28
Fix type-check task

anthonyshew:
Ensures type check tasks run in parallel while respecting their dependencies.


trek (dismissed):

chris-olszewski (dismissed):
We'll need to fixup the broken types in this PR probably. I think there needs to be a changeset as well for CI to work properly

codybrouwers (dismissed):

packages/build-utils/src/types.ts
```diff
@@ -49,15 +49,31 @@ export interface Config {
...
     }
   | {
       type: 'header' | 'cookie' | 'query';
       key: string;
-      value?: string;
+      value?: MatchableValue;
     }
 >;
 
+type MatchableValue =
```
chris-olszewski: This is just copy-pasta'd from the type change in https://github.com/vercel/vercel/pull/13375

As the type error confirmed, these types must in fact match.

.changeset/thirty-donkeys-flash.md
```diff
@@ -0,0 +1,26 @@
...
+'@vercel/remix-builder': minor
+'@vercel/edge': minor
+'@vercel/next': minor
+'@vercel/node': minor
+'@vercel/ruby': minor
+'vercel': minor
+'@vercel/go': minor
+---
+
+Adds support for conditionValues for `has` and `missing` and `mitigate` in your `vercel.json` file.
```
chris-olszewski: @trek is this the right list for a change to the `vercel.json` definition?

packages/build-utils/src/types.ts
charliemeyer2000: Suggested fix - import { HasField } from routing utils and then export it rather than doing the copy/paste type change

turbo.json
```diff
@@ -53,6 +53,11 @@
       "dependsOn": ["build"],
       "outputLogs": "new-only"
     },
-    "type-check": {}
+    "type-check": {
+      "dependsOn": ["source"]
```
chris-olszewski: I think we might need to ammend this to include `^build`. You can't typecheck if you consume `build-utils`
https://github.com/vercel/vercel/blob/ccade8f6033e0bca8a0a08562dca2fd56af9d1fc/packages/build-utils/build.mjs#L3

turbo.json
```diff
@@ -53,6 +53,11 @@
       "dependsOn": ["build"],
       "outputLogs": "new-only"
     },
-    "type-check": {}
+    "type-check": {
+      "dependsOn": ["source", "^build"]
```
chris-olszewski, line 57: This is necessary as typecheck uses the `.d.ts` files generated during build

trek (approved):

PR #13404, opened 2025-05-30
feat(functions): add `VERCEL_OIDC_TOKEN` verification to development

balazsorban44:
A common issue occurs when `VERCEL_OIDC_TOKEN` is defined in multiple files (e.g., in Next.js, `.env.development.local`, and `.env.local`) by accident.

Even if `vercel env pull` succeeds in updating one of the `VERCEL_OIDC_TOKEN` values, the environment may still be incorrect, as the framework may be resolving the expired token instead.

This PR introduces a dev-only warning for such cases to aid the developer in troubleshooting.

Related [IAM-3841](https://linear.app/vercel/issue/IAM-3841/document-vercel-oidc-token-troubleshooting)


packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -23,15 +25,31 @@ import { getContext } from '../get-context';
...
+    getContext().headers?.['x-vercel-oidc-token'] ?? env.VERCEL_OIDC_TOKEN;
 
   if (!token) {
     throw new Error(
-      `The 'x-vercel-oidc-token' header is missing from the request. Do you have the OIDC option enabled in the Vercel project settings?`
+      `The Vercel OIDC token is missing in your environment. Do you have the OIDC option enabled in the Vercel project settings?`
     );
   }
 
+  if (env.VERCEL_ENV === 'development') {
```
Schniz: consider adding `process.env.NODE_ENV !== "production"` so bundlers will remove that piece of code

packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -29,9 +30,25 @@ export async function getVercelOidcToken(): Promise<string> {
 
   if (!token) {
     throw new Error(
-      `The 'x-vercel-oidc-token' header is missing from the request. Do you have the OIDC option enabled in the Vercel project settings?`
+      `The Vercel OIDC token is missing in your environment. Do you have the OIDC option enabled in the Vercel project settings?`
     );
   }
 
+  if (process.env.NODE_ENV !== 'production') {
```
shaper, line 37: Would this log in a preview deployment? The message is specific to development.

packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -29,9 +30,25 @@ export async function getVercelOidcToken(): Promise<string> {
...
+    const { createRemoteJWKSet, jwtVerify, errors } = await import('jose');
+    try {
+      const jwks = createRemoteJWKSet(new URL('https://oidc.vercel.com'));
+      await jwtVerify(token, jwks);
```
shaper, line 41: In AI Gateway use, the expiry failed here: https://github.com/vercel/ai-gateway/blob/6dae0076614222e5a810d7d0c0f45749444c3221/app/v1/ai/language-model/extract-headers.ts#L43
which runs before the `jwtVerify` call at line 57 there.

That code grew organically -- I wanted to be able to explicitly identify and handle expiration with custom message, error detail, etc.

For the use case here where a dev had an otherwise-valid but expired token, will expiry fail the `jwtVerify` call?

shaper (approved):
This is a nice proactive enhancement, thanks!

We would need to copy some of this to the implementation [here](https://github.com/vercel/ai/blob/v5/packages/gateway/src/get-vercel-oidc-token.ts) FYI, I can do it once we land this.

packages/functions/src/oidc/get-vercel-oidc-token.ts
balazsorban44: Preview deployments are `NODE_ENV='production'` too, as we deploy the production-ready outputs. Only `VERCEL_ENV` would be `preview`, so this is fine.

packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -29,9 +30,25 @@ export async function getVercelOidcToken(): Promise<string> {
...
+    } catch (error) {
+      if (error instanceof errors.JOSEError) {
+        console.warn(
+          'Invalid Vercel OIDC Token. Do you have multiple `.env.*` files with `VERCEL_OIDC_TOKEN` defined?',
+          error
+        );
+      }
+      throw error;
+    }
```
Schniz, line 50: gonna be an annoying person, but can we extract that into a separate function `async function validateToken` to easily grok what's happening in this function?

packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -29,9 +30,25 @@ export async function getVercelOidcToken(): Promise<string> {
...
+    const { createRemoteJWKSet, jwtVerify, errors } = await import('jose');
+    try {
+      const jwks = createRemoteJWKSet(new URL('https://oidc.vercel.com'));
+      await jwtVerify(token, jwks);
+    } catch (error) {
+      if (error instanceof errors.JOSEError) {
+        console.warn(
+          'Invalid Vercel OIDC Token. Do you have multiple `.env.*` files with `VERCEL_OIDC_TOKEN` defined?',
+          error
+        );
```
Schniz, line 47: nit: should we have a custom error class instead of printing a warning? This will help having a nice `@throws` and allowing people to keep, ignore, reroute the warnings etc

PR #13410, opened 2025-05-30
[package/next] output filePath in deterministic order

marc-vercel:
Ensure that the output of `collectTracedFiles` is deterministic.


packages/next/src/utils.ts
```diff
@@ -3524,9 +3526,7 @@ export async function getNodeMiddleware({
   const tracedFiles: Record<string, FileFsRef> = {};
 
   await Promise.all(
-    fileList.map(
-      collectTracedFiles(baseDir, lstatResults, lstatSema, reasons, tracedFiles)
-    )
+    fileList.map(collectTracedFiles(baseDir, lstatResults, lstatSema, reasons))
   );
```
sokra: `tracedFiles = Object.fromEntries` is missing here

marc-vercel: Fixed in https://github.com/vercel/vercel/pull/13410/commits/ef2e258574a2f6a9bab1781211cc37ba239be427

sokra (approved):

LukeSheard (approved):

PR #13420, opened 2025-06-04
feat: rolling releases - use subcommands

brookemosby:
Using subcommands instead of options for a cleaner flow. Also using linked project.


codybrouwers (dismissed):

packages/cli/src/commands/rolling-release/command.ts
```diff
@@ -1,82 +1,178 @@
...
-      description: "Action to perfom on a project's rolling release.",
-      enum: ['configure', 'start', 'approve', 'abort', 'complete', 'fetch'],
+      name: 'start',
+      description: 'Start a rolling release',
+      aliases: [],
+      arguments: [],
+      examples: [
+        {
+          name: 'Start a rolling release',
+          value: `${packageName} rolling-release start --deployId=dpl_123`,
```
styfle: Does this work with a deployment url instead of deployment id?

For example, can I do

```sh
vc release start --deployId=$(vc deploy)
```

packages/cli/src/commands/rolling-release/command.ts
```diff
@@ -1,82 +1,178 @@
...
+        },
+      ],
+      options: [
+        {
+          name: 'deployId',
```
styfle: Can we change this from `deployId` to `dpl`? That way you could do

```sh
vc release start --dpl=dpl_123
```

Or even shorter

```sh
vc release start --dpl=123
```

packages/cli/src/commands/rolling-release/command.ts
```diff
@@ -1,82 +1,178 @@
+import { packageName } from '../../util/pkg-name';
+
 export const rollingReleaseCommand = {
   name: 'rolling-release',
-  aliases: ['rr'],
+  aliases: ['rr', 'release'],
```
styfle: I think we should make `vc release` without any additional options be the new `vc promote`.

Then you can pass options to make it rolling.

https://vercel.com/docs/cli/promote

Thoughts @trek @TooTallNate ?

brookemosby: Promote currently requires deploymentId/URL

styfle: Right and I think we should make `vc release` also work with both.

See previous comment: https://github.com/vercel/vercel/pull/13420#discussion_r2132467712

styfle: Imagine you're only using the cli to deploy and roll the release. Where would you even get the deployment id? I don't think we print it out, just the url.

packages/cli/src/commands/rolling-release/command.ts
brookemosby: Just updated so it will 👍

packages/cli/src/commands/rolling-release/command.ts
brookemosby: Just updated so it will work with either param

packages/cli/src/commands/rolling-release/index.ts
```diff
@@ -22,13 +24,37 @@ export default async (client: Client): Promise<number> => {
...
   let parsedArguments = null;
 
-  const flagsSpecification = getFlagsSpecification(
-    rollingReleaseCommand.options
+  // Get the subcommand, accounting for both full command name and aliases
+  const args = client.argv.slice(2);
+  const firstArg = args[0];
+
+  // Check if first arg is the command name or an alias
+  const isCommandOrAlias = firstArg === 'rolling-release' || firstArg === 'rr';
```
onsclom: Looks like this should include the `release` alias:
```suggestion
  const isCommandOrAlias = firstArg === 'rolling-release' || firstArg === 'rr' || firstArg === 'release';
```

Also, maybe the `getSubcommand` utility would come in handy here? Looks like this is used in other commands with subcommands: https://github.com/vercel/vercel/blob/3fedd2a6ea794032bc2c102808b0077f06548e5a/packages/cli/src/commands/alias/index.ts#L46

packages/cli/src/commands/rolling-release/command.ts
onsclom: If I understand right, @styfle is suggesting `vc release` shouldn't be an alias of of `rolling-release`, and should instead be an alias of `promote`. I agree that `vc release` feels more like `vc promote` than `vc rolling-release`.

Maybe this PR just shouldn't include the `release` alias, and a future PR could add `release` as an alias to `vc promote` with options for rolling releases?

styfle: Yes I agree that is the best path forward. Use the existing `vc promote` to add this new functionality. It feels like rolling releases are just a newer way to promote.

For more context, see original announcement here:
https://vercel.com/changelog/stage-and-manually-promote-deployments-to-production

onsclom (approved):

PR #13426, opened 2025-06-06
[@vercel/functions] support runtime cache usage at build time

kldavis4:
# problem

We want to support usage of the runtime cache api during builds but currently this will use the in memory cache implementation

# solution

Add client implementation that is used during builds when cache implementation is not present on the client and the appropriate environment variables are present


packages/functions/src/cache/types.ts
```diff
@@ -67,4 +65,11 @@ export interface CacheOptions {
    * Optional separator string for the namespace.
    */
   namespaceSeparator?: string;
+
+  /**
+   * This function can be used to add cache tags that will be used for cache invalidation.
+   *
+   * @param {(tags: string|string[]) => void} [addCacheTags] - Function to add cache tags.
+   */
+  addCacheTags?: (tags: string | string[]) => void;
```
lubakravche: we can remove this one right?

packages/functions/src/cache/build-client.ts
```diff
@@ -0,0 +1,119 @@
...
+      if (res.status === 404) {
+        return null;
+      }
+      if (res.status === 200) {
+        const cacheState = res.headers.get('x-vercel-cache-state');
+        const content = (await res.json()) as unknown;
+        if (content && cacheState !== 'fresh') {
+          await this.delete(key);
+          return null;
+        }
```
lubakravche: we should add types for the cache states:
```
export enum PkgCacheState {
  Fresh     = 'fresh',
  Stale     = 'stale',
  Expired   = 'expired',
  NotFound  = 'notFound',
  Error     = 'error',
}
```

we can make this more efficient by not reading the content into memory unless it's fresh and only delete it if it exists:

```
// Header may be missing, so union with `null`
  const cacheState = res.headers.get(
    'x-vercel-cache-state',
  ) as PkgCacheState | null;
  
  if (cacheState !== PkgCacheState.Fresh) {
    if (
      cacheState === PkgCacheState.Stale ||
      cacheState === PkgCacheState.Expired
    ) {
      await del(key);
    }

    // Always release the body so the connection can be reused
    res.body?.cancel?.();
    return null;
  }

  // Only here do we incur the cost of buffering & parsing JSON.
  return (await res.json()) as T;
}
```

lubakravche: discussed async but let's remove the key deletion too, it feels racey

lubakravche: also lets make these changes for the non-build case too

packages/functions/src/cache/build-client.ts
```diff
@@ -0,0 +1,119 @@
...
+    options?: {
+      name?: string;
+      ttl?: number;
+      tags?: string[];
+    }
+  ) => {
+    try {
+      const optionalHeaders: Record<string, string> = {};
+      if (options?.ttl) {
+        optionalHeaders['x-vercel-revalidate'] = options.ttl.toString();
```
lubakravche: can we pull all these headers into constants

kldavis4: done

packages/functions/src/cache/build-client.ts
kldavis4: updated

lubakravche (approved):

cramforce (approved):

smaeda-ks (approved):

PR #13429, opened 2025-06-09
feat: introduce `getVercelOidcTokenSync`

markandrus:
`getVercelOidcToken` doesn't need to be async, and there are places we'd like to use it where it would be more convenient if it were sync. Therefore, this PR introduces a new, synchronous version: `getVercelOidcTokenSync`.

There are still some places that could benefit from an asynchronous `getVercelOidcToken`, and so this version is still there, and not marked as deprecated.


packages/functions/docs/modules/oidc.md
```diff
@@ -97,3 +98,38 @@ A promise that resolves to the OIDC token.
...
+### getVercelOidcTokenSync
+
+▸ **getVercelOidcTokenSync**(): `string`
+
+Returns the OIDC token from the request context or the environment variable.
+
+This function is used to retrieve the OIDC token from the request context or the environment variable.
+It checks for the `x-vercel-oidc-token` header in the request context and falls back to the `VERCEL_OIDC_TOKEN` environment variable if the header is not present.
+
+context headers.
```
markandrus: Looks like there is a typo here and in the original `getVercelOidcToken` documentation to be cleaned up.

trek (approved):

marcgreenstock (approved):

packages/functions/src/oidc/get-vercel-oidc-token.ts
```diff
@@ -23,6 +23,29 @@ import { getContext } from '../get-context';
  * ```
  */
 export async function getVercelOidcToken(): Promise<string> {
+  return getVercelOidcTokenSync();
```
dimitropoulos, line 26: can we please just remove the old one, in that case?  it's a noop if someone awaits a sync function, anyway.

so I'm saying instead of creating a new function, can we just update the existing one to not be async?

markandrus: I was also wondering about this, but @marcgreenstock pointed me to some code samples we have showing how to use this with [Azure](https://vercel.com/docs/oidc/azure#query-an-azure-cosmosdb-instance) and [GCP](https://vercel.com/docs/oidc/gcp#return-gcp-vertex-ai-generated-text). In those code samples, we pass `getVercelOidcToken` precisely where a `() => Promise<string>` is expected, and so I think changing them to `() => string` would be breaking.

dimitropoulos: can we update our code snippets to show `() => string | Promise<string>`?  that'd be a backwards compatible change (and we can make that change here too, of course).

markandrus: We could cut a new major release of `@vercel/functions` and update the code snippets like this:

**Azure**

```diff
 const tokenCredentials = new ClientAssertionCredential(
   AZURE_TENANT_ID,
   AZURE_CLIENT_ID,
-  getVercelOidcToken,
+  async () => getVercelOidcToken(),
 );
```

**GCP**

```diff
 // Initialize the External Account Client
 const authClient = ExternalAccountClient.fromJSON({
   type: 'external_account',
   audience: `//iam.googleapis.com/projects/${GCP_PROJECT_NUMBER}/locations/global/workloadIdentityPools/${GCP_WORKLOAD_IDENTITY_POOL_ID}/providers/${GCP_WORKLOAD_IDENTITY_POOL_PROVIDER_ID}`,
   subject_token_type: 'urn:ietf:params:oauth:token-type:jwt',
   token_url: 'https://sts.googleapis.com/v1/token',
   service_account_impersonation_url: `https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/${GCP_SERVICE_ACCOUNT_EMAIL}:generateAccessToken`,
   subject_token_supplier: {
     // Use the Vercel OIDC token as the subject token
-    getSubjectToken: getVercelOidcToken,
+    getSubjectToken: async () => getVercelOidcToken(),
   },
 });
```

I'm not sure this change is worth already doing that, though. Having the async interface be the default is more ergonomic for Azure and GCP, and it (hopefully) discourages caching the result at the module level (unless you do top-level await, which I think is atypical). IMO, we could wait to make a decision on the default version's return type and avoid a major version bump.

PR #13438, opened 2025-06-11
fix(vercel): exit dev command on error

chris-olszewski:
Reenable test disabled in #13436

According to `git bisect` https://github.com/vercel/vercel/pull/13226 broke this behavior. Unsure of how the passed on CI/up until recently as it was definitely broken.

It looks like the `AbortController` prevented the process from exiting.

### DIFF
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -108,7 +108,11 @@ export default async function dev(
         telemetry.trackOidcTokenRefresh(++refreshCount);
       }
     } catch (error) {
-      output.debug(`Failed to refresh OIDC token: ${error}`);
+      // Throw any error aside from an abort error.
+      if (!(error instanceof Error && error.name === 'AbortError')) {
+        throw error;
+      }
+      output.debug('OIDC token refresh was aborted');
     }
   });
 
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -93,18 +93,22 @@ export default async function dev(
   const timeout = setTimeout(async () => {
     if (link.status !== 'linked') return;
 
-    let refreshCount = 0;
-    for await (const token of refreshOidcToken(
-      controller.signal,
-      client,
-      link.project.id,
-      envValues,
-      'vercel-cli:dev'
-    )) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/dev/dev.ts
```diff
@@ -126,5 +130,10 @@ export default async function dev(
     }
   }
 
-  await devServer.start(...listen);
+  try {
+    await devServer.start(...listen);
+  } finally {
+    clearTimeout(timeout);
+    controller.abort();
```
chris-olszewski, line 141: Always clean up oidc refresh even if dev server has an error

packages/cli/src/commands/dev/dev.ts
```diff
@@ -93,18 +93,22 @@ export default async function dev(
...
-      client,
-      link.project.id,
-      envValues,
-      'vercel-cli:dev'
-    )) {
-      output.debug(`Refreshing ${chalk.green(VERCEL_OIDC_TOKEN)}`);
-      envValues[VERCEL_OIDC_TOKEN] = token;
-      await devServer.runDevCommand(true);
-      telemetry.trackOidcTokenRefresh(++refreshCount);
+    try {
```
chris-olszewski, line 96: Required to ensure an error from the dev server is surfaced to the user and not an `AbortError`

packages/cli/src/commands/dev/dev.ts
```diff
@@ -93,18 +93,22 @@ export default async function dev(
...
+        envValues,
+        'vercel-cli:dev'
+      )) {
+        output.debug(`Refreshing ${chalk.green(VERCEL_OIDC_TOKEN)}`);
+        envValues[VERCEL_OIDC_TOKEN] = token;
+        await devServer.runDevCommand(true);
+        telemetry.trackOidcTokenRefresh(++refreshCount);
+      }
+    } catch (error) {
+      output.debug(`Failed to refresh OIDC token: ${error}`);
```
chris-olszewski: Unsure of what should happen if `refreshOidcToken` fails. @markandrus do you have an opinion on what the behavior should be?

trek (dismissed):
LGTM, but let's give @markandrus time to chime in.

trek (approved):

PR #13440, opened 2025-06-11
feat(cli): load env vars to memory via `vc env`

balazsorban44:
Right now, when you want to copy environment variables to your machine, your only option is to pull them into a file via `vc env pull .env.local`.

Also, `vc env` is just printing what `vc env --help` would.

We could utilize this command to load the environment variables for a subprocess, instead of having to write them to a file.

**Example:**
```sh
vc env next dev
```

_This has a **security benefit** since we do not preserve the environment variables after the process exits._

The implementation reuses `vc env pull` and introduces a new `--memory` flag, which is added if the `vc env` command is called without any known subcommand.

Usage:

[Screencast from 2025-06-11 15-47-39.webm](https://github.com/user-attachments/assets/210f6a52-c7fd-44b0-80ac-cf8858db715c)

You can test it via 

```sh 
cd packages/cli &&
  pnpm vercel link &&
  pnpm vercel env node -e 'console.log(Boolean(process.env.VERCEL_OIDC_TOKEN))'` // true
```

### DIFF
--- a/packages/cli/src/commands/env/command.ts
+++ b/packages/cli/src/commands/env/command.ts
@@ -148,7 +148,7 @@ export const pullSubcommand = {
       required: false,
     },
     {
-      name: 'process',
+      name: 'command',
       required: false,
     },
   ],
@@ -194,7 +194,7 @@ export const pullSubcommand = {
     },
     {
       name: 'Run a process with the pulled Environment Variables without writing to a file',
-      value: [`${packageName} env <process>`, `${packageName} env next dev`],
+      value: [`${packageName} env <command>`, `${packageName} env next dev`],
     },
   ],
 } as const;
--- a/packages/cli/src/commands/env/command.ts
+++ b/packages/cli/src/commands/env/command.ts
@@ -141,10 +141,14 @@ export const pullSubcommand = {
   name: 'pull',
   aliases: [],
   description:
-    'Pull all Development Environment Variables from the cloud and write to a file [.env.local]',
+    'Load all Development Environment Variables from the cloud to a file [.env.local] or a process',
   arguments: [
     {
       name: 'filename',
+      required: false,
+    },
+    {
+      name: 'process',
       required: false,
     },
   ],
@@ -169,7 +173,15 @@ export const pullSubcommand = {
     {
       ...yesOption,
       description:
-        'Skip the confirmation prompt when removing an environment variable',
+        'Skip the confirmation prompt when updating an existing environment variable file',
+    },
+    {
+      name: 'memory',
+      shorthand: null,
+      description:
+        'Load Environment Variables into memory instead of a file. When this command exits, the variables are lost.',
+      type: Boolean,
+      deprecated: false,
     },
   ],
   examples: [
@@ -179,6 +191,10 @@ export const pullSubcommand = {
         `${packageName} env pull <file>`,
         `${packageName} env pull .env.development.local`,
       ],
+    },
+    {
+      name: 'Run a process with the pulled Environment Variables without writing to a file',
+      value: [`${packageName} env <process>`, `${packageName} env next dev`],
     },
   ],
 } as const;
--- a/packages/cli/src/commands/env/command.ts
+++ b/packages/cli/src/commands/env/command.ts
@@ -199,6 +199,20 @@ export const pullSubcommand = {
   ],
 } as const;
 
+export const commandSubcommand = {
+  name: '<command>',
+  aliases: [],
+  description: 'The subcommand to run with the pulled Environment Variables',
+  arguments: [],
+  options: [],
+  examples: [
+    {
+      name: 'Run a process with the pulled Environment Variables without writing to a file',
+      value: [`${packageName} env <command>`, `${packageName} env next dev`],
+    },
+  ],
+} as const;
+
 export const envCommand = {
   name: 'env',
   aliases: [],
@@ -209,6 +223,7 @@ export const envCommand = {
     listSubcommand,
     pullSubcommand,
     removeSubcommand,
+    commandSubcommand,
   ],
   options: [],
   examples: [],
--- a/packages/cli/src/commands/env/index.ts
+++ b/packages/cli/src/commands/env/index.ts
@@ -62,28 +62,39 @@ export default async function main(client: Client) {
     output.print(
       help(command, { parent: envCommand, columns: client.stderr.columns })
     );
-    return 2;
-  }
-
-  if (needHelp && subcommand) {
-    telemetry.trackCliFlagHelp('env', subcommandOriginal);
   }
 
   switch (subcommand) {
     case 'ls':
-      if (needHelp) return printHelp(listSubcommand);
+      if (needHelp) {
+        telemetry.trackCliFlagHelp('env', subcommandOriginal);
+        printHelp(listSubcommand);
+        return 2;
+      }
       telemetry.trackCliSubcommandList(subcommandOriginal);
       return ls(client, args);
     case 'add':
-      if (needHelp) return printHelp(addSubcommand);
+      if (needHelp) {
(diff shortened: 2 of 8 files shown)


packages/cli/src/commands/env/command.ts
```diff
@@ -141,12 +141,16 @@ export const pullSubcommand = {
...
   description:
-    'Pull all Development Environment Variables from the cloud and write to a file [.env.local]',
+    'Load all Development Environment Variables from the cloud to a file [.env.local] or a process',
   arguments: [
     {
       name: 'filename',
       required: false,
     },
+    {
+      name: 'process',
```
chris-olszewski: I think command is more accurate than process here
```suggestion
      name: 'command',
```

packages/cli/src/commands/env/command.ts
```diff
@@ -180,6 +192,10 @@ export const pullSubcommand = {
         `${packageName} env pull .env.development.local`,
       ],
     },
+    {
+      name: 'Run a process with the pulled Environment Variables without writing to a file',
+      value: [`${packageName} env <process>`, `${packageName} env next dev`],
```
chris-olszewski:
```suggestion
      value: [`${packageName} env <command>`, `${packageName} env next dev`],
```

packages/cli/test/unit/commands/env/index.test.ts
```diff
@@ -59,12 +62,12 @@ describe('env', () => {
...
 
-    it('shows help', async () => {
-      const args: string[] = ['not-a-command'];
+    it('inherits variables', async () => {
+      const args = ['not-a-command'];
 
       client.setArgv('env', ...args);
       const exitCode = await env(client);
-      expect(exitCode).toEqual(2);
+      expect(exitCode).toEqual(0);
```
chris-olszewski, line 70: Could you update this test to verify that the variables are passed to the child process?

packages/cli/test/unit/commands/env/index.test.ts
```diff
@@ -24,13 +24,16 @@ describe('env', () => {
...
 
-  it('errors when invoked without subcommand', async () => {
+  it('Show subprocess help when invoked without subcommand', async () => {
     client.setArgv('env');
     const exitCodePromise = env(client);
-    await expect(exitCodePromise).resolves.toBe(2);
+    await expect(exitCodePromise).resolves.toBe(1);
+    expect(client.stderr).toOutput(
+      'Invalid number of arguments. Usage: `vercel env <process>`'
+    );
```
chris-olszewski, line 33: I think this is a regression in the helpfulness of the text output of `vercel env`. It goes from
```
Error: Please specify a valid subcommand: ls | add | rm | pull

  ▲ vercel env command

  Interact with Environment Variables for a Project                                                               

  Commands:

  add     name [environment]          Add an Environment Variable (see examples below)                      
  list    [environment] [git-branch]  List all Environment Variables for a Project                          
  pull    [filename]                  Pull all Development Environment Variables from the cloud and write to
                                      a file [.env.local]                                                   
  remove  name [environment]          Remove an Environment Variable (see examples below)
```
to
```
Error: Invalid number of arguments. Usage: `vercel env <process>`
       See also: `vercel env --help`
```
which requires an additional `vercel env --help` to discover what commands are available.

packages/cli/src/commands/env/pull.ts
```diff
@@ -174,6 +198,51 @@ export async function envPullCommandLogic(
...
+
+  if (memory) {
+    output.stopSpinner();
+
+    const [command, ...rest] = args;
+
+    const child = spawn(command, rest, {
+      env: { ...process.env, ...environmentVariables },
+      stdio: 'inherit',
+    });
```
chris-olszewski, line 225: This doesn't properly setup signal forwarding from the `vercel` process to the child process. Can you make sure that sending `SIGINT` to `vercel` results in the child process shutting down? I think you'll need to setup a process group where `vercel` is the leader.

```
# Get process tree of the vercel cli
[0 olszewski@macbookpro] /Users/olszewski/code/vercel/vercel $ pstree -p 88056
-+= 00001 root /sbin/launchd
 \-+= 16827 olszewski tmux
   \-+= 98314 olszewski -zsh
     \-+= 88056 olszewski node /Users/olszewski/.nvm/versions/node/v20.19.2/bin/pnpm vercel env pnpm dev --cwd /Users/olszewski/code/vercel/turborepo/docs/si
       \-+- 88077 olszewski node /Users/olszewski/code/vercel/vercel/packages/cli/node_modules/.bin/../../../../node_modules/.pnpm/ts-node@10.9.1_@swc+core@1
         \-+- 88123 olszewski node /Users/olszewski/.nvm/versions/node/v20.19.2/bin/pnpm dev
           \-+- 88145 olszewski node /Users/olszewski/code/vercel/turborepo/docs/site/node_modules/.bin/../next/dist/bin/next dev --turbopack
             |--- 88210 olszewski next-server (v15.4.0-canary.23) 
             \--- 88233 olszewski /Users/olszewski/code/vercel/turborepo/node_modules/.pnpm/@esbuild+darwin-arm64@0.25.2/node_modules/@esbuild/darwin-arm64/b
# Send an interrupt to vercel cli
[0 olszewski@macbookpro] /Users/olszewski/code/vercel/vercel $ kill -s INT 88056
# Check if next server is running, note that the `pnpm dev` is no longer a child of vercel, but the root process
[0 olszewski@macbookpro] /Users/olszewski/code/vercel/vercel $ pstree -p 88123
-+= 00001 root /sbin/launchd
 \-+- 88123 olszewski node /Users/olszewski/.nvm/versions/node/v20.19.2/bin/pnpm dev
   \-+- 88145 olszewski node /Users/olszewski/code/vercel/turborepo/docs/site/node_modules/.bin/../next/dist/bin/next dev --turbopack
     |-+- 88210 olszewski next-server (v15.4.0-canary.23) 
     | |--- 89346 olszewski node /Users/olszewski/code/vercel/turborepo/docs/site/.next/transform.js 52496
     | \--- 89347 olszewski node /Users/olszewski/code/vercel/turborepo/docs/site/.next/transform.js 52498
     \--- 88233 olszewski /Users/olszewski/code/vercel/turborepo/node_modules/.pnpm/@esbuild+darwin-arm64@0.25.2/node_modules/@esbuild/darwin-arm64/bin/esbui
```

packages/cli/src/commands/env/index.ts
```diff
@@ -99,8 +98,6 @@ export default async function main(client: Client) {
       telemetry.trackCliSubcommandPull(subcommandOriginal);
       return pull(client, args);
     default:
-      output.error(getInvalidSubcommand(COMMAND_CONFIG));
-      output.print(help(envCommand, { columns: client.stderr.columns }));
-      return 2;
+      return pull(client, args.concat('--memory'));
```
chris-olszewski: I think creating a new `exec` subcommand is the safer option here. Having inferred subcommands can really end up restricting changes we can make to `vercel env` in the future without being a breaking change.

chris-olszewski (changes requested):
~I feel like this would be better suited as a new subcommand for `env`.~ Did not realize this had already been deliberated, ignore this suggestion.

packages/cli/src/commands/env/index.ts
balazsorban44: I think this is a good use case for no subcommands, but we can do this as a follow-up after further discussion, thanks!

packages/cli/test/unit/commands/env/index.test.ts
chris-olszewski: I do feel we should have a check for if there's no args provided to `vercel env` and show `vercel env --help` instead of an "Invalid number of arguments"

balazsorban44: Agree! Will fix!

PR #13448, opened 2025-06-13
[@vercel/functions] prefix tags in runtime cache

kldavis4:
# problem

We want to keep existing ISR tags separate from the runtime cache tags because currently there is not good support for using runtime cache tags with ISR pages

# solution

Namespace tags by default


kldavis4:
note that since ISR tags don't have a namespace, it is possible for these tags to collide. Maybe it makes sense to have a more unique default namespace?

PR #13455, opened 2025-06-17
[@vercel/functions] add warning message to indicate in-memory cache is being used

kldavis4:
Warn when falling back to in-memory caching to prevent invalid expectations if for some reason runtime cache is not available in the current context.

### DIFF
--- a/.changeset/big-bottles-explain.md
+++ b/.changeset/big-bottles-explain.md
@@ -1,0 +1,5 @@
+---
+'@vercel/functions': patch
+---
+
+Add warning when calling getCache to indicate that in-memory cache is being used.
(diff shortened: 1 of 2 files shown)


lubakravche (approved):

shraddha55 (approved):

shraddha55:
@kldavis4 with this change we should we able to verify in-memory cache is getting used in sc middleware runtime cache e2e tests? and see this log on browser console

kldavis4:
> @kldavis4 with this change we should we able to verify in-memory cache is getting used in sc middleware runtime cache e2e tests? and see this log on browser console

It will log a warning message but won't fail - we could add a no-fallback env var to force it to fail when the env var is set

Schniz (approved):

PR #13457, opened 2025-06-17
Make deploy logs default.

anthonyshew:


anthonyshew:
@trek is about to do a better job than me.

PR #13458, opened 2025-06-17
[cli] make `--logs` the default behavior for `vercel deploy`

trek:
* Updates `vercel deploy` to show build logs by default.
* Makes `vercel deploy --logs` warn about the option being deprecated.
* Introduces `vercel deploy --no-logs` to preserve old behavior as migration path for customers who prefer no logs.

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -102,6 +102,7 @@ export default async (client: Client): Promise<number> => {
     );
     telemetryClient.trackCliFlagPublic(parsedArguments.flags['--public']);
     telemetryClient.trackCliFlagLogs(parsedArguments.flags['--logs']);
+    telemetryClient.trackCliFlagNoLogs(parsedArguments.flags['--no-logs']);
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/deploy/index.ts
```diff
@@ -505,7 +511,7 @@ export default async (client: Client): Promise<number> => {
       target,
       skipAutoDetectionConfirmation: autoConfirm,
       noWait,
-      withLogs: parsedArguments.flags['--logs'],
+      withLogs: withLogs,
```
tknickman:
```suggestion
      withLogs,
```

tknickman (dismissed):
one suggestion, but not blocking - lgtm!

trek:
@tknickman lol, not sure how I went through the trouble of making a variable and then never used it. But also surprised we don't have linting configured to complain about this.

packages/cli/src/commands/deploy/index.ts
trek: I'm surprised we don't lint for this

tknickman (approved):

PR #13461, opened 2025-06-18
[@vercel/node] allow nodejs as runtime for middleware

Kikobeats:
This PR adds support for runtime: 'nodejs' configuration option in the Node.js builder for non-Next.js middleware files (root-level middleware.js/middleware.ts). 

Previously, middleware was forced to run as Edge Functions only. Now developers can choose between Edge runtime (default) and Node.js runtime for their middleware.


packages/node/src/serverless-functions/serverless-handler.mts
```diff
@@ -69,10 +70,25 @@ async function compileUserCode(
...
-    return getWebExportsHandler(listener, HTTP_METHODS);
+
+    let handler = listener;
+    if (options.isMiddleware) {
+      handler = HTTP_METHODS.reduce(
+        (acc, method) => {
+          acc[method] = listener;
+          return acc;
+        },
+        {} as Record<(typeof HTTP_METHODS)[number], ServerlessFunctionSignature>
```
jeffsee55, line 88: nit: perhaps TS struggling to infer from `.reduce`, could this work?

```ts
let handler: Record<
  (typeof HTTP_METHODS)[number],
  ServerlessFunctionSignature
> = listener;
if (options.isMiddleware) {
  handler = Object.fromEntries(
    HTTP_METHODS.map(method => [method, listener])
  );
}
 ```

File: packages/node/test/fixtures/63-nodejs-middleware/middleware.ts [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+};
+
+export default (req: Request) => {
+  const url = new URL(req.url);
+  const headers = new Headers({
+    'x-got-middleware': 'true',
+  });
+  if (url.pathname === '/' || url.pathname.startsWith('/api/')) {
+    // For `index.html` and `/api/node.js`, pass through
+    headers.set('x-middleware-next', '1');
```
jeffsee55, line 12: > For `index.html` and `/api/node.js`, pass through

What does that mean? Is this the only way to trigger this behavior? Ideally we don't need a next-specific header right?

jeffsee55 (approved):
👍

packages/node/test/fixtures/63-nodejs-middleware/middleware.ts
QuiiBz: `x-middleware-next` is not specific to Next.js but rather Vercel, and signals to continue processing the request (like an express `next()`). We also have e.g. `x-middleware-rewrite` to instead rewrite to a given URL, or `x-middleware-redirect` to redirect to a given URL: https://github.com/vercel/vercel/blob/main/packages/functions/src/middleware.ts

In this case it means requests hitting `/` or `/api/*` will continue and eventually hit the function `/api/node.js`, which is what the test asserts.

QuiiBz (approved):

PR #13462, opened 2025-06-18
feat(cli): add blob token option and env variable

luismeyer:
Prev the token was only read from .env.local. For better ci and local experience you can also pass the token as an option or as an env var now

### DIFF
--- a/.changeset/yellow-cars-cross.md
+++ b/.changeset/yellow-cars-cross.md
@@ -2,4 +2,4 @@
 "vercel": patch
 ---
 
-feat(cli) optimize blob token retrieval
+feat(cli): optimize blob token retrieval
--- a/.changeset/yellow-cars-cross.md
+++ b/.changeset/yellow-cars-cross.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+feat(cli) optimize blob token retrieval
--- a/.changeset/yellow-cars-cross.md
+++ b/.changeset/yellow-cars-cross.md
@@ -2,4 +2,4 @@
 "vercel": patch
 ---
 
-feat(cli): optimize blob token retrieval
+feat(cli): add blob token option and env variable
--- a/packages/cli/src/commands/blob/command.ts
+++ b/packages/cli/src/commands/blob/command.ts
@@ -235,6 +235,7 @@ export const blobCommand = {
   options: [
     {
       name: 'rw-token',
+      shorthand: null,
       type: String,
       deprecated: false,
       description: 'Read/write token for the Blob store',
--- a/packages/cli/src/commands/blob/command.ts
+++ b/packages/cli/src/commands/blob/command.ts
@@ -238,7 +238,7 @@ export const blobCommand = {
       shorthand: null,
       type: String,
       deprecated: false,
-      description: 'Read/write token for the Blob store',
+      description: 'Read_Write_Token for the Blob store',
       argument: 'String',
     },
   ],
--- a/packages/cli/src/commands/blob/command.ts
+++ b/packages/cli/src/commands/blob/command.ts
@@ -232,6 +232,14 @@ export const blobCommand = {
     copySubcommand,
     storeSubcommand,
   ],
-  options: [],
+  options: [
+    {
+      name: 'rw-token',
+      type: String,
+      deprecated: false,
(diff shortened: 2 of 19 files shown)


.changeset/yellow-cars-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"vercel": patch
+---
+
+feat(cli): optimize blob token retrieval
```
vvo: Probably should rename this and PR title too? So it's clear about what we did (we don't optimize, we allow to pass a custom token even if you haven't connected the store to a project right?)

vvo:
Are we able to get better error messages? Like if I run `vercel blob list` by default it should tell me something like:

```
No Vercel Blob token was found. Make sure to either:
- Link your current folder to a Vercel Project that has a Vercel Blob store connected
- Or pass a token like:
  - $command --token $VERCEL_BLOB_READ_WRITE_TOKEN; or
  - VERCEL_BLOB_READ_WRITE_TOKEN $command
```

Plus we should update the front/ documentation to explain what happens when you use `vercel blob list` or any command (it uses the currently configured project, or the specific token).

packages/cli/src/commands/blob/command.ts
```diff
@@ -232,6 +232,15 @@ export const blobCommand = {
...
     storeSubcommand,
   ],
-  options: [],
+  options: [
+    {
+      name: 'rw-token',
+      shorthand: null,
+      type: String,
+      deprecated: false,
+      description: 'Read/write token for the Blob store',
```
vvo:
```suggestion
      description: 'read-write token for the Blob store',
```

vvo: (to not be confused with "Read or write the token")

packages/cli/src/util/blob/token.ts
```diff
@@ -3,12 +3,48 @@ import { resolve } from 'node:path';
...
+import { getFlagsSpecification } from '../get-flags-specification';
+import { blobCommand } from '../../commands/blob/command';
+import { parseArguments } from '../get-args';
+import { getCommandName, packageName } from '../pkg-name';
+import cmd from '../output/cmd';
+import listItem from '../output/list-item';
+
+const ErrorMessage = `No Vercel Blob token found. To fix this issue, choose one of the following options:
+${listItem('Link your current folder to a Vercel Project that has a Vercel Blob store connected', 1)}
+${listItem(`Pass the token directly as an option: ${getCommandName('blob ls --token BLOB_TOKEN')}`, 2)}
```
vvo: I don't see "ls" in the Vercel Blob docs, is that correct? perhaps let's use list if that's what we document for now

packages/cli/src/util/blob/token.ts
```diff
@@ -3,12 +3,48 @@ import { resolve } from 'node:path';
...
+${listItem(`Set the Token as an environment variable: ${cmd(`VERCEL_BLOB_READ_WRITE_TOKEN=BLOB_TOKEN ${packageName} blob ls`)}`, 3)}`;
```
vvo: same

vvo (dismissed):

packages/cli/src/util/blob/token.ts
```diff
@@ -3,12 +3,48 @@ import { resolve } from 'node:path';
...
+import { getFlagsSpecification } from '../get-flags-specification';
+import { blobCommand } from '../../commands/blob/command';
+import { parseArguments } from '../get-args';
+import { getCommandName, packageName } from '../pkg-name';
+import cmd from '../output/cmd';
+import listItem from '../output/list-item';
+
+const ErrorMessage = `No Vercel Blob token found. To fix this issue, choose one of the following options:
+${listItem('Link your current folder to a Vercel Project that has a Vercel Blob store connected', 1)}
+${listItem(`Pass the token directly as an option: ${getCommandName('blob list --token BLOB_TOKEN')}`, 2)}
```
jeffsee55:
```suggestion
${listItem(`Pass the token directly as an option: ${getCommandName('blob list --rw-token BLOB_TOKEN')}`, 2)}
```

jeffsee55 (dismissed):

jeffsee55 (approved):

PR #13472, opened 2025-06-20
[frameworks] Add Nitro as framework

onsclom:
`nitro build` detects the env var `VERCEL=1` and emits build output API code, so Nitro already works great on Vercel with zero configuration.

### DIFF
--- a/examples/__tests__/integration/nitro.test.ts
+++ b/examples/__tests__/integration/nitro.test.ts
@@ -1,0 +1,4 @@
+import { deployExample } from '../test-utils';
+it('[examples] should deploy nitro', async () => {
+  await deployExample('nitro');
+});
--- a/examples/__tests__/integration/nitro.test.ts
+++ b/examples/__tests__/integration/nitro.test.ts
@@ -1,4 +1,5 @@
 import { deployExample } from '../test-utils';
-it('[examples] should deploy nitro', async () => {
+// eslint-disable-next-line jest/no-disabled-tests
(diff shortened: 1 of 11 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -1529,6 +1529,7 @@ export const frameworks = [
     website: 'https://nuxtjs.org',
     sort: 2,
     envPrefix: 'NUXT_ENV_',
+    supersedes: ['nitro'],
```
onsclom, line 1532: Looks like Nuxt projects sometimes list `nitro` as dep, and we want to prefer Nuxt for detection.

tknickman: Yea this is the right call - nitro is likely to be used by other frameworks as well so should generally be lower priority.

tknickman (dismissed):

examples/__tests__/integration/nitro.test.ts
```diff
@@ -0,0 +1,5 @@
+import { deployExample } from '../test-utils';
+// eslint-disable-next-line jest/no-disabled-tests
```
onsclom, line 2: Getting an error on with this test that goes away with a redeploy. Skipping for now because it works when manually deploying.

https://vercel.com/zero-conf-vtest314/nitro/FsHW5Q856PEFQ7cwVSb8HZa4dTNP?filter=errors#L40

tknickman (approved):

PR #13476, opened 2025-06-23
feat(routes-utils): move validation for max items for routes

tknickman:
We are moving this limit to be managed via the API instead.


trek (dismissed):

.changeset/rude-spoons-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/routing-utils': minor
```
trek, line 2: Is the value staying the same at the API? This might be a `major` version bump either way now that I think about it.

tknickman: Value stays the same - yes. Should we consider this breaking given it's going from less to more permissive? I would agree if it was the opposite. But I may not be thinking of a case when this could be considered breaking.

healeycodes (dismissed):

ijjk (dismissed):

trek:
Updated the test to remove the test for the removed behavior.

chris-olszewski (approved):

PR #13484, opened 2025-06-24
[cli][feature-flagged] adds stub `vercel guidance` command

trek:
Adds the base behavior for a "guidance" mode that will suggest to customers and agents common followup steps to commands.

For example, if you ran `vercel deploy` guidance might be something like (totally made up at the moment):

```
To check runtime logs run of your deployed application

vercel logs dpl_123

If these logs contain errors you can revert to the previous deployment with

vercel rollback

If these logs also contain errors, consider

vercel bisect

```

No output is suggested as part of this PR. These proposed changes introduce 

* `vercel guidance enable` to turn this behavior on
* `vercel guidance disable` to turn this behavior off
* `vercel guidance status` to see the current on/off status

Like the CLI's telemetry behavior, this may also be turned off with an environment variable `VERCEL_GUIDANCE_DISABLED=1`

This behavior is additionally disabled where we detect the CLI running in CI or on Vercel itself (identified by `CI=1`)

One first run in non-CI environment, the CLI will inform the user that the feature is on and provide steps for disabling if they choose.

A `patch` mostly because this is all behind a local env var flag `FF_GUIDANCE_MODE`.

### DIFF
--- a/packages/cli/src/commands/guidance/status.ts
+++ b/packages/cli/src/commands/guidance/status.ts
@@ -1,0 +1,13 @@
+import chalk from 'chalk';
+import output from '../../output-manager';
+import type Client from '../../util/client';
+
+export default async function status(client: Client) {
+  const enabled = client.config.guidance?.enabled !== false;
+
+  const status = enabled ? chalk.green('Enabled') : chalk.red('Disabled');
+  output.print('\n');
+  output.log(`${chalk.bold('Guidance status')}: ${status}\n`);
+
+  return 0;
+}
--- a/packages/cli/src/commands/guidance/status.ts
+++ b/packages/cli/src/commands/guidance/status.ts
@@ -7,7 +7,8 @@ export default async function status(client: Client) {
 
   const status = enabled ? chalk.green('Enabled') : chalk.red('Disabled');
   output.print('\n');
-  output.log(`${chalk.bold('Guidance status')}: ${status}\n`);
+  output.log(`${chalk.bold('Guidance status')}: ${status}`);
+  output.print('\n');
 
   return 0;
 }
--- a/packages/cli/src/util/guidance/check-status.ts
+++ b/packages/cli/src/util/guidance/check-status.ts
@@ -5,6 +5,13 @@ import output from '../../output-manager';
 
 export function checkGuidanceStatus({ config }: { config: GlobalConfig }) {
   if (!process.env.FF_GUIDANCE_MODE) {
+    // disabling guidance if not flagged into experimenting with it.
+    return;
+  }
+
+  if (process.env.CI) {
+    // disabling guidance initial enabling if in a CI environment
+    // which includes Vercel's build container.
     return;
   }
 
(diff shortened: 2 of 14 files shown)


packages/cli/src/util/guidance/check-status.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (config.guidance) {
+    // telemetry has been set previously by this check of
+    // user running vercel telemetry commands
+    return;
+  }
+
+  output.note(
+    'The Vercel CLI can suggest common follow-up commands and steps to help guide new users.'
+  );
+  output.log('You cand disable this feature by running:');
```
chris-olszewski:
```suggestion
  output.log('You can disable this feature by running:');
```

packages/cli/src/commands/guidance/status.ts
```diff
@@ -0,0 +1,14 @@
+import chalk from 'chalk';
+import output from '../../output-manager';
+import type Client from '../../util/client';
+
+export default async function status(client: Client) {
+  const enabled = client.config.guidance?.enabled !== false;
```
chris-olszewski, line 6: Should this be a shared function/method on config or client type?

Unsure if config verified type of config written to disk e.g. could a user write `guidance: { "enabled": "false" }` and we end up checking `!!client.config.guidance?.enabled` somewhere else down the line.

packages/cli/test/unit/util/guidance-check-status.test.ts
```diff
@@ -0,0 +1,144 @@
...
+          checkGuidanceStatus({
+            config: {},
+          });
+        });
+        it('does not inform the user', async () => {
+          await expect(client.stderr).not.toOutput(
+            'The Vercel CLI can suggest common follow-up commands and steps to help guide new users.'
+          );
+          await expect(client.stderr).not.toOutput(
+            'You cand disable this feature by running:'
```
chris-olszewski:
```suggestion
            'You can disable this feature by running:'
```

packages/cli/test/unit/util/guidance-check-status.test.ts
```diff
@@ -0,0 +1,144 @@
...
+            config: {},
+          });
+        });
+
+        it('informs the user', async () => {
+          await expect(client.stderr).toOutput(
+            'The Vercel CLI can suggest common follow-up commands and steps to help guide new users.'
+          );
+          await expect(client.stderr).toOutput(
+            'You cand disable this feature by running:'
```
chris-olszewski:
```suggestion
            'You can disable this feature by running:'
```

chris-olszewski (dismissed):
Minor typo and an honest question

packages/cli/src/commands/guidance/status.ts
trek: We do type this structure https://github.com/vercel/vercel/blob/83bccbfac19dbd63ff290ded3a7b96c5a1fad092/internals/types/index.d.ts#L56-L65

and only use the reader/writer methods internally when acting on the file. But, yeah, a human could edit the file by hand and cause issues.

chris-olszewski (approved):

PR #13489, opened 2025-06-25
[cli] add Google login functionality

javierbyte:
Adds "Continue with Google" option to the CLI. Sign in with Google on Vercel has been out for a couple of weeks now, it works the same as our Gitlab / Bitbucket integrations.


packages/cli/src/util/login/google.ts
```diff
@@ -0,0 +1,14 @@
+import { URL } from 'url';
```
balazsorban44, line 1: No need for this import, `URL` is on the global

balazsorban44 (dismissed):
LGTM

balazs4 (dismissed):
LGTM

jeffsee55 (dismissed):

javierbyte:
@balazs4 @jeffsee55 I changed from `patch` to `minor` and it seems that invalidated the reviews, can I get a ✅ again?

balazs4 (approved):
LGTM

PR #13491, opened 2025-06-25
[rr] fix rate limits for undefined project name fetches

brookemosby:
We were being rate limited, and potentially passing in empty projectIds, so this provides a fix for both.

### DIFF
--- a/packages/cli/src/commands/redeploy/index.ts
+++ b/packages/cli/src/commands/redeploy/index.ts
@@ -18,7 +18,11 @@ import { redeployCommand } from './command';
 import { getFlagsSpecification } from '../../util/get-flags-specification';
 import output from '../../output-manager';
 import { RedeployTelemetryClient } from '../../util/telemetry/commands/redeploy';
-import type { CustomEnvironment, Project } from '@vercel-internals/types';
+import type {
+  CustomEnvironment,
+  Project,
+  ProjectRollingRelease,
+} from '@vercel-internals/types';
 import {
   getCustomEnvironments,
   pickCustomEnvironment,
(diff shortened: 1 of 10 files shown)


dimitropoulos (dismissed):

packages/cli/src/commands/redeploy/index.ts
```diff
@@ -171,14 +175,21 @@ export default async function redeploy(client: Client): Promise<number> {
...
       );
-      const project = await getProjectByNameOrId(client, deployment.projectId);
-      if (project instanceof ProjectNotFound) {
-        throw project;
+      let project: Project | ProjectNotFound | undefined;
+      let rollingRelease: ProjectRollingRelease | undefined;
+
+      client.stdout.write(JSON.stringify(deployment));
+      if (deployment.projectId && deployment.projectId != '') {
+        project = await getProjectByNameOrId(client, deployment.projectId);
```
brookemosby, line 182: Verified that projectId is returned in deployment fetch

packages/cli/src/util/deploy/process-deployment.ts
```diff
@@ -239,20 +240,13 @@ export default async function processDeployment({
...
-
-      // Workaround to avoid hammering `api-projects-list`
-      if (rollingRelease === undefined) {
-        const project = await getProjectByNameOrId(
-          client,
-          deployment?.projectId || ''
-        );
-
+      let project: Project | ProjectNotFound | undefined;
+      if (project === undefined) {
```
brookemosby, line 244: only check if project is undefined, since rolling releases won't be enabled for most projects

trek (dismissed):

trek (dismissed):

trek (dismissed):

trek (dismissed):

trek (approved):

PR #13500, opened 2025-06-26
[cli] correct log checking suggestion for failed `deploy` runs

trek:
We shifted `vercel logs` to mean only _runtime_ logs about 11 months ago, but have been advising people to check them for build errors. 🧹 to correct the output.

Also decided to change the suggestion's order to prefer CLI interactions, where before we suggested looking in the UI first.

### DIFF
--- a/packages/cli/src/commands/deploy/index.ts
+++ b/packages/cli/src/commands/deploy/index.ts
@@ -668,11 +668,13 @@ export default async (client: Client): Promise<number> => {
 
     if (err instanceof BuildError) {
       output.error(err.message || 'Build failed');
-      output.error(
-        `Check your logs at https://${now.url}/_logs or run ${getCommandName(
-          `logs ${now.url}`
+      output.log('\n');
+      output.log(
+        `To check build logs run: ${getCommandName(
+          `inspect ${now.url} --logs`
         )}`
       );
+      output.log(`Or inspect them in your browser at https://${now.url}/_logs`);
 
       return 1;
     }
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/deploy/index.ts
```diff
@@ -668,11 +668,13 @@ export default async (client: Client): Promise<number> => {
 
     if (err instanceof BuildError) {
       output.error(err.message || 'Build failed');
-      output.error(
-        `Check your logs at https://${now.url}/_logs or run ${getCommandName(
-          `logs ${now.url}`
+      output.log('\n');
+      output.log(
```
chris-olszewski, line 672: I think the `error` -> `log` makes sense as long as `log` also goes to `stderr`.
(Ignore if `output.error` goes to `stdout`)

chris-olszewski (approved):

packages/cli/src/commands/deploy/index.ts
trek: I believe it does. `output` [here](https://github.com/vercel/vercel/blob/83bccbfac19dbd63ff290ded3a7b96c5a1fad092/packages/cli/src/commands/deploy/index.ts#L69) is the "output manger" [singleton](https://github.com/vercel/vercel/blob/83bccbfac19dbd63ff290ded3a7b96c5a1fad092/packages/cli/src/output-manager.ts#L3), and all the methods just a decoration over `stream.write` in [this class](https://github.com/vercel/vercel/blob/83bccbfac19dbd63ff290ded3a7b96c5a1fad092/packages/cli/src/util/output/create-output.ts#L32).

I have no idea why it's organized like this though

chris-olszewski: TIL

chris-olszewski (approved):

PR #13523, opened 2025-07-03
[cli] Use streams to upload Blob files instead of reading it all in memory

vvo:
Before this commit, we would read the FULL file before sending it to Vercel
Blob. This can make your memory go wild.

Now we will stream the file from disk to Vercel Blob directly.
I also updated the unit tests so they do not rely on mocking the filesystem calls which are an internal implementation, we only need to check input => output. Ideally I could check the output read stream and verify its content but that's not so easy to do. The ultimate play here is to avoid unit tests for these simple cli commands and only have integrations tests, the real ones.


luismeyer (dismissed):

vvo:
* **#13524** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13523** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

luismeyer (approved):

trek (approved):

PR #13524, opened 2025-07-03
[cli] Add support for piping data to `blob put` command via stdin

vvo:
This PR adds support for:

```sh
cat file.txt | vc blob put
```

Just like any Unix command line tool accepting files, it should accept stdin.

### DIFF
--- a/packages/cli/src/commands/blob/put.ts
+++ b/packages/cli/src/commands/blob/put.ts
@@ -33,13 +33,6 @@ export default async function put(
     return 1;
   }
 
-  if (!parsedArgs.args.length) {
-    printError(
-      `Missing required arguments: ${getCommandName('blob put pathToFile')}`
-    );
-    return 1;
-  }
-
   const {
     flags,
     args: [filePath],
@@ -53,7 +46,10 @@ export default async function put(
     '--force': force,
   } = flags;
 
-  telemetryClient.trackCliArgumentPathToFile(filePath);
+  // Only track file path if one was provided
+  if (filePath) {
+    telemetryClient.trackCliArgumentPathToFile(filePath);
+  }
   telemetryClient.trackCliFlagAddRandomSuffix(addRandomSuffix);
   telemetryClient.trackCliOptionPathname(pathnameFlag);
   telemetryClient.trackCliFlagMultipart(multipart);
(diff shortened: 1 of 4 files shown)


vvo:
* **#13524** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13523** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/cli/src/commands/blob/put.ts
```diff
@@ -53,49 +46,69 @@ export default async function put(
...
       return 1;
     }
-  } catch (err) {
-    output.debug(`Error reading file: ${err}`);
 
-    if (isErrnoException(err)) {
-      output.error(`File doesn't exist at '${filePath}'`);
+    // Reading from stdin
+    putBody = process.stdin;
+    pathname = pathnameFlag ?? `${new Date().toISOString()}-from-stdin`;
```
luismeyer: should we default here? Seems weird to get a random file name. I personally would prefer an error so that I can use the command correctly

packages/cli/src/util/telemetry/commands/blob/put.ts
```diff
@@ -15,6 +15,10 @@ export class BlobPutTelemetryClient
     }
   }
 
+  trackCliInputSourceStdin() {
+    this.trackCliFlag('input-source-stdin');
```
luismeyer: mhm, this isn't really a flag is it? Can we track this differently?

packages/cli/src/commands/blob/put.ts
vvo: I wondered the same but then it makes the --pathname in one case required in another case not required. Given this usage is edge case it's probably ok? And we can ensure in the docs we show it with --pathname

packages/cli/src/util/telemetry/commands/blob/put.ts
vvo: To be honest I have 0 idea where this data ends up. I just did a search on trackCli* and it's not obvious which one to use.

I've updated it to:
```
  trackCliInputSourceStdin() {
    this.trackCliArgument({
      arg: 'pathToFile',
      value: '__vercel_stdin__',
    });
  }
```

Thoughts?

packages/cli/src/commands/blob/put.ts
vvo: After a night of sleep your suggestion is the good one I will update this

luismeyer (dismissed):

trek (dismissed):

trek (approved):

PR #13545, opened 2025-07-09
add error filter to logs CLI

ctgowrie:
Add a flag to vc logs `--error` that filters logs to only error log levels


chris-olszewski (approved):
A changeset will be required before this can merge

PR #13548, opened 2025-07-10
create @vercel/oidc and @vercel/oidc-aws-credentials-provider 

alice-wondered:
This PR moves the existing packages for getting vercel OIDC tokens as well as AWS credentials into their own packages. In order to maintain backwards compatibility for the time being, these packages are re-exported as deprecated from `@vercel/functions`


alice-wondered:
Lots of documentation stuff to change, this draft is mostly just focused on the structure being backwards compatible.

Will also put up a draft that introduces a non-backwards-compatible approach

alice-wondered:
I'll have to clean up build files too because of how some of this is built

packages/functions/src/cache/index.ts
```diff
@@ -142,3 +142,5 @@ export const HEADERS_VERCEL_CACHE_STATE = 'x-vercel-cache-state';
 export const HEADERS_VERCEL_REVALIDATE = 'x-vercel-revalidate';
 export const HEADERS_VERCEL_CACHE_TAGS = 'x-vercel-cache-tags';
 export const HEADERS_VERCEL_CACHE_ITEM_NAME = 'x-vercel-cache-item-name';
+
+export type { RuntimeCache } from './types';
```
markandrus: Do we need this change, or can it be reverted?

packages/oidc/README.md
```diff
@@ -0,0 +1,6 @@
+# `@vercel/oidc`
+
+Runtime OIDC helper methods intended to be used with your Vercel Functions
+
+~~ TODO: fill this in with an actual valid link to docs ~~
+Please [follow the documentation](https://vercel.com/docs/functions/vercel-functions-package) for examples and usage.
```
markandrus: Maybe leave blank for the initial release? Not sure 👀

packages/oidc/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    },
+    "./headers": {
+      "import": "./dist/headers.js",
+      "require": "./dist/headers.js"
+    }
```
markandrus: What is headers? Is that used?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -1,5 +1,4 @@
-import { getContext } from '../get-context';
-
+import { getContext } from './get-context';
```
markandrus, line 1: Revert change here?

packages/oidc/src/get-context.ts
markandrus, line 1: Maybe I'm missing something, but are RuntimeCache and its dependent files necessary for the OIDC code to work? For example, `getVercelOidcToken` only accesses `headers`:

```ts
  const token =
    getContext().headers?.['x-vercel-oidc-token'] ??
    process.env.VERCEL_OIDC_TOKEN;
```

And RuntimeCache is only imported to `getContext` so it can be used as the type of `cache`. So I imagine we could do an even simpler version of `getContext`. What do you think about

1. Not exporting `getContext` from `@vercel/oidc` token, just having an unexported version of it inside of `get-vercel-oidc-token.ts`?
2. Leaving RuntimeCache and all its dependent files in `@vercel/functions`?

IMO, there's a version of `@vercel/oidc` that's almost exclusively `get-vercel-oidc-token.ts`, and nothing else. WDYT?

packages/oidc/package.json
alice-wondered: I copied a bunch of stuff from other places as I was moving things around so random things keep cropping up in all of the JSON 😢

packages/oidc/src/get-context.ts
alice-wondered: Yeah I'm good with that! I mostly moved things the way I did to try and change as little as possible but agreed it can be stripped out

packages/oidc/src/get-vercel-oidc-token.ts
alice-wondered: I think this is needed right? it tracked this change as me renaming the file to packages/oidc/src where get-context is at the same level as this file

packages/functions/src/cache/index.ts
alice-wondered: I'm fine with reverting it, I think it's a hold over from when I initially put everything together with runtime-utils as a package

packages/oidc/src/get-vercel-oidc-token.ts
markandrus: ah yeah, i just meant the newline removal. 

but if we end up inlining a minimal `getContext` in this file, maybe we this changes some more

packages/oidc-aws-credentials-provider/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "build": "pnpm run build:code && pnpm run build:docs",
+    "build:code": "node ../../utils/build.mjs",
+    "build:docs": "typedoc && prettier --write docs/**/*.md docs/*.md"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@aws-sdk/credential-provider-web-identity": "3.609.0",
```
markandrus: - Should this remain a peer dependency?
- Should this use a SemVer range? (e.g., `^3.609.0`)

Also interesting (for a future refactor?) [the NPM page of @aws-sdk/credential-provider-web-identity](https://www.npmjs.com/package/@aws-sdk/credential-provider-web-identity) says

> You probably shouldn't, at least directly. Please use [@aws-sdk/credential-providers](https://www.npmjs.com/package/@aws-sdk/credential-providers) instead.

So maybe we eventually swap this dependency.

markandrus (commented):
One comment/question about a dependency, but otherwise LGTM!

packages/oidc-aws-credentials-provider/package.json
alice-wondered: The peer dependency makes sense, had to take some time to read up on why you'd want to use it. Changed. 

Yeah a refactor later if they don't want you to usually use this package makes perfect sense to me

markandrus (dismissed):
Looks good to me, but I've tagged @marcgreenstock in for an extra set of eyes.

shaper (dismissed):

markandrus (dismissed):

smaeda-ks (dismissed):

packages/oidc-aws-credentials-provider/package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@vercel/oidc": "workspace:*"
+  },
+  "peerDependencies": {
+    "@aws-sdk/credential-provider-web-identity": "^3.609.0"
```
marcgreenstock: I think we should make this a dependency rather than a peer dependency.

packages/functions/src/oidc/index.ts
```diff
@@ -1,6 +1,17 @@
...
-export { awsCredentialsProvider } from './aws-credentials-provider';
-export {
-  getVercelOidcToken,
-  getVercelOidcTokenSync,
-} from './get-vercel-oidc-token';
+/**
+ * @deprecated Use @vercel/oidc-aws-credentials-provider instead
+ * AWS authentication for Vercel Functions
+ */
+export type { AwsCredentialsProviderInit } from '@vercel/oidc-aws-credentials-provider';
```
marcgreenstock: I think we should re-implement rather than importing from `@vercel/oidc-aws-credentials-provider`.

Re-implementing will remove the dependency on `@vercel/oidc-aws-credentials-provider` and allow it to retain optional peer dependency `@aws-sdk/credential-provider-web-identity` while allowing `@vercel/oidc-aws-credentials-provider` to have a dependency (not peer) on `@aws-sdk/credential-provider-web-identity`.

The import on `@vercel/oidc` is fine though.

packages/oidc-aws-credentials-provider/src/aws-credentials-provider.ts
```diff
@@ -1,6 +1,6 @@
 import type { AwsCredentialIdentityProvider } from '@smithy/types';
 import type { FromWebTokenInit } from '@aws-sdk/credential-provider-web-identity';
-import { getVercelOidcTokenSync } from './get-vercel-oidc-token';
+import { getVercelOidcTokenSync } from '@vercel/oidc';
```
marcgreenstock, line 5: I suggest we import `fromWebToken` from `@aws-sdk/credential-provider-web-identity` here rather than a dynamic import. 

It was originally a dynamic import only because `@aws-sdk/credential-provider-web-identity` was an optional peer dependency.

packages/functions/src/oidc/index.ts
```diff
@@ -1,6 +1,17 @@
+/**
+ * @deprecated Use @vercel/oidc-aws-credentials-provider instead
+ * AWS authentication for Vercel Functions
+ */
 export type { AwsCredentialsProviderInit } from './aws-credentials-provider';
+
+/**
+ * @deprecated Use @vercel/oidc-aws-credentials-provider instead
```
marcgreenstock, line 8: I'm curious, does this actually add to the JSDoc or replace it? It might be better to add the deprecation warnings on the original exports.

packages/oidc-aws-credentials-provider/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@aws-sdk/credential-provider-web-identity": "^3.609.0",
+    "@vercel/oidc": "workspace:*"
+  },
+  "peerDependencies": {}
```
marcgreenstock:
```suggestion
  }
```

marcgreenstock (dismissed):

packages/functions/src/oidc/index.ts
alice-wondered: My understanding is that it should work. There's a closed issue in the typescript github where they talk about this behavior and it was accepted as a valid issue when it wasn't working this way and correctly updating the warnings, syntax highlighting, etc.

marcgreenstock (dismissed):

packages/oidc-aws-credentials-provider/src/aws-credentials-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @property {Function} [roleAssumerWithWebIdentity] - A function that assumes a role with web identity and returns a promise fulfilled with credentials for the assumed role.
+ * @property {string} [roleSessionName] - An identifier for the assumed role session.
+ * @property {string} [providerId] - The fully qualified host component of the domain name of the identity provider.
+ * @property {Array} [policyArns] - ARNs of the IAM managed policies that you want to use as managed session policies.
+ * @property {string} [policy] - An IAM policy in JSON format that you want to use as an inline session policy.
+ * @property {number} [durationSeconds=3600] - The duration, in seconds, of the role session. Defaults to 3600 seconds.
+ */
+
+export interface AwsCredentialsProviderInit // eslint-disable-line @typescript-eslint/no-empty-interface
+  extends Omit<FromWebTokenInit, 'webIdentityToken'> {}
```
marcgreenstock, line 22:
```suggestion
  extends Omit<FromWebTokenInit, 'webIdentityToken' | 'providerId'> {}
```

packages/oidc-aws-credentials-provider/src/aws-credentials-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * The init object for the `awsCredentialsProvider` function.
+ *
+ * @typedef {Object} AwsCredentialsProviderInit
+ * @property {string} roleArn - ARN of the role that the caller is assuming.
+ * @property {Object} [clientConfig] - Custom STS client configurations overriding the default ones.
+ * @property {Array} [clientPlugins] - Custom STS client middleware plugin to modify the client default behavior.
+ * @property {Function} [roleAssumerWithWebIdentity] - A function that assumes a role with web identity and returns a promise fulfilled with credentials for the assumed role.
+ * @property {string} [roleSessionName] - An identifier for the assumed role session.
+ * @property {string} [providerId] - The fully qualified host component of the domain name of the identity provider.
```
marcgreenstock, line 15: I just realized we shouldn't be exposing this property.

> Do not specify this value for OpenID Connect ID tokens.
> https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-web-identity/src/fromWebToken.ts#L36-L43

```suggestion
```

packages/oidc-aws-credentials-provider/src/aws-credentials-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Obtains the Vercel OIDC token and creates an AWS credential provider function
+ * that gets AWS credentials by calling STS AssumeRoleWithWebIdentity API.
+ *
+ * @param {AwsCredentialsProviderInit} init - The initialization object.
+ * @param {string} init.roleArn - ARN of the role that the caller is assuming.
+ * @param {Object} [init.clientConfig] - Custom STS client configurations overriding the default ones.
+ * @param {Array} [init.clientPlugins] - Custom STS client middleware plugin to modify the client default behavior.
+ * @param {Function} [init.roleAssumerWithWebIdentity] - A function that assumes a role with web identity and returns a promise fulfilled with credentials for the assumed role.
+ * @param {string} [init.roleSessionName] - An identifier for the assumed role session.
+ * @param {string} [init.providerId] - The fully qualified host component of the domain name of the identity provider.
```
marcgreenstock, line 34:
```suggestion
```

packages/oidc-aws-credentials-provider/src/aws-credentials-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+ * import { awsCredentialsProvider } from '@vercel/functions/oidc';
+ *
+ * const s3Client = new s3.S3Client({
+ *   credentials: awsCredentialsProvider({
+ *     roleArn: "arn:aws:iam::1234567890:role/RoleA",
+ *     clientConfig: { region: "us-west-2" },
+ *     clientPlugins: [addFooHeadersPlugin],
+ *     roleAssumerWithWebIdentity: customRoleAssumer,
+ *     roleSessionName: "session_123",
+ *     providerId: "graph.facebook.com",
```
marcgreenstock, line 53:
```suggestion
 *     providerId: "graph.facebook.com",
```

packages/oidc-aws-credentials-provider/docs/README.md
```diff
@@ -0,0 +1,58 @@
...
+import { awsCredentialsProvider } from '@vercel/functions/oidc';
+
+const s3Client = new s3.S3Client({
+  credentials: awsCredentialsProvider({
+    roleArn: 'arn:aws:iam::1234567890:role/RoleA',
+    clientConfig: { region: 'us-west-2' },
+    clientPlugins: [addFooHeadersPlugin],
+    roleAssumerWithWebIdentity: customRoleAssumer,
+    roleSessionName: 'session_123',
+    providerId: 'graph.facebook.com',
```
marcgreenstock, line 35:
```suggestion
```

LukeSheard (dismissed):

LukeSheard (approved):

PR #13554, opened 2025-07-10
[react-router] Apply Vercel Skew Protection cookie at the request handler top-level

TooTallNate:
Re-attempt at what https://github.com/vercel/vercel/pull/13552 was trying to accomplish. That PR didn't work because those `entry.server` files are only used when "Edge" runtime pages are present.

This approach directly in the request handler should handle all cases.


GxDesign:
🙏 our production app is reporting many errors constantly due to skew with RR7, daily deploys. Thanks for working on this!

GxDesign:
Hi @TooTallNate, any update on this? Would be happy to help test it out

TooTallNate:
@GxDesign Apologies, this fell off of my radar. If you'd like to help test, please set environment variable `VERCEL_CLI_VERSION` with a value of `https://vercel-git-fix-react-router-skew-protection-redux.vercel.sh/tarballs/vercel.tgz` and let me know if that fixes things for you.

GxDesign:
I tried adding the environment variable and enabling Skew Protection, but didnt see any cookies I recognize from the skew protection docs, was there additional setup needed?

PR #13558, opened 2025-07-11
[packages/routing-utils] do not throw when convertRewrites fails

marc-vercel:
Do not throw when we fail to convert a rewrite and return a proper error. I am seeing this errors:

```
[error reporter]: Failed to parse rewrite: {"source":"REDACTED","destination":"REDACTED"}
```

This error is thrown inside convertRewrites: https://github.com/vercel/vercel/blob/826539f0236c5532c473e2490da6ea797d363423/packages/routing-utils/src/superstatic.ts#L196

### DIFF
--- a/packages/routing-utils/src/index.ts
+++ b/packages/routing-utils/src/index.ts
@@ -425,7 +425,7 @@ export function getTransformedRoutes(
         routes,
         error: createError(
           code,
-          'Failed to parse rewrite',
+          (err as Error).message, // Error: Failed to parse rewrite: <which rewrite>
           'https://vercel.link/invalid-route-source-pattern',
           'Learn More'
         ),
--- a/packages/routing-utils/src/index.ts
+++ b/packages/routing-utils/src/index.ts
@@ -416,7 +416,23 @@ export function getTransformedRoutes(
         ),
       };
     }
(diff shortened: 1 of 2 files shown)


packages/routing-utils/src/index.ts
```diff
@@ -416,7 +416,23 @@ export function getTransformedRoutes(
         ),
       };
     }
-    const normalized = normalizeRoutes(convertRewrites(rewrites));
+
+    let convertedRewrites: Route[] = [];
+    try {
+      convertedRewrites = convertRewrites(rewrites);
+    } catch (err) {
```
healeycodes, line 423: `err` is unused – is that by mistake? 👀

healeycodes (dismissed):

packages/routing-utils/src/index.ts
marc-vercel: Updated - I did not want to use `err` for the error message, but given that the error does not leak anything internal: it is just "Error: Failed to parse rewrite: <which rewrite>", it is fine. Rewrites are provided by the customer.

healeycodes: SGTM

PR #13571, opened 2025-07-17
fix(react-router): avoid injecting 404 route if additional runtimes are found

chris-olszewski:
Currently if additional functions are declared with other runtimes e.g. Python we inject a `/api(/.*)? -> 404` route to prevent `/api` from returning the directory (https://github.com/vercel/vercel/pull/5064). This is unnecessary for React Router which handles this for us and overfits resulting in declared React Router routes returning a 404.

### DIFF
--- a/.changeset/green-dryers-shave.md
+++ b/.changeset/green-dryers-shave.md
@@ -1,0 +1,5 @@
+---
+'@vercel/frameworks': patch
+---
+
+No longer inject 404 route for react router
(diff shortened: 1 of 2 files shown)


tknickman (approved):

dimitropoulos (approved):
it doesn't look like a lot of code, but we spent a lot of time coming to this conclusion.  I think the fact that the remix side already has it (and the react-router entry has the remix-bundler too) is good evidence that this was simply just missing.

PR #13572, opened 2025-07-17
[next]: Fix case for _next/data for i18n-basepath-middleware

ztanner:
This fixes an issue with generating `_next/data` routes when `basePath` is used with `i18n` and middleware. We were failing to detect if a route is a prerender due to anchoring a regex replace which fails when a `basePath` causes the locale not to be at the start.

### DIFF
--- a/packages/next/test/fixtures/00-i18n-basepath-middleware/middleware.js
+++ b/packages/next/test/fixtures/00-i18n-basepath-middleware/middleware.js
@@ -1,3 +1,5 @@
+import { NextResponse } from 'next/server' 
+
 export function middleware(req) {
   return NextResponse.next();
 }
--- a/packages/next/test/fixtures/00-i18n-basepath-middleware/middleware.js
+++ b/packages/next/test/fixtures/00-i18n-basepath-middleware/middleware.js
@@ -1,0 +1,7 @@
+export function middleware(req) {
+  return NextResponse.next();
+}
+
+export const config = {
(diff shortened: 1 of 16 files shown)


packages/next/test/fixtures/00-i18n-basepath-middleware/middleware.js
```diff
@@ -0,0 +1,7 @@
+export function middleware(req) {
```
ijjk, line 3:
```suggestion
import { NextResponse } from 'next/server' 

export function middleware(req) {
```

packages/next/test/fixtures/00-i18n-basepath-middleware/vercel.json
```diff
@@ -0,0 +1,101 @@
...
+  "builds": [{ "src": "package.json", "use": "@vercel/next" }],
+  "probes": [
+     {
+      "path": "/docs/_next/data/testing-build-id/en/dynamic/first.json?slug=first",
+      "status": 200,
+      "mustContain": "slug\":\"first\"",
+      "mustNotContain": "<html"
+    },
+    {
+      "path": "/docs/_next/data/testing-build-id/en/dynamic/123.json?slug=123",
```
ijjk, line 12: This is the specific failure case as a prerendered route will resolve correctly but a fallback generated route will not

ijjk (approved):

PR #13578, opened 2025-07-19
Add `--json` flag to projects list command

anthonyshew:
Add `--json` flag to `vercel projects ls` command to enable structured JSON output. This allows for easier programmatic consumption of project data, especially when filtering for projects requiring updates.

### DIFF
--- a/packages/cli/src/commands/project/command.ts
+++ b/packages/cli/src/commands/project/command.ts
@@ -1,5 +1,5 @@
 import { packageName } from '../../util/pkg-name';
-import { nextOption, yesOption } from '../../util/arg-common';
+import { nextOption, yesOption, jsonOption } from '../../util/arg-common';
 
 export const addSubcommand = {
   name: 'add',
@@ -51,6 +51,7 @@ export const listSubcommand = {
   arguments: [],
   options: [
     nextOption,
+    jsonOption,
     {
       name: 'update-required',
       description: 'A list of projects affected by an upcoming deprecation',
@@ -63,6 +64,10 @@ export const listSubcommand = {
     {
       name: 'Paginate projects, where `1584722256178` is the time in milliseconds since the UNIX epoch',
       value: `${packageName} project ls --next 1584722256178`,
+    },
+    {
+      name: 'List projects using a deprecated Node.js version in JSON format',
+      value: `${packageName} project ls --update-required --json`,
     },
   ],
 } as const;
--- a/packages/cli/src/commands/project/command.ts
+++ b/packages/cli/src/commands/project/command.ts
@@ -53,7 +53,6 @@ export const listSubcommand = {
     nextOption,
     {
       name: 'json',
-      shorthand: 'j',
       type: Boolean,
       deprecated: false,
       description: 'Output in JSON format',
--- a/packages/cli/src/commands/project/command.ts
+++ b/packages/cli/src/commands/project/command.ts
@@ -53,6 +53,7 @@ export const listSubcommand = {
     nextOption,
     {
       name: 'json',
+      shorthand: null,
       type: Boolean,
       deprecated: false,
       description: 'Output in JSON format',
--- a/packages/cli/src/commands/project/command.ts
+++ b/packages/cli/src/commands/project/command.ts
@@ -51,7 +51,6 @@ export const listSubcommand = {
   arguments: [],
   options: [
     nextOption,
-    jsonOption,
     {
       name: 'json',
       shorthand: 'j',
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/project/command.ts
```diff
@@ -51,6 +51,13 @@ export const listSubcommand = {
   arguments: [],
   options: [
     nextOption,
+    {
+      name: 'json',
+      shorthand: 'j',
```
chris-olszewski: I don't think we need a shorthand for this flag
```suggestion
```

packages/cli/src/commands/project/list.ts
```diff
@@ -13,6 +13,24 @@ import getScope from '../../util/get-scope';
...
+  'Node Version',
+];
+const PAGINATION_FLAGS_TO_EXCLUDE = [
+  '_',
+  '--next',
+  '-N',
+  '-d',
+  '-y',
+  '--json',
+  '-j',
```
chris-olszewski:
```suggestion
```

packages/cli/test/unit/commands/project/list.test.ts
```diff
@@ -87,6 +87,74 @@ describe('list', () => {
...
+      const user = useUser();
+      useTeams('team_dummy');
+      const project = useProject({
+        ...defaultProject,
+      });
+
+      client.setArgv('project', 'ls', '--json');
+      await projects(client);
+
+      const output = client.stderr.getFullOutput();
```
chris-olszewski: We should emit JSON to `stdout` so it could be piped into something like `jq`

packages/cli/test/unit/commands/project/list.test.ts
```diff
@@ -87,6 +87,74 @@ describe('list', () => {
...
+      expect(output).toContain('"projects":');
+      expect(output).toContain('"pagination":');
+      expect(output).toContain('"contextName":');
+      expect(output).toContain('"elapsed":');
+
+      // Parse JSON to validate structure
+      const jsonMatch = output.match(/\{[\s\S]*\}/);
+      expect(jsonMatch).toBeTruthy();
+
+      const parsedOutput = JSON.parse(jsonMatch![0]);
```
chris-olszewski:
```suggestion
      const parsedOutput = JSON.parse(output);
```

packages/cli/test/unit/commands/project/list.test.ts
```diff
@@ -87,6 +87,74 @@ describe('list', () => {
...
+
+      // Should contain JSON output
+      expect(output).toContain('"projects":');
+      expect(output).toContain('"pagination":');
+      expect(output).toContain('"contextName":');
+      expect(output).toContain('"elapsed":');
+
+      // Parse JSON to validate structure
+      const jsonMatch = output.match(/\{[\s\S]*\}/);
+      expect(jsonMatch).toBeTruthy();
```
chris-olszewski: Parsing the output as JSON and asserting the result is enough
```suggestion
```

packages/cli/src/commands/project/list.ts
```diff
@@ -75,6 +83,102 @@ export default async function list(
...
+) {
+  const jsonOutput = {
+    projects: projectList.map(project =>
+      createProjectJson(project, metadata.deprecated)
+    ),
+    pagination: metadata.pagination,
+    contextName: metadata.contextName,
+    elapsed: metadata.elapsed,
+  };
+  output.print(JSON.stringify(jsonOutput, null, 2));
```
chris-olszewski: I'm not sure if we have to make another output manager targeting `stdout` or if `console.log` would be fine, but pretty adamant that we want this data going to `stdout` instead of `stderr`

chris-olszewski (changes requested):
I feel pretty strongly that we should output JSON to `stdout`

packages/cli/test/unit/commands/__snapshots__/help.test.ts.snap
```diff
@@ -3768,6 +3768,7 @@ exports[`help command > project help output snapshots > project list help output
 
   Options:
 
+  -j,  --json             Output in JSON format
```
anthonyshew:
```suggestion                                                                            
```

packages/cli/src/commands/project/list.ts
anthonyshew: That was what I originally had wanted to do, and on my first parse of the repo, I thought I saw the other `--json` flag(s) were going to stderr. I just looked again and that was totally wrong. 😂

Adjusting now...

chris-olszewski (approved):

PR #13594, opened 2025-07-22
Add hono to framework-detection and associated builder

jeffsee55:
This adds framework detection and an associated builder for Hono so that it can be used in the root as opposed to needing to be in the `api` folder. The framework detection is pretty strict at the moment to ensure we don't get any false positives for non-hono apps.

- Checks that hono is a dependency
- Checks that there is a file (either `index.{js,ts}` or `src/index.{js,ts}` that has some sort of hono import/require

The builder is a light wrapper around the node builder that uses a shim file to tweak the handler functions. This allows users to use the normal `default export app` from the Hono [docs](https://hono.dev/docs/)

This also doesn't affect code inside the `api` folder. If you have a hono app there, you'll still need to user the `hono/vercel` handler. TBD on what we should do about that.

More tests to follow, but much of our test infra is based on the builder being available on the build container, so chicken & egg situation


packages/frameworks/src/frameworks.ts
```diff
@@ -2244,6 +2244,68 @@ export const frameworks = [
...
+        handle: 'filesystem',
+      },
+      {
+        src: '/(.*)',
+        dest: '/',
+      },
+    ],
+    detectors: {
+      every: [{ matchPackage: 'hono' }],
+      some: [
```
jeffsee55, line 2267: The file we're looking for needs to have some sort of `import/require` of Hono, this way we don't get false positives if they have a Hono dependency but are only using it in the `api/` folder

packages/hono/src/build.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  export const GET = handle;
+  export const POST = handle;
+  export const PUT = handle;
+  export const DELETE = handle;
+  export const PATCH = handle;
+  export const OPTIONS = handle;
+  export const HEAD = handle;`;
+
+  return nodeBuild({
```
jeffsee55, line 27: This just calls the node builder with additional arguments for `shim` and `useWebApi`.

`useWebApi` allows us to use the fetch-style `(req: Request) => Response` instead of the Express style `(req: IncomingMessage, res: Response) => void`

packages/hono/src/build.ts
```diff
@@ -0,0 +1,49 @@
+import { type BuildV3 } from '@vercel/build-utils';
+// @ts-expect-error - FIXME: hono-framework build is not exported
```
jeffsee55, line 2: It was a bit finicky to touch these types since they're hand-written an an `index.d.ts` file. Will revisit

.changeset/small-baboons-heal.md
```diff
@@ -0,0 +1,9 @@
+---
+"vercel": patch
```
chris-olszewski: This should probably be a minor

packages/frameworks/src/frameworks.ts
```diff
@@ -2244,6 +2244,68 @@ export const frameworks = [
...
+      ],
+    },
+    settings: {
+      installCommand: {
+        placeholder:
+          '`yarn install`, `pnpm install`, `npm install`, or `bun install`',
+      },
+      buildCommand: {
+        placeholder: 'TODO',
+        value: 'TODO',
```
chris-olszewski: Do these need to be filled out before merging?

packages/node/src/build.ts
```diff
@@ -445,6 +450,13 @@ export const build: BuildV3 = async ({
     ];
   }
 
+  if (shim) {
+    preparedFiles['shim.js'] = new FileBlob({
+      data: `import app from "./${handler}";${shim}`,
```
chris-olszewski: Would it make more sense to have `shim` as a function that takes a handler file path and outputs a shim instead of doing this hard coded behavior? It feels weird the `app` import needs to match the `shim` template provided by the caller.

chris-olszewski (dismissed):
Mostly small questions with respect to the new API

onsclom:
Tried to test manually but running into some problems: https://vercel.com/austin-merricks-projects/hono-simple/AQdyiY3mugTnhEDTJhwkBdmdXupm

Used the basic hono app starter and `vercel.json` with
```
{
  "builds": [
    {
      "src": "*",
      "use": "https://vercel-9kjfljheq.vercel.sh/tarballs/%40vercel/hono.tgz"
    }
  ]
}
```
then did `vc -b VERCEL_CLI_VERSION=https://vercel-9kjfljheq.vercel.sh/tarballs/vercel.tgz`, but it fails to deploy a function for the default route:

image

Maybe this process is missing something that will happen when everything is available on the build container?

Approving this PR since Hono projects of this pattern do not currently work, so it should not make things worse.

onsclom (dismissed):

chris-olszewski (dismissed):

onsclom (approved):

PR #13598, opened 2025-07-23
[cli] Send CI metadata with the git meta

chibicode:
Start sending metadata using CI's environment variables if applicable.


packages/cli/src/util/create-git-meta.ts
```diff
@@ -54,6 +54,14 @@ export async function createGitMeta(
   const dirty = dirtyResult.value;
   const commit = commitResult.value;
 
+  let ciMetadata: CiMetadata = {};
+  if (process.env.GITHUB_ACTIONS && process.env.GITHUB_ACTOR) {
```
chibicode, line 58: https://docs.github.com/en/actions/reference/variables-reference

Name | Description
-- | --
`GITHUB_ACTIONS` | Always set to `true` when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.
`GITHUB_ACTOR` | The name of the person or app that initiated the workflow. For example, `octocat`.

packages/cli/src/util/create-git-meta.ts
```diff
@@ -54,6 +54,14 @@ export async function createGitMeta(
...
+    ciMetadata = {
+      ciType: 'github-actions' as const,
+      githubActionsActor: process.env.GITHUB_ACTOR,
```
wentsul, line 61: Not that familiar with this part of the code, but isn't the git metadata usually generic here? Wouldn't we want to extend this support to other git providers? In the current flow, the [parsing/normalization happens here](https://github.com/vercel/api/blob/d0b5f7828a39c3ca60cc8b3796bedd9b71822fd8/services/api-deployments/src/endpoints/create-deployment.ts#L1630-L1633)

wentsul (dismissed):

internals/types/index.d.ts
```diff
@@ -510,7 +510,33 @@ export interface Token {
...
+    }
+  | {
+      ci: true;
+      ciType: 'bitbucket-pipelines';
+      ciGitProviderUsername?: never;
+      ciGitRepoVisibility?: never;
+    }
+  | {
+      ci?: true;
+      ciType?: never;
```
wentsul, line 534: should this be unknown. feels like empty would either break or make the discriminated union hard to use. not positive on that one.

PR #13601, opened 2025-07-23
[next] fix failing `clientSegmentCache` prefetch case

ztanner:
When `clientSegmentCache` is enabled, special routing rules are emitted to the routes-manifest which are then consumed in the builder to craft the segment cache rewrite rules. However, `.prefetch.rsc` is assumed to _always_ be the correct destination for `/_tree` requests, when in actuality those routes only exist when PPR is enabled. This should be adjusted on the Next.js side to ensure we emit the proper route but because there's a fair amount of plumbing already present in the builder based on `isAppPPREnabled`, and because the forked logic is temporary, I've opted to fix it here. 

Failed run before fix: https://github.com/vercel/vercel/actions/runs/16483462499/job/46605839542?pr=13601

Fixes NAR-176


packages/next/test/fixtures/00-app-dir-segment-cache/index.test.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-env jest */
+const path = require('path');
```
ijjk, line 2: copied into https://github.com/vercel/vercel/pull/13602

PR #13602, opened 2025-07-23
[next] Reduce segment routes

ijjk:
Currently when segment routes are enabled with PPR the number of routes added becomes exponential un-necessarily causing users to be more likely to hit the route limit and also just reduces routing performance. 

This optimizes the routes used for segment and RSC prefetch outputs by using more flexible regex handling instead of exact static route matching. This also removes the duplication of rewrites to handle the `x-nextjs-rewritten-path/query` headers.


packages/next/test/integration/integration-2.test.js
```diff
@@ -637,41 +635,12 @@ describe('rewrite headers', () => {
...
       'x-nextjs-rewritten-path': '/$1',
       'x-nextjs-rewritten-query': 'suffix=$1',
     });
   });
-
-  it('should not add rewrite headers when it is excluded with missing', () => {
-    const route = routes.filter(r => r.src?.includes('missing'));
-    expect(route.length).toBe(1);
-    expect(route[0].headers).toBeUndefined();
-  });
```
gnoff, line 651: Not sure what this test was intending to assert but are we certain this assertion is fine to drop?

gnoff (dismissed):

packages/next/test/integration/integration-2.test.js
ijjk: Yea this was from duplicating routes in specific cases but we aren't duplicating them anymore

chris-olszewski (dismissed):

ztanner (dismissed):

ztanner (approved):

PR #13606, opened 2025-07-24
fix(frameworks): update Nuxt

atinux:
- Dev command should be `nuxt dev`
- We use Nuxt and not Nuxt.js
- Also detect `nuxt-nightly`


LukeSheard (approved):

packages/frameworks/src/frameworks.ts
```diff
@@ -1518,16 +1518,16 @@ export const frameworks = [
...
     slug: 'nuxtjs',
     demo: 'https://nuxtjs-template.vercel.app',
     logo: 'https://api-frameworks.vercel.sh/framework-logos/nuxt.svg',
     screenshot:
       'https://assets.vercel.com/image/upload/v1647366075/front/import/nuxtjs.png',
     tagline:
-      'Nuxt.js is the web comprehensive framework that lets you dream big with Vue.js.',
-    description: 'A Nuxt.js app, bootstrapped with create-nuxt-app.',
-    website: 'https://nuxtjs.org',
+      'Nuxt is the web comprehensive framework that lets you dream big with Vue.js.',
```
danielroe:
```suggestion
      'Nuxt is the open source framework that makes full-stack development with Vue.js intuitive.',
```

PR #13608, opened 2025-07-24
Add refresh token behavior to getVercelOidcToken

alice-wondered:
This PR adds all necessary functionality for determining if a deployment OIDC token is expired and then optionally refreshing it via API call. 

Because the projects using a vercel OIDC token are already using vercel this refresh mechanism depends upon the vercel cli in order to refresh the token to a tmp directory file. It's then read into the existing process env variables and passed back seamlessly.


packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -7,6 +9,10 @@ import { getContext } from './get-context';
  * This function is used to retrieve the OIDC token from the request context or the environment variable.
  * It checks for the `x-vercel-oidc-token` header in the request context and falls back to the `VERCEL_OIDC_TOKEN` environment variable if the header is not present.
  *
+ * Unlike the `getVercelOidcTokenSync` function, this function will optionally refresh the token if it is expired.
+ * by pulling the latest token from the Vercel CLI.
```
markandrus: WDYT about documenting refresh behavior on both functions?

**getVercelOidcToken**

```
* In development, this function will refresh the OIDC token if it is expired.
```

**getVercelOidcTokenSync**

```
* In development, this function will not refresh the OIDC token if it is expired.
* If you need this functionality, use {@link getVercelOidcToken} instead.
```

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -21,8 +27,23 @@ import { getContext } from './get-context';
  * });
  * ```
  */
-export async function getVercelOidcToken(): Promise<string> {
-  return getVercelOidcTokenSync();
+export async function getVercelOidcToken(refresh = false): Promise<string> {
```
markandrus: For consumers of this function, I think we'd want the default to be `true`. IIUC, the dream is to just call  `getVercelOidcToken()` and always have a usable OIDC token (so long as you're logged in to development), without having to think about it.

Actually, what do you think about removing the `refresh` parameter entirely?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -21,8 +27,23 @@ import { getContext } from './get-context';
...
+    token = getVercelOidcTokenSync();
+  } catch (error) {
+    if (!refresh) {
+      throw error;
+    }
+  }
+  if (refresh) {
+    if (!token || isExpired(getTokenPayload(token))) {
+      // better to handle the errors here or let them propogate up to the user so they can handle it?
+      await refreshToken();
```
markandrus: I think it probably makes sense to allow an error to bubble up from here.

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+interface VercelTokenResponse {
+  token: string;
+}
+
+async function getVercelOidcToken(
+  authToken: string,
+  projectId: string
+): Promise<VercelTokenResponse | null> {
+  try {
+    const url = `https://api.vercel.com/projects/token?projectId=${projectId}&source=vercel-oidc-refresh`;
```
alice-wondered: I've noticed a ton of APIs have a `/v{number}/` in front of the path. I dug for this a bit but is there one I should be using here?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  return typeof window !== 'undefined';
+}
+
+async function getVercelDataDir(): Promise<string | null> {
+  if (isBrowserEnv()) {
+    return '';
+  }
+  const vercelFolder = 'com.vercel.cli';
+  const os = await import('os');
+  const path = await import('path');
```
markandrus: Parallelize these?
```suggestion
  const [os, path] = await Promise.all([
    import('os'),
    import('path'),
  ]);
```

Also, we have to do this because the build breaks if we put them statically at the top-level? There are no browser shims that get inserted for the browser case?

Could we do a synchronous `require` instead?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  if (process.env.XDG_DATA_HOME) {
+    return path.join(process.env.XDG_DATA_HOME, vercelFolder);
+  }
+  switch (os.platform()) {
+    case 'darwin':
+      return path.join('~/Library/Application Support', vercelFolder);
+    case 'linux':
+      return path.join('~/.local/share', vercelFolder);
+    case 'win32':
+      return path.join('%LOCALAPPDATA%', vercelFolder);
```
markandrus: The `~` and `%LOCALAPPDATA%` get expanded correctly?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+  const vercelFolder = 'com.vercel.cli';
+  const os = await import('os');
+  const path = await import('path');
+  if (process.env.XDG_DATA_HOME) {
+    return path.join(process.env.XDG_DATA_HOME, vercelFolder);
+  }
+  switch (os.platform()) {
+    case 'darwin':
+      return path.join('~/Library/Application Support', vercelFolder);
```
markandrus: Does the space need escaping? I'm just thinking of issues I have in my shell 😅

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+async function getVercelCliToken(): Promise<string | null> {
+  const dataDir = await getVercelDataDir();
+  if (!dataDir) {
+    return null;
+  }
+  const path = await import('path');
+  const fs = await import('fs');
```
markandrus:
```suggestion
  const [path, fs] = await Promise.all([
    import('path'),
    import('fs'),
  ]);
```
Maybe also better to move to top of function declaration, before the `getVercelDataDir()`?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  const tokenPath = path.join(dataDir, 'auth.json');
+  if (!fs.existsSync(tokenPath)) {
+    return null;
+  }
+  const token = fs.readFileSync(tokenPath, 'utf8');
+  return JSON.parse(token).token;
```
markandrus: It's possible that this is expired and that we should also use a refresh token in the same file to get the latest token (and update `auth.json`); however, I don't think it's necessary for now.

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+      headers: {
+        Authorization: `Bearer ${authToken}`,
+      },
+    });
+    if (!res.ok) {
+      throw new Error(`failed to refresh token: ${res.statusText}`);
+    }
+    if (!res.json) {
+      throw new Error('failed to refresh token: no json response');
+    }
```
markandrus: AFAIK the `json()` method is always defined, and this check can be eliminated.

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    const tokenRes = (await res.json()) as VercelTokenResponse;
```
markandrus: WDYT about having a separate function for validating the response. We don't need Zod, we could just do something like

```ts
function assertVercelTokenResponse(input: unknown): asserts input is VercelTokenResponse {
  if (!input || typeof input !== 'object') {
    throw new TypeError('Expected an object');
  }
  if (!('token' in input) || typeof input.token !== 'string') {
    throw new TypeError('Expected a string-valued token property');
  }
  return input;
}
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    return tokenRes;
+  } catch (e) {
+    const message = e instanceof Error ? e.message : e;
```
markandrus: I suppose the API might also return a JSON object representing an error. IIRC, the shape is something like

```json
{
  "code": "not_found",
  "message": "Project not found"
}
```

or something like this. It may make sense to have a function that parses the JSON response, checks if it is a successful response or an error response, and uses that to construct the error.

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    throw new Error(`failed to refresh token: ${message}`);
```
markandrus: You could also do this:
```suggestion
    const cause = originalError;
    throw new Error('failed to refresh token', { cause });
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    let dir = process.cwd();
+    while (dir !== path.dirname(dir)) {
+      const pkgPath = path.join(dir, '.vercel');
+      if (fs.existsSync(pkgPath)) {
+        return dir;
+      }
+      dir = path.dirname(dir);
+    }
+  } catch (e) {
+    throw new Error('token refresh only supported in node server environments');
```
markandrus: Style question: are we typically doing lowercase error messages in `@vercel/` packages, including CLI? I would guess we should do sentence-case and optionally punctuate

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+  try {
+    const fs = await import('fs');
+    const path = await import('path');
+    const pkgPath = path.join(dir, '.vercel', 'project.json');
+    if (!fs.existsSync(pkgPath)) {
+      throw new Error('project.json not found');
+    }
+    const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+    return pkg.projectId;
```
markandrus: Maybe worth asserting this is indeed a string?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  } catch (e) {
+    const message = e instanceof Error ? e.message : e;
+    throw new Error(`unable to find project id: ${message}`);
+  }
+}
+
+async function findRootDir(): Promise<string> {
+  try {
+    const fs = await import('fs');
+    const path = await import('path');
```
markandrus: Parallelize and move to top?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -7,6 +9,10 @@ import { getContext } from './get-context';
  * This function is used to retrieve the OIDC token from the request context or the environment variable.
  * It checks for the `x-vercel-oidc-token` header in the request context and falls back to the `VERCEL_OIDC_TOKEN` environment variable if the header is not present.
  *
+ * Unlike the `getVercelOidcTokenSync` function, this function will optionally refresh the token if it is expired.
+ * by pulling the latest token from the Vercel CLI.
+ *
+ * @param {boolean} refresh - Whether to refresh the token if it is expired or missing.
  * @returns {Promise<string>} A promise that resolves to the OIDC token.
  * @throws {Error} If the `x-vercel-oidc-token` header is missing from the request context and the environment variable `VERCEL_OIDC_TOKEN` is not set.
```
AAorris: Any new error conditions from `refreshToken` to add here?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -21,8 +27,23 @@ import { getContext } from './get-context';
...
+  try {
+    token = getVercelOidcTokenSync();
+  } catch (error) {
+    if (!refresh) {
+      throw error;
+    }
+  }
+  if (refresh) {
+    if (!token || isExpired(getTokenPayload(token))) {
+      // better to handle the errors here or let them propogate up to the user so they can handle it?
```
AAorris: 👍  Since we were throwing an error before I think there is a precedent for letting it propagate up

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  projectId: string
+): Promise<void> {
+  try {
+    const os = await import('os');
+    const path = await import('path');
+    const tmpDir = os.tmpdir();
+    const tokenPath = path.join(
+      tmpDir,
+      `vercel-oidc-token-${projectId}`,
+      'token.json'
```
markandrus: I wonder if just `${projectId}.json` would be sufficient? The project ID is prefixed `prj_`, right? So it's pretty unambiguous. WDYT?

Also, I started thinking some more about the `/tmp` directory… maybe it would be better to not choose new directories to write this information to, and to instead re-use the existing `~/Library/Application Support`, `$XDG_DATA_HOME`, etc., directory, but with `com.vercel.oidc` or something like that? In that case, `~/Library/Application Support/com.vercel.oidc/${projecId}.json` seems pretty clear? WDYT?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+async function loadToken(
+  projectId: string
+): Promise<VercelTokenResponse | null> {
+  try {
+    const os = await import('os');
+    const path = await import('path');
+    const fs = await import('fs');
```
markandrus: If we could somehow replace `await import` with synchronous `require` or module-level `import`, then we could actually use this inside of `getVercelOidcTokenSync`, and try

- the request header,
- the environment variable, or
- the local disk.

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (typeof Uint8Array === undefined || typeof TextDecoder === undefined) {
+      throw new Error(
+        'The browser environment is not supported. Requires Uint8Array and TextDecoder'
+      );
+    }
+
+    const binaryString = window.atob(base64);
+    const toUtf8 = Uint8Array.from(binaryString, c => c.charCodeAt(0));
+    const decoder = new TextDecoder();
+    const payload = decoder.decode(toUtf8);
```
markandrus: I think there may be some alternative Node.js APIs you can use for doing a Base64URL-decoding. E.g.,

```js
function decodeBase64Url(base64url) {
  const base64 = base64url
    .replace(/-/g, '+')
    .replace(/_/g, '/') + '=='.slice(0, (4 - (base64url.length % 4)) % 4))

  return Buffer.from(base64, 'base64')
}
```

Is that an option? Or are we trying to run in other runtimes where these are unavailable?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+    return JSON.parse(payload);
+  }
+
+  return JSON.parse(Buffer.from(base64, 'base64').toString('utf8'));
+}
+
+export function isExpired(token: TokenPayload): boolean {
+  return token.exp * 1000 < Date.now();
+}
```
markandrus: Instead of just checking that the token is currently expired, we could align the behavior here with what the CLI is doing: e.g., proactively refresh an OIDC token which will expire within the next 15 minutes. WDYT?

https://github.com/vercel/vercel/blob/main/packages/cli/src/util/env/refresh-oidc-token.ts#L10-L13

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,226 @@
...
+
+function isBrowserEnv(): boolean {
+  return typeof window !== 'undefined';
+}
+
+async function getVercelDataDir(): Promise<string | null> {
+  if (isBrowserEnv()) {
+    return '';
+  }
```
markandrus: I think this could be removed, because I cannot see where this is called without first passing through such a check?

markandrus (commented):
Is it possible or intended to use `@vercel/oidc` outside of Node.js? It seems we cannot support the browser at all, and if that's true, I thnk we can simplify a lot of the code. For example, get rid of the environment checks, move to static imports, even make some functions synchronous (and available to `getVercelOidcTokenSync`).

packages/oidc/src/get-vercel-oidc-token.ts
alice-wondered: sounds good to me

packages/oidc/src/token.ts
alice-wondered: I thought it was just browser but forgot that edge functions also don't have access to these as far as I understand

packages/oidc/src/token.ts
alice-wondered: great question... I'm really assuming a lot there huh? 😅 I'll look into this

packages/oidc/src/token.ts
alice-wondered: I'm mostly impartial honestly. There's an argument to be made for letting those tokens get wiped when you reboot since they're for development but also the standard is already to store it in XDG_DATA_HOME so I'm up for changing it

packages/oidc/src/token.ts
alice-wondered: The dynamic imports will also guard for edge functions unfortunately

packages/oidc/src/token.ts
alice-wondered: This is all browser code but what you wrote out should be more or less what I did for the node path. I think I'll just remove the browser path though

packages/oidc/src/token.ts
alice-wondered: it's different in literally every programming language I've used, I'll check what's going on in the rest of the project and match it lol

packages/oidc/src/token.ts
alice-wondered: should be good to go on spaces

packages/oidc/src/token.ts
alice-wondered: Changed these to be expanded properly

packages/oidc/src/token.ts
alice-wondered: I went ahead and defined a custom error type but figure the current error flow is probably enough information to figure out if something is wrong on a basic level. If the messages from our error responses start getting more complex than what's covered it would be good to do this I think

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -33,6 +54,8 @@ export async function getVercelOidcToken(): Promise<string> {
  * This function is used to retrieve the OIDC token from the request context or the environment variable.
  * It checks for the `x-vercel-oidc-token` header in the request context and falls back to the `VERCEL_OIDC_TOKEN` environment variable if the header is not present.
  *
+ * This function will not refresh the token if it is expired. For refreshing the token, use the @link{getVercelOidcToken} function.
```
markandrus: Does `@link{foo}` work? I thought it had to be `{@link foo}`.

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -22,7 +26,24 @@ import { getContext } from './get-context';
...
+    err = error;
+  }
+
+  try {
+    if (!token || isExpired(getTokenPayload(token))) {
+      await refreshToken();
+      token = getVercelOidcTokenSync();
+    }
+  } catch (error) {
+    throw new Error(`Failed to refresh OIDC token: ${err}: ${error}`);
```
markandrus: One option here (sorry, I cannot remember if I suggested it before and we resolved it):

```suggestion
    throw new Error('Failed to refresh OIDC token', { cause: error });
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+interface VercelTokenResponse {
+  token: string;
+}
+
+async function getVercelOidcToken(
+  authToken: string,
+  projectId: string
+): Promise<VercelTokenResponse | null> {
+  try {
+    const url = `https://api.vercel.com/v1/projects/${projectId}/token?source=vercel-oidc-refresh`;
```
markandrus: Many of the Vercel APIs require scoping requests with a `?teamId=$TEAM_ID` or `?slug=$TEAM_SLUG` query parameter. Are we doing that on this endpoint? (We can also special case it there…).

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+
+function findProjectId(): string {
+  const dir = findRootDir();
+  if (!dir) {
+    throw new VercelOidcTokenError('Unable to find root directory');
+  }
+  try {
+    const pkgPath = path.join(dir, '.vercel', 'project.json');
+    if (!fs.existsSync(pkgPath)) {
+      throw new VercelOidcTokenError('Project.json not found');
```
markandrus:
```suggestion
      throw new VercelOidcTokenError('project.json not found');
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+  try {
+    const dir = getUserDataDir();
+    if (!dir) {
+      throw new VercelOidcTokenError('Unable to find user data directory');
+    }
+    const tokenPath = path.join(dir, 'com.vercel.token', `${projectId}.json`);
+    const tokenJson = JSON.stringify(token);
+    fs.mkdirSync(path.dirname(tokenPath), { recursive: true });
+    fs.writeFileSync(tokenPath, tokenJson);
+    fs.chmodSync(tokenPath, 0o600);
```
markandrus: What does this mean?

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+  return token.exp * 1000 < Date.now() + timeout;
+}
+
+export async function refreshToken(): Promise<void> {
+  let maybeToken = loadToken(findProjectId());
+  if (!maybeToken || isExpired(getTokenPayload(maybeToken.token))) {
+    const authToken = getVercelCliToken();
+    if (!authToken) {
+      throw new VercelOidcTokenError(
+        'Failed to refresh OIDC token: login to vercel cli'
```
markandrus, line 20:
```suggestion
        'Failed to refresh OIDC token: login to Vercel CLI'
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+      );
+    }
+    const projectId = findProjectId();
+    if (!projectId) {
+      throw new VercelOidcTokenError(
+        'Failed to refresh OIDC token: project id not found'
```
markandrus, line 25:
```suggestion
        'Failed to refresh OIDC token: project ID not found'
```

packages/oidc/src/token.ts
```diff
@@ -0,0 +1,224 @@
...
+  return;
+}
+
+export class VercelOidcTokenError extends Error {
+  cause?: unknown;
+
+  constructor(message: string, cause?: unknown) {
+    super(message);
+    this.name = 'VercelOidcTokenError';
+    this.cause = cause;
```
markandrus: I think you can do this? I.e., we don't have to implement `cause?: unknown` or override `toString` ourselves, do we?
```suggestion
    super(message, { cause });
    this.name = 'VercelOidcTokenError';
```

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -1,4 +1,6 @@
 import { getContext } from './get-context';
+import { getTokenPayload, isExpired, refreshToken } from './token';
```
markandrus: For edge runtime, should we now make this a dynamic import inside of `getVercelOidcToken`? I.e., something like

```ts
export async function getVercelOidcToken(): Promise<string> {
  let token = '';
  let err: any;

  try {
    token = getVercelOidcTokenSync();
  } catch (error) {
    err = error;
  }

  try {
    const { getTokenPayload, isExpired, refreshToken } = await import('./token');
```

markandrus (commented):
Looking good 👍 Any ideas on how to test?

packages/oidc/src/get-vercel-oidc-token.ts
alice-wondered: oh whoops, you did suggest it and I resolved it after introducing the custom error type but didn't track that it was the other file. I'm good with the suggestion!

packages/oidc/src/token.ts
alice-wondered: I don't believe this endpoint is doing that. The way I wrote it up it uses the authenticated user's team/user to check for access on the requested project. Is there more information on why the teamId query param is necessary? That has been confusing me seeing it on so many endpoints and then going somewhat ignored

packages/oidc/src/token.ts
alice-wondered: I can add a comment, it's setting read permissions to user/group only. I think I should actually make it `0o660` though so that write permissions are user/group as well

alice-wondered:
> Looking good 👍 Any ideas on how to test?

@markandrus only way I can think to add tests would be something e2e after the API gets merged but even then you'd need access to a user credential. I could export and test all of the file I/O or just mock the return of fetch as well

packages/oidc/src/get-vercel-oidc-token.ts
alice-wondered: Ahhh ok I thought I had tried it before making the custom error. There's no options bag on the error constructor pre ES2022 and this module uses ES2021. Is this something we would normally just bump?

packages/oidc/src/token.ts
alice-wondered: The same ES2022 issue would apply here

alice-wondered:
added some tests :)

packages/oidc/src/get-vercel-oidc-token.ts
markandrus: Hmm, not sure. Maybe we should bump in a different PR. Until then, maybe we should just reuse VercelOidcTokenError here?

packages/oidc/src/token.ts
markandrus: Ack, then let's leave as-is. Maybe it's best to just do the string prefixing 🤔

packages/oidc/package.json
```diff
@@ -14,7 +14,7 @@
       "require": "./dist/index.js"
     }
   },
-  "version": "2.0.0",
+  "version": "2.0.1",
```
markandrus: Are we supposed to increment this, or is the changeset process supposed to increment this?

```suggestion
  "version": "2.0.0",
```

alice-wondered: Great question, I honestly have no idea. We somehow ended up on v2 so my guess is it's being auto incremented and I should revert this

.changeset/tidy-needles-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/oidc': patch
```
markandrus: I suppose this could be minor instead, since it's technically a new feature?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -22,7 +26,29 @@ import { getContext } from './get-context';
...
+
+  try {
+    token = getVercelOidcTokenSync();
+  } catch (error) {
+    err = error;
+  }
+
+  try {
+    const { getTokenPayload, isExpired } = await import('./token-util');
+    const { refreshToken } = await import('./token');
```
markandrus: Nit: could these be parallelized?
```suggestion
    const [
      { getTokenPayload, isExpired },
      { refreshToken },
    ] = await Promise.all([
      import('./token-util'),
      import('./token'),
    ]);
```

packages/oidc/src/token-util.ts
```diff
@@ -0,0 +1,151 @@
...
+interface VercelTokenResponse {
+  token: string;
+}
+
+export async function getVercelOidcToken(
+  authToken: string,
+  projectId: string
+): Promise<VercelTokenResponse | null> {
+  try {
+    const url = `https://api.vercel.com/v1/projects/${projectId}/token?source=vercel-oidc-refresh`;
```
markandrus: Do we have a solution for the team scope?

packages/oidc/src/token-util.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export function saveToken(token: VercelTokenResponse, projectId: string): void {
+  try {
+    const dir = getUserDataDir();
+    if (!dir) {
+      throw new VercelOidcTokenError('Unable to find user data directory');
+    }
+    const tokenPath = path.join(dir, 'com.vercel.token', `${projectId}.json`);
+    console.log(tokenPath);
```
AAorris:
```suggestion
```

.changeset/tidy-needles-share.md
alice-wondered: when I used the changeset builder I definitely selected minor

packages/oidc/src/token-io.ts
AAorris, line 1: I wonder if there's an existing function we can use, like [getGlobalPathConfig](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/config/global-path.ts#L18)?

packages/oidc/src/token-util.ts
```diff
@@ -0,0 +1,151 @@
...
+    const tokenJson = JSON.stringify(token);
+    fs.mkdirSync(path.dirname(tokenPath), { mode: 0o660, recursive: true }); // read/write perms for owner only
+    fs.writeFileSync(tokenPath, tokenJson);
+    fs.chmodSync(tokenPath, 0o660); // read/write perms for owner only
```
AAorris, line 103: I wonder if we could be even stricter and use 0o600, like [writeToAuthConfigFile](https://github.com/vercel/vercel/blob/main/packages/cli/src/util/config/files.ts#L59)?

packages/oidc/src/token-util.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (typeof prj.projectId !== 'string') {
+      throw new TypeError('Expected a string-valued projectId property');
+    }
+    return prj.projectId;
+  } catch (e) {
+    throw new VercelOidcTokenError(`Unable to find project ID`, e);
+  }
+}
+
+export function saveToken(token: VercelTokenResponse, projectId: string): void {
```
AAorris, line 93: Should we write conditional on the global auth config's [skipWrite](https://github.com/vercel/vercel/blob/main/internals/types/index.d.ts#L33)? 
(Maybe this config option is unrelated)

.changeset/tidy-needles-share.md
alice-wondered: Oh unless you meant major? Which does make sense. I can't say I'm overly opinionated, the actual major feature is technically rather large but also only affects a small part of the exposed use of the function. I could go either way

packages/oidc/src/get-vercel-oidc-token.ts
alice-wondered: updated 👍

AAorris (approved):

.changeset/tidy-needles-share.md
markandrus: I mean, I think this should be `minor`, since it's an API compatible feature addition:
```suggestion
'@vercel/oidc': minor
```

alice-wondered: Oh I'm fully up for changing it, I used the wizard that the GHA points to and just selected my package as a minor change in the CLI and it did all of this.

LukeSheard (approved):

PR #13609, opened 2025-07-24
Python lambda: always use predefined exclusion list

ali-vercel:
Python lambdas have a predefined list of excluded files (e.g. `.git`).
- Current behavior: when `excludeFiles` is specified, this overrides the predefined list. This is confusing: excluding files results in more files being included.
- Proposed behavior: `excludeFiles` is added to the predefined list.

### DIFF
--- a/.changeset/funny-brooms-smoke.md
+++ b/.changeset/funny-brooms-smoke.md
@@ -1,0 +1,15 @@
+---
+'@vercel/python': major
+---
+
+By default, the Python builder excludes certain directories from the zip output.
+In vercel.json it's also possible to specify a custom `excludeFiles` rule.
+Previously `excludeFiles` would replace the default exclusions entirely. Now the
+default exclusions will always apply. The default exclusions consist of:
+
+- .git
+- .vercel
+- .pnpm-store
+- node_modules (also excluded when nested)
+- .next
+- .nuxt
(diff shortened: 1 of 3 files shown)


ali-vercel:
* **#13609** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13609?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chris-olszewski (commented):
Is the current behavior a bug? If not this should be a major bump, not a patch.

Is there any reason why a user might need a file in one of these directories? I agree that this is the behavior users expect, but the proposed behavior would break users who require a file in the builtin excludes.

Deferring to @jeffsee55 / @onsclom who probably have more informed opinions on this.

onsclom:
Found [a bit of documentation for excludeFiles](https://vercel.com/docs/project-configuration#value-definition), but it doesn't go into specifics about this. I think major is a good idea here.

ali-vercel:
Thanks for the input @chris-olszewski @onsclom, I made it a major change and rephrased the changeset.

chris-olszewski (approved):

PR #13054, opened 2025-02-13
fix(release): correct related publish config

tknickman:
Fix the publish config for the new related projects package

PR #13613, opened 2025-07-25
Add access public to hono

jeffsee55:
https://github.com/vercel/vercel/actions/runs/16530986764/job/46756059686

### DIFF
--- a/.changeset/wet-beers-listen.md
+++ b/.changeset/wet-beers-listen.md
@@ -1,0 +1,5 @@
+---
+"@vercel/hono": patch
+---
+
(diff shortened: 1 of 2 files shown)


dimitropoulos (commented):
I'm sorry, but I'm not a good person to review this - I'm not familiar with `hono` and I don't know what this change means for it (i.e. whether this is permitted, per se).

jeffsee55:
@dimitropoulos this is just saying that even though it's a scoped package, it should be public.

Prior art [here](https://github.com/vercel/vercel/pull/13054)

PR #13622, opened 2025-07-28
Hono example

anthonyshew:
Adding a Hono example to our `examples/` directory.


examples/hono/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "hono",
+  "type": "module",
+  "scripts": {
+    "dev": "tsx watch src/index.ts",
```
anthonyshew: Should this be `vc dev`? Maybe it shouldn't because we can't assume the user has `vc` installed?

anthonyshew: The Cloudflare example uses `workers dev`: https://github.com/honojs/starter/blob/d6d385c92cb134f0c92f6f1e1b9b9643138fadbb/templates/cloudflare-workers/package.json#L4

jeffsee55: yeah `vc dev` is the recommendation for now

examples/hono/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "build": "tsc",
+    "start": "node dist/index.js"
```
jeffsee55: Theres not really a start command IMO

onsclom (dismissed):

onsclom (approved):

examples/hono/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "dev": "vc dev",
+    "build": "vc build"
+  },
+  "dependencies": {
+    "hono": "^4.8.9"
+  },
+  "devDependencies": {
+    "@types/node": "^20.11.17",
+    "tsx": "^4.7.1",
```
styfle, line 13: @anthonyshew Why does this have tsx?

I'm not seeing where its used.

anthonyshew: Looks to be a mistake.

styfle: - https://github.com/vercel/vercel/pull/13911

PR #13625, opened 2025-07-28
[next] fix failing clientSegmentCache prefetch case

ijjk:
Continuing https://github.com/vercel/vercel/pull/13601

Validated against E2E deploy tests here as well https://github.com/vercel/next.js/actions/runs/16664251402/job/47169335770

Closes: NAR-176


packages/next/src/utils.ts
```diff
@@ -519,12 +538,13 @@ export async function getDynamicRoutes({
           // if PPR or client segment cache is enabled
           if (isAppPPREnabled || isAppClientSegmentCacheEnabled) {
             routes.push({
-              ...route,
```
ztanner, line 522: we don't need any other properties on the original route def?

ijjk: nope also causes confusion it what we're applying

ztanner (approved):

PR #13632, opened 2025-07-29
[@vercel/functions] Introduce `attachDatabasePool` (experimental)

cramforce:
Context https://www.notion.so/vercel/The-serverless-compute-and-database-connection-myth-23de06b059c4805c9b6af156260fe683

Experimental for now to allow testing with customers and bake it a bit


packages/functions/src/db-connections/index.ts
```diff
@@ -0,0 +1,310 @@
...
+    redisPool.on!('end', () => {
+      if (DEBUG) {
+        console.log('Redis connection ended');
+      }
+      waitUntilIdleTimeout(dbPool);
+    });
+    return;
+  }
+
+  throw new Error('Unsupported database pool type');
```
QuiiBz, line 309: Not sure about throwing an error, can we just log an error to not potentially completely destroy the instance?

QuiiBz (commented):
Only one comment, otherwise looks good. We should also update the docs: https://github.com/vercel/vercel/blob/main/packages/functions/docs/modules/index.md

packages/functions/src/db-connections/index.ts
cramforce: This would break startup both in development and in production (failing all tests, etc.)

I think is is extremely unlikely to ever get into production by mistake

QuiiBz: Hum, fair point. Let's roll

QuiiBz (dismissed):

ofhouse (approved):

PR #13653, opened 2025-08-01
feat(next): improve fallback root params handling with hasFallbackRootParams flag

wyattjoh:
## Summary
This PR improves the handling of fallback root parameters by introducing a `hasFallbackRootParams` flag to simplify route regex generation and enhance prerender manifest processing.

## Changes
- Add `hasFallbackRootParams` flag to `RoutesManifestRoute` type for better type safety
- Simplify route regex generation when `hasFallbackRootParams` is true by using a more direct `.rsc` suffix
- Update prerender manifest handling for both fallback and blocking fallback routes
- Improve conditional logic for fallback root params processing in the prerender route handler

## Test plan
- [ ] Verify existing tests pass
- [ ] Test route generation with and without fallback root params
- [ ] Validate prerender manifest structure includes new fields

Related Pull Request: https://github.com/vercel/next.js/pull/82282

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -224,6 +224,18 @@ type RoutesManifestRoute = {
   regex: string;
   namedRegex?: string;
   routeKeys?: { [named: string]: string };
+
+  /**
+   * If true, this indicates that the route has fallback root params. This is
+   * used to simplify the route regex for matching.
+   */
+  hasFallbackRootParams?: boolean;
+
+  /**
+   * The prefetch segment data routes for this route. This is used to rewrite
+   * the prefetch segment data routes (or the inverse) to the correct
+   * destination.
+   */
   prefetchSegmentDataRoutes?: {
     source: string;
     destination: string;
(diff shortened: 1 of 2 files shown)


packages/next/src/utils.ts
```diff
@@ -540,9 +553,14 @@ export async function getDynamicRoutes({
             routes.push({
               src: route.src.replace(
                 new RegExp(escapeStringRegexp('(?:/)?$')),
-                '(?<rscSuffix>\\.rsc|\\.prefetch\\.rsc|\\.segments/.+\\.segment\\.rsc)(?:/)?$'
+                hasFallbackRootParams
+                  ? '\\.rsc(?:/)?$'
```
ijjk, line 557: Shouldn't this still match `prefetch.rsc`?

wyattjoh: The rule for matching the `/_tree` request is now inserted by the changes in Next.js to route to the correct `.prefetch.rsc` route.

ijjk: Ah I see

ijjk (approved):

PR #13654, opened 2025-08-01
[cli] Add framework detection text styling

onsclom:
### Before

image

### After

image

image

### Potential future work

If there are more places than detection text we would like to use this, `color` and `specialText` could be an optional field we add to frameworks in [`frameworks.ts`](https://github.com/vercel/vercel/blob/1d14ffeb867eccc1279cd7276b786cae00bd94c8/packages/frameworks/src/frameworks.ts).

### DIFF
--- a/packages/cli/src/util/input/edit-project-settings.ts
+++ b/packages/cli/src/util/input/edit-project-settings.ts
@@ -83,11 +83,25 @@ export async function editProjectSettings(
     return settings;
   }
 
+  const styledFramework = (frameworkName: string) => {
+    const frameworkStyle = {
+      text: frameworkName,
+      color: chalk.blue,
+    };
+
+    if (frameworkName === 'Hono') {
+      frameworkStyle.text = '🔥 Hono';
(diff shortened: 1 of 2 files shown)


onsclom:
https://linear.app/vercel/issue/FLOW-4664/make-hono-output-prettier-to-emphasise-detection

chris-olszewski (approved):
🔥

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -83,11 +83,25 @@ export async function editProjectSettings(
     return settings;
   }
 
+  const styledFramework = (frameworkName: string) => {
+    const frameworkStyle = {
+      text: frameworkName,
+      color: chalk.blue,
+    };
+
+    if (frameworkName === 'Hono') {
```
tknickman, line 92: Let's ship it and ITG this when there is more, would be great to keep this in the framework config, or maintain another framework config for style data. But love it

tknickman (approved):

PR #13659, opened 2025-08-02
chore(hono): remove build from example

tknickman:
Remove build, it's not required


LukeSheard (dismissed):

anthonyshew (changes requested):
I think we need to get rid of `dev`, too, and update the README. Take a look at the README's for our [other examples](https://vercel.com/templates/hono) and [the Hono starter](https://github.com/honojs/starter/tree/main/templates/vercel).

tknickman:
@anthonyshew you're right, fixed!

anthonyshew:
@tknickman The README needs updated, too, and do we want to remove `vercel` as a devDep?

anthonyshew (approved):

jeffsee55 (approved):

PR #13660, opened 2025-08-02
refactor: extract AWS dynamic import to async loader with error handling

Schniz:
## Summary

- Move the dynamic import of  into a dedicated async function with try/catch.
- If the import fails, throws a clear error: “package '@aws-sdk/credential-provider-web-identity' not found”.
- **Motivation:** This extraction is done to explicitly hint to bundlers (such as Webpack, Vite, or others) that it is acceptable for this package to be missing at runtime. By isolating the dynamic import and handling the error, we prevent bundlers from treating the package as a hard dependency, which can reduce bundle size and avoid unnecessary errors in environments where AWS support is not needed.

## Test plan

- All existing tests should pass (no breaking changes expected).

🤖 Generated with [opencode](https://opencode.ai)


TooTallNate (dismissed):

pranaygp (commented):
LGTM but idk the codebase enough to 👍

dglsparsons (approved):

PR #13664, opened 2025-08-02
Support fetchable apps out of the box for Node dev server

jeffsee55:
This removes the shim file now that BOA entries with a `.fetch` property are supported. There were some issues with the shim when the compiler compiled to CJS which will be resolved by this.

It also fails the built when TSC finds an error. Right now, this behavior is only turned on for Hono apps so as to not disrupt existing behavior.


packages/hono/test/unit/build.test.ts
```diff
@@ -24,13 +24,21 @@ const config = {
...
 };
-const meta = { skipDownload: true, cliVersion: '44.4.1' };
+const meta = { skipDownload: true };
 
 const createFiles = (workPath: string, fileList: string[]) => {
   const files: Files = {};
+
   for (const path of fileList) {
+    if (!path) {
+      console.log('❌ createFiles - Found undefined/null path!');
```
chris-olszewski, line 34: Do we want to continue along if given a falsey path? Should we fail the test?

packages/node/src/typescript.ts
```diff
@@ -312,7 +312,11 @@ export function register(opts: Options = {}): Register {
 
       const diagnosticList = filterDiagnostics(diagnostics, ignoreDiagnostics);
 
-      reportTSError(diagnosticList, config.options.noEmitOnError);
+      if (process.env.EXPERIMENTAL_NODE_TYPESCRIPT_ERRORS) {
```
chris-olszewski, line 315: Do we have a standard prefix for `vc` env vars to avoid collisions?

packages/hono/src/build.ts
```diff
@@ -1,51 +1,46 @@
...
 // @ts-expect-error - FIXME: hono-framework build is not exported
 import { build as nodeBuild } from '@vercel/node';
+import { sep } from 'path';
 
 export const build: BuildV3 = async args => {
   const entrypoint = findEntrypoint(args.files);
 
+  // Introducing new behavior for the node builder where Typescript errors always
+  // fail the build. Previously, this relied on noEmitOnError being true in the tsconfig.json
+  process.env.EXPERIMENTAL_NODE_TYPESCRIPT_ERRORS = '1';
```
chris-olszewski, line 11: Does the builder process get reused ever? If so we should switch to just setting this for the this specific `nodeBuild`

onsclom (approved):

PR #13690, opened 2025-08-07
refactor: don't decode access token to get team ID

markandrus:
I don't think we need to decode the access token to get the team ID.

### DIFF
--- a/packages/cli/src/commands/login/future.ts
+++ b/packages/cli/src/commands/login/future.ts
@@ -15,7 +15,6 @@ import {
   deviceAccessTokenRequest,
   processTokenResponse,
   isOAuthError,
-  inspectToken,
 } from '../../util/oauth';
 import o from '../../output-manager';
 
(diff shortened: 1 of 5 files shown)


balazsorban44 (approved):

packages/cli/src/commands/login/future.ts
```diff
@@ -129,30 +128,18 @@ export async function login(client: Client): Promise<number> {
...
-
-      if (tokens.refresh_token) {
-        client.updateAuthConfig({ refreshToken: tokens.refresh_token });
-      }
+      client.updateConfig({ currentTeam: undefined });
 
       // If we have a brand new login, update `currentTeam`
       if (isInitialLogin) {
-        await updateCurrentTeamAfterLogin(client, client.config.currentTeam);
+        await updateCurrentTeamAfterLogin(client);
```
chris-olszewski, line 142: Just confirming my understanding:
 - for initial installs we want to possibly set the team id to user's id
 - all other cases team id should be empty

packages/cli/test/unit/commands/login/future.test.ts
```diff
@@ -93,7 +79,7 @@ describe('login --future', () => {
     expect(await exitCodePromise, 'exit code for "login --future"').toBe(0);
     await expect(client.stderr).toOutput('Congratulations!');
 
-    expect(fetch).toHaveBeenCalledTimes(pollCount + 3);
+    expect(fetch).toHaveBeenCalledTimes(pollCount + 4);
```
chris-olszewski, line 82: Just confirming this is expected since we're now making an additional call to fetch user information in `updateCurrentTeamAfterLogin`?

packages/cli/test/unit/commands/login/future.test.ts
```diff
@@ -150,10 +121,6 @@ describe('login --future', () => {
       }).toString()
     );
 
-    const teamAfter = client.config.currentTeam;
-    expect(teamAfter, 'Team id differs from original').not.toBe(teamBefore);
-    expect(teamAfter).toBe(accessTokenPayload.team_id);
```
chris-olszewski, line 155: Would keeping an assertion of `expect(teamAfter).toBeUndefined()` make sense here just to keep tabs on behavior?

chris-olszewski (approved):

packages/cli/src/commands/login/future.ts
markandrus: Yes, this behavior seems fine. We will double-check that it's OK before unflagging `--future`.

packages/cli/test/unit/commands/login/future.test.ts
markandrus: IDK. It actually remains `undefined` in this test, so I removed it. It's possible we haven't mocked out the default team ID up lookup here.

packages/cli/test/unit/commands/login/future.test.ts
markandrus: That's my understanding, yes; although I find this style of test assertion to be a little too coupled to the implementation. That's part of why I removed the subsequent `for (let i = 3; i <= fetch.mock.calls.length; i++) { …` checks.

PR #13693, opened 2025-08-07
node-builder: Always compile `.mts` to ESM

jeffsee55:
When theres' no `tsconfig` & no `"type": "module"` we should still compile to ESM. The previous [PR](https://github.com/vercel/vercel/pull/13658) was just for `.ts` files, which may compile to CJS if there's no `"type": "module"`, but for `.mts` we should never do that.

EDIT: this PR was dramatically simplified to just use nodenext when no `tsconfig.json` is found


packages/hono/test/unit/build.test.ts
```diff
@@ -102,7 +108,7 @@ const fixtures = {
   },
   '10-index-ts-no-tsconfig': {
     handler: ['index.js'],
-    moduleType: 'cjs',
+    moduleType: 'esm',
```
chris-olszewski, line 105: Just confirming that `cjs` behavior was a bug for this test case

chris-olszewski (approved):

packages/hono/test/unit/build.test.ts
jeffsee55: Yes! 👍

packages/node/src/build.ts
```diff
@@ -143,14 +143,50 @@ async function compile(
...
     const relPath = relative(baseDir, path);
     if (!tsCompile) {
+      // Find package.json to determine module type
+      const pathDir = join(workPath, dirname(path));
+      let pkg: { type?: string } = {};
+
+      // Check if we already have the package.json in cache
+      if (pkgCache.has(pathDir)) {
+        pkg = pkgCache.get(pathDir) || {};
+      } else {
```
styfle:
```suggestion
      let pkg = pkgCache.get(pathDir)

      if (!pkg) {
```

You can simplify this quite a bit

packages/node/src/build.ts
```diff
@@ -143,14 +143,50 @@ async function compile(
...
+        try {
+          const pathToPkg = await walkParentDirs({
+            base: workPath,
+            start: pathDir,
+            filename: 'package.json',
+          });
+          if (pathToPkg) {
+            pkg = require_(pathToPkg);
+          } else {
+            pkg = {};
```
styfle, line 170: You could leave pkg undefined instead of assigning here

packages/node/src/build.ts
```diff
@@ -143,14 +143,50 @@ async function compile(
...
+      const pathDir = join(workPath, dirname(path));
+      let pkg: { type?: string } = {};
+
+      // Check if we already have the package.json in cache
+      if (pkgCache.has(pathDir)) {
+        pkg = pkgCache.get(pathDir) || {};
+      } else {
+        // Load package.json and add to cache
+        try {
+          const pathToPkg = await walkParentDirs({
```
styfle: Doesn't this logic already existing somewhere? I know we already support esm so seems like we could reuse that instead of walking the tree again.

styfle: Oh I'm thinking here which already supports the cache

https://github.com/vercel/vercel/blob/02123258d2358d85ef13548e5f38bfd09641b74b/packages/node/src/build.ts#L319

packages/node/src/typescript.ts
```diff
@@ -491,8 +502,13 @@ export function fixConfig(
   }
 
   // If not specified, the default Node.js module is CommonJS.
+  // However, if package.json has "type": "module" or the file is .mts, use ESNext instead.
   if (config.compilerOptions.module === undefined) {
-    config.compilerOptions.module = 'CommonJS';
+    if (pkg?.type === 'module' || isMtsFile) {
```
styfle: This will break if the file is `.cjs` and `type: module` which is allowed by node

packages/node/src/typescript.ts
```diff
@@ -502,8 +522,13 @@ export function fixConfig(
   }
 
   // If not specified, the default Node.js module is CommonJS.
+  // However, if package.json has "type": "module" or the file is .mts, use ESNext instead.
   if (config.compilerOptions.module === undefined) {
-    config.compilerOptions.module = 'CommonJS';
+    if ((pkg?.type === 'module' && moduleTypeExtension !== 'cts') || moduleTypeExtension === 'mts') {
+      config.compilerOptions.module = 'ESNext';
+    } else {
+      config.compilerOptions.module = 'CommonJS';
```
styfle: I wonder, can we solve this problem by setting `nodenext` instead of `esnext`?

Since `nodenext` is supposed to consider the package.json type and file extension.

jeffsee55: ohh, good point! Let me try that

packages/express/test/unit/build.test.ts
```diff
@@ -111,7 +111,7 @@ const fixtures: Record<
   },
   '10-index-ts-no-tsconfig': {
     handler: ['index.js'],
-    moduleType: 'cjs',
```
jeffsee55, line 114: This is what the bug was exposing

packages/node/src/typescript.ts
jeffsee55: This works, the validation around this is mostly sitting in the express/hono builder tests, but will move some over once the dust settle.

It was failing on some tests but it was because for some reason `nodenext` seems to also turn on `strict` mode. I can't find any mention of that in documentation, but I also think it's safe to explicitly set `strict: false` here anyway.

styfle: Thats odd. I wouldn’t expect strict mode to change. Maybe it’s the “use strict” for modules as opposed to TS strict:true.

packages/hono/test/unit/build.test.ts
```diff
@@ -112,10 +112,18 @@ const fixtures = {
     handler: ['index.mjs'],
     moduleType: 'esm',
   },
+  '13-index-mts-no-tsconfig-node-no-module': {
```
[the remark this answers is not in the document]
jeffsee55: Will fix in follow-up

PR #13698, opened 2025-08-08
feat(cli): add unlink command to disconnect projects

tknickman:
Implements a new `vc unlink` command that reverses the `vc link` operation by removing the `.vercel/` directory and its contents (project.json, README.txt).

Features:
- Simple confirmation prompt before unlinking (skippable with --yes flag)
- Proper error handling for directories that aren't linked
- Clean success messaging
- Comprehensive test coverage including unit and integration tests

The implementation was simplified to focus on core functionality - removing the .vercel directory without modifying .gitignore files, as leaving .vercel entries in .gitignore is harmless and avoids unnecessary user prompts.


dimitropoulos (approved):
only think I could think of reading the code is that maybe there are (or will be in the future) other files that will be under .vercel that we'd not like to get rid of for `vc unlink` but I'm not aware of any today

jeffsee55 (commented):
Should this work for projects linked with `vc link --repo`? That populates a `.vercel/repo.json` like this:

```
{
  "orgId": "team_zmt7092n75xNHueuIifr6TJW",
  "remoteName": "origin",
  "projects": [
    {
      "id": "prj_fyjbzG3P2BRi7NAjml467wrJqlIi",
      "name": "turborepo-with-hono-3-web",
      "directory": "apps/web"
    },
    {
      "id": "prj_zeZ5EyIugQVpV1VvuP6plHfRofvq",
      "name": "turborepo-with-hono-3-api",
      "directory": "apps/api"
    }
  ]
}
```

When that's present, this would ask you which project like:

> ? Please select a Project: turborepo-with-hono-3-api

But then we'd wipe out the whole `.vercel` directory, effectively unlinking all projects. 

We also populate a `.vercel` folder in the project's root dir (eg. `apps/api`) when `vc build` is used, so would want to clear that folder out as well.

PR #13704, opened 2025-08-11
[next] skip building pages router entries when not present

huozhi:
This PR is a dependency of the Next.js PR (https://github.com/vercel/next.js/pull/82444). Since we're not output Pages Router entries when only only App Router presents in that PR, we need to skip the Pages internal entries `/_app`, `/_document` and `/_error` building in next builder.

Currently next builder always emit the internal entries. For backword compatiability, we need to always try to emit the entries for the new Next.js with that change, and for previous versions we still emit Pages Router internal entries. 
When we stop emitting the internal entries, `lambdas[page]` will be undefined, then reading `lambdas[page].fsPath` and the rest gona be broken on those internal entries. So we add optional check for accessing state of each page.

So we filterout the non-existed `internalPages` by `lambdaPages` which is the actual output from next.js, hence the rest logic will be safe. It's creating lambdas based on **what nextjs actually emitted**.

### Verification with Next.js

Manually tested with the 2 deployment failing from nextjs side, it works well.

Using a custom build with this change and test on Next.js with test-deploy jobs:
https://github.com/vercel/next.js/actions/runs/16992428463/job/48176499037?pr=82444

### DIFF
--- a/.changeset/poor-badgers-drop.md
+++ b/.changeset/poor-badgers-drop.md
@@ -2,4 +2,4 @@
 '@vercel/next': patch
 ---
 
-Handle building for app router without interal pages entries
+Handle building for app router without internal pages entries
--- a/.changeset/poor-badgers-drop.md
+++ b/.changeset/poor-badgers-drop.md
@@ -1,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+Handle building for app router without interal pages entries
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -1618,13 +1618,11 @@ export const build: BuildV2 = async buildOptions => {
         nonApiPages.push(page);
       }
 
-      if (pages[page]) {
-        compressedPages[page] = (
-          await createPseudoLayer({
-            [page]: pages[page],
-          })
-        ).pseudoLayer[page] as PseudoFile;
-      }
+      compressedPages[page] = (
+        await createPseudoLayer({
+          [page]: pages[page],
+        })
+      ).pseudoLayer[page] as PseudoFile;
     }
     const mergedPageKeys: string[] = [...nonApiPages, ...apiPages];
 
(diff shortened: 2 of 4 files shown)


.changeset/poor-badgers-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+Handle building for app router without interal pages entries
```
ztanner:
```suggestion
Handle building for app router without internal pages entries
```

packages/next/src/index.ts
```diff
@@ -1637,7 +1643,11 @@ export const build: BuildV2 = async buildOptions => {
       const nftCache = Object.create(null);
       const lstatSema = new Sema(25);
       const lstatResults: { [key: string]: ReturnType<typeof lstat> } = {};
-      const pathsToTrace = mergedPageKeys.map(page => pages[page].fsPath);
+      const pathsToTrace = mergedPageKeys
+        .map(page => {
+          return pages[page]?.fsPath;
```
ijjk: Why does `mergedPageKeys` contain non-existent pages, feel like we should address this at the source instead of adding optionals throughout.

packages/next/src/server-build.ts
```diff
@@ -951,17 +955,10 @@ export async function serverBuild({
...
       } else {
+        const lambdaPage = lambdaPages[page];
         fileList = Array.from(
-          parentFilesMap?.get(
-            path.relative(baseDir, lambdaPages[page].fsPath)
-          ) || []
+          parentFilesMap?.get(path.relative(baseDir, lambdaPage.fsPath)) || []
         );
-
-        if (!fileList) {
```
huozhi, line 960: this error will not be thrown, the fileList is array on both if/else branches.
* In if branch - it will be assigned to `[]`
* In else branch - it will be return from `Array.from`

ijjk (approved):

PR #13706, opened 2025-08-11
Hono: fix 404 status for /api routes

jeffsee55:
Only server /api route from hono builder if the node builder isn't a match


packages/hono/src/index.ts
```diff
@@ -7,9 +7,12 @@ import type { ShouldServe, StartDevServer } from '@vercel/build-utils';
...
   const requestPath = opts.requestPath.replace(/\/$/, ''); // sanitize trailing '/'
-  // Don't override API routes, otherwise serve it
   if (requestPath.startsWith('api')) {
-    return false;
+    // Don't override API routes, otherwise serve it
+    if (opts.hasMatched) {
+      return false;
+    }
+    return true;
   }
```
chris-olszewski: Minor suggestion for readability
```
  if (requestPath.startsWith('api' && opts.hasMatched)) {
    // Don't override API routes, otherwise serve it
    return false;
  }
```

chris-olszewski (approved):

PR #13740, opened 2025-08-13
feat(next): add client param parsing support with experimental flag

wyattjoh:
Enables client-side parameter parsing for Next.js applications when using PPR (Partial Prerendering) with client segment caching. This optimization allows segments to be cached more effectively when route parameters don't affect the segment content.

### What
- Add `isAppClientParamParsingEnabled` flag to build configurations and pipeline
- Thread flag through server build and utils for segment prerender logic
- Use blank `allowQuery` for segment prerenders when client parsing is enabled with PPR
- Refactor variable names for clarity (`htmlFsRef` → `htmlFallbackFsRef`, `jsonFsRef` → `dataFallbackFsRef`)
- Remove strict RSC extension validation for enhanced route compatibility

### Why
When client param parsing is enabled alongside PPR and client segment caching, segments don't need to vary based on route parameters since parsing happens client-side. This allows using empty `allowQuery` arrays for better caching efficiency and enables the same prerendered segment to serve all parameter variations.

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -2067,6 +2067,7 @@ export const build: BuildV2 = async buildOptions => {
       bypassToken: prerenderManifest.bypassToken || '',
       isServerMode,
       isAppPPREnabled: false,
+      prerenderManifest,
     }).then(arr =>
       localizeDynamicRoutes(
         arr,
@@ -2097,6 +2098,7 @@ export const build: BuildV2 = async buildOptions => {
         bypassToken: prerenderManifest.bypassToken || '',
         isServerMode,
         isAppPPREnabled: false,
+        prerenderManifest,
       }).then(arr =>
         arr.map(route => {
           route.src = route.src.replace('^', `^${dynamicPrefix}`);
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -1466,6 +1466,13 @@ export const build: BuildV2 = async buildOptions => {
           true
       : false;
 
+    // When this is true, then it means all routes are PPR enabled.
+    const isAppFullPPREnabled = requiredServerFilesManifest
+      ? requiredServerFilesManifest?.config.experimental?.ppr === true ||
+        requiredServerFilesManifest.config.experimental?.cacheComponents ===
+          true
+      : false;
+
     const isAppClientSegmentCacheEnabled = requiredServerFilesManifest
       ? requiredServerFilesManifest.config.experimental?.clientSegmentCache ===
         true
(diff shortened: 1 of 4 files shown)


wyattjoh:
* **#13741** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13740** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/next/src/index.ts
```diff
@@ -1463,11 +1463,24 @@ export const build: BuildV2 = async buildOptions => {
...
     const isAppClientSegmentCacheEnabled = requiredServerFilesManifest
       ? requiredServerFilesManifest.config.experimental?.clientSegmentCache ===
         true
       : false;
 
+    // We read this from the routes manifest instead of the config because we
+    // expect that the flag will be deprecated in the future and the manifest
+    // will be the source of truth.
+    const isAppClientParamParsingEnabled =
+      routesManifest?.rsc?.clientParamParsing ?? false;
```
ijjk, line 1485: If the goal is to remove it seems like it'd make more sense to come from an experimental config which isn't covered by semver unlike the routes manifest which is versioned.

wyattjoh: The experimental flag is going to be removed eventually, but the client must declare support for it, hence why it should be backwards compatible with this flag being omitted and the RSC manifest seems like a reasonable place for the client to announce support of this (to prevent cache poisoning).

packages/next/src/utils.ts
```diff
@@ -553,13 +575,25 @@ export async function getDynamicRoutes({
...
                 new RegExp(escapeStringRegexp('(?:/)?$')),
-                hasFallbackRootParams
+                // If we have fallback root params (implying we've already
+                // emitted a rewrite for the /_tree request), or if the route
+                // has PPR enabled and client param parsing is enabled, then
+                // we don't need to include any other suffixes.
+                hasFallbackRootParams ||
+                  (isRoutePPREnabled &&
+                    isAppClientParamParsingEnabled &&
+                    !prefetchDataRoute)
```
ijjk: Would be good to hoist this out of the ternary since it's used here and the `dest`

packages/next/src/index.ts
ijjk: Ah I see so the experimental flag will go away but we need to keep the behavior

ijjk (approved):

PR #13764, opened 2025-08-14
[next] should not create segment cache for non prerender routes

huozhi:
### Issue

When `clientSegmentCache` is enabled, the build output is not aligned with local next build. Noticed that even pages router API are generated with segment cache routes. 

During local build we don't create that for dynamic ones. In next.js it's only generating for ssg paths.


packages/next/src/server-build.ts
```diff
@@ -1677,7 +1677,10 @@ export async function serverBuild({
       });
       pagesPlaceholderRscEntries[`${pathName}.rsc`] = dummyBlob;
 
-      if (isAppClientSegmentCacheEnabled) {
+      if (
+        isAppClientSegmentCacheEnabled &&
+        prerenderRoutes.has(normalizePage(page))
+      ) {
         pagesPlaceholderRscEntries[`${pathName}.segments/_tree.segment.rsc`] =
           dummyBlob;
```
ijjk, line 1685: The description of the PR says the pages API route was hitting an ISR route although this isn't ISR this is just a static blob. Also it can only be navigated to if the `Next-Router-Segment-Prefetch` header is set.

packages/next/test/fixtures/00-app-dir-segment-cache/vercel.json
```diff
@@ -18,6 +18,15 @@
...
       "mustContain": "foobar-1"
+    },
+    {
+      "path": "/api/status",
+      "status": 200,
+      "responseHeaders": {
+        "x-vercel-cache": "MISS",
+        "cache-control": "public, max-age=0, must-revalidate"
+      },
+      "mustContain": "\"status\":\"ok\""
```
ijjk, line 29: This test case passes without the changes here

CleanShot 2025-08-15 at 11 09 51@2x

huozhi: We reached out to o11y folks today that this is not a problem that causing the behavior difference ([slack-ref](https://vercel.slack.com/archives/C08V7CU57E1/p1755192842231719)), so this PR is not tight to the issue anymore.

The change is now only targetting for improving the confusing output. Will update the tests once we confirm it's not ideal to build out segment routes.

huozhi:
Closing this as we need to keep the routes with dummy blob (literally `{}`) that match at the filesystem routing step to cause hard navigations without a 404

PR #13766, opened 2025-08-14
[next] upgrade ts version in test fixture to workaround e2e failure

huozhi:
Update typescript version in test fixture to workaround the failing ts error caused by latest canary

```
- include was updated to add '.next/types/**/*.ts'

  Linting and checking validity of types  ...
  Failed to compile.
Type error: An import path can only end with a '.tsx' extension when 'allowImportingTsExtensions' is enabled.
```

It errors with tes 5.1 but not error with 5.9. That type route feature is using a TS feature only supports above 5.7 (x-ref: https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#path-rewriting-for-relative-paths)


ijjk (approved):

ijjk:
Closing in favor of fix landed to canary

PR #13771, opened 2025-08-15
feat(cli): add SSO re-authorization support to `--future`

balazsorban44:
[Slack thread](https://vercel.slack.com/archives/C07LMT9LBUZ/p1755249046831269?thread_ts=1755246366.663849&cid=C07LMT9LBUZ)

This PR adds support for re-authorizing via SSO during any subcommand that tries to access an SSO-enforced resource for the new flow that is behind the `--future` flag. This is one of the last remaining tasks unflagging the new flow.

I.e. this:
> You must re-authenticate with SAML to use 'teamSlug' scope.

### DIFF
--- a/internals/types/index.d.ts
+++ b/internals/types/index.d.ts
@@ -49,7 +49,6 @@ interface OAuthAuthConfig extends AuthConfigBase {
   expiresAt?: number;
   refreshToken?: string;
   type: 'oauth';
-  sessionId?: string;
 }
 
 type AuthConfig = LegacyAuthConfig | OAuthAuthConfig;
--- a/internals/types/index.d.ts
+++ b/internals/types/index.d.ts
@@ -49,6 +49,7 @@ interface OAuthAuthConfig extends AuthConfigBase {
   expiresAt?: number;
   refreshToken?: string;
   type: 'oauth';
+  sessionId?: string;
 }
 
 type AuthConfig = LegacyAuthConfig | OAuthAuthConfig;
--- a/packages/cli/src/commands/login/index.ts
+++ b/packages/cli/src/commands/login/index.ts
@@ -73,7 +73,11 @@ export default async function login(client: Client): Promise<number> {
     if (validateEmail(input)) {
       result = await doEmailLogin(client, input);
     } else {
-      result = await doSamlLogin(client, input, parsedArgs.flags['--oob']);
+      result = await doSamlLogin(
+        client,
+        { teamIdOrSlug: input, slug: input },
+        parsedArgs.flags['--oob']
+      );
     }
   } else if (parsedArgs.flags['--github']) {
     result = await doGithubLogin(client, parsedArgs.flags['--oob']);
--- a/packages/cli/src/commands/login/index.ts
+++ b/packages/cli/src/commands/login/index.ts
@@ -73,11 +73,7 @@ export default async function login(client: Client): Promise<number> {
     if (validateEmail(input)) {
       result = await doEmailLogin(client, input);
     } else {
-      result = await doSamlLogin(
-        client,
-        { teamIdOrSlug: input, slug: input },
-        parsedArgs.flags['--oob']
-      );
+      result = await doSamlLogin(client, input, parsedArgs.flags['--oob']);
     }
   } else if (parsedArgs.flags['--github']) {
     result = await doGithubLogin(client, parsedArgs.flags['--oob']);
--- a/packages/cli/src/util/login/prompt.ts
+++ b/packages/cli/src/util/login/prompt.ts
@@ -53,7 +53,12 @@ export default async function prompt(
   } else if (choice === 'saml') {
     const slug =
       error?.teamId || (await readInput(client, 'Enter your Team slug:'));
-    result = await doSamlLogin(client, slug, outOfBand, ssoUserId);
+    result = await doSamlLogin(
+      client,
+      { teamIdOrSlug: slug, slug },
+      outOfBand,
+      ssoUserId
+    );
   }
 
   return result;
--- a/packages/cli/src/util/login/prompt.ts
+++ b/packages/cli/src/util/login/prompt.ts
@@ -13,7 +13,7 @@ import output from '../../output-manager';
 
 export default async function prompt(
   client: Client,
-  error?: Pick<SAMLError, 'teamId'>,
+  error?: Pick<SAMLError, 'teamId' | 'scope'>,
   outOfBand?: boolean,
   ssoUserId?: string
 ) {
@@ -52,13 +52,10 @@ export default async function prompt(
     result = await doEmailLogin(client, email, ssoUserId);
   } else if (choice === 'saml') {
     const slug =
-      error?.teamId || (await readInput(client, 'Enter your Team slug:'));
-    result = await doSamlLogin(
-      client,
-      { teamIdOrSlug: slug, slug },
-      outOfBand,
-      ssoUserId
-    );
+      error?.scope ||
+      error?.teamId ||
+      (await readInput(client, 'Enter your Team slug:'));
+    result = await doSamlLogin(client, slug, outOfBand, ssoUserId);
   }
 
   return result;
--- a/packages/cli/src/util/login/saml.ts
+++ b/packages/cli/src/util/login/saml.ts
@@ -1,14 +1,55 @@
 import { URL } from 'url';
 import type Client from '../client';
+import { inspectTokenRequest, processInspectTokenResponse } from '../oauth';
 import doOauthLogin from './oauth';
 
-export default function doSamlLogin(
+export default async function doSamlLogin(
   client: Client,
-  teamIdOrSlug: string,
+  team: { teamIdOrSlug: string; slug: string },
   outOfBand?: boolean,
   ssoUserId?: string
 ) {
+  if (client.authConfig.type === 'oauth') {
+    const sessionId = await getSessionId(client);
+    const url = new URL(
+      `https://vercel.com/sso/${team.slug}?${new URLSearchParams({
+        session_id: sessionId,
+      }).toString()}`
+    );
+    return doOauthLogin(
+      client,
+      url,
+      'SAML Single Sign-On',
+      outOfBand,
+      ssoUserId
+    );
+  }
   const url = new URL('/auth/sso', client.apiUrl);
-  url.searchParams.set('teamId', teamIdOrSlug);
+  url.searchParams.set('teamId', team.teamIdOrSlug);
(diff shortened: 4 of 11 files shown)


internals/types/index.d.ts
```diff
@@ -49,6 +49,7 @@ interface OAuthAuthConfig extends AuthConfigBase {
   expiresAt?: number;
   refreshToken?: string;
   type: 'oauth';
+  sessionId?: string;
```
markandrus: Nix this?

packages/cli/src/commands/login/index.ts
```diff
@@ -73,7 +73,11 @@ export default async function login(client: Client): Promise<number> {
     if (validateEmail(input)) {
       result = await doEmailLogin(client, input);
     } else {
-      result = await doSamlLogin(client, input, parsedArgs.flags['--oob']);
+      result = await doSamlLogin(
+        client,
+        { teamIdOrSlug: input, slug: input },
+        parsedArgs.flags['--oob']
+      );
```
markandrus: Do we actually need to update this?

packages/cli/src/util/login/prompt.ts
```diff
@@ -53,7 +53,12 @@ export default async function prompt(
   } else if (choice === 'saml') {
     const slug =
       error?.teamId || (await readInput(client, 'Enter your Team slug:'));
-    result = await doSamlLogin(client, slug, outOfBand, ssoUserId);
+    result = await doSamlLogin(
+      client,
+      { teamIdOrSlug: slug, slug },
+      outOfBand,
+      ssoUserId
+    );
```
markandrus: Do we actually need to update this?

packages/cli/src/util/oauth.ts
```diff
@@ -398,3 +401,56 @@ function canParseURL(url: string) {
...
+   * since January 1 1970 UTC, indicating when this token will expire.
+   */
+  exp: number;
+  /** Whether or not the presented token is active. */
+  active: boolean;
+  token_type: 'access_token';
+  /** The authorizing principal's team. */
+  team_id?: string;
+  session_id: string;
+}
```
markandrus:
```suggestion
export interface InrospectionResponse {
  /** Whether or not the presented token is active. */
  active: boolean;

  client_id?: string;
  
  session_id?: string;
}
```

packages/cli/src/util/oauth.ts
```diff
@@ -398,3 +401,56 @@ function canParseURL(url: string) {
...
+ * @see https://datatracker.ietf.org/doc/html/rfc7662#section-2.2
+ */
+export async function processInspectTokenResponse(
+  response: Response
+): Promise<[InspectionError] | [null, AccessToken]> {
+  try {
+    const token = await response.json();
+    if (!token || typeof token !== 'object' || !('session_id' in token)) {
+      throw new Error('Invalid token response');
+    }
```
markandrus: Actually, it's valid, but `active` may be `false`. Why not do this?

```suggestion
    if (!token || typeof token !== 'object' || !('active' in token)) {
      throw new Error('Invalid token introspection response');
    }
```

packages/cli/src/util/oauth.ts
```diff
@@ -398,3 +401,56 @@ function canParseURL(url: string) {
...
+    return [null, token];
+  } catch (cause) {
+    return [new InspectionError('Could not inspect token.', { cause })];
```
markandrus: Nit: use "introspect" everywhere?

```suggestion
    return [new IntrospectionError('Could not introspect token.', { cause })];
```

packages/cli/src/util/oauth.ts
```diff
@@ -398,3 +401,56 @@ function canParseURL(url: string) {
...
+  }
+}
+
+class InspectionError extends Error {}
```
markandrus:
```suggestion
class IntrospectionError extends Error {
  constructor(message, options) {
    super(message, options);
    this.name = 'IntrospectionError';
  }
}
```

packages/cli/src/util/oauth.ts
```diff
@@ -398,3 +401,56 @@ function canParseURL(url: string) {
     return false;
   }
 }
+
+/**
+ * Perform Token Introspection Request.
+ *
+ * @see https://datatracker.ietf.org/doc/html/rfc7662#section-2.1
+ */
+export async function inspectTokenRequest(token: string): Promise<Response> {
```
markandrus, line 410: Why have both `inspectTokenRequest` and `processInspectTokenResponse`?

Can we just have one function, `introspectToken`? It would make life easier for the caller, IMO:

```ts
introspectToken(token: string): Promise<[TokenIntrospectionError] | [null, IntrospectionResponse]>
```

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
...
+      ssoUserId
+    );
+  }
   const url = new URL('/auth/sso', client.apiUrl);
-  url.searchParams.set('teamId', teamIdOrSlug);
+  url.searchParams.set('teamId', team.teamIdOrSlug);
   return doOauthLogin(client, url, 'SAML Single Sign-On', outOfBand, ssoUserId);
 }
+
+export async function getSessionId(client: Client): Promise<string> {
```
markandrus: Nit: return both session ID and client ID here?
```suggestion
export async function getSessionAndClientId(client: Client): Promise<{ session_id: string, client_id: string }> {
```

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
...
+  if (client.authConfig.type !== 'oauth') {
+    throw new Error(
+      'Session ID can only be retrieved for OAuth login. This error should not happen. Please contact support.'
+    );
+  }
```
markandrus: Technically we could just pass it through and see what the endpoint returns 🤷 WDYT about removing the check, then?

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
...
+
+  const inspectResponse = await inspectTokenRequest(token);
+
+  const [inspectError, inspectResult] =
+    await processInspectTokenResponse(inspectResponse);
+
+  if (inspectError) throw inspectError;
+
+  return inspectResult.session_id;
```
markandrus: Something like

```ts
if (!response.session_id || !response.client_id) {
  throw new Error(':-(');
}

return {
  session_id: response.session_id,
  client_id: response.client_id,
};
```

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
...
-  teamIdOrSlug: string,
+  team: { teamIdOrSlug: string; slug: string },
   outOfBand?: boolean,
   ssoUserId?: string
 ) {
+  if (client.authConfig.type === 'oauth') {
+    const sessionId = await getSessionId(client);
+    const url = new URL(
+      `https://vercel.com/sso/${team.slug}?${new URLSearchParams({
+        session_id: sessionId,
```
markandrus:
```suggestion
        session_id: session_id,
        client_id: client_id,
```

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
...
-export default function doSamlLogin(
+export default async function doSamlLogin(
   client: Client,
-  teamIdOrSlug: string,
+  team: { teamIdOrSlug: string; slug: string },
   outOfBand?: boolean,
   ssoUserId?: string
 ) {
+  if (client.authConfig.type === 'oauth') {
+    const sessionId = await getSessionId(client);
```
markandrus:
```suggestion
    const { session_id, client_id } = await getSessionId(client);
```

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,55 @@
 import { URL } from 'url';
 import type Client from '../client';
+import { inspectTokenRequest, processInspectTokenResponse } from '../oauth';
 import doOauthLogin from './oauth';
 
-export default function doSamlLogin(
+export default async function doSamlLogin(
   client: Client,
-  teamIdOrSlug: string,
+  team: { teamIdOrSlug: string; slug: string },
```
markandrus: Should this just stay as-is, for simplicity?

```ts
  teamIdOrSlug: string,
```

and then just do

```ts
      return doSamlLogin(client, error.scope);
```

packages/cli/src/util/oauth.ts
balazsorban44: Just a pattern following the rest of the file. Each OAuth action is separated into a "make request" and "process response" function.

I can combine them in a separate PR if you like. :+1:

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,59 @@
...
+      outOfBand,
+      ssoUserId
+    );
+  }
   const url = new URL('/auth/sso', client.apiUrl);
   url.searchParams.set('teamId', teamIdOrSlug);
   return doOauthLogin(client, url, 'SAML Single Sign-On', outOfBand, ssoUserId);
 }
+
+export async function decodeToken(client: Client) {
```
markandrus:
```suggestion
async function decodeToken(client: Client) {
```

markandrus (approved):
+1 pending tests

LukeSheard (approved):

ali-vercel (approved):

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,59 @@
...
   url.searchParams.set('teamId', teamIdOrSlug);
   return doOauthLogin(client, url, 'SAML Single Sign-On', outOfBand, ssoUserId);
 }
+
+async function decodeToken(client: Client) {
+  const { token } = client.authConfig;
+
+  if (!token) {
+    throw new Error(
+      'You need to be logged in to perform this action. Please log in and try again.'
```
swarnava: Assuming we are referring to `vc login` here? Can we clarify something like `Run vercel login and try again` or similar

packages/cli/src/util/login/saml.ts
```diff
@@ -1,14 +1,59 @@
...
+    await processInspectTokenResponse(inspectResponse);
+
+  if (inspectError) throw inspectError;
+
+  if (
+    !inspectResult.active ||
+    !inspectResult.session_id ||
+    !inspectResult.client_id
+  ) {
+    throw new Error('Invalid token. Please log in and try again.');
```
swarnava: Assuming we are referring to `vc login` here? Can we clarify something like `Run vercel login and try again` or similar

PR #13775, opened 2025-08-16
[next] Add action metadata routes

ijjk:
This leverages the additional fields in the server reference manifest to add request meta for the requested action. 

x-ref: https://github.com/vercel/next.js/pull/82695


packages/next/test/integration/integration-2.test.js
```diff
@@ -714,3 +714,29 @@ describe('cache-control', () => {
...
+  it('should add action name meta routes', async () => {
+    const foundActionNames = [];
+
+    for (const route of buildResult.routes || []) {
+      if (route.has?.[0].key == 'next-action' && route.transforms) {
+        foundActionNames.push(route.transforms[0].args);
+      }
+    }
+    expect(foundActionNames.length).toBe(5);
+    expect(foundActionNames.sort()).toMatchSnapshot();
```
huozhi, line 740: Can we use toMatchInlineSnapshot for easier view in the test? I guess we don't need to create a separate snapshot

huozhi (approved):

packages/next/test/integration/integration-2.test.js
ijjk: Inline doesn't work in this repo due to jest version missing prettier v3 support

ali-vercel (approved):

PR #13776, opened 2025-08-16
[@vercel/functions] Strengthen check for attachDatabasePool to only be active in a function

cramforce:
While the current code did exclude local dev, it still ran during builds.

### DIFF
--- a/packages/functions/src/db-connections/index.ts
+++ b/packages/functions/src/db-connections/index.ts
@@ -158,7 +158,11 @@ const bootTime = Date.now();
 const maximumDuration = 15 * 60 * 1000 - 1000; // 15 minutes - 1 second
 
 function waitUntilIdleTimeout(dbPool: DbPool) {
-  if (!process.env.VERCEL_URL || !process.env.VERCEL_REGION) {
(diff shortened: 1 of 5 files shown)


ali-vercel (approved):

dimitropoulos (approved):

packages/functions/src/db-connections/index.ts
```diff
@@ -158,8 +158,8 @@ const bootTime = Date.now();
 const maximumDuration = 15 * 60 * 1000 - 1000; // 15 minutes - 1 second
 
 function waitUntilIdleTimeout(dbPool: DbPool) {
-  if (!process.env.VERCEL_URL) {
-    // We're not running in Vercel, so we don't need to wait for idle connections.
+  if (!process.env.VERCEL_URL || !process.env.VERCEL_REGION) {
```
tknickman: This should have worked since `VERCEL_URL` is available at both build and runtime. 

NOTE: If a project has system environment variables disabled neither of these will be available.

tknickman (approved):

PR #13784, opened 2025-08-18
[oidc] Dual publish `.mjs` files as ESM

TooTallNate:
For the `@vercel/oidc` package, publish both CommonJS and ESM builds of the package. ESM works better with bundlers using tree shaking such as esbuild.


ijjk (approved):

AAorris (approved):

gr2m:
just ran into the same problem: https://github.com/vercel/vercel/pull/14119. I'll update this PR to get it ready for merging, I like this approach better (local build setup instead of updating the shared `build.mjs` script)

gr2m:
not sure about what this error is

image

but otherwise this should be ready

gr2m (approved):

gr2m (changes requested):
my changes won't work, because the `.js` extensions import the `require` code again 😪

PR #13785, opened 2025-08-18
Add express example

jeffsee55:


examples/hono/pnpm-lock.yaml
jeffsee55, line 1: We don't want this present when users do `vc init hono` because it's based on our workspace

anthonyshew (approved):

tknickman (approved):

PR #13795, opened 2025-08-19
[CICD-DX] Use http-proxy-node16 to fix deprecation notice

jeffsee55:


dimitropoulos (approved):
looks like this was fixed here: https://github.com/Jimbly/http-proxy-node16/commit/baf65417daf70216911e6c68a2554a1306b83642

PR #13798, opened 2025-08-20
Add support for app.js as a server entrypoint

jeffsee55:


packages/hono/src/build.ts
```diff
@@ -41,7 +46,13 @@ export const findEntrypoint = (files: Files) => {
   });
 
   if (!entrypoint) {
-    throw new Error('No valid entrypoint found');
+    throw new Error(`No valid entrypoint found. Valid entrypoints are:
+
+- app.{cjs,js,mjs,mts,ts}
+- index.{cjs,js,mjs,mts,ts}
+- server.{cjs,js,mjs,mts,ts}
+- src/index.{cjs,js,mjs,mts,ts}
```
tknickman: Same

packages/hono/src/build.ts
```diff
@@ -18,6 +18,11 @@ export const build: BuildV3 = async args => {
 
 export const findEntrypoint = (files: Files) => {
   const validEntrypoints = [
+    ['app.cjs'],
+    ['app.js'],
+    ['app.mjs'],
+    ['app.mts'],
+    ['app.ts'],
```
tknickman: Not blocking, but we should probably be doing this programmatically

tknickman (approved):

packages/express/src/build.ts
```diff
@@ -41,7 +46,13 @@ export const findEntrypoint = (files: Files) => {
   });
 
   if (!entrypoint) {
-    throw new Error('No valid entrypoint found');
+    throw new Error(`No valid entrypoint found. Valid entrypoints are:
+
+- app.{cjs,js,mjs,mts,ts}
```
chris-olszewski: I realize these are alphabetical, but I'm not sure if that's the easiest way sort them. Up to you if sorting by lang then module is easier:
```suggestion
- app.{js,cjs,mjs,ts,mts}
```

packages/hono/src/build.ts
```diff
@@ -41,7 +46,13 @@ export const findEntrypoint = (files: Files) => {
   });
 
   if (!entrypoint) {
-    throw new Error('No valid entrypoint found');
+    throw new Error(`No valid entrypoint found. Valid entrypoints are:
+
+- app.{cjs,js,mjs,mts,ts}
```
chris-olszewski: Same as above, I slightly prefer `{js,cjs,mjs,ts,mts}`

chris-olszewski (approved):

PR #13799, opened 2025-08-20
feat(next): add clientParamParsingOrigins support for external rewrite headers

wyattjoh:
This change introduces a new experimental configuration option `clientParamParsingOrigins` that allows controlling which external origins can receive rewrite headers during client parameter parsing. Previously, all external rewrites were blocked from receiving these headers for security reasons.

The implementation adds origin extraction logic and regex-based matching against the configured allowed origins list, enabling secure cross-origin rewrite header forwarding for trusted domains while maintaining security for untrusted external destinations.

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -151,6 +151,7 @@ export async function serverBuild({
   isAppFullPPREnabled,
   isAppClientSegmentCacheEnabled,
   isAppClientParamParsingEnabled,
+  clientParamParsingOrigins,
 }: {
   appPathRoutesManifest?: Record<string, string>;
   dynamicPages: string[];
@@ -200,6 +201,7 @@ export async function serverBuild({
   isAppFullPPREnabled: boolean;
   isAppClientSegmentCacheEnabled: boolean;
   isAppClientParamParsingEnabled: boolean;
+  clientParamParsingOrigins: string[] | undefined;
 }): Promise<BuildResult> {
   if (isAppPPREnabled) {
     debug(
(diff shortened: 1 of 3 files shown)


wyattjoh:
* **#13799** <a href="https://app.graphite.dev/github/pr/vercel/vercel/13799?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/vercel/13799?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/next/src/server-build.ts
```diff
@@ -291,6 +293,30 @@ export async function serverBuild({
...
+          // Find the first occurrence of any URL delimiter
+          const delimiters = ['/', '?', '#'];
+          let endIndex = -1;
+
+          for (const delimiter of delimiters) {
+            const index = urlWithoutProtocol.indexOf(delimiter);
+            if (index !== -1) {
+              endIndex = endIndex === -1 ? index : Math.min(endIndex, index);
+            }
+          }
```
ijjk, line 309:
```suggestion
          for (const delimiter of delimiters) {
            const index = urlWithoutProtocol.indexOf(delimiter);
            if (index !== -1) {
              endIndex = index
              break
            }
          }
```
The delimiter ordering matters so seems like it should break on first occurrence instead of using min index and checking all un-necessarily.

ijjk (approved):
Would be good to have a test case in Next.js or here once landed upstream

packages/next/src/server-build.ts
wyattjoh: I thought the same with the code that I updated too, but the following url would break with these changes:

```
https://google.com/?param=thing/other-thing
```

PR #13805, opened 2025-08-21
Add support for output and install project settings

jeffsee55:


packages/express/src/build.ts
```diff
@@ -13,6 +13,9 @@ export const build: BuildV3 = async args => {
     ...args,
     entrypoint,
     considerBuildCommand: true,
+    entrypointCallback: (preparedFiles: Files) => {
+      return findEntrypoint(preparedFiles);
+    },
```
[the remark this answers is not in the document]
jeffsee55: We want to fail if there's no entrypoint

packages/node/src/build.ts
```diff
@@ -78,14 +80,28 @@ async function downloadInstallAndBundle({
...
+  const installCommand = config.projectSettings?.installCommand;
+  if (typeof installCommand === 'string' && considerBuildCommand) {
+    if (installCommand.trim()) {
+      console.log(`Running "install" command: \`${installCommand}\`...`);
+      await execCommand(installCommand, {
+        ...spawnOpts,
+        cwd: entrypointFsDirname,
+      });
+    } else {
+      console.log(`Skipping "install" command...`);
```
chris-olszewski, line 93: Are we advertising `"install": ""` as a way to skip install? If so a small explanation of why might be good.

chris-olszewski (approved):

packages/node/src/build.ts
jeffsee55: I don't know of the use case, just copied from other places. I think the logic here is just to we don't do `exec("")` but logging as a convention for what the others do 🤷

PR #13807, opened 2025-08-22
[cli] add supportsCancellation option to functions

craigandrews:
Cancellation of requests to Fluid / N1 functions running on Lambda requires breaking changes to how the function lifecycle works. For this reason, and to allow rapid but limited rollout of the feature, configuration of which functions be cancelled will exist in `vercel.json` in the `functions` blocks.

This PR adds support for the optional boolean `supportsCancellation` field:

```json
{
  "functions": {
    "api/cancellable.js": {
      "supportsCancellation": true
    }
  }
}
```

This is then written to the `.vc-config.json` file when building.

```json
{
  "handler": "api/cancellable.js",
  "runtime": "nodejs20.x",
  "architecture": "arm64",
  "environment": {},
  "launcherType": "Nodejs",
  "shouldAddHelpers": true,
  "shouldAddSourcemapSupport": true,
  "awsLambdaHandler": "",
  "supportsCancellation": true
}
```


craigandrews:
*question:* is `supportsCancellation` an acceptable field name or should it be shorter, e.g. `cancellable`?

packages/build-utils/src/lambda.ts
```diff
@@ -305,6 +316,13 @@ export class Lambda {
       }
     }
 
+    if (supportsCancellation !== undefined) {
+      assert(
+        typeof supportsCancellation === 'boolean',
+        '"supportsCancellation" is not a boolean'
+      );
+    }
+
```
ecklf, line 325: Might be worth double checking if this is not already covered in our schema validation

https://github.com/vercel/vercel/blob/f5d6674b41467653c820c90ae769643e141014fb/packages/cli/src/util/validate-config.ts#L158-L178

craigandrews: it is, but this is for when the Lambda class is created in code rather than by parsing JSON

craigandrews: all the other fields have similar checks so I'm just following suite

ali-vercel (approved):

PR #13816, opened 2025-08-25
[next] output static metadata as file instead of prerender

huozhi:
Generate static metadata files as static files in output and directly served by proxy, instead of generating as ISR routes.

This PR uses an approach to match the static metadata output path to the original input file by determine if the route is static metadata file and if it's can be located in the input files.

We normalized the input files like what Next.js normalized the pathnames, and used those pathnames as key of original file refs:

- `/blog/[id]/icon.png` -> `/blog/[id]/icon.png`
- `/foo/icon.png` -> `/foo/icon.png`
- `/(group)/foo/icon.png` -> `/foo/icon-<hash>.png`

The pathname normalization and hash generation is mirrored from Next.js.

Then we could find the original metadata file based on the route key, and change the output from prerender to static file.


packages/next/src/metadata.ts
```diff
@@ -0,0 +1,47 @@
...
+  },
+  twitter: {
+    filename: 'twitter-image',
+    extensions: ['jpg', 'jpeg', 'png', 'gif'],
+  },
+};
+
+export function isStaticMetadataRoute(pathname: string) {
+  // The -\w{6} is the suffix that normalized from group routes;
+  const groupSuffix = '(-\\w{6})?';
```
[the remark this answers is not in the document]
huozhi: it's already `"(-\\w{6})?"` which will be used with new RegExp

packages/next/src/utils.ts
```diff
@@ -79,6 +80,59 @@ function isDynamicRoute(route: string): boolean {
...
+  const pathname = route.replace(/\/route$/, '');
+  const isMetadataPattern = isStaticMetadataRoute(pathname);
+
+  if (isMetadataPattern) {
+    // strip the suffix from pathname
+    const normalizedPathname = pathname.replace(/-\w{6}$/, '');
+    // A set of files in relative paths of source files
+    // app/page.tsx app/icon.svg
+    const filesSet = new Set(Object.keys(files));
+    const hasStaticSourceFile = filesSet.has(`app${normalizedPathname}`);
```
ijjk: Why is this check necessary? Also I don't think we should be checking from `files` and instead could check from `lambdaPageKeys` like done elsewhere if this is necessary.

huozhi: the pathname can be `/icon.png` but the actual source file could either be `/icon.png/route.js` or `/icon.png`, we have a test case covered this.

huozhi: `lambdaPageKeys` seems from the output, the above two cases output is the same but input varies. I think we still need to verify the source files

ijjk: Why do we need to assert the source file is static though isn't matching the prerender pathname sufficient?

huozhi: User custom routes `icon.png/route.js`, `robots.txt/route.js` (source files) technically is a normal route not a metadata route, which can have arbitrary response. They share the same output path with the static metadata routes. e.g.

* src `icon.png` -> out `icon.png/route.js`
* src `icon.png/route.js` -> out `icon.png/route.js`

If there's a source file look like metadata route but it's not, generating it as a metadata image file will be incorrect.

ijjk: Ah I see makes sense to guard on that then if we consider that valid currently.

ijjk (approved):

packages/next/src/utils.ts
```diff
@@ -79,6 +80,59 @@ function isDynamicRoute(route: string): boolean {
...
 
+/**
+ * Check if a route is a static metadata route and has corresponding source file
+ */
+function getContentTypeFromFile(fileRef: File): string | undefined {
+  if (!fileRef || !('fsPath' in fileRef)) {
+    return undefined;
+  }
+
+  const ext = path.extname(fileRef.fsPath).slice(1);
```
devjiwonchoi: Nit1: no need slice if match with "."

Nit2: using map instead of switch is faster here as well

devjiwonchoi (approved):

packages/next/src/utils.ts
huozhi: > Nit1: no need slice if match with "."

this is avoid to have `.` for every key which is not necessary, I don't see a strong difference here

PR #13828, opened 2025-08-26
Expand framework detection to src/app and src/server and .cts extensions

jeffsee55:
- Expand framework detection.
- Improve handling when multiple entrypoints are detected.
- Fix an issue with `@types` packages when they're symlinked


packages/node/src/typescript.ts
```diff
@@ -271,6 +271,17 @@ export function register(opts: Options = {}): Register {
       getCompilationSettings: () => config.options,
       getDefaultLibFileName: () => ts.getDefaultLibFilePath(config.options),
       getCustomTransformers: () => transformers,
+      resolveModuleNames: (moduleNames: string[], containingFile: string) => {
```
jeffsee55, line 274: I can't find a good way to write a test for what this fixes. A couple of times we've seen an inconsistent build error when tsconfig has `"strict": true` because the typescript program can't find `@types/express`. 

I was able to reproduce this using `pnpm` but our unit tests in this repo often use npm because the repo itself uses pnpm and the same error doesn't reproduce when used in this workspace.

packages/express/src/build.ts
```diff
@@ -13,25 +16,59 @@ export const build: BuildV3 = async args => {
...
+    ['app'],
+    ['index'],
+    ['server'],
+    ['src', 'app'],
+    ['src', 'index'],
+    ['src', 'server'],
+  ];
 
-  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'mts'];
+  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'mts', 'cts'];
```
chris-olszewski: Just to keep the same ordering as the `js` variants
```suggestion
  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'cts', 'mts'];
```

chris-olszewski (approved):

packages/hono/src/build.ts
```diff
@@ -13,25 +16,53 @@ export const build: BuildV3 = async args => {
...
+    ['app'],
+    ['index'],
+    ['server'],
+    ['src', 'app'],
+    ['src', 'index'],
+    ['src', 'server'],
+  ];
 
-  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'mts'];
+  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'mts', 'cts'];
```
jeffsee55:
```suggestion
  const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'cts', 'mts'];
```

tknickman (approved):

PR #13837, opened 2025-08-28
Add option to disable Runtime Cache warning

plmrry:
This adds an option to set an env var to disable the "Runtime Cache unavailable" message.

`DISABLE_RUNTIME_CACHE_WARNING`

During local dev, this tends to generate a million logs:

Screenshot 2025-08-28 at 12 53 21 AM


plmrry:
Re-opening!

PR #13842, opened 2025-08-29
feat(next): add filterRegexMatches support for query parameter filtering

wyattjoh:
## What?

This PR adds query parameter filtering support in prerender routes through a new `filterRegexMatches` field in the `Chain` interface. The implementation allows filtering of query parameters based on the `allowQuery` configuration.

## Why?

Currently, all query parameters from named regular expression matches are passed to resume lambda calls, which can pollute the regular expressions used for route matching. This change ensures that only allowed query parameters (as defined in the `allowQuery` configuration) are included, providing better control over query parameter handling in PPR (Partial Prerendering) chains.

## How?

- Added `filterRegexMatches: string[]` field to the `Chain` interface in `packages/build-utils/src/types.ts`
- Updated `onPrerenderRoute` function in `packages/next/src/utils.ts` to:
  - Extract allowed query parameters from the `allowQuery` configuration
  - Populate the `filterRegexMatches` array based on this configuration
  - Include the field in all chain objects created for PPR routes
- Added appropriate changeset documenting the minor version bump for both affected packages

The implementation ensures backward compatibility while providing the new filtering capability for improved query parameter management.

### DIFF
--- a/packages/build-utils/src/types.ts
+++ b/packages/build-utils/src/types.ts
@@ -597,6 +597,12 @@ export interface Chain {
    * The headers to send when making the request to append to the response.
    */
   headers: Record<string, string>;
+
+  /**
+   * This is used to filter the query parameters that are added from the named
+   * regular expression matches.
+   */
+  filterRegexMatches: string[];
 }
 
 /**
--- a/.changeset/query-parameter-filtering.md
+++ b/.changeset/query-parameter-filtering.md
@@ -1,0 +1,6 @@
+---
+"@vercel/build-utils": minor
+"@vercel/next": minor
+---
+
+Add filterRegexMatches support for query parameter filtering in prerender routes. This feature allows filtering query parameters from named regular expression matches based on the allowQuery configuration, ensuring only allowed parameters are passed to resume lambda calls.
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2836,6 +2836,14 @@ export const onPrerenderRoute =
         }
       }
 
+      // Using the allowQuery configuration based on the routes manifest keys,
+      // we will use it's values to filter the query parameters that are added
+      // from the named regular expression matches to the resume lambda call.
+      // This ensures that the resume lambda will only receive the query
+      // parameters that are allowed and not other query parameters that might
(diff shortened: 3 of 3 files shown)


huozhi (approved):
looks good, a non blocking question: should we update the `chain`'s type here in `next/src/utils.ts` on line 330, looks like the new fields are missing there

```
/**
   * Configuration related to Partial Prerendering.
   */
  ppr?: {
    /**
     * The chained response for the PPR resume.
     */
    chain: {
      /**
       * The headers that will indicate to Next.js that the request is for a PPR
       * resume.
       */
      headers: Readonly<Record<string, string>>;
    };
  };
  ```

**Comment by wyattjoh** [comment]
We shouldn't as the new field is inferred by the routes manifest, and not globally set like the other RSC fields are.

File: packages/build-utils/src/types.ts [hunk 1]
```diff
@@ -597,6 +597,12 @@ export interface Chain {
    * The headers to send when making the request to append to the response.
    */
   headers: Record<string, string>;
+
+  /**
+   * This is used to filter the query parameters that are added from the named
+   * regular expression matches.
+   */
+  filterRegexMatches: string[];
```
ijjk, line 605: Why is this needed separate of allowQuery?

wyattjoh: `allowQuery` contributes to the cache key on Vercel. In cases where we have a single PPR fallback shell serving many different URL's (think `allowQuery: []`), we still want to filter the query parameters that are passed to the resume lambda. This is really only critical when we're handling a middleware rewrite where the request URL cannot be used to infer the dynamic path segment values.

ijjk: Shouldn't the resume allow all query values/matches?

wyattjoh: Not all the ones from the regular expression named matches. This is the source of the `?rscstuff=.rsc` query parameter that was being added by some regular expressions.

marc-vercel (approved):
I do not have full context on this PR, but code-wise LGTM. Let's also wait for @ijjk approval, as he had some open questions?

Also, thanks for the clear PR description and comments 🙏!

PR #13855, opened 2025-09-02
chore(cli): telemetry on login state

balazsorban44:
This lets us monitor the login process and how successful it is, in terms of finished attempts.


packages/cli/src/commands/login/index.ts
```diff
@@ -50,7 +51,15 @@ export default async function login(client: Client): Promise<number> {
 
   if (parsedArgs.flags['--future']) {
     telemetry.trackCliFlagFuture('login');
-    return await future(client);
+    const status = await future(client);
+    telemetry.trackState(status);
+    switch (status) {
+      case 'canceled':
```
balazsorban44, line 59: Intentionally return 0 on canceled, to avoid printing:
> " ELIFECYCLE  Command failed."

Same as [before](https://github.com/vercel/vercel/pull/13855/files#diff-7f30718abce504732ae57656f2280ae0acf29cbdf09f917f94f07a016754d4efL50-L51)

packages/cli/src/commands/login/future.ts
balazsorban44, line 1: The cancellation refactor lets us make this function always return a state, instead of interrupting with `process.exit()`

packages/cli/src/commands/login/index.ts
marc-vercel, line 59: Nit: can we add a `// no break` comment, so that in future code changes/reviews we understand we did not forget the break?

marc-vercel (approved):
LGTM - Just left a small comment.

krsjenswbp (commented):
ok

PR #13865, opened 2025-09-02
Undo change that set min node verion to node 20 since Vercel CLI still needs to work on Node 18 for upgrade paths

jeffsee55:
There are still scenarios where the CLI needs to work with Node 18, it's rare but should technically still work to assist in migrating to 20.


tknickman (approved):

marc-vercel (approved):
LGTM, thanks! Could you provide a bit of contexts in the PR description

PR #13884, opened 2025-09-05
fix(python): build installs dependencies into a vendor dir instead of project root

ricardo-agz:
This PR makes it so that `vercel build` on a python project runs pip install into a vendor dir inside /.vercel/cache instead of to the project root.

**Now (after vercel build)**
Screenshot 2025-09-05 at 9 34 15 AM

**Prev (after vercel build)**
Screenshot 2025-09-04 at 3 56 37 PM


TooTallNate:
I suggest installing into somewhere under `.vercel/cache` (which is already cached in the build container for subsequent deployments, and `.vercel` is already added to `.gitignore`).

packages/python/src/index.ts
```diff
@@ -1,5 +1,5 @@
-import fs from 'fs';
-import execa from 'execa';
+import * as fs from 'fs';
+import execa = require('execa');
```
styfle: Why change from import to require?

ricardo-agz: I reverted this since it was unnecessary for this PR. Had only made the change since IDE was getting upset.

packages/python/README.md
styfle, line 1: Note that this readme will get published to npm so not sure that is what we want

https://www.npmjs.com/package/@vercel/python

This feels more like a contributing doc rather than a readme.

ricardo-agz: Good point, removed it for now. Can make a better/proper contributing doc later if necessary

tknickman (approved):

PR #13802, opened 2025-08-21
feat(cli): unflag new `--future` login experience

balazsorban44:
**NOTES**

- I've kept the `future` files (both login, logout, and their respective tests) to reduce the diff on this PR. We can clean up in a follow-up
- `vercel login` will warn on obsolete options (arguments/flags) that we can remove in a minor.
  - There are inline login calls - ie. an auth'ed command in a logged-out state - that trigger login. I ensured that those don't print warnings
- dropped the `--future` flag from `logout` since it was unnecessary in the last 4 months (https://github.com/vercel/vercel/pull/13366) 

---

The `vercel login` command has been overhauled, see below for how to use the new flow.

## Breaking changes

The following are now obsolete:

- Passing the `email` argument or `--github`, `--gitlab`, `--bitbucket` flags. Run the command without any options:

```diff
- $ vercel login your@email.com
- $ vercel login --github
- $ vercel login --gitlab
- $ vercel login --bitbucket
+ $ vercel login
```

Choose your preferred authentication method in the browser instead.

- Passing the `teamId`/`teamSlug` argument. Run the command without any options:

```diff
- $ vercel login team
+ $ vercel login
```

After signing in with any of the non-SAML methods, you will be prompted to authorize individual teams.

- `--oob`: Visit [vercel.com/device](https://vercel.com/device) on any browser-capable device
  and enter the code shown in the terminal. This flag is now obsolete.

> [!TIP]
> Hit <kbd>Enter</kbd> to open the link automatically, if you are signing in from a browser-capable device.

> [!IMPORTANT]
> Read the instructions carefully to match your location, IP, and request time when approving the device, as you will be granting access to your Vercel account.

## Resources

- [`vercel login` Documentation](https://vercel.com/docs/cli/login)
- `vercel login --help`

**Good to know:** The new login experience is built on top of the [OAuth 2.0 Device Flow](https://datatracker.ietf.org/doc/html/rfc8628) specification.

PR #13887, opened 2025-09-05
chore(release): add flow to manually publish canary CLI

balazsorban44:
**Notes:**

- limited to CLI only, to solve the problem I was having, and not need to think about cross-dependency releases
- it creates a canary tag that includes the commit SHA (eg. vercel@46.0.2-canary.40a86f9314769b41ca7a153a415c4a8aa7b21b60)
- [comments back to the PR](https://github.com/vercel/vercel/pull/13802#issuecomment-3257627618)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,6 @@ jobs:
     permissions:
       contents: write
       id-token: write
-      pull-requests: write
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: 'Commit (full) SHA to release the CLI from'
+        required: true
 
 env:
   TURBO_REMOTE_ONLY: 'true'
@@ -14,6 +19,7 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   release:
+    if: github.event_name == 'push'
     name: Release
     runs-on: ubuntu-latest
     permissions:
@@ -76,7 +82,61 @@ jobs:
           script: |
             const script = require('./utils/update-latest-release.js')
             await script({ github, context })
+  manual-cli-release:
+    if: github.event_name == 'workflow_dispatch'
+    name: Manual CLI Release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      pull-requests: write
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          ref: ${{ github.event.inputs.commit }}
 
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: install npm@9
+        run: npm i -g npm@9
+
+      - name: install pnpm@8.3.1
+        run: npm i -g pnpm@8.3.1
+
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build Packages
+        run: pnpm build
+        env:
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+
+      - name: Version CLI, publish and comment on PR
+        run: |
+          cd packages/cli &&
+          PR_NUMBER="$(gh api \
+              -H "Accept: application/vnd.github+json" \
+              "repos/${{ github.repository }}/commits/${{github.event.inputs.commit}}/pulls" \
+              --jq '.[0].number' || true)" &&
+          VERSION=$(node -p "require('./package.json').version")-canary.${{ github.event.inputs.commit }} &&
+          pnpm version $VERSION --no-git-tag-version &&
+          pnpm publish --access public --tag canary --dry-run &&
+          gh pr comment $PR_NUMBER --body "🚀 Published canary version:
+          \`\`\`bash
+          pnpm add vercel@$VERSION
+          \`\`\`
+          "
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          NPM_CONFIG_PROVENANCE: 'true'
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
   summary:
     name: Summary (release)
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,11 @@ on:
       commit:
         description: 'Commit (full) SHA to release the CLI from'
         required: true
+      dryRun:
+        description: 'Whether to run the publish process in dry run mode.'
+        type: boolean
+        required: true
+        default: true
 
 env:
   TURBO_REMOTE_ONLY: 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -76,7 +82,61 @@ jobs:
...
+      - name: Version CLI, publish and comment on PR
+        run: |
+          cd packages/cli &&
+          PR_NUMBER="$(gh api \
+              -H "Accept: application/vnd.github+json" \
+              "repos/${{ github.repository }}/commits/${{github.event.inputs.commit}}/pulls" \
+              --jq '.[0].number' || true)" &&
+          VERSION=$(node -p "require('./package.json').version")-canary.${{ github.event.inputs.commit }} &&
+          pnpm version $VERSION --no-git-tag-version &&
+          pnpm publish --access public --tag canary --dry-run --no-git-checks &&
```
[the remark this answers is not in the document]
balazsorban44: Did this on purpose to test, will drop before the PR is merged.

.github/workflows/release.yml
```diff
@@ -76,7 +82,61 @@ jobs:
...
+          \`\`\`bash
+          pnpm add vercel@$VERSION
+          \`\`\`
+          "
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          NPM_CONFIG_PROVENANCE: 'true'
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN_ELEVATED }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
```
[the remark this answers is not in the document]
balazsorban44: I think this does not matter for the manual release workflow.

.github/workflows/release.yml
```diff
@@ -76,7 +82,61 @@ jobs:
...
+
+      - name: Build Packages
+        run: pnpm build
+        env:
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+
+      - name: Version CLI, publish and comment on PR
+        run: |
+          cd packages/cli &&
+          PR_NUMBER="$(gh api \
```
tknickman: This could be empty if this is run against a branch that hasn't been opened in a PR yet.

.github/workflows/release.yml
tknickman: Let's make `--dry-run` an input, so we can choose to run a release as a dry run (default), or for real.

.github/workflows/release.yml
```diff
@@ -76,7 +82,61 @@ jobs:
...
+              -H "Accept: application/vnd.github+json" \
+              "repos/${{ github.repository }}/commits/${{github.event.inputs.commit}}/pulls" \
+              --jq '.[0].number' || true)" &&
+          VERSION=$(node -p "require('./package.json').version")-canary.${{ github.event.inputs.commit }} &&
+          pnpm version $VERSION --no-git-tag-version &&
+          pnpm publish --access public --tag canary --dry-run --no-git-checks &&
+          gh pr comment $PR_NUMBER --body "🚀 Published canary version:
+          \`\`\`bash
+          pnpm add vercel@$VERSION
```
tknickman: Is this a check just to make sure it's installable? I would move this to a different step so we get better error messaging in the workflow and can tell the difference between publish failed vs install failed.

.github/workflows/release.yml
```diff
@@ -76,7 +82,61 @@ jobs:
...
+        run: |
+          cd packages/cli &&
+          PR_NUMBER="$(gh api \
+              -H "Accept: application/vnd.github+json" \
+              "repos/${{ github.repository }}/commits/${{github.event.inputs.commit}}/pulls" \
+              --jq '.[0].number' || true)" &&
+          VERSION=$(node -p "require('./package.json').version")-canary.${{ github.event.inputs.commit }} &&
+          pnpm version $VERSION --no-git-tag-version &&
+          pnpm publish --access public --tag canary --dry-run --no-git-checks &&
+          gh pr comment $PR_NUMBER --body "🚀 Published canary version:
```
tknickman: We should add a log message here as well just so we can see it in the workflow output (and in the case when we don't have a PR)

.github/workflows/release.yml
```diff
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: 'Commit (full) SHA to release the CLI from'
+        required: true
```
tknickman, line 11: We could let this default to HEAD as well.

tknickman (changes requested):
A few change requests, but looks good thanks for tackling this one!

.github/workflows/release.yml
balazsorban44: I prefer if we can point to any commit and that it is explicitly set, to make it very intentional what you are releasing so I would keep it without a default value.

.github/workflows/release.yml
balazsorban44: I think I will just drop this and comment in the logs.

.github/workflows/release.yml
balazsorban44: I think I will just drop this and comment in the logs.

.github/workflows/release.yml
balazsorban44: I think I will just drop this and comment in the logs.

.github/workflows/release.yml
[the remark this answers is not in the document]
balazsorban44: made it configurable https://github.com/vercel/vercel/pull/13887#discussion_r2325063213

ijjk (commented):
Why is this needed when we already create tarballs under publicly accessibly URL for every commit e.g. for the latest commit on main (not released) https://vercel-c4rpej6u8.vercel.sh/tarballs/vercel.tgz

Which can be used via `npm i -g vercel@https://vercel-c4rpej6u8.vercel.sh/tarballs/vercel.tgz` or `VERCEL_CLI_VERSION=https://vercel-c4rpej6u8.vercel.sh/tarballs/vercel.tgz`

balazsorban44:
Hey @ijjk, thanks for checking! This new approach lets you target _any_ commit, not just the ones in `main`. This comes in handy for cases like #13802. A commit in `main` will always end up in a published release, potentially before it is desired.

> we already create tarballs under publicly accessibly URL for every commit e.g. for the latest commit on main

I just learned about this, is this documented somewhere? In addition to not finding details in the repo, I also asked the CI/CD team and was told that we don't currently have a good pre-release workflow.

We were seemingly also publishing `canary` releases to npm, which seemed to be identical to `latest` releases, so I thought it'd be a good target for pre-releases, and matches other public projects we have (eg. Next.js canaries).

I will let @tknickman and you decide about this PR, but I think the approach makes a lot of sense so we have a way of manually triggering releases like in the afforementioned case, under `vercel@canary` on npm, which makes it easy to test changes, using the npm registry.

We also get [provenance](https://search.sigstore.dev/?logIndex=483594392) and a good overview of pre-release via npm APIs.

Alternatively, we should extend the current pre-release flow to support any commit and document it, however I cannot commit to that work, as I was only trying to unblock myself for #13802 and opened this PR to contribute back.

NOTE: I marked the PR for autogmerge, however if you disagree, feel free to just close it! 🙏

TooTallNate:
> This new approach lets you target _any_ commit, not just the ones in `main`

The tarball URLs are already for every commit, since they are created in the Vercel deployment for the `vercel/vercel` repo. For example, the CLI tarball URL for the latest commit on this very PR is https://vercel-c9apevuh1.vercel.sh/tarballs/vercel.tgz.

There is some documentation around this as well here: https://www.notion.so/vercel/Vercel-CLI-Tarballs-c214e2aea93b41a288087fec290316e9#c214e2aea93b41a288087fec

So ya, unless there's a use-case that publishing to npm allows that these tarballs can not fulfill, then I don't think we need this PR. The tarballs effectively already _are_ canary releases for every commit.

TooTallNate:
For your example PR #13802, the tarball URL for the [latest commit](https://github.com/vercel/vercel/pull/13802/commits/a0bfa7a5ba0f3cab79a4c03891d8bf046563733f) there is: https://vercel-21xtqpa5h.vercel.sh/tarballs/vercel.tgz

So if you wanted to test locally or whatever you can do:

```
npm i -g https://vercel-21xtqpa5h.vercel.sh/tarballs/vercel.tgz
```

balazsorban44:
That makes sense, thanks for the info!

PR #13914, opened 2025-09-11
[python] FastAPI entrypoint discovery

ricardo-agz:
Auto-detect different entrypoints for FastAPI

### DIFF
--- a/packages/python/src/index.ts
+++ b/packages/python/src/index.ts
@@ -33,7 +33,10 @@ const fastapiContentRegex =
   /(from\s+fastapi\s+import\s+FastAPI|import\s+fastapi|FastAPI\s*\()/;
 
 const fastapiCandidateEntrypoints = fastapiEntrypointFilenames.flatMap(
-  filename => fastapiEntrypointDirs.map(dir => `${dir}/${filename}.py`)
+  filename =>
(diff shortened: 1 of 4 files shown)


packages/python/src/index.ts
```diff
@@ -26,6 +27,29 @@ import { getLatestPythonVersion, getSupportedPythonVersion } from './version';
...
 
+const fastapiEntrypointFilenames = ['app', 'index', 'server', 'main'];
+const fastapiEntrypointDirs = ['', 'src', 'app'];
+const fastapiContentRegex =
+  /(from\s+fastapi\s+import\s+FastAPI|import\s+fastapi|FastAPI\s*\()/;
+
+const fastapiCandidateEntrypoints = fastapiEntrypointFilenames.flatMap(
+  filename =>
+    fastapiEntrypointDirs.map(dir =>
+      dir ? `${dir}/${filename}.py` : `${filename}.py`
```
tknickman: Should use path.join for ``${dir}/${filename}.py`` instead?

tknickman (approved):

packages/python/src/index.ts
ricardo-agz: Good callout, you're right, much cleaner that way

PR #13921, opened 2025-09-11
[python]: don't cache python dependencies yet

ricardo-agz:
Need to add cache invalidation on requirements/Pipfile changes before we can cache the dependencies vendor dir. Moving base vendor dir out of .vercel/cache

### DIFF
--- a/.changeset/fluffy-ears-burn.md
+++ b/.changeset/fluffy-ears-burn.md
@@ -1,0 +1,5 @@
+---
+"@vercel/python": patch
+---
+
+[python]: move python packages vendor dir out of /cache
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
> .next/cache

Double checking you mean `.vercel/cache`.

> Need to add cache invalidation on requirements/Pipfile changes

What happens now as a result of this being cached? Build cache issues?

ricardo-agz:
> > .next/cache
> 
> Double checking you mean `.vercel/cache`.
> 
> > Need to add cache invalidation on requirements/Pipfile changes
> 
> What happens now as a result of this being cached? Build cache issues?

Yes, meant `.vercel/cache`. Issue would be that if someone has a deployment and then adds new dependencies, redeploys, then hits the lambda max size issue, then removes the dependencies from requirements.txt to fix it and redeploys, the old dependencies will still be included and the build will still fail.

PR #13924, opened 2025-09-12
Fix _next/data middleware resolving without header

ijjk:
This ensures we route correctly even if the `x-nextjs-data` header isn't sent with middleware `_next/data` request. 

Fixes: NAR-389


packages/next/test/fixtures/00-middleware/vercel.json
```diff
@@ -104,7 +109,7 @@
         "redirect": "manual"
       },
       "responseHeaders": {
-        "Location": "/from-middleware/"
+        "Location": "/from-next-config/"
```
ijjk, line 112: This highlighted the bug without this fix as well.

ztanner (approved):

PR #13942, opened 2025-09-16
[frameworks][h3] Add h3 detection and builder

onsclom:
This adds h3 zero config support following the patterns from Hono and Express. There is currently lots of duplicated code between the three, but we can refactor shared code into build utils in a future PR.


jeffsee55 (commented):
Sorry I forgot to mention we had to hack [here](https://github.com/vercel/vercel/blob/a1c9eb42ebf273d1cfe21d5544315d81a6385883/packages/cli/src/util/dev/builder.ts#L427-L432) to get the dev server to work. I know, I know... messy

jeffsee55 (approved):

packages/frameworks/logos/h3.svg
pi0, line 1: H3 has a new logo: https://github.com/h3js/logos

packages/h3/test/fixtures/01-basic/server.mjs
```diff
@@ -0,0 +1,5 @@
+import { H3, serve } from "h3";
+
+const app = new H3().get("/", () => "⚡️ Tadaa!");
+
+serve(app);
```
pi0, line 5: We can/should do `export default app` (`serve` will use node.js APIs in compat mode)

Template: https://github.com/h3js/starter

(we have one more place to update)

PR #13943, opened 2025-09-16
Support rate-limiting when vercel auth is present and support all frameworks for header discovery

cramforce:
Automatically forwards the auth cookie when present.

Switches from using `next/headers` to using the request context since it is always present


quuu (approved):
clean upgrade

jeffsee55 (approved):

PR #13946, opened 2025-09-16
Add support for inlclude/exclude files config from vercel.json

jeffsee55:
This adds function config for exclude and include files settings, other function configs like region and max duration haven't actually been supported for the node builder so adding that next.
```
// vercel.json
{
  "functions": {
     "server.ts": { "includeFiles": "views/**/*" }
  }
}
```

### DIFF
--- a/packages/fs-detectors/src/detect-builders.ts
+++ b/packages/fs-detectors/src/detect-builders.ts
@@ -721,6 +721,30 @@ function checkUnusedFunctions(
       }
     }
   }
+  if (
+    frontendBuilder &&
+    (isOfficialRuntime('express', frontendBuilder.use) ||
+      isOfficialRuntime('hono', frontendBuilder.use))
+  ) {
+    // Copied from builder entrypoint detection
+    const validFilenames = [
+      'app',
+      'index',
+      'server',
+      'src/app',
+      'src/index',
+      'src/server',
+    ];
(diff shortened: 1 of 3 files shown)


packages/fs-detectors/src/detect-builders.ts
```diff
@@ -721,6 +721,30 @@ function checkUnusedFunctions(
       }
     }
   }
+  if (
```
jeffsee55, line 724: There's downstream logic to throw an error when a vercel.json entry isn't associated with a builder, but we don't want it to throw here, we need this info for when the builder actually runs.

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -721,6 +721,30 @@ function checkUnusedFunctions(
...
+    const validFilenames = [
+      'app',
+      'index',
+      'server',
+      'src/app',
+      'src/index',
+      'src/server',
+    ];
+    const validExtensions = ['js', 'cjs', 'mjs', 'ts', 'cts', 'mts'];
+    const validEntrypoints = validFilenames.flatMap(filename =>
```
jeffsee55, line 739: cc @onsclom this would be handy if it was moved into build utils since its copied from the builders

marc-vercel (approved):
LGTM - Remember to add a PR description before merging 🙏

PR #13953, opened 2025-09-17
[python] switch python build install step to uv

ricardo-agz:
Make python builds faster by using [uv](https://github.com/astral-sh/uv) to install pythons reqs.
For local dev, if uv is available, use it, otherwise, pip install uv, then reqs


packages/python/src/install.ts
```diff
@@ -95,6 +167,95 @@ async function pipInstall(pipPath: string, workPath: string, args: string[]) {
...
+      candidates.push('/usr/local/bin/uv');
+      candidates.push('/opt/homebrew/bin/uv');
+    } else {
+      candidates.push('C:\\Users\\Public\\uv\\uv.exe'); // minimal Windows fallback
+    }
+    for (const p of candidates) {
+      if (fs.existsSync(p)) return p;
+    }
+  } catch {
+    // ignore
```
marc-vercel: Do we want to add a debug here also?

marc-vercel (approved):
LGTM, left a small comment.

jeffsee55 (commented):
Not sure of the best way to test this, will uv be used even when there's only a `requirements.txt`? Does it produce a uv lock when used?

ricardo-agz:
> Not sure of the best way to test this, will uv be used even when there's only a `requirements.txt`? Does it produce a uv lock when used?

@jeffsee55 Yea this PR only switches out the install step to use uv, which is much faster and is compatible with `requirements.txt` via `uv pip install -r requirements.txt`, it does not produce a uv lock. But I do have a follow-up PR to this that does support using uv to read the `pyproject.toml` and `uv.lock` files. 
I also just added some probes to check the logs to make sure it is actually using uv and not falling back to pip

jeffsee55 (approved):

PR #13967, opened 2025-09-19
[python] Support installing dependencies from uv.lock + pyproject.toml

ricardo-agz:
Follow up to https://github.com/vercel/vercel/pull/13953
Adds support for reading requirements from uv.lock and pyproject.toml


packages/python/src/index.ts
```diff
@@ -158,6 +140,17 @@ export const build: BuildV3 = async ({
   const hasReqLocal = !!fsFiles[join(entryDirectory, 'requirements.txt')];
   const hasReqGlobal = !!fsFiles['requirements.txt'];
 
+  const uvLockDir = fsFiles[join(entryDirectory, 'uv.lock')]
```
tknickman: This is more code, but I think it's more readable and reduces the duplication a bit (and gets rid of the nested ternary):

```ts
function findDir({
  file,
  entryDirectory,
  workPath,
  fsFiles,
}: {
  file: string;
  entryDirectory: string;
  workPath: string;
  fsFiles: Record<string, unknown>;
}): string | null {
  if (fsFiles[join(entryDirectory, file)]) {
    return join(workPath, entryDirectory);
  }

  if (fsFiles[file]) {
    return workPath;
  }

  // Case 3: File not found in either location
  return null;
}


const uvLockDir = findDir({
  file: "uv.lock",
  entryDirectory,
  workPath,
  fsFiles,
});

const pyprojectDir = findDir({
  file: "pyproject.toml",
  entryDirectory,
  workPath,
  fsFiles,
});
```

packages/python/src/index.ts
```diff
@@ -169,72 +162,45 @@ export const build: BuildV3 = async ({
...
-        lock = JSON.parse(json);
-      } catch (err) {
-        throw new NowBuildError({
-          code: 'INVALID_PIPFILE_LOCK',
-          message: 'Unable to parse Pipfile.lock',
-        });
-      }
-      pythonVersion = getSupportedPythonVersion({
+    if (requiresPython && /\b\d+\.\d+\b/.test(requiresPython.trim())) {
+      const exact = requiresPython.trim().match(/\b\d+\.\d+\b/)![0];
```
tknickman: Let's just define the regex once here:

Could do something like this:

```ts
const VERSION_REGEX = /\b\d+\.\d+\b/;
const exact = requiresPython?.trim().match(VERSION_REGEX)?.[0];

if (exact) {
  // use exact here
}
```

tknickman (commented):
A few comments for clarity but overall looks good!

packages/python/src/index.ts
ricardo-agz: I agree, this is much cleaner, fixed in new commit

tknickman (approved):

183amir:
Hi this PR introduced an issue for us: https://github.com/vercel/vercel/issues/14044 where this can lead to excessive size due to inclusion of development dependencies, which may quickly exceed the 250MB deployment limit for Python functions.

PR #13985, opened 2025-09-24
feat(cli-auth): introduce `@vercel/cli-auth`

balazsorban44:
https://github.com/vercel/vercel/pull/13986 depends on this


packages/cli-auth/credentials-store.ts
```diff
@@ -0,0 +1,79 @@
...
+// NOTE: We are pinned on v5 because newer versions behave differently
+// when there are dots in the application name, eg. `com.vercel.cli` becomes `com.vercel`
+import XDGAppPaths from 'xdg-app-paths';
+import { z } from 'zod/mini';
+
+const Credentials = z.object({
+  /** An `access_token` obtained using the OAuth Device Authorization flow.  */
+  token: z.optional(z.string()),
+  /** A `refresh_token` obtained using the OAuth Device Authorization flow. */
+  refreshToken: z.optional(z.string().check(z.minLength(1))),
```
markandrus: Why have the minLength check on the `refreshToken` but not the `token`? I think we can be relaxed here and remove the check.
```suggestion
  /** An `access_token` obtained using the OAuth Device Authorization flow.  */
  token: z.optional(z.string()),
  /** A `refresh_token` obtained using the OAuth Device Authorization flow. */
  refreshToken: z.optional(z.string()),
```

markandrus (approved):

packages/cli-auth/oauth.ts
```diff
@@ -0,0 +1,351 @@
...
+        {
+          headers: {
+            'Content-Type': 'application/json',
+            'user-agent': userAgent,
+          },
+        }
+      );
+
+      const json = await response.json();
+      as = AuthorizationServerMetadata.parse(json);
```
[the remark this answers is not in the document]
balazsorban44: We expect never to get HTML from the discovery endpoint

LukeSheard (approved):

packages/cli-auth/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@vercel/cli-auth",
+  "description": "Used by Vercel's CLIs to handle authentication",
+  "version": "0.0.0",
```
gr2m, line 4: was this meant to be public? The released failed due to this new package:
https://github.com/vercel/vercel/actions/runs/18416256421/job/52480516883#step:9:62

PR #13986, opened 2025-09-24
refactor(cli): use `@vercel/cli-auth`

balazsorban44:
Depends on https://github.com/vercel/vercel/pull/13985


packages/cli/src/commands/login/index.ts
```diff
@@ -78,5 +78,11 @@ export default async function login(
...
   telemetry.trackState('started');
 
-  return await future(client, telemetry);
+  try {
+    return await future(client, telemetry);
+  } catch (error: unknown) {
+    printError(error);
+    telemetry.trackState('error');
+    return 1;
+  }
```
markandrus, line 87: Tackle this in a separate PR? **EDIT:** Could be combined w/ the OAuth changes.

packages/cli/src/util/client.ts
```diff
@@ -174,20 +174,9 @@ export default class Client extends EventEmitter implements Stdio {
...
 
-    // If we had an error, during the refresh process, empty the auth config
-    // to force the user to re-authenticate
-    if (tokensError) {
-      output.debug('Error refreshing token, emptying auth config.');
-      this.emptyAuthConfig();
-      this.writeToAuthConfigFile();
-      return;
-    }
+    const tokens = await oauthClient.refreshToken(authConfig.refreshToken);
```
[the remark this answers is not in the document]
markandrus: This seems like a regression and that we should continue allowing token refreshes to fail

packages/cli/src/util/config/files.ts
```diff
@@ -51,33 +52,24 @@ export const writeToConfigFile = (stuff: GlobalConfig): void => {
...
-export const readAuthConfigFile = (): AuthConfig => {
-  const config = loadJSON.sync(AUTH_CONFIG_FILE_PATH);
-  return config;
-};
+export const readAuthConfigFile = credentialsStore.get;
 
 export const writeToAuthConfigFile = (authConfig: AuthConfig) => {
-  if (authConfig.skipWrite) {
-    return;
-  }
```
markandrus, line 62: Is this safe to remove?

markandrus (commented):
How about separating the OAuth changes into a separate PR?

packages/cli/src/util/config/files.ts
markandrus: Looks like it, since it is passed into `update`.

PR #13995, opened 2025-09-25
[python] update FastAPI starter example

ricardo-agz:
Updates example to match deployed example: https://vercel-plus-fastapi.vercel.app/


tknickman (approved):

ismaelrumzan (approved):

gr2m:
FYI please don't add changesets for examples. We don't publish these packages and it caused a release error (https://github.com/vercel/vercel/issues/14023) due to a `changeset/action` bug that is known and has a pending pull request since December 2022: https://github.com/changesets/action/issues/241

I'll think about preventing this from happening again from our side, but want to give you a heads up for the time being.

PR #13996, opened 2025-09-25
hono o11y poc

brookemosby:


packages/cli/src/commands/build/index.ts
```diff
@@ -630,7 +630,7 @@ async function doBuild(
             if (
               process.env.VERCEL_EXPERIMENTAL_EXPRESS_BUILD === '1' &&
               'name' in builder &&
-              builder.name === 'express' &&
+              (builder.name === 'express' || builder.name === 'hono') &&
               'experimentalBuild' in builder &&
               typeof builder.experimentalBuild === 'function'
             ) {
```
onsclom, line 644: Seems confusing that `VERCEL_EXPERIMENTAL_EXPRESS_BUILD` also triggers experimental hono builds. Maybe there should be another env var for hono?

packages/hono/src/experimental/introspection.ts
```diff
@@ -0,0 +1,182 @@
...
+    const methodPatterns = [
+      { method: 'GET', pattern: /app\.get\(["']([^"']+)["']/g },
+      { method: 'POST', pattern: /app\.post\(["']([^"']+)["']/g },
+      { method: 'PUT', pattern: /app\.put\(["']([^"']+)["']/g },
+      { method: 'DELETE', pattern: /app\.delete\(["']([^"']+)["']/g },
+      { method: 'PATCH', pattern: /app\.patch\(["']([^"']+)["']/g },
+      { method: 'OPTIONS', pattern: /app\.options\(["']([^"']+)["']/g },
+      { method: 'HEAD', pattern: /app\.head\(["']([^"']+)["']/g },
+      { method: 'ALL', pattern: /app\.all\(["']([^"']+)["']/g },
+    ];
```
onsclom: This seems like a different strategy than the express 011y poc? Looks like this wouldn't catch cases where `app` is named something else. [Routes](https://hono.dev/examples/grouping-routes-rpc) wouldn't be supported either.

Maybe the `app` naming is common enough, this works well enough for most cases?

onsclom (approved):
Approving since the changes are behind `VERCEL_EXPERIMENTAL_EXPRESS_BUILD`, so this will not affect existing hono builds!

packages/hono/src/experimental/introspection.ts
brookemosby: changed to use hono built in function

packages/cli/src/commands/build/index.ts
brookemosby: Good point - will fix 👍

PR #14026, opened 2025-10-01
[python] better build logging

ricardo-agz:
Makes python builds a bit more interpretable
* Log if no python version specified: `No Python version specified in pyproject.toml or Pipfile.lock. Using latest installed version: 3.12`
* Logs where it is installing dependencies from, e.g. `Installing required dependencies from requirements.txt...`
* Only log `using uv at ... ` once


tknickman (approved):

marc-vercel (approved):
LGTM, I it is mostly a small refactor and more logs addeed.

PR #14042, opened 2025-10-04
[ruby] zero-config ruby on rails

ricardo-agz:
Adds zero-config support for ruby on rails
(it looks like a lot but it's mostly just the framework example generated with `rails new` that has a ton of  files)

It:
* Adds rails as an auto-detected framework
* Adds an example for rails
* Ruby runtime excludes /public from bundle and uses static builder
* For rails, proxies errors to the cdn route
* Rails requires `SECRET_KEY_BASE` to be set, otherwise it fails to boot, so for zero config, if the user does not set a `SECRET_KEY_BASE` env, an auto-generated one is used

Deployed zero-config rails app: https://rails-zero-config-example.vercel.app/
(Using Supabase integration for db)


jeffsee55:
Seems like this does bundle assets, but only if they don't need to be compiled right? Will Rails serve those with appropriate headers? 

I wonder if we need to think a bit on how assets will be managed here in general, seems like ideally they're pushed to the CDN at build time, but perhaps if the headers are set properly it's not a big deal. 

In general, looks good!

ricardo-agz:
> Seems like this does bundle assets, but only if they don't need to be compiled right? Will Rails serve those with appropriate headers?
> 
> I wonder if we need to think a bit on how assets will be managed here in general, seems like ideally they're pushed to the CDN at build time, but perhaps if the headers are set properly it's not a big deal.
> 
> In general, looks good!

@jeffsee55 Yea Rails comes with html pages for error codes in the public folder and it reads from them directly to serve the html content so we can exclude the /public folder and use the static builder like we do for node/python, we just have to add a middleware to fetch the html content for those pages and return that, which is what the new commit does.

The alternative could be we don't exclude anything that looks like {code}.html in public but then maybe the inconsistencies of how we handle stuff in the public folder could be confusing

MEMZ-CHROER:
sorry,but i want to give an advice:
Our vercel is more and more well,but i think if it can optimize the connection of vercel.app domain in China,It will very good than before.
Because In fact Every Chinese people want to use vercel and vercel.app domain,but in china,It seems not good to explore vercel.app in china.
Thanks you
---- Replied Message ----
From Ricardo ***@***.***> Date 10/08/2025 06:40 To ***@***.***> Cc ***@***.***> Subject Re: [vercel/vercel] [ruby] zero-config ruby on rails (PR #14042) 
ricardo-agz left a comment (vercel/vercel#14042)
Seems like this does bundle assets, but only if they don't need to be compiled right? Will Rails serve those with appropriate headers?
I wonder if we need to think a bit on how assets will be managed here in general, seems like ideally they're pushed to the CDN at build time, but perhaps if the headers are set properly it's not a big deal.
In general, looks good!
@jeffsee55 Yea Rails comes with html pages for error codes in the public folder and it reads from them directly to serve the html content so we can exclude the /public folder and use the static builder like we do for node/python, we just have to add a middleware to fetch the html content for those pages and return that, which is what the new commit does.
The alternative could be we don't exclude anything that looks like {code}.html in public but then maybe the inconsistencies of how we handle stuff in the public folder could be confusing

kmk142789 (approved):

PR #14065, opened 2025-10-07
Add @vercel/backends package to consolidate backend framework builders

jeffsee55:
This introduces a new package (codename: `@vercel/cervel`) which can be used for generic node backends to both build and serve apps. It's in this monorepo, but isn't specific to any Vercel tooling, and it might make sense to move to it's own repo.

- Wraps `srvx` with the `tsx` loader
- Builds with `rolldown`
- Shares the same entrypoint detection logic that we're using for the backend presets.

The idea here would be that we'd update express/hono/etc framework presets with `build` and `dev` commands:

```
"dev": "cervel",
"build": "cervel build index.ts"
```

While internally, if there's no build script we supply `cervel build` under the hood (which will autodetect an entrypoint if not supplied).

---

This work is coming up during the o11y POC because we have exhausted how far we can get with using the `@vercel/node` builder. Essentially, the node builder is using version 3 of the build output API, and we need to use version 2.

All of this is behind the `VERCEL_EXPERIMENTAL_BACKENDS=1` env var so should be safe to move forward with.


packages/backends/CHANGELOG.md
onsclom, line 1: This file should be removed

onsclom (approved):

PR #14067, opened 2025-10-08
log functions exceeding the 250MB size limit

vikhyathdevadiga:
Adds opt-in serverless function size analysis (requires an environment variable `VERCEL_ANALYZE_BUILD_OUTPUT=1`) to detect and report functions exceeding the [250MB uncompressed size limit](https://vercel.link/serverless-function-size). The output will contain:
- Functions that exceed the size limit with their total uncompressed size
- A breakdown of the top 10 largest dependencies
- A summary of normal functions under the limit

> [!NOTE]
> Low Risk Change
>
> This PR adds opt-in build output analysis functionality gated by an environment variable to detect and report serverless functions exceeding size limits, with no security, auth, or data integrity changes.
> 
> - Adds new analysis functions to detect functions exceeding 250MB size limit
> - Feature gated behind VERCEL_ANALYZE_BUILD_OUTPUT=1 environment variable
> - Throws NowBuildError when functions exceed size limit
>
> <sup>Risk assessment for [commit facf1d5](https://github.com/vercel/vercel/commit/facf1d51da001d0024e2b016465c2873f242307c).</sup>


packages/cli/src/commands/build/index.ts
```diff
@@ -869,6 +871,203 @@ async function doBuild(
...
+    throw new NowBuildError({
+      code: 'NOW_SANDBOX_WORKER_MAX_LAMBDA_SIZE',
+      message: `${exceededFunctions.length} function${exceededFunctions.length === 1 ? '' : 's'} exceeded the uncompressed maximum size of ${LAMBDA_SIZE_LIMIT_MB} MB.`,
+      link: 'https://vercel.link/serverless-function-size',
+      action: 'Learn More',
+    });
+  }
+}
+
+async function analyzeSingleFunction(
```
jeffsee55, line 1367: I think we also need to count files that are in the `functions/output/index.func` folder itself, which this isn't doing.

The other thing I can't quite tell from this review is, are we deduping on symlinks? If we have 2 pointers to the same file that won't bloat the lambda size, not sure how common that is in practice.

vikhyathdevadiga: Do you mean `.vercel/output/functions/index.func`? Since we’re already calculating `filePathMap` in `.functions/index.func/.vc-config.json`, wouldn't that cover it?

Regarding symlinks, I'm not sure if `filePathMap` in `.vc-config.json` will contain any. But, running `vercel build`, I noticed `.vercel/output/functions/index.func` and `.vercel/output/functions/index.rsc.func` will be generated, and one of those will be a symlink. So, I've filtered out rsc files (line [937](https://github.com/vercel/vercel/pull/14067/files#diff-452a115a951a23488dd3257a83934cc9439918fbd41143a0ded219164dd13c6fR937)) and used statSync (line [1058](https://github.com/vercel/vercel/pull/14067/files#diff-452a115a951a23488dd3257a83934cc9439918fbd41143a0ded219164dd13c6fR1058)). Should this be enough to handle the deduplication?

jeffsee55: The `filePathMap` is not the full picture, the files in the `index.func` are all part of the zip, so we may underreport. An example that doesn't use `filePathMap` is the `@vercel/node` builder for api functions.

jeffsee55 (approved):

packages/cli/src/commands/build/index.ts
```diff
@@ -934,6 +936,203 @@ async function doBuild(
...
+  path: string;
+  size: number;
+  files: Map<string, number>;
+} | null> {
+  try {
+    const content = await fs.readFile(file, 'utf8');
+    const parsed = JSON.parse(content);
+
+    // Extract file paths from .vc-config.json
+    const filePathMap =
```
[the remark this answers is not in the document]
jeffsee55: This is what I was saying as well. Though this may be easier since they're all in the function's folder, so can just get the size of everything in there

packages/cli/src/commands/build/index.ts
vikhyathdevadiga: Thanks! That's correct. I just updated the code to include the files in the `.func` directory.

PR #14069, opened 2025-10-08
[functions] fix: move BuildCache timeout cleanup to prevent hanging on JSON parsing

kldavis4:
### TL;DR

Fix BuildCache timeout handling to prevent build hangs during response parsing.

### What changed?

Moved `clearTimeout()` calls to occur immediately after fetch completion rather than after `response.json()` parsing in the BuildCache class. This ensures that timeouts are properly cleared as soon as the network request completes, regardless of how long JSON parsing takes.

### How to test?

1. Test scenarios where JSON parsing might be slow or fail
2. Verify that requests no longer hang indefinitely during static generation
3. Confirm that network requests still respect the timeout while JSON parsing is allowed to complete

### Why make this change?

The previous implementation was causing 60-second build timeouts during static generation because the timeout wasn't being cleared until after JSON parsing completed. Since the abort signal doesn't affect JSON parsing anyway, this change allows the timeout to properly cover just the network request phase while letting JSON parsing complete without constraints. This prevents indefinite hangs when parsing is slow or fails.

### DIFF
--- a/packages/functions/src/cache/build-client.ts
+++ b/packages/functions/src/cache/build-client.ts
@@ -38,9 +38,9 @@ export class BuildCache {
         method: 'GET',
         signal: controller.signal,
       });
+      clearTimeout(timeoutId);
 
       if (res.status === 404) {
-        clearTimeout(timeoutId);
         return null;
       }
       if (res.status === 200) {
@@ -49,14 +49,10 @@ export class BuildCache {
         ) as PkgCacheState | null;
         if (cacheState !== PkgCacheState.Fresh) {
           res.body?.cancel?.();
-          clearTimeout(timeoutId);
           return null;
         }
-        const result = (await res.json()) as unknown;
-        clearTimeout(timeoutId);
-        return result;
+        return (await res.json()) as unknown;
       } else {
-        clearTimeout(timeoutId);
         throw new Error(`Failed to get cache: ${res.statusText}`);
       }
     } catch (error: any) {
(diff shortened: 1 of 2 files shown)


kldavis4:
* **#14069** <a href="https://app.graphite.com/github/pr/vercel/vercel/14069?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14069?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/functions/src/cache/build-client.ts
```diff
@@ -38,9 +38,9 @@ export class BuildCache {
         method: 'GET',
         signal: controller.signal,
       });
+      clearTimeout(timeoutId);
```
smaeda-ks, line 41: I'm slightly confused about this change and how it changes the behavior.

> This prevents indefinite hangs when JSON parsing is slow or fails

if `res.json()` throws then we still call `clearTimeout` in the catch block, no? Also, doesn't `res.json()` throw AbortError?
https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_operation_using_an_explicit_signal

> If the request is aborted after the fetch() call has been fulfilled but before the response body has been read, then attempting to read the response body will reject with an AbortError exception.

Should we just call `clearTimeout` in the parent finally block instead?

PR #14099, opened 2025-10-14
[frameworks] Detect nitro when nitro package exists

onsclom:
Nitro is switching from `nitropack (v2)` to `nitro (v3)`: https://v3.nitro.build/docs/quick-start#create-a-nitro-project

This updates Nitro framework detection to detect the `nitro` package.

### DIFF
--- a/packages/frameworks/src/frameworks.ts
+++ b/packages/frameworks/src/frameworks.ts
@@ -2333,7 +2333,7 @@ export const frameworks = [
       'Nitro lets you create web servers that run on multiple platforms.',
     website: 'https://nitro.build/',
     detectors: {
-      every: [{ matchPackage: 'nitropack' }],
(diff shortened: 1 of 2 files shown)


packages/frameworks/src/frameworks.ts
```diff
@@ -2351,7 +2351,7 @@ export const frameworks = [
         value: 'dist',
       },
     },
-    dependency: 'nitropack',
+    dependency: 'nitro',
```
tknickman: Do we need to support either nitropack or nitro?

onsclom: This field is actually unused for new frameworks with `detectors`, so deleted this field to make things more clear!

https://github.com/vercel/vercel/blob/43db6c62b29cbf093e729e362b5120c8ce46f1aa/packages/frameworks/src/types.ts#L194-L200

tknickman (approved):

PR #14101, opened 2025-10-14
Revert "Temporarily skip failing nextjs tests (#14034)"

ijjk:
Re-enables and updates tests for latest v16 canary as turbopack was made the default and PPR flag was disabled so we need to lock some tests that relied on those specific semantics to a previous canary and update newer ones that test more general features that have been moved to `cacheComponents` flag. 

This reverts commit ca7b9042fa3ac2877eeb672453cb13f163c00eab.


.changeset/empty-laws-think.md
```diff
@@ -0,0 +1,2 @@
+---
```
ztanner: we need an actual changeset since `next/src/utils` changed right?

ijjk: Ah true

ztanner (approved):

PR #14108, opened 2025-10-16
[node,next] add support for Bun as a JavaScript runtime

QuiiBz:
⚠️ This PR is still work-in-progress. Bun will ship as a beta runtime once ready

This PR adds support for using [Bun](https://bun.com/) as a JavaScript runtime, similar to the [Edge Runtime](https://vercel.com/docs/functions/runtimes/edge)

---

For vanilla functions using `@vercel/node`, you can opt in similarly to the Edge Runtime by exporting a `config.runtime`:
```ts
export const config = {
  runtime: 'bun' // or 'edge', 'nodejs' by default
}
```

Additionnaly and for other frameworks, like Next.js, you can opt in via `vercel.json`:
```json
{
  "functions": {
    "api/*.ts": {
      "runtime": "@vercel/bun"
    }
  }
}
```

The way this works is by registering a fake `@vercel/bun` builder, that maps to `@vercel/node`. This is for the same reason as why we have Edge Functions within `@vercel/node`: other framework builders like NestJS, Hono, etc, are re-exporting the builder methods from `@vercel/node`,  but they still need to support the Edge Runtime and now Bun

Within `@vercel/node` and `@vercel/next`, we can then detect wether the static config for a route uses Bun, or if one of the `vercel.json`'s `functions` config matches that route and uses Bun. If yes, we update the Build Output API function primitive  with the `runtime: 'bun1.x'` property, instead of e.g. `runtime: 'node22.x'`, which tells Vercel's API to deploy that function using Bun. For Next.js specifically, we also use the runtime as a bundling criteria

This PR also adds support for starting Bun in `vercel dev`. Because we spawn a `bun` process from within the `node` process of the CLI, we cannot send IPC messages to retrieve the HTTP port. Instead, we read stdout until the "dev server started message" is emitted by Bun and extract the HTTP port. If Bun is not available in the `PATH` or in `$HOME/.bun/bin`, we automatically execute the [installation script](https://bun.com/docs/installation) for the correct platform

TODOs:
- [x] Integrations tests
- [x] Download the correct `bun` binary and use it when starting a dev server, instead of assuming it's available in the `PATH`
- [ ] To discuss: should we use `@vercel/bun` (like existing builders) or just `bun` inside `vercel.json`?


packages/node/src/fork-dev-server.ts
```diff
@@ -17,54 +23,96 @@ export function forkDevServer(options: {
...
+      env: cloneEnv(process.env, options.meta.env, {
+        VERCEL_DEV_ENTRYPOINT: options.entrypoint,
+        VERCEL_DEV_CONFIG: JSON.stringify(options.config),
+        VERCEL_DEV_BUILD_ENV: JSON.stringify(options.meta.buildEnv || {}),
+        VERCEL_DEV_PUBLIC_DIR: options.publicDir || '',
+      }),
+      stdio: ['pipe', 'pipe', 'pipe'],
+    };
+
+    child = spawn('bun', ['--bun', devServerPath], spawnOptions);
```
jeffsee55: Why not just include `--port` arg for a port we know is free?

QuiiBz:
This was superseded by https://github.com/vercel/vercel/pull/14130, unfortunaly my commits haven't been cherry-picked

PR #14120, opened 2025-10-20
Add Fastify zero-config support

onsclom:
Previously Fastify projects would need to use the `/api` dir and a `vercel.json` with:
```
{
  "rewrites": [
    {
      "source": "/(.*)",
      "destination": "/api"
    }
  ]
}
```

This adds a framework preset which handles a list of default entry points like we do for Express, Hono, and NestJS.


packages/fastify/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@vercel/fastify",
+  "version": "0.1.6",
```
jeffsee55: Not sure it matters at all but maybe good to put this at `0.0.1`?

jeffsee55 (approved):

joe10832 (approved):
@vercel-release-bot

PR #14164, opened 2025-10-28
Add Elysia + use bun for nft conditions

jeffsee55:
Add Elysia framework, use `bun` as NFT [import condition](https://github.com/vercel/nft#exports--imports) from the node builder


packages/frameworks/src/frameworks.ts
```diff
@@ -3311,6 +3311,229 @@ export const frameworks = [
     dependency: 'nestjs',
     getOutputDirName: async () => 'public',
   },
+  {
```
LukeSheard, line 3314: Is now the time to start splitting this file up? lol

jeffsee55: 🤷 personally I don't mind it

tknickman (approved):

PR #14168, opened 2025-10-29
[python] allowing /api folder entrypoints for FastAPI + Flask

ricardo-agz:
This PR adds support for entrypoints inside the /api folder for FastAPI/Flask apps.
It adds `ignoreRuntimes: ['@vercel/python'],` because otherwise the API builder routes will overshadow the framework routes and 404 when they shouldn't

### DIFF
--- a/.changeset/red-forks-help.md
+++ b/.changeset/red-forks-help.md
@@ -1,0 +1,7 @@
+---
+"vercel": patch
+"@vercel/frameworks": patch
+"@vercel/python": patch
+---
+
+[python] allowing /api folder entrypoints for FastAPI + Flask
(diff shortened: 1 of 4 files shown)


tknickman (commented):
Should this be a major? Is there a chance someone is forcing FastAPI into an api route now and may see different behavior with this release?

ricardo-agz:
> Should this be a major? Is there a chance someone is forcing FastAPI into an api route now and may see different behavior with this release?

@tknickman Yea I guess we should, it is possible (although I'd be really surprised if someone was actually doing this) that someone could have their main fastapi app at `main.py`, then individual fastapi apps inside each file in `/api` to work with the current way we handle the Python api builder and this would change that behavior, those python files inside `/api` would stop getting the special api builder treatment

tknickman:
Let's do it just to make sure we're correctly following semver since it _is_ a behavior change.

tknickman (approved):

PR #14169, opened 2025-10-29
[cli] add new commands `vercel cache invalidate --srcimg`

styfle:
This PR adds two new commands:

- `vercel cache invalidate --srcimg /api/avatar/1`
- `vercel cache dangerously-delete --srcimg /api/avatar/1 --revalidation-deadline-seconds 604800`

### DIFF
--- a/packages/cli/src/commands/cache/dangerously-delete.ts
+++ b/packages/cli/src/commands/cache/dangerously-delete.ts
@@ -48,20 +48,41 @@ export default async function dangerouslyDelete(
   client.config.currentTeam = org.type === 'team' ? org.id : undefined;
   const yes = Boolean(parsedArgs.flags['--yes']);
   const tag = parsedArgs.flags['--tag'];
+  const srcimg = parsedArgs.flags['--srcimg'];
+  const revalidate = parsedArgs.flags['--revalidation-deadline-seconds'];
   telemetry.trackCliFlagYes(yes);
   telemetry.trackCliOptionTag(tag);
+  telemetry.trackCliOptionSrcimg(srcimg);
+  telemetry.trackCliOptionRevalidationDeadlineSeconds(revalidate);
 
-  if (!tag) {
-    output.error(`The --tag option is required`);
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/cache/dangerously-delete.ts
```diff
@@ -48,20 +48,41 @@ export default async function dangerouslyDelete(
...
+  const srcimg = parsedArgs.flags['--srcimg'];
+  const revalidate = parsedArgs.flags['--revalidation-deadline-seconds'];
   telemetry.trackCliFlagYes(yes);
   telemetry.trackCliOptionTag(tag);
+  telemetry.trackCliOptionSrcimg(srcimg);
+  telemetry.trackCliOptionRevalidationDeadlineSeconds(revalidate);
 
-  if (!tag) {
-    output.error(`The --tag option is required`);
+  if (tag && srcimg) {
```
shraddha55, line 58: Should we add a check that atleast one of them is required?

shraddha55: Actually i see it here nvm - https://github.com/vercel/vercel/pull/14169/files#diff-2bc4be7c350edf0eed6a087180df08098c510367d5f7785b3abfe54774f2d4deR81

styfle: Yeah at least one is required. So we error if neither are provided (there are tests for that).

We also error if both are provided (though I forgot to add a test for that case)

shraddha55 (approved):
Great PR! Looks good to me!

packages/cli/src/commands/cache/dangerously-delete.ts
styfle: @shraddha55 Added a test in [43c2aad](https://github.com/vercel/vercel/pull/14169/commits/43c2aad64b02a696e5a091470ddbda6e57b14fad)

balazs4 (approved):
LGTM

PR #14176, opened 2025-10-30
[build-utils]: Remove getSpawnOptions

LukeSheard:
# Overview

Remove the `getSpawnOptions` function from being used for subprocesses. This function is designed to setup any subprocesses running from the CLI with the correct version of node in `PATH` for the build. We've ported this logic to the build environment now so that any processes running are run with the right `PATH` environment variable set for their build settings (respecting `package.json`).

I've refactored the builders to now only use `getEnvForPackageManager` in places which require the subprocess. The intention here is to remove "vercel environment specific" logic from the CLI and have that owned by the build container - so that the CLI is a smaller wrapper around builder specific logic. 

This logic is next to be moved to the Vercel environment so that the right zero-config detection is run when spawning the CLI, which makes this logic detection redundant.


LukeSheard:
* **#14176** <a href="https://app.graphite.com/github/pr/vercel/vercel/14176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14176?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

marc-vercel (approved):
LGTM code wise; has this been tested with the `build-container` in lhr1?

PR #14177, opened 2025-10-30
fix(next): treat interception routes with dynamic params as dynamic

wyattjoh:
## What

Fixes an issue where interception routes with dynamic path parameters were not being recognized as dynamic routes during prerendering.

## Why

When cache components is enabled, interception routes with dynamic parameters (like `/foo/(..)bar/[id]`) were being treated as static routes. This caused incorrect payloads to be served because the prerender logic wasn't handling them as dynamic routes that need special payload generation.

The root cause is that `isDynamicRoute()` doesn't currently support the interception route syntax (e.g., `(..)`, `(...)`, `(.)`).

## How

Patched the `isDynamicRoute` to handle interception routes the same way that Next.js does from Next.js 16 onwards.

NAR-482

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2883,7 +2883,11 @@ export const onPrerenderRoute =
           allowQuery = Object.values(routeKeys);
         }
       } else {
-        const isDynamic = isDynamicRoute(pageKey);
+        // We added the `!!route` check because the isDynamicRoute does not
+        // currently support intercepting routes, but if it exists in the
+        // `dynamicRoutes` part of the routes manifest, then we know that it's
+        // dynamic!
+        const isDynamic = !!route || isDynamicRoute(pageKey);
 
         if (routeKeys) {
           // if we have routeKeys in the routes-manifest we use those
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2858,7 +2858,11 @@ export const onPrerenderRoute =
         (r): r is RoutesManifestRoute =>
           r.page === pageKey && !('isMiddleware' in r)
       ) as RoutesManifestRoute | undefined;
-      const isDynamic = isDynamicRoute(routeKey);
+      // We added the `!!route` check because the isDynamicRoute does not
+      // currently support intercepting routes, but if it exists in the
+      // `dynamicRoutes` part of the routes manifest, then we know that it's
+      // dynamic!
+      const isDynamic = !!route || isDynamicRoute(routeKey);
       const routeKeys = route?.routeKeys;
 
       // by default allowQuery should be undefined and only set when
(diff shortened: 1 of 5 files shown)


packages/next/src/utils.ts
```diff
@@ -2858,7 +2858,12 @@ export const onPrerenderRoute =
         (r): r is RoutesManifestRoute =>
           r.page === pageKey && !('isMiddleware' in r)
       ) as RoutesManifestRoute | undefined;
-      const isDynamic = isDynamicRoute(routeKey);
+      const isRouteKeyDynamic =
+        routesManifest?.dynamicRoutes.some(
+          (r): r is RoutesManifestRoute =>
+            r.page === routeKey && !('isMiddleware' in r)
+        ) ?? false;
```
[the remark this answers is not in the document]
wyattjoh: Nope! The purpose of `routeKey` and `srcRoute` are different!

packages/next/src/utils.ts
```diff
@@ -2879,7 +2884,11 @@ export const onPrerenderRoute =
           allowQuery = Object.values(routeKeys);
         }
       } else {
-        const isDynamic = isDynamicRoute(pageKey);
+        // We added the `!!route` check because the isDynamicRoute does not
```
ztanner: what `!!route` check?

packages/next/src/utils.ts
```diff
@@ -2858,7 +2858,12 @@ export const onPrerenderRoute =
         (r): r is RoutesManifestRoute =>
           r.page === pageKey && !('isMiddleware' in r)
       ) as RoutesManifestRoute | undefined;
-      const isDynamic = isDynamicRoute(routeKey);
+      const isRouteKeyDynamic =
+        routesManifest?.dynamicRoutes.some(
+          (r): r is RoutesManifestRoute =>
+            r.page === routeKey && !('isMiddleware' in r)
+        ) ?? false;
+      const isDynamic = isRouteKeyDynamic || isDynamicRoute(routeKey);
```
ztanner: does this handling need to be baked into `isDynamicRoute`? I'm wondering if there are other cases that are are incorrectly marking these routes as static when they should be dynamic. 

I suspect that util came from Next.js and has drifted as implementations changed (eg the one in Next has handling for interception routes: https://github.com/vercel/next.js/blob/13174c55256308ba6864ce7cd3fe53e30f2cd7dd/packages/next/src/shared/lib/router/utils/is-dynamic.ts)

I feel like we need to either fully rely on a util, or fully rely on Next.js telling us something's dynamic. Feels strange to do both

ztanner (approved):

CosmicJesterX (commented):
this ok? i DO have other systems im working with also...

CosmicJesterX (commented):
=)

PR #14192, opened 2025-11-01
[python] asgi lifespan events support

ricardo-agz:
Swaps out our custom http server handler for asgi apps with [uvicorn](https://uvicorn.dev/), an actual asgi server, which automatically handles asgi startup / shutdown events as opposed to implementing our own lifespan manager like we implement in https://github.com/vercel/vercel/pull/14037

**Note:** running Sanic on the pinned version of uvicorn breaks when running versions < 21.0.0 like we do in fixture 11-asgi so this PR upgrades Sanic to the latest version (25.3.0)


brookemosby (approved):
LGTM

PR #14195, opened 2025-11-02
feat(cli): add vercel httpstat command

swarnava:
Introduce a new `vercel httpstat` command that provides HTTP timing statistics for Vercel deployments, similar to `vercel curl` but with timing visualization.

Features:
- Automatic deployment URL resolution (latest or specific deployment)
- Automatic protection bypass authentication handling
- Pass-through arguments to httpstat after '--' separator
- Full telemetry integration
- Comprehensive test coverage
- Detailed documentation with examples

The command wraps the httpstat tool (which must be installed separately) and automatically handles:
- Deployment URL lookup
- Protection bypass secrets
- Header injection for Vercel Protection

Usage examples:
  ```sh
 $ vercel httpstat /api/hello vercel httpstat /api/users -- -X POST -d '{"name": "John"}' 
 
 $ vercel httpstat /api/status --deployment dpl_ABC123 
 
 $ vercel httpstat /api/protected --protection-bypass <secret>
  ```
  
  
image


packages/cli/src/commands/httpstat/index.ts
```diff
@@ -0,0 +1,207 @@
...
+        `x-vercel-protection-bypass: ${deploymentProtectionToken}`
+      );
+    }
+  }
+
+  httpstatFlags.unshift(fullUrl);
+
+  output.debug(
+    `Executing: httpstat ${httpstatFlags.map(requoteArgs).join(' ')}`
+  );
```
dimitropoulos, line 49: it looks like this is nearly identical to the same 150 lines in the curl command - can we please, then, centralize the behavior for getting the flags to a common utility so they don't fall out of sync?

packages/cli/src/commands/httpstat/utils.ts
```diff
@@ -0,0 +1,10 @@
+export const requoteArgs = (arg: string): string => {
+  if (arg.includes(' ')) {
+    if (arg.includes('"')) {
+      return `'${arg}'`;
+    } else {
+      return `"${arg}"`;
+    }
+  }
+  return arg;
+};
```
dimitropoulos: this is identical to the same utility in https://github.com/vercel/vercel/blob/06a8810b86134b7f3f800f0be82835198ea84a75/packages/cli/src/commands/curl/utils.ts#L1

can we please remove it?

dimitropoulos (approved):
tested and it works great!

I tested the "command not found" and that was great

Screenshot_20251105_113026

and the core use-case works 
Screenshot_20251105_113109

flag passing also worked great, here's `--json-output` for example:

Screenshot_20251105_113207

PR #14205, opened 2025-11-03
Upgrade Typescript

jeffsee55:
- Upgrades to latest Typescript version, including update to the node builder, which will be used to transpile user code
- Replaces ts-node with tsx for node dev server


balazsorban44 (approved):

packages/cli/src/commands/env/pull.ts
```diff
@@ -35,7 +35,7 @@ function readHeadSync(path: string, length: number) {
   const buffer = Buffer.alloc(length);
   const fd = openSync(path, 'r');
   try {
-    readSync(fd, buffer, 0, buffer.length, null);
+    readSync(fd, buffer as any, 0, buffer.length, null);
```
marc-vercel, line 38: Is it possible to avoid the `as any` and pass a convert this to a proper type (i.e. `as Buffer`)?

.changeset/cuddly-masks-yawn.md
```diff
@@ -0,0 +1,17 @@
...
+'@vercel/frameworks': patch
+'@vercel/firewall': patch
+'@vercel/cervel': patch
+'@vercel/client': patch
+'@vercel/edge': patch
+'@vercel/node': patch
+'vercel': patch
+---
+
+Upgrade Typescript
```
marc-vercel, line 17: I think this should be a breaking change, but not a strong opinion (happy to keep the patch). Bumping to Typescript v4 -> v5, means that we are updating dependencies that before we could not because we were not at Typescript v5. Any of these dependencies might use new language features not available in v4 and it might break customer apps.

CLI can be a patch, but any other libraries should be a breaking change.

marc-vercel (approved):
Approved, left some comments. Let's try to merge it today or tomorrow to avoid having a big PR stale and having to do endless rebases.

PR #14209, opened 2025-11-05
feat: [`vercel curl` & `vercel httpstat`] validation for full url

dimitropoulos:
This PR adds:

- `vercel curl` & `vercel httpstat` validation such that that if someone passes a url like `vercel curl https://tek-nitro-app-git-update-data-uncurated-tests.vercel.app/` it would previously send a request to `https://tek-nitro-app-git-update-data-uncurated-tests.vercel.app/https://tek-nitro-app-git-update-data-uncurated-tests.vercel.app/` which is obviously wrong.  So now we detect this and error and tell the user that they need to provide a path to the deployment
- adds guidance for users that they can target other specific deployments within the currently linked project using the --deployment flag since the backend actually accepts a url and will look up the deployment based on it


swarnava (approved):

packages/cli/src/commands/curl/README.md
```diff
@@ -23,14 +23,15 @@ vercel curl <path> [options] [-- <curl-args>]
 - **`path`** (required): The API path to request
   - Can start with `/` (e.g., `/api/hello`) or without (e.g., `api/hello`)
   - Automatically normalized to include leading slash
+  - Must be a relative path — do not pass a full URL. To target a specific deployment base URL, use `--deployment <id>`
```
jeffsee55, line 26: Why can't they just pass the full url? At which point, also providing a `--deployment` option should be considered invalid

jeffsee55 (approved):
LGTM, just a question, non-blocking

packages/cli/src/commands/curl/README.md
dimitropoulos: we discussed it and we don't (at least in the beginning) want to encourage a user experience that causes people to conclude they first need to get the full URL.  a huge portion of what this tool does (roughly half) is figure that out _for you_, so it should (by that definition) be somewhat rare/unusual that you'd require doing that.

PR #14210, opened 2025-11-05
Pull introspection module out of backends for external use

jeffsee55:
Nitro may experiment with using the introspection stuff with their v3 `server.ts` entry, so pullling it out of `backends` package.

This now moves introspection tests as well, and unit tests in `backends` are more focused on making sure the lambdas they create actually run.


tknickman (approved):
Mostly moves / refactors - lgtm as long as CI is happy

PR #14214, opened 2025-11-06
Patch OIDC data dir permissions

alice-wondered:
previously was 0o660 which didn't allow directory operations in linux

### DIFF
--- a/packages/oidc/src/token-util.ts
+++ b/packages/oidc/src/token-util.ts
@@ -97,7 +97,7 @@ export function saveToken(token: VercelTokenResponse, projectId: string): void {
     }
     const tokenPath = path.join(dir, 'com.vercel.token', `${projectId}.json`);
     const tokenJson = JSON.stringify(token);
-    fs.mkdirSync(path.dirname(tokenPath), { mode: 0o660, recursive: true }); // read/write perms for owner only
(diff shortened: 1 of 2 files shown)


packages/oidc/src/token-util.ts
```diff
@@ -97,7 +97,7 @@ export function saveToken(token: VercelTokenResponse, projectId: string): void {
     }
     const tokenPath = path.join(dir, 'com.vercel.token', `${projectId}.json`);
     const tokenJson = JSON.stringify(token);
-    fs.mkdirSync(path.dirname(tokenPath), { mode: 0o660, recursive: true }); // read/write perms for owner only
+    fs.mkdirSync(path.dirname(tokenPath), { mode: 0o770, recursive: true }); // read/write/exec perms for owner/group only, x required for dir ops
```
shaper, line 100: do we need to worry about existing directories with bad perms, like look for them and remove or edit their perms or anything?

shaper (approved):

packages/oidc/src/token-util.ts
alice-wondered: great question... wonder if that should be a follow up? This feature is new enough that it might only affect a handful of users

AAorris (approved):

PR #14216, opened 2025-11-06
[ruby] support local `vercel build` with ruby

ricardo-agz:
Fixes issue where running `vc build` locally doesn't work because it can't identify your system ruby
Also adds `predefindedExcludes` to exclude files from lambda


packages/ruby/src/index.ts
```diff
@@ -95,19 +100,47 @@ async function bundleInstall(
...
-      console.error(result.stderr);
-      throw result;
+    if (lockResult.exitCode !== 0) {
+      console.log(lockResult.stdout);
+      console.error(lockResult.stderr);
+      throw lockResult;
     }
   }
 
+  console.log('Running bundle install...');
```
LukeSheard: I think you should run the frozen-install before you run the non-frozen one. I assume that the non-resolution behaves the same way as node package managers which might modify user-space deps?

packages/ruby/src/install-ruby.ts
```diff
@@ -47,12 +49,69 @@ function getLatestRubyVersion(): RubyVersion {
...
+  major: number;
+  minor: number;
+  rubyPath: string;
+  gemPath: string;
+}) {
+  const gemHome = join(tmpdir(), `vercel-ruby-${major}${minor}-${process.pid}`);
+  return {
+    major,
+    gemHome,
+    runtime: `ruby${major}.${minor}`,
```
LukeSheard: We should have this set to whatever we have in the Vercel build environment so that we're deploying consistently across the two. 

The way that I've thought about this for CLI outside of Vercel is we bring in your configuration setting from Vercel and locally you're responsible for setting your your environment equivalently.

LukeSheard (approved):

packages/ruby/src/index.ts
ricardo-agz: good catch, I updated this to first add the dependencies we need like `webrick` to the gemfile if not already included and then lock, which won't modify existing user pinned deps, and then run a single frozen `bundle install`

packages/ruby/src/install-ruby.ts
ricardo-agz: this makes sense

PR #14221, opened 2025-11-07
Include account info in requests

jeffsee55:
When issuing a request while logged into a team that doesn't own the given project, this would fail to issue a bypass token

### DIFF
--- a/packages/cli/src/commands/curl/bypass-token.ts
+++ b/packages/cli/src/commands/curl/bypass-token.ts
@@ -22,24 +22,17 @@ async function createDeploymentProtectionToken(
     );
   }
 
-  const query = new URLSearchParams();
-  if (orgId) {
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/curl/bypass-token.ts
```diff
@@ -22,24 +22,17 @@ async function createDeploymentProtectionToken(
...
-        },
-      }
-    );
+    }>(`/v1/projects/${projectId}/protection-bypass`, {
+      method: 'PATCH',
+      body: '{}',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      accountId: orgId,
```
dimitropoulos, line 34: tyvm! this was the problem, eh?

dimitropoulos (approved):

PR #14096, opened 2025-10-13
[function] Add addCacheTag api

shraddha55:
### Problem

Added addCacheTag Api to function package.

### What changed?
* Introduced a new API: `addCacheTag`
* Added a changeset file to document the minor version change

### How to test?

1. Import the addcachetag functions in a serverless function:
   ```typescript
   import { addCacheTag } from '@vercel/functions';
   ```

2. Test add cache tags:
   ```typescript
   await addCacheTag('my-tag');
   ```

### Why make this change?

This change provides a standardized API for adding tags to content within the functions package, allowing developers to associate cached content with tag or tags.

PR #14241, opened 2025-11-12
[function] Patch addCacheTag api

shraddha55:
## Summary
  Fix incorrect type definition and implementation in `addCacheTag` module.
  Related PR - https://github.com/vercel/vercel/pull/14096

  ## Problem
  - The type definition `AddCacheTagApi` was defined as an object interface with an `addCacheTag` method, but at runtime
  `getContext().addCacheTag` returns a function directly
  - The implementation was incorrectly trying to call `api.addCacheTag(tag)` when `api` is already the function itself, causing the
   error: `"addCacheTag is not a function"`

  ## Solution
  - Changed `AddCacheTagApi` from an interface to a type alias that represents a function directly: `(tag: string | string[]) =>
  Promise<void>`
  - Updated the implementation to call the API function directly: `api(tag)` instead of `api.addCacheTag(tag)`
  
  ## Test Plan
  - [ ] Verify `addCacheTag` works correctly in Next.js
  - [ ] Test with single tag and array of tags


packages/functions/src/addcachetag/index.ts
```diff
@@ -3,7 +3,7 @@ import { getContext } from '../get-context';
 export const addCacheTag = (tag: string | string[]) => {
   const api = getContext().addCacheTag;
   if (api) {
-    return api.addCacheTag(tag);
+    return api(tag);
```
styfle:
```suggestion
const addCacheTag = getContext().addCacheTag;
if (addCacheTag) {
    return addCacheTag(tag);
```

I think this reads better

.changeset/quiet-cars-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/functions': minor
+---
+
+Patch addCacheTag api
```
kldavis4: can you provide more detail here - this will go in the public changelog

.changeset/quiet-cars-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/functions': minor
```
styfle:
```suggestion
'@vercel/functions': patch
```

This is a patch since its fixing a bug

.changeset/quiet-cars-grin.md
styfle:
```suggestion
Fix `addCacheTag()` which was previously not working (no-op).
```

kldavis4 (approved):

marc-vercel (approved):

styfle (approved):

PR #14272, opened 2025-11-14
Improve warnings and errors around Bun and Elysia

onsclom:
1. If we detect Elysia with Node, we print the warning:  "Warning: Currently using Elysia with Node.js. To use Bun, add `"bunVersion": "1.x"` to `vercel.json`."
2. Removes old warning from when Bun was not supported as a runtime: "Warning: Bun is used as a package manager at build time only, not at runtime with Functions"

~~Error if detecting `Bun.serve()` in dev: "Bun.serve() is currently not supported in Vercel serverless functions. Export your handler as the default export instead."~~ (This is breaking other unit tests for unclear reasons. Removing this to unblock the other 2 improvements.)


jeffsee55 (approved):

packages/node/src/build.ts
```diff
@@ -210,6 +210,14 @@ async function compile(
       ? ['bun']
       : undefined;
 
+  if (config.framework === 'elysia') {
```
LukeSheard: Why doesn't `elysia` have it's own builder for this detection? 

I have been thinking about this and I feel like we don't have the right encapsulation for builders to support extending them, which is why we `builds` is an Array instead of a single value within the zero-config build framework still.

onsclom: Good point, this wasn't the best place for this. Refactored things to move this detection into the `elysia` builder, which now allows other backend frameworks to add their own checks more easily.

PR #14279, opened 2025-11-14
Include `.vercel/routes.json` during `vc deploy`

jeffsee55:
`vc deploy` auto-ignores `.vercel`, but we want it to include the `.vercel/routes.json` if it exists

### DIFF
--- a/packages/client/src/utils/index.ts
+++ b/packages/client/src/utils/index.ts
@@ -117,10 +117,12 @@ export async function buildFileTree(
     };
     fileList = await readdir(path, [ignores]);
 
+    // Check for special files that should be included even if ignored
+    const refs = new Set<string>();
+
     if (prebuilt) {
       // Traverse over the `.vc-config.json` files and include
       // the files referenced by the "filePathMap" properties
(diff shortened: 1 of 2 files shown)


packages/client/src/utils/index.ts
```diff
@@ -153,10 +155,22 @@ export async function buildFileTree(
...
         debug(`Error detecting microfrontend config: ${e}`);
       }
+    }
 
-      if (refs.size > 0) {
-        fileList = fileList.concat(Array.from(refs));
+    // Check for .vercel/routes.json and include it if it exists
+    try {
+      const routesJsonPath = join(path, '.vercel', 'routes.json');
+      const routesJsonContent = await maybeRead(routesJsonPath, null);
```
tknickman, line 163: Do we actually care about the contents here? If not `fs.stat` is probably faster.

tknickman (approved):
One note, but not blocking

packages/client/src/utils/index.ts
jeffsee55: Will fast-follow

PR #14290, opened 2025-11-17
Skip failing nextjs tests

onsclom:
These tests are failing on main: https://github.com/vercel/vercel/pull/14291


marc-vercel (approved):
Approved - do we have a ticket so that we remember to fix that?

.changeset/weak-baboons-fix.md
```diff
@@ -0,0 +1,2 @@
+---
```
marc-vercel, line 1: We are missing a description to the patch update -  please add it before merging 🙏

onsclom: Empty changesets like this get ignored and allow us to pass the changeset lint check. I think an empty changeset is correct here because this PR only touches tests and not code that users run.

PR #14312, opened 2025-11-19
[cli] create util to execute and set config path conditionally based on vercel.ts

pranavkarthik10:
Created a new util to conditionally check whether a `vercel.ts` configuration file exists, and if so return it as the config path. Existing logic to check between a `now.json` and a `vercel.json` is still available. 

This util executes the TypeScript configuration file and generates an intermediary `vercel.json` inside `VERCEL_DIR`.  It's protected by a feature flag, and will be protected similarly in the actual commands where it's used (`dev`, `build`, and `deploy`).

### DIFF
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -61,8 +61,6 @@ export async function compileVercelConfig(
     };
   }
 
-  output.debug(`Compiling vercel.ts to ${compiledConfigPath}`);
-
   try {
     const { build } = await import('esbuild');
 
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -61,12 +61,12 @@ export async function compileVercelConfig(
     };
   }
 
+  const tempOutPath = join(vercelDir, 'vercel-temp.js');
+
   try {
     const { build } = await import('esbuild');
 
     await mkdir(vercelDir, { recursive: true });
-
-    const tempOutPath = join(vercelDir, 'vercel-temp.js');
 
     await build({
       entryPoints: [vercelTsPath],
@@ -91,12 +91,6 @@ export async function compileVercelConfig(
 
     output.debug(`Compiled vercel.ts -> ${compiledConfigPath}`);
 
-    try {
-      await unlink(tempOutPath);
-    } catch (err) {
-      output.debug(`Failed to cleanup temp file: ${err}`);
-    }
-
     return {
       configPath: compiledConfigPath,
       wasCompiled: true,
@@ -107,6 +101,14 @@ export async function compileVercelConfig(
       message: `Failed to compile vercel.ts: ${error.message}`,
       link: 'https://vercel.com/docs/projects/project-configuration',
     });
+  } finally {
+    try {
+      await unlink(tempOutPath);
+    } catch (err: any) {
+      if (err.code !== 'ENOENT') {
+        output.debug(`Failed to cleanup temp file: ${err}`);
+      }
+    }
   }
 }
 
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -157,15 +157,24 @@ export async function compileVercelConfig(
         stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
       });
 
+      const timeout = setTimeout(() => {
+        child.kill();
+        reject(new Error('Config loader timed out after 10 seconds'));
+      }, 10000);
+
       child.on('message', message => {
+        clearTimeout(timeout);
+        child.kill();
         resolve(message);
       });
 
       child.on('error', err => {
+        clearTimeout(timeout);
         reject(err);
       });
 
       child.on('exit', code => {
+        clearTimeout(timeout);
         if (code !== 0) {
           reject(new Error(`Config loader exited with code ${code}`));
         }
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -154,7 +154,6 @@ export async function compileVercelConfig(
     const config = await new Promise((resolve, reject) => {
       const child = fork(loaderPath, [tempOutPath], {
         stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
-        env: process.env,
       });
 
       child.on('message', message => {
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -146,7 +146,7 @@ export async function compileVercelConfig(
 
     const loaderScript = `
       const configModule = await import(process.argv[2]);
-      const config = configModule.default || configModule.config || configModule;
+      const config = ('default' in configModule) ? configModule.default : (configModule.config || configModule);
       process.send(config);
     `;
     await writeFile(loaderPath, loaderScript, 'utf-8');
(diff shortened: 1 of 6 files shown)


pranavkarthik10:
* **#14324** <a href="https://app.graphite.com/github/pr/vercel/vercel/14324?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14314** <a href="https://app.graphite.com/github/pr/vercel/vercel/14314?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14313** <a href="https://app.graphite.com/github/pr/vercel/vercel/14313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14312** <a href="https://app.graphite.com/github/pr/vercel/vercel/14312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,132 @@
...
+      format: 'cjs',
+      outfile: tempOutPath,
+      packages: 'external',
+      target: 'node18',
+      sourcemap: 'inline',
+    });
+
+    delete require.cache[require.resolve(tempOutPath)];
+    const configModule = require(tempOutPath);
+    const config = configModule.default || configModule.config || configModule;
```
jeffsee55: Will you be adding any validation to the output?

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,132 @@
...
+      platform: 'node',
+      format: 'cjs',
+      outfile: tempOutPath,
+      packages: 'external',
+      target: 'node18',
+      sourcemap: 'inline',
+    });
+
+    delete require.cache[require.resolve(tempOutPath)];
+    const configModule = require(tempOutPath);
```
jeffsee55: Will env variables be supported? If so, we'd need to invoke this as a child process I think.

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,132 @@
...
+      configPath: hasVercelJson
+        ? vercelJsonPath
+        : hasNowJson
+          ? nowJsonPath
+          : null,
+      wasCompiled: false,
+    };
+  }
+
+  const vercelTsPath = join(workPath, 'vercel.ts');
```
MatthewStanciu: can be in a follow up but we also need to make sure we're supporting `vercel.js`

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: yes, the output will be validated automatically the same way vercel.json is in the commands: dev, build, deploy, where this util will be called.

pranavkarthik10: using this function in those commands is in a later pr on the stack

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: Yes, env variables should be supported. After digging into it a bit I've found that after making local environment files available within this function via dotenvConfig, it works. But I'm open to updating to use a child process as well.

jeffsee55: Ok, I'm curious how this will actually work in practice with `esm/cjs` interop. I suppose. esbuild doing the transform is solid enough. But I do wonder about whether this should be invoked as a fork process.

packages/cli/src/util/compile-vercel-config.ts
jeffsee55: We should support most commmon extensions: `mjs|cjs|js|mts|cts|ts`. perhaps cts can be omitted

pranavkarthik10: was going to do it as a follow up but it makes sense to tackle it now. I guess this also adds in another layer of validation to check whether multiple `vercel.{mjs|cjs|js|mts|cts|ts}` exist at the same time.

pranavkarthik10: added support for the extensions listed aside from cts, with a top level const

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: For sure, after some testing on performance and security we can revisit this

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,178 @@
...
+  sourceFile?: string;
+}
+
+const VERCEL_CONFIG_EXTENSIONS = ['ts', 'mts', 'js', 'mjs', 'cjs'] as const;
+
+function findAllVercelConfigFiles(workPath: string): string[] {
+  const foundFiles: string[] = [];
+  for (const ext of VERCEL_CONFIG_EXTENSIONS) {
+    const configPath = join(workPath, `vercel.${ext}`);
+    if (existsSync(configPath)) {
```
srintern: can we use `fs.access()` instead?

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,178 @@
...
+  const vercelJsonPath = join(workPath, 'vercel.json');
+  const nowJsonPath = join(workPath, 'now.json');
+  const hasVercelJson = existsSync(vercelJsonPath);
+  const hasNowJson = existsSync(nowJsonPath);
+
+  // Check for conflicting vercel.json and now.json
+  if (hasVercelJson && hasNowJson) {
+    throw new ConflictingConfigFiles([vercelJsonPath, nowJsonPath]);
+  }
+
```
srintern, line 81: do we have this check anywhere else today?

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: updated

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: yes in certain places later, this is an additional failsafe

packages/cli/test/unit/util/compile-vercel-config.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    tmpDir = getNewTmpDir();
+    process.env.VERCEL_TS_CONFIG_ENABLED = '1';
+  });
+
+  afterEach(async () => {
+    delete process.env.VERCEL_TS_CONFIG_ENABLED;
+    await remove(tmpDir);
+  });
+
+  it('should compile vercel.ts to vercel.json', async () => {
```
jeffsee55, line 21: Can you add another test for `.mjs` at least? I believe that will fail on node 20 and perhaps pass on node 22 with the way this is doing `require(pathToFile)`, but would be good to know.

If it captures the failure, I think that's fine to move forward with this as is. But might be worthwhile to think about using `child_process.fork()`.

pranavkarthik10: Added a test, and modified it to use a child process instead. Verified environment variable access works without explicitly passing it in for both `.env` and `.env.local`

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -0,0 +1,224 @@
...
+      child.on('error', err => {
+        reject(err);
+      });
+
+      child.on('exit', code => {
+        if (code !== 0) {
+          reject(new Error(`Config loader exited with code ${code}`));
+        }
+      });
```
jeffsee55, line 181: It might be good to also just issue a timeout where we only wait for a message for a couple of seconds. I'm not sure what kinds of things could go wrong, but might as well be safe.

jeffsee55 (approved):
Non-blocking, looks good!

packages/cli/src/util/compile-vercel-config.ts
pranavkarthik10: added timeout for 10s

PR #14327, opened 2025-11-21
[python] uv workspaces support

ricardo-agz:
Adds support for uv workspaces in Python monorepos. Previously, when building an app whose root was not at the repo root (e.g. apps/python-app), we only installed dependencies from the app’s own pyproject.toml/uv export, so internal workspace packages (like packages/python-lib) were not installed into _vendor.

This PR reads the workspace pyproject.toml at repoRootPath, uses [tool.uv.workspace] and [tool.uv.sources] to discover workspace members, and installs any members that are listed in the app’s project.dependencies. For example, with:

* app at /apps/python-app
* internal package at packages/python-lib
* root-level `pyproject.toml` with uv workspaces:
```
[tool.uv.workspace]
members = [
    "packages/python-lib",
    "apps/python-app2",
]

[tool.uv.sources]
python-lib = { workspace = true }
```

```
turbo-with-python-uv/
├── apps/
│   └── python-app2/
│       ├── main.py
│       └── pyproject.toml
├── packages/
│   ├── python-lib/
│   │   ├── package.json
│   │   ├── pyproject.toml
│   │   └── src/
│   │       └── python_lib/
│   │           ├── __init__.py
│   └── tooling-config/
│       └── package.json
├── pyproject.toml
├── turbo.json
└── uv.lock
```


packages/python/src/uv-workspace.ts
```diff
@@ -0,0 +1,172 @@
...
+};
+
+function getDependencyName(spec: string): string | null {
+  // PEP 508 dependency spec starts with the package name, followed by optional
+  // version constraints / markers. We only need the leading "name" portion.
+  const match = spec.match(/^[A-Za-z0-9_.-]+/);
+  return match ? match[0] : null;
+}
+
+export async function installUvWorkspaceDependencies({
```
jeffsee55, line 40: I'm surprsised we have to do all of this work, why can't `uv` just do it?

ricardo-agz: The reason `uv` can't just handle this out of the box now is because of how we're installing dependencies: we use `uv export` + `uv pip install` into /_vendor inside /.vercel rather than letting uv create and manage a full virtual environment

Longer term, we could consider letting uv manage the environment directly (e.g. creating /.venv, running `uv sync`, and then packaging that env into the Lambda; we currently exclude /.venv), which would let uv do more of this natively

But there could also be benefits to keeping deps vendored (especially if we go forward with something like the Services proposal and someone wants different versions of a package in different services) so this probably requires a bit more thought

jeffsee55: > we could consider letting uv manage the environment directly (e.g. creating /.venv, running uv sync, and then packaging that env into the Lambda

non-blocking: Is is possible to let uv manage but override where they vendor to?

jeffsee55 (approved):

packages/python/src/uv-workspace.ts
ricardo-agz: actually there might be yea, I think we could do something like:
```bash
uv venv .vercel/output/python/.venv # create the venv inside .vercel
export UV_PROJECT_ENVIRONMENT=".vercel/output/python/.venv" # point uv to it
uv sync --directory apps/path-to-python-app # install deps 
```

ricardo-agz: I think it might be best to hold off on that for a later PR though since the scope might be a bit bigger since we'd be changing the way the whole install/vendoring logic works but I think it might be the way

PR #14356, opened 2025-11-25
[next] fix trailing slash handling for segment prefetch 

huozhi:
### Fix

This PR adds the trailing slash handling to the segments prefetch requests. When trailing slash is enabled, the segment prefretch is failing as 404 because the slash is not matched in the regexes like other routes.

Closes NXT-118

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -2279,7 +2279,11 @@ export async function serverBuild({
             prefetchSegmentSuffix
               ? [
                   {
-                    src: path.posix.join('/', entryDirectory, '/(?<path>.+)$'),
+                    src: path.posix.join(
+                      '/',
+                      entryDirectory,
+                      '/(?<path>.+?)(?:/)?$'
+                    ),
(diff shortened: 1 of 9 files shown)


packages/next/src/server-build.ts
```diff
@@ -2279,7 +2279,11 @@ export async function serverBuild({
             prefetchSegmentSuffix
               ? [
                   {
-                    src: path.posix.join('/', entryDirectory, '/(?<path>.+)$'),
+                    src: path.posix.join(
+                      '/',
+                      entryDirectory,
+                      '/(?<path>.+?)(?:/)?$'
```
timneutkens, line 2285: Will defer to @ztanner / @ijjk. One thing I'm wondering is if we should add this optional match for trailing `/` or instead we should know when trailingSlash is configured to include/exclude it.

ijjk: Always handling the trailing slash is consistent with what we do with other routes

ijjk (approved):

PR #14360, opened 2025-11-25
[CICD-DX] Standardize ls argument errors

onsclom:
Standardize errors using ls/list subcommands. Introduces `validateLsArgs` util and uses it in each instance.

https://linear.app/vercel/issue/FLOW-1399/standardize-vercel-group-ls-random-extra-arg-behavior


packages/cli/src/util/validate-ls-args.ts
```diff
@@ -0,0 +1,18 @@
+import chalk from 'chalk';
+import { getCommandName } from './pkg-name';
+import output from '../output-manager';
+
+export function validateLsArgs(
```
tknickman: Can you update so this function takes an object as the first param so we can use named arguments. Then uses like this:

```ts
 const validationResult = validateLsArgs(
    'integration list',
    parsedArguments.args,
    2,
    1,
    getCommandName('integration list [project]')
  );
```

Become much more readable:

```ts
  const validationResult = validateLsArgs({
    commandName: 'integration list',
    args: parsedArguments.args,
    maxArgs: 2,
    exitCode: 1,
    usageString: getCommandName('integration list [project]')
  })
```

packages/cli/src/commands/dns/ls.ts
```diff
@@ -30,6 +31,18 @@ export default async function ls(client: Client, argv: string[]) {
     return 1;
   }
   const { args, flags: opts } = parsedArgs;
+
+  const validationResult = validateLsArgs(
+    'dns ls [domain]',
+    args,
+    1,
+    1,
+    getCommandName('dns ls [domain]')
```
tknickman: It looks like we don't have to pass this (it's the same as the default)

packages/cli/src/commands/integration/list.ts
```diff
@@ -37,11 +39,15 @@ export async function list(client: Client) {
   // Note: the `--integration` flag is tracked later, after validating
   // whether the value is a known integration name or not.
 
-  if (parsedArguments.args.length > 2) {
-    output.error(
-      'Cannot specify more than one project at a time. Use `--all` to show all resources.'
-    );
-    return 1;
+  const validationResult = validateLsArgs(
+    'integration list',
```
tknickman: Should we just use `integration list [project]` here, and then we can omit the last arg and rely on the default?

tknickman (commented):
A few questions, but looks good!

tknickman (approved):

PR #14365, opened 2025-11-26
Block network io during introspection

jeffsee55:
Add logic for blocking network io during introspection. This is a basic approach, hopefully to kick off more discussion. 

Also adds a few tests around network IO during init.


brookemosby (approved):
LGTM

PR #14372, opened 2025-11-26
Fix Node.js proxy/middleware with basePath and matcher on Vercel

gaojude:
When using `proxy.ts` (or Node.js runtime middleware) with both `basePath` and a custom `matcher` config, the proxy wouldn't run on Vercel deployments even though it worked perfectly locally.

The root cause is a double basePath prefix issue. When Next.js builds a Node.js proxy/middleware, it generates `functions-config-manifest.json` with matchers that already have basePath baked into the regexp (e.g., `^\\/new(?:\\/...` for `basePath: '/new'`). However, when the Vercel builder processes these matchers through `getRouteMatchers()`, it doesn't pass a `page` property.

The `getRouteMatchers()` function has a check `if (info.page === '/')` that skips adding basePath - this was designed for edge middleware where the manifest entry has `page: '/'`. But since Node.js middleware doesn't pass `page`, the condition fails and basePath gets added again, resulting in routes like `/new/new/...` instead of `/new/...`.

The fix passes `page: '/'` when calling `getRouteMatchers()` for Node.js middleware, which tells it to skip the basePath addition since the matchers already have it. This matches the behavior for edge middleware where `page` is `/` (root middleware).

Fixes NEXT-4795


packages/next/src/utils.ts
```diff
@@ -3741,8 +3741,12 @@ export async function getNodeMiddleware({
     return null;
   }
   const routes: RouteWithSrc[] = [];
+  // Pass page: '/' to skip basePath addition in getRouteMatchers(), since
+  // functions-config-manifest matchers already have basePath baked in by
+  // Next.js build. This matches the behavior for edge middleware where
+  // page is '/' (root middleware) and basePath is not added.
   const routeMatchers = getRouteMatchers(
-    { matchers: middlewareFunctionConfig.matchers },
+    { matchers: middlewareFunctionConfig.matchers, page: '/' },
```
ijjk, line 3749: Can we add a test case for the described case?

ijjk (approved):

PR #14377, opened 2025-11-27
fix(nestjs): install builder in cli

tknickman:
Install Nestjs in CLI

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -47,6 +47,7 @@
     "@vercel/hono": "0.2.14",
     "@vercel/hydrogen": "1.3.2",
     "@vercel/next": "4.15.6",
+    "@vercel/nestjs": "workspace:*",
     "@vercel/node": "5.5.13",
     "@vercel/python": "6.1.0",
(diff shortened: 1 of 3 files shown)


ricardo-agz (approved):

packages/cli/package.json
```diff
@@ -47,6 +47,7 @@
     "@vercel/hono": "0.2.14",
     "@vercel/hydrogen": "1.3.2",
     "@vercel/next": "4.15.6",
+    "@vercel/nestjs": "workspace:*",
```
ofhouse, line 50: Shouldn't this point to the version number instead like the other builders?

tknickman: This should get converted to the real version on release, I think we do the others incorrectly actually, but let me double check.

tknickman: Yea this should work fine, spot checked against the others using workspace protocol and verified they're converted to the correct versions on release (pnpm pack / pnpm release does this automatically)

ofhouse (approved):

PR #14392, opened 2025-12-02
[CICD-DX] Don't prompt to `env pull` on newly created projects

codywong:
Newly created projects won't have environment variables. In these cases, we can streamline the process and avoid prompting the user to pull environment variables

fixes https://linear.app/vercel/issue/CIFEAT-469/dont-pull-environment-variables-on-vc-link-if-user-just-created-the

### DIFF
--- a/packages/cli/src/util/link/setup-and-link.ts
+++ b/packages/cli/src/util/link/setup-and-link.ts
@@ -245,7 +245,8 @@ export default async function setupAndLink(
       project.name,
       org.slug,
       successEmoji,
-      autoConfirm
+      autoConfirm,
+      false // don't prompt to pull env for newly created projects
     );
 
     await connectGitRepository(client, path, project, autoConfirm, org);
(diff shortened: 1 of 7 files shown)


codywong:
* **#14392** <a href="https://app.graphite.com/github/pr/vercel/vercel/14392?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14392?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -245,7 +245,8 @@ export default async function setupAndLink(
       project.name,
       org.slug,
       successEmoji,
-      autoConfirm
+      autoConfirm,
+      false // don't prompt to pull env for newly created projects
```
tknickman, line 249: We should refactor this in the future to use an object as the param so we can have more context on what this false connects to at the call site. But this looks good for now.

tknickman (approved):

PR #14399, opened 2025-12-02
[CICD-DX] Fix vc build to respect .vercelignore

onsclom:
Doing `vc build` locally, `.vercelignore` is incorrectly ignored. Static assets ignored in the `/public` folder would still get copied to `.vercel/output/static`.

https://linear.app/vercel/issue/FLOW-1324/vercel-build-should-respect-vercelignore-entries


packages/cli/src/util/build/write-build-result.ts
```diff
@@ -645,15 +650,59 @@ async function writeLambda(
...
+  dir: string,
+  staticRoot: string,
+  filter: (path: string) => boolean
+): Promise<void> {
+  const entries = await fs.readdir(dir);
+
+  await Promise.all(
+    entries.map(async entry => {
+      const entryPath = join(dir, entry);
+      const stat = await fs.stat(entryPath);
```
codywong, line 691: nit: can a builder return symlinks? I'm guessing not, but if they did, would lstat be preferred here?

codywong (approved):

PR #14407, opened 2025-12-04
[client] Add retry logic to deployment status polling

bgw:
This is an attempt to fix https://vercel.slack.com/archives/C09MK639NMN/p1764806810028989?thread_ts=1761256638.364529&cid=C09MK639NMN

We do a lot of deployments in parallel in the Next.js CI, and hit the rate limit very often:
https://github.com/vercel/next.js/actions/workflows/test_e2e_deploy_release.yml

This is a simple GET request (safe to retry), and the API returns an HTTP 429 in this case, so we should just use the `Retry-After` header. I also added a check for HTTP 5xx errors.

I used claude to help write some unit tests, though I cleaned up and simplified what claude generated.

The rate limit for this endpoint is pretty high (500/min), but I tested this manually by removing the normal `getPollingDelay` sleep at the end of the generator loop, and by adding a bunch of artificial parallel fetches to the code.

### DIFF
--- a/packages/client/src/check-deployment-status.ts
+++ b/packages/client/src/check-deployment-status.ts
@@ -38,7 +38,9 @@ export function parseRetryAfterMs(response: any): number | null {
     let retryAfterMs = Number(header) * 1000;
     if (Number.isNaN(retryAfterMs)) {
       let retryAfterDateMs = Date.parse(header);
-      if (!Number.isNaN(retryAfterDateMs)) {
+      if (Number.isNaN(retryAfterDateMs)) {
+        retryAfterMs = RETRY_DELAY_DEFAULT_MS;
+      } else {
         retryAfterMs = retryAfterDateMs - Date.now();
       }
     }
--- a/packages/client/src/check-deployment-status.ts
+++ b/packages/client/src/check-deployment-status.ts
@@ -27,6 +27,34 @@ const RETRY_DELAY_MAX_MS = 60_000;
 const RETRY_DELAY_MIN_MS = 5_000;
 const RETRY_DELAY_DEFAULT_MS = 5_000;
 
+export function parseRetryAfterMs(response: any): number | null {
+  // HTTP 429 (Too Many Requests) or 503 (Service Unavailable)
+  if (response.status === 429 || response.status === 503) {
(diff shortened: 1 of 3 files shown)


packages/client/src/check-deployment-status.ts
```diff
@@ -46,16 +52,51 @@ export async function* checkDeploymentStatus(
...
+
+      retriesLeft--;
+      if (retriesLeft == 0) {
+        break;
+      }
+
+      // HTTP 429: Too Many Requests
+      if (deploymentResponse.status === 429) {
+        let retryAfterMs =
+          parseInt(deploymentResponse.headers.get('Retry-After'), 10) * 1000;
```
eps1lon: `Retry-After` is also sometimes used on 503 Service Unavailable: https://httpwg.org/specs/rfc9110.html#field.retry-after

packages/client/src/check-deployment-status.ts
```diff
@@ -46,16 +52,51 @@ export async function* checkDeploymentStatus(
...
+        if (Number.isNaN(retryAfterMs)) {
```
eps1lon: That should cover when `retry-after` is an HTTP date. Still, JS implicit type conversion is funky so maybe be extra careful and only attempt to parse if the header only contains digits (https://httpwg.org/specs/rfc9110.html#rule.delay-seconds)?

packages/client/src/check-deployment-status.ts
```diff
@@ -46,16 +52,51 @@ export async function* checkDeploymentStatus(
...
+          retryAfterMs = RETRY_DELAY_DEFAULT_MS;
+        }
+        retryAfterMs = Math.min(
+          RETRY_DELAY_MAX_MS,
+          Math.max(RETRY_DELAY_MIN_MS, retryAfterMs)
+        );
+        await sleep(retryAfterMs);
```
lukesandberg, line 110:
```suggestion
       debug(`Rate limited by the server while querying deployment status, retrying after ${retryAfterMs}ms`)
        await sleep(retryAfterMs);
```

something like that?

ijjk (approved):

PR #14410, opened 2025-12-04
[next] Type config in required-server-files

mischnic:
Basically no code changes, adds types instead of `any` for `requiredServerFiles.config`

Came here from https://github.com/vercel/next.js/pull/86812


.changeset/fifty-lizards-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+Properly type next config instead of using any
```
bgw:
```suggestion
Properly type next config in `NextRequiredServerFilesManifest` instead of using any
```

Is this a public API? If it's not, this doesn't need a changelog entry.

bgw (approved):

.changeset/fifty-lizards-smoke.md
mischnic: No, but CI fails if there is no changelog 🤷

ijjk: You can add an empty one with `pnpm changeset --empty`

timneutkens (approved):

PR #14415, opened 2025-12-05
[python] update dependency installation

ricardo-agz:
Standardizes the dependency installation for the different supported formats to all use `uv sync` to install dependencies inside `.vercel/python/.venv`. This means that for:
* requirements.txt -> we create a `pyproject.toml` and `uv.lock`
* Pipfile/lock -> export to a temp requirements, then create a `pyproject.toml` and `uv.lock`
* for all: run `uv sync`

`uv sync` lets us handle a few things automatically that we otherwise were hand-rolling/would need to hand-roll:
* pruning unused dependencies in _vendor/.venv (this will let us do caching with `prepareCache()` in a follow-up PR, otherwise stale dependencies would accumulate in the cache even when removing them from `requirements.txt`)
* deterministic/idempotent installs when running the builder multiple times in sequence such as for python projects in the /api folder (since `uv sync` will no-op when all the dependencies are already installed in `.vercel/python/.venv`)
* support for uv workspaces (replacing our previous custom logic to enable it)


packages/python/src/index.ts
```diff
@@ -276,191 +238,59 @@ export const build: BuildV3 = async ({
...
   }
 
-  await installRequirement({
+  // Runtime framework dependencies are managed via the uv project so that the
+  // lockfile is the single source of truth for all installed packages. These
+  // are intentionally unpinned so they can resolve alongside user-declared
+  // dependencies (for example, modern Flask versions that require newer
+  // Werkzeug releases).
+  const runtimeDependencies =
+    framework === 'flask' ? ['werkzeug'] : ['werkzeug', 'uvicorn'];
```
1st1: Should these be pinned?

packages/python/src/install.ts
```diff
@@ -65,11 +73,433 @@ async function areRequirementsInstalled(
...
+    lines.push('dependencies = [');
+    for (const dep of dependencies) {
+      lines.push(`  "${dep}",`);
+    }
+    lines.push(']');
+  } else {
+    lines.push('dependencies = []');
+  }
+
+  lines.push('');
```
1st1: I'd also add a "private" classifier just for good hygiene https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers

packages/python/src/index.ts
ricardo-agz: Good point, yes they should, I set werkzeug to >= 1.0.1 which is what it was previously pinned to, uvicorn to >= 0.24, and updated one of the e2e test fixtures to pin those 2 versions respectively

brookemosby (approved):
LGTM

jeffsee55 (approved):

PR #14417, opened 2025-12-05
Fix example api url

onsclom:
https://now-example-files.zeit.sh/ has deployment protection so it causes:

```
vc init
Vercel CLI 48.12.1
Error: Response Error (401)
```

### DIFF
--- a/packages/ruby/package.json
+++ b/packages/ruby/package.json
@@ -17,7 +17,7 @@
   },
   "scripts": {
     "build": "node ../../utils/build-builder.mjs",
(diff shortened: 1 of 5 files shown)


ricardo-agz (approved):

tknickman (approved):

packages/ruby/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "scripts": {
     "build": "node ../../utils/build-builder.mjs",
-    "test": "jest --reporters=default --reporters=jest-junit --env node --verbose --runInBand --bail",
```
onsclom, line 20: Failing for unrelated reasons and blocking this PR

PR #14418, opened 2025-12-05
[python] fix custom install command

ricardo-agz:
Correctly wires the project install command so that it gets executed in the correct virtual environment so dependencies can be vendored correctly. Also overrides the default install flow


packages/python/src/index.ts
```diff
@@ -240,59 +248,82 @@ export const build: BuildV3 = async ({
...
-      }`
-    );
-  }
+  // If a dashboard install command is configured for FastAPI/Flask, treat it as
+  // an override for the default dependency installation: run the command inside
+  // the build virtualenv
+  const hasCustomInstallCommand =
+    (framework === 'fastapi' || framework === 'flask') &&
+    typeof projectInstallCommand === 'string' &&
+    projectInstallCommand.length > 0;
```
codywong: nit: `!!projectInstallCommand`or `!== ""` could be a bit more canonical vs  `.length > 0`

codywong (approved):

jeffsee55 (approved):

PR #14429, opened 2025-12-07
fix: check used functions for outputDirectory

hyoban:
Because we also checked here.

https://github.com/vercel/vercel/blob/c4176593cc17a729568932ea78bbabb39d89a6c1/packages/hono/src/build.ts#L80-L100


hyoban:
Just a friending ping @agadzik @chloetedder

PR #14434, opened 2025-12-08
feat(functions): Add `continent` to `geolocation()` return

NWylynko:
Updates `geolocation()` in the functions package to include the continent of the users request. This is from the vercel reference docs (https://vercel.com/docs/headers/request-headers#x-vercel-ip-continent).


NWylynko:
I've set the changeset as a `minor` but if this is just a `patch` I'm happy to change it.

PR #14437, opened 2025-12-08
[cli] Add non-interactive mode for adding redirects and fix a few other small issues

mknichel:
This PR primarily adds the ability to specify options to `vc redirects add` for status code, case sensitivity, and preserving query params so it doesn't need to be interactive. 

It also applies two other small changes:

- Changes `--staged` to `--staging`
- Shows you the diff of the redirects compared to prod when you use `ls --staging`

Screenshot 2025-12-08 at 11 37 35 AM

### DIFF
--- a/.changeset/short-hairs-deny.md
+++ b/.changeset/short-hairs-deny.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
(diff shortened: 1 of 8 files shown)


bcjroberts (approved):

mknichel:
@onsclom do you think you could take a look at this PR?

onsclom (approved):

PR #14443, opened 2025-12-09
Improve handling of 429 error codes in the Vercel CLI

lukesandberg:
* Handle `date` formatted `Retry-After` headers, this is similar to #14407 but on the CLI side
    * it is unfortunate to have two copies of parsing logic, but there isn't a clear place to share the logic and it is pretty simple
* Change the 'retry' loop logic in client.ts and `util/index.ts` to always respect and retry when there is a `Retry-After` header or the error code is 429.  Currently they always abort on 400 errors
* rename the `retryAfter` properties to `retryAfterMs` to be clearer

This should improve the CLIs ability to recover from temporary rate limit conditions.  This is important for the Next.js CI system where we generate a lot of deployments as part of release validation:  e.g. https://github.com/vercel/next.js/actions/runs/20041033845/job/57474897550

### DIFF
--- a/.changeset/popular-cougars-heal.md
+++ b/.changeset/popular-cougars-heal.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Improve handling of 429 error codes and Retry-After headers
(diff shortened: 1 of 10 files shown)


ijjk (approved):

lukesandberg:
### Merge activity

* **Dec 10, 12:30 AM UTC**: @lukesandberg merged this pull request with [Graphite](https://app.graphite.com/github/pr/vercel/vercel/14443).

PR #14449, opened 2025-12-10
OIDC error refinement

alice-wondered:
Previously I attempted to handle error messages as if TS was golang, stacking context between layers. This approach led to poor end user messaging for actionable errors.

### DIFF
--- a/packages/oidc/src/get-vercel-oidc-token.ts
+++ b/packages/oidc/src/get-vercel-oidc-token.ts
@@ -49,10 +49,11 @@ export async function getVercelOidcToken(): Promise<string> {
       token = getVercelOidcTokenSync();
     }
   } catch (error) {
-    if (err?.message && error instanceof Error) {
-      error.message = `${err.message}\n${error.message}`;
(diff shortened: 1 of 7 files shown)


aayush-kapoor (approved):

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -49,10 +49,14 @@ export async function getVercelOidcToken(): Promise<string> {
...
-    if (err?.message && error instanceof Error) {
-      error.message = `${err.message}\n${error.message}`;
+    let message = err instanceof Error ? err.message : '';
+    if (error instanceof Error) {
+      message = `${message}\n${error.message}`;
     }
-    throw new VercelOidcTokenError(`Failed to refresh OIDC token`, error);
+    if (message) {
+      throw new VercelOidcTokenError(message);
+    }
```
gr2m: @cramforce isn't it better to set `.cause = error` to retain all information from the thrown error?

PR #14402, opened 2025-12-03
[CICD-DX] Improve env pull with conditional quoting of environment values

codywong:
# Improve Environment Variable Formatting in `vercel env pull`

This PR enhances how environment variables are formatted when using `vercel env pull` to create `.env` files:

- Added smart quoting for environment variables:
    - Values with spaces or starting with `#` are now properly quoted
    - Double quotes within values are escaped when necessary
    - Simple values remain unquoted for better readability
- Extracted formatting logic into a dedicated `formatEnvValue()` utility function with comprehensive test coverage
- Fixed newline handling to properly escape `\n` and `\r` characters

This change ensures that environment variables are formatted in a way that's compatible with standard `.env` file parsers while maintaining readability.

fixes [FLOW-879: `vercel env pull` cannot properly handle value that contains `"`​](https://linear.app/vercel/issue/FLOW-879/vercel-env-pull-cannot-properly-handle-value-that-contains)

PR #14456, opened 2025-12-11
[CICD-DX] Preserve JSON objects/arrays unquoted in env values

codywong:
This PR enhances the `formatEnvValue` function to better handle JSON objects and arrays in environment variables. JSON values are now preserved as-is without unnecessary quoting or escaping, making them more readable and usable in configuration files.

A follow-up to handle https://github.com/vercel/vercel/pull/14402#issuecomment-3642460884. 

If we still encounter users running into issues with env parsing patterns, we may need to move to defining env-pull quoting rules/strategy into a project's config or similar.

ref https://linear.app/vercel/issue/FLOW-879/vercel-env-pull-cannot-properly-handle-value-that-contains


codywong:
* **#14456** <a href="https://app.graphite.com/github/pr/vercel/vercel/14456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14456?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

onsclom (approved):

bryandowning:
@onsclom @codywong 

This change appears to have broken some of our multiline JSON environment variables.

**Value stored in Vercel**
```
{
  "a": "some-value",
  "b": "another-value"
}
```

Then when doing a `vercel pull`, the output in the .env.local file looks like this:

**vercel v48 output**
`NEXT_PUBLIC_TEST_JSON="{\n  "a": "some-value",\n  "b": "another-value"\n}"`

**vercel v50 output**
`NEXT_PUBLIC_TEST_JSON="{\n  \"a\": \"some-value\",\n  \"b\": \"another-value\"\n}"`

Both of the above values parse correctly, but when `vercel build` runs with the v50 output and inlines the values into the chunks, the output looks like this:

```
'{\n  \\"a\\": \\"some-value\\",\n  \\"b\\": \\"another-value\\"\n}'
```

...which fails in `JSON.parse`.

PR #14462, opened 2025-12-12
[cli] update install flow for vercel.ts to ensure all packages are installed pre-execution

pranavkarthik10:
For `vercel.ts` we want to make sure the packages used in `vercel.ts` are installed and available before it's executed. For this, we've added an optional install step pre-execution only when the feature flag is enabled and the config file is a `vercel.ts` file.

### DIFF
--- a/.changeset/rude-humans-count.md
+++ b/.changeset/rude-humans-count.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Run install command earlier if using vercel.ts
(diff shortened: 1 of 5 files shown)


pranavkarthik10:
In it's current state for projects with `vercel.ts`, it runs install twice. This is feature flagged and will only affect projects with `vercel.ts` in them. Will fast follow with an optimization for that.

onsclom (approved):

PR #14464, opened 2025-12-12
[backends] Override tsconfig settings that affect typecheck 

jeffsee55:
Some settings like `"moduleResolution": "nodenext"` will cause the typecheck to error when a relative path import is missing. But since we're using rolldown for the actual compilation that's ok. This overrides those settings to be more permissive.


packages/backends/test/fixtures/09-hono-ts-nodenext/index.ts
```diff
@@ -0,0 +1,16 @@
+import { Hono } from 'hono'
+// this error is expected
+import { echo } from './lib/echo'
+// this error is expected, importing .ts is ok
+import { echo2 } from './lib/echo-2.ts'
```
mehulkar, line 5: what error are the comments here talking about? are there typecheck errors on these lines that you are referring to? I understand we can't add `@ts-expected-error` since that's the point of this test, but then how does this pass CI? do we not typecheck test fixtures?

mehulkar (approved):
some questions/comments

packages/backends/test/fixtures/09-hono-ts-nodenext/index.ts
jeffsee55: Yeah this should just use expect-error, will fix. 

And yes I think we ignore test files for typechecks

jeffsee55: EDIT: wait, I want this to be in invalid `.ts`. The `tsc` flag overrides are necessary for this to pass typechecking

Since our builder will resolve these imports regardless of whether or not their extension is present and will even resolve `.ts` extensions it doesn't make sense to fail on these.

PR #14466, opened 2025-12-12
[cli] Adding Vercel CLI support for custom error pages

raochand:
This PR adds CLI support for an upcoming feature on Vercel to customize pages for platform errors (e.g., 5xx errors from function timeouts).

Users can configure custom error pages in vercel.json:
```
{
  "customErrorPage": "_errors/error.html"
}
```
Or with separate pages for different error types:

```
{
  "customErrorPage": {
    "default5xx": "_errors/server-error.html",
    "default4xx": "_errors/not-found.html"
  }
}
```

### DIFF
--- a/packages/cli/src/util/validate-config.ts
+++ b/packages/cli/src/util/validate-config.ts
@@ -155,6 +155,26 @@ const cronsSchema = {
   },
 };
 
+const customErrorPageSchema = {
+  oneOf: [
+    { type: 'string', minLength: 1 },
+    {
+      type: 'object',
+      additionalProperties: false,
+      properties: {
+        default5xx: {
+          type: 'string',
+          minLength: 1,
+        },
+        default4xx: {
+          type: 'string',
+          minLength: 1,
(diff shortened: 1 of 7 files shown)


packages/cli/src/util/validate-config.ts
```diff
@@ -155,6 +155,27 @@ const cronsSchema = {
   },
 };
 
+const customErrorPageSchema = {
```
pranavkarthik10, line 158: Based on the other schemas, this could go in packages/routing-utils/src/schemas.ts, and then be imported here. Makes sense if this isn't considered part of routing-utils though.

agadzik (approved):

packages/cli/src/util/validate-config.ts
raochand: Chatted with pranav.. we will keep it here as there are examples of other schemas in here.

jaygengelbach (approved):

PR #14467, opened 2025-12-12
[python] uv workspaces fix - don't throw on missing `uv.lock`

ricardo-agz:
If you have a project with multiple `pyproject.toml` using uv workspaces and the `uv.lock` file is at the repo root instead of the app root, it currently throws the following error:
```
vc build
Vercel CLI 50.0.0
WARN! Build not running on Vercel. System environment variables will not be available.
Using Python 3.12 from pyproject.toml
Using uv at "/opt/homebrew/bin/uv"
Installing required dependencies from pyproject.toml...
Error: Expected "uv.lock" to exist in "/Users/ricardo/Documents/code/scratchpad/turbo-with-python-uv/apps/python-app2" after preparing uv project
```

This PR resolves the lockfile even if it is at the repo root or outside the project root

### DIFF
--- a/.changeset/many-feet-divide.md
+++ b/.changeset/many-feet-divide.md
@@ -1,0 +1,5 @@
+---
+"@vercel/python": patch
+---
+
+[python] uv workspaces fix don't throw on missing `uv.lock`
--- a/packages/python/src/index.ts
+++ b/packages/python/src/index.ts
@@ -69,6 +69,7 @@ export async function downloadFilesInWorkPath({
 
 export const build: BuildV3 = async ({
   workPath,
+  repoRootPath,
   files: originalFiles,
   entrypoint,
   meta = {},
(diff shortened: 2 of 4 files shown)


jeffsee55 (approved):
What is supposed to happen in an ideal world? Is there supposed to be a `uv.lock` for every workspace or a single one at the root?

ricardo-agz:
> What is supposed to happen in an ideal world? Is there supposed to be a `uv.lock` for every workspace or a single one at the root?

@jeffsee55  You only have one `uv.lock` for the whole project, [from the docs](https://docs.astral.sh/uv/concepts/projects/workspaces/#when-not-to-use-workspaces): 
"In a workspace, each package defines its own pyproject.toml, but the workspace shares a single lockfile, ensuring that the workspace operates with a consistent set of dependencies."

PR #14493, opened 2025-12-17
[python] only create api builders for `.py` files that export an app or handler

ricardo-agz:
A helper file in /api that does not export an app or http handler should *not* create a lambda. This PR adds a helper to build-utils to detect if a file contains a top-level app or handler

### DIFF
--- a/packages/fs-detectors/src/detect-builders.ts
+++ b/packages/fs-detectors/src/detect-builders.ts
@@ -1,6 +1,6 @@
 import minimatch from 'minimatch';
 import { valid as validSemver } from 'semver';
-import { parse as parsePath, extname } from 'path';
+import { parse as parsePath, extname, join } from 'path';
 import type { Route, RouteWithSrc } from '@vercel/routing-utils';
 import frameworkList, { Framework } from '@vercel/frameworks';
 import type {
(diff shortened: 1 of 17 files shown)


smaeda-ks (approved):

packages/fs-detectors/src/detect-builders.ts
```diff
@@ -180,7 +182,16 @@ export async function detectBuilders(
 
   // API
   for (const fileName of sortedFiles) {
-    const apiBuilder = maybeGetApiBuilder(fileName, apiMatches, options);
+    let apiBuilder = maybeGetApiBuilder(fileName, apiMatches, options);
+
+    // For Python files, verify they are valid entrypoints before creating a builder
+    if (apiBuilder && fileName.endsWith('.py') && options.workPath) {
```
craigandrews: suggestion: given `maybeGetApiBuilder` is already a conditional function, this block should be moved inside rather than splitting the logic.

ricardo-agz: Yea much cleaner that way, moved it inside `maybeGetApiBuilder`

onsclom (approved):

PR #14494, opened 2025-12-17
Allow `vercel.json` function configuration on output function paths

TooTallNate:
Improve function deduplication by considering function configuration when creating symlinks.

### What changed?

- Enhanced the function deduplication logic to consider function configuration when determining if functions are identical
- Added tracking of function configuration alongside paths in the `existingFunctions` map
- Introduced a new `ExistingFunctionEntry` interface to store both config and path information
- Modified `writeFunctionSymlink` to only create symlinks when both the function and its configuration match
- Removed the warning for unused functions in the `checkUnusedFunctions` function
- Added function configuration retrieval for Lambda and Edge functions in the build result writing process

### How to test?

1. Create a project with multiple serverless functions that have identical code but different configurations
2. Run a build and verify that functions with different configurations are not deduplicated via symlinks
3. Verify that functions with identical code and configurations are still properly deduplicated

### Why make this change?

Previously, the system would deduplicate functions based solely on the function instance, creating symlinks even when functions had different configurations. This could lead to incorrect behavior since functions with different configurations (like memory size, timeout, etc.) should be treated as distinct entities even if their code is identical. This change ensures that only truly identical functions (same code and configuration) are deduplicated, maintaining the correct behavior for each function endpoint.

### DIFF
--- a/.changeset/blue-dingos-grin.md
+++ b/.changeset/blue-dingos-grin.md
@@ -1,0 +1,5 @@
+---
+'@vercel/fs-detectors': minor
+---
+
+Remove "unused_function" check when collecting Builders list
--- a/.changeset/great-kids-perform.md
+++ b/.changeset/great-kids-perform.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Allow `vercel.json` function configuration on output function paths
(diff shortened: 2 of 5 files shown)


TooTallNate:
* **#14494** <a href="https://app.graphite.com/github/pr/vercel/vercel/14494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14494?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aria157 (approved):
thankyou i appreciate this

jeffsee55 (approved):
What's the context for removing the `checkUnusedFunctions` bit?

PR #14497, opened 2025-12-18
[prebuilt-skew-protection] feat: adding user configured deployment id to build output type

brookemosby:
brookemosby/flow-5149-add-user-configured-deployment-id-to-build-output-type
Needed so we can later store on deployment document.


.changeset/curly-eggs-smell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/build-utils': minor
+'vercel': minor
```
styfle: I would call this a patch since its not a new feature, just new types

tknickman (approved):

PR #14502, opened 2025-12-18
[python] consolidate python api builders into single lambda func

ricardo-agz:
Right now we're creating an individual lambda function for every `.py` file that has an asgi/wsgi app or http handler when using the `/api` directory. This is unnecessary since every python entrypoint pulls in the same set of dependencies.
This PR consolidates each python entrypoint in `/api` into a single* /api/py function so we only run a single consolidated Python builder

It accomplishes this by creating a single Starlette "meta-app" that mounts each entrypoint at its respective path

*When `rewrites` are used in `vercel.json`, if a Python file is the target of a rewrite, we cannot consolidate that file it as this would break current behavior. Other entrypoints that are not the target of a rewrite can still be consolidated

This is only applicable to non-zero-config uses of Python on Vercel, e.g. when being used in the `/api` folder in a Next.js app or with a non-Flask/FastAPI framework

**Note:** This *does* introduce a behavior change, since every python api entrypoint is consolidated into a single lambda, we no longer support each individual `.py` file having its own set of dependencies (although this is not something we want to support), but this is why this PR changes `05-cowsay-pipfile-local` fixture


packages/python/src/index.ts
```diff
@@ -48,6 +48,9 @@ import {
 
 export const version = 3;
 
+// Additional dependencies required for consolidated Python API mode
+const CONSOLIDATED_API_DEPENDENCIES = ['starlette>=0.27.0', 'asgiref>=3.0.0'];
```
elprans: Probably worth pinning major ver.

elprans: Oh and also see if you can lower too in case user code depends on older stuff.

PR #14504, opened 2025-12-18
[cli] remove early compilation in get-config for vercel.ts projects

pranavkarthik10:
- the reason we were compiling the vercel config in the first place was because we were reading vercel.json in that place (cli entrypoint)
- the only property extracted by this JSON parse–turned–TypeScript compilation was to extract the scope property, which is a pre-vc link behavior, not even in the schema anymore. super legacy
- so we can just skip it
- we now remove compilation at that step (get-config)
- verified non-monorepos work

### DIFF
--- a/packages/cli/src/util/get-config.ts
+++ b/packages/cli/src/util/get-config.ts
@@ -3,6 +3,7 @@ import { fileNameSymbol } from '@vercel/client';
 import {
   CantParseJSONFile,
   CantFindConfig,
+  ConflictingConfigFiles,
   WorkingDirectoryDoesNotExist,
 } from './errors-ts';
 import humanizePath from './humanize-path';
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/get-config.ts
```diff
@@ -56,6 +61,13 @@ export default async function getConfig(
   const nowFilePath = path.resolve(localPath, 'now.json');
 
   // Then try with `vercel.ts`, `vercel.json` or `now.json` in the same directory
+  const sourceConfigFile = await findSourceVercelConfigFile(localPath);
+  if (sourceConfigFile) {
+    await initCorepack({ repoRootPath: localPath });
+    output.log(`Installing dependencies before config compilation...`);
+    await runNpmInstall(localPath, [], { env: process.env });
```
MatthewStanciu: note: unlike the previous place where we installed early, here we don't yet have access to `installCommand` :(

we figured this was okay because the install later uses the right package manager. unsure if this causes conflicts.

i wonder if it makes more sense to statically analyze here instead of build?

MatthewStanciu: resolved; we changed the purpose of the pr to remove this compilation step entirely. see: https://vercel.slack.com/archives/C09SG01P39C/p1766099939924419?thread_ts=1766098025.776669&cid=C09SG01P39C

tknickman (approved):
Looks good

onsclom (approved):

PR #14517, opened 2025-12-23
[config] Merge computed Rewrite/Redirect types to Route if they are mixed with Route

MatthewStanciu:
@pbzona was testing vercel.ts and ran into a DX issue where computed rewrites with transforms correctly get turned into the Route type, but if one of them doesn’t have transforms it stays Rewrite, and you can’t mix both. See [Slack thread](https://vercel.slack.com/archives/C09SG01P39C/p1766440463591849) for a code snippet.

Technically this is expected behavior and these types are generated correctly, but it’s a DX issue because this is presumably a common use case that will burn people unnecessarily.

This PR:
- converts all rewrites to routes if they’re computed
- does not if routes and redirects/rewrites are both explicitly present (so it will still throw a schema validation error)

This should be fine because it’s following expected behavior for this pattern. But if we merge this we should add it to the docs too.

### DIFF
--- a/packages/config/src/cli.test.ts
+++ b/packages/config/src/cli.test.ts
@@ -1,0 +1,135 @@
+import { describe, it, expect } from 'vitest';
+import { normalizeConfig } from './cli';
+
+describe('normalizeConfig', () => {
+  it('should normalize rewrites array when it contains mixed Rewrite and Route formats', () => {
+    const config = {
+      rewrites: [
+        { source: '/simple', destination: '/dest' },
+        { src: '/complex', dest: '/dest', transforms: [] },
+      ],
+    };
+
+    const result = normalizeConfig(config);
+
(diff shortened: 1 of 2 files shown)


pranavkarthik10 (approved):
Yep, this is a thing only in the CLI commands. In practice, similar to `vercel.json`, users would have to set the top level property to routes. The solution for the future is to auto-convert everything to routes or support transforms and other properties in rewrites.

MatthewStanciu:
The Vercel CLI is actually where we want to do this. I moved this PR's logic to https://github.com/vercel/vercel/pull/14518.

MatthewStanciu:
I also want this logic in here. Currently unsure how to share the code if it lives primarily in CLI. Reopening for now with duplicate logic.

PR #14530, opened 2026-01-02
[python] set `UV_PYTHON` for Python installs

ricardo-agz:
We currently only support Python 3.12, but uv installs dependencies for the python version specified in `.python-version`, meaning that if a project includes this file and it is set to something other than 3.12, dependencies won't be found at runtime.
https://docs.astral.sh/uv/reference/environment/#uv_python


elprans:
@ricardo-agz , we might want to also consider setting `UV_PYTHON_PREFERENCE` to `system` in the build-container to make sure we are always building packages against a runtime build that is known (as opposed to uv-managed).

Issue #14541, opened 2026-01-05
OIDC: implement Vercel CLI token refresh


gr2m:
The `oidc` package ([/packages/oidc](https://github.com/vercel/vercel/tree/main/packages/oidc)) is the reason why `ai` just works™️ in local development as long as the current folder is linked to a vercel project.

It retrieves the `vercel` CLI token here:
https://github.com/vercel/vercel/blob/b9eae1d7c9f618355c6179eb58afb3c54318a046/packages/oidc/src/token-util.ts#L15-L31

And then uses it to retrieve an OIDC token here:
https://github.com/vercel/vercel/blob/b9eae1d7c9f618355c6179eb58afb3c54318a046/packages/oidc/src/token-util.ts#L37-L57

However, the `vercel` CLI token expires after 8h, and the `oidc` package does currently not check for expiration / does not implement a token refresh. 

The `vercel` CLI token is stored in a file like `/Users/<user>/Library/Application Support/com.vercel.cli/auth.json` (path depends on OS). The contents look like this

```
{
  "token": "vca_...",
  "expiresAt": 1767659841,
  "refreshToken": "vcr_..."
}
```

The logic that needs to be implemented is

1. Check for `expiresAt`
2. If expired, refresh `token` using `refreshToken` (for reference, here is the `vercel` CLI implementation: https://github.com/vercel/vercel/blob/d69becaa795a2ae53dc1442b2973756be15d3423/packages/cli/src/util/client.ts#L183-L185)
3. Update the `auth.json` file with the new `token`, `expiresAt`, and new `refreshToken`

alice-wondered:
Resolved by #14543

PR #14543, opened 2026-01-05
[OIDC] Refresh the CLI Auth Token When Refreshing OIDC

alice-wondered:
this PR solves a bug where it was possible to accidentally use an expired CLI auth token when refreshing the OIDC token locally

### DIFF
--- a/packages/oidc/src/oauth.ts
+++ b/packages/oidc/src/oauth.ts
@@ -1,0 +1,102 @@
+import { hostname, platform, arch } from 'os';
+
+const VERCEL_ISSUER = 'https://vercel.com';
+const VERCEL_CLI_CLIENT_ID = 'cl_HYyOPBNtFMfHhaUn9L4QPfTZz6TP47bp';
+
+// Simplified user agent for OIDC package
+const userAgent = `@vercel/oidc node-${process.version} ${platform()} (${arch()}) ${hostname()}`;
+
+export interface TokenSet {
+  access_token: string;
+  token_type: 'Bearer';
+  expires_in: number;
+  refresh_token?: string;
+  scope?: string;
+}
+
+/**
+ * Cached authorization server metadata
+ */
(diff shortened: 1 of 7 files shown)


alice-wondered:
Resolves #14541

packages/oidc/src/oauth.ts
```diff
@@ -0,0 +1,102 @@
...
+    }),
+  });
+}
+
+/**
+ * Process the token response and extract the token set
+ */
+export async function processTokenResponse(
+  response: Response
+): Promise<[Error] | [null, TokenSet]> {
```
gr2m, line 76: why the `[error, data]` response type, why not just throwing an error? We only use `processTokenResponse` in one place and can wrap it in a try/catch block

packages/oidc/src/token-util.test.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('should return token if valid and not expired', async () => {
+    const validToken = {
+      token: 'valid-access-token',
+      refreshToken: 'refresh-token',
+      expiresAt: Math.floor(Date.now() / 1000) + 3600, // expires in 1 hour
```
gr2m, line 25: preferable we would mock date/time for deterministic tests
https://vitest.dev/guide/mocking/dates.html

gr2m (approved):
code changes look good, left a few remarks but nothing critical. 

I did test it locally by building the changes and installing the `@vercel/oidc` pkg locally, and it worked. The `auth.json` file was updated with the new tokens and expiration correctly, and the OIDC tocken was updated successfully

packages/oidc/src/oauth.ts
alice-wondered: yeah agreed, sometimes claude just does things. No idea where it got this idea from

PR #14549, opened 2026-01-06
[python] `getPythonVersion()` in build-utils

ricardo-agz:
* Moves the python version detection to build utils so we can use it in the build container
* Adds support for reading python version from `.python-version` files

### DIFF
--- a/packages/build-utils/src/fs/python-version.ts
+++ b/packages/build-utils/src/fs/python-version.ts
@@ -1,0 +1,375 @@
+import { statSync, promises as fs } from 'fs';
+import { join } from 'path';
+import { NowBuildError } from '../errors';
+import { PythonVersion } from '../types';
+import debug from '../debug';
+
+export type PythonVersionMajorMinor = `${number}.${number}`;
+
+/**
+ * Reads the Python version from a `.python-version` file in the given directory.
+ */
+export async function readPythonVersionFile(
+  workPath: string
+): Promise<string | undefined> {
+  try {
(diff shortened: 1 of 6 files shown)


btwitsPratyush:
Nice improvement, moving Python version detection into build-utils makes a lot of sense.

A couple of clarifications:-
- If multiple sources are present (e.g. `.python-version`, env var, default), what is the intended precedence order?
- How do we handle an invalid or malformed `.python-version` file — do we fail fast or fall back?

Also, are there tests covering the `.python-version` path specifically? Might be useful to prevent regressions later.

packages/build-utils/src/fs/python-version.ts
```diff
@@ -0,0 +1,375 @@
...
+  version: string;
+  source: '.python-version' | 'pyproject.toml';
+}
+
+/**
+ * Detects the Python version
+ * workPath/.python-version > workPath/pyproject.toml >
+ * rootDirectoryPath/.python-version > rootDirectoryPath/pyproject.toml
+ */
+export async function detectPythonVersion(
```
btwitsPratyush, line 67: Quick clarification here:-

- I see the precedence is `.python-version` → `pyproject.toml` (workPath first, then root). Is this the intended order going forward?
- For `.python-version`, invalid values throw a `BUILD_UTILS_PYTHON_VERSION_INVALID` error — just confirming this is an intentional fail-fast behavior rather than falling back to defaults.

Looks solid overall 👍

elprans (approved):

PR #14552, opened 2026-01-07
fix(node): quote paths with spaces in NODE_OPTIONS

dferber90:
When forking dev servers, paths containing spaces (like `/Users/username/Library/Application Support`) were not properly quoted in NODE_OPTIONS, causing Node.js to fail with "Cannot find module" errors.

This adds quotes around the --require path arguments to ensure paths with spaces are handled correctly.

Fixes issue with `vc dev` on macOS when using Hono or other frameworks that require transpilation.


AndyBitz (dismissed):

tknickman (dismissed):

LukeSheard (approved):

TooTallNate (approved):

dferber90:
Moved back to draft state as this has issues on Windows. Trying a cursor agent PR here https://github.com/vercel/vercel/pull/14563

PR #14562, opened 2026-01-08
[python-analysis] Add @vercel/python-analysis package

elprans:
We currently have Python analyses spread and often duplicated between
`@vercel/fs-detectors`, `@vercel/build-utils` and `@vercel/python`.
This is suboptimal.

The new package is intended to consolidate all code necessary for static
analysis of Python project metadata and code.  This initial version
implements essential discovery of Python package manifests
(`pyproject.toml`, `Pipfile` and text requirements files) as well as
configuration such as `uv`'s `.python-version`.  Notably, we attempt to
parse and reason about data in manifests and configuration in order to
expose a normalized view of the project to the callers.

This commit introduces two useful entrypoints:

* `discoverPythonPackage` -- takes `entrypointDir` and `rootDir`, does
  discovery and manifest normalization magic and returns a
  `PythonPackage` object containing project and workspace manifests,
  as well as configuration and resolved Python runtime constraints..

* `selectPython` -- takes an array of Python constraints (as may be
  returned by `discoverPythonPackage` above, an array of available
  Python runtime builds and attempts to resolve the best runtime build
  suitable for the package (this is basically equivalent to
  `uv python find`).

This only adds new code and does not attempt to wire it in yet.


packages/python-analysis/src/manifest/types/python-specifiers.ts
```diff
@@ -0,0 +1,201 @@
...
+  },
+
+  toStringPretty(impl: PythonImplementation): string {
+    switch (impl) {
+      case 'cpython':
+        return 'CPython';
+      case 'pypy':
+        return 'PyPy';
+      case 'pyodide':
+        return 'PyIodide';
```
ricardo-agz: tiny typo: 'Pyodide'

packages/python-analysis/test/uv-python-version.test.ts
```diff
@@ -0,0 +1,644 @@
...
+          constraint: [{ operator: '==', version: '3.10', prefix: '' }],
+          variant: 'default',
+        },
+      });
+    });
+
+    it('parses pypy310 (compact form)', () => {
+      const result = parseUvPythonRequest('pypy310');
+      expect(result).not.toBeNull();
+      expect(result?.implementation).toBe('pypy');
```
ricardo-agz: Should we also be testing the version here? e.g. `expect(result?.version?.constraint[0].version).toBe('3.10');`

ricardo-agz (approved):

packages/python-analysis/src/manifest/types/pyproject.ts
```diff
@@ -0,0 +1,78 @@
+import type { UvConfig } from './uv-config';
+
+/**
+ * [build-system] per PEP 518.
+ */
+export interface PyProjectBuildSystem {
```
1st1: I'd start suing zod aggressively.

packages/python-analysis/src/manifest/package.ts
```diff
@@ -0,0 +1,639 @@
...
+  let configs: PythonConfigs[] = [];
+
+  for (;;) {
+    const prefixConfigs = await loadPythonConfigs(rootPath, prefix);
+    if (Object.keys(prefixConfigs).length !== 0) {
+      configs.push(prefixConfigs);
+    }
+
+    const prefixManifest = await loadPythonManifest(rootPath, prefix);
+    if (prefixManifest !== null) {
```
1st1: nit -- i usually use `== null` or `!= null` instead of `===` and `!==` because the former compares against both `null` and `undefined` and the latter is strict `null` comparison. A bare `return` in JS returns `undefined`

.changeset/light-lies-wonder.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
1st1, line 2: what's this?

packages/python-analysis/src/manifest/requirements-txt-parser.ts
```diff
@@ -195,7 +194,7 @@ function extractPipArguments(fileContent: string): {
   }
 
   return {
-    cleanedContent: cleanedLines.join(EOL),
+    cleanedContent: cleanedLines.join('\n'),
```
1st1, line 194: Why not use `EOL` everywhere? Right now i see EOL in some places, `\n` in others, `\r?\n` in yet others

1st1 (approved):

packages/python-analysis/test/uv-python-version.test.ts
elprans: Oh good catch. This actually exposed a bug!

.changeset/light-lies-wonder.md
elprans: Required by the automatic changelog system.

packages/python-analysis/src/manifest/requirements-txt-parser.ts
elprans: The requirements.txt parser will fail with Windows EOL.

packages/python-analysis/src/manifest/types/pyproject.ts
elprans: Done

PR #14564, opened 2026-01-08
Ensure internal build step runs if a build script is missing

jeffsee55:
When we detect a monorepo, we inject `turbo run` as a build command. However if there's no `build` in the package.json, nothing will happen. This change allows us to know whether that we're in a monorepo, and if so, we double-check that there actually is a build script, if not, we run the internal build.

Also-

- Removes `rolldown` dep where it's no longer used
- Doesn't fail on builds where there's no framework, allowing `export const fetch` to work


packages/backends/src/build.ts
```diff
@@ -156,3 +163,14 @@ export const doBuild = async (
     tsPromise,
   };
 };
+
+async function getPackageJson(entryPath: string): Promise<PackageJson> {
```
mehulkar: isn't there already a helper for this somewhere?

packages/cli/test/fixtures/unit/commands/build/turborepo-hono-monorepo/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "with-hono",
```
mehulkar: guessing other fixtures use this name, does it need an update? I've seen very weird bugs in turbo's test suite when an name in a fixture accidentally gets re-used

packages/express/package.json
```diff
@@ -27,7 +27,6 @@
     "@vercel/static-config": "workspace:*",
     "fs-extra": "11.1.0",
     "path-to-regexp": "8.3.0",
-    "rolldown": "1.0.0-beta.35",
```
mehulkar: why does rolldown get removed everywhere?

packages/cervel/src/find-entrypoint.ts
```diff
@@ -54,7 +54,16 @@ export const findEntrypoint = async (
   );
 
   if (!framework) {
-    throw new Error('No framework found in package.json');
+    for (const entrypoint of entrypoints) {
```
mehulkar, line 57: I don't understand this change, is it related to the task at hand (fallback to auto behavior when a build script is missing)?

packages/backends/src/build.ts
```diff
@@ -107,9 +106,17 @@ export const doBuild = async (
...
 
-  // If there's an output directory configured but no build command result, build ourselves
   const outputDir = join(args.workPath, outputSetting);
-  if (!buildCommandResult) {
+
+  const packageJson = await getPackageJson(args.workPath);
+
+  // Monorepo support injects a build command like 'turbo run build', but if
+  // this workspace doesn't have a build script, we need to build ourselves
+  const monorepoWithoutBuildScript =
```
mehulkar, line 114: IIUC, we're solving for when there's a custom `buildCommand`, but no `.scripts.build` in package.json, is that right?  It shouldn't matter whether it's a monorepo or not?

mehulkar (commented):
Wanted to clarify a few things before approving!

packages/cervel/src/find-entrypoint.ts
jeffsee55: Something unrelated I noticed while testing, updated in PR description

packages/backends/src/build.ts
jeffsee55: It matters because when there's a build command we never run the internal build logic, assuming that the specified build command is handling everything.

However, because the build command came from our custom monorepo detection logic, we don't have that same assurance.

packages/express/package.json
jeffsee55: Something unrelated I noticed while testing, updated in PR description

packages/cli/test/fixtures/unit/commands/build/turborepo-hono-monorepo/package.json
jeffsee55: Updated it to make more sense, but the test copies the entire `packages/cli/test/fixtures/unit/commands/build/turborepo-hono-monorepo` into a tmp dir so it doesn't clash with other `pnpm` stuff and works in isolation.

packages/backends/src/build.ts
jeffsee55: There's a few running around, moved this to `build-utils` for the next time someone wants it.

packages/backends/src/build.ts
mehulkar: ok, then could this be something like `if (autoInjectedBuildCommand) {}` ? (invert the logic and also name it so it corresponds with the situation). Conflating it with monorepos seems confusing

packages/express/package.json
mehulkar: if possible a separate PR would be nice for cleanups!

mehulkar (approved):
Don't want to block the PR, but left another comment!

PR #14566, opened 2026-01-09
[cli] Migrate build output from CommonJS to ESM

tknickman:
## Summary

This PR migrates the Vercel CLI build output from CommonJS to ESM (ECMAScript Modules), enabling the use of modern ESM-only packages while maintaining full backwards compatibility.

### Key Changes

- **Build Configuration**: Added `"type": "module"` to package.json, updated esbuild to output ESM format with a CJS compatibility shim for `require`, `__filename`, and `__dirname`
- **Worker Files**: Renamed standalone worker files (`builder-worker.js`, `get-latest-worker.js`) to `.cjs` extension since they use CommonJS syntax
- **Template Compilation**: Updated doT.js template compilation to handle ESM package context
- **TypeScript Config**: Updated module settings to `esnext` with `node16` resolution
- **Jest Config**: Renamed to `.cjs` for CommonJS compatibility in ESM package

### Why ESM?

- Enables use of modern npm packages that are ESM-only (e.g., Ink 5+/6+ for terminal UI)
- Better tree-shaking and static analysis capabilities
- Aligns with the Node.js ecosystem direction
- Prerequisite for future CLI enhancements

### Backwards Compatibility

This change is fully backwards compatible:
- CLI works identically for end users
- All existing functionality preserved
- Works with both Node.js and Bun runtimes
- No changes to public API or CLI interface

## Test plan

- [x] `pnpm build --filter=vercel` succeeds
- [x] `vercel --version` works correctly
- [x] `vercel whoami` works correctly
- [x] `vercel dev --help` loads without errors
- [x] Unit tests pass (1,582 passing)
- [x] Works with Node.js runtime
- [x] Works with Bun runtime


TooTallNate:
<picture data-single-emoji=":nice-90s-bounce:" title=":nice-90s-bounce:">:nice-90s-bounce:</picture>

.changeset/cli-esm-migration.md
```diff
@@ -0,0 +1,11 @@
+---
+"vercel": patch
```
mehulkar, line 2: can we do canary in case there are issues?

anthonyshew (approved):

packages/cli/scripts/build.mjs
```diff
@@ -65,11 +77,11 @@ copyFileSync(
...
+  new URL('src/util/dev/builder-worker.cjs', repoRoot),
+  new URL('builder-worker.cjs', distRoot)
 );
 copyFileSync(
-  new URL('src/util/get-latest-version/get-latest-worker.js', repoRoot),
-  new URL('get-latest-worker.js', distRoot)
+  new URL('src/util/get-latest-version/get-latest-worker.cjs', repoRoot),
+  new URL('get-latest-worker.cjs', distRoot)
 );
 copyFileSync(new URL('src/vc.js', repoRoot), new URL('vc.js', distRoot));
```
jeffsee55, line 87: I wonder if this should be `.mjs` just to be more explicit, I know it works because there's a `package.json` with `type: "module"` but does this ever run in a context where that package.json isn't present?

packages/cli/scripts/start.js
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../dist/index.js');
+await import('../dist/index.js');
```
mehulkar, line 3: what's our node support for CLI?

mehulkar (approved):
cool, let me know if you want to do more testing (not even sure how that happens)

.changeset/cli-esm-migration.md
tknickman: We can test this directly before release so don't have to do a canary

packages/cli/scripts/start.js
tknickman: `>= 18`, so we have top level await

styfle: We should drop node 18 support since deployments don't even support it https://vercel.com/changelog/node-js-18-is-being-deprecated

tknickman:
Validated this with full framework matrix, local testing, and manual builds

packages/cli/scripts/compile-templates.mjs
```diff
@@ -14,14 +17,18 @@ export async function compileDevTemplates() {
...
     const fnPath = new URL(file, templatesDirURL);
+    const cjsPath = new URL(file.replace(/\.js$/, '.cjs'), templatesDirURL);
     const tsPath = fnPath.href.replace(/\.js$/, '.ts');
     const def = await readFile(
       new URL(fnPath.href.replace(/\.js$/, '.tsdef')),
       'utf8'
     );
     const interfaceName = def.match(/interface (\w+)/)[1];
 
-    const { default: fn } = await import(fnPath);
```
styfle, line 24: I thought `await import()` would work with both CJS and ESM 🤔

packages/cli/scripts/compile-templates.mjs
```diff
@@ -14,14 +17,18 @@ export async function compileDevTemplates() {
...
+    // Rename to .cjs so require() treats it as CommonJS in ESM package
+    await rename(fnPath, cjsPath);
+    const fn = require(fileURLToPath(cjsPath));
+    await unlink(cjsPath);
```
styfle, line 32: Isn't `unlink()` basically delete?

The opposite of `rename()` should be another `rename()`

packages/cli/scripts/build.mjs
tknickman: I don't think it does, and I think the redundancy may be more confusing....but we can itg this one if needed

packages/cli/scripts/start.js
tknickman: Yea will update in a followup

packages/cli/scripts/compile-templates.mjs
tknickman: Yea updated the comment to clarify this

packages/cli/scripts/compile-templates.mjs
tknickman: You're right, it does I could have used that as well.

tknickman: Updated

packages/cli/scripts/compile-templates.mjs
```diff
@@ -14,14 +14,22 @@ export async function compileDevTemplates() {
...
+    const cjsPath = new URL(file.replace(/\.js$/, '.cjs'), templatesDirURL);
     const tsPath = fnPath.href.replace(/\.js$/, '.ts');
     const def = await readFile(
       new URL(fnPath.href.replace(/\.js$/, '.tsdef')),
       'utf8'
     );
     const interfaceName = def.match(/interface (\w+)/)[1];
 
-    const { default: fn } = await import(fnPath);
+    // doT generates CommonJS code, but since this is an ESM package (.js files
```
styfle, line 25: How many files is doT generating?

Perhaps it would be faster to also generate a package.json with `module: commonjs` once and then all the files can be read as the expected cjs without renaming anything.

packages/cli/src/vc.js
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
-'use strict';
 // This shim defers loading the real module until the compile cache is enabled.
 // https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+// enableCompileCache was added in Node.js 22.1.0, so we need to handle older versions.
```
styfle:
```suggestion
// enableCompileCache was added in Node.js 22.8.0, so we need to handle older versions.
```

Please review output from AI to avoid hallucinations

tknickman: Ooh good catch!

packages/cli/test/fixtures/unit/monorepo-link/.gitignore
```diff
@@ -1,3 +1,5 @@
 dist
 !/.vercel
 .vercel/output
+# Ignore .vercel directories generated in subdirectories by tests
+*/.vercel
```
styfle, line 5: Is this necessary?

Seems to contradict line 2

tknickman: I added this because these tests generate `.vercel `directories and don't clean them up, I prefer a clean working env after testing - I missed the first line here  🤔 will take a look at why that one is there.

packages/cli/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "allowJs": true,
+    "module": "es2022",
```
styfle, line 4: This should probably be `nodenext` per the documentation

https://www.typescriptlang.org/tsconfig/#module

```suggestion
    "module": "nodenext",
```

packages/cli/scripts/compile-templates.mjs
tknickman: It's only 5 files, so not too bad at all

packages/cli/tsconfig.json
tknickman: This would be a bigger refactor than I wanted to do in one PR because nodenext requires extensions on relative imports. We can get around that with module resolution of bundler but that needs TS 5 and we're not on that yet either so yet another large change. 

But I'm planning to queue those up next. 

If this was a lib that users were importing from I would be concerned about it but since the api contract is CLI only here imo this is an ok tradeoff for more iterative release.

styfle (approved):

PR #14582, opened 2026-01-11
feat(cli): add `vercel env run` command

julianbenegas:
Adds a new subcommand that runs any command with environment variables from the linked Vercel project, without writing them to the filesystem.

## Usage

```bash
vercel env run -- next dev
vercel env run -e preview --git-branch feature-x -- npm test
```

## Options

- `--environment, -e` - Target environment (default: development)
- `--git-branch` - Git branch for env overrides

## How it works

1. Fetches environment variables using the same API as `vercel env pull`
2. Does NOT write them to the filesystem
3. Spawns the user's command with all env vars exported
4. Returns the child process exit code


tknickman (commented):
Looks good, no main blockers. One thing that may be nice to add would be mentioning when `--` is required. For example If a user runs `vercel env run echo hello` (without --), This will just show "No command provided". 

It might be slightly more helpful to hint that -- is required, e.g., "No command provided. Use -- to separate vercel flags from your command."

tknickman (approved):
Approving to unblock

julianbenegas:
> Looks good, no main blockers. One thing that may be nice to add would be mentioning when `--` is required. For example If a user runs `vercel env run echo hello` (without --), This will just show "No command provided".
> 
> It might be slightly more helpful to hint that -- is required, e.g., "No command provided. Use -- to separate vercel flags from your command."

fixed https://github.com/vercel/vercel/pull/14582/commits/7c6b6fc783faa2cdd2aead999f3ecc8a06faf92e

PR #14583, opened 2026-01-12
[build-utils]: use https instead of http

Anshuman71:
A user reached out in the [Vercel Community](https://community.vercel.com/t/urls-included-in-the-warning-text-is-in-http-protocol/31156/3) sharing that the URLs in the error message for incompatible Node version should use https. 

I couldn't see a reason to keep them http when the https link works, so I made them all https.


LukeSheard (approved):

Anshuman71:
Hi @LukeSheard, I've added the changeset. However, I'm not sure if the failing checks are related to the changes I made. Can you take another look? Maybe I'm missing something else.

PR #14584, opened 2026-01-12
feat(cli): add upgrade command

tknickman:
Add upgrade command that support package manager aware auto upgrading, dry run, and json format. 

Also add prompt to update when terminal is interactive

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -62,6 +62,7 @@ import * as ERRORS from './util/errors-ts';
 import { APIError } from './util/errors-ts';
 import { SENTRY_DSN } from './util/constants';
 import getUpdateCommand from './util/get-update-command';
+import { executeUpgrade } from './util/upgrade';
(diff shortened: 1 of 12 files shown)


packages/cli/src/index.ts
```diff
@@ -918,7 +924,30 @@ Changelog: ${output.link(changelog, changelog, { fallback: false })}
...
+          });
+
+          if (answer === '' || answer === 'y' || answer === 'yes') {
+            output.print('\n');
+            const upgradeExitCode = await executeUpgrade();
+            process.exitCode = upgradeExitCode;
+            return;
+          }
+
+          output.print('\n');
```
onsclom: I think this is a good case for using `client.input.confirm`

onsclom (approved):

packages/cli/src/index.ts
tknickman: Good call!

tknickman: Done

PR #14590, opened 2026-01-12
Exit early in `vc bisect` if good and bad URLs are the same

TooTallNate:
### TL;DR

Added validation to prevent `vc bisect` from running when good and bad URLs are identical.

### What changed?

- Added an early exit condition in the `bisect` command that checks if the good and bad deployment URLs are the same
- Added a corresponding error message: "Good and Bad deployments must be different"
- Created a new test case to verify this validation works correctly

### How to test?

Run the `vc bisect` command with the same URL for both `--good` and `--bad` parameters:
```
vc bisect --good https://example.vercel.app --bad https://example.vercel.app --path /docs
```
The command should exit with an error message and return code 1.

### Why make this change?

Running a bisect operation when both the good and bad deployments are the same doesn't make sense and would waste resources. This validation prevents users from accidentally running an invalid bisect operation.

### DIFF
--- a/.changeset/gentle-cups-itch.md
+++ b/.changeset/gentle-cups-itch.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+Exit early in `vc bisect` if good and bad URLs are the same
--- a/packages/cli/src/commands/bisect/index.ts
+++ b/packages/cli/src/commands/bisect/index.ts
@@ -172,6 +172,11 @@ export default async function bisect(client: Client): Promise<number> {
     return 1;
   }
 
+  if (badDeployment.url === goodDeployment.url) {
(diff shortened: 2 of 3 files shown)


TooTallNate:
* **#14590** <a href="https://app.graphite.com/github/pr/vercel/vercel/14590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tknickman (approved):

styfle (approved):

ijjk (approved):

PR #14604, opened 2026-01-13
[builders]: remove getSpawnOptions usage

LukeSheard:
# Overview

This is a re-apply of https://github.com/vercel/vercel/pull/14176 without the changes to remove `getSpawnOptions` from the `build-utils` package due to the breaking changes that happened. 

## Background 

The vercel environment now populates the right tools into the toolchain when subprocess are spawned in user-space. This means that builders now no longer need to think about the toolchains they need to provision. 

In this case we no longer re-create `$PATH` to include the node version which is needed for the build.


LukeSheard:
* **#14604** <a href="https://app.graphite.com/github/pr/vercel/vercel/14604?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14604?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

marc-vercel (approved):
Code wise LGTM - Can you add in the PR what we talked privately: which change caused the previous revert and how we are avoiding the same problem now?

PR #14608, opened 2026-01-13
[cli] Do not show default settings when framework is detected

onsclom:
Default project settings are often not useful or confusing:
image

This PR hides those project settings when a framework is auto-detected, but still leaves the option to modify the settings

After:
image


packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -108,30 +108,35 @@ export async function editProjectSettings(
...
-              : chalk.italic(`${defaultSetting.placeholder}`)
-          }`
-        )}\n`
-      );
+      const defaultSetting = framework.settings[setting];
+      const override = localConfigurationOverrides?.[setting];
+
+      if (!override && defaultSetting) {
+        output.print(
+          `${chalk.dim(
```
mehulkar, line 131: we need to kill chalk

ricardo-agz (approved):

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -108,30 +108,35 @@ export async function editProjectSettings(
...
-            isSettingValue(defaultSetting)
-              ? defaultSetting.value
-              : chalk.italic(`${defaultSetting.placeholder}`)
-          }`
-        )}\n`
-      );
+      const defaultSetting = framework.settings[setting];
+      const override = localConfigurationOverrides?.[setting];
+
+      if (!override && defaultSetting) {
```
mehulkar, line 129: should this be `!override && !defaultSetting`? IIUC, if a framework is detected, and it has default settings, we don't want to print

packages/cli/src/util/input/edit-project-settings.ts
```diff
@@ -108,30 +108,35 @@ export async function editProjectSettings(
   output.print(
     !framework.slug
       ? `No framework detected. Default Project Settings:\n`
-      : `Auto-detected Project Settings (${styledFramework(framework.name)}):\n`
+      : `Auto-detected Project Settings for ${styledFramework(framework.name)}\n`
```
mehulkar, line 111: feel like we need some kind of glyph on the left side so it aligns with the rest of the output also (or maybe some leading `...` ?

mehulkar (approved):
nothing blocking, some qs

packages/cli/src/util/input/edit-project-settings.ts
tknickman: Yea I agree with this. Maybe ✓?

packages/cli/src/util/input/edit-project-settings.ts
onsclom: If it gets here, it didn't detect a framework, so `defaultSetting` will be the default settings for no framework

packages/cli/src/util/input/edit-project-settings.ts
onsclom: Sounds like the existing util `output.log` instead of `output.print` is a good fit for this and it prepends with `>`:

image

tknickman: Nice!

PR #14612, opened 2026-01-14
Add `experimentalServices` to `vercel.json`

ricardo-agz:
Adds `experimentalServices` to `vercel.json`
Implements `detectServices` which, for now, just resolves the services configured in `vercel.json`


MatthewStanciu (approved):

1st1:
Hm, this is very different from what was proposed in the RFC
CleanShot 2026-01-13 at 20 29 00@2x

ricardo-agz:
> Hm, this is very different from what was proposed in the RFC

@1st1 Do we want to offer both `services` and `serviceGroups` keys for simplicity & extra configuration when needed? 
I added `experimentalServiceGroups` as a key as well.

The "experimental" prefix is temporary while we iterate and the extra config settings like `maxDurarion`, `memory`, etc. are porting over the configs for [functions](https://vercel.com/docs/project-configuration/vercel-json#functions)

packages/fs-detectors/src/services/index.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export async function detectServices(
+  options: DetectServicesOptions
+): Promise<DetectServicesResult> {
+  const { fs, workPath = '' } = options;
+  const services: ResolvedService[] = [];
+  const errors: ServiceDetectionError[] = [];
+
+  // Read vercel.json
+  const configPath = workPath ? `${workPath}/vercel.json` : 'vercel.json';
```
jeffsee55, line 27: Just calling out that this can also be `.ts` file. There may already be some abstract tooling to read whatever is provided automatically cc @MatthewStanciu ?

Seems ok for the experimental phase to not block on this.

jeffsee55 (approved):

packages/fs-detectors/src/services/index.ts
ricardo-agz: Good callout, yes, we should definitely add `vercel.ts` support going forward

PR #14617, opened 2026-01-14
[experimental-services] multi service builds

ricardo-agz:
Implements the baseline zero-config services detection when a project's framework = "services"
Also implements the actual build for the multiple services, either specified in `vercel.json` or auto-detected

Future PR TODOs:
* Integrate with Python's new entrypoint detection
* Fix an issue where separate python services use the same virtual environment in `.vercel/python` which causes issues since other services can add/remove dependencies from the virtual environment that the other services depend on
* Fix an issue where services 404 unless you hard code the paths to include the routePrefix


packages/fs-detectors/src/services/get-services-builders.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  // Convert warnings to error format
+  const warningResponses: ErrorResponse[] = (result.warnings || []).map(w => ({
+    code: w.code,
+    message: w.message,
+  }));
+
+  if (result.errors.length > 0) {
+    return {
+      builders: null,
```
mehulkar, line 76: weird interface that an error response is returning the field..

packages/fs-detectors/src/services/get-services-builders.ts
```diff
@@ -0,0 +1,194 @@
...
+      defaultRoutes: null,
+      redirectRoutes: null,
+      rewriteRoutes: null,
+      errorRoutes: null,
+    };
+  }
+
+  if (result.services.length === 0) {
+    return {
+      builders: null,
```
mehulkar, line 91: can this be an empty array? (my first time seeing this code, so i assume this is too invasive)

packages/fs-detectors/src/services/builders.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+export function generateServicesRoutes(services: ResolvedService[]): {
+  rewriteRoutes: Route[];
+  defaultRoutes: Route[];
+} {
+  const rewriteRoutes: Route[] = [];
+  const defaultRoutes: Route[] = [];
+
+  // Sort services by route prefix length (longest first, then primary last)
+  const sortedServices = [...services].sort((a, b) => {
```
mehulkar: - can we add a comment about why this is the order chosen? 
- why not return it sorted from `detectServices()`? (probably depends on the answer to the first q)

packages/fs-detectors/src/services/builders.ts
```diff
@@ -0,0 +1,194 @@
...
+    if (prefixB === '' && prefixA !== '') return -1;
+    // Otherwise sort by length (longest first)
+    return prefixB.length - prefixA.length;
+  });
+
+  for (const service of sortedServices) {
+    const prefix = service.routePrefix || '';
+    const builderSrc = service.builder.src || '';
+    // Strip extension to get function path
+    const functionPath = '/' + builderSrc.replace(/\.[^/.]+$/, '');
```
mehulkar: for my understanding, is this a filepath from source code or from Build Output API? a comment would be helpful

packages/fs-detectors/src/services/entrypoints.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+export async function detectAllEntrypoints(
+  fs: DetectorFilesystem,
+  workDir: string,
+  runtimes: ServiceRuntime[]
+): Promise<DetectedEntrypoint[]> {
+  const detected: DetectedEntrypoint[] = [];
+
+  for (const runtime of runtimes) {
+    const entrypoint = await detectEntrypoint(fs, runtime, workDir);
```
mehulkar: want to parallelize this? since it's doing `fs.hasPath()` it might be a little faster

packages/fs-detectors/src/services/index.ts
```diff
@@ -3,53 +3,333 @@ import type {
...
+ * manifest files and entrypoints (zero-config mode).
  */
 export async function detectServices(
   options: DetectServicesOptions
 ): Promise<DetectServicesResult> {
   const { fs, workPath = '' } = options;
+
+  // Use explicit services if provided, otherwise read from vercel.json
+  let experimentalServices: ExperimentalServices | undefined =
+    options.explicitServices;
```
mehulkar: having trouble tracing while reading this diff... where does `options.explicitServices` originate?

packages/fs-detectors/src/services/index.ts
```diff
@@ -3,53 +3,333 @@ import type {
...
+
+  if (!experimentalServices) {
+    const configPath = workPath ? `${workPath}/vercel.json` : 'vercel.json';
+
+    const hasVercelJson = await fs.hasPath(configPath);
+    if (hasVercelJson) {
+      try {
+        const configBuffer = await fs.readFile(configPath);
+        const config = JSON.parse(configBuffer.toString('utf-8'));
+        experimentalServices = config.experimentalServices;
```
mehulkar: - isn't there a helper for this somewhere that will safely read/return something?

packages/fs-detectors/src/services/auto-detection/manifests.ts
```diff
@@ -0,0 +1,137 @@
...
+  maxDepth: number = MAX_MANIFEST_SEARCH_DEPTH
+): Promise<DetectedManifest[]> {
+  const manifests: DetectedManifest[] = [];
+  const manifestFileNames = new Set(Object.keys(MANIFEST_FILES));
+
+  async function walk(dir: string, depth: number): Promise<void> {
+    if (depth > maxDepth) return;
+
+    try {
+      const entries = await fs.readdir(dir);
```
mehulkar, line 43: Seems like a `glob` would be better here than walking `fs.readdir`, check out `tinyglobby`, it's the recommended glob lib with the smallest footprint: https://github.com/SuperchupuDev/tinyglobby

https://e18e.dev/docs/replacements/glob.html

ricardo-agz: Yea glob would be much better, we'd just have to implement glob support for the [DetectorFilesystem in api](https://github.com/vercel/api/blob/main/packages/detect-framework/sparse-clone-detector-filesystem.ts)
I can follow up with this though

packages/fs-detectors/src/services/get-services-builders.ts
ricardo-agz: refactored how we use `detectBuilders` and `getServicesBuilders`, now `detectBuilders runs the actual detection/resolution solig and `getServicesBuilders` is an adapter for us to use in `detectBuilders`

packages/fs-detectors/src/services/get-services-builders.ts
ricardo-agz: refactored `getServicesBuilders` to match the return type of `getBuilders` so it's just an adapter over `detectServices`

packages/fs-detectors/src/services/index.ts
ricardo-agz: cleaned this up so now `detectServices` doesn't require the caller to have read `vercel.json` and passed in the explicit services which was pretty unintuitive so hopefully it is more readable now

packages/cli/src/commands/build/index.ts
```diff
@@ -566,9 +566,12 @@ async function doBuild(
     isZeroConfig = true;
 
     // Detect the Vercel Builders that will need to be invoked
+    // Note: detectBuilders handles the "services" framework internally when
+    // projectSettings.framework === 'services'
     const detectedBuilders = await detectBuilders(files, pkg, {
       ...localConfig,
       projectSettings,
+      vercelConfig: localConfig,
```
elprans: `vercelConfig` seems to be missing from `detect-builders.ts/Options`

packages/fs-detectors/src/services/resolve-configured.ts
```diff
@@ -30,22 +87,50 @@ export function resolveService(
...
+  } else if (config.runtime) {
+    const runtime = config.runtime as ServiceRuntime;
+    builderUse = RUNTIME_BUILDERS[runtime];
+    builderSrc = config.entrypoint!;
+  } else {
+    const runtime = inferRuntimeFromExtension(config.entrypoint!);
+    builderUse = getBuilderForRuntime(runtime!);
+    builderSrc = config.entrypoint!;
+  }
+
```
elprans, line 117: Should we validate if `builderSrc` actually exists to catch config skew?

packages/fs-detectors/src/services/index.ts
```diff
@@ -1,55 +1,427 @@
...
  *
- * Reads service configurations from vercel.json `experimentalServices`
- * and resolves them into ResolvedService objects.
+ * This is the main entry point for service detection. It:
+ * 1. Reads vercel.json to check for configured services
+ * 2. If `experimentalServices` exists, resolves those (configured)
+ * 3. Otherwise, auto-detects from manifests and entrypoints (detected)
+ * 4. Generates routing rules for the services
  */
 export async function detectServices(
```
elprans: I would reorg this into a submodule, otherwise the import graph is a bit weird `index.ts -> builders.ts -> index.ts`

packages/fs-detectors/src/services/manifests.ts
```diff
@@ -0,0 +1,137 @@
...
+  'pyproject.toml': 'python',
+  'requirements.txt': 'python',
+  Pipfile: 'python',
+  'go.mod': 'go',
+  'Cargo.toml': 'rust',
+  Gemfile: 'ruby',
+};
+
+// Directories to skip during manifest detection.
+const IGNORED_DIRECTORIES = new Set([
```
elprans: Seems like we are adding these lists all over the place (`api/detect-framework` has its [own version](https://github.com/vercel/api/blob/e4dcb798e6e3ff7ee6023b97f63e2f24142d0e98/packages/detect-framework/sparse-clone-detector-filesystem.ts#L46)).  Consider unifying.

packages/fs-detectors/src/services/index.ts
```diff
@@ -1,55 +1,427 @@
...
+        dest: functionPath,
+        check: true,
+      });
+    } else {
+      // Non-primary service: prefix-based rewrite
+      const normalizedPrefix = prefix.startsWith('/')
+        ? prefix.slice(1)
+        : prefix;
+      rewrites.push({
+        src: `^/${normalizedPrefix}(?:/(.*))?$`,
```
elprans: Should we preserve the captured path (the `(.*)` bit)?

elprans:
>     * Fix an issue where separate python services use the same virtual environment in `.vercel/python` which causes issues since other services can add/remove dependencies from the virtual environment that the other services depend on

Is the plan to use separate venv for each service?

packages/fs-detectors/src/services/resolve-configured.ts
ricardo-agz: we could but it is also handled gracefully by the builders themselves, for example [in the python builder](https://github.com/vercel/vercel/blob/583c0521046dde3574afa506eb4fa7b9c150394f/packages/python/src/index.ts#L157)

packages/fs-detectors/src/services/manifests.ts
ricardo-agz: yea good point, moved it to `constants.ts`, we can import from here in the api repo since it already imports @vercel/fs-detectors

ricardo-agz:
> > ```
> > * Fix an issue where separate python services use the same virtual environment in `.vercel/python` which causes issues since other services can add/remove dependencies from the virtual environment that the other services depend on
> > ```
> 
> Is the plan to use separate venv for each service?

@elprans The case for this would be if you have two separate python services, e.g. a Flask one and a FastAPI one with separate `pyproject.toml`s and separate dependencies they could each be isolated

packages/fs-detectors/src/services/detect-services.ts
```diff
@@ -0,0 +1,135 @@
...
+        check: true,
+      });
+    } else {
+      // Non-root service: prefix-based rewrite
+      const normalizedPrefix = routePrefix.startsWith('/')
+        ? routePrefix.slice(1)
+        : routePrefix;
+      rewrites.push({
+        src: `^/${normalizedPrefix}(?:/.*)?$`,
+        dest: routePrefix,
```
elprans, line 128: Need to figure out a way to make rewrites actually rewrite the route so that the service sees `dest` and not `src` (this is outside of this PR, but I think it's important to nail this before this lands as this is part of the contract).

PR #14619, opened 2026-01-14
Run NFT and introspection in parrallel and increase introspection timeout

jeffsee55:
- Parrallelize NFT and introspection steps
- Increase timeout to 8 seconds. It's taking up to 5 seconds for a large app we have been testing with
- Add more debug logs

### DIFF
--- a/.changeset/serious-tigers-collect.md
+++ b/.changeset/serious-tigers-collect.md
@@ -1,0 +1,5 @@
+---
+'@vercel/introspection': patch
+---
+
+Add more debug logs to introspection process
--- a/.changeset/serious-tigers-collect.md
+++ b/.changeset/serious-tigers-collect.md
@@ -1,5 +1,8 @@
 ---
 '@vercel/introspection': patch
+'@vercel/backends': patch
 ---
 
-Add more debug logs to introspection process
+- Parrallelize NFT and introspection steps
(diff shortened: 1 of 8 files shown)


tknickman (approved):

packages/backends/src/index.ts
```diff
@@ -32,17 +32,21 @@ export const build: BuildV2 = async args => {
...
-    debug(`Route mapping failed to detect routes`);
+    debug(`Introspection failed to detect routes`);
   }
 
   const handler = relative(
     args.repoRootPath,
     join(outputConfig.dir, outputConfig.handler)
   );
 
+  const { files } = await nftPromise;
```
mehulkar, line 47: deferring await feels like weird way to parallelize. should we Promise.all instead?

.changeset/serious-tigers-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/introspection': patch
+---
+
+Add more debug logs to introspection process
```
mehulkar: the parallelization piece seems like the more important change, should it be in the notes?

mehulkar (approved):

packages/backends/src/index.ts
jeffsee55: I wanted to log the debug statements in a timely manner so we get somewhat realistic timing info. Open to ideas there but didn't really want to move the debug logs inside those functions.

PR #14626, opened 2026-01-15
[go]Use deterministic tmpdir path for go cache dirs.

gscho:
#14484 introduced support for `GOCACHE` and `GOMODCACHE`. This actually slowed down builds for customers because we compile a binary for every go handler separately and sequentially. Since the `getWriteableDirectory` function always returns a randomly named tmpdir, this meant we lost caching altogether.

The fix implemented here is to use a deterministic directory based on the `workPath`. I was able to reproduce the issue and test that this change greatly improved the compile time.

### DIFF
--- a/.changeset/quiet-spoons-clean.md
+++ b/.changeset/quiet-spoons-clean.md
@@ -1,0 +1,5 @@
+---
+'@vercel/go': patch
+---
+
+Use deterministic tmpdir when setting GOMODCACHE and GOCACHE dirs.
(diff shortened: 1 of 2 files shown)


brookemosby (approved):
LGTM

tknickman (approved):
LGTM! 

The fix correctly addresses the root cause - `getWriteableDirectory()` returning random tmpdirs was breaking cache sharing between sequential Go handler builds.

**Key observations:**
- Using MD5 hash of workPath for deterministic cache paths is a clean solution
- The path structure change from `.cache/go-build` to `go-build` aligns with the global cache structure (line 305)
- Sequential operations instead of `Promise.all()` is safer for symlink creation
- Removing symlinks before creating handles broken symlinks correctly

Nice fix! 🎉

PR #14627, opened 2026-01-15
bulk redirects

brookemosby:
Making bulk redirects work with prebuilt


packages/client/src/utils/index.ts
```diff
@@ -169,6 +171,41 @@ export async function buildFileTree(
       debug(`Error checking for .vercel/routes.json: ${e}`);
     }
 
+    // Include bulkRedirectsPath file if specified (for prebuilt deployments)
+    if (prebuilt && bulkRedirectsPath) {
```
mehulkar, line 175: Why not make it part of BOA? (In other words.. why limit this to prebuilt deployments)?

mehulkar (approved):

packages/client/src/utils/index.ts
brookemosby: The file path handling is different in BOA

PR #14638, opened 2026-01-15
Fix issue when stdout chunks are split from introspection data

jeffsee55:
The introspection is sent via stdout and when there are other log messages it would mess up the chunking and change the string's start and end characters, causing `JSON.parse` to fail.

NOTE: this would be a lot easier if I'd used `fork` instead of `spawn`, but I'm thinking this will soon need to invoke a bun process (perhaps), so might not be worth the switch


packages/introspection/src/index.ts
```diff
@@ -70,15 +73,38 @@ export const introspectApp = async (args: {
...
 
+      // Buffer to accumulate stdout data across chunks
+      let stdoutBuffer = '';
+
       child.stdout?.on('data', data => {
         try {
-          introspectionData = introspectionSchema.parse(
-            JSON.parse(data.toString() || '{}')
+          // Accumulate data in buffer to handle chunked output
+          stdoutBuffer += data.toString();
```
mehulkar: Would it make sense to open a file as a write stream instead?

jeffsee55: For memory concerns? I don't anticipate this to be super large but probably a good idea in the long-term. Let me circle back on that.

mehulkar: yeah, since the stdout comes from userland it is effectively unbounded

jeffsee55: Oh right, forgot I'm now taking all stdout. Updated.

packages/introspection/src/index.ts
```diff
@@ -95,14 +108,71 @@ export const introspectApp = async (args: {
...
+            try {
+              const stdoutBuffer = readFileSync(tempFilePath, 'utf8');
+
+              // Check if we have a complete introspection result
+              const beginIndex = stdoutBuffer.indexOf(
+                BEGIN_INTROSPECTION_RESULT
+              );
+              const endIndex = stdoutBuffer.indexOf(END_INTROSPECTION_RESULT);
+
+              if (beginIndex !== -1 && endIndex !== -1) {
```
mehulkar, line 147: I was playing around with this in chatgpt yesterday.. there might be a way to use `Transform`s to do the filter. Not sure if it's more ergonomic or has any benefits though, i was just wondering https://chatgpt.com/share/e/69712e5c-e310-800b-85a5-da79e2cece4f

mehulkar (approved):

PR #14653, opened 2026-01-18
feat(cli): add `vercel blob store list` command

Grappeggia:
## Summary

Add a new CLI command `vercel blob store list` that lists all Blob stores for a team, enabling users to see their blob stores directly from the CLI without needing to switch to the dashboard.

**Resolves: STO-1475**

## Features

- Lists all blob stores with name, ID, status, region, size, file count, connected projects, and age
- `--json` flag for machine-readable output
- `--no-projects` flag to hide the Projects column  
- `ls` alias for the `list` subcommand
- Follows CLI conventions consistent with other list commands (`project list`, `domains ls`, `integration list`)

## Usage

```bash
# List all blob stores (table format)
vercel blob store list

# Using alias
vercel blob store ls

# JSON output
vercel blob store list --json

# Hide projects column
vercel blob store list --no-projects
```

## Example Output

```
Blob stores in my-team:

  Name              ID                       Status      Region   Size    Files   Projects           Age
  my-uploads        store_abc123def456gh     ● Active    iad1     2.3MB   142     my-app, api (+1)   5d
  user-avatars      store_xyz789abc012de     ● Active    sfo1     856KB   89      frontend           2w
  temp-storage      store_qrs456tuv789wx     ● Active    cdg1     12MB    1.2k    –                  3mo
```

## Test Plan

- Added comprehensive unit tests covering:
  - Table output format
  - JSON output
  - `--no-projects` flag
  - `ls` alias
  - Empty result handling
  - Error cases
  - Telemetry tracking


packages/cli/src/commands/blob/command.ts
```diff
@@ -229,12 +229,54 @@ export const getStoreSubcommand = {
...
+  description: 'List all Blob stores',
+  arguments: [],
+  options: [
+    {
+      name: 'json',
+      shorthand: null,
+      type: Boolean,
+      deprecated: false,
+      description: 'Output results as JSON',
+    },
```
tknickman, line 408: We can leave this for now, but we're going to be adding a new param here - `format`, that can support a value of `json` in another PR. We're moving all commands to this so we have the flexibility to support multiple output formats. But we can consider leaving a `json` value as well for easy discoverability.

packages/cli/src/commands/blob/command.ts
```diff
@@ -229,12 +229,54 @@ export const getStoreSubcommand = {
...
+    {
+      name: 'no-projects',
```
tknickman, line 410: This is fine for now, would like to support a more flexible `--table=id,name,projects` etc in the future

tknickman (commented):
Two comments but nothing blocking - @styfle can you take a look here for the blob API usage / data filtering?

vvo (approved):

packages/cli/src/commands/blob/command.ts
styfle: I think we should remove `no-projects` since no other command has this and removing it later would be a breaking change.

In the future, we can add what @tknickman is suggesting.

packages/cli/src/commands/blob/store-list.ts
```diff
@@ -0,0 +1,216 @@
...
+    );
+
+    if (blobStores.length === 0) {
+      if (jsonOutput) {
+        client.stdout.write(`${JSON.stringify({ stores: [] }, null, 2)}\n`);
+      } else {
+        output.log(`No blob stores found in ${chalk.bold(contextName)}.`);
+      }
+      return 0;
+    }
```
styfle:
```suggestion
    if (blobStores.length === 0 && !jsonOutput){
        output.log(`No blob stores found in ${chalk.bold(contextName)}.`);
      }
      return 0;
    }
```

You can let the json option go through the normal logic to `outputJson()` and it will do the same thing.

packages/cli/src/commands/blob/store-list.ts
```diff
@@ -0,0 +1,216 @@
...
+      store.region || '-',
+      bytes(store.size) || '0B',
+      formatCount(store.count),
+    ];
+
+    if (noProjects) {
+      return [...baseRow, age];
+    }
+
+    return [...baseRow, formatProjects(store.projectsMetadata), age];
```
styfle:
```suggestion
    if (!noProjects) {
      baseRow.push(formatProjects(store.projectsMetadata));
    }
    baseRow.push(age);
    return baseRow;
```

No need to create a new array since you can push into the existing one.

packages/cli/src/commands/blob/command.ts
Grappeggia: Sg, leaving it as is for now

packages/cli/src/commands/blob/command.ts
Grappeggia: +1, removed

brookemosby (approved):
LGTM

PR #14654, opened 2026-01-19
allow installing products in one command

mtoth-vercel:
allows users to reference product directly in the install command, saving a step if the integration is already installed

## installing new product on existing integration
Screenshot 2026-01-18 at 9 18 28 PM

## installing new integration + product
(same as before)
Screenshot 2026-01-18 at 9 22 40 PM

## getting product names (slugs actually) for an integration
Screenshot 2026-01-18 at 9 34 50 PM

## considerations
* requires the user to know the name of the integration they want to install in the first place. Opening up the list of integrations on the marketplace/adding search is a TODO.
* command is listed under the `vc integrations` which is described as "managing your existing integrations" but this will return products for **any** integration slug you pass in, not just the ones you have installed (not a blocker imo and is kinda nice)

### DIFF
--- a/.changeset/violet-pets-compare.md
+++ b/.changeset/violet-pets-compare.md
@@ -1,0 +1,5 @@
+---
+'vercel': minor
+---
+
+Add support for installing marketplace products directly and listing available products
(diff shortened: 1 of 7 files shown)


tonypan2:
Lgtm. What does single product integration installation look like?

PR #14658, opened 2026-01-19
[build-utils] Optimize getNodeVersion by separating package.json lookup

ofhouse:
## Summary

- Add `findPackageJson` function for lightweight package.json lookup without lockfile scanning
- Update `getNodeVersion` to use `findPackageJson` instead of `scanParentDirs`
- Improves performance by avoiding unnecessary lockfile parsing when only `package.json` is needed

## Problem

`getNodeVersion` was calling `scanParentDirs` internally, which scans and parses all possible lockfiles (`yarn.lock`, `package-lock.json`, `pnpm-lock.yaml`, `bun.lock`, `bun.lockb`, `vlt-lock.json`). However, `getNodeVersion` only needs `packageJson?.engines?.node` - it doesn't use any lockfile information.

## Solution

Created a new `findPackageJson` function that only traverses directories to find and optionally read `package.json`, without any lockfile scanning. Updated `getNodeVersion` to use this lightweight function.

## Performance Impact

**Before**: `getNodeVersion` triggered:
- `walkParentDirs` for `package.json` + `walkParentDirsMulti` for 6 lockfiles
- Reading and parsing up to 5 lockfiles  
- `readProjectRootInfo` traversal for turbo support

**After**: `getNodeVersion` triggers:
- Only `walkParentDirs` for `package.json`
- Only reading/parsing that single file

## Breaking Changes

None - all exports maintain the same function signatures.

### DIFF
--- a/.changeset/optimize-get-node-version.md
+++ b/.changeset/optimize-get-node-version.md
@@ -1,0 +1,5 @@
+---
+'@vercel/build-utils': patch
+---
+
+Add `findPackageJson` function for optimized package.json lookup without lockfile scanning. This improves `getNodeVersion` performance by avoiding unnecessary lockfile parsing.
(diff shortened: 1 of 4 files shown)


janory (approved):
LGTM! 👍

PR #14671, opened 2026-01-20
Capture backends tracing

jeffsee55:
Adds spans for tracing during backends builds. Captures build, nft, typescript, and introspection as well as how many routes were captured in the introspection step.


packages/backends/src/index.ts
```diff
@@ -11,26 +11,56 @@ import {
...
   debug('Introspection starting..');
-  const { routes, framework } = await introspectApp({
-    ...outputConfig,
-    framework: args.config.framework,
-    env: {
-      ...(args.meta?.env ?? {}),
-      ...(args.meta?.buildEnv ?? {}),
-    },
+  const introspectAppSpan = span.child('vc.builder.backends.introspectApp');
+  const { routes, framework } = await introspectAppSpan.trace(async span => {
```
mehulkar, line 51: how does this work? is there something that adds `.trace` to the function prototype?

mehulkar (approved):

packages/backends/src/index.ts
mehulkar: q doesn't make sense... the `.trace` is on the span object, i thought it was on `introspectApp` function, and there was something magical making `.trace()` possible

PR #14677, opened 2026-01-20
[python] Use uv instead of pip for static-build

gscho:
The `python3` and `pip3` binaries are now managed by `uv`. Use `uv add` and `uv sync` to replace `pip3` since we are working towards moving everything to `uv`.


pyproject.toml
```diff
@@ -1,10 +0,0 @@
-[project]
```
elprans: Why were those deleted?

gscho: Accident

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -1548,13 +1548,13 @@ export async function runPipInstall(
...
   assert(path.isAbsolute(destPath));
-  const opts = { ...spawnOpts, cwd: destPath, prettyCommand: 'pip3 install' };
+  const opts = { ...spawnOpts, cwd: destPath };
 
-  await spawnAsync(
-    'pip3',
-    ['install', '--disable-pip-version-check', ...args],
-    opts
-  );
+  await spawnAsync('uv', ['add', ...args], {
```
elprans: This makes the function very specifically limited to what `uv add` can do.  It would IMO be less risky to just do `uv pip install`, which is explicitly compatible with `pip install`

gscho: - `uv pip install` -> fails with no venv
- Adding a venv before `uv pip install` failed
- `uv add -r requirements.txt && uv sync` failed
- `uv pip install --system` failed
- `uv pip install --break-system-packages` failed
- `uv pip install --break-system-packages --system` failed

Why are we using `uv`?

elprans: Hm. So `pip install` succeeded where `uv pip install` is now failing?  Odd, but fine.  `uv add` followed by `uv sync` is a fine approach, but the function should not be called `runPipInstall` then.

gscho: `pip3` used to be installed via a rpm. Since we moved to `uv` it won't work because its a "managed" environment. `uv` won't seem to work without a virtual environment. I've tried many incantations at this point and still failing.

PR #14683, opened 2026-01-21
fix ruby + python static builds e2e fixture failures

ricardo-agz:
**Static Builds with Ruby:**
Add gem executable bin directory to PATH

**Static Builds with Python:**
Run `uv pip install --target ...` and add the target to `PYTHONPATH` instead of `pip install`

### DIFF
--- a/packages/build-utils/src/fs/run-user-scripts.ts
+++ b/packages/build-utils/src/fs/run-user-scripts.ts
@@ -1536,13 +1536,16 @@ export async function runBundleInstall(
   await spawnAsync('bundle', args.concat(['install']), opts);
 }
 
-export interface PipInstallResult {
-  /**
-   * The directory where packages were installed.
-   * Add this to PYTHONPATH when running Python commands.
-   */
-  targetDir?: string;
-}
+export type PipInstallResult =
+  | { installed: false }
+  | {
+      installed: true;
+      /**
+       * The directory where packages were installed.
+       * Add this to PYTHONPATH when running Python commands.
+       */
+      targetDir: string;
+    };
 
 export async function runPipInstall(
   destPath: string,
@@ -1552,7 +1555,7 @@ export async function runPipInstall(
 ): Promise<PipInstallResult> {
   if (meta && meta.isDev) {
     debug('Skipping dependency installation because dev mode is enabled');
-    return {};
+    return { installed: false };
   }
 
   assert(path.isAbsolute(destPath));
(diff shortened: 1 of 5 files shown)


gscho (approved):

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -1536,25 +1536,68 @@ export async function runBundleInstall(
   await spawnAsync('bundle', args.concat(['install']), opts);
 }
 
+export interface PipInstallResult {
+  /**
+   * The directory where packages were installed.
+   * Add this to PYTHONPATH when running Python commands.
+   */
+  targetDir?: string;
```
mehulkar: why optional? it might make more sense to change the return type to be a union and only return the result if there is a targetDir

packages/build-utils/src/fs/run-user-scripts.ts
```diff
@@ -1536,25 +1536,68 @@ export async function runBundleInstall(
...
+    // Only fall back to pip3 if uv is not found (ENOENT)
+    // Other errors (e.g., install failures) should propagate
+    const isUvNotFound =
+      err instanceof Error &&
+      'code' in err &&
+      (err as NodeJS.ErrnoException).code === 'ENOENT';
+    if (!isUvNotFound) {
+      throw err;
+    }
+    debug('uv not found, falling back to pip3');
```
mehulkar: when would it not be available? isn't this just for vercel's build container?

packages/cli/test/unit/commands/link/index.test.ts
```diff
@@ -1022,7 +1022,7 @@ describe('link', () => {
       await expect(client.stderr).toOutput(
         'Which scope should contain your project?'
       );
-      client.stdin.write('y\n');
+      client.stdin.write('\r'); // Select first option (user) in select prompt
```
mehulkar: are we losing an assertion that pressing "y" works or is it covered in other tests?

packages/static-build/src/index.ts
```diff
@@ -626,22 +628,44 @@ export const build: BuildV2 = async ({
...
 
-    if (isPipInstall) {
-      // TODO: Add bins to PATH once we implement pip caching
+    if (isPipInstall && pipTargetDir) {
+      // Add pip bin directory to PATH for CLI commands (e.g., `mkdocs` instead of `python3 -m mkdocs`)
+      const pipBinDir = path.join(pipTargetDir, 'bin');
+      pathList.push(pipBinDir);
+      debug(
+        `Added "${pipBinDir}" to PATH env because a requirements.txt was found`
+      );
```
mehulkar, line 646: Is there a user facing log for this too? maybe not here, but at the point where we actually modify `PATH`, we might want to list out everything that was added

mehulkar (approved):
Some suggestions/questions in line, but otherwise lgtm

packages/cli/test/unit/commands/link/index.test.ts
ricardo-agz: Since this flaky test is not blocking CI I can remove it for now if we want, issue is whenever you press "y" if you have a team name starting with "y" it'll select that one which causes failures

packages/build-utils/src/fs/run-user-scripts.ts
ricardo-agz: It should always be available in the build container, I guess this is not necessary

PR #14695, opened 2026-01-21
fix(next): use base metadata headers for segment fallback initial headers

wyattjoh:
fix(next): use base metadata headers for segment fallback initial headers

When serving segment fallbacks, the initial headers should be inferred
from the base metadata file which includes the cache tags for the
fallback render. Without this, build output entries are produced without
cache tags and become unexpirable.

chore: add changeset for segment fallback headers fix

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -3285,15 +3285,25 @@ export const onPrerenderRoute =
             // then therefore unexpirable.
             let segmentInitialHeaders = initialHeaders;
             if (
-              !segmentInitialHeaders &&
-              // Blocking/fallback routes don't emit per-segment headers from Next,
-              // so inherit the route meta headers to carry cache tags
               (isBlocking || isFallback) &&
               'headers' in meta &&
               typeof meta.headers === 'object' &&
               meta.headers !== null
             ) {
-              segmentInitialHeaders = meta.headers as Record<string, string>;
+              const metaHeaders = meta.headers as Record<string, string>;
+              const hasMissingMetaHeader =
+                !segmentInitialHeaders ||
+                Object.keys(metaHeaders).some(
+                  key => segmentInitialHeaders?.[key] == null
+                );
+              if (hasMissingMetaHeader) {
+                // Merge to fill any missing meta headers without overriding
+                // segment-specific values.
+                segmentInitialHeaders = {
+                  ...metaHeaders,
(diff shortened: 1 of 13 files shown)


wyattjoh:
* **#14695** <a href="https://app.graphite.com/github/pr/vercel/vercel/14695?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14695?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14689** <a href="https://app.graphite.com/github/pr/vercel/vercel/14689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/next/src/utils.ts
```diff
@@ -3277,6 +3277,23 @@ export const onPrerenderRoute =
...
+            // No initial headers here means that the fallback is being served
+            // and it should infer the initial headers from the base metadata
+            // file which will be the ones which include the cache tags for
+            // the fallback render. Failing to do this will result in
+            // producing build output entries without any cache tags which are
+            // then therefore unexpirable.
+            let segmentInitialHeaders = initialHeaders;
+            if (
+              !segmentInitialHeaders &&
+              isBlocking &&
```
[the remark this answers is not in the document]
wyattjoh: No, because fallbacks will use the `fallbackHeaders` if that's the case.

ztanner: is there a test we can write for this? I'm trying to understand the scenario this is triggered by. When is a fallback blocking in a PPR path?

packages/next/src/utils.ts
```diff
@@ -3277,6 +3277,23 @@ export const onPrerenderRoute =
...
 
+            // No initial headers here means that the fallback is being served
+            // and it should infer the initial headers from the base metadata
+            // file which will be the ones which include the cache tags for
+            // the fallback render. Failing to do this will result in
+            // producing build output entries without any cache tags which are
+            // then therefore unexpirable.
+            let segmentInitialHeaders = initialHeaders;
+            if (
+              !segmentInitialHeaders &&
```
ztanner: what about when headers exist, but tags are absent? Isn't that another scenario where we produced build output entries without any cache tags?

ijjk (approved):

ztanner (approved):

PR #14696, opened 2026-01-21
Adding/updating environment variables shows warnings for key and values

JohnPhamous:
Goal is to provide better guidance for humans and Agents for adding/updating env vars.

Handles these scenarios that might have unassuming behaviors

1. The value contains a whitespace character
2. The key starts with a prefix where a framework will expose the value to the client
3. The key starts with a public prefix and contains a sensitive keyword (like PASSWORD)

**Value warnings** (whitespace, quotes, newlines, empty):
```
WARN! Value includes surrounding quotes (these will be stored literally)
? How to proceed?
❯ Leave as is
  Re-enter
  Trim whitespace  ← only shown for whitespace warnings
```

**Sensitive key warnings** (e.g., `NEXT_PUBLIC_PASSWORD`):
```
WARN! The NEXT_PUBLIC_ prefix will make PASSWORD visible to anyone visiting your site
? How to proceed?
❯ Leave as is
  Rename to PASSWORD
  Re-enter
```

https://github.com/user-attachments/assets/dab215a5-21b6-4e9b-87b6-afbdfce1ffe8

https://github.com/user-attachments/assets/e285aef9-2335-4009-b56b-f6286770217a

### DIFF
--- a/packages/cli/src/commands/env/add.ts
+++ b/packages/cli/src/commands/env/add.ts
@@ -11,6 +11,14 @@ import readStandardInput from '../../util/input/read-standard-input';
 import param from '../../util/output/param';
 import { emoji, prependEmoji } from '../../util/emoji';
 import { isKnownError } from '../../util/env/known-error';
+import {
+  getEnvValueWarnings,
+  getEnvKeyWarnings,
+  formatWarnings,
+  hasOnlyWhitespaceWarnings,
+  trimValue,
+  removePublicPrefix,
+} from '../../util/env/validate-env';
 import { getCommandName } from '../../util/pkg-name';
 import { isAPIError } from '../../util/errors-ts';
 import { getCustomEnvironments } from '../../util/target/get-custom-environments';
@@ -51,6 +59,7 @@ export default async function add(client: Client, argv: string[]) {
   telemetryClient.trackCliFlagSensitive(opts['--sensitive']);
   telemetryClient.trackCliFlagForce(opts['--force']);
   telemetryClient.trackCliFlagGuidance(opts['--guidance']);
+  telemetryClient.trackCliFlagYes(opts['--yes']);
 
   if (args.length > 3) {
     output.error(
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/env/add.ts
```diff
@@ -161,6 +225,61 @@ export default async function add(client: Client, argv: string[]) {
     });
   }
 
+  // Validate and handle value warnings with re-entry option
+  let finalValue = envValue;
+
+  if (!skipConfirm) {
+    let valueAccepted = false;
```
mehulkar: looks pretty much the same as the procedure for `update`, is it worth consolidating?

packages/cli/src/util/env/validate-env.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (messages.length === 2) {
+    return `Value ${messages[0]} and ${messages[1]}`;
+  }
+  const last = messages.pop();
+  return `Value ${messages.join(', ')}, and ${last}`;
+}
+
+/** Framework prefixes that expose variables to the browser. */
+const PUBLIC_PREFIXES = [
+  'NEXT_PUBLIC_', // Next.js
```
mehulkar: can we get this definition from somewhere else? it must exist right? I know turbo has one too

mehulkar (approved):
Seems like it's not in this PR, but in the first video, there was a prompt `Your value will be encrypted. Mark as sensitive?` I found that very confusing... will it be encrypted if I select yes, or is it already going to be encrypted (in which case, why should I mark it as sensitive??)

Other than that, this code looks good!

JohnPhamous:
> Seems like it's not in this PR, but in the first video, there was a prompt Your value will be encrypted. Mark as sensitive? I found that very confusing... will it be encrypted if I select yes, or is it already going to be encrypted (in which case, why should I mark it as sensitive??)

@mehulkar Yes all environment variables are encrypted. Vercel has [sensitive environment variables](https://vercel.com/docs/environment-variables/sensitive-environment-variables) which means the values can never be read again. This is existing behavior. We can improve the copywriting for sensitive environment variables in another PR.

packages/cli/src/util/env/validate-env.ts
JohnPhamous: Good idea, found it!

packages/cli/src/commands/env/add.ts
JohnPhamous: Done, ty!

PR #14698, opened 2026-01-22
Vercel 

HACKER007-SHIP-IT:
Packard repositoris


HACKER007-SHIP-IT:
import { streamText } from 'ai'

const result = streamText({
  model: 'openai/gpt-5',
  prompt: 'Why is the sky blue?'
})

PR #14703, opened 2026-01-22
[next]: add partial fallback support

ztanner:
This PR adds the `partialFallback` property onto the Prerender document. This will signal different fallback semantics: namely, the expectation with a partial fallback is that it will immediately kick off a background revalidation to produce a more complete static document.

See https://github.com/vercel/next.js/pull/89063 corresponding Next.js changes.


ijjk (approved):

packages/next/src/utils.ts
```diff
@@ -3072,6 +3072,12 @@ export const onPrerenderRoute =
         }
       }
 
+      const partialFallback =
```
ijjk: This needs to be separate pr or it will fail until build utils is bumped in build container

ijjk (approved):

PR #14705, opened 2026-01-22
[cli] Normalize items in the `routes` array to the `routes` syntax

MatthewStanciu:
2 days ago I opened https://github.com/vercel/vercel/pull/14679 to put to rest a gnarly DX issue with vercel.ts where it was very easy to generate a mixed array (—> invalid config) when you had header transforms. See that PR description for a very detailed description of the issue.

The solution proposed in that PR was to always convert everything to the routes format no matter what. This solves all current and future problems related to mixed arrays, but creates a new problem because `routes` and `rewrites`/`redirects`/`headers` have different routing orders in different frameworks, so landing that change would change where in the stack routes are processed.

After chatting with @mknichel, we landed on optimizing for this use case:

```ts
export const config: VercelConfig = {
  routes: [
    routes.rewrite('/test-header', 'https://httpbin.org/headers', {
      requestHeaders: {
        authorization: `Bearer token`,
      },
    }),
    routes.redirect('/test-build', 'https://httpbin.org/headers'),
  ]
};
```

Currently this fails because the `rewrite` gets compiled to the `routes` format since it has a header transform, but the `redirect` compiles to the `redirects` format. So although both are placed in `routes`, it still generates a mixed array.

This PR solves this issue. This is more reasonable than converting everything to `routes`. It does force you to move everything into `routes` if you have a header transform, but this is also the current `vercel.json` behavior, so it is actually more in line with what people expect from vercel.json.

### DIFF
--- a/packages/cli/src/util/compile-vercel-config.ts
+++ b/packages/cli/src/util/compile-vercel-config.ts
@@ -6,12 +6,20 @@ import output from '../output-manager';
 import { NowBuildError } from '@vercel/build-utils';
 import { VERCEL_DIR } from './projects/link';
 import { ConflictingConfigFiles } from './errors-ts';
+import type { RouteWithSrc, Rewrite, Redirect } from '@vercel/routing-utils';
 
-function isRouteFormat(item: any): boolean {
-  return item && typeof item === 'object' && 'src' in item;
+type RouteItemType = 'route' | 'rewrite' | 'redirect';
+type RouteInput = RouteWithSrc | Rewrite | Redirect;
+
+function detectRouteType(item: RouteInput): RouteItemType {
+  if ('src' in item) return 'route';
+  if ('permanent' in item) return 'redirect';
+  return 'rewrite';
 }
 
-function toRouteFormat(item: any, isRedirect: boolean): any {
+function toRouteFormat(item: RouteInput, type: RouteItemType): RouteWithSrc {
+  if (type === 'route') return item as RouteWithSrc;
+
   const {
     source,
     destination,
@@ -19,16 +27,15 @@ function toRouteFormat(item: any, isRedirect: boolean): any {
     permanent,
     respectOriginCacheControl,
     ...rest
-  } = item;
+  } = item as Rewrite & Redirect;
 
-  const route: any = {
+  const route: RouteWithSrc = {
     src: source,
     dest: destination,
     ...rest,
   };
 
-  if (isRedirect) {
-    route.redirect = true;
+  if (type === 'redirect') {
     route.status = statusCode || (permanent ? 308 : 307);
   } else {
     if (respectOriginCacheControl !== undefined) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -97,7 +97,13 @@ export function normalizeConfig(config: any): any {
   }
 
   if (allRoutes.length > 0) {
-    normalized.routes = allRoutes;
+    normalized.routes = allRoutes.map(item => {
+      if (isRouteFormat(item)) {
+        return item;
+      }
+      const isRedirect = 'permanent' in item || 'statusCode' in item;
+      return toRouteFormat(item, isRedirect);
```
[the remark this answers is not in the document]
MatthewStanciu: correct problem, unnecessary test addition in your solution

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -97,7 +97,13 @@ export function normalizeConfig(config: any): any {
   }
 
   if (allRoutes.length > 0) {
-    normalized.routes = allRoutes;
+    normalized.routes = allRoutes.map(item => {
+      if (isRouteFormat(item)) {
+        return item;
+      }
+      const isRedirect = 'permanent' in item;
```
MatthewStanciu, line 104: This seems brittle but it's the only reliable way to determine if it's a redirect without it being inside the `redirects` array. Someone could pass source and dest without the `@vercel/config` sugar and it would be treated as a rewrite, but then they wouldn't be indicating anywhere that they expect it to be a redirect, which matches the behavior of vercel.json. `@vercel/config` automatically adds the `permanent` property with a default value when compiling if you use the `routes.redirect()` sugar. So for the purpose of compiling vercel.ts, it works.

mknichel: `permanent` is optional but always set by `routes.redirect` in `@vercel/config`. Talked in person and Matthew is going to send another PR that refactors this code.

packages/cli/test/unit/util/compile-vercel-config.test.ts
```diff
@@ -175,6 +175,75 @@ describe('normalizeConfig', () => {
...
+          statusCode: 301,
+          permanent: true,
+        },
+      ],
+    };
+
+    const result = normalizeConfig(config);
+
+    expect(result.routes).toEqual([
+      { src: '/old', dest: '/new', redirect: true, status: 301 },
```
mknichel, line 244: `route.redirect` is not a real field

mknichel (approved):

packages/cli/test/unit/util/compile-vercel-config.test.ts
MatthewStanciu: I was going to remove this from all the tests but since we already have the refactor PR https://github.com/vercel/vercel/pull/14707 I did it there instead

PR #14707, opened 2026-01-23
[cli] Mild refactor of compile-vercel-config

MatthewStanciu:
Follow-up to https://github.com/vercel/vercel/pull/14705 which does a small refactor of the parts of `compile-vercel-config` that that PR touches. Many of these issues were already there but should be fixed. In particular, this PR:
- Removes all references to `redirect: true`; this property was hallucinated by unrelated changes a month ago and does not exist
- Adds proper types - removes all the `any`s
- Removes unnecessary route type detection


MatthewStanciu:
Tested all behavior still works as expected

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -3,61 +3,48 @@ import { join, basename } from 'path';
...
 import { VERCEL_DIR } from './projects/link';
 import { ConflictingConfigFiles } from './errors-ts';
+import { NowBuildError } from '@vercel/build-utils';
+import type { RouteWithSrc, Rewrite, Redirect } from '@vercel/routing-utils';
+import type { VercelConfig } from '@vercel/client';
 
-function isRouteFormat(item: any): boolean {
-  return item && typeof item === 'object' && 'src' in item;
-}
+type RouteInput = RouteWithSrc | Rewrite | Redirect;
```
mknichel: This should support `Header` also.

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -70,21 +57,22 @@ function normalizeArrayField(
  * If routes and rewrites/redirects are BOTH explicitly defined,
  * returns unchanged to let schema validation fail.
  */
-export function normalizeConfig(config: any): any {
+export function normalizeConfig(config: VercelConfig): VercelConfig {
   const normalized = { ...config };
-  let allRoutes: any[] = normalized.routes || [];
+  const { rewrites, redirects } = normalized;
```
mknichel: This should support `headers` too, right? Or is that not an error when you specify `headers` and `routes`?

packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -3,61 +3,48 @@ import { join, basename } from 'path';
...
-): any[] | null {
+  items: RouteInput[] | undefined
+): RouteWithSrc[] | null {
   if (!items || !Array.isArray(items) || items.length === 0) {
     return null;
   }
 
-  const hasRouteFormat = items.some(isRouteFormat);
+  const hasRouteFormat = items.some(item => 'src' in item);
   if (!hasRouteFormat) {
```
mknichel: If `normalizeArrayField` is being called, shouldn't this function always convert it to `RouteWithSrc` format? It's confusing to me that `normalizeConfig` doesn't make the decision that the routes need to be converted and this function just normalizes all the routes.

MatthewStanciu: good catch, this is a relic from previous iterations of the logic. i removed `normalizeArrayField` here: https://github.com/vercel/vercel/pull/14707/changes/bc5ce93db987e91d66167cba86e22ea9aac34521

packages/cli/src/util/compile-vercel-config.ts
MatthewStanciu: added header support!

packages/cli/src/util/compile-vercel-config.ts
MatthewStanciu: added header support!

mknichel (approved):

PR #14709, opened 2026-01-23
[cli] Throw explicit error when vercel.ts `routes` conversion creates a conflict

MatthewStanciu:
https://github.com/vercel/vercel/pull/14705 fixes this use case of vercel.ts:

```ts
// ✅ fixed
export const config: VercelConfig = {
  routes: [
    routes.rewrite('/test-header', 'https://httpbin.org/headers', {
      requestHeaders: {
        authorization: `Bearer token`,
      },
    }),
    routes.redirect('/test-build', 'https://httpbin.org/headers'),
  ]
};
```

But the user can still have a confusing experience if they do this:

```ts
// ❌ not fixed
export const config: VercelConfig = {
  rewrites: [
    routes.rewrite('/test-env', 'https://httpbin.org/headers', {
      requestHeaders: {
        authorization: `Bearer ${deploymentEnv('API_TOKEN')}`,
      },
    }),
    routes.rewrite('/test-rewrite-regular', 'https://httpbin.org/headers')
  ],
  redirects: [
    routes.redirect('/test-build', 'https://httpbin.org/headers'),
  ]
};
```

This will fail with the error message:

```
Error: If `rewrites`, `redirects`, `headers`, `cleanUrls` or `trailingSlash` are used, then `routes` cannot be present.
```

But that's not clear because from the user's perspective they separated their rewrites and redirects like the docs say, so where is `routes` coming from?

We're not going to compile the `redirects` to routes (https://github.com/vercel/vercel/pull/14679), so we need the error message to be clearer.

This PR fails the build during config compilation if after normalization (AKA compiling anything to `routes` that needs this transformation) there's still a conflict. In this case, since it has the full context of what just happened (unlike the generic vercel.json error it currently throws) it can throw a more descriptive error message with clear instructions for what to do.

Screenshot 2026-01-23 at 5 12 34 PM


packages/cli/src/util/compile-vercel-config.ts
```diff
@@ -74,6 +74,23 @@ export function normalizeConfig(config: VercelConfig): VercelConfig {
   const convertedRewrites = normalizeArrayField(rewrites);
   const convertedRedirects = normalizeArrayField(redirects);
 
+  if (
+    hasRewrites &&
+    hasRedirects &&
+    Boolean(convertedRewrites) !== Boolean(convertedRedirects) // one of them was converted to routes
```
mknichel: Instead of implicitly detecting whether `normalizeArrayField` changed the array, can we make a specific boolean that checks the condition when all routes should be converted to the `routes` field and use that here?

mknichel (approved):

PR #14728, opened 2026-01-26
Revert "[python] remove pip from isInstalled check"

gscho:
Reverts vercel/vercel#14712

Removing `pip` from the `isInstalled` check breaks the "fallback to 3.12" behaviour


tknickman (approved):
approving revert

nsidnev (approved):

PR #14730, opened 2026-01-26
[backends] Improve esm/cjs interop 

jeffsee55:
This PR does 2 things:

1. When importing from a CJS dependency, Rolldown wasn't transforming named imports properly because we're intentionally leaving `node_modules` untouched. As a fix, we now detect an import of a CJS dependency and replace it with a virtual shim file. The shims are CJS files which reexport the given CJS dependency, but unlike the dependency, they _are_ transformed by Rolldown.

1. We use NFT to trace `node_modules`. This used to trace from the Rolldown build result, but there are some inconsistencies with how `node_modules` are resolved when traced from the `.vercel` folder, so this now traces the source code. As a result, it needs to transform Typescript files, so we're doing that with `oxc-transform` as well.

Incidentally, the addition of the rolldown plugin logic meant that it was now trivial to treat workspace deps as if they were part of the project, so this PR also supports handling of `.ts` files from other workspaces in a monorepo. However introspection will still fail in some cases for monorepos, which will be fixed in a later PR.


packages/backends/src/build.ts
```diff
@@ -83,23 +84,57 @@ export const doBuild = async (
...
+            handler,
+            tsPromise,
+            files: buildResult.rolldownResult.outputFiles,
+          };
+        }
       }
 
       await writeFile(cervelJsonPath, JSON.stringify({ handler }, null, 2));
 
+      const files = await nodeFileTrace({
```
mehulkar, line 114: should we put a span around this? curious how it impacts perf

packages/cervel/src/node-file-trace.ts
```diff
@@ -0,0 +1,96 @@
+import { relative, join } from 'node:path';
+import { readFile, lstat } from 'node:fs/promises';
+import { isNativeError } from 'node:util/types';
+import { FileFsRef, glob, debug, type Files } from '@vercel/build-utils';
+import {
+  nodeFileTrace as nft,
+  resolve as nftResolveDependency,
+} from '@vercel/nft';
+import { transform } from 'oxc-transform';
```
mehulkar, line 9: guessing this is a native dep. will it work in build container?

packages/cervel/src/node-file-trace.ts
```diff
@@ -0,0 +1,96 @@
...
+      const relativeFile = relative(args.repoRootPath, file);
+      result.fileList.delete(relativeFile);
+    }
+  }
+
+  debug('NFT traced files count:', result.fileList.size);
+
+  for (const file of result.fileList) {
+    const absolutePath = join(args.repoRootPath, file);
+    const stats = await lstat(absolutePath);
```
mehulkar, line 83: should we parallelize these async functions instead of `await` on each one?

mehulkar (approved):

packages/backends/src/build.ts
jeffsee55: Will add in a follow up 👍

packages/cervel/src/node-file-trace.ts
jeffsee55: It's running on the framework tests and I ran it against the workflow app as well so seems ok, we also have native deps for rolldown and esbuild in other places

packages/cervel/src/node-file-trace.ts
jeffsee55: I had this, but I actually don't know, does it matter to Promise.all potentially thousands of files?

jeffsee55: Will follow up on this as well.

PR #14736, opened 2026-01-26
feat(cli): minimize update prompt in tty

tknickman:
Minimize update prompt when being asked after command completion.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -949,7 +949,7 @@ main()
             choices: [
               { key: 'u', name: 'Upgrade now', value: 'upgrade' },
               { key: 'c', name: 'View changelog', value: 'changelog' },
-              { key: 'n', name: 'Skip', value: 'skip' },
+              { key: 's', name: 'Skip', value: 'skip' },
             ],
           });
 
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -77,6 +77,7 @@ import { checkTelemetryStatus } from './util/telemetry/check-status';
 import output from './output-manager';
 import { checkGuidanceStatus } from './util/guidance/check-status';
 import { determineAgent } from '@vercel/detect-agent';
+import open from 'open';
 
 const VERCEL_DIR = getGlobalPathConfig();
 const VERCEL_CONFIG_PATH = configFiles.getConfigFilePath();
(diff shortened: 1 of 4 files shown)


packages/cli/src/util/upgrade.ts
```diff
@@ -28,6 +39,20 @@ export async function executeUpgrade(): Promise<number> {
...
+        // Show output only on error
+        const stdoutStr = Buffer.concat(stdout).toString();
+        const stderrStr = Buffer.concat(stderr).toString();
+        if (stdoutStr) {
+          output.print(stdoutStr);
+        }
+        if (stderrStr) {
+          output.print(stderrStr);
+        }
+        output.error(`Upgrade failed with exit code ${code ?? 1}`);
```
mehulkar: default to 1 seems bad, i would default to 'unknown' or just 'null'

packages/cli/src/index.ts
```diff
@@ -916,38 +917,69 @@ const main = async () => {
...
-          box(
-            `Update available! ${chalk.gray(`v${pkg.version}`)} ≫ ${chalk.green(
-              `v${latest}`
-            )}
+
+        if (isTTY) {
+          // Interactive mode: show condensed prompt
+          const errorMsg =
+            exitCode && exitCode !== 2
+              ? chalk.magenta(
```
mehulkar, line 933: project-kill-chalk

packages/cli/src/index.ts
```diff
@@ -916,38 +917,69 @@ const main = async () => {
...
+            )} → ${chalk.green(`v${latest}`)})${errorMsg}\n`
+          );
+
+          const action = await client.input.expand({
+            message: 'What would you like to do?',
+            default: 'u',
+            choices: [
+              { key: 'u', name: 'Upgrade now', value: 'upgrade' },
+              { key: 'c', name: 'View changelog', value: 'changelog' },
+              { key: 'n', name: 'Skip', value: 'skip' },
```
mehulkar: why not `s` for skip?

packages/cli/src/index.ts
```diff
@@ -916,38 +917,69 @@ const main = async () => {
...
+
+          if (action === 'upgrade') {
+            const upgradeExitCode = await executeUpgrade();
+            process.exitCode = upgradeExitCode;
+            return;
+          } else if (action === 'changelog') {
+            await open(changelog);
+          }
+        } else {
+          // Non-interactive mode: show full update box
```
mehulkar, line 964: why full update box? condense it in both cases?

mehulkar (approved):

packages/cli/src/index.ts
tknickman: the full update box contains the command to run, so if we're not going to prompt you to run it we should show it to you so you know what to run.

packages/cli/src/index.ts
tknickman: Yea I started with Y/N and left that one. Good catch, thanks!

packages/cli/src/util/upgrade.ts
tknickman: Yep good call. Fixed

PR #14737, opened 2026-01-26
[cli][client] Add random skew to Retry-After logic

bgw:
We're still seeing rate limit errors in the Next.js e2e deployment tests:

E.g. https://github.com/vercel/next.js/actions/runs/21373542150/job/61524648707#step:33:270

This PR is a continuation of the efforts to fix rate limiting in:
- https://github.com/vercel/vercel/pull/14407
- https://github.com/vercel/vercel/pull/14443

My hypothesis with the current remaining 429 errors is:
* We've got a retry loop everywhere this API is called
* We use the HTTP `Retry-After` header to decide when to retry
* The backend doesn't use fixed 1 minute buckets, but rather a sliding window.
* The backend's `Retry-After` header tells us when the next rate limit token will be available.
* Every parallel deployment then ends up waiting on that one next rate limit token and there's a thundering herd when trying to recover from 429 errors.

So I think the retry logic on the client has to be to read the `Retry-After` header and then add a random duration to it, up to one minute, to prevent a thundering herd.

## Alternate Solutions

We could add this random skew on the backend. That might be better as clients don't have to be aware of this potential problem.

But also: What the backend is doing is technically correct. And this isn't really about the backend protecting itself, but rather about preventing clients from running out of retries.

### DIFF
--- a/packages/client/src/check-deployment-status.ts
+++ b/packages/client/src/check-deployment-status.ts
@@ -22,7 +22,7 @@ interface DeploymentStatus {
 }
 
 // If an error occurs, how should our retries behave?
-const RETRY_COUNT = 3;
+const RETRY_COUNT = 5;
 // Maximum value to cap `Retry-After` to in order to avoid hanging if we get a
 // `Retry-After` value in the far future. This limit is applied before
 // `RETRY_DELAY_SKEW_MS`, so the total duration can exceed this.
--- a/packages/client/src/check-deployment-status.ts
+++ b/packages/client/src/check-deployment-status.ts
@@ -23,8 +23,13 @@ interface DeploymentStatus {
 
 // If an error occurs, how should our retries behave?
 const RETRY_COUNT = 3;
+// Maximum value to cap `Retry-After` to in order to avoid hanging if we get a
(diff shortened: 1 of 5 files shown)


bgw:
* **#14737** <a href="https://app.graphite.com/github/pr/vercel/vercel/14737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/14737?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/client/src/check-deployment-status.ts
```diff
@@ -23,8 +23,13 @@ interface DeploymentStatus {
 
 // If an error occurs, how should our retries behave?
 const RETRY_COUNT = 3;
```
lukesandberg: for check-deployment-status should we allow for more retries?   this is a somewhat spammy api  since we are querying every 5 seconds.

lukesandberg (approved):

ijjk (approved):

PR #14739, opened 2026-01-27
[experimental-services] vercel.json multi-service builds

ricardo-agz:
Based off of https://github.com/vercel/vercel/pull/14617 except without the zero-config auto-detection

Implements the logic to actually run multiple builds when `experimentalServices` is configured in `vercel.json` and the framework preset is set to "services"


packages/fs-detectors/src/services/detect-services.ts
```diff
@@ -0,0 +1,131 @@
...
+      ? builderSrc
+      : `/${builderSrc}`;
+
+    // Web services
+    if (routePrefix === '/') {
+      // Root service: catch-all route
+      defaults.push({
+        src: '^/(.*)$',
+        dest: functionPath,
+        check: true,
```
mehulkar, line 121: what does `check: true` do?

packages/fs-detectors/test/unit.detect-services.test.ts
```diff
@@ -56,10 +62,12 @@ describe('detectServices', () => {
             frontend: {
               workspace: 'apps/web',
               framework: 'nextjs',
+              routePrefix: '/',
             },
             api: {
               workspace: 'apps/api',
               entrypoint: 'src/server.ts',
+              routePrefix: '/api',
```
mehulkar, line 70: it would be useful to have a test with to routable services so we can see 2 non-`/` route prefixes tested

packages/fs-detectors/src/services/index.ts
```diff
@@ -1,55 +1,13 @@
...
 
-      const resolved = resolveService(name, serviceConfig);
-      services.push(resolved);
-    }
-  }
+export { getBuilderForRuntime, inferRuntimeFromExtension } from './utils';
 
-  return { services, errors };
-}
+export { getServicesBuilders } from './get-services-builders';
```
mehulkar: nit: would generally stay away from re-exports, they're a huge pain to trace and also have a perf penalty

packages/fs-detectors/src/services/resolve.ts
```diff
@@ -30,24 +103,88 @@ export function resolveService(
...
+    builderUse = RUNTIME_BUILDERS[runtime];
+    builderSrc = config.entrypoint!;
+  } else {
+    const runtime = inferRuntimeFromExtension(config.entrypoint!);
+    builderUse = getBuilderForRuntime(runtime!);
+    builderSrc = config.entrypoint!;
+  }
+
+  // routePrefix is required for web services (validated above), optional for others
+  const routePrefix = config.routePrefix || '/';
```
mehulkar: This default here feels a bit happenstance, since it doesn't guarantee that the right service is getting the fallback behavior. Could we hoist it somewhere where the "default" service resovles the prefix before it gets here?

mehulkar (approved):
some minor comments, otherwise lgtm

packages/fs-detectors/src/services/detect-services.ts
ricardo-agz: it tells the router to verify the destination exists on the filesystem before applying the route
it's not strictly necessary since we control what gets built anyway, mostly kept for consistency:
https://github.com/vercel/vercel/blob/ad16ec997faa460f9fe5035310b0dd3e12c4f7bf/packages/routing-utils/src/superstatic.ts#L186

packages/fs-detectors/src/services/resolve.ts
ricardo-agz: Yea you're right, this is not necessary. Crons and workers will have routing handled internally (still TODO) so they shouldn't accept a routePrefix. Updated this

PR #14740, opened 2026-01-27
[python] Enable python3.13 and 3.14 support

gscho:
## Description

Currently python 3.13 and 3.14 are installed in the build container but they can't be used (unless you use `--prebuilt`).
That's because the `isInstalled` function looks on the path for pip AND python. Now that we've moved to `uv` managed pythons, `pip3.x` is not on the `$PATH` and that makes it fail the `isInstalled` check.

## Summary of Changes

- Used `uv python list` to list installed pythons, filtering out the system python
- Preserves the current behaviour of "falling back" to python3.12 when an unsupported version of python is selected
- Refactors `uv` based functions into its own class
- Check for the expected python version in each test fixture
- Add a fixture to test python3.13 and python3.14 deploys
- If python3.12 satisfies the required python, use that

## Potential User Impact

Users who currently supply a python version with no upper bound will start deploying python 3.14 lambdas instead of 3.12 lambdas. For example if your python version is `>=3.11` you would end up with a 3.14 lambda.

This could break customers who's dependencies are pinned to versions that don't support 3.14. For example one of our test fixtures was pinned to a version of fastapi that breaks on 3.14 because we don't have rust inside our build container. Updating fastapi solved this because the newer version had a wheel available for it.

## TODO

We should start looking at adding a `discontinueDate` for pythons we no longer support


packages/python/src/version.ts
elprans, line 12: This is inconsistent with how `uv` would handle open Python pins (`>=3.12`).  I think we should either: 1) continue to pin 3.12, but then we'll need to rewrite `.python-version` or bump this to 3.14.  I think we should bite the bullet and bump the default to 3.14 here.

gscho: @elprans AFAIK `.python-version` is not currently used to determine which python a project needs. Just `pyproject.toml` and `Pipfile.lock`. Not sure I understand the inconsistency. Python 3.12 is selected as the default because the potential customer impact is smaller. Its also the default version linked to the path in the build-container currently.

elprans: > AFAIK `.python-version` is not currently used to determine which python a project needs. Just `pyproject.toml` and `Pipfile.lock`

The CLI would not use it to determine the version, but `uv` itself would!  You'll end up with an inconsistency -- `vc` thinks we use 3.12 and deploys the project as such, but `uv` would have installed the deps for 3.14.

gscho: Bumping the default caused 7 of our e2e tests to fail due to 500 errors with the deployed functions. I think we have a much greater chance of breaking people with 3.14 as the default.

gscho: > The CLI would not use it to determine the version, but uv itself would! You'll end up with an inconsistency -- vc thinks we use 3.12 and deploys the project as such, but uv would have installed the deps for 3.14.

Can we use the Major.Minor version defined in the `.python-version` file over `pyproject.toml` and `Pipfile.lock` when it exists then?

elprans: That's what https://github.com/vercel/vercel/pull/14549 does.  Perhaps we should merge it.  @ricardo-agz, WDYT?

gscho: @elprans Take a look at https://github.com/vercel/vercel/pull/14740/commits/2fa3bb4c3f0f74c31648b2a6408b342be7b8ed37 maybe we ship this first and then switch to #14549 in a follow up?

elprans: @gscho, plausible, but we also want to actually *write* a .python-version with DEFAULT_PYTHON_VERSION iff one does not exist and the range in `pyproject.toml` includes 3.12.  This way we can ensure consistency with uv selecting 3.12 by default even if you specified `>=3.12` in your manifest.

packages/python/src/index.ts
```diff
@@ -270,6 +270,22 @@ export const build: BuildV3 = async ({
...
   });
 
+  // Write a .python-version file on behalf of the user when:
+  // no .python-version file exists and the required version in pyproject.toml
+  // is <= DEFAULT_PYTHON_VERSION
+  if (
+    !pythonVersionFileDir &&
+    pyprojectDir &&
+    declaredPythonVersion?.source === 'pyproject.toml' &&
+    parseFloat(pythonVersion.version) <= parseFloat(DEFAULT_PYTHON_VERSION)
```
elprans: Neat, but this would break horribly if there is a patch version in the version somehow and also float comparison might be fiddly in general and compare to false when it should be true.  Use `parseVersionTuple` and `compareTuples` or `satisfies` from version.ts instead

elprans (approved):

PR #14754, opened 2026-01-27
[next] Remove 404.html prerenders from functions

mischnic:
Closes PACK-6560

They were only there as an optimization, but introduce nondeterminism because the HTML contains the build id and deployment id

### DIFF
--- a/packages/next/src/index.ts
+++ b/packages/next/src/index.ts
@@ -87,7 +87,6 @@ import {
   normalizePackageJson,
   normalizePage,
   onPrerenderRoute,
-  onPrerenderRouteInitial,
   PseudoFile,
   PseudoLayer,
   PseudoLayerResult,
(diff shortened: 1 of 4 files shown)


packages/next/src/index.ts
```diff
@@ -1336,33 +1335,33 @@ export const build: BuildV2 = async buildOptions => {
...
-        ? path.posix.join(entryDirectory, i18n.defaultLocale, '404')
-        : undefined;
-    }
-
-    if (!hasStatic500 && i18n) {
-      hasStatic500 =
-        !!staticPages[
-          path.posix.join(entryDirectory, i18n.defaultLocale, '500')
-        ];
-    }
```
ijjk, line 1365: Let's just remove from being included in the function here https://github.com/vercel/vercel/blob/ad16ec997faa460f9fe5035310b0dd3e12c4f7bf/packages/next/src/server-build.ts#L742-L747

mischnic: done

ijjk (commented):
Maybe we add a couple assertions in the integrations tests ensuring the 404.html doesn't slip into any lambda's files

ijjk (approved):

PR #14755, opened 2026-01-27
[cli] Use promote flow when targeting production alias

onsclom:
`vc alias set [alias] [production alias]` causes problems when users expect that to promote the alias. For example, https://linear.app/vercel/issue/PIPE-5641/unexpected-rollback-when-an-aliased-domain-is-edited-oror-00922719.

This updates `vc alias set` so that if we detect the target alias is a production alias, we use the promote flow.

image

### DIFF
--- a/packages/cli/src/commands/alias/set.ts
+++ b/packages/cli/src/commands/alias/set.ts
@@ -2,6 +2,7 @@ import chalk from 'chalk';
 import type { SetDifference } from 'utility-types';
 import type { AliasRecord } from '../../util/alias/create-alias';
 import * as ERRORS from '../../util/errors-ts';
+import { ProjectNotFound } from '../../util/errors-ts';
 import assignAlias from '../../util/alias/assign-alias';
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/alias/set.ts
```diff
@@ -379,3 +400,42 @@ function getTargetsForAlias(args: string[], { alias }: VercelConfig = {}) {
...
+  if (project instanceof ProjectNotFound) {
+    return null;
+  }
+
+  if (!isProductionAlias(project, aliasTarget)) {
+    return null;
+  }
+
+  output.log(
+    `Alias ${chalk.bold(aliasTarget)} is a production alias. Using promote flow.`
```
tknickman: This might be a little more concise:

```suggestion
    `${chalk.bold(aliasTarget)} is a production alias - promoting.`
```

tknickman (commented):
Just one note, but looks good!

tknickman (approved):

PR #14762, opened 2026-01-28
experimental "ruby" runtime framework preset

ricardo-agz:
* Adds zero config support for the generic Ruby runtime framework preset (feature flagged)
* Fixes small bug in the Ruby version parsing where `ruby "~> 3.3.x"` in Gemfile would fail due to a trailing space not being trimmed after removing the `~>` prefix.
* Sets RACK_ENV env var to 'production' (necessary for sinatra to work correctly)

https://vercel-plus-sinatra.vercel.app/


packages/frameworks/src/frameworks.ts
```diff
@@ -4121,6 +4121,56 @@ export const frameworks = [
       },
     ],
   },
+  {
+    name: 'Ruby',
+    slug: 'ruby',
+    experimental: true,
+    runtimeFramework: true,
+    logo: 'https://api-frameworks.vercel.sh/framework-logos/ruby.svg',
```
mehulkar, line 4129: TIL. does the file you're adding in this PR somehow get deployed here?

mehulkar (approved):

packages/frameworks/src/frameworks.ts
ricardo-agz: I actually just now found out where that code for that lives lol. But yes, there's a build script that serves them as static files for the Vercel app. For this deployment:
https://vercel-o4rgy2ei6.vercel.sh/framework-logos/ruby.svg

PR #14763, opened 2026-01-28
[cli] move builders to peer dependencies

mehulkar:
- Moves builders to peerDependencies and tells pnpm that they're optional dependencies. With this config, `pnpm` will not bundle them into the publishd `vercel` CLI, but they will be in the manifest. 
- Add a feature to the `resolveBuilders` function that reads the peerDependencies section of package.json at runtime and installs that correct version of the builder. It stores this builder in a local cache (`.vercel/builders`) so it only has to get them from the registry once.


packages/cli/src/output-manager.ts
```diff
@@ -1,7 +1,7 @@
 import { Output } from './util/output';
 
 const output = new Output(process.stderr, {
-  debug: false,
+  debug: true,
```
LukeSheard, line 4: why did this change? Maybe it should be an env var?

tknickman: Agree, was this intentional?

packages/cli/src/util/build/import-builders.ts
```diff
@@ -64,19 +68,41 @@ export async function importBuilders(
...
+function getPeerDependencies(): Record<string, string> {
+  if (!peerDependencies) {
+    try {
+      const cliPkgPath = require_.resolve('vercel/package.json', {
+        paths: [__dirname],
+      });
+      const cliPkg = require_(cliPkgPath) as PackageJson;
+      peerDependencies =
+        (cliPkg.peerDependencies as Record<string, string>) || {};
+    } catch {
```
tknickman: I would at least log an error here if the require fails (maybe json is invalid?)

packages/cli/src/output-manager.ts
mehulkar: no, i'll revert this before merging, thanks!. 

I was going to run some more tests in API  + build container after reviews / any adjustments

jeffsee55 (approved):

tknickman (approved):

packages/cli/src/util/build/import-builders.ts
mehulkar: done

PR #14765, opened 2026-01-28
experimental rust runtime framework preset

ricardo-agz:
Adds zero config support for the generic Rust runtime framework preset (feature flagged)


examples/axum/public/index.html
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
```
elprans, line 1: Are we intentionally moving away from `vercel/examples` to hosting examples here?

ricardo-agz: Each framework gets an example in here which is what we clone when you do `vc init {framework}`
`vercel/examples` also has examples that are not just the framework starters

elprans (approved):

mehulkar (approved):

PR #14783, opened 2026-01-29
[next] Strip routes-manifest.json for determinism

mischnic:
Closes PACK-6754

In minimal mode, these fields aren't needed at runtime (they were already incorporated into the config.json for Vercel CDN)

### DIFF
--- a/packages/next/src/server-build.ts
+++ b/packages/next/src/server-build.ts
@@ -1231,22 +1231,36 @@ export async function serverBuild({
 
       const updatedManifestFiles: { [name: string]: FileBlob } = {};
 
-      if (isCorrectManifests) {
-        // filter dynamic routes to only the included dynamic routes
-        // in this specific serverless function so that we don't
-        // accidentally match a dynamic route while resolving that
-        // is not actually in this specific serverless function
-        for (const manifest of [
-          'routes-manifest.json',
-          'server/pages-manifest.json',
-          ...(appPathRoutesManifest ? ['server/app-paths-manifest.json'] : []),
-        ] as const) {
-          const fsPath = path.join(entryPath, outputDirectory, manifest);
+      for (const manifest of isCorrectManifests
+        ? [
+            'routes-manifest.json',
+            'server/pages-manifest.json',
+            ...(appPathRoutesManifest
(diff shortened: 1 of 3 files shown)


packages/next/src/server-build.ts
```diff
@@ -1187,22 +1187,36 @@ export async function serverBuild({
 
       const updatedManifestFiles: { [name: string]: FileBlob } = {};
 
-      if (isCorrectManifests) {
```
ijjk, line 1234: This should all still be gated on this as it's a version check and we can't strip for older versions

mischnic: I didn't change anything on the nextjs side here to not use it, but yeah we can gate it
- the deployment id was recently added in https://github.com/vercel/vercel/pull/14497/changes just to forward the deploymentId in to the builder, it's never read by Next.js
- the headers should never have been used by Next.js?

mischnic: The whole thing is now version gated
And the routes filtering is additionally behind `mayFilterManifests` which is `!experimentalAllowBundling`

packages/next/test/integration/integration-2.test.js
```diff
@@ -778,3 +778,48 @@ describe('action-headers', () => {
...
+  it('should strip routes-manifest', async () => {
+    let originalManifest = JSON.parse(
+      await fs.readFile(
+        path.join(workPath, '.next', 'routes-manifest.json'),
+        'utf8'
+      )
+    );
+    expect(originalManifest.deploymentId).toBeDefined();
+    // Not merged yet in Next.js
+    // expect(originalManifest.headers.length).not.toBe(0);
```
ijjk: Why would this be changed in the original manifest? `next-server` and similar do still need this field

ijjk (approved):
Change looks good other than comment

packages/next/test/integration/integration-2.test.js
mischnic: it's just that the test case is a noop right now
before that Next.js PR (pull request, not manifest) is merged, `headers` is empty. so I can't test if the builder indeed replaces the `headers` with an empty array because it empty either way.

PR #14769, opened 2026-01-28
feat(cli): Add confirmation prompt for DELETE API operations

tknickman:
## Summary

- Adds user confirmation prompts for DELETE API operations in the CLI
- Adds `--dangerously-skip-permissions` flag to the `api` command to bypass confirmation prompts
- When running under an AI agent (detected via `@vercel/detect-agent`), displays a warning to stderr when DELETE confirmation is bypassed
- In non-TTY mode, DELETE operations fail with an error unless `--dangerously-skip-permissions` is used

### New Flag for `vercel api` Command
- `--dangerously-skip-permissions` - Skips confirmation prompts for DELETE operations (use with caution)

### Behavior
1. **DELETE operations** now require interactive confirmation before proceeding
2. **Agent mode warning**: When an AI agent bypasses DELETE confirmation using `--dangerously-skip-permissions`, a prominent warning is displayed to stderr
3. **Non-TTY mode**: DELETE operations fail with a helpful error message unless `--dangerously-skip-permissions` is specified

### Example Usage
```bash
# Will prompt for confirmation
vercel api /v13/deployments/dpl_abc123 -X DELETE

# Skips confirmation (shows warning when run by AI agent)
vercel api /v13/deployments/dpl_abc123 -X DELETE --dangerously-skip-permissions
```

### Files Changed
- `packages/cli/src/util/client.ts` - Added `confirmMutatingOperation()` method and agent/permission flags
- `packages/cli/src/index.ts` - Pass agent detection to Client
- `packages/cli/src/commands/api/command.ts` - Added `--dangerously-skip-permissions` option
- `packages/cli/src/commands/api/index.ts` - Parse flag and set client property
- `packages/cli/src/util/telemetry/commands/api/index.ts` - Added telemetry tracking for new flag
- `packages/cli/test/unit/util/client-confirmation.test.ts` - Added 15 unit tests
- `packages/cli/test/mocks/client.ts` - Added reset for new flags

## Test plan
- [x] Build passes
- [x] Type checking passes
- [x] 15 unit tests covering all scenarios pass
- [ ] Manual testing of DELETE confirmation prompt
- [ ] Manual testing with `--dangerously-skip-permissions` flag

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14784, opened 2026-01-29
fix(cli): Move DELETE confirmation prompt to `vercel api` command only

developerjhp:
## Problem

PR #14769 introduced a DELETE confirmation prompt for AI agent safety, but incorrectly placed the `confirmMutatingOperation()` call inside `client.fetch()`. This caused **ALL commands using DELETE API calls** (e.g., `vercel env rm`, `vercel remove`) to fail in non-TTY environments (CI/CD pipelines).

### Error Message
```
Error: DELETE operations require confirmation. Use --dangerously-skip-permissions to skip confirmation in non-interactive mode.
```

### Root Cause
1. `confirmMutatingOperation()` was called on every `client.fetch()` with DELETE method
2. In non-TTY mode, it requires `--dangerously-skip-permissions` flag to proceed
3. But this flag was only added to the `vercel api` command
4. Other commands like `env rm` have their own `--yes` flag for confirmation, which became ineffective

## Solution

Move the DELETE confirmation logic to only apply to `vercel api` command:

1. Remove `confirmMutatingOperation()` call from `client.fetch()`
2. Change `confirmMutatingOperation()` from private to public method
3. Call `confirmMutatingOperation()` directly in `vercel api`'s `executeSingleRequest()`
4. Update tests to call `confirmMutatingOperation()` directly instead of through `fetch()`

### `packages/cli/src/util/client.ts`
- Remove confirmation call from `fetch()` method (6 lines removed)
- Change `confirmMutatingOperation` visibility from `private` to `public`

### `packages/cli/src/commands/api/index.ts`
- Add `confirmMutatingOperation()` call at the start of `executeSingleRequest()`
- Place it inside try-catch block for proper error handling

### `packages/cli/test/unit/util/client-confirmation.test.ts`
- Update all tests to call `client.confirmMutatingOperation()` directly
- Remove mock endpoint setup (no longer needed)
- Update assertions from response objects to boolean values

## Testing

- [x] All 15 unit tests pass
- [x] `vercel env rm` now works with `--yes` flag in CI/CD environments
- [x] `vercel api -X DELETE` still prompts for confirmation as intended

## Related

- Fixes regression introduced in #14769

### DIFF
--- a/.changeset/fix-env-rm-non-tty.md
+++ b/.changeset/fix-env-rm-non-tty.md
@@ -1,0 +1,7 @@
+---
+"vercel": patch
+---
+
+fix(cli): Move DELETE confirmation prompt to `vercel api` command only
+
+The DELETE confirmation prompt introduced in #14769 was incorrectly applied to all `client.fetch()` calls, which broke commands like `vercel env rm` in non-TTY environments (CI/CD). This fix moves the confirmation logic to only apply to the `vercel api` command as originally intended, allowing other commands with their own confirmation flows (like `--yes` flag) to work properly in non-interactive mode.
--- a/packages/cli/src/commands/api/index.ts
+++ b/packages/cli/src/commands/api/index.ts
@@ -237,6 +237,15 @@ async function executeSingleRequest(
   flags: ParsedFlags
 ): Promise<number> {
   try {
+    // Confirm DELETE operations before proceeding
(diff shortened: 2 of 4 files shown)


tknickman:
Thanks for the fix! I pulled this in and shipped as part of vercel@50.9.2

PR #14785, opened 2026-01-29
[CLI] Add Command to get an OIDC token

marcgreenstock:
Adds the command `vercel project token [name]` which will retrieve the development OIDC token for the project.

If run in the context of a project no args are necessary, if run outside the `name` argument (project name) is required.

Returns the token and a line break so it's easy to parse.


packages/cli/src/commands/project/token.ts
```diff
@@ -0,0 +1,65 @@
...
+          source: 'vercel-cli',
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    );
+    output.print(res.token);
+    output.print('\n');
+    return 0;
```
gr2m, line 57: what about the `com.vercel.token/${projectId}.json` file? Should we check if it contains a valid token before fetching a new one? Or are the `com.vercel.token/${projectId}.json` files no longer used?

marcgreenstock: That I think can remain an implementation of the client. i.e. `@vercel/oidc` can continue to persist it https://github.com/vercel/vercel/blob/main/packages/oidc/src/token-util.ts#L133-L162

dvoytenko (approved):

gr2m:
Current output is 

```
$ vc-local project token my-app
> Vercel CLI 50.23.0
token here
```

I changed the `> Vercel CLI 50.23.0` output to be only shown on `output.debug` (a4ebb23) - is that an acceptable change?  I'd like `vc-local project token my-app` to have no other output other than the token

gr2m:
When I run `vc-local project token` in a directory that is not setup it prompts me

```
? Set up “~/code/vercel/vercel”? (Y/n)
```

Does that make sense for that sub command? Or should it exit with 1 and log out that the folder is not setup?

marcgreenstock:
> When I run `vc-local project token` in a directory that is not setup it prompts me
> 
> ```
> ? Set up “~/code/vercel/vercel”? (Y/n)
> ```
> 
> Does that make sense for that sub command? Or should it exit with 1 and log out that the folder is not setup?

Mmm... perhaps it should exit 1 rather than prompting... this command isn't intended to be used manually.

PR #14776, opened 2026-01-29
[services] new project flow

ricardo-agz:
Implements the new project display and setup for services projects, e.g. projects with `experimentalServices` in `vercel.json`
This is gated under a `VERCEL_USE_EXPERIMENTAL_SERVICES` env var

Screenshot 2026-01-28 at 5 09 03 PM

PR #14805, opened 2026-01-30
[services] add support for `vercel dev`

nsidnev:
Implements automatic discovery of multi-service layouts and services orchestrator for `vercel dev` for the new project flow from https://github.com/vercel/vercel/pull/14776:

image


ricardo-agz (approved):
Nice
This can be a separate PR, but it would be good to add dev support for generic Python apps as well (when framework is null but runtime is python). Sort of like this other [generic "pyproject app" idea ](https://github.com/vercel/vercel/pull/14383) where we consolidate `vc_init_dev_asgi.py` and `vc_init_dev_wsgi.py` and run the Python dev server for everything even non Flask and FastAPI apps

packages/cli/src/util/dev/server.ts
```diff
@@ -927,7 +934,21 @@ export default class DevServer {
     this._address = new URL(replaceLocalhost(address));
 
     const vercelConfig = await this.getVercelConfig();
-    const devCommandPromise = this.runDevCommand();
+
+    let devCommandPromise: Promise<void> | undefined;
+    if (this.services && this.services.length > 1) {
```
ricardo-agz: Do we want to opt out of this behavior when the project is a "services" project but we only have 1 service?

PR #14807, opened 2026-01-30
[next] Fix assertion for cacheComponents with getStaticProps route

mischnic:
Fixes this invariant when trying to `vc build` a Next.js app with a Pages `getStaticProps` route and `cacheComponents: true`

```
Error: invariant: htmlFsRef != null && jsonFsRef != null /gsp
```

### DIFF
--- a/packages/next/src/utils.ts
+++ b/packages/next/src/utils.ts
@@ -2725,7 +2725,9 @@ export const onPrerenderRoute =
       !isBlocking &&
       (!isNotFound || static404Page) &&
       dataRoute &&
-      (!isAppClientParamParsingEnabled || prefetchDataRoute)
+      // When this is an App route, either isAppClientParamParsingEnabled is disabled
+      // or there's a prefetchDataRoute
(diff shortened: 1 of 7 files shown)


packages/next/src/utils.ts
```diff
@@ -2725,7 +2725,9 @@ export const onPrerenderRoute =
       !isBlocking &&
       (!isNotFound || static404Page) &&
       dataRoute &&
-      (!isAppClientParamParsingEnabled || prefetchDataRoute)
+      // When this is an App route, either isAppClientParamParsingEnabled is disabled
+      // or there's a prefetchDataRoute
+      (!isAppPathRoute || !isAppClientParamParsingEnabled || prefetchDataRoute)
```
mischnic, line 2730: I have no idea what `isAppClientParamParsingEnabled` means.
But the when enabling cache components, this becomes `true`. Which shouldn't make a difference to how Pages routes are handled?

ztanner (approved):

PR #14819, opened 2026-01-30
[services] static-build frontends support

ricardo-agz:
Adds support for static build services and mounting them at route prefixes

### DIFF
--- a/packages/fs-detectors/src/services/detect-services.ts
+++ b/packages/fs-detectors/src/services/detect-services.ts
@@ -71,14 +71,13 @@ export async function detectServices(
 /**
  * Generate routing rules for services.
  *
- * Web services: Routes are ordered by prefix length (longest first) to ensure
- * more specific routes match before broader ones. For example, `/api/users`
- * must be checked before `/api`, which must be checked before the catch-all `/`.
+ * Routes are ordered by prefix length (longest first) to ensure more specific
+ * routes match before broader ones. For example, `/api/users` must be checked
+ * before `/api`, which must be checked before the catch-all `/`.
  *
- * Serverless services: Route requests to the function path.
- *
- * Cron/Worker services: TODO
- * Use internal routes under `/_svc/crons` and `/_svc/workers`
+ * - Static/SPA services: SPA fallback routes to index.html
+ * - Serverless services: Rewrite to the function entrypoint
+ * - Cron/Worker services: TODO - internal routes under `/_svc/`
  */
 export function generateServicesRoutes(
   services: ResolvedService[]
--- a/packages/fs-detectors/src/services/detect-services.ts
+++ b/packages/fs-detectors/src/services/detect-services.ts
@@ -75,8 +75,9 @@ export async function detectServices(
  * more specific routes match before broader ones. For example, `/api/users`
  * must be checked before `/api`, which must be checked before the catch-all `/`.
  *
- * Static/SPA services: Use filesystem routing with SPA fallback to index.html.
- * The assets are mounted at the routePrefix path.
+ * Static/SPA services: SPA fallback routes to index.html are generated here.
+ * Framework-specific routes (cache headers, etc.) come from the builder via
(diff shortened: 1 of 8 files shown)


packages/fs-detectors/src/services/detect-services.ts
```diff
@@ -91,15 +92,23 @@ export function generateServicesRoutes(
...
-    if (b.routePrefix === '/') return -1;
-    return b.routePrefix.length - a.routePrefix.length;
-  });
-
-  for (const service of sortedWebServices) {
+  const sortServices = <T extends { routePrefix: string }>(arr: T[]): T[] =>
+    [...arr].sort((a, b) => {
+      if (a.routePrefix === '/') return 1;
+      if (b.routePrefix === '/') return -1;
+      return b.routePrefix.length - a.routePrefix.length;
```
1st1: I have a feeling that the comment on top no longer matches the implementation, as length is no longer accounted for paths starting with `/`. Does this matter?

1st1: (And I see that this code has been this way for a while, I just don't get why this sorting is needed at all :))

packages/fs-detectors/src/services/detect-services.ts
```diff
@@ -133,5 +142,34 @@ export function generateServicesRoutes(
     }
   }
 
+  // Generate SPA fallback routes for static services
```
1st1: Why we didn't have to do this prior to this PR? Speaking about the entire new for-loop. Just curious

packages/fs-detectors/src/services/detect-services.ts
ricardo-agz: The '/' check was sort of unnecessary since we only care about longer route prefixes being first and (now) we normalize so routePrefixes always start with /
We care about sorting so that when we generate the route config, the more specific routes get matched first:
```
{
  "version": 3,
  "routes": [
    {
      "handle": "filesystem"
    },
    {
      "src": "^/fastapi-api(?:/.*)?$",
      "dest": "/fastapi-api/main.py",
      "check": true
    },
    {
      "src": "^/admin(?:/.*)?$",
      "dest": "/admin/index.html"
    },
    {
      "src": "/(.*)",
      "dest": "/index.html"
    }
  ],
  "crons": []
}

```

packages/fs-detectors/src/services/detect-services.ts
ricardo-agz: Previously we were building every service as a function, which only worked for backend services. We need to handle static builds a bit differently, mainly with the index.html catchall fallbacks
The double loop was not strictly necessary, consolidated into just one loop with an 
`if (isStaticBuild(service)) {` check

packages/fs-detectors/src/services/detect-services.ts
```diff
@@ -86,50 +87,47 @@ export function generateServicesRoutes(
...
+
+    if (isStaticBuild(service)) {
+      // Static/SPA service: serve index.html for client-side routing
+      if (routePrefix === '/') {
+        defaults.push({ handle: 'filesystem' });
+        defaults.push({ src: '/(.*)', dest: '/index.html' });
+      } else {
+        rewrites.push({
+          src: `^/${normalizedPrefix}(?:/.*)?$`,
+          dest: `/${normalizedPrefix}/index.html`,
```
jeffsee55, line 111: Is this logic pulled from somewhere? How do we know it's always `index.html`?

jeffsee55 (approved):

packages/fs-detectors/src/services/detect-services.ts
ricardo-agz: Yea, you're right, it's not an accurate assumption, just a common case that happens to work for a few framework presets like Vite
I am working on a couple PRs for better support for other frameworks that don't follow this:
https://github.com/vercel/vercel/pull/14849
https://github.com/vercel/vercel/pull/14850

PR #14827, opened 2026-02-02
[python-runtime] Vendor runtime dependencies

elprans:
The Python runtime depends on `werkzeug` and `uvicorn` to provide
the WSGI and ASGI server glue.  Currently, those dependencies are
injected as normal dependencies.  This is problematic because we
cannot pin the dependencies as that might conflict with dependencies in
user projects.

Since we only depend on a small set of packages which themselves are
largely dependency-free, we can get away with vendoring the dependencies
directly into the runtime package.  Use the `vendoring` tool, which is
what `pip` uses to vendor its dependencies.


python/vercel-runtime/src/vercel_runtime/_vendor/markupsafe/_speedups.cpython-314-darwin.so
1st1, line 1: This file shouldn't be committed

packages/python/src/index.ts
```diff
@@ -368,16 +368,16 @@ export const build: BuildV3 = async ({
...
-
-      // Runtime framework dependencies are managed via the uv project so that the
-      // lockfile is the single source of truth for all installed packages. These
-      // are intentionally unpinned so they can resolve alongside user-declared
-      // dependencies (for example, modern Flask versions that require newer
-      // Werkzeug releases).
-      runtimeDependencies.push('werkzeug>=1.0.1');
-      if (framework !== 'flask') {
-        runtimeDependencies.push('uvicorn>=0.24');
+      } else {
```
1st1, line 371: > 'vercel-runtime==0.1.0'

above -- you should prob bump the version?

ricardo-agz (approved):
Nice, LGTM

packages/python/src/index.ts
elprans: done

1st1 (approved):

PR #14828, opened 2026-02-02
[remix] Prevent 404 responses from being cached with immutable headers

jeffsee55:
## Summary

- Move cache-control headers for static assets from before `handle: 'filesystem'` to after `handle: 'hit'`
- This ensures immutable cache headers are only applied when a file actually exists
- Prevents 404 responses from being cached for 1 year with immutable headers

## Background

When cache-control headers were placed BEFORE `handle: 'filesystem'`, the headers would be applied to all matching requests, including 404s. This caused 404 responses for non-existent assets to be cached for 1 year with immutable headers, making them impossible to fix without cache purging.

The correct pattern (used by Next.js) is to place cache headers AFTER `handle: 'hit'`, which only applies them when a file is found:

```
// routes to call after a file has been matched
{ handle: 'hit' },
// Before we handle static files we need to set proper caching headers
{
  src: '/_next/static/...',
  headers: { 'cache-control': 'public,max-age=31536000,immutable' },
  ...
}
```

See: https://github.com/vercel/vercel/blob/main/packages/next/src/server-build.ts#L2748-L2766

## Test plan

- [x] Added integration test probe to verify 404 responses don't get immutable cache headers
- Similar test assertion pattern used here: https://github.com/vercel/vercel/blob/623e43f865e927406200849aa3d7b6a9e9c2dcfc/packages/next/test/fixtures/01-cache-headers/vercel.json#L18-L21

Fixes: https://linear.app/vercel/issue/BE-415

Manual curl of test fixture vs older behavior

```
➜  vercel-2 git:(main) curl -I https://01-no-preset-g45ffbb2a-zero-conf-vtest314.vercel.app/assets/does-not-exist-abc123.js
HTTP/2 404 
age: 0
cache-control: public, max-age=0, must-revalidate
content-type: text/html
date: Mon, 02 Feb 2026 17:38:53 GMT
server: Vercel
strict-transport-security: max-age=63072000; includeSubDomains; preload
x-robots-tag: noindex
x-vercel-cache: MISS
x-vercel-id: sfo1::iad1::b2sp6-1770053933572-302fafb09751

➜  vercel-2 git:(main) curl -I https://01-no-preset-hw782p5gz-zero-conf-vtest314.vercel.app/assets/does-not-exist-abc123.js
HTTP/2 404 
age: 0
cache-control: public, max-age=31536000, immutable
content-type: text/html
date: Mon, 02 Feb 2026 17:39:04 GMT
server: Vercel
strict-transport-security: max-age=63072000; includeSubDomains; preload
x-robots-tag: noindex
x-vercel-cache: MISS
x-vercel-id: sfo1::iad1::cf57s-1770053944387-bb701ce00343
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/rare-rivers-attack.md
+++ b/.changeset/rare-rivers-attack.md
@@ -1,0 +1,5 @@
+---
+"@vercel/remix-builder": patch
+---
+
+[remix] Prevent 404 responses from being cached with immutable headers
--- a/packages/remix/src/build-legacy.ts
+++ b/packages/remix/src/build-legacy.ts
@@ -546,12 +546,8 @@ module.exports = config;`;
     ...staticFiles,
     ...transformedBuildAssets,
   };
+  const assetsPath = remixConfig.publicPath.replace(/^\/|\/$/g, '');
   const routes: any[] = [
-    {
-      src: `^/${remixConfig.publicPath.replace(/^\/|\/$/g, '')}/(.*)$`,
-      headers: { 'cache-control': 'public, max-age=31536000, immutable' },
-      continue: true,
-    },
     {
       handle: 'filesystem',
     },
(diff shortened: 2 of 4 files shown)


tknickman (commented):
Can we add a test as well?

jeffsee55:
> Can we add a test as well?

I added a [probe](https://github.com/vercel/vercel/pull/14828/changes#diff-d943e8e2a265781af9ec69e774330fd084e0742eb8b6d905184612050799ea11) to assert that we don't put that cache-control header on the response. Add a unit test as well you think?

tknickman (approved):

PR #14836, opened 2026-02-03
[next] Remove 404.html prerenders from functions, take 2

mischnic:
Reapply https://github.com/vercel/vercel/pull/14754
Revert https://github.com/vercel/vercel/pull/14754
Closes PACK-6560

Relies on https://github.com/vercel/next.js/pull/89263

---

They were only there as an optimization, but introduce nondeterminism because the HTML contains the build id and deployment id

### DIFF
--- a/.changeset/pink-eels-impress.md
+++ b/.changeset/pink-eels-impress.md
@@ -1,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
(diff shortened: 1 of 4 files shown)


ijjk (approved):
Looks good and test run has only unrelated flakey failure https://github.com/vercel/next.js/actions/runs/21678610490/job/62521427178

PR #14839, opened 2026-02-03
ai: add runtime skill

ecklf:
This skill will help when creating new language runtimes or updating existing ones (e.g., migrating the Go runtime to use runtime: 'executable' with full IPC support).

Adds an agent skill for building new Vercel runtimes that implement the Fluid IPC protocol.
- Documents the IPC message protocol (server-started, handler-started, end, log, metric)
- Provides implementation checklists for both runtime (language-side) and builder (TypeScript-side)
- Links to reference implementations (Node.js, Python, Rust) instead of duplicating code
- Covers key Lambda options for interpreted vs compiled languages (runtime: 'executable')
- Lists common pitfalls to avoid

### DIFF
--- a/.claude/skills/vercel-runtime-implementation-guide.md
+++ b/.claude/skills/vercel-runtime-implementation-guide.md
@@ -1,0 +1,156 @@
+# Vercel Runtime Implementation Guide
+
+Guide for building Vercel runtimes that implement the Fluid IPC protocol.
+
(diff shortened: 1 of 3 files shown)


ofhouse (approved):

sjskkwpkskka:
1f31870e07ded78ff2fe28a1f37ea25cb36a52c1

PR #14843, opened 2026-02-03
[build-utils] Replace Python-based AST parser with @vercel/python-analysiss

elprans:
Replace Python-based AST parser with WASM-based
`@vercel/python-analysis` for basic detection of Python entrypoints. This
eliminates the need for a Python runtime when analyzing Python files for
WSGI/ASGI application patterns.


1st1 (approved):

packages/build-utils/src/python.ts
```diff
@@ -68,12 +26,7 @@ export async function isPythonEntrypoint(
       return false;
     }
```
ricardo-agz: maybe we could get rid of this as well and move it inside `containsAppOrHandler` so it's self contained?

ricardo-agz (approved):

Ashutosh0x:
Nice move @1st1! Moving the AST detection to WASM via @vercel/python-analysis is a big win for build performance and removing runtime dependencies during the analysis phase. This should make the Python build pipeline significantly more resilient across different environments.

PR #14857, opened 2026-02-04
Add experimental manual deployment

javivelasco:
Add experimental support for manual (two-phase) deployments via two new hidden subcommands on `vercel deploy`:

- **`vercel deploy init`** — Creates a deployment in "manual" mode without uploading build outputs. The deployment is initialized on the server and returns a deployment ID.
- **`vercel deploy continue --id dpl_xxx`** — Uploads the `.vercel/output` directory (from a local `vercel build`) to an existing manual deployment and triggers the server-side build/ready flow.

This enables workflows where the user controls the build step locally and then pushes outputs to an existing deployment, useful for CI pipelines and monorepo setups where build orchestration happens outside Vercel.

On the `@vercel/client` side, the `uploadFiles` function was extracted from the monolithic `upload` generator into a standalone reusable export, and a new `continueDeployment` async generator handles the continue flow (hash files, upload missing ones, POST to `/deployments/:id/continue`, poll for ready state). A missing semaphore release on abort was also fixed.

Both subcommands are hidden/experimental for now.

### DIFF
--- a/packages/cli/src/commands/deploy/command.ts
+++ b/packages/cli/src/commands/deploy/command.ts
@@ -2,6 +2,141 @@ import { confirmOption, forceOption, yesOption } from '../../util/arg-common';
 
 export const deprecatedArchiveSplitTgz = 'split-tgz';
 
+export const initSubcommand = {
+  name: 'init',
+  aliases: [],
+  description: 'Create a manual deployment that can be continued later',
+  hidden: true as const,
+  arguments: [],
+  options: [
+    {
+      ...forceOption,
+      description: 'Force a new deployment even if nothing has changed',
+    },
+    {
+      name: 'with-cache',
+      shorthand: null,
+      type: Boolean,
+      deprecated: false,
+      description: 'Retain build cache when using "--force"',
+    },
+    {
+      name: 'public',
+      shorthand: 'p',
+      type: Boolean,
+      deprecated: false,
+      description: 'Deployment is public (`/_src`) is exposed)',
+    },
+    {
+      name: 'env',
+      shorthand: 'e',
+      type: [String],
+      argument: 'KEY=VALUE',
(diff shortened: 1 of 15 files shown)


tknickman (dismissed):
Claude having a great time with oauth. Better review coming

packages/cli/src/commands/deploy/command.ts
```diff
@@ -2,6 +2,141 @@ import { confirmOption, forceOption, yesOption } from '../../util/arg-common';
...
+      type: Boolean,
+      deprecated: false,
+      description: 'Retain build cache when using "--force"',
+    },
+    {
+      name: 'public',
+      shorthand: 'p',
+      type: Boolean,
+      deprecated: false,
+      description: 'Deployment is public (`/_src`) is exposed)',
```
jeffsee55:
```suggestion
      description: 'Deployment is public (exposes `/_src`)',
```

packages/client/src/create-deployment.ts
```diff
@@ -47,6 +48,31 @@ export default function buildCreateDeployment() {
       });
     }
 
+    /**
+     * Manual deployment is an experimental feature that supports only prebuilt
+     * deployments. We could implicitly pass prebuilt=true when hitting the
+     * API but it is more intentional to require the user to set it.
```
jeffsee55, line 54: > We could implicitly pass prebuilt=true when hitting the API but it is more intentional to require the user to set it.

Seems like the `init` command doesn't take the `--prebuilt` flag and it _is_ implied

jeffsee55 (approved):

packages/client/src/create-deployment.ts
javivelasco: Yeah let me update that comment.

PR #14861, opened 2026-02-04
[python] prefer `fastapi dev` if available and reload changes

nsidnev:
Adds optional support for `fastapi dev` for ASGI applications if available. Also includes automatic reloading when code changes, so users don't need to restart `vercel dev` to see updates


packages/python/vc_init_dev_wsgi.py
```diff
@@ -112,7 +112,7 @@ def write(chunk: bytes):
     host = "127.0.0.1"
     try:
         from werkzeug.serving import run_simple
-        run_simple(host, 0, app, use_reloader=False)
+        run_simple(host, 0, app, use_reloader=True)
```
1st1, line 115: Why was it `False`?

nsidnev: It was like this from the beginning, when WSGI support [was added](https://github.com/vercel/vercel/pull/14089/changes/7abd44b3c63064599d1d47eee19bad6955f02cb9#diff-55b14ec81c8f4d7d0cc8ef5abf32d6df0d3029f466fa885654f200211cec6a5a), although the next line somewhat suggests that reloader should be enabled. Perhaps it was a typo that went unnoticed? I tested reloading with `fastapi dev`, `uvicorn`, and `werkzeug` in a single multi-service project, and everything worked fine.

cc @ricardo-agz

1st1: @elprans @ricardo-agz imo should be true? what are we missing?

elprans: I don't see any reasons not to enable it.

ricardo-agz (approved):

1st1 (approved):

PR #14864, opened 2026-02-04
feat(oidc): add optional argument `getVercelOidcToken({ team, project })`

gr2m:
This will allow us to remove some code from vercel/sandbox, see https://github.com/vercel/sandbox/pull/34


gr2m (commented):
tested the changes with latest `ai` as well as the `sandbox` cli with the changes from https://github.com/vercel/sandbox/pull/34

packages/oidc/docs/classes/AccessTokenMissingError.md
```diff
@@ -0,0 +1,160 @@
...
+
+• **message**: `string`
+
+#### Inherited from
+
+Error.message
+
+#### Defined in
+
+node_modules/.pnpm/typescript@4.9.5/node_modules/typescript/lib/lib.es5.d.ts:1054
```
dvoytenko, line 56: A weird code reference

packages/oidc/src/auth-errors.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Error thrown when no authentication configuration is found.
+ * This typically means the user needs to log in.
+ */
+export class AccessTokenMissingError extends Error {
+  name = 'AccessTokenMissingError';
+  constructor() {
+    super('No authentication found. Please log in.');
```
dvoytenko: Provide login URL?

packages/oidc/src/auth-errors.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+/**
+ * Error thrown when attempting to refresh the authentication token fails.
+ * This includes cases where no refresh token is available.
+ */
+export class RefreshAccessTokenFailedError extends Error {
```
dvoytenko, line 18: What's the value of this error? The net result is the same - user has to re-login, right? So at the very least this should be composed on top of `AccessTokenMissingError`?

packages/oidc/src/auth-errors.ts
```diff
@@ -0,0 +1,23 @@
...
+  name = 'RefreshAccessTokenFailedError';
+  cause?: unknown;
```
dvoytenko: There's already `cause` on Error

packages/oidc/src/auth-errors.ts
```diff
@@ -0,0 +1,23 @@
...
+  constructor(cause?: unknown) {
+    super('Failed to refresh authentication token.');
```
dvoytenko:
```
super('...', { cause })
```

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -1,6 +1,28 @@
...
+  /**
+   * Optional team ID to use for token refresh.
+   * When provided, this team ID will be used instead of reading from `.vercel/project.json`.
+   */
+  teamId?: string;
+  /**
+   * Optional project ID to use for token refresh.
+   * When provided, this project ID will be used instead of reading from `.vercel/project.json`.
+   */
+  projectId?: string;
```
dvoytenko: Should we also accept `projectSlug`?

packages/oidc/src/get-vercel-oidc-token.ts
```diff
@@ -1,6 +1,28 @@
...
 
+/**
+ * Options for getting the Vercel OIDC token.
+ */
+export interface GetVercelOidcTokenOptions {
+  /**
+   * Optional team ID to use for token refresh.
+   * When provided, this team ID will be used instead of reading from `.vercel/project.json`.
+   */
+  teamId?: string;
```
dvoytenko: Also `teamSlug`? Totally fine if we just limit this to IDs. This is just FMI.

packages/oidc/src/get-vercel-oidc-token.ts
dvoytenko: As we using `process.env.VERCEL_PROJECT_ID` when available? (https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_PROJECT_ID). Really odd, but I'm not seeing `VERCEL_TEAM_ID` anywhere there. Maybe worth adding for a more universal behavior.

packages/oidc/src/index.ts
```diff
@@ -3,3 +3,8 @@ export {
   getVercelOidcTokenSync,
 } from './get-vercel-oidc-token';
 export { getContext } from './get-context';
+export {
+  AccessTokenMissingError,
+  RefreshAccessTokenFailedError,
+} from './auth-errors';
+export { getVercelCliToken } from './token-util';
```
dvoytenko: Why should this be an exported public function?

packages/oidc/src/token.ts
```diff
@@ -9,17 +9,44 @@ import {
...
+}
+
+export async function refreshToken(
+  options?: RefreshTokenOptions
+): Promise<void> {
+  // Use provided options or fall back to reading from .vercel/project.json
+  let projectId: string;
+  let teamId: string;
+
+  if (options?.projectId && options?.teamId) {
```
dvoytenko: This is where we could try to access `process.env.VERCEL_PROJECT_ID`.

dvoytenko (changes requested):

packages/oidc/docs/classes/AccessTokenMissingError.md
gr2m: this is automated 🤷🏼 `dosc/classes` gets updated when running `npm run build`

packages/oidc/src/auth-errors.ts
gr2m: we want to distinguish between different error paths. Refresh won't occur if access token is present, so I don't think it should derive from that. But they should both derive from the same error. I want to use custom error classes for all error cases but that would be a breaking change. I'll do that as a follow up

packages/oidc/src/get-vercel-oidc-token.ts
gr2m: Good question, I'll check in with @Schniz. The `sandbox` CLI in fact accepts both a slug and an id ([docs](https://vercel.com/docs/vercel-sandbox/cli-reference)).

packages/oidc/src/index.ts
gr2m: right now to centralize implementation. `sandbox` CLI is using it: https://github.com/vercel/sandbox/pull/34/changes#diff-c5500f78f50f5f92794a76b191133a93e494399e55cd92928e16d44a071af25dR7. I agree it's odd to export the function from the oidc package, but I want to address the DX problems first, then do a follow up with refactoring and a new package.

packages/oidc/src/token.ts
gr2m: at this point we want to check if explicit values have been passed, these should override environment settings.

So if anything I would implement it in `findProjectInfo()`

https://github.com/vercel/vercel/blob/d968c1f9070dbe9bdafb0c1c81b56e51678e2092/packages/oidc/src/token-util.ts#L133-L153

but at this point we get the `project.json` anyway which always has the projectId

packages/oidc/src/auth-errors.ts
gr2m: This case should be handled by the consuming application/library. Given the OIDC context We can suggest to login with the vercel CLI, but I'll make sure our own libraries and apps handle `AccessTokenMissingError`. E.g. for AI Gateway, we will suggest to set a gateway access token and link to where the user can do that.

addressed in 919724af0

packages/oidc/src/auth-errors.ts
gr2m: maybe in newer versions of typescript, but here it causes a type error.

image

image

image

packages/oidc/src/auth-errors.ts
gr2m: same as above. I know it's supported by modern JS, but types are outdated

packages/oidc/src/get-vercel-oidc-token.ts
gr2m: looks like teamId and teamSlug can be used interchangeably. But a project name would need to be resolved using an API call to [GET https://api.vercel.com/v9/projects/{idOrName}](https://docs.vercel.com/docs/rest-api/reference/endpoints/projects/find-a-project-by-id-or-name)

I can add it to the OIDC package. Added via 116c8bbad

packages/oidc/src/get-vercel-oidc-token.ts
gr2m: good thinking, but I'll leave it out of scope for this PR if that's ok. Shall we create a follow up issue or do you want us to track it in a different place?

packages/oidc/src/auth-errors.ts
dvoytenko: You can't set it on the `Error`. It can only be provided in the constructor. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#options

If our TS is completely missing it - we should update TS compiler. But we should follow the JS spec. A lot of downstream features depend on the cause being correctly initialized.

packages/oidc/src/auth-errors.ts
dvoytenko: Ditto: let's update compiler, but not lose JS features.

packages/oidc/src/get-vercel-oidc-token.ts
dvoytenko: Yes, let's create followup.

dvoytenko: I also sent https://vercel.slack.com/archives/CLDDX2Y0G/p1770665960750159

packages/oidc/src/index.ts
dvoytenko: Sandbox cannot yet call this function if it doesn't exist yet. Why shouldn't sandbox just call the new cleanly named function? Why create a public function only to remove it next?

packages/oidc/src/auth-errors.ts
dvoytenko: So, what would I do in this code?

```
try {
  getVercelToken()
} catch (e) {
  if (e instanceof RefreshAccessTokenFailedError) {
    // TODO
  }
}
```

I have no moves - only to say "can't do anything", right? I.e. it's not a type, it's a reason.

packages/oidc/src/token-util.ts
```diff
@@ -59,12 +74,59 @@ export async function getVercelCliToken(): Promise<string | null> {
...
+ * If the input starts with 'prj_', it's assumed to be an ID and returned directly.
+ * Otherwise, calls the Vercel API to resolve the slug to an ID.
+ *
+ * @param authToken - The authentication token to use for API calls
+ * @param projectIdOrSlug - Either a project ID (prj_*) or a project slug
+ * @param teamId - Optional team ID to scope the project lookup
+ * @returns The resolved project ID
+ * @throws {VercelOidcTokenError} If the project cannot be resolved
+ */
+export async function resolveProjectId(
```
dvoytenko: How is this used exactly? Can we instead pushing this resolution to the backend when the token is minted?

packages/oidc/src/index.ts
gr2m: how would you call the function? I'm not concerned about how to name it, I'm concerned about creating an entire new package to not get distracted about addressing the DX problems we experience across our tools. I want to centralize the code first to remove duplication, then we can think about how to organize the packages.

dvoytenko (changes requested):

packages/oidc/src/index.ts
gr2m: will rename to `getVercelToken`. We canonically refer to the CLI token as vercel token already, e.g. with `VERCEL_TOKEN` environment variable

packages/oidc/src/token-util.ts
gr2m: will revert this change and investigate if it's needed at all. Originally implemented to remove more code from sandbox, but haven't investigated deeply why they do it in the first place. Will do that first.

.changeset/oidc-optional-team-project-params.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vercel/oidc': minor
+---
+
+- Add optional `teamId` and `projectId` parameters to `getVercelOidcToken()` to allow explicit control over token refresh behavior instead of always reading from `.vercel/project.json`
```
dvoytenko: Nit, but it's "team" and "project", right?

dvoytenko (approved):

packages/oidc/src/auth-errors.ts
gr2m: here the DX differs based on circumstance, the consuming library/app should decide.

- AI SDK is currently guiding the user to create an AI Gateway Access token, but might also enable users to kick of an OAuth Flow for CLI apps that utilize AI SDK
- Sandbox is kicking of an OAuth flow

.changeset/oidc-optional-team-project-params.md
gr2m: yes, thanks, fixed now!

PR #14869, opened 2026-02-05
[python-analysis] Fix ESM compatibility

elprans:
Fixes this:

    (node:67) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of
    file:///.../dist/wasm/vercel_python_analysis.js is not specified
    and it doesn't parse as CommonJS.
    Reparsing as ES module because module syntax was detected. This
    incurs a performance overhead.

Now `@vercel/python-analysis` should be properly `import`-able and
`require()`-able.


nsidnev (approved):

utils/build.mjs
```diff
@@ -1,9 +1,16 @@
 import execa from 'execa';
 import ts from 'typescript';
 import path from 'node:path';
+import { readFileSync } from 'node:fs';
 import { build } from 'esbuild';
 import { fileURLToPath } from 'node:url';
 
+export function getDependencies(cwd = process.cwd()) {
+  const pkgPath = path.join(cwd, 'package.json');
+  const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
```
scotttrinh, line 10: FWIW, it appears that we already have a shared utility function for fetching package.json contents with some built-in caching that we could use here:

https://github.com/vercel/vercel/blob/d8e2592614779687283e52901bccd3ebe79b8233/internals/get-package-json/src/index.ts

packages/python-analysis/scripts/build.mjs
```diff
@@ -0,0 +1,27 @@
+import { esbuild, tsc, getDependencies } from '../../../utils/build.mjs';
```
scotttrinh, line 1: fwiw, since we switched to `module`, you can use a bare `.js` for ESM files.

edit: I didn't mean for this import, I was trying to point to this file itself:

```
packages/python-analysis/scripts/build.mjs -> packages/python-analysis/scripts/build.js
```

utils/build.mjs
```diff
@@ -1,9 +1,16 @@
...
+  return Object.keys(pkg.dependencies || {});
```
scotttrinh, line 11: Tiny tiny micro-optimization, just because it seems like we want to speed up the CLI in general:


```suggestion
  return pkg.dependencies ? Object.keys(pkg.dependencies) : [];
```

scotttrinh (approved):
No blockers, just a few small things to consider.

utils/build.mjs
elprans: I don't think we can rely on that.  This a build script for the CLI itself, not runtime.

utils/build.mjs
elprans: This is a CLI build script, it does not affect runtime performance in any way.

packages/python-analysis/scripts/build.mjs
elprans: I decided to keep mjs for consistency with other packages.

ricardo-agz (approved):

PR #14870, opened 2026-02-05
cli: load local .env files in `vercel env run`

julianbenegas:
## Summary

- Loads local `.env` files (`.env.local`, `.env.development.local`, etc.) when running `vercel env run`
- Uses `@next/env` which was already a dependency but unused
- Fixes env var precedence so local overrides can take effect

## Env var precedence (lowest to highest)

1. Vercel project env vars (from API)
2. Local `.env` files
3. Shell environment variables

Previously, Vercel env vars would override everything, and then the subprocess (e.g. `next dev`) would load `.env.local` but couldn't override already-set vars. Now local `.env` files are loaded by the CLI itself and applied with the right precedence.

## Test plan

- [ ] Run `vercel env run -- env | grep MY_VAR` with `MY_VAR` set in Vercel and `.env.local`
- [ ] Verify `.env.local` value wins over Vercel
- [ ] Verify `MY_VAR=shell vercel env run -- env | grep MY_VAR` shows "shell"

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/gentle-items-knock.md
+++ b/.changeset/gentle-items-knock.md
@@ -1,0 +1,5 @@
+---
+"vercel": patch
+---
+
+cli: load local .env files in `vercel env run`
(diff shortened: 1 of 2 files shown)


plmrry (approved):
Nice, thanks

tknickman (approved):

PR #14876, opened 2026-02-05
[python] Enable standalone runtime unconditionally

elprans:
Drop `VERCEL_RUNTIME_PYTHON_ENABLED` flag and enable standalone runtime
unconditionally.  Change injection method to `uv pip install` to avoid
messing up manifests when building locally.


packages/python/src/index.ts
```diff
@@ -417,9 +404,7 @@ export const build: BuildV3 = async ({
   const entrypointWithSuffix = `${entrypoint}${suffix}`;
   debug('Entrypoint with suffix is', entrypointWithSuffix);
 
-  let handlerPyContents: string;
-  if (useRuntime) {
-    handlerPyContents = `
+  const runtimeTrampoline = `
 import importlib
 import os
```
ricardo-agz, line 409: Not a blocker, just wondering if it would be nicer to move this to its own actual .py file, e.g vc_init.py could just become this

ricardo-agz (approved):

packages/python/src/index.ts
elprans: It's templated so we'd need to read/replace/write anyway.  Also, this stub is temporary, we'll get rid of most of it once we've cleaned up the runtime some.

PR #14881, opened 2026-02-05
[cli] Fix misleading error when unknown flags passed to commands

tknickman:
## Summary
- Fixed misleading "Invalid number of arguments" error when unknown flags are passed to CLI commands
- Root cause: `permissive: true` in `parseArguments()` caused unknown flags to be silently treated as positional arguments
- Added `-y/--yes` flag support to `dns rm` for skipping confirmation prompt
- Added ESLint rule to prevent this pattern in the future

## Changes
- Removed `permissive: true` from leaf commands (dns/rm, dns/add, dns/ls, dns/import, microfrontends/pull, rolling-release subcommands)
- Added `yesOption` to `dns rm` command
- Added telemetry tracking for the `--yes` flag
- Added ESLint rule targeting `packages/cli/src/commands/**/*` (excluding index.ts) that errors on `permissive: true`

## Test plan
- [x] `vercel dns ls domain.com --unknown` now shows "Unknown or unexpected option: --unknown" instead of "Invalid number of arguments"
- [x] `vercel dns rm <id> -y` skips confirmation prompt
- [x] Added unit tests for `--yes` flag and unknown flag rejection

Fixes CICD-1566

### DIFF
--- a/.changeset/clever-crabs-compete.md
+++ b/.changeset/clever-crabs-compete.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
+fix(cli): correct permissive issues
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,7 +759,7 @@ checksum = "9312f7c4f6ff9069b165498234ce8be658059c6728633667c526e27dc2cf1df5"
 
 [[package]]
 name = "vercel_runtime"
-version = "2.1.0"
+version = "2.1.1"
 dependencies = [
  "axum",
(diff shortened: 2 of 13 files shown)


jeffsee55 (approved):

Ashutosh0x:
Nice one @tknickman! That 'Invalid number of arguments' error has definitely caused some head-scratching when it was actually just a typo in a flag. This will make the CLI's feedback much clearer and easier to debug.

PR #14893, opened 2026-02-05
[services] build time service url env vars

ricardo-agz:
* Injects `{SERVICE_NAME}_URL` (and `{FRAMEWORK_PREFIX}_{SERVICE_NAME}_URL`) environment variables for all the web services that are part of the deployment
* These consist of `{VERCEL_URL}/{SERVICE_ROUTE_PREFIX}`
* This is so they are available during the build, as some frontend frameworks, e.g. Next and Vite require them to be available during build time, not just runtime

### DIFF
--- a/packages/build-utils/src/get-service-url-env-vars.ts
+++ b/packages/build-utils/src/get-service-url-env-vars.ts
@@ -1,0 +1,108 @@
+import type { Service } from './types';
+
+type Envs = { [key: string]: string | undefined };
+
+interface FrameworkInfo {
+  slug: string | null;
+  envPrefix?: string;
+}
+
+export interface GetServiceUrlEnvVarsOptions {
+  services: Service[];
+  frameworkList: readonly FrameworkInfo[];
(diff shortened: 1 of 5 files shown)


packages/build-utils/src/get-service-url-env-vars.ts
```diff
@@ -0,0 +1,108 @@
...
+  currentEnv?: Envs;
+  deploymentUrl?: string;
+}
+
+/**
+ * Convert service name to environment variable name.
+ * e.g., "frontend" → "FRONTEND_URL", "api-users" → "API_USERS_URL"
+ */
+function serviceNameToEnvVar(name: string): string {
+  return `${name.replace(/-/g, '_').toUpperCase()}_URL`;
```
elprans, line 22: Do we have a restricted alphabet for service names? is it `[alpha][alnum]*`?

jeffsee55 (approved):

elprans (approved):

PR #14900, opened 2026-02-06
Speedup CLI loading by 30-50% where it matters

1st1:
| Command | `main` branch | this PR | Improvement | Why |
  |---------|--------|---------|-------------|-----------|
  | `vercel --version` | 267ms | 21ms | **12.7x faster** | Fast path; load nothing |
  | `vercel --help` | 224ms | 23ms | **9.7x faster** | Fast path; load almost nothing, just "chalk" and whatever it tags along |
  | `vercel deploy --help` | 263ms | 177ms | 1.5x faster | Separate entry point, code splitting |
  | `vercel login --help` | 229ms | 155ms | 1.5x faster | Separate entry point, code splitting |
  | `vercel whoami --help` | 277ms | 232ms | 1.2x faster | Not directly optimized, but reads less code because deps for priority commands don't load |
  | `vercel dev --help` | 268ms | 230ms | 1.16x faster | Separate entry point, code splitting, but has very heavy deps |

-------

### TL;DR

* The main slowdown happens because the current one-file 9mb JavaScript bundle is chock‑full, no amount of v8 optimizations and laziness can import it all instantly.
* Solution: load and run less of JS code.
* This PR makes `vercel deploy`, `dev`, `login`, `env`, and `init` **30% faster** to load and start running (explained why below). The other commands are becoming a bit faster too.
* `proxy-agent` is a huge dependency; what if we could conditionally use it? If we do we have another **10-15% boost**.
* `vercel --version` and `vercel --help` is now as fast as `node -c "..."` because they don't cause much code to be loaded at all now.

------

### In more detail

1️⃣ **[upd: let's do it in a different PR 🛑]** Looks like we have a number of duplicate packages bundled with slightly different version, roughly 500kb (out of almost 9mb total bundle size).

I make us ask `esbuild` to produce a `metafile` build artifact, which can later be analyzed to see what's bundled and why. The new `pnpm --filter=vercel build:check-dupes` command exposed that we're having some relatively heavy-weight dupes.

I've removed some of the duplicates trimming the total bundle by 250kb or so. This can be continued and also automated in CI -- for example we can make `build:check-dupes` fail if the duplicates total size is greater than X% of the total bundle size. When it is we go fix it.

2️⃣ **[upd: let's do it in a different PR 🛑]** We're using `node-fetch` which tags along some heavy dependencies, close to 400kb. Node has native `fetch()` since version 18, which by itself is no longer supported. Basically there's _almost_ no excuse to use `node-fetch` so I removed it. `agent` bit is missing though, maybe it's ok.

3️⃣ Main optimization: having just one file in our bundle means that ALL CODE must be loaded even for commands that don't need much of it. What if we: (a) enable code splitting in esbuild; (b) separate a few important frequent commands into separate build targets; (c) group less important commands into one other build target; (d) lazy-load everything in the CLI index.ts.

This yields a very measurable and perceivable speedup of 300%; e.g. `vercel deploy --help` went from `300ms` to roughly `200ms`.

It's possible to enable this optimization for all commands (not just a few). Then all commands become faster, but commands like `vc deploy` aren't as fast as with the partial strategy - the code splitting becomes less efficient and they start to load more and thus become slower.

4️⃣ `vercel --version` can be special-cased to run in under 40ms. It doesn't need anything except the version that can be separately produced as a build artifact. Ditto for `vercel --help` that only has one dependency -- `chalk`.

5️⃣ `proxy-agent` is a huge dependency but only few need it. We can detect if the environment is asking for a proxy `HTTP_PROXY` and alike are set and then load/use it, if not - use the standard node http agent.

➡️ Out of all of these 3️⃣ and 5️⃣ are obviously the most impactful optimizations that we should explore further.

-------

### Further thoughts:

6️⃣ We can produce a dedicated build for our build containers & cloud infra where we only package specific commands we need. This potentially can save us a some seconds if we run a lot of vercel commands (do we?)

7️⃣ Minifying bundled code doesn't yield any improvements.

-------

### New bundle layout

| With this PR | `main` branch |
|--------|-------|
| <pre>[ 384]  ./packages/cli/dist/<br>├── [1.9K]  builder-worker.cjs<br>├── [1.1K]  chunks<br>│   ├── [3.7K]  chunk-37LEEXV7.js<br>│   ├── [ 68K]  chunk-3A4GQRZF.js<br>│   ├── [200K]  chunk-3CP4GH2F.js<br>│   ├── [510K]  chunk-43N2PQCI.js<br>│   ├── [ 28K]  chunk-63NZ32AN.js<br>│   ├── [120K]  chunk-6DUXWYRP.js<br>│   ├── [4.3K]  chunk-6LR634YY.js<br>│   ├── [2.9K]  chunk-7YU7CICQ.js<br>│   ├── [1.3K]  chunk-BKB5L3FB.js<br>│   ├── [1.3K]  chunk-D2J3GLXC.js<br>│   ├── [ 76K]  chunk-DMS2IHIP.js<br>│   ├── [ 735]  chunk-EMLBMP2Y.js<br>│   ├── [5.2K]  chunk-F74TGEHO.js<br>│   ├── [5.1K]  chunk-FSNLBJ2P.js<br>│   ├── [ 14K]  chunk-H3QXBVFB.js<br>│   ├── [ 26K]  chunk-HTNMR3YG.js<br>│   ├── [ 45K]  chunk-JB5TGCBH.js<br>│   ├── [ 59K]  chunk-JEKDRBPW.js<br>│   ├── [ 879]  chunk-MXVI7MZS.js<br>│   ├── [338K]  chunk-NTKLBAYU.js<br>│   ├── [6.2K]  chunk-O5X574PA.js<br>│   ├── [ 63K]  chunk-QLL6B4LZ.js<br>│   ├── [ 12K]  chunk-RSRCJV22.js<br>│   ├── [124K]  chunk-VQUSHCT6.js<br>│   ├── [409K]  chunk-VZC6O7OJ.js<br>│   ├── [1.1M]  chunk-WZXTUQDG.js<br>│   ├── [3.4K]  chunk-X3J4XLCN.js<br>│   ├── [9.4K]  chunk-YKN6JHFQ.js<br>│   ├── [ 791]  chunk-YPQSDAEW.js<br>│   ├── [ 864]  compile-vercel-config-LDH47NKP.js<br>│   ├── [1.8M]  dist-7KT3LUVO.js<br>│   └── [ 950]  link-RWCHAJPR.js<br>├── [ 224]  commands<br>│   ├── [  96]  deploy<br>│   │   └── [ 30K]  index.js<br>│   ├── [  96]  dev<br>│   │   └── [820K]  index.js<br>│   ├── [  96]  env<br>│   │   └── [ 41K]  index.js<br>│   ├── [  96]  init<br>│   │   └── [7.5K]  index.js<br>│   └── [  96]  login<br>│       └── [ 656]  index.js<br>├── [618K]  commands-bulk.js<br>├── [7.8K]  get-latest-worker.cjs<br>├── [ 467]  help.js<br>├── [1.7M]  index.js<br>├── [1.0K]  vc.js<br>├── [ 520]  VERCEL_DIR_README.txt<br>└── [  34]  version.mjs</pre> | <pre>[ 224]  ./packages/cli/dist/<br>├── [1.9K]  builder-worker.cjs<br>├── [7.8K]  get-latest-worker.cjs<br>├── [8.9M]  index.js<br>├── [ 425]  vc.js<br>└── [ 520]  VERCEL_DIR_README.txt</pre> |

### Total bundle size

| With this PR | `main` branch |
|--------|-------|
| 8.8Mb | 8.9Mb |


1st1:
~I think I'll rollback dependency dedup & node-fetch work from this PR -- those don't impact perf improvements much anyway, but it will keep this PR in a much more minimal reviewable shape.~

✅

packages/detect-agent/package.json
```diff
@@ -37,7 +37,7 @@
     "url": "https://github.com/vercel/vercel/issues"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=18"
```
LukeSheard: This is a breaking change

1st1: I can roll it back. But why? 18 is no longer supported, 14 is just ancient

elprans: I think the point is to keep breaking changes like this separate from the main bulk so we can revert them more easily.  If you're dropping native fetch, then we can keep 14, right?

1st1: that's the plan: https://github.com/vercel/vercel/pull/14900#issuecomment-3858881340

1st1: Done; this PR is now as focused as it can be :)

packages/cli/src/vc.js
```diff
@@ -9,4 +10,26 @@ try {
   }
 } catch {}
 
+// Fast path for --version to avoid loading the entire CLI
+if (
+  process.argv.length === 3 &&
+  (process.argv[2] === '--version' || process.argv[2] === '-v')
+) {
+  const { version } = await import('./version.mjs');
+  console.log(version);
```
tknickman, line 20: Current version writes:

`{version}` to stdout, and 
`Vercel CLI {version}` to stderr

This is kind of nice because we get both machine parseable and human parseable, but I don't die on this hill. 

If we do land on changing it let's make this a major version just to account for anyone parsing the version in a script.

packages/cli/src/vc.js
```diff
@@ -9,4 +10,26 @@ try {
...
+
+// Fast path for --help to avoid loading the entire CLI
+if (
+  process.argv.length === 3 &&
+  (process.argv[2] === '--help' || process.argv[2] === '-h')
+) {
+  const { version } = await import('./version.mjs');
+  const { help } = await import('./help.js');
+  console.log(`Vercel CLI ${version}`);
+  console.log(help());
```
tknickman: This is all written to stderr currently, but I don't think we need to do that tbh so I'm fine with this, let's just make it a major change.

.changeset/warm-pants-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'vercel': minor
```
tknickman, line 2: If we change output of version I think this should be a major, but otherwise this is good with me

tknickman (commented):
Only a few comments

packages/cli/src/vc.js
1st1: Good catch! I've pushed a commit that fixed the code to match the current behavior, please take a look

packages/cli/src/vc.js
1st1: ditto -- pushed a commit to match the current behavior

.changeset/warm-pants-speak.md
1st1: Let's not change anything, this should be a transparent perf boost, no behavior impact

jeffsee55 (approved):

tknickman (approved):

packages/cli/scripts/build.mjs
```diff
@@ -29,43 +42,94 @@ createConstants();
...
+    mkdirSync(cmdDistDir, { recursive: true });
+  }
+}
+
+await build({
+  entryPoints,
+  bundle: true,
+  format: 'esm',
+  platform: 'node',
+  target: 'node18',
```
elprans: node18?

packages/cli/src/vc.js
```diff
@@ -9,4 +10,27 @@ try {
...
+// Fast path for --help to avoid loading the entire CLI
+if (
+  process.argv.length === 3 &&
+  (process.argv[2] === '--help' || process.argv[2] === '-h')
+) {
+  const { version } = await import('./version.mjs');
+  const { help } = await import('./help.js');
+  console.error(`Vercel CLI ${version}`);
+  console.error(help());
+  process.exit(2);
```
elprans: Why 2?

.changeset/warm-pants-speak.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Improved performance 1.2-1.5x for CLI commands by implementing a different bundling & lazy loading strategy
```
elprans:
```suggestion
1.2-1.5x startup time improvement for CLI commands via chunked bundling & lazy loading strategy
```

elprans (approved):

packages/cli/scripts/build.mjs
1st1: ouch, good catch. I've rollbacked this approach a bunch to make sure we don't diverge from the current `main` branch logic.

packages/cli/src/vc.js
1st1: @claude why indeed

PR #14924, opened 2026-02-06
[python] Verify uv binary is executable in findUvBinary

elprans:
Instead of only checking file existence, run `uv --version` to confirm
the binary actually works before returning it.  This protects from
possibly broken local uv installs.


nsidnev (approved):

brookemosby (approved):
LGTM

PR #14940, opened 2026-02-08
[services] entrypoint detection and devserver fix for Ruby

ricardo-agz:


1st1 (approved):
I'm not a ruby expert but this looks ok

PR #14942, opened 2026-02-08
[cli] Add interactive and flag based configuration of rolling releases

mknichel:
This PR adds interactive and flag based configuration to the `vercel rolling-release configure` command so that users and agents don't need to know the JSON configuration syntax.

- `vercel rolling-release configure` is an interactive flow where the user can select options for each stage
- `vercel rolling-release configure --enable --advancement-type automatic --stage 10,3m --stage 50,5m` would enable rolling releases with automatic advancement with 2 stages at 10% for 3m and 50% for 5m.

### DIFF
--- a/.changeset/quiet-wolves-flow.md
+++ b/.changeset/quiet-wolves-flow.md
@@ -1,0 +1,5 @@
+---
+'vercel': patch
+---
+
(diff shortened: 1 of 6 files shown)


brookemosby (approved):
LGTM

PR #14946, opened 2026-02-09
[services] v2 + v3 builders support

ricardo-agz:
Handles a couple issues to enable support for more complex frontend frameworks for services, especially at route prefixes:
* Next.js at a workspace: if you have your Next.js app at /frontend, we need to invoke the Next builder as if its workPath was at /frontend so that routes don't get prefixed with /frontend
* Treats route-owning (v2), e.g. Next.js vs non-route-owning (v3), e.g. Python builders differently. Route-owning builders don't need synthetic routes injected by the services resolved
* Creates a route-ownership scoping layer for route-owning builders so their emitted src routes only match inside their owning routePrefix namespace (prevents cross-service leakage/redirect loops like root Next.js vs /api/*)
* Normalizes runtime synthetic rewrites to extensionless function destinations (e.g. main.py -> /main) to match v3 runtime output naming


jeffsee55 (approved):

packages/cli/src/util/build/service-route-ownership.ts
```diff
@@ -0,0 +1,107 @@
...
+  allServices: Service[];
+}
+
+function isWebServiceWithPrefix(
+  service: Service
+): service is Service & { type: 'web'; routePrefix: string } {
+  return service.type === 'web' && typeof service.routePrefix === 'string';
+}
+
+function normalizeRoutePrefix(routePrefix: string): string {
```
elprans, line 16: non-blocking observation: we need URL-utils or something to avoid duplicating this

packages/cli/src/commands/build/index.ts
```diff
@@ -695,6 +697,17 @@ async function doBuild(
   }
   const diagnostics: Files = {};
 
+  const hasDetectedServices =
+    detectedServices !== undefined && detectedServices.length > 0;
+  const servicesByBuilderSrc = new Map<string, Service>();
+  if (hasDetectedServices) {
+    for (const service of detectedServices!) {
+      if (service.builder.src) {
+        servicesByBuilderSrc.set(service.builder.src, service);
```
elprans, line 693: should we add a guard here if there are multiple services with same builder.src here?

elprans (approved):
Great stuff!

packages/cli/src/util/build/service-route-ownership.ts
ricardo-agz: Agree, I'll follow up with a slight refactor here

PR #14948, opened 2026-02-09
feat(cli): Add integration discover subcommand with JSON output

bhrigu123:
## Summary

Adds a new `vercel integration discover` subcommand to list installable marketplace products.

### What this adds

- New command:
  - `vercel integration discover`
  - `vercel integration discover --json`
- JSON output format optimized for agent consumption:
  - `integrations[].slug`
  - `integrations[].name`
  - `integrations[].description` (from `shortDescription`)
  - `integrations[].category` (array of category titles)
  - `integrations[].products[]` (`slug`, `name`)
- Human-readable table output by default.
- Telemetry:
  - `subcommand:discover`
  - `flag:json` when used

### API usage

- `GET /v2/integrations/integrations?integrationType=marketplace`
- `GET /v2/integrations/categories`

### Notes
- The output is product focused (not integration). We list each product that are available on marketplace across all integrations.
- We fetch available integrations and their tags separately and combine the results. `integrations` endpoint returns `tag_ids[]` for each integration, and we fetch `tag_id` to to `title` from the `categories` endpoint.
- If tag lookup fails, the command warns and still returns integrations with empty tag arrays.
- The command filters to integrations **that are both marketplace and installable (`isMarketplace && canInstall`)**, then maps `tagIds` to category titles.

### Example output
```sh
$ pnpm vercel integration discover

> Available marketplace integrations:
Product Name                Slug                        Provider                  Description                       Tags                                                        
Browserbase                 browserbase                 Browserbase               Automate browsers with AI         Web Automation, AI, Agents, Workflow                        
AssistLoop                  assistloop                  AssistLoop                AI Agent for Customer Service     Support Agent, Agents, AI
Amazon Aurora DSQL          aws/aws-dsql                AWS                       Serverless PostgreSQL             Storage, Postgres                                           
Amazon OpenSearch           aws/open-search             AWS                       open search                       Storage, Searching
```

```sh
$ pnpm vercel integration discover --json

{
  "products": [
    {
      "name": "Browserbase",
      "slug": "browserbase",
      "provider": "Browserbase",
      "description": "Automate browsers with AI",
      "tags": [
        "Web Automation",
        "AI",
        "Agents",
        "Workflow"
      ]
    },
    {
      "name": "Amazon OpenSearch",
      "slug": "aws/open-search",
      "provider": "AWS",
      "description": "open search",
      "tags": [
        "Storage",
        "Searching"
      ]
    },
    {
      "name": "Amazon Aurora PostgreSQL",
      "slug": "aws/aws-apg",
      "provider": "AWS",
      "description": "Full PostgreSQL compatibility",
      "tags": [
        "Storage",
        "Postgres"
      ]
    }
}
```
## Test Plan

- Added unit tests in `packages/cli/test/unit/commands/integration/discover.test.ts`:
  - help telemetry
  - JSON output shape
  - table output
  - category-fetch failure behavior
  - integrations-fetch error handling
  - debug argv shape regression
  - telemetry for subcommand/json flag
- Added mock handlers in `packages/cli/test/mocks/integration.ts` for discover endpoints.


packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  const categoryTitleById = new Map<string, string>(
+    categories.map(category => [category.id, category.title])
+  );
+
+  const results = integrations
+    .filter(integration => integration.isMarketplace && integration.canInstall)
+    .map(integration => {
+      const category = (integration.tagIds ?? [])
```
dvoytenko: "categories"?

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+        .map(tagId => categoryTitleById.get(tagId))
+        .filter((title): title is string => Boolean(title));
+
+      return {
+        slug: integration.slug,
+        name: integration.name,
+        description: integration.shortDescription ?? '',
+        category,
```
dvoytenko: What do we do with the rest of the tags? IMHO they are just as important and we should output them all here too.

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+        products: (integration.products ?? []).map(product => ({
+          slug: product.slug,
+          name: product.name,
```
dvoytenko, line 144: Ditto: products have tags too, right?

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+  return table(
+    [
+      ['Name', 'Slug', 'Description', 'Products'].map(header =>
+        chalk.bold(chalk.cyan(header))
+      ),
+      ...integrations.map(integration => [
+        integration.name,
+        integration.slug,
+        integration.description || chalk.gray('-'),
+        integration.products.length > 0
```
dvoytenko: Is this enough focus on products? Especially for multi-product cases this may need to be more, including descriptions, tags, etc. Overall it seems maybe we should invert this concept and do product-first.

dvoytenko (changes requested):

packages/cli/src/commands/integration/discover.ts
bhrigu123: Good point. The idea of `discover` was to mirror the https://vercel.com/marketplace experience, which primarily focuses on Integrations.

We also discussed about a separate command to fetch integration details, which will be more detailed including product details, snippets, etc. Hence trying to keep this one for discoverability only without too much details.

But open to discussing broadly. Will create a slack thread.

packages/cli/src/commands/integration/discover.ts
bhrigu123: we are showing all the tags as list (the name should be categories here).

(I didn't add a column for categories on the table / non-json output though because the table on the terminal was getting overflown. We do have the description so category felt redundant. We do have it in the json output)

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+import { printError } from '../../util/error';
+import { validateJsonOutput } from '../../util/output-format';
+import table from '../../util/output/table';
+import output from '../../output-manager';
+import { discoverSubcommand } from './command';
+import { IntegrationDiscoverTelemetryClient } from '../../util/telemetry/commands/integration/discover';
+import { fetchMarketplaceIntegrations } from '../../util/integration/fetch-marketplace-integrations-list';
+import { fetchIntegrationCategories } from '../../util/integration/fetch-integration-categories';
+
+type IntegrationListItem = {
```
tonypan2: this type is defined in `fetch-marketplace-integrations-list` and `fetch-integration-categories.ts` as well

packages/cli/src/util/integration/fetch-marketplace-integrations-list.ts
```diff
@@ -0,0 +1,20 @@
...
+  slug: string;
+  name: string;
+  shortDescription?: string;
+  tagIds?: string[];
+  products?: { slug: string; name: string }[];
+  isMarketplace?: boolean;
+  canInstall?: boolean;
+};
+
+export async function fetchMarketplaceIntegrations(client: Client) {
```
tonypan2: rename to `fetchMarketplaceIntegrationsList`? there is already a `fetchMarketplaceIntegrations`

packages/cli/src/commands/integration/command.ts
```diff
@@ -123,6 +123,32 @@ export const listSubcommand = {
...
 } as const;
 
+export const discoverSubcommand = {
+  name: 'discover',
+  aliases: [],
+  description: 'Discover available marketplace integrations',
+  arguments: [],
+  options: [
+    {
+      name: 'json',
```
tonypan2: seems like we should use this instead https://github.com/vercel/vercel/blob/45963f54f26a3e54df58a3b6a30e383de4094dd1/packages/cli/src/util/arg-common.ts#L128-L135

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,189 @@
...
+  canInstall?: boolean;
+};
+
+type IntegrationCategory = {
+  id: string;
+  title: string;
+};
+
+export async function discover(client: Client, args: string[]) {
+  let parsedArguments = null;
```
tonypan2, line 58: should do some scope checking like this?

https://github.com/vercel/vercel/blob/1e93f403e1532828b58f9886aaaf8be1d7e2dc3f/packages/cli/src/commands/integration/list.ts#L62-L67

tonypan2: i guess this command should be able to run in "logged out" mode. so maybe team check is not necessary. but please check that it actually works when you are not logged in.

packages/cli/src/util/integration/fetch-marketplace-integrations-list.ts
```diff
@@ -0,0 +1,20 @@
...
+  return client.fetch<IntegrationListItem[]>(
+    '/v2/integrations/integrations?integrationType=marketplace',
```
tonypan2, line 20: we should create a ticket to make this endpoint support pagination

tonypan2 (changes requested):
some comments

packages/cli/src/commands/integration/discover.ts
bhrigu123: Made it product focused

packages/cli/src/util/integration/fetch-marketplace-integrations-list.ts
bhrigu123: https://linear.app/vercel/issue/MKT-2725/support-pagination-in-integrations-api

packages/cli/src/commands/integration/discover.ts
bhrigu123: merging tags of products and integrations now

packages/cli/src/commands/integration/discover.ts
bhrigu123: the auth enforcement happens at the global CLI entry point level. We do have a hardcoded `subcommandsWithoutToken` list but `integration` is not under that list under which `discover` will exist. I think it should be okay to be behind auth since we can expect a user to be working with a project to be able to discover and add integrations.

tonypan2 (approved):
could be done as a follow up: we should surface billing plans for each product too. this will help ai agents choose.

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  const allTags = [...(integrationTagIds ?? []), ...(productTags ?? [])];
+
+  for (const tag of allTags) {
+    if (tag.startsWith('tag_')) {
+      const title = categoryTitleById.get(tag);
+      if (title && !seen.has(title)) {
+        seen.add(title);
+        result.push(title);
+      }
```
dvoytenko: maybe also `else { result.push(tag.substring(4))}`?

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,226 @@
...
+  'messaging',
+  'compute',
+]);
+
+function resolveTags(
+  productTags: string[] | undefined,
+  integrationTagIds: string[] | undefined,
+  categoryTitleById: Map<string, string>
+): string[] {
+  const seen = new Set<string>();
```
dvoytenko: why both `seen` and `result`? Just make `result` a Set :)

packages/cli/src/commands/integration/discover.ts
```diff
@@ -0,0 +1,226 @@
...
+        provider: integration.name,
+        description: integration.shortDescription ?? '',
+        tags: resolveTags(undefined, integration.tagIds, categoryTitleById),
+      });
+    } else {
+      const isMultiProduct = products.length > 1;
+      for (const product of products) {
+        results.push({
+          name: product.name,
+          slug: isMultiProduct
```
dvoytenko: should be `isMultiProduct || product.slug !== integration.slug`

dvoytenko (approved):

packages/cli/src/commands/integration/discover.ts
bhrigu123: Good point. That way we won't skip any tags. Updated

packages/cli/src/commands/integration/discover.ts
bhrigu123: ah I was trying to maintain the order of the tags. But don't think that's important. Simply using a set now.

packages/cli/src/commands/integration/discover.ts
bhrigu123: I see. Good catch. Updated.

PR #14970, opened 2026-02-10
v2beta queue format

ctgowrie:
### Summary
- Adds support for `queue/v2beta` trigger type where the consumer name is implicitly derived from the function path
- For v2beta, only one queue trigger per function is allowed
- Consumer name is sanitized from the function path (e.g., `api/test.js` → `api-testjs`)

### Changes
- **Schema**: Accept both `queue/v1beta` and `queue/v2beta` types; v2beta doesn't allow `consumer` field in config
- **Types**: Separate input types (`TriggerEventInput`) from output types (`TriggerEvent`); output always has `consumer`
- **`getLambdaOptionsFromFunction`**: Populates `consumer` for v2beta triggers using `sanitizeConsumerName(functionPath)`
- **`sanitizeConsumerName`**: New utility that strips dots and replaces invalid characters with hyphens

### v1beta vs v2beta

| | v1beta | v2beta |
|---|---|---|
| `consumer` in config | Required | Not allowed |
| Consumer name | User-specified | Derived from function path |
| Triggers per function | Unlimited | 1 max |

> [!NOTE]
> Low Risk Change
>
> This PR adds a new queue trigger format (v2beta) with schema validation, type definitions, and a utility function to derive consumer names from function paths - all additive changes with comprehensive test coverage.
> 
> - Adds `queue/v2beta` trigger type alongside existing `queue/v1beta` with schema validation
> - New `sanitizeConsumerName` utility encodes function paths into valid consumer names
> - Extensive unit tests covering v2beta validation and sanitization edge cases
>
> <sup>Risk assessment for [commit 739c6be](https://github.com/vercel/vercel/commit/739c6be5c2b5db614e36e8033109e640db9c3c2e).</sup>


.changeset/forty-seahorses-shave.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vercel/build-utils': patch
+'@vercel/config': patch
+'vercel': patch
+---
+
+Add new expirementalTrigger format for queeus v2beta
```
Timer: @cursoragent fix this typo in spelling

packages/build-utils/src/lambda.ts
```diff
@@ -11,9 +11,26 @@ import type {
...
+ * Allowed characters: A-Z, a-z, 0-9, _, -
+ * - Dots are stripped (removed)
+ * - Other disallowed characters are replaced with hyphens
+ *
+ * @example
+ * sanitizeConsumerName('api/test.js') // => 'api-testjs'
+ * sanitizeConsumerName('api/users/handler.ts') // => 'api-users-handlerts'
+ */
+export function sanitizeConsumerName(functionPath: string): string {
+  return functionPath.replace(/\./g, '').replace(/[^A-Za-z0-9_-]/g, '-');
```
Timer: This probably isn't fool proof in terms of preventing collisions, is it? Can we change to an alt encoding approach? Maybe a url encoding style? What would be best?

@cursoragent want to propose a fix here?

.changeset/forty-seahorses-shave.md
Timer: @cursoragent try again i fixed auth

packages/build-utils/src/lambda.ts
Timer: @cursoragent try again i fixed auth

ctgowrie: URL encoding is tricky because of how proxy interprets encoded paths (worry about bugs), and we have the consumer name in paths. I prefer we keep this simple and risk collisions. I think they will be exceedingly rare / near zero.

Alternatively, if you want shorter/cleaner names for common characters, we could use mnemonic escapes:
/ → _S (slash)
. → _D (dot)
_ → __ (double underscore)
Everything else → _XX

Wyt?

Timer: sgtm

ctgowrie: Let's stick with the simple `-`. Anything else will not be human readable in the dashboard

ctgowrie: Used mnemonic

Timer (approved):

PR #14976, opened 2026-02-10
[python] Support python functions >250mb in size

gscho:
## Summary

When a Python function is over the 250mb limit for lambda, we package the `uv` binary into the lambda and install the dependencies at runtime.

If a package is determined to be "private" it will always be packaged into the lambda zip because we can't install it at runtime.

## Enabling the feature

To enable this feature the user must set `VERCEL_EXPERIMENTAL_PYTHON_UV_INSTALL_ON_STARTUP=1` in their vercel environment. ALSO in order to increase the ephemeral storage to 10GB you need to set `MAXIMISE_TMP_STORAGE=1`. This functionality already existed but maybe we can make the builder set that for us somehow.

## Blast Radius

The blast radius of this PR is small because customer's need to opt-in via environment variable.

> [!NOTE]
> Low Risk Change
>
> This PR adds new build-time logic for Python Lambda packaging with runtime dependency installation, gated behind an explicit opt-in environment variable (VERCEL_EXPERIMENTAL_PYTHON_UV_INSTALL_ON_STARTUP), limiting blast radius to users who enable the feature.
> 
> - New feature: runtime dependency installation for oversized Python Lambdas (>249MB)
> - Feature gated behind explicit env var opt-in (VERCEL_EXPERIMENTAL_PYTHON_UV_INSTALL_ON_STARTUP)
> - Adds uv.lock parsing, package classification (public vs private), and bundle size calculation
>
> <sup>Risk assessment for [commit a3ffb56](https://github.com/vercel/vercel/commit/a3ffb56228cf79e2ad1996341d759052900662a6).</sup>


packages/python/trampoline.py.tmpl
```diff
@@ -0,0 +1,104 @@
...
+os.environ.update(
+    {
+        "__VC_HANDLER_MODULE_NAME": "{{{moduleName}}}",
+        "__VC_HANDLER_ENTRYPOINT": "{{{entrypointWithSuffix}}}",
+        "__VC_HANDLER_ENTRYPOINT_ABS": os.path.join(_here, "{{{entrypointWithSuffix}}}"),
+        "__VC_HANDLER_VENDOR_DIR": "{{{vendorDir}}}",
+    }
+)
+
+{{#runtimeInstallEnabled}}
```
elprans: We should move all of this into vercel-runtime.  Primarily because we want the tunnel and logging properly set up before we do any work, otherwise we lose visibility.

packages/python/src/index.ts
```diff
@@ -472,12 +498,129 @@ from vercel_runtime.vc_init import vc_handler
...
+
+  if (runtimeInstallEnabled && uvLockPath && uvProjectDir) {
+    console.log(
+      `Bundle size (${totalBundleSizeMB} MB) exceeds limit. ` +
+        `Enabling runtime dependency installation.`
+    );
+
+    // Regenerate lock file with --no-build --upgrade to ensure all packages have binary wheels.
+    // --no-build: Only select package versions that have pre-built wheels
+    // --upgrade: Re-resolve all packages, ignoring pins in the existing lock file
```
msullivan: Why `--upgrade`?
We probably do want to respect pins if the lock file was user provided, right?

(It's possible I'm totally off base here, since I'm probably missing a bunch of context. (Indeed, I'm just reviewing this to try to build some more context))

gscho: Great question and I also debated this myself. The goal is to do this runtime install stuff only for packages which have prebuilt dependencies because lambda has no build environment. If we don’t pass this flag we won’t be certain we can install their dependencies at runtime and it could result in a nasty runtime error that is hard for the user to detect. I decided to implicitly pass this flags so more often than not the customers code would just work on vercel. Also if we can’t resolve a lock file that would work with this feature it will fail at build time giving them a nice error.

msullivan: *nod*

I'd be pretty frustrated as a user though if my app broke because the version of something changed even though I was using a lock file.

gscho: Re-run the `uv sync` with the `only-binary` flag and fail their build preferable I guess? I think fixing the problem for them automatically is a bit nicer but perhaps too magic? WDYT @elprans

elprans: You want `uv sync --frozen --no-build --reinstall --no-cache`

elprans: If we want to optimize, we can run `uv sync --frozen --no-build` _first_, note a failure, then run `uv sync --frozen` as usual, and if we overflow and if prior `--no-build` failed we error out.

gscho: It now produces this at build time:

```
Error: Bundle size exceeds the Lambda limit and requires runtime dependency installation, but some packages in your uv.lock file do not have pre-built binary wheels available. Runtime dependency installation requires all public packages to have binary wheels.

To fix this, either:
1. Remove the problematic packages, or
2. Regenerate your lock file with: uv lock --upgrade --no-build
```

packages/python/src/index.ts
```diff
@@ -480,12 +543,138 @@ from vercel_runtime.vc_init import vc_handler
...
+          `Runtime dependency installation requires all public packages to have binary wheels.\n\n` +
+          `To fix this, either:\n` +
+          `1. Remove the problematic packages, or\n` +
+          `2. Regenerate your lock file with: uv lock --upgrade --no-build`,
+      });
+    }
+
+    // Read and parse the uv.lock file
+    const lockContent = await fs.promises.readFile(uvLockPath, 'utf8');
+    const lockFile = parseUvLock(lockContent);
```
elprans: Let's push this down into python-analysis with proper error handling. See #15019

packages/python/src/index.ts
```diff
@@ -587,7 +588,23 @@ from vercel_runtime.vc_init import vc_handler
 
     // Read and parse the uv.lock file
     const lockContent = await fs.promises.readFile(uvLockPath, 'utf8');
```
elprans: This needs to be wrapped in a try .. catch

elprans: Also use `await readFile` for consistency

packages/python/src/install.ts
```diff
@@ -536,3 +576,126 @@ export async function mirrorSitePackagesIntoVendor({
...
+  let totalSize = 0;
+
+  for (const filePath of Object.keys(files)) {
+    const file = files[filePath];
+    if ('fsPath' in file && file.fsPath) {
+      try {
+        const stats = await fs.promises.stat(file.fsPath);
+        totalSize += stats.size;
+      } catch (err) {
+        debug(`Failed to stat file ${file.fsPath}: ${err}`);
```
elprans: i would raise this to warning
```suggestion
        console.warn(`Warning: Failed to stat file ${file.fsPath}, size will not be included in bundle calculation: ${err}`);
```

packages/python/src/install.ts
```diff
@@ -536,3 +576,126 @@ export async function mirrorSitePackagesIntoVendor({
...
+}): Promise<Files> {
+  const vendorFiles: Files = {};
+
+  if (privatePackages.length === 0) {
+    debug('No private packages to bundle');
+    return vendorFiles;
+  }
+
+  // Normalize package names for comparison (PEP 503: lowercase, replace - and _ with -)
+  const normalizePackageName = (name: string): string =>
```
elprans: Use normalizePackageName from python-analyusis

packages/python/src/install.ts
```diff
@@ -536,3 +576,126 @@ export async function mirrorSitePackagesIntoVendor({
...
+      for (const entry of entries) {
+        // Skip __pycache__ and .pyc files
+        if (entry.name === '__pycache__' || entry.name.endsWith('.pyc')) {
+          continue;
+        }
+
+        // Check if this entry belongs to a private package
+        // Package directories are named like: package_name or package_name-version.dist-info
+        const entryBaseName = entry.name
+          .replace(/-[\d.]+\.dist-info$/, '')
```
elprans, line 649: Dist info would include underscores I think, e.g `package_name-version.dist-info` per [PEP 427](https://peps.python.org/pep-0427/#escaping-and-unicode)

elprans (approved):

ricardo-agz (approved):

PR #15004, opened 2026-02-11
CLI: non-interactive ensureLink

brookemosby:
When non-interactive and linking requires confirmation or project selection, ensureLink/setupAndLink now return HEADLESS and output JSON action_required (confirmation_required), with coverage in ensureLink tests.

> [!NOTE]
> Low Risk Change
>
> CLI changes add non-interactive mode handling that throws errors or returns early when confirmation is needed, which is defensive behavior that prevents unintended actions without explicit confirmation.
> 
> - Adds HEADLESS error throw in inputProject when client.nonInteractive is true
> - Modifies setup-and-link to handle nonInteractive flag and return error status when prompts needed
> - Adds comprehensive unit tests for ensureLink non-interactive behavior
>
> <sup>Risk assessment for [commit 0b9e8aa](https://github.com/vercel/vercel/commit/0b9e8aaf1a59fdaa15b1ec1d894136ae85af2ed7).</sup>


packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -70,6 +75,8 @@ export default async function setupAndLink(
     successEmoji = 'link',
     setupMsg = 'Set up',
     projectName = basename(path),
+    projectId,
```
jeffsee55: Right now you can pass `--project`, which gets used as the name if it's creating a new one (right?). Can that value be used instead of this and if it's an existing project name or ID we can find it?

jeffsee55 (approved):

packages/cli/src/util/link/ensure-link.ts
```diff
@@ -48,11 +53,32 @@ export async function ensureLink(
 
   if (link.status === 'error') {
     if (link.reason === 'HEADLESS') {
-      output.error(
-        `Command ${getCommandName(
-          commandName
-        )} requires confirmation. Use option ${param('--yes')} to confirm.`
-      );
+      if (nonInteractive) {
+        outputActionRequired(
```
codywong, line 57: this already calls exit so don't need to call again @ line80

packages/cli/src/util/link/ensure-link.ts
```diff
@@ -29,6 +31,9 @@ export async function ensureLink(
   opts: SetupAndLinkOptions = {}
 ): Promise<ProjectLinked | number> {
   let { link } = opts;
+  // All commands respect global --non-interactive; link can override via opts
+  const nonInteractive = opts.nonInteractive ?? client.nonInteractive ?? false;
```
codywong, line 35: outputActionRequired short-circuits on !client.nonInteractive

Is it possible to have opts.nonInteractive === true but client.nonInteractive === false  
  
If so, currently we no-op?

packages/cli/src/util/link/setup-and-link.ts
```diff
@@ -97,12 +100,13 @@ export default async function setupAndLink(
...
+  if (!isTTY && !autoConfirm && !nonInteractive) {
     return { status: 'error', exitCode: 1, reason: 'HEADLESS' };
   }
 
   const shouldStartSetup =
     autoConfirm ||
+    nonInteractive ||
     (await client.input.confirm(
       `${setupMsg} ${chalk.cyan(`“${toHumanPath(path)}”`)}?`,
       true
```
codywong, line 112: <p class='graphite__hidden'><i>[Re: line +124]</i></p>
does this need some nonInteractive handling too?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/vercel/vercel/15004?utm_source=unchanged-line-comment">Graphite</a>.</p>

PR #15005, opened 2026-02-11
fix(codeowners): remove old members

tknickman:
Remove old members


tknickman:
Admin merging

.github/CODEOWNERS
ijjk, line 8: Can we fix that these codeowners can't also approve for a `pnpm-lock.yaml` change?

Maybe a specific line just below for the `pnpm-lock.yaml` file?

tknickman: That should be changed as of yesterday

tknickman: https://github.com/vercel/vercel/blob/main/.github/CODEOWNERS#L44-L49

PR #15010, opened 2026-02-11
[routing-utils] Undeprecate `routes`

MatthewStanciu:
Some time ago, we deprecated the `routes` property and replaced it with new properties of its individual components: `rewrites`, `redirects`, `headers`. This made sense at the time. But then last year we added [header transforms](https://vercel.com/changelog/transform-rules-are-now-available-in-vercel-json) and made them only available on the deprecated `routes` in vercel.json.

THEN we built on top of this feature when we built `vercel.ts` by adding syntactic sugar that makes it easy to add transforms. This quickly caused some frustrating DX issues because we were abstracting away the vercel.json property where `transforms` lives but still compiling to `vercel.json`, which caused some confusing errors.

I normalized these in https://github.com/vercel/vercel/pull/14705, but that's just a band-aid. The real issue is that we're in this weird state where an important CDN feature can only be configured via a property in vercel.json/ts that your editor yells at you for using because it's "deprecated" and the only reason its replacements are not backwards-compatible is because have different names for `source`, `destination`, and `statusCode`.

@mknichel wrote a proposal for undeprecating `routes` here: [**Undeprecate routes in vercel.json**](https://www.notion.so/vercel/Undeprecate-routes-in-vercel-json-2f8e06b059c480e58013d975a8d606df)

This PR implements part 1 of undeprecating routes:
- removes `deprecated` in the schema
- converts `source`, `destination` and `statusCode` to `src`, `dest`, and `status` 
- updates error messages & unit tests

Next steps will be implemented in follow-ups:
- Deprecate `handle`, `continue`, `override`, `important` properties within routes
- Remove `mixed-routing-properties` error; allow `routes` and `headers` / `redirects` / `rewrites` together
- Update docs

## Full stack

- https://github.com/vercel/vercel/pull/15010 (this PR)
- https://github.com/vercel/vercel/pull/15015
- https://github.com/vercel/vercel/pull/15020

Smaller/loose ends:
- https://github.com/vercel/vercel/pull/15014
- https://github.com/vercel/vercel/pull/15016

> [!NOTE]
> Low Risk Change
>
> This PR removes the deprecated flag from the routes schema and adds property aliases (source/destination/statusCode) with corresponding validation and tests, which is a non-breaking schema change with no security implications.
> 
> - Schema: removes `deprecated: true` from routesSchema, adds `source`/`destination`/`statusCode` aliases
> - Validation: adds `convertRouteAliases` function to normalize aliases to canonical forms with conflict detection
> - Tests: extensive unit test updates for new alias support and error messages
>
> <sup>Risk assessment for [commit 816ca41](https://github.com/vercel/vercel/commit/816ca4161489ba987b27eda45f3925b671cb070b).</sup>


packages/routing-utils/src/schemas.ts
```diff
@@ -350,25 +350,32 @@ const transformsSchema = {
  */
 export const routesSchema = {
   type: 'array',
-  deprecated: true,
```
MatthewStanciu, line 353: this one line is the biggest deal of the pr!

packages/routing-utils/test/index.spec.ts
```diff
@@ -269,11 +352,11 @@ describe('normalizeRoutes', () => {
       ],
       [
         {
-          keyword: 'required',
```
mknichel, line 272: Should we expect to see an error that either `src` or `source` is required for this route?

packages/routing-utils/src/schemas.ts
mknichel: Could we mark `important` and `override` and `handle` as still deprecated?

MatthewStanciu: this is currently done in https://github.com/vercel/vercel/pull/15015. do we want to move that into this pr?

packages/routing-utils/test/index.spec.ts
MatthewStanciu: yes, good catch. i've done a pass through all the errors in routing-utils related to all the properties that were converted (src, dest, status) and updated the user-facing errors to include both.

packages/routing-utils/src/schemas.ts
mknichel: You could have empty changesets for the separate changes and then release them all at the end.

packages/routing-utils/src/types.ts
```diff
@@ -85,6 +85,12 @@ export type RouteWithSrc = {
     redirect?: Record<string, string>;
     cookie?: string;
   };
+  /**
+   * Aliases for `src`, `dest`, `statusCode`, converted during normalization
```
mknichel: Could you expand on the reason for having both: `source` is used to be consistent with `rewrites` and `redirects` and is more user friendly, but underneath the covers, everything will get converted to `src`.

packages/routing-utils/test/index.spec.ts
mknichel: I don't see the error added back in the latest diff. Did I miss it?

MatthewStanciu: They're in the diff in `index.ts`, added in https://github.com/vercel/vercel/pull/15010/changes/6b117f5f821bc3ed735144f268e9765732784e8f and https://github.com/vercel/vercel/pull/15010/changes/d7d37c67e8783468898a548242fe4135dba54173. Was that what you meant?

packages/routing-utils/src/types.ts
MatthewStanciu: done in https://github.com/vercel/vercel/pull/15010/changes/12525db4725c10fe052e0dd9458208ae39de0b9a!

packages/routing-utils/test/index.spec.ts
MatthewStanciu: I see what you mean now. I added a new test here: https://github.com/vercel/vercel/pull/15010/changes/29151c1ee4c3ea37ea66ef2892e2a503531b13fa

mknichel (approved):

PR #15015, opened 2026-02-11
[routing-utils] Deprecate `handle`, `important`, and `override` properties in routes schema

MatthewStanciu:
See https://github.com/vercel/vercel/pull/15010 for more context

> [!NOTE]
> Low Risk Change
>
> This PR deprecates schema properties and updates error messages to remove `handle` as a valid route option, which is a breaking API change (marked as major version bump) but only adds deprecation warnings rather than removing functionality.
> 
> - Error message updated to remove `handle` as valid route property
> - Added `deprecated: true` flags to `handle`, `important`, and `override` in schema
> - Added JSDoc `@deprecated` comments to TypeScript types
>
> <sup>Risk assessment for [commit f014d3d](https://github.com/vercel/vercel/commit/f014d3d180573de1754433fffeb17c218bd9aaac).</sup>


packages/routing-utils/src/schemas.ts
```diff
@@ -400,18 +400,22 @@ export const routesSchema = {
...
           },
           important: {
+            deprecated: true,
             type: 'boolean',
           },
           user: {
             type: 'boolean',
           },
           continue: {
+            deprecated: true,
```
mknichel: Let's leave this not deprecated.

packages/routing-utils/src/schemas.ts
```diff
@@ -400,18 +400,22 @@ export const routesSchema = {
...
             type: 'boolean',
           },
           override: {
+            deprecated: true,
             type: 'boolean',
           },
           check: {
+            deprecated: true,
```
mknichel: Let's leave this as not deprecated too.

mknichel (approved):

PR #15016, opened 2026-02-11
[config] Undeprecate `routes`

MatthewStanciu:
> [!NOTE]
> Low Risk Change
>
> This PR adds alias support for route configuration properties (source/src, destination/dest, statusCode/status) with validation that throws errors when both aliases are used, plus removes a deprecation marker - no security-sensitive changes.
> 
> - Adds route property aliases with validation to prevent conflicting definitions
> - Removes @deprecated annotation from `routes` config option
> - New tests covering alias validation and normalization behavior
>
> <sup>Risk assessment for [commit 9e541ad](https://github.com/vercel/vercel/commit/9e541ad8acdc60ec27bc9055e7b62347b8948909).</sup>


packages/config/src/router.ts
```diff
@@ -1094,11 +1100,27 @@ export class Router {
    *    });
    */
   public route(config: Route): this {
-    this.validateSourcePattern(config.src);
+    const src = config.src ?? config.source;
```
mknichel, line 1183: Does this code also validate that a user doesn't try to specify both fields?

mknichel (approved):

packages/config/src/router.ts
MatthewStanciu: Oops no, fixed here https://github.com/vercel/vercel/pull/15016/changes/9e541ad8acdc60ec27bc9055e7b62347b8948909

PR #15018, opened 2026-02-11
[services] frontend + backend routing bugfixes

ricardo-agz:
Fixes a few issues relating to Next.js/Vite + FastAPI project routing
* Fixes issue where if you have a Vite app mounted at the root + a Next.js app mounted at /dashboard and you ping a route like: /dashboard/nonexistent-route, it falls back to Vite's index.html instead of Next.js's 404
* Moves service functions to an internal _svc namespace to prevent an issue where hitting /backend would be handled by FastAPI instead of Next.js 404
* Also adds a few e2e fixture tests for services
* Moves some helper functions like normalizeRoutePrefix() to routing-utils

> [!WARNING]
> High Risk Change
>
> This PR fixes routing bugs by scoping service routes with ownership guards and moving function outputs to a reserved internal namespace, which are infrastructure-level routing changes that could affect request handling across services.
> 
> - Removes `isRouteOwningBuilder` condition, applying route scoping to all builders
> - Adds reserved `/_svc` prefix validation and internal function path namespace
> - Modifies route source generation with ownership guards to prevent cross-service route capture
>
> <sup>Risk assessment for [commit b25567b](https://github.com/vercel/vercel/commit/b25567b489ace5ea60ef995c7ee6dcc7dbd476a9).</sup>


packages/cli/src/commands/build/index.ts
```diff
@@ -988,7 +987,6 @@ async function doBuild(
       if (
         hasDetectedServices &&
         service &&
```
ricardo-agz, line 989: This check is no longer necessary since non-route-owning builders also emit routes

nsidnev (approved):

brookemosby (approved):
LGTM

PR #15026, opened 2026-02-12
fs-detectors flaky test fix

ricardo-agz:
Fixes a CI flake in `test/unit.local-file-system-detector.test.ts` (`should skip entry if socket`).
[Example](https://github.com/vercel/vercel/actions/runs/21936432882/job/63351617725?pr=14986)

Cause:
- The test called `fs.mkdir()` without `await`, then immediately called `server.listen()` on a Unix socket path.
- When bind failed (seen as intermittent `EACCES`), the promise never rejected, so the test timed out at 5s.
- It also used a shared temp dir (`/tmp/socket-dir`), which can collide across concurrent CI runs.

What changed:
- Switched to `fs.mkdtemp()` for a unique temp directory per test run.
- Wrapped `server.listen()` with both resolve + reject (`server.once('error', reject)`).
- Read with `readdir('')` from detector root.
- Made cleanup more robust: close server first (if listening), then remove temp dir.

> [!NOTE]
> Low Risk Change
>
> This PR contains only test file improvements to fix a flaky CI test by using unique temp directories and proper async error handling.
> 
> - Test fix: switched to `fs.mkdtemp()` for unique temp directories per test run
> - Added proper error handling with `server.once('error', reject)` for socket binding
> - Improved cleanup ordering: close server before removing temp directory
>
> <sup>Risk assessment for [commit edbc0fe](https://github.com/vercel/vercel/commit/edbc0febea14d918cedf7f96f170b811cc17c495).</sup>


packages/fs-detectors/test/unit.local-file-system-detector.test.ts
```diff
@@ -77,22 +77,31 @@ describe('LocalFileSystemDetector', () => {
       return;
     }
 
-    const socketdir = path.join(os.tmpdir(), 'socket-dir');
+    const socketdir = await fs.mkdtemp(path.join(os.tmpdir(), 'socket-dir-'));
```
1st1, line 82: btw do we care about cleaning up directories after tests?

1st1 (approved):

PR #15027, opened 2026-02-12
[vercel flags] store hashed keys

dferber90:
Patches the experimental embedding of flag definitions.

The changed functionality is enabled only enabled when the `VERCEL_EXPERIMENTAL_EMBED_FLAG_DEFINITIONS`  environment variable  is set. The existing `@vercel/flags-core` library is already forwards-compatible with the new format.

See https://vercel.com/docs/flags/vercel-flags/sdks/core#embedded-definitions

> [!NOTE]
> Low Risk Change
>
> This PR adds SHA-256 hashing of SDK keys before embedding them in generated flag definition modules, which is a security improvement that prevents raw keys from being exposed in build output.
> 
> - Adds SHA-256 hashing of SDK keys to avoid embedding raw keys in output
> - New `vercel flags prepare` CLI subcommand for manual flag definition preparation
> - Comprehensive test coverage for hashing and module generation
>
> <sup>Risk assessment for [commit 46358e9](https://github.com/vercel/vercel/commit/46358e977c8ca4dfede7380af1091dafc9a0509f).</sup>


packages/cli/src/commands/flags/index.ts
```diff
@@ -136,6 +139,14 @@ export default async function main(client: Client) {
     case 'sdk-keys':
       telemetry.trackCliSubcommandSdkKeys(subcommandOriginal);
       return sdkKeys(client);
+    case 'emit-datafiles':
```
mehulkar: `--output` or `--write` seems like a better convntion here instead of `emit-datafiles`, what do you think?

mehulkar (approved):

packages/cli/src/commands/flags/index.ts
dferber90: I'm thinking just `vercel flags prepare` makes the most sense. It's a pretty specific operation that not a lot of people will ever need to run manually

dferber90: Renamed here 61045b8d7911dea5e2257959ca0c65e10497f2e7

tknickman (approved):

PR #15032, opened 2026-02-12
Update release workflows so release PR syncing is manual

jeffsee55:
- Stops automatic release PR churn by gating release.yml so changesets/action only runs on merged Version Packages commits.
- Adds a new manual workflow, release-version-pr-sync.yml, to explicitly sync/refresh the Version Packages PR when needed.
- Keeps status checks predictable with a summary job that passes on skipped releases and fails only on real release failures/cancellations.

> [!NOTE]
> Low Risk Change
>
> This PR modifies GitHub Actions release workflows to gate automatic releases behind a commit message check and adds a manual workflow dispatch option, which are CI/CD configuration changes but don't affect security controls or application logic.
> 
> - New manual workflow_dispatch trigger for release PR syncing
> - Gates release job with commit message condition check
> - Summary job updated to handle skipped release states
>
> <sup>Risk assessment for [commit 0531157](https://github.com/vercel/vercel/commit/0531157e5184dda49157ab483131dce0a519fa11).</sup>

### DIFF
--- a/.github/workflows/release-version-pr-sync.yml
+++ b/.github/workflows/release-version-pr-sync.yml
@@ -2,11 +2,6 @@ name: Release Version PR Sync
 
 on:
   workflow_dispatch:
-    inputs:
-      reason:
-        description: 'Why sync the release PR now?'
-        required: false
-        type: string
 
 env:
   TURBO_REMOTE_ONLY: 'true'
@@ -64,12 +59,3 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
 
-  summary:
-    name: Summary (release version PR sync)
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    needs:
-      - sync-release-pr
-    steps:
-      - name: Check All
-        run: echo OK
--- a/.github/workflows/release-version-pr-sync.yml
+++ b/.github/workflows/release-version-pr-sync.yml
@@ -4,8 +4,8 @@ on:
   workflow_dispatch:
 
 env:
-  TURBO_REMOTE_ONLY: "true"
-  TURBO_TEAM: "vercel"
+  TURBO_REMOTE_ONLY: 'true'
+  TURBO_TEAM: 'vercel'
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release-version-pr-sync.yml
```diff
@@ -0,0 +1,75 @@
+name: Release Version PR Sync
+
+on:
+  workflow_dispatch:
+    inputs:
+      reason:
```
tknickman: We don't need this right? This should just create if it doesn't exist, update otherwise.

.github/workflows/release-version-pr-sync.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Sync Release Pull Request
+        uses: changesets/action@v1
+        with:
+          version: pnpm ci:version
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+
+  summary:
```
tknickman: I don't think we need this do we?

.github/workflows/release.yml
```diff
@@ -16,6 +16,7 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    if: startsWith(github.event.head_commit.message, 'Version Packages')
```
tknickman, line 19: I don't think we need this. It can use the existing changesets logic that switches behavior based on the presence of .changeset files, right?

jeffsee55: This is what keeps normal PRs from triggering the Version Packages flow, so the other action creates the version packages PR and then this only lets that one through

.github/workflows/release-version-pr-sync.yml
jeffsee55: Removing

.github/workflows/release-version-pr-sync.yml
jeffsee55: Removed!

.github/workflows/release.yml
mehulkar: I like this approach of limiting which pushes to main trigger release. Note that this commit message is configurable in changeset config and also in repo merge settings, so it can break in the future. Might want to make the changeset config explicit and add a comment here

.github/workflows/release.yml
```diff
@@ -107,8 +108,16 @@ jobs:
...
     needs:
       - release
     steps:
       - name: Check All
-        run: echo OK
+        run: |
+          # Fail summary only when the gated `release` job actually fails/cancels.
+          if [ "${{ needs.release.result }}" = "failure" ] || [ "${{ needs.release.result }}" = "cancelled" ]; then
+            echo "Release job failed with result: ${{ needs.release.result }}"
+            exit 1
```
mehulkar, line 121: We should also add a datadog monitor for when this workflow fails

mehulkar (approved):

PR #15033, opened 2026-02-12
[ci:release] Integrate Python runtime release properly

elprans:
Currently the vercel-runtime Python package gets pushed to PyPI _after_
all e2e and npm publishing was done via workflow dispatch to
`release-python-runtime.yml`.  This is problematic because the Python
builder pins the runtime version _exactly_ and would fail to find the
yet-unreleased package version on PyPI when running e2e tests.

A simple fix would be to publish to PyPI before e2e, but that defeats
the purpose of pre-release testing.  Instead, here we follow the
JavaScript package procedure where we push artifacts to Vercel and test
on those.  For Python this means that we inject `VERCEL_RUNTIME_PYTHON`
pointing to `vercel-runtime @ ${{ dplUrl }}/tarballs/vercel-runtime.whl`
alongside `VERCEL_CLI_VERSION`.

The remaining issue is to make sure that PyPI publication succeeds
_before_ we proceed with the main CLI publication, otherwise we will
release a broken build.  Achieve this by moving the PyPI publication step
into `ci:publish` so it runs before `pnpm publish`.

Specific changes:

  - Add `python/vercel-runtime/publish.mjs` (version check, build, smoke
    test, publish)
  - Add `utils/publish-runtimes.mjs` which gets called from `ci:publish`
    for dispatch (mostly just generalizing non-JS publication here).
  - Remove async `workflow_dispatch` trigger and `push` trigger from
    `release-python-runtime.yml` (`workflow_dispatch` kept for now for
    manual publication if we need it).
  - `api/_lib/script/build.ts` learns to scan and copy
    `python/**/dist/*.whl` so they are accessible in pre-release e2e.

> [!WARNING]
> High Risk Change
>
> This PR modifies CI/CD release workflows and publishing scripts to integrate Python runtime publication before npm publish, which is infrastructure/deployment configuration affecting the release process.
> 
> - Modifies `ci:publish` script to run Python runtime publish before npm publish
> - Changes release workflow to add `environment: release` and Python setup steps
> - Adds new publish scripts for PyPI publication with version validation
>
> <sup>Risk assessment for [commit 478d73d](https://github.com/vercel/vercel/commit/478d73de694b165f257fcd9b167fd104820951c8).</sup>

### DIFF
--- a/.github/workflows/release-python-runtime.yml
+++ b/.github/workflows/release-python-runtime.yml
@@ -1,10 +1,6 @@
 name: Release Python Runtime Package
 
 on:
-  push:
-    branches: [main]
-    paths:
-      - 'python/vercel-runtime/**'
   workflow_dispatch:
     inputs:
       force:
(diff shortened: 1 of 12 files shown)


.github/workflows/release-python-runtime.yml
```diff
@@ -1,10 +1,6 @@
 name: Release Python Runtime Package
```
gscho, line 1: Should this workflow use publish.mjs now?

elprans: good call. no need to duplicate.

gscho (approved):

jeffsee55 (approved):

PR #15040, opened 2026-02-13
Mark SvelteKit as superseding vite

msullivan:
Currently, when trying to use sveltekit with
`VERCEL_USE_EXPERIMENTAL_SERVICES=1` in zero-config mode, we get errors like:
```
Error: Multiple frameworks detected in frontend/: SvelteKit, Vite. Use explicit experimentalServices config.
```
from the services auto-detection.

SvelteKit depends on Vite, so it's expected for both to be present. In this case the culprit is that SvelteKit isn't marked as superseding Vite.

Tested manually but I didn't write a test because I couldn't find any similar sort of testing.

There might be other lurking issues of this form, since it seems like many call sites just take the first framework found.

> [!NOTE]
> Low Risk Change
>
> This PR adds a framework metadata property to mark SvelteKit as superseding Vite for auto-detection purposes, with corresponding tests.
> 
> - Adds `supersedes: ['vite']` to two SvelteKit framework definitions
> - Adds test cases for SvelteKit + backend auto-detection scenarios
> - Changeset for minor version bump
>
> <sup>Risk assessment for [commit 70edb0d](https://github.com/vercel/vercel/commit/70edb0dd8795feb3eb9b7753d8c251e310e3ec2b).</sup>


nsidnev (approved):
LGTM! Can we also add a unit test for auto-detection in `packages/fs-detectors/test/unit.auto-detect-services.test.ts` that would check for the superseding case?

ricardo-agz (approved):

PR #15041, opened 2026-02-13
fix(cli): fix import resolution strategy for builders

mehulkar:
Previously, if a builder was not found in the local cache directory, we
were always installing it. This was incorrect. We should first check the
CLI's node modules for the builder. This will support environments like
the Vercel build container that pre-install all the builders for
performance.

This also fixes the e2e test for eleventy-v0. This fixture was
incidentally failing to deploy with this bug, because a local
.vercel/builders directory was created with node_modules. The test setup
uses vc deploy, which does not push up the .gitignore file, so eleventy
was then trying to read all of .vercel/builders and building it. This
failed because there HAPPENED to be a file in there that Eleventy could
not process. Through a series of fortunate circumstances, we found this
bug.

> [!NOTE]
> Low Risk Change
>
> This PR changes builder import resolution order to check CLI's node_modules before installing, which is a logic change in build tooling but does not affect security, auth, or data integrity.
> 
> - Reorders builder resolution: checks CLI node_modules before triggering install
> - Removes early-exit install path when peerDep not found in cache
> - Adds unit test and .eleventyignore entry for test fixture
>
> <sup>Risk assessment for [commit cd7917c](https://github.com/vercel/vercel/commit/cd7917c94fa3773ad1c41dcb0698463860ce92bf).</sup>


tonypan2 (approved):

packages/elysia/.deploy
```diff
@@ -0,0 +1 @@
+Thu Feb 12 23:04:47 CST 2026
```
ijjk, line 1: Are these intended to commit?

ijjk (approved):

packages/elysia/.deploy
mehulkar: yeah.. this repo uses turbo but it's caching is very busted. The reason this bug was missed is because some e2e tests never ran. I'm adding these deploy files (similar to api repo) to manually invalidate all builders to run their unit and e2e tests. My friend claude helped me with a script:

```bash
#!/usr/bin/env bash
# Writes current date to .deploy in each package that is a peerDependency of
# packages/cli. Uses `turbo ls --output json` to map package name → directory.
set -euo pipefail
REPO_ROOT="$PWD"
CLI_PKG="$REPO_ROOT/packages/cli/package.json"
DATE_STR="$(date)"

turbo_json="$(pnpm exec turbo ls --output json 2>/dev/null)"
peer_names=$(jq -r '
  .peerDependencies // {} | to_entries[]
  | select(.value | type == "string" and (. == "workspace:*" or startswith("workspace:")))
  | .key
' "$CLI_PKG")

for name in $peer_names; do
  dir="$(echo "$turbo_json" | jq -r --arg n "$name" '.packages.items[] | select(.name == $n) | .path')"
  if [[ -z "$dir" || "$dir" == "null" ]]; then
    echo "No dir for $name (turbo ls)" >&2
    continue
  fi
  echo "$DATE_STR" > "$REPO_ROOT/$dir/.deploy"
  echo "$name → $dir/.deploy"
done
```

mehulkar: They might get published to the dist, so probably not the best strategy, but i'll come back to that.

PR #15050, opened 2026-02-13
feat(cli): easier agentic login

tknickman:
Support better login experience from cursor / claude

> [!NOTE]
> Low Risk Change
>
> This PR modifies CLI login UX to auto-open browser for authentication instead of waiting for user input, which is a flow change but does not alter authentication security - the same device authorization flow is used.
> 
> - Removes manual [ENTER] prompt, auto-opens browser for device auth flow
> - Adds Cursor agent detection to control browser-opening behavior in CI
> - Graceful fallback with instructions when browser fails to open
>
> <sup>Risk assessment for [commit 265367b](https://github.com/vercel/vercel/commit/265367b4bc1c8ac0a658080a147da04e192d13fb).</sup>

### DIFF
--- a/packages/cli/src/commands/login/future.ts
+++ b/packages/cli/src/commands/login/future.ts
@@ -2,6 +2,7 @@ import readline from 'node:readline';
 import chalk from 'chalk';
 import * as open from 'open';
 import { eraseLines } from 'ansi-escapes';
+import { KNOWN_AGENTS } from '@vercel/detect-agent';
 import type Client from '../../util/client';
 import { printError } from '../../util/error';
 import { updateCurrentTeamAfterLogin } from '../../util/login/update-current-team-after-login';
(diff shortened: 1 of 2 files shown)


AAorris (approved):

packages/cli/src/commands/login/future.ts
```diff
@@ -47,7 +48,42 @@ export async function login(
...
+  const shouldSkipBrowser = process.env.CI && !isCursorAgent;
+
+  o.log(
+    `\n  Visit ${chalk.bold(
+      o.link(
+        verification_uri.replace('https://', ''),
+        verification_uri_complete,
+        { color: false, fallback: () => verification_uri_complete }
+      )
+    )}${o.supportsHyperlink ? ` and enter ${chalk.bold(user_code)}` : ''}\n`
```
AAorris, line 64: MUST display user_code
> Clients MUST still display the "user_code", as the authorization server will require the user to confirm it to disambiguate devices or as remote phishing mitigation
https://datatracker.ietf.org/doc/html/rfc8628#section-3.3.1

tknickman: This is unchanged from the current flow, but agree - will fast follow a fix here

PR #15054, opened 2026-02-13
[cli] Fix integration remove, balance, and open commands not finding integrations

bhrigu123:
## Summary

The `vercel integration remove`, `vercel integration balance`, and `vercel integration open` fail today with:
- `Error: No integration <name> found.`
- `Error: No installations found for this integration`
- `Error: No configuration found for "<name>".`

### Root Cause

These three commands use `fetchMarketplaceIntegrations()` to look up integration configurations via `GET /v2/integrations/configurations`. This function relied on `client.fetch` to automatically inject the `teamId` query parameter from `client.config.currentTeam`.

However, `config.currentTeam` is only set when users explicitly run `vercel teams switch` (which persists it to disk) or pass the `--scope` flag (in-memory for that invocation). For most users who haven't done either of these, `currentTeam` is undefined. The team is still resolved correctly by `getScope()` (via `user.defaultTeamId`), but `client.fetch` never adds `teamId` to the API request. The API then returns results scoped to the user's personal account context (which has no marketplace integrations) instead of the team — resulting in an empty response every time.

In contrast, `vercel integration list` worked correctly because `getResources()` explicitly passes `teamId` as a query parameter rather than relying on the automatic injection.

**This bug affects most users** — it only works for users who have explicitly run `vercel teams switch` or always use `--scope`.

### Fix

- Added an explicit `teamId` parameter to `fetchMarketplaceIntegrations()` and `getFirstConfiguration()`
- Updated all three callers (`remove`, `balance`, `open`) to pass `team.id` from the already-resolved team object
- This matches the pattern used by `getResources()` in the working `integration list` command

### Verified

- API returns correct data when `teamId` is explicitly included (confirmed via direct API calls)
- Built locally and tested `integration remove` and `integration balance` against real integrations — both now work
- All 29 existing unit tests pass

## Test plan

- [x] `vercel integration remove <slug>` finds and removes the integration
- [x] `vercel integration balance <slug>` finds and shows balance info
- [x] Existing unit tests pass (remove: 8, balance: 10, open: 11)
- [x] Test `vercel integration open <slug>`

> [!NOTE]
> Low Risk Change
>
> Bug fix that explicitly passes teamId parameter to API calls, fixing integration commands that failed when currentTeam was not set; this is a defensive change that ensures proper team scoping rather than loosening access.
> 
> - Adds explicit `teamId` query parameter to `fetchMarketplaceIntegrations()` API call
> - Sets `useCurrentTeam: false` to prevent automatic team injection
> - Adds test verifying teamId is required and passed correctly
>
> <sup>Risk assessment for [commit 8556e0c](https://github.com/vercel/vercel/commit/8556e0c9503f5ae31d3b37674a59baf2cbcea743).</sup>


packages/cli/src/util/integration/fetch-marketplace-integrations.ts
```diff
@@ -3,12 +3,14 @@ import type { Configuration } from './types';
...
   client: Client,
-  slug: string
+  slug: string,
+  teamId: string
 ) {
   const searchParams = new URLSearchParams();
   searchParams.set('view', 'account');
   searchParams.set('installationType', 'marketplace');
   searchParams.set('integrationIdOrSlug', slug);
+  searchParams.set('teamId', teamId);
```
tonypan2, line 13: Nit: The [`useConfiguration()`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/test/mocks/integration.ts#L1104) mock doesn't validate that `teamId` is present in the request (it only reads `integrationIdOrSlug`), and all tests set `client.config.currentTeam = team.id` in `beforeEach` — so the auto-injection in [`client._fetch`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/src/util/client.ts#L338) makes tests pass regardless of this fix. Would it be worth adding a test where `currentTeam` is NOT set (relying on `defaultTeamId`) and having the mock assert `teamId` presence, similar to how [`useResources()`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/test/mocks/integration.ts#L1069) does?

packages/cli/src/util/integration/fetch-marketplace-integrations.ts
```diff
@@ -3,12 +3,14 @@ import type { Configuration } from './types';
...
   return await client.fetch<Configuration[]>(
```
tonypan2, line 14: Nit: Now that `teamId` is explicitly provided via search params, you could pass `useCurrentTeam: false` in the [`client.fetch`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/src/util/integration/fetch-marketplace-integrations.ts#L14) options to prevent the redundant overwrite from [`client._fetch` L338](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/src/util/client.ts#L338) when `currentTeam` IS set. Not blocking — [`getResources()`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/src/util/integration-resource/get-resources.ts#L4) also doesn't set it, so this is consistent with existing patterns.

packages/cli/src/util/integration/fetch-marketplace-integrations.ts
tonypan2, line 13: Tangential: The sibling function [`fetchMarketplaceIntegrationsList()`](https://github.com/vercel/vercel/blob/75677bf056fd9f92f33adf1c63e682dee18dcc59/packages/cli/src/util/integration/fetch-marketplace-integrations-list.ts#L18) (used by the `discover` command) calls `/v2/integrations/integrations` without explicitly passing `teamId`. Different endpoint, but worth checking if it has the same class of bug when `currentTeam` is unset.

tonypan2 (approved):
LGTM

packages/cli/src/util/integration/fetch-marketplace-integrations.ts
bhrigu123: Added a test in the `"without currentTeam (defaultTeamId fallback)"` describe block that sets up a northstar user with `defaultTeamId` but does NOT set `client.config.currentTeam`. The mock validates that `teamId` is present in the request — if missing, it returns 400. This confirms the fix works for the defaultTeamId fallback path.

packages/cli/src/util/integration/fetch-marketplace-integrations.ts
bhrigu123: Good call — added `useCurrentTeam: false` to prevent the redundant override.

packages/cli/src/util/integration/fetch-marketplace-integrations.ts
bhrigu123: Checked — `fetchMarketplaceIntegrationsList()` is used by the `discover` command and hits a different endpoint (`/v2/integrations/integrations`) which returns a global catalog of available integrations, not team-specific configurations. It does not need a `teamId`. So this is not affected.

PR #15056, opened 2026-02-14
[cli] Add --format=json to integration balance, remove & resource commands

bhrigu123:
## Summary
- Add `--format=json` support to `integration balance`, `integration remove`, `integration-resource remove`, and `integration-resource disconnect` commands
- Refactor `integration balance` to use `parseArguments` (standard flag-parsing pattern) instead of raw `args`
- For action commands (`remove`, `disconnect`), require `--yes` when `--format=json` is passed to avoid interactive prompts in machine-readable mode

[MKT-2529: JSON formatting in responses](https://linear.app/vercel/issue/MKT-2529/json-formatting-in-responses)

## Test plan
- [x] All 251 existing integration/integration-resource tests pass
- [x] New tests for each command: valid JSON output, `--yes` requirement, invalid format error, telemetry tracking
- [ ] Manual test: `node packages/cli/dist/index.js integration balance <name> --format=json`
- [x] Manual test: `node packages/cli/dist/index.js integration remove <name> --format=json --yes`
- [x] Manual test: `node packages/cli/dist/index.js integration-resource remove <name> --format=json --yes`
- [x] Manual test: `node packages/cli/dist/index.js integration-resource disconnect <name> <project> --format=json --yes`

## Examples

```sh 
vc-dev integration-resource disconnect prisma-postgres-yellow --format=json --yes
Vercel CLI 50.17.1
{
  "resource": "prisma-postgres-yellow",
  "disconnected": true,
  "project": "blog-nextjs-sanity-9"
}
```

```sh
vc-dev integration-resource remove prisma-postgres-yellow --format=json --yes
Vercel CLI 50.17.1
{
  "resource": "prisma-postgres-yellow",
  "removed": true
}
```

```sh
vc-dev integration remove prisma --format=json --yes
Vercel CLI 50.17.1
{
  "integration": "prisma",
  "removed": true
}
```

**Note**: could not manually run the balance command as all the test resources have no balance available. Have added unit tests

## Error examples 
Output to std::err (not printed as json)

```sh
vc-dev integration-resource remove prisma-postgres-yellow --format=json
Vercel CLI 50.17.1
Error: --format=json requires --yes to skip confirmation prompts
```

```sh
vc-dev integration remove prisma --format=json
Vercel CLI 50.17.1
Error: --format=json requires --yes to skip confirmation prompts
```

> [!NOTE]
> Low Risk Change
>
> This PR adds --format=json output support to CLI integration commands with proper validation requiring --yes flag for destructive operations, representing a feature addition with no security or data risk.
> 
> - Adds --format=json option to integration balance, remove, and disconnect commands
> - Requires --yes flag when using --format=json to prevent interactive prompts in machine-readable mode
> - Extensive test coverage added for new JSON output functionality
>
> <sup>Risk assessment for [commit e38a251](https://github.com/vercel/vercel/commit/e38a2511aefd8f010408a9d9d57a9a9da7598c6c).</sup>


packages/cli/src/commands/integration-resource/disconnect.ts
```diff
@@ -111,6 +126,16 @@ export async function disconnect(client: Client) {
       }
       throw error;
     }
+
+    if (asJson) {
```
tonypan2, line 131: The `--all` path calls [`handleDisconnectAllProjects`](https://github.com/vercel/vercel/blob/b1b76de947e9d94123b43977b776aeb761ed1e22/packages/cli/src/commands/integration-resource/disconnect.ts#L213) which unconditionally calls `output.success()` to stderr before this JSON block runs. Every other code path in this PR correctly suppresses the success message in JSON mode (e.g., [`handleDisconnectProject`](https://github.com/vercel/vercel/blob/b1b76de947e9d94123b43977b776aeb761ed1e22/packages/cli/src/commands/integration-resource/disconnect.ts#L204) gates it behind `!asJson`). Consider passing `asJson` to `handleDisconnectAllProjects` to suppress it, matching the pattern used elsewhere. Same applies to `remove-resource --disconnect-all --format=json`.

packages/cli/src/commands/integration-resource/disconnect.ts
```diff
@@ -111,6 +126,16 @@ export async function disconnect(client: Client) {
...
+      output.stopSpinner();
+      const projects =
+        targetedResource.projectsMetadata?.map(project => project.name) ?? [];
+      client.stdout.write(
+        `${JSON.stringify({ resource: resourceName, disconnected: true, projects }, null, 2)}\n`
```
tonypan2: Nit: The `--all` path outputs `{ projects: [...] }` (array) while the single-project path at [line 200](https://github.com/vercel/vercel/blob/b1b76de947e9d94123b43977b776aeb761ed1e22/packages/cli/src/commands/integration-resource/disconnect.ts#L200) outputs `{ project: "name" }` (string). Different field names and types for the same command means consumers need to handle two response shapes. Worth considering always using `projects: string[]` for consistency?

packages/cli/src/commands/integration-resource/disconnect.ts
```diff
@@ -111,6 +126,16 @@ export async function disconnect(client: Client) {
       }
       throw error;
     }
+
+    if (asJson) {
+      output.stopSpinner();
```
tonypan2: Nit: By the time this runs, the spinner is already stopped by `handleDisconnectAllProjects` (which calls `output.success` internally). The call is harmless (idempotent) but misleading since every other JSON path in this PR calls `stopSpinner` when the spinner is actually still running.

packages/cli/src/commands/integration-resource/disconnect.ts
tonypan2: Nit: The `--all` path uses `resourceName` from `parsedArguments.args[1]` (user input), while the single-project path at [line 200](https://github.com/vercel/vercel/blob/b1b76de947e9d94123b43977b776aeb761ed1e22/packages/cli/src/commands/integration-resource/disconnect.ts#L200) uses `resource.name` from the API response. These could differ in casing or normalization. Using the same source for both would be more predictable.

packages/cli/src/commands/integration/command.ts
```diff
@@ -238,6 +242,7 @@ export const removeSubcommand = {
       description:
         'Skip the confirmation prompt when uninstalling an integration',
     },
+    formatOption,
```
tonypan2, line 304: Nit (pre-existing): There's no help snapshot test for `integration remove`. The other subcommands (`balance`, `disconnect`, `resource remove`) all have snapshots in `help.test.ts.snap`, but `integration remove` doesn't, so this new `--format` option isn't snapshot-tested for this subcommand.

packages/cli/src/commands/integration-resource/disconnect.ts
bhrigu123: Fixed — `handleDisconnectAllProjects` now accepts an `asJson` parameter. When true, it calls `output.stopSpinner()` instead of `output.success()`, so no stderr output leaks in JSON mode. Both callers (disconnect.ts and remove-resource.ts) now pass `asJson` through.

packages/cli/src/commands/integration-resource/disconnect.ts
bhrigu123: Good call — unified to always use `projects: string[]` for both single-project and `--all` paths. Single-project now returns `{ projects: [projectName] }` instead of `{ project: projectName }`.

packages/cli/src/commands/integration-resource/disconnect.ts
bhrigu123: Removed — the `output.stopSpinner()` now lives inside `handleDisconnectAllProjects` (in the `asJson` branch), so the caller no longer has the misleading redundant call.

packages/cli/src/commands/integration-resource/disconnect.ts
bhrigu123: Fixed — the `--all` path now uses `targetedResource.name` (from the API response) instead of `resourceName` (from user input), matching the single-project path.

packages/cli/src/commands/integration/command.ts
bhrigu123: Noted — this is a pre-existing gap (not introduced by this PR). Leaving it out of scope for now.

bhrigu123: Created [MKT-2822: Add missing help snapshot test for `integration remove` subcommand](https://linear.app/vercel/issue/MKT-2822/add-missing-help-snapshot-test-for-integration-remove-subcommand) to track this.

tonypan2 (approved):

PR #15073, opened 2026-02-17
[cli] Add metrics command for querying observability data

ergunsh:
## Summary

  - Adds a new `vercel metrics` command (behind `FF_METRICS` feature flag) that queries the observability metrics API for time-series data across 24 event types
  - Includes two subcommands: `query` (default) for running aggregated queries with CSV/JSON output, and `schema` for listing available events, dimensions, and measures
  - Supports grouping, filtering (OData), custom time ranges, granularity control, and project/team scoping

  ## Details

  **Command structure:**
  - `vercel metrics -e incomingRequest --group-by httpStatus` — query with CSV output
  - `vercel metrics -e functionExecution -m requestDurationMs -a p95 --format=json` — query with JSON output
  - `vercel metrics schema` — list all 24 events
  - `vercel metrics schema -e incomingRequest` — show dimensions/measures for an event

  **Key design decisions:**
  - Uses `vercel.com/api/observability/metrics` endpoint (not `api.vercel.com`), matching the logs command pattern
  - Static schema registry for client-side validation and the `schema` subcommand (no API call needed)
  - Default aggregation inferred from measure units (sum for counts/bytes/tokens, avg for durations/ratios), mirroring the dashboard UI
  - Auto-selects granularity based on time range, with automatic adjustment when user-specified granularity is too small
  - Scope resolution: `--project`/`--all` use `vercel switch` team context; default uses linked project

  **Files added (22 files, ~4300 lines):**
  - `src/commands/metrics/` — command definition, query, schema, validation, output formatting, time utilities, types
  - `src/util/telemetry/commands/metrics/` — telemetry client with tracking for all flags
  - `test/unit/commands/metrics/` — 123 tests covering query, schema, validation, output, time utils, schema data, and routing

  ## Test plan

  - [x] Unit tests pass: `npx vitest run packages/cli/test/unit/commands/metrics/`
  - [x] Manual test with linked project: `FF_METRICS=1 vercel metrics -e incomingRequest`
  - [x] Manual test team-wide: `FF_METRICS=1 vercel metrics --all -e incomingRequest --group-by projectName`
  - [x] Manual test schema: `FF_METRICS=1 vercel metrics schema -e functionExecution --format=json`
  - [x] Verify feature flag gating: command is inaccessible without `FF_METRICS=1`

> [!NOTE]
> Low Risk Change
>
> This PR adds a new feature-flagged CLI command for querying observability metrics, consisting entirely of new files with command definitions, validation logic, output formatting, and comprehensive tests.
> 
> - New `vercel metrics` command behind FF_METRICS feature flag (~4300 lines)
> - Static schema registry with 24 event types for client-side validation
> - 123 unit tests covering query, schema, validation, and output formatting
>
> <sup>Risk assessment for [commit 9c18515](https://github.com/vercel/vercel/commit/9c18515e3f118d98c3073a301da8ccd3be8855b5).</sup>


packages/cli/src/commands/metrics/output.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export function getRollupColumnName(
+  measure: string,
+  aggregation: string
+): string {
+  return `${measure}_${aggregation}`;
+}
+
+export function formatCsv(
```
adriancooney, line 28: Just double checking: we don't have any existing pattern for rendering to CSV in the CLI? I'm wondering if we should be using a well-tested library here (e.g. `d3-csv`) which is battle hardened and covers the edge cases.

packages/cli/src/commands/metrics/query.ts
```diff
@@ -0,0 +1,373 @@
...
+): number {
+  let code: string;
+  let message: string;
+
+  switch (err.status) {
+    case 402:
+      code = 'PAYMENT_REQUIRED';
+      message =
+        'This feature requires an Observability Plus subscription. Upgrade at https://vercel.com/dashboard/settings/billing';
+      break;
```
adriancooney, line 68: Nice!

packages/cli/src/commands/metrics/query.ts
```diff
@@ -0,0 +1,373 @@
...
+    all: boolean | undefined;
+    jsonOutput: boolean;
+  }
+): Promise<{ scope: Scope; accountId: string } | number> {
+  // --project or --all: resolve team context via getScope
+  if (opts.project || opts.all) {
+    const { team } = await getScope(client);
+    if (!team) {
+      const errMsg =
+        'No team context found. Run `vercel switch` to select a team, or use `vercel link` in a project directory.';
```
adriancooney, line 109: Is there a common error message/system we can use here?

packages/cli/src/commands/metrics/query.ts
```diff
@@ -0,0 +1,373 @@
...
+    event,
+    rollups: { [rollupColumn]: { measure, aggregation } },
+    startTime: rounded.start.toISOString(),
+    endTime: rounded.end.toISOString(),
+    granularity: granResult.duration,
+    ...(groupBy.length > 0 ? { groupBy } : {}),
+    ...(filter ? { filter } : {}),
+    limit: limit ?? 10,
+    limitRanking: 'top_by_summary' as const,
+    tailRollup: 'truncate' as const,
```
adriancooney: Do we need to configure these? Are we using the same defaults in the dashboard?

packages/cli/src/commands/metrics/output.ts
ergunsh: Ah, good call! I couldn't see an existing pattern for rendering to CSV in the CLI. WDYT of moving forward with the hand rolled approach in this PR and improving it down line? (e.g. using `d3-csv`)

packages/cli/src/commands/metrics/query.ts
ergunsh: Ah, there isn't any system we can use here; but the error message shown by other commands is `Team not found.`. Should I update here too?

packages/cli/src/commands/metrics/query.ts
ergunsh: Ah, indeed they are the default values for the query engine. I've removed them!

packages/cli/src/commands/metrics/query.ts
adriancooney: Nope, this message is much better. We should always give the agents clues.

packages/cli/src/commands/metrics/query.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+    const errMsg = err instanceof Error ? err.message : String(err);
+    if (jsonOutput) {
+      client.stdout.write(formatErrorJson('NETWORK_ERROR', errMsg));
+    } else {
+      output.error(errMsg);
+    }
+    return 1;
+  }
+  output.stopSpinner();
```
adriancooney: We're calling this twice on error.

packages/cli/src/commands/metrics/schema-data.ts
adriancooney, line 1: How do we expect to solve drift in this file? Is this something we could request from the server? I'd much prefer we don't embed it in the CLI. Fine to ignore for the initial version but certainly something for release.

packages/cli/src/commands/metrics/time-utils.ts
```diff
@@ -0,0 +1,159 @@
...
+    return new Date(Date.now() - ms);
+  }
+  const date = new Date(input);
+  if (isNaN(date.getTime())) {
+    throw new Error(
+      `Invalid time format "${input}". Use relative (1h, 30m, 2d, 1w) or ISO 8601 datetime.`
+    );
+  }
+  return date;
+}
```
adriancooney, line 20: This can be replaced by the `ms` dependency.

packages/cli/src/commands/metrics/query.ts
```diff
@@ -0,0 +1,371 @@
...
+  const granMs = granularity
+    ? toGranularityMs(granularity)
+    : toGranularityMs(getAutoGranularity(rangeMs) || '1h');
+  const rounded = roundTimeBoundaries(
+    startTime,
+    endTime,
+    granResult.adjusted
+      ? toGranularityMsFromDuration(granResult.duration)
+      : granMs
+  );
```
adriancooney: I'm not sure I follow the granularity rounding here. Can you elaborate on how it works?

packages/cli/src/commands/metrics/time-utils.ts
ergunsh: Oh great, done!

packages/cli/src/commands/metrics/schema-data.ts
ergunsh: Discussed offline

packages/cli/src/commands/metrics/query.ts
ergunsh: Good catch on the readability! Technically it wasn't called twice since the catch block always returns before the post-try `stopSpinner()` but the structure definitely made it look that way. Moved it into a finally block so there's a single call to `stopSpinner()`.

packages/cli/src/commands/metrics/query.ts
ergunsh: I simplified the logic and added comments for clarity. There are two key pieces:

* **`computeGranularity`** determines the effective granularity based on the time range and optional `--granularity`.
    * If none is provided, it picks a sensible default (e.g. a 3-day range → `1h`). 
    * If the provided granularity is too small for the time range, it adjusts (e.g. a 3-day range, `--granularity` is `1m`, then it's auto adjusted to `1h`)

* **`roundTimeBoundaries`** aligns the start/end times to clean granularity boundaries before calling the API.

For example, with `--granularity 1h` and a range of `14:23–16:47`, the API would return partial buckets (37m, 60m, 47m), which aren’t directly comparable.
We fix this by flooring the start (`14:00`) and ceiling the end (`17:00`), so all buckets are full 1-hour intervals.

packages/cli/src/commands/metrics/command.ts
```diff
@@ -0,0 +1,200 @@
...
+    {
+      name: 'query',
+      aliases: [],
+      description: '',
+      default: true,
+      hidden: true,
+      arguments: [],
+      options: [],
+      examples: [],
+    },
```
adriancooney, line 53: Are we keeping this subcommand?

packages/cli/src/help.ts
```diff
@@ -4,6 +4,10 @@ import chalk from 'chalk';
 const packageName = 'vercel';
 const logo = '▲';
 
+const metricsLine = process.env.FF_METRICS
+  ? '\n      metrics                          Queries observability metrics for your project or team'
+  : '';
```
adriancooney, line 9: yikes. this is not ideal! we'll need to get a better solution in future.

packages/cli/src/commands/metrics/command.ts
ergunsh: This is a workaround for the `--help` case. It treats `command` as required unless a subcommand has `default: true`.

Without this, help renders as:
```
vercel metrics command
```

and with this, it renders as
```
vercel metrics [command]
```

packages/cli/src/help.ts
ergunsh: Agreed! (e.g. list the commands as an array, and render the help text based on that)

adriancooney (approved):
Great work! We're looking to get this out quick however I'm happy that the risk of affecting any production commands is extremely low (boundaries are tight).

PR #15074, opened 2026-02-17
fix(cli): Use device code flow for missing scope re-authentication

anatrajkovska:
If the user switches to a team using `vercel switch` which has SSO or MFA enforced and the user hasn't yet authorized the team, instead of redirecting them to the old SSO flow which is deprecated, now we will use the device code flow which supports passing the `team_id` query param that enforces the user to authorize the request before they can login.

> [!WARNING]
> High Risk Change
>
> This PR changes the authentication flow for team re-authentication, replacing the deprecated SAML/SSO login flow with device code flow, which is a refactor of the auth mechanism that requires careful review to ensure the new flow properly enforces authorization.
> 
> - Replaces SAML login and prompt-based re-auth with device code flow
> - Removes user confirmation prompt before SAML login
> - Always passes teamId for re-authentication (previously conditional on SAML)
>
> <sup>Risk assessment for [commit 794b171](https://github.com/vercel/vercel/commit/794b171237f9de27fc7250262bb68980cbbfe66b).</sup>


marcgreenstock (approved):

AAorris (approved):

mehulkar (approved):
stamping to get past code owners since Accounts team has already approved

PR #14614, opened 2026-01-14
Add private blobs to vercel cli

shawnfeldman:
requires https://github.com/vercel/storage/pull/924

Adds all private blob commands

PR #15075, opened 2026-02-17
feat(cli): add private Blob storage support

vvo:
## Summary

- New `blob get` command that downloads blobs by URL or pathname — streams content to stdout by default (pipeable), supports `--output` flag to save to file
- `--access` flag (`public` | `private`) on `put`, `copy`, `get`, and `create-store` commands — defaults to `public` for backward compatibility
- Display `Access: Public/Private` in `get-store` output
- Bump `@vercel/blob` to `2.3.0-cc8f7be-20260217140255` prerelease with private storage SDK support (`get()` method, `access: 'private'`)
- Shared `parseAccessFlag()` helper in `src/util/blob/access.ts`

Replaces #14614 with a clean implementation addressing review feedback:
- `--access` defaults to `public` (no breaking change to existing `vercel blob put file.txt` workflows)
- `blob get` streams content to stdout instead of only printing metadata
- `parseAccessFlag()` helper is shared, not duplicated across files
- Proper type import instead of type assertion hack

## Test plan

- [x] All existing blob tests pass (`npx vitest run packages/cli/test/unit/commands/blob/`)
- [x] New tests for `blob get` command (stdout streaming, `--output` file save, `--access private`, error cases, telemetry)
- [x] Updated `create-store` test assertions to include `access` in request body
- [ ] Manual testing with a real store:
  - `vercel blob put test-file.txt` (backward compat, no --access)
  - `vercel blob put test-file.txt --access private`
  - `vercel blob get test-file.txt --access private`
  - `vercel blob get test-file.txt --access private --output ./downloaded.txt`
  - `vercel blob get test-file.txt --access private | head -c 100`
  - `vercel blob create-store my-store --access private`
  - `vercel blob get-store`

> [!NOTE]
> Low Risk Change
>
> This PR adds new CLI commands and options for private blob storage support with a major dependency bump, but defaults to 'public' access for backward compatibility and adds proper validation.
> 
> - Major version bump of @vercel/blob from 1.0.2 to 2.3.0 prerelease
> - New `blob get` command and `--access` flag across multiple subcommands
> - Access flag defaults to 'public' with validation via parseAccessFlag helper
>
> <sup>Risk assessment for [commit 2f8ab76](https://github.com/vercel/vercel/commit/2f8ab76edc03626c597bd115dc30cda3981188be).</sup>

### DIFF
--- a/.changeset/blob-private-storage.md
+++ b/.changeset/blob-private-storage.md
@@ -4,7 +4,7 @@
 
 Add private Blob storage support:
 
-- Create private stores: `vercel blob store add my-store --access private`
+- Create private stores: `vercel blob create-store my-store --access private`
 - Upload to private stores: `vercel blob put file.txt --access private`
 - Download blobs with the new `blob get` command: `vercel blob get file.txt --access private` (works with both public and private stores)
 - Copy blobs: `vercel blob copy source.txt dest.txt --access private`
--- a/.changeset/blob-private-storage.md
+++ b/.changeset/blob-private-storage.md
@@ -2,4 +2,12 @@
 'vercel': minor
 ---
 
-Add private Blob storage support: new `blob get` command for downloading blobs, `--access` flag on `put`, `copy`, `get`, and `store add` commands (defaults to `public` for backward compatibility), and display access type in `store get` output
+Add private Blob storage support:
+
+- Create private stores: `vercel blob store add my-store --access private`
+- Upload to private stores: `vercel blob put file.txt --access private`
+- Download blobs with the new `blob get` command: `vercel blob get file.txt --access private` (works with both public and private stores)
+- Copy blobs: `vercel blob copy source.txt dest.txt --access private`
+- Display access type (Public/Private) in `vercel blob store get` output
+
+The `--access` flag defaults to `public` for backward compatibility.
--- a/.changeset/blob-private-storage.md
+++ b/.changeset/blob-private-storage.md
@@ -11,3 +11,8 @@ Add private Blob storage support:
 - Display access type (Public/Private) in `vercel blob store get` output
 
 The `--access` flag defaults to `public` for backward compatibility.
+
+Flatten blob store commands: `blob create-store`, `blob delete-store`, `blob get-store`.
+Rename `--force` to `--allow-overwrite`.
+Add conditional headers: `--if-match` for put/del/copy, `--if-none-match` for get.
+Old commands and options are deprecated but still work.
--- a/.changeset/blob-private-storage.md
+++ b/.changeset/blob-private-storage.md
@@ -1,0 +1,5 @@
+---
+'vercel': minor
+---
+
+Add private Blob storage support: new `blob get` command for downloading blobs, `--access` flag on `put`, `copy`, `get`, and `store add` commands (defaults to `public` for backward compatibility), and display access type in `store get` output
--- a/packages/cli/src/commands/blob/get.ts
+++ b/packages/cli/src/commands/blob/get.ts
@@ -38,7 +38,11 @@ export default async function get(
     flags,
     args: [urlOrPathname],
   } = parsedArgs;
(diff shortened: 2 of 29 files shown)


.changeset/blob-private-storage.md
```diff
@@ -0,0 +1,13 @@
+---
+'vercel': minor
+---
+
+Add private Blob storage support:
+
+- Create private stores: `vercel blob store add my-store --access private`
```
mehulkar: seems like a lot of nested subcommands... can we do `vercel blob add <name>` instead of `vercel blob store add <name>`

packages/cli/src/commands/blob/get.ts
```diff
@@ -0,0 +1,102 @@
...
+    } else {
+      await pipeline(nodeStream, client.stdout, { end: false });
+    }
+  } catch (err) {
+    output.stopSpinner();
+    printError(err);
+    return 1;
+  }
+
+  return 0;
```
mehulkar, line 107: does this need to be explicit?

packages/cli/src/commands/blob/put.ts
```diff
@@ -46,10 +48,19 @@ export default async function put(
     '--force': force,
   } = flags;
 
+  const access = accessFlag ?? 'public';
+  if (!isAccess(access)) {
+    output.error(
+      `Invalid access value: '${access}'. Must be 'public' or 'private'.`
+    );
+    return 1;
+  }
```
mehulkar: does this validation need consolidating? It's being added to every command separaetly

mehulkar (commented):
lgtm, but the DX of adding a store seems weird

mehulkar (approved):

.changeset/blob-private-storage.md
luismeyer: I designed it like this because I wanted to avoid confusion between managing stores and managing objects within the store.
Having `vercel blob add <name>` and `vercel blob put <name>` on the same subcommand level felt weird to me because it's so different

luismeyer (approved):

.changeset/blob-private-storage.md
vvo: We probably can get await with create-store, delete-store, connect-store to avoid one sublevel, thoughts @luismeyer?

vvo:
* **#15101** <a href="https://app.graphite.com/github/pr/vercel/vercel/15101?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15075** <a href="https://app.graphite.com/github/pr/vercel/vercel/15075?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/15075?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/blob-private-storage.md
vvo: Done in https://github.com/vercel/vercel/pull/15101

vvo (commented):
Addressed the review feedback — consolidated the access validation into a shared `parseAccessFlag()` helper in `src/util/blob/access.ts`.

packages/cli/src/commands/blob/get.ts
vvo: It seems to be the standard CLI convention, all command functions return explicit exit codes (0 for success, 1 for error). Example: https://github.com/vercel/vercel/blob/main/packages/cli/src/commands/whoami/index.ts#L63

packages/cli/src/commands/blob/put.ts
vvo: Good call! Consolidated this into a shared `parseAccessFlag()` helper in `src/util/blob/access.ts`. All 4 commands (`get`, `put`, `copy`, `store-add`) now use it:

```ts
const access = parseAccessFlag(accessFlag);
if (!access) return 1;
```

PR #15080, opened 2026-02-17
[python] Lambda runtime install cold-start performance improvements

gscho:
> [!WARNING]
> High Risk Change
>
> Large refactor of Python Lambda dependency bundling logic introduces new knapsack packing algorithm and changes runtime installation from `uv pip install` to `uv sync --inexact --frozen`, affecting cold-start behavior for oversized bundles.
> 
> - New `PythonDependencyExternalizer` class with knapsack algorithm for optimal Lambda packing
> - Runtime install switches from `uv pip install -r requirements.txt` to `uv sync --inexact --frozen`
> - Removes `generateRuntimeRequirements` function, replaces with JSON config-based approach
>
> <sup>Risk assessment for [commit 870da37](https://github.com/vercel/vercel/commit/870da373b1cdb3409212d2381f7e34556c2a9d24).</sup>


packages/python-analysis/src/manifest/uv-lock-parser.ts
```diff
@@ -173,30 +173,3 @@ export function classifyPackages(
...
   return { privatePackages, publicPackages, packageVersions };
 }
-
-/**
- * Generate requirements.txt content for runtime installation.
- *
- * Only includes public packages that will be installed at runtime from PyPI.
- * Private packages should be bundled separately.
- */
-export function generateRuntimeRequirements(
```
gscho, line 183: No longer needed b/c we use `uv sync` now.

python/vercel-runtime/src/vercel_runtime/vc_init.py
```diff
@@ -229,44 +229,78 @@ def send_message(message: dict):
...
+            # Writing pyvenv.cfg directly avoids spawning a subprocess.
+            os.makedirs(_site_packages, exist_ok=True)
+            with open(os.path.join(_deps_dir, "pyvenv.cfg"), "w") as _f:
+                _f.write(f"home = {os.path.dirname(sys.executable)}\n")
+                _f.write("include-system-site-packages = false\n")
+
+            # Use uv sync --inexact --frozen to install only the
+            # missing public packages. --inexact avoids removing
+            # packages already present in _vendor (bundled deps).
+            _sync_cmd = [
```
gscho, line 443: Instead of an inclusions list (`_runtime_requirements.txt`) we now use `uv sync` with a lockfile and pass a list of exclusions which are private packages and packages already existing inside the lambda zip.

python/vercel-runtime/src/vercel_runtime/vc_init.py
```diff
@@ -267,44 +267,81 @@ def send_message(message: dict):
...
+                _uv_path,
+                "sync",
+                "--inexact",
```
elprans, line 446: I think we should use `--no-cache` here also to prevent uv from persisting its cache.  This also makes `--link-mode=hardlink` unnecessary I think (it's also a bit dicey to use since we can't 100% guarantee there is no cross-fs boundary between source and dest).

packages/python/package-python-devel/.gitignore
```diff
@@ -0,0 +1,2 @@
+.vercel
```
elprans: Stray dir?

packages/python/src/dependency-externalizer.ts
```diff
@@ -0,0 +1,567 @@
...
+  /**
+   * Generate the optimally-packed Lambda bundle.
+   * Mutates `files` in place: adds vendor files (private + knapsack-selected
+   * public), runtime config, and uv binary.
+   * Must be called after analyze().
+   */
+  async generateBundle(files: Files): Promise<void> {
+    if (!this.analyzed) {
+      throw new Error(
+        'RuntimeDepInstall.analyze() must be called before generateBundle()'
```
elprans:
```suggestion
        'PythonDependencyExternalizer.analyze() must be called before generateBundle()'
```

elprans (approved):

PR #15085, opened 2026-02-17
chore: improve import builders tests so we can hide implementation details

mehulkar:
`resolveBuilders` is a private function, we don't need to unit test it separately. also makes it hard to tell if refactors break the implementation.

> [!NOTE]
> Low Risk Change
>
> This PR refactors tests to use public API instead of testing private function directly, with no changes to production logic beyond removing an export keyword.
> 
> - Removes `export` from `resolveBuilders` function to make it private
> - Adds mocks for `untracedInstallBuilders` in test file
> - Rewrites tests to call `importBuilders` public API instead of `resolveBuilders` directly
>
> <sup>Risk assessment for [commit 3bbd898](https://github.com/vercel/vercel/commit/3bbd8982d415ccaa5c5b57cf466c5f9b0f0f5d3d).</sup>


tknickman (approved):

packages/cli/src/util/build/import-builders.ts
```diff
@@ -81,7 +81,7 @@ export async function importBuilders(
   return importResult.builders;
 }
 
-export async function resolveBuilders(
+async function resolveBuilders(
```
dferber90, line 84: Removing the export caused issues on https://github.com/vercel/vercel/pull/15027, where the linter is now failing on the test. I wonder why it wasn't caught here. See [log](https://github.com/vercel/vercel/actions/runs/22154679735/job/64055199318?pr=15027).

Image

PR #15097, opened 2026-02-18
[services] strip services route prefix in python runtime

ricardo-agz:
* Strip service route prefix in python runtime when services are auto-detected so app sees `/ping` instead of `/_/backend/ping`.
* When `routePrefix` is manually set in `vercel.json`, we do _not_ strip the route prefix
* Sets `SCRIPT_NAME` with route prefix for WSGI
* Sets `root_path` with route prefix for ASGI

**Note:** There is duplicate code in the vercel python runtime and `vc_init_dev_asgi.py` and `vc_init_dev_wsgi.py`. Ideally, we should consolidate `vc_init_dev_asgi.py` and `vc_init_dev_wsgi.py` into a single `vc_init_dev.py` (later PR)

> [!WARNING]
> High Risk Change
>
> This PR adds route prefix stripping logic for Python runtime services, modifying request path handling in WSGI/ASGI environments which affects how requests are routed to user applications.
> 
> - Adds new `routePrefixSource` field to Service type to distinguish configured vs auto-detected prefixes
> - Injects `VERCEL_SERVICE_ROUTE_PREFIX_STRIP` env var to enable path stripping in Python runtime
> - Modifies WSGI/ASGI scope handling to strip service prefix from request paths and set root_path/SCRIPT_NAME
>
> <sup>Risk assessment for [commit 1b738d7](https://github.com/vercel/vercel/commit/1b738d79d78c9487752f121029fef783c49a49c7).</sup>


nsidnev (approved):

elprans (approved):

packages/python/vc_init_dev_asgi.py
```diff
@@ -45,6 +45,85 @@ def _color(text: str, code: str) -> str:
...
+
+    if not prefix.startswith('/'):
+        prefix = f'/{prefix}'
+
+    if prefix != '/':
+        prefix = prefix.rstrip('/')
+        if not prefix:
+            prefix = '/'
+
+    return '' if prefix == '/' else prefix
```
msullivan: I think this whole chunk is equivalent to `return prefix.rstrip('/')`

packages/python/vc_init_dev_wsgi.py
```diff
@@ -24,6 +24,59 @@ def _color(text: str, code: str) -> str:
...
+
+    if not prefix.startswith('/'):
+        prefix = f'/{prefix}'
+
+    if prefix != '/':
+        prefix = prefix.rstrip('/')
+        if not prefix:
+            prefix = '/'
+
+    return '' if prefix == '/' else prefix
```
msullivan: Likewise

packages/python/vc_init_dev_asgi.py
```diff
@@ -45,6 +45,85 @@ def _color(text: str, code: str) -> str:
...
+
+    updated_scope = dict(scope)
+    updated_scope['path'] = path_value
+
+    raw_path = scope.get('raw_path')
+    if isinstance(raw_path, (bytes, bytearray)):
+        try:
+            decoded = bytes(raw_path).decode('latin-1')
+            stripped_raw, _ = _strip_service_route_prefix(decoded)
+            updated_scope['raw_path'] = stripped_raw.encode('latin-1')
```
msullivan: I'm not sure how worried we are about this, but I think that this won't work if there are any non-ASCII characters in the route prefix.

`os.environ.get` uses utf-8 by default (unless we've overridden it somewhere).

One approach would be to use `('utf-8', 'surrogateescape')` as the arguments to decode and encode here, which would match what getenv does.
(surrogateescape turns invalid utf-8 sequences into reserved unicode characters, basically, so it shouldn't fail.)

Other option would be always doing the stripping on either bytes or latin1ified strs, I think?

packages/python/vc_init_dev_asgi.py
msullivan: Though maybe it's a microoptimization? I think then we'd want `return '' if prefix == '/' else prefix.rstrip('/')`

jeffsee55 (approved):

PR #15115, opened 2026-02-18
fix(cli): add prepayment billing validation to integration balance

bhrigu123:
## Summary
- The `vercel integration balance` command was failing with an unhelpful "No balance information found" error for non-prepayment integrations. This is because the balance/threshold feature only applies to integrations with prepayment billing plans, but the command had no validation for this.
- Added an early check that the integration has resources with `billingPlan.type === 'prepayment'` before calling the balance API, with a clear error message explaining this is a prepayment-only command.
- Updated the `--help` description to clarify the command is for prepayment integrations only.

**Context:** Only 3 real marketplace integrations currently use prepayment billing (deepinfra, xai, fal), so this command was silently failing for the other 44 integrations.

Closes MKT-2823

## Test plan
- [x] Existing balance tests pass (11 tests)
- [x] New test for non-prepayment integration error path
- [x] All integration command tests pass (183 tests)
- [x] All integration-resource command tests pass (52 tests)
- [x] Updated `create-threshold` test that was affected by mock data changes

#### Integration which does not support pre-payment billing
```sh
node ~/workspace/cli/vercel/packages/cli/dist/vc.js integration balance aws
Vercel CLI 50.18.2
Error: The integration aws does not use prepayment billing. This command is only for integrations with prepayment billing plans.
```

#### Valid integration with pre-payment
```sh
node ~/workspace/cli/vercel/packages/cli/dist/vc.js integration balance fal --team team_ZrswwWlkgu3cRre7Tk7Rr0HV
Vercel CLI 50.18.2
WARN! The "--team" option is deprecated. Please use "--scope" instead.
> Balances and thresholds for fal:
> ● installation
>     Balance: $112.77
> ● fal-orange-umbrella
>     Threshold: Spend $25.00 if balance goes below $5.00
```
#### When integration is not installed
```sh
node ~/workspace/cli/vercel/packages/cli/dist/vc.js integration balance fal
Vercel CLI 50.18.2
Error: No balance information found for this integration
```

> [!NOTE]
> Low Risk Change
>
> This PR adds input validation that RESTRICTS access to a CLI command for non-prepayment integrations, with safe fallback behavior (returns true on errors, allowing the command to proceed rather than blocking).
> 
> - Adds early validation check for prepayment billing plan before calling balance API
> - Error handling defaults to permissive (returns true) to avoid false rejections
> - Updates help text and adds test coverage for new validation path
>
> <sup>Risk assessment for [commit bd2d586](https://github.com/vercel/vercel/commit/bd2d5869981476f3a6f7747157b892b282b77828).</sup>


packages/cli/src/commands/integration/command.ts
```diff
@@ -243,7 +243,7 @@ export const balanceSubcommand = {
   name: 'balance',
   aliases: [],
   description:
-    'Shows the balances and thresholds of a specified marketplace integration',
+    'Shows the balances and thresholds of a marketplace integration that uses prepayment billing',
```
tonypan2, line 246: The description change here breaks 4 snapshot tests in `help.test.ts` — CI is red on all 6 `vitest-unit` platforms:

1. `integration help column width 40`
2. `integration help column width 80`
3. `integration help column width 120`
4. `integration balance subcommand help column width 120`

Fix: run `pnpm --filter vercel vitest-run --update` to regenerate snapshots.

packages/cli/src/commands/integration/balance.ts
```diff
@@ -59,6 +59,16 @@ export async function balance(client: Client, args: string[]) {
     return 1;
   }
 
+  if (
+    resources.length > 0 &&
```
tonypan2: The guard `resources.length > 0` means this check only runs when resources exist. When there are zero matched resources, the code falls through to the balance API (`GET /v1/integrations/installations/:id/billing/balance`) which returns a generic "No balance information found" — the same unhelpful message this PR aims to fix.

A cleaner approach might be to add the validation API-side in [`get-installation-balances.ts`](https://github.com/vercel/api/blob/main/services/api-integrations/src/endpoints/marketplace/get-installation-balances.ts), following the existing pattern from `update-resource-threshold.ts`. Something like:

```ts
// After the ownership check, before fetching balances:
const integration = await getIntegrationById(configuration.integrationId) as MarketplaceIntegration | null;
if (!integration) {
  throw new IntegrationNotFoundError(configuration.integrationId);
}

const hasPrepayment = integration.products?.some(
  p => p.billingPlan?.type === 'prepayment'
);
if (!hasPrepayment) {
  return res.status(400).json({
    error: { code: 'not_prepayment', message: 'This integration does not use prepayment billing.' }
  });
}
```

This would make the CLI-side validation in this PR unnecessary — the API would return a specific error that the CLI could surface directly, covering all edge cases (zero resources, missing `billingPlan` metadata, etc.).

tonypan2 (changes requested):

tonypan2 (approved):

PR #15122, opened 2026-02-18
[dev] allow to skip authentication and project linking for `vc dev`

nsidnev:
Allows to skip project setup for `vc dev` by setting `VERCEL_EXPERIMENTAL_DEV_SKIP_LINK` env var, so that users can start prototyping or test examples without having to authenticate and create projects they may not need later

> [!NOTE]
> Low Risk Change
>
> This PR adds an experimental env var that allows skipping authentication and project linking for the `vc dev` command, which is a deliberate security bypass for local development convenience - but it's opt-in via explicit env var and only affects local dev workflow, not production.
> 
> - Adds `dev` to subcommandsWithoutToken when experimental flag enabled, bypassing auth requirement
> - Skips setupAndLink flow for unlinked projects when VERCEL_EXPERIMENTAL_DEV_SKIP_LINK is set
> - Creates new experimental.ts utility to check env var flag
>
> <sup>Risk assessment for [commit db73250](https://github.com/vercel/vercel/commit/db73250c7133a75731992f2b3e07d0e4cef637b0).</sup>


packages/cli/src/util/dev/experimental.ts
```diff
@@ -0,0 +1,6 @@
+export function isExperimentalSkipDevLinkEnabled(): boolean {
```
ricardo-agz, line 3: To consolidate flags for services mode (since we already have a few going around), let's also auto enable this by default when `VERCEL_USE_EXPERIMENTAL_SERVICES` is enabled no?

ricardo-agz (approved):

PR #15167, opened 2026-02-19
[python] Look in all root dirs for entry points if the framework is django.

dnwpark:
Django projects tend to not have a standard `app` directory that they put their entry point into.

Find the entry point by either:
- parsing the manage.py to find settings.py, parse that then get the entrypoint file.
- fallback to looking for `app`/`application`

> [!NOTE]
> Low Risk Change
>
> This PR adds new Django entrypoint detection logic by parsing manage.py and settings.py files to find WSGI/ASGI application paths, with no security, auth, or data-related changes.
> 
> - New functions to parse Django settings module from manage.py and resolve WSGI/ASGI entrypoints
> - Rust AST parsing additions for extracting string constants and DJANGO_SETTINGS_MODULE values
> - Comprehensive test coverage for new Django entrypoint discovery logic
>
> <sup>Risk assessment for [commit 88a34b0](https://github.com/vercel/vercel/commit/88a34b0616dee819d1e58f4a7c9d5e65f2ec2d17).</sup>


packages/python/src/entrypoint.ts
```diff
@@ -81,9 +92,23 @@ export async function detectGenericPythonEntrypoint(
...
 
     // Search candidate locations using AST-based detection
-    const candidates = PYTHON_CANDIDATE_ENTRYPOINTS.filter(
-      (c: string) => !!fsFiles[c]
-    );
+    let base_candidates: string[];
+    if (_framework !== 'django') {
+      base_candidates = PYTHON_CANDIDATE_ENTRYPOINTS;
+    } else {
+      const rootGlobs = await glob('*', {
```
elprans: I think for Django we ought to look for `manage.py` and extract the server entrypoint from there.

elprans: Specifically look for `DJANGO_SETTINGS_MODULE` (but we should also respect it in the environment)

dnwpark: Makes sense. I'll parse the `manage.py` and look for where `DJANGO_SETTINGS_MODULE` is set.

elprans: Basically, you need to follow django's own resolution process.  I think it looks for `asgi.py`/`wsgi.py` in the app package (which would be next to settings), but please read the docs.

elprans: Actually, I think it;s just `WSGI_APPLICATION` in `settings.py`

packages/python/src/entrypoint.ts
```diff
@@ -80,23 +107,64 @@ export async function detectGenericPythonEntrypoint(
...
-        debug(`Detected Python entrypoint: ${candidate}`);
-        return candidate;
+        debug(`Using configured Python entrypoint: ${entry}`);
+        return entry;
       }
     }
 
-    // No valid entrypoint found via AST detection
-    return null;
+    // If the Django WSGI entrypoint exists and is valid, use it
```
elprans: comments seems misplaced

packages/python/src/entrypoint.ts
```diff
@@ -80,23 +107,64 @@ export async function detectGenericPythonEntrypoint(
...
+    const wsgiEntry = await getDjangoEntrypoint(workPath);
```
elprans: we should probably try harder to find `manage.py` and look one level of subdirectories deep?  Unless I'm misunderstanding what `workPath` means here that is.  @ricardo-agz, WDYT?

packages/python-analysis/crates/vercel_python_analysis/src/entrypoint.rs
```diff
@@ -77,13 +79,122 @@ pub(crate) fn contains_app_or_handler_impl(source: &str) -> bool {
...
+    };
+    walk_body(&mut finder, parsed.suite());
+    match finder.matches.len() {
+        1 => Some(finder.matches.into_iter().next().unwrap()),
+        _ => None,
+    }
+}
+
+/// Visitor that walks the AST and collects every DJANGO_SETTINGS_MODULE value
+/// from os.environ.setdefault('DJANGO_SETTINGS_MODULE', ...) calls.
```
msullivan, line 158: How canonical is this `os.environ.setdefault` pattern? There's obviously a million other ways to set an environment variable.

dnwpark: Hard to say. It is what you get when you start a new django project with `django-admin startproject`.

I did some searching for sample projects around github and none of them seem to have done anything else.

dnwpark: There is some older documentation about importing settings from other settings files in case there are multiples, and also you can override the settings within the code, and also pass an explicit settings file to `runserver`

packages/python/src/entrypoint.ts
dnwpark: I am now checking for immediate subdirs as well.

packages/python/src/entrypoint.ts
dnwpark: expanded comments to maybe make them clearer

packages/python-analysis/crates/vercel_python_analysis/src/entrypoint.rs
dnwpark: Looking further out, it seems many projects use a `Procfile` to specify the entry point to hosting apps. We should prioritize that first

dnwpark: Did some more digging, and a lot of projects use a `Procfile` to specify the entry point. Also I did find one more project that didn't specify `WSGI_APPLICATION` anywhere, trying to figure out what that's all about.

packages/build-utils/src/python.ts
```diff
@@ -22,3 +27,256 @@ export async function isPythonEntrypoint(
...
 }
+
+/**
+ * Read Procfile at workPath and parse the web process. Supports:
+ * - web: gunicorn <module> or <module>:<attr>
+ * - web: uvicorn <module> or <module>:<attr>
+ * - web: uwsgi <settings>.ini (reads module = <appModule> from the INI)
+ * Returns the corresponding Python file path or null.
+ */
+export async function getProcfileWebEntrypoint(
```
ricardo-agz: While I think parsing the Procfile _is_ something we should do, I think it should probably be more so to create a `vercel.json` with services or something rather that for it to be something we strictly rely on and depend on going forward

We currently have a way to specify a specific non-standard entrypoint with an `app` script in `pyproject.toml`. Maybe if we _can't_ find a standard entrypoint but we do see a `Procfile` it should prompt you in the CLI something like "We detected your entrypoint as __ from Procfile, does this look right? Y/n" and if you select yes it adds the app script to `pyproject.toml`

@elprans thoughts?

ricardo-agz: Actually, I thought Procfile was Heroku-specific but seems like it is sort of widely adopted by other platforms? So maybe we do actually want to support it natively but I still think it might be outside the scope of just Django entrypoint detection

msullivan (approved):
Probably get @elprans to take another look too though

packages/build-utils/src/python.ts
```diff
@@ -22,3 +27,58 @@ export async function isPythonEntrypoint(
...
+): Promise<string | null> {
+  const settingsModule = await getDjangoSettingsModule(workPath);
+  if (!settingsModule) return null;
+  const settingsPath = join(
+    workPath,
+    `${settingsModule.replace(/\./g, '/')}.py`
+  );
+  try {
+    const settingsContent = await fs.promises.readFile(settingsPath, 'utf-8');
+    const wsgiApplication = await getStringConstant(
```
ricardo-agz, line 80: Should we not also try to get the asgiApplication here? And I think we should probably prioritize `ASGI_APPLICATION` over `WSGI_APPLICATION` if it exists

packages/python-analysis/src/semantic/entrypoints.ts
```diff
@@ -42,3 +42,43 @@ export async function containsAppOrHandler(source: string): Promise<boolean> {
...
+ */
+export async function getStringConstant(
+  source: string,
+  name: string
+): Promise<string | null> {
+  const mod = await importWasmModule();
+  return mod.getStringConstant(source, name) ?? null;
+}
+
+/** Regex to detect os.environ.setdefault('DJANGO_SETTINGS_MODULE', '...') so we can skip WASM when absent */
```
ricardo-agz: Should we make this check a bit more liberal? So it also catches stuff like:
```python
os.environ.setdefault(
    "DJANGO_SETTINGS_MODULE",
    "hello.settings",
)
```

or 

```python
os.environ.setdefault(
    "DJANGO_SETTINGS_MODULE",  # default settings module
    "hello.settings"
)
```

Since it's just an optimization maybe it should just check for the existence of DJANGO_SETTINGS_MODULE

ricardo-agz (approved):

packages/python-analysis/src/semantic/entrypoints.ts
dnwpark: Good call!

packages/build-utils/src/python.ts
dnwpark: So `ASGI_APPLICATION` isn't a core django setting, but is used by other frameworks. I've added it for now but in the future we should be smarter about how we check for it.

PR #14244, opened 2025-11-12
[python] fastapi + flask build command support

ricardo-agz:
Fix lack of support for build & install commands for FastAPI / Flask projects.
Flask & FastAPI projects now support running a project configured build command or also defining a build command in `pyproject.toml` under `[tool.vercel.scripts]`:

```toml
[project]
name = "42-build-command"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "fastapi==0.121.1",
]

[tool.vercel.scripts]
build = "python build.py"
```

PR #15171, opened 2026-02-20
[python] run build command after install commands

msullivan:
Previously the project build command ran before dependencies were
installed, which meant build scripts couldn't import installed
packages, and which was inconsistent with our other builders.

Move the env setup and install command extraction closer
to where they're used, and run the build command after uv sync
completes. And now that the venv is created before the build,
run the build in the env also.

Support for python `buildCommand`/`installCommand` was added
in #14244, and they ran in the correct order but not I think in the
virtual env. The `installCommand` was made to run in the venv in #14418,
but this resulted in it running *after* `buildCommand`.
I think that we also broke putting `pip install` in buildCommand at some point,
because of it needing `--break-system-packages`

TODO:
- [x] Tests
- [x] Develop more confidence about how breaking this might be

> [!NOTE]
> Low Risk Change
>
> This PR reorders Python build/install command execution and changes the environment from spawnEnv to pythonEnv for build commands, which is a build process logic change but fixes incorrect ordering rather than introducing new risk.
> 
> - Moves build command execution to run after dependency installation
> - Changes build command environment from spawnEnv to pythonEnv (virtual env)
> - Adds test fixtures to verify build commands can import installed packages
>
> <sup>Risk assessment for [commit 20b7f05](https://github.com/vercel/vercel/commit/20b7f057963644d4a39ce85d84fbc37d81daa08c).</sup>

### DIFF
--- a/.changeset/long-dots-brake.md
+++ b/.changeset/long-dots-brake.md
@@ -1,0 +1,5 @@
+---
+'@vercel/python': minor
+---
+
+Run Python build commands _after_ install commands and in the virtual env
--- a/packages/python/src/index.ts
+++ b/packages/python/src/index.ts
@@ -118,56 +118,6 @@ export const build: BuildV3 = async ({
     throw err;
   }
 
-  // For Python frameworks, also honor project install/build commands (vercel.json/dashboard)
-  if (isPythonFramework(framework)) {
-    const {
-      cliType,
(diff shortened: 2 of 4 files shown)


dnwpark (approved):
Seems reasonable

ricardo-agz (approved):

PR #14749, opened 2026-01-27
[functions] URL encode cache tags

kldavis4:
### TL;DR

URL encode cache tags to properly handle special characters.

### What changed?

- Added URL encoding for cache tags using `encodeURIComponent` before they're sent to the cache API
- Created helper functions `encodeTags`, `encodeTag`, and `encodeOptions` to handle encoding
- Updated the cache wrapper to use these encoding functions when setting cache entries or expiring tags
- Added unit tests to verify proper encoding of tags with spaces, commas, ampersands, and other special characters

### How to test?

1. Create cache entries with tags containing special characters:
   ```js
   await cache.set('key', 'value', {
     tags: ['tag with spaces', 'tag&special=chars', 'tag,with,commas']
   });
   ```

2. Expire tags with special characters:
   ```js
   await cache.expireTag('tag&special');
   await cache.expireTag(['tag one', 'tag&two']);
   ```

3. Verify that the tags are properly encoded when sent to the cache API

### Why make this change?

Previously, cache tags containing special characters (spaces, commas, ampersands, etc.) were not properly handled when setting cache entries or expiring tags. This could lead to unexpected behavior when using tags like `"my tag"` or `"category,item"`. By URL encoding the tags, we ensure they are correctly processed by the cache API, improving reliability and preventing potential issues with tag-based cache operations.

PR #15180, opened 2026-02-20
[functions] Scope cache tag encoding to commas only and add encoding to purge calls

kldavis4:
# Problem

A customer reported that after the encodeURIComponent change in https://github.com/vercel/vercel/pull/14749, their tags containing colons (e.g. "key:id") stopped working with invalidateByTag. The tag was being stored as "key%3Aid" in the cache but the purge API was sending "key:id", causing a mismatch. The original change was only meant to support commas in tags (since commas are used as tag delimiters), but encodeURIComponent encoded far more characters than necessary, creating a breaking change in a minor release.

Because the [Purge](https://github.com/vercel/functions/blob/main/packages/purge/src/purge.ts) class already does encodeURIComponent on the whole joined tag string in getInvalidateUrl (line 102) for URL safety. If we also encodeURIComponent in the client wrapper, the tag value gets double-encoded — "key:id" → "key%3Aid" → "key%253Aid" in the URL. That happens to work (server decodes once to "key%3Aid" which matches the stored value), but it's fragile and means the encoding behavior depends on two separate layers both encoding in a specific way. If the Purge class ever changes its encoding, everything breaks.

More importantly, it doesn't fix the root issue — encodeURIComponent encodes characters like `:`, `&`, `=` that were never a problem. The only character that actually breaks things is `,` because it's used as the tag delimiter in both the `x-vercel-cache-tags` header and the `?tags=` query parameter. Scoping to commas solves the actual problem without changing the semantics of tags that were already working.

# Solution

This change modifies cache tag encoding to only encode commas (`%2C`) instead of using full `encodeURIComponent`, preserving special characters like colons and ampersands in cache tags. The encoding is now consistently applied to both cache operations and purge functions (`invalidateByTag` and `dangerouslyDeleteByTag`).

**Changes:**
- Replaced `encodeURIComponent` with comma-specific encoding in cache functions
- Added comma encoding to `invalidateByTag` and `dangerouslyDeleteByTag` purge operations
- Updated tests to verify comma encoding while preserving other special characters

**The net change here from prior to #14749 is to revert the excessive encoding of tags and ensure all tag calls now encode just commas, which should solve this customer's issue and continue to solve the original customer's issue.**

---
changeset

> [!WARNING]
> High Risk Change
>
> Changes cache tag encoding from full encodeURIComponent to comma-only replacement, affecting how cache tags are stored and matched across cache operations and purge functions - a breaking change in encoding behavior that could cause cache mismatches if not deployed consistently.
> 
> - Replaces `encodeURIComponent` with comma-to-exclamation replacement for cache tags
> - Applies new encoding to `invalidateByTag`, `dangerouslyDeleteByTag`, and `addCacheTag`
> - Encoding change affects cache key matching - mismatch could cause purge failures
>
> <sup>Risk assessment for [commit b9367d2](https://github.com/vercel/vercel/commit/b9367d2d13740be2c01e5b5d40c5539c5914b05e).</sup>

### DIFF
--- a/.changeset/fix-cache-tag-encoding.md
+++ b/.changeset/fix-cache-tag-encoding.md
@@ -2,4 +2,4 @@
 '@vercel/functions': patch
 ---
 
-fix(functions): scope cache tag encoding to commas only instead of full `encodeURIComponent` to avoid breaking tags containing colons, ampersands, and other non-comma special characters. Also encode commas in `invalidateByTag`, `dangerouslyDeleteByTag`, and `addCacheTag` for consistency.
+fix(functions): scope cache tag encoding to commas only instead of full `encodeURIComponent` to avoid breaking tags containing colons, ampersands, and other non-comma special characters. Commas are replaced with `!`. Also encode commas in `invalidateByTag`, `dangerouslyDeleteByTag`, and `addCacheTag` for consistency.
--- a/.changeset/fix-cache-tag-encoding.md
+++ b/.changeset/fix-cache-tag-encoding.md
@@ -1,0 +1,5 @@
+---
+'@vercel/functions': patch
+---
+
+fix(functions): scope cache tag encoding to commas only instead of full `encodeURIComponent` to avoid breaking tags containing colons, ampersands, and other non-comma special characters. Also encode commas in `invalidateByTag`, `dangerouslyDeleteByTag`, and `addCacheTag` for consistency.
--- a/packages/functions/docs/modules/index.md
+++ b/packages/functions/docs/modules/index.md
@@ -78,7 +78,7 @@ A promise that resolves when the tag is added.
 
 #### Defined in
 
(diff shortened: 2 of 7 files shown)


kldavis4:
* **#15180** <a href="https://app.graphite.com/github/pr/vercel/vercel/15180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/15180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

styfle (commented):
I think this is going to break `invalidateBySrcImage(tag)` which calls `invalidateByTag('_vc_srcimg:tag')`

It currently encodes the colon so removing the encoding could break it.

QuiiBz (approved):

styfle (approved):
Lets ship this as a quick fix, but in the future we need to changes this to `!` instead so we don't cause tags to exceed the maximum length.

PR #15187, opened 2026-02-20
[services] detect project root to prevent accidental service as a project setup

nsidnev:
In a multi-service project setup, running `vc` commands outside the root directory can result in a separate project being created in for another directory, which is undesirable. With these changes, the CLI will detect that the command is being run in a subdirectory of the project's root directory and switch location, ensuring `vc` commands are executed from the root

> [!NOTE]
> Low Risk Change
>
> This PR adds a new utility function to detect project root directories and integrates it into multiple CLI commands (build, dev, link), which changes working directory resolution behavior but is gated behind an experimental feature flag.
> 
> - New `resolveProjectCwd` function changes cwd resolution in build, dev, and link commands
> - Feature gated behind `isExperimentalServicesEnabled()` flag
> - Comprehensive unit tests added for new functionality
>
> <sup>Risk assessment for [commit ea66577](https://github.com/vercel/vercel/commit/ea665776e64ee0af155f8a39c238849bd55e28db).</sup>


packages/cli/src/util/projects/find-project-root.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * When experimental services are enabled and the current directory is inside
+ * a service subdirectory, this returns the project root instead so that
+ * commands operate on the whole project rather than a single service.
+ */
+export async function resolveProjectCwd(cwd: string): Promise<string> {
+  if (!isExperimentalServicesEnabled()) return cwd;
+
+  const projectRoot = await findProjectRoot(cwd);
+  if (projectRoot && projectRoot !== cwd) {
```
gscho, line 68: What if this is false -> should we error?

nsidnev: Not really, If this is `false`, it basically means that the current directory could be considered a valid project root directory, and nothing else was suitable

gscho (approved):

ricardo-agz (approved):

PR #15190, opened 2026-02-20
[build-utils] support lambda ephemeral storage setting

gscho:
Allows the python builder to increase the ephemeral storage setting from 512MB to 1024MB for certain python functions.

We only want to do this when a function is over the current 500MB limit and the max will likely stay 1024MB as functions larger than that have very slow cold start times.

This functionality is currently behind the flag `VERCEL_PYTHON_INCREASED_EPHEMERAL_STORAGE` since the API does not support this yet.

> [!WARNING]
> High Risk Change
>
> This PR adds a new Lambda configuration option for ephemeral storage size, gated behind a feature flag, with extensive debug logging added to the Python runtime - infrastructure change affecting Lambda deployment configuration.
> 
> - Infrastructure: adds `ephemeralStorageSize` option to Lambda class affecting AWS Lambda configuration
> - Feature flag gated: `VERCEL_PYTHON_INCREASED_EPHEMERAL_STORAGE` controls new 2048MB storage request
> - Extensive debug logging added to Python runtime initialization
>
> <sup>Risk assessment for [commit 8f8aa5b](https://github.com/vercel/vercel/commit/8f8aa5b0309baffaaf1215f91d823fea1a1ddbb6).</sup>


packages/python/src/dependency-externalizer.ts
```diff
@@ -22,12 +22,27 @@ const readFile = promisify(fs.readFile);
...
 export const LAMBDA_SIZE_THRESHOLD_BYTES = 249 * 1024 * 1024;
+
 // Pack Lambda up to 245MB to leave a buffer
 export const LAMBDA_PACKING_TARGET_BYTES = 245 * 1024 * 1024;
 
-// AWS Lambda ephemeral storage (/tmp) is 512MB. Use 500MB to leave a buffer
-// for runtime overhead (.pyc generation, uv cache, metadata, etc.)
-export const LAMBDA_EPHEMERAL_STORAGE_BYTES = 500 * 1024 * 1024;
+// The /tmp size in MB we request from AWS when increased ephemeral storage is enabled.
+export const REQUESTED_TMP_SIZE_MB = 1024;
```
elprans: Where is this used?

packages/python/src/dependency-externalizer.ts
```diff
@@ -55,6 +70,9 @@ export class PythonDependencyExternalizer {
   private totalBundleSize: number = 0;
   private analyzed = false;
 
+  // Populated by generateBundle()
+  public needsExtraEphemeralStorage = false;
```
elprans, line 74: Why is this a bool and not a specific number?

packages/python/src/dependency-externalizer.ts
gscho: Thanks - that would have been an annoying runtime bug since we want to request 1024MB and leave some buffer.

packages/python/src/dependency-externalizer.ts
gscho: The intention is to leave the ephemeral storage size undefined since non of the other builders will be setting this value. More closely aligns with them.

elprans: Still not clear why this isn't a `number | undefined`?  Why are we choosing to enable ephemeral storage in discrete steps as opposed to "storage needed + buffer" asserting for the limit?

gscho: > Still not clear why this isn't a number | undefined?

I think thats more confusing because we'd compute the ephemeral storage size needed and if its less than 500 it returns undefined. Also I didn't want to get into the exact storage needed + buffer calculations. You either get `512MB` or `1024MB`. I don't see a reason to do that unless we're very concerned about the cost of this storage right now.

elprans: ok

elprans (approved):

PR #15203, opened 2026-02-22
fix(config): convert path-to-regexp patterns to regex in getConfig()

mitul-s:
The previous fix (#15134) only addressed pattern conversion when rewrite() and redirect() methods returned Route objects with transforms or env vars. This fix extends the conversion to the getConfig() method, which converts redirects, rewrites, and headers to routes format when routes are present.

Without this fix, patterns like `:path*` would be passed through literally instead of being converted to proper regex, causing 404 errors.

https://claude.ai/code/session_01CeTsDHjUb1wrKjMEBCsv87

### DIFF
--- a/.changeset/fix-404-getconfig-conversion.md
+++ b/.changeset/fix-404-getconfig-conversion.md
@@ -1,0 +1,5 @@
+---
+"@vercel/config": patch
+---
+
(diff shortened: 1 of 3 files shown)


mitul-s:
oops ignore claude

PR #15206, opened 2026-02-23
[cli] Send cache tags as array with comma encoding in invalidate and dangerously-delete commands

kldavis4:
### TL;DR

Modified cache invalidation and deletion commands to send tags as arrays instead of comma-separated strings to the API.

### What changed?

The `cache invalidate` and `cache dangerously-delete` commands now parse comma-separated tag strings into arrays before sending them to the API endpoints. The tag processing logic splits on commas and replaces any remaining commas with exclamation marks to handle edge cases.

### How to test?

Run the cache commands with single and multiple tags:
- `vercel cache invalidate --tag foo`
- `vercel cache invalidate --tag foo,bar,baz`
- `vercel cache dangerously-delete --tag foo`
- `vercel cache dangerously-delete --tag foo,bar,baz --revalidate 60`

Verify that the API receives the tags as an array format instead of a string.

### Why make this change?

This change aligns the CLI with the expected API format where tags should be sent as an array rather than a comma-separated string, ensuring proper communication between the client and server.

### DIFF
--- a/packages/cli/src/commands/cache/dangerously-delete.ts
+++ b/packages/cli/src/commands/cache/dangerously-delete.ts
@@ -66,11 +66,12 @@ export default async function dangerouslyDelete(
   let postUrl = '';
   let postBody = {};
   if (tag) {
-    itemName = plural('tag', tag.split(',').length, false);
+    const tags = tag.split(',').map((t: string) => t.replace(/,/g, '!'));
+    itemName = plural('tag', tags.length, false);
     itemValue = tag;
     flag = '--tag';
     postUrl = '/v1/edge-cache/dangerously-delete-by-tags';
-    postBody = { tags: tag, revalidationDeadlineSeconds: revalidate };
+    postBody = { tags, revalidationDeadlineSeconds: revalidate };
   } else if (srcimg) {
     itemName = 'source image';
     itemValue = srcimg;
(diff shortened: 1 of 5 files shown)


kldavis4:
* **#15206** <a href="https://app.graphite.com/github/pr/vercel/vercel/15206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/vercel/15206?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/commands/cache/dangerously-delete.ts
```diff
@@ -66,11 +66,12 @@ export default async function dangerouslyDelete(
   let postUrl = '';
   let postBody = {};
   if (tag) {
-    itemName = plural('tag', tag.split(',').length, false);
+    const tags = tag.split(',').map((t: string) => t.replace(/,/g, '!'));
```
styfle, line 69: This won't do anything since splitting on `,` means there are no more commas to replace.

I think for this PR to do anything, we would have too support something like:

```
--tag one --tag two --tag three
```

because the existing format will always be split

```
--tag one,two,three
```

PR #15213, opened 2026-02-23
Revert "[functions] URL encode cache tags"

styfle:
Reverts vercel/vercel#14749 since this caused a regression when using `:` and other characters in tags.

The expected behavior is that commas split into multiple tags.

> [!WARNING]
> High Risk Change
>
> Revert removes URL encoding of cache tags, which could allow special characters like commas to be interpreted as tag delimiters, potentially affecting cache invalidation behavior.
> 
> - Removes encodeURIComponent from cache tags in set() and expireTag()
> - Tags with special characters (commas, spaces, ampersands) now passed through unencoded
> - Removes associated URL encoding tests
>
> <sup>Risk assessment for [commit 83d1d76](https://github.com/vercel/vercel/commit/83d1d7601b8ecaf69a01e90cdd612dcd5aeaea30).</sup>

### DIFF
--- a/.changeset/revert-url-encode.md
+++ b/.changeset/revert-url-encode.md
@@ -1,0 +1,6 @@
+---
+'@vercel/functions': patch
+---
+
+[functions] Revert "[functions] URL encode cache tags"
+
--- a/packages/functions/docs/modules/index.md
+++ b/packages/functions/docs/modules/index.md
@@ -261,7 +261,7 @@ An instance of the Vercel Runtime Cache.
 
 #### Defined in
 
-[packages/functions/src/cache/index.ts:55](https://github.com/vercel/vercel/blob/main/packages/functions/src/cache/index.ts#L55)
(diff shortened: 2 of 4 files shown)


kldavis4 (commented):
need a changeset

.changeset/revert-url-encode.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vercel/functions': patch
```
kldavis4, line 2: the change being reverted was minor (based on @QuiiBz 's direction) so probably makes sense for this to be the same

styfle: @kldavis4 I would consider this a patch since the previous change introduced broken behavior and we are fixing the broken behavior, therefore patch.

kldavis4:
@QuiiBz here is the relevant thread for the original PR that this is reverting: https://vercel.slack.com/archives/C02K2HCH5V4/p1769213460081909

We discussed today and came to the conclusion that this was the wrong change and we should not support commas in tags.

lubakravche (approved):

QuiiBz (approved):

PR #15220, opened 2026-02-23
feat(cli): browser-based terms acceptance for non-interactive environments

bhrigu123:
## Summary

- When `vercel integration add` runs in auto-provision mode and the client is an AI agent or non-TTY terminal, the CLI now opens a browser for terms acceptance instead of erroring out
- The CLI polls for installation creation (every 2s, up to 5 min timeout) and continues the provisioning flow once terms are accepted in the browser
- Interactive TTY behavior remains unchanged

## Changes

- **New**: `packages/cli/src/util/integration/accept-terms-via-browser.ts` — utility that opens browser, shows spinner, and polls `fetchInstallations()` for the new installation
- **Modified**: `packages/cli/src/commands/integration/add-auto-provision.ts` — conditional browser/TTY flow based on `client.isAgent || !client.stdin.isTTY`
- **Modified**: test mocks and tests — updated existing non-TTY/agent tests to expect browser flow, added timeout/skip/URL-params tests

## Companion PR

- Frontend (accept-terms page + API route): https://github.com/vercel/front/pull/63048

## Test plan

- [x] All 77 unit tests pass (`vitest run packages/cli/test/unit/commands/integration/add-auto-provision.test.ts`)
- [ ] Manual E2E: run `vercel integration add <integration> --agent` and verify browser opens, terms page loads, and CLI continues after acceptance
- [ ] Verify interactive TTY flow is unchanged

> [!NOTE]
> Low Risk Change
>
> This PR adds a new browser-based terms acceptance flow for non-interactive CLI environments (AI agents or non-TTY), which is a feature addition that maintains the existing interactive TTY behavior and doesn't weaken any security controls.
> 
> - New utility `accept-terms-via-browser.ts` opens browser and polls for installation creation
> - Conditional flow in `add-auto-provision.ts` routes non-TTY/agent clients to browser flow
> - Test updates verify browser flow behavior and timeout handling
>
> <sup>Risk assessment for [commit a7ceacd](https://github.com/vercel/vercel/commit/a7ceacdcebab0ff64dc5b72cad01ea90354df935).</sup>

### DIFF
--- a/packages/cli/src/commands/integration/add-auto-provision.ts
+++ b/packages/cli/src/commands/integration/add-auto-provision.ts
@@ -7,6 +7,7 @@ import getScope from '../../util/get-scope';
 import { autoProvisionResource } from '../../util/integration/auto-provision-resource';
 import { fetchIntegrationWithTelemetry } from '../../util/integration/fetch-integration';
 import { fetchInstallations } from '../../util/integration/fetch-installations';
+import { acceptTermsViaBrowser } from '../../util/integration/accept-terms-via-browser';
 import { promptForTermAcceptance } from '../../util/integration/prompt-for-terms';
 import { selectProduct } from '../../util/integration/select-product';
 import type {
@@ -155,15 +156,33 @@ export async function addAutoProvision(
 
   let acceptedPolicies: AcceptedPolicies = {};
   if (!teamInstallation) {
-    const policies = await promptForTermAcceptance(client, integration);
-    if (!policies) {
-      telemetry.trackMarketplaceEvent('marketplace_install_flow_dropped', {
-        ...baseProps,
-        reason: 'policy_declined',
-      });
-      return 1;
+    if (client.isAgent || !client.stdin.isTTY) {
+      // Browser flow: open browser for terms acceptance, poll for installation
+      const browserInstallation = await acceptTermsViaBrowser(
+        client,
+        integration,
+        team.id
+      );
+      if (!browserInstallation) {
+        telemetry.trackMarketplaceEvent('marketplace_install_flow_dropped', {
+          ...baseProps,
+          reason: 'browser_terms_timeout',
+        });
+        return 1;
+      }
+      // acceptedPolicies stays {} — installation already created with policies in browser
+    } else {
+      // Interactive TTY: keep existing prompt behavior
+      const policies = await promptForTermAcceptance(client, integration);
+      if (!policies) {
+        telemetry.trackMarketplaceEvent('marketplace_install_flow_dropped', {
+          ...baseProps,
+          reason: 'policy_declined',
+        });
+        return 1;
+      }
+      acceptedPolicies = policies;
     }
-    acceptedPolicies = policies;
   }
 
   // Validate metadata flags (if provided) BEFORE prompting for resource name
--- a/packages/cli/src/commands/integration/add-auto-provision.ts
+++ b/packages/cli/src/commands/integration/add-auto-provision.ts
@@ -155,13 +155,15 @@ export async function addAutoProvision(
   );
 
   let acceptedPolicies: AcceptedPolicies = {};
+  let browserInstallationId: string | undefined;
   if (!teamInstallation) {
     if (client.isAgent || !client.stdin.isTTY) {
       // Browser flow: open browser for terms acceptance, poll for installation
       const browserInstallation = await acceptTermsViaBrowser(
         client,
         integration,
-        team.id
+        team.id,
+        contextName
       );
       if (!browserInstallation) {
         telemetry.trackMarketplaceEvent('marketplace_install_flow_dropped', {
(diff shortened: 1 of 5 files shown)


bhrigu123 (commented):
VADE suggestion applied in f62ea67 — changed `output.debug` to `output.log` so users can see the URL when the browser fails to open in headless environments.

packages/cli/src/util/integration/accept-terms-via-browser.ts
```diff
@@ -0,0 +1,59 @@
...
+  client: Client,
+  integration: Integration,
+  teamId: string,
+  timeoutMs: number = DEFAULT_TIMEOUT_MS
+): Promise<IntegrationInstallation | null> {
+  const url = new URL('https://vercel.com/api/marketplace/cli');
+  url.searchParams.set('cmd', 'accept-terms');
+  url.searchParams.set('integrationId', integration.id);
+  url.searchParams.set('source', 'cli');
+
```
tonypan2, line 22: Should `teamId` be included in the URL here? Both other CLI-to-browser handoffs to `/api/marketplace/cli` in this file pass `teamId` (`provisionResourceViaWebUI` at line 312, `handleManualVerificationAction` at line 676). Without it, if the user's browser session has a different active team than the CLI, the installation gets created for the wrong team and the CLI polls for `ownerId === teamId` forever until the 5-minute timeout.

The `teamId` is already available as a parameter — just not included in the URL.

*🤖 Written by Claude Code on behalf of @tonypan2*

packages/cli/src/util/integration/accept-terms-via-browser.ts
```diff
@@ -0,0 +1,59 @@
...
+const POLL_INTERVAL_MS = 2000;
+const DEFAULT_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
+
+export async function acceptTermsViaBrowser(
+  client: Client,
+  integration: Integration,
+  teamId: string,
+  timeoutMs: number = DEFAULT_TIMEOUT_MS
+): Promise<IntegrationInstallation | null> {
+  const url = new URL('https://vercel.com/api/marketplace/cli');
```
tonypan2: Alternative to adding `teamId`: could the CLI open the accept-terms page directly instead of going through the `/api/marketplace/cli` intermediary? The CLI already has `team` (from `getScope()`) and `integration.slug`, so it could construct something like `https://vercel.com/{teamSlug}/~/integrations/accept-terms/{integrationSlug}` — similar to how the auto-provision fallback at line 298 opens `fallback.url` directly.

This would solve the team context issue (team slug is in the URL), and remove the deployment dependency on the companion frontend PR adding a `cmd=accept-terms` case to the route handler.

*🤖 Written by Claude Code on behalf of @tonypan2*

packages/cli/src/commands/integration/add-auto-provision.ts
```diff
@@ -150,15 +151,33 @@ export async function addAutoProvision(
...
+        team.id
+      );
+      if (!browserInstallation) {
+        telemetry.trackMarketplaceEvent('marketplace_install_flow_dropped', {
+          ...baseProps,
+          reason: 'browser_terms_timeout',
+        });
+        return 1;
+      }
+      // acceptedPolicies stays {} — installation already created with policies in browser
```
tonypan2: `browserInstallation` is fetched during polling but only null-checked — the installation ID isn't passed to `autoProvisionResource`. The server-side auto-provision endpoint already accepts an optional `installationId` in the request body (see `resolveInstallation` in `auto-provision-resource.ts`) which lets it skip the candidate discovery and use the exact installation. Consider passing `browserInstallation.id` through to eliminate the re-discovery step and avoid any eventual-consistency edge case.

*🤖 Written by Claude Code on behalf of @tonypan2*

tonypan2 (changes requested):
some comments

packages/cli/src/util/integration/accept-terms-via-browser.ts
bhrigu123: Fixed in a7ceacd — went with the direct URL approach from the next comment. The CLI now opens `/{teamSlug}/~/integrations/accept-terms/{integrationSlug}?source=cli` directly, so the team context is in the URL path and there's no session mismatch risk.

packages/cli/src/util/integration/accept-terms-via-browser.ts
bhrigu123: Applied in a7ceacd — the CLI now constructs the page URL directly using `contextName` (team slug) and `integration.slug`, bypassing the `/api/marketplace/cli` intermediary entirely. This also removes the deployment dependency on the companion frontend PR's route handler.

packages/cli/src/commands/integration/add-auto-provision.ts
bhrigu123: Fixed in a7ceacd — `browserInstallation.id` is now passed through to `autoProvisionResource` as `installationId`, so the server can skip candidate discovery.

tonypan2 (approved):

tonypan2 (approved):

PR #15238, opened 2026-02-24
[python] Parse entrypoint variable name.

dnwpark:
Python entrypoint detection methods may find a variable other than `app`, `application` or `handler`. We should use it in `vc_init` when available.

Not sure what to do about `start-dev-server.ts` though...

Also, maybe we should support `app.py:run` as an explicit entrypoint format.

> [!WARNING]
> High Risk Change
>
> This PR changes Python entrypoint detection logic from boolean checks to returning variable names, modifying how the runtime resolves and loads application handlers - a change to critical build/runtime infrastructure that could affect all Python deployments.
> 
> - Changes entrypoint detection from boolean to returning variable name across multiple packages
> - Modifies vc_init.py runtime logic for resolving app/handler variables
> - Alters return types of entrypoint detection functions from string to [string, string] tuples
>
> <sup>Risk assessment for [commit 73df0ea](https://github.com/vercel/vercel/commit/73df0ea5ad6bafd8222726165a06cb016958fdea).</sup>

### DIFF
--- a/.changeset/little-wasps-wonder.md
+++ b/.changeset/little-wasps-wonder.md
@@ -1,0 +1,8 @@
+---
+'@vercel/python-analysis': minor
+'@vercel/python-runtime': minor
+'@vercel/build-utils': minor
+'@vercel/python': minor
+---
+
+Use specified entrypoint variable for python when available.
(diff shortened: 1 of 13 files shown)


dnwpark:
superceded by #15614

PR #15255, opened 2026-02-25
[ci] set release fetch depth

ricardo-agz:
Fixes issue where we always run the python PyPI release process even if the version has not changed.

Default fetch depth when unset is 1, which causes `HEAD^` to fail and previous version to fall back to '', which triggers the release process since currentVersion !== previousVersion in `publish.mjs`

> [!NOTE]
> Low Risk Change
>
> This PR adds a fetch-depth configuration to the GitHub Actions checkout step to fix a CI release process bug, which is a minor infrastructure change with low risk.
> 
> - CI: adds fetch-depth: 2 to checkout action in release workflow
> - Empty changeset file added
>
> <sup>Risk assessment for [commit d4bbdbd](https://github.com/vercel/vercel/commit/d4bbdbd1a8130918208a1c96e50a4b79ec357381).</sup>


.github/workflows/release.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.GH_TOKEN_PULL_REQUESTS }}
+          fetch-depth: 2
```
elprans, line 29: I would add a comment here as to why so that it doesn't get dropped unknowingly

elprans (approved):

tknickman (approved):

PR #15260, opened 2026-02-25
[cli] fix: dev server for large buffers

ecklf:
## Problem

When vercel dev builds a Lambda with large dependencies, the builder worker process creates a `zipBuffer` containing all the bundled files. This buffer is sent back to the main CLI process via Node.js IPC using `process.send()`.

## Solution

For large zipBuffers (>256MB), write the buffer to a temporary file instead of sending it through IPC:
1. Worker side (`builder-worker.cjs`): After creating the zip, if it exceeds the threshold, write it to a temp file and send the file path instead of the buffer
2. Main process side (`builder.ts`): When receiving the build result, check for zipBufferPath. If present, read the file from disk and delete the temp file
This completely bypasses the JSON serialization memory amplification for large payloads while maintaining the non-fs overhead for smaller builds.

> [!NOTE]
> Low Risk Change
>
> This PR adds a performance optimization to write large zip buffers to temp files instead of IPC, with proper cleanup, and refactors deserialization logic with improved type safety - no security, auth, or schema changes.
> 
> - Adds temp file handling for zip buffers >256MB to avoid OOM in dev server
> - Refactors IPC deserialization with explicit TypeScript interfaces
> - Fixes promise error handling in Awaiter class
>
> <sup>Risk assessment for [commit 2ebd31b](https://github.com/vercel/vercel/commit/2ebd31b26e8e3fa72d221c99c1f9fddc5aefee50).</sup>


packages/cli/src/util/dev/builder-worker.cjs
```diff
@@ -44,14 +53,38 @@ async function processMessage(message) {
...
+    // Delete files after creating zip to avoid OOM when serializing via IPC.
+    // The zipBuffer contains all file data, so files is no longer needed.
+    delete output.files;
+
+    // For large zip buffers, write to a temp file instead of sending via IPC.
+    // JSON serialization of large Buffers causes OOM because each byte becomes
+    // a separate array element (e.g., {"type":"Buffer","data":[1,2,3,...]}).
+    if (zipBuffer.length > ZIP_BUFFER_FILE_THRESHOLD) {
+      const tempDir = os.tmpdir();
+      const randomId = crypto.randomBytes(8).toString('hex');
```
ofhouse, line 68: **Question:** Do we have to take care of the cleanup?
I mean true it's a temporary file, but is it fine to just leave it there when dev command stops?

Is it handled the same way in other places as well?

packages/cli/src/util/dev/builder.ts
```diff
@@ -48,9 +47,124 @@ interface BuildMessage {
...
+    }
+    case 'Lambda': {
+      const lambda: BuiltLambda = Object.assign(
+        Object.create(Lambda.prototype),
+        obj
+      );
+      // Convert the JSON-ified Buffer object back into an actual Buffer,
+      // or read from temp file if it was too large for IPC
+      if (obj.zipBufferPath) {
+        const { readFileSync, unlinkSync } = await import('fs');
```
ofhouse: Can we use top-level import for that, as dynamic import is not a common pattern in the repo yet?

ofhouse (approved):

packages/cli/src/util/dev/builder.ts
ecklf: Fixed

packages/cli/src/util/dev/builder-worker.cjs
ecklf: We do clean up the zip when we deserialize the IPC message via `unlinkSync` here, so the zip itself is very short-lived and only ever exists if we can't handle things in-memory via IPC.
https://github.com/vercel/vercel/pull/15260/changes#diff-07e93ba9f6a897953dde3558cdc68aff1a482544c0130b27a11c5f930b6ea65aR117-R128

There are edge-cases like the process being killed or the IPC failing, but I don't think we do something similar for `@vercel/fun` either in that scenario.

packages/node/src/awaiter.ts
```diff
@@ -15,16 +15,16 @@ export class Awaiter {
     );
 
   public waitUntil = (promise: Promise<unknown>) => {
-    this.promises.add(promise);
+    // Attach an error handler immediately to prevent unhandled promise rejections
+    // while still tracking the promise for awaiting later
+    const handledPromise = Promise.resolve(promise).catch(this.onError);
```
tknickman, line 20: Thank you so much

tknickman (approved):

PR #15277, opened 2026-02-26
[next] fix: lower max uncompressed lambda size to leave headroom for primitives

ecklf:
We can run into the following error on Vercel deployments.

> [vc] (patchBuild) ungraceful error was thrown: A Serverless Function has exceeded the unzipped maximum size of 250 MB.

This happens because other Lambda primitives like layers add to the 250MB limit.

Fix: Lowering the limit to 240MB for bundling so we leave some headroom.

> [!NOTE]
> Low Risk Change
>
> This PR reduces a Lambda size limit constant from 250MB to 240MB and updates corresponding test assertions, which is a defensive change that tightens constraints.
> 
> - Constant change: reduces DEFAULT_MAX_UNCOMPRESSED_LAMBDA_SIZE from 250 to 240 MiB
> - Test updates: adjusts expected warning/error messages to reflect new limit
> - Changeset added for patch release
>
> <sup>Risk assessment for [commit 8720b4f](https://github.com/vercel/vercel/commit/8720b4f874496c996a29e535dafd4f5576c72e41).</sup>

### DIFF
--- a/.changeset/odd-trainers-greet.md
+++ b/.changeset/odd-trainers-greet.md
@@ -1,0 +1,5 @@
+---
+'@vercel/next': patch
+---
+
+lower max uncompressed lambda size to leave headroom for primitives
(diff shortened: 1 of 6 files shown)


smaeda-ks (changes requested):
We already do this via LAMBDA_RESERVED_UNCOMPRESSED_SIZE

PR #15295, opened 2026-02-27
Revert "[python] run build command after install commands (#15171)"

msullivan:
This reverts commit 90238372d2bd0df7b9c0b31331106c242179d526.

This caused CI breakage.

> [!WARNING]
> High Risk Change
>
> This revert changes the execution order of Python build commands - moving them to run BEFORE dependencies are installed instead of AFTER, which is a significant change to build logic that could cause build failures or unexpected behavior.
> 
> - Build command execution moved from after dependency install to before
> - Environment changed from pythonEnv (venv) to spawnEnv for build command execution
> - Reverts CI-breaking change per PR description
>
> <sup>Risk assessment for [commit 840a2c0](https://github.com/vercel/vercel/commit/840a2c03a6f3967faa3efa71597a15af642dd18c).</sup>

### DIFF
--- a/.changeset/long-dots-brake.md
+++ b/.changeset/long-dots-brake.md
@@ -1,5 +1,0 @@
----
-'@vercel/python': minor
----
-
-Run Python build commands _after_ install commands and in the virtual env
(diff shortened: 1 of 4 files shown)


msullivan:
nevermind we're good

PR #15300, opened 2026-02-27
do not require auth for flags prepare subcommand

dferber90:
The `vercel flags prepare` command doesn't require auth since it's reading the env vars for a flags sdk key and uses that to fetch the flags definitions.

> [!WARNING]
> High Risk Change
>
> This PR adds the `flags prepare` subcommand to the list of commands that can run without authentication, which is an intentional auth bypass for a specific CLI command that uses SDK keys from environment variables instead of user tokens.
> 
> - Auth bypass: adds `flags prepare` to `subcommandsWithoutToken` list
> - Tests verify command executes without credentials and emits expected module files
>
> <sup>Risk assessment for [commit 5cf4a96](https://github.com/vercel/vercel/commit/5cf4a96af2894672c78d396a38079cef6b91d80b).</sup>


luismeyer (approved):

vvo (approved):
Looks good! Clean change that follows existing patterns.

Two minor suggestions:
- Consider asserting the test also checks for expected behavior (e.g. command actually progresses past auth), not just the absence of the credentials error — otherwise a silent failure could pass the test.
- The test is missing a trailing semicolon on the last line, unlike the rest of the file.

dferber90:
@vvo thanks for the note about tests, added happy path tests in 061b748ea1f3f649b685c6d5fe5c36340cd576c3

PR #15309, opened 2026-02-27
ci: check out PR commit instead of merge commit in test workflows

msullivan:
This should result in substantially reduced frequency of needing to
rebase on master to get tests to pass.

The default behavior of github actions on pull requests is to checkout
a temporary merge commit against the current HEAD of the target branch
(typically main).

This has pluses and minuses, but it interacts poorly with vercel PR
previews, which build based on the actual PR commit.

Our tests wait for the vercel preview and download the built CLI from
it, and use that for running tests. This is *not* the same CLI as
would be built from the merge commit that the tests run against.

In particular, this means that if a PR is merged that adds a new
feature/bugfix and a test that exercises it, this will often cause
other PRs that aren't rebased on it yet to have spurious test
failures, since the new tests run (they are in the merge commit) but
the CLI being tested doesn't incorporate the changes (because it was
built just from the PR branch).

Fix this by making all of the github actions checkouts check out the
PR commit instead of using the default behavior of a merge commit.
Not *all* of the actions are affected by the problem, since not all
use the vercel build (`test.yml` and `test-e2e.yml` are the culprits),
but I changed it for every action for consistency.

The potential downside here is that *some* projection against bad
merges are lost, since the tests run against potentially an older main
branch. But honestly I don't think *that* much protection is lost,
because even then, it's still not running against `main` at the time
of the merge.

> [!NOTE]
> Low Risk Change
>
> This PR modifies GitHub Actions workflow files to change checkout behavior from merge commit to PR commit, which is a CI configuration change with no impact on application code, security, or database structure.
> 
> - CI: adds `ref: ${{ github.event.pull_request.head.sha }}` to checkout steps across 6 workflow files
> - Adds new `lint-actions` job to enforce checkout ref convention
> - No changes to application code, auth, or data handling
>
> <sup>Risk assessment for [commit 0f030f6](https://github.com/vercel/vercel/commit/0f030f697049b58a418909d880b362459db0ffbb).</sup>


msullivan:
For even more fun, see this failure: https://github.com/vercel/vercel/actions/runs/22503525928/job/65196383032?pr=15309
In which it is demonstrated that github gets its actions configs from the temporary merge commit, even though if you click "Workflow file" it shows just the version from the PR commit!

msullivan:
Here is a demonstration project I made of the skew between actions and vercel: https://github.com/uncurated-tests/test-actions-skew/pull/1

ricardo-agz (approved):

jeffsee55 (approved):

